PR #4853, opened 2022-12-28
Use WAL journal mode for briefcases

kabentley:


pmconne (approved):

grigasp:
@kabentley, shouldn't we enable WAL mode for `StandaloneDb` too?

PR #4854, opened 2022-12-29
Remove FeatureGates API

pmconne:
It was marked `@internal` and doesn't make any sense as part of the core libraries - only an app can decide how/if to implement feature-gating.
Removed its usage from one example (example code should not use internal APIs).
@jason-crow I want to assume apps like yours did not use this internal API as part of their feature-gating implementation.


calebmshafer (approved):

jason-crow:
> It was marked `@internal` and doesn't make any sense as part of the core libraries - only an app can decide how/if to implement feature-gating. Removed its usage from one example (example code should not use internal APIs). @jason-crow I want to assume apps like yours did not use this internal API as part of their feature-gating implementation.

That is correct

pmconne:
sigh.

HIGH Security Vulnerability: Prototype Pollution in JSON5 via Parse Method in json5 (from ..__..__test-apps__appui-test-app__connected>@bentley/react-scripts>eslint-config-react-app>eslint-plugin-import>tsconfig-paths>json5).  See https://github.com/advisories/GHSA-9c47-m6qq-7p4h for more info.

calebmshafer:
@pmconne the jsonwebtoken one should now be fixed in the latest version of the authorization packages. I'm away from my computer at the moment but can update later.

As for the react-scripts one, sigh... Let's just add it to the ignore list in `tools/internal/scripts/rush/audit.js`.

pmconne:
> the jsonwebtoken one should now be fixed in the latest version of the authorization packages.

It is, but azurite has same issue and they [don't appear interested](https://github.com/Azure/Azurite/pull/1777) in addressing it.

PR #4862, opened 2023-01-03
Added getOriginPropertyName() method to column info

rschili:
Method allows getting the backing property name, if any. It's different from getPropertyName as it does still return the property name even if an alias is used for the column.
Native addon changes are already live and included in the latest build that we consume.

### DIFF
--- a/core/backend/src/ECSqlStatement.ts
+++ b/core/backend/src/ECSqlStatement.ts
@@ -689,6 +689,13 @@ export interface ECSqlColumnInfo {
    */
   getPropertyName(): string;
 
+  /** Gets the name of the original property that the column data is from.
+   * > Other than getPropertyName(), this ignores aliases and allows getting the name
+   * > of the property which is being used for the column. A column may not be backed
+   * > by a property, in which case this returns undefined.
+   */
+  getOriginPropertyName(): string | undefined;
+
   /** Gets the full access string to the corresponding ECSqlValue starting from the root class.
    * > If this column is backed by a generated property, i.e. it represents ECSQL expression,
    * > the access string consists of the ECSQL expression.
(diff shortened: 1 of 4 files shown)


core/backend/src/ECSqlStatement.ts
```diff
@@ -689,6 +689,13 @@ export interface ECSqlColumnInfo {
    */
   getPropertyName(): string;
 
+  /** Gets the name of the original property that the column data is from.
+   * > Other than getPropertyName(), this ignores aliases and allows getting the name
```
pmconne, line 693: Does `>` have some special meaning here and elsewhere?

pmconne, line 693: Use links for referenced APIs: `[[getPropertyName]]`. Consider updating documentation for `getPropertyName` to explain how it is affected by aliases etc.

pmconne (approved):

core/backend/src/ECSqlStatement.ts
rschili: vscode indents that portion in the tooltip.
image

pmconne: @williamkbentley what happens to these `>`s on the docs site? Do they show up verbatim? Indent? Something else?

rschili: Looks like it's turning into annotations on the website. https://www.itwinjs.org/reference/core-backend/ecsql/ecsqlcolumninfo/getpropertyname/

pmconne: Thanks, so typical markdown behavior.
I don't see much reason to do that vs just having multiple lines of documentation like we do elsewhere.
If something should be called out as, e.g., an important deviation from expected behavior, the `@note` tag is appropriate. I don't think it's needed here.

PR #4866, opened 2023-01-03
Document package and API versioning policies.

pmconne:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/391.
Incidentally fix bug in README: you can't successfully `rush cover` without first `rush build:ci`-ing.

TODO
- [x] Integrate into docs
- [x] Guidelines for iTwin.js devs on how to handle breaking changes, incl when reviewing PRs
- [x] Clean up docs/learning/guidelines/release-tags-guidelines.md


Versioning.md
```diff
@@ -0,0 +1,38 @@
...
+In practice, `@beta` and `@public` APIs are typically designed in a way that promotes flexibility in the future - e.g., by accepting their arguments as "options" object types. This reduces the risk of breaking changes and deprecation.
+
+## Package versioning policy
+
+The individual packages that constitute the iTwin.js libraries use [semantic versioning](./docs/learning/guidelines/semantic-versioning-guidelines.md) to convey information about how their public APIs change over time.
+
+- Patch releases occur frequently to fix bugs or security issues and have no impact on the public API.
+- Minor releases occur on roughly six- to eight-week intervals. They may introduce new APIs, but never break any existing `@public` APIs.
+- Major releases occur only as frequently as necessary. They can include breaking changes to `@public` APIs, but only according to the [deprecation policy](#api-deprecation-policy).
+
```
pmconne, line 33: @calebmshafer I would like to link to (or include?) the LTS major release support policy here.

pmconne: @calebmshafer separate PR?

pmconne: Found https://github.com/iTwin/itwinjs-backlog/issues/391 where you were working on LTS policy, I guess I will fold it in here...

Versioning.md
```diff
@@ -0,0 +1,38 @@
...
+Occasionally, a `@beta` or `@alpha` API may be abandoned rather than being promoted. Not all experiments are successful.
+
+In practice, `@beta` and `@public` APIs are typically designed in a way that promotes flexibility in the future - e.g., by accepting their arguments as "options" object types. This reduces the risk of breaking changes and deprecation.
+
+## Package versioning policy
+
+The individual packages that constitute the iTwin.js libraries use [semantic versioning](./docs/learning/guidelines/semantic-versioning-guidelines.md) to convey information about how their public APIs change over time.
+
+- Patch releases occur frequently to fix bugs or security issues and have no impact on the public API.
+- Minor releases occur on roughly six- to eight-week intervals. They may introduce new APIs, but never break any existing `@public` APIs.
```
calebmshafer, line 31: Can you remove the time intervals into this document?

README.md
```diff
@@ -31,7 +31,7 @@ Each package has its own **node_modules** directory that contains symbolic links
 1. Clone repository (first time) with `git clone` or pull updates to the repository (subsequent times) with `git pull`
 2. Install dependencies: `rush install`
 3. Clean: `rush clean`
-4. Rebuild source: `rush rebuild`
+4. Build source: `rush build:ci` (or `rush build` if you don't intend to run the tests).
```
aruniverse, line 34: thx paul

Versioning.md
```diff
@@ -0,0 +1,38 @@
...
+iTwin.js employs versioning policies to faciliate the following objectives:
+
+- Provide a broad set of stable, well-supported APIs.
+- Continuously deliver new features, bug fixes, and security updates.
+- Evolve APIs over time in response to user feedback.
+- Encourage users to keep up to date with the most recent versions.
+
+## API promotion process
+
+Each API exported by a package is marked with a [release tag](./docs/learning/guidelines/release-tags-guidelines.md) indicating its level of stability. A package's "public" API comprises all of its exported APIs that have been marked with the `@public` release tag. All other APIs are considered "non-public". The [package versioning](#package-versioning-policy) and [API deprecation](#api-deprecation-policy) policies apply only the a package's public API.
```
calebmshafer, line 12: What's the classification you're giving "non-public" here? I feel that just confuses things when you talk about `internal` below.

Versioning.md
pmconne: IMO semantic-versioning-guidelines.md should go away entirely. It mostly just restates how semver works - a link to somebody else's description would suffice.

release-tags-guidelines.md should tell contributors to iTwin.js how to use release tags. The info in Versioning.md is targeted at consumers of iTwin.js, not its developers. That's why it's not in the docs directory.

Are you asking me to clean both of those other documents up as part of this PR; or, remove references to them from this document; or something else?

Versioning.md
pmconne: Bye.

Versioning.md
pmconne: "non-public" = not "@public".
I will clean this up to describe each tag in a consumer-facing manner.
I'd prefer not to talk about internal and alpha at all here, since they are omitted from the docs. But they're going to show up in auto-complete in IDEs so probably need to be defined.

Versioning.md
calebmshafer: > Are you asking me to clean both of those other documents up as part of this PR; or, remove references to them from this document; or something else?

I was referring to removing the semantic versioning one. Just pushed an update to do that.

> The info in Versioning.md is targeted at consumers of iTwin.js, not its developers. That's why it's not in the docs directory.

If that's the goal, I think this one is too developer focused. The other one I'm working on that's the support policy would be the high level item that both dev and non-dev would be looking at IMO. In both cases, I feel like we should have them in our deployed docs site as well as referenced in the README.

pmconne: > I think this one is too developer focused

If you want me to do something about that I'll need you to elaborate. ("Consumers" here means, developers who **use** iTwin.js, but do not **contribute** to iTwin.js).

calebmshafer: Yes, I meant the developer using iTwin.js, not developing it, as well. If that's the goal of this document why wouldn't it be on our doc site? I think it'd be helpful to have it there and then also keep the link that you added in our main README.

Versioning.md
calebmshafer: when you did the investigation into the APIs, how many were `alpha`?

I feel that it isn't used properly (i.e an excuse not to write docs) and based on your description below could really just be removed and instead we only use `internal` `beta` and `public` as the main release tags.

> "non-public" = not "@public".

I don't think that clarification is necessary and that sentence doesn't add much so could be removed.

Versioning.md
```diff
@@ -0,0 +1,38 @@
...
+
+Occasionally, a `@beta` or `@alpha` API may be abandoned rather than being promoted. Not all experiments are successful.
+
+In practice, `@beta` and `@public` APIs are typically designed in a way that promotes flexibility in the future - e.g., by accepting their arguments as "options" object types. This reduces the risk of breaking changes and deprecation.
+
+## Package versioning policy
+
+The individual packages that constitute the iTwin.js libraries use [semantic versioning](https://semver.org/) to convey information about how their public APIs change over time.
+
+- Patch releases occur frequently to fix bugs or security issues and have no impact on the public API.
```
calebmshafer, line 30: What do you think of having the description below rather than trying to convey the frequency of a particular release type?  I had started out with something similar to what you have but felt that the clarification on each was unneeded as the time is predictable and therefore not too useful to call out.


    MAJOR: Potentially contains backwards incompatible changes. A public API will not change in a backwards incompatible way unless it has been marked as deprecated for at minimum 1 major version. See [release tags](https://github.com/iTwin/itwinjs-backlog/issues/391#release-tags) for details on how to determine if an API is public.
    MINOR: Backwards compatible features, bug fixes and other enhancements to public APIs. However, in some cases beta and alpha APIs are subject to change. See [release tags](https://github.com/iTwin/itwinjs-backlog/issues/391#release-tags) for details on compatibility and how to determine if an API is public.
    PATCH: Backwards compatible bug fixes and security updates.


However we could remove the talk about release tags and their support in this case to merge it with yours. Something like,

    MAJOR: Potentially contains backwards incompatible changes. A public API will not change in a backwards incompatible way unless it has been marked as deprecated according to our `deprecation policy`.
    MINOR: Backwards compatible features, bug fixes and other enhancements to public APIs. May introduce new APIs, but never break any existing `@public` APIs.
    PATCH: Backwards compatible bug fixes and security updates. No impact on the `@public` API.

Versioning.md
pmconne: There are 774 occurrences of `@alpha` in common/api. I would like to see that number reduced significantly as we reevaluate our current API tags, but that effort is going to be much bigger than the scope of this PR.

I don't recall why extensions use `@preview` instead of `@beta` - do you? Was there some case in which some API would be `@preview` but not also `@beta`?

Versioning.md
pmconne: Sure. I mainly want to convey that major releases are infrequent occurrences to assuage fears that we'll constantly be breaking APIs.

I'm moving this file to docs/learning and reworking it based on feedback.

Versioning.md
calebmshafer: > I don't recall why extensions use @preview instead of @beta - do you? Was there some case in which some API would be @preview but not also @beta?

We separated them because we wanted to leave the potential for a `@public` API to be added to the Extension API as `@preview` in case we weren't sure we wanted to keep it in there. However, we have nothing at all in preview today and I'm re-thinking my stance on that now... Feels like it's an unnecessary complication to add onto an already confusing distinction between the two APIs. 

I still think we need a "preview" Extension API but maybe that should be for things tagged `@beta`.

Versioning.md
```diff
@@ -0,0 +1,38 @@
...
+- Minor releases occur less frequently than patch releases. They may introduce new APIs, but never break any existing `@public` APIs.
+- Major releases occur only as frequently as necessary. They can include breaking changes to `@public` APIs, but only according to the [deprecation policy](#api-deprecation-policy).
+
+## API deprecation policy
+
+Most `@public` APIs are intended to remain stable indefinitely, but occasionally a breaking change is required to fix a bug, introduce new functionality, or address performance problems. The deprecation policy ensures that such changes are introduced deliberately and predictably to help users plan ahead.
+
+Before an API can be removed it must first be marked as `@deprecated` in version `N` of the package. It must remain stable for the duration of version `N+1`. Finally, in version `N+2` it *may* be removed. This means that, for example, an API deprecated in version 3.1 of a package cannot be removed until version 5.0 at the earliest.
```
NancyMcCallB, line 38: If we're going to begin enforcing this in V4, perhaps we should use 4.1 and 6.0 in this example.

pmconne: Do you intend to remove ui-related APIs in 4.x that were deprecated in 3.x?

aruniverse: would it be better to just say `major version N` ? and not provide an explicit example?

pmconne: I would like to make it crystal clear to users when they can expect this policy to start being enforced, hence my question to @NancyMcCallB.

calebmshafer: @pmconne yes there is a plan to do it. The extent to which we do is yet to be determined. We're taking a look at the APIs now.

Also, we should talk with @grigasp I know he had some APIs he wanted to remove that are deprecated and not sure how long they've been that way.

calebmshafer: I agree with @NancyMcCallB, we should only start enforcing as of 4.x and onward. We were not clear previously.

However, this does not mean that we should go and add deprecated to everything we'd like to remove in 4.0 right now. We still need to avoid breaking things unnecessarily that will impact people.

pmconne: > We still need to avoid breaking things unnecessarily that will impact people.

Damn straight.

> this does not mean that we should go and add deprecated to everything we'd like to remove in 4.0

Maybe I misunderstand but if you are planning to remove something in 4.0, please deprecate it prior to that.

calebmshafer: @pmconne  sorry, wasn't clear. I meant that we should avoid adding anything else as deprecated in 3.x that we plan to remove in 4.0, i.e. we should be done trying to remove things in 4.0 that aren't already marked.

If you add deprecated now then you will need to remove it at the earliest in 5.0. Let me get the list together and see if that's the case. We should track all of this for the release.

grigasp: We have almost no deprecated APIs that we carried over from 2.x - all deprecated ones we currently have got deprecated in the 3.x timeframe. As much as I'd like to get rid of all of them, there's no strong reasons to remove them in 4.x except for one case - the APIs related to the `Table` component, which is going to be removed completely and there's no way for us to keep anything related to it.

Also, I'd like to suggest a couple of things documentation wise:

- When adding `@deprecated` tag, we should also specify the version it was deprecated in. Knowing that an API may be removed in version `N+1` doesn't help if you don't know the value of `N`. This can be found using `git`, but is really inconvenient.

- We should be clear on what we consider a breaking change. Here're a couple types of breaking changes (technically) that I believe we're making from time to time in our non-major releases:

  - Adding optional attributes to interfaces that are used as data structures. In theory, [every change to a public interface is a breaking change](https://log.havrlant.cz/interface-breaking-change/). I think such breaking changes should be allowed even in patches.
 
  - Upgrading peer dependency version even if it's just a patch release change. In theory, peer dependencies are considered part of the public API. See more details [here](https://github.com/semver/semver/issues/502). At the moment we have the packages all locked at the same version - increasing patch version of `core-bentley`, which is a peer dependency of nearly all other package in the repo, doesn't cause other packages' major version change. And that's technically violating the rule... I think this is going to become a bigger problem when UI and some Presentation packages move out of the core repo and start being versioned separately. Increasing major version of `presentation-components` with every patch release of `presentation-common` would cause a huge number of major releases, which we want to avoid. Either we have to keep violating the rule, or need to convert peer deps to regular deps by duplicating some of the interfaces.

calebmshafer: > Adding optional attributes to interfaces that are used as data structures. In theory, [every change to a public interface is a breaking change](https://log.havrlant.cz/interface-breaking-change/). I think this specific kind of breaking change should be allowed even in patches.

@grigasp , definitely agree with this one. I'd say this is fine as long as the API is still backwards compatible.

> Upgrading peer dependency version even if it's just a patch release change

Good point on this one, it is definitely confusing especially as we get more packages in separate places. We already handle this two different ways today when considering all of iTwin.js. The first way is what you mentioned for the itwinjs-core packages, every release we bump the version of peer dependencies for every other package in the repo (i.e. if they're in a lockstep release policy).  The second way, any time the minimum of a peer dependency is changed we require a major version bump to change as we consider that a breaking change (e.g. this is how we handle the itwinjs-core versions in the iTwin Viewer, every time we update the minimum core pkg versions we make a major version bump of the iTwin Viewer packages).

The general, unwritten, rule that we have been following is that anything that is within a lockstep release policy (i.e. itwinjs-core packages) can freely move up their peer dependency as long as that peer dependency is also within the lockstep policy. This ensures consistent versions and since the packages are all released exactly at the same time you need to do this action once to update them all.

When a package is outside of the release policy of it's peer dependency, it must follow the "second way". The main reason is that with them being on a separate version policy, and released independently, the coordination between the two packages is much more loose.  This does normally force the package that has a peer dependency to support a range of versions most times and I do always suggest when you're breaking this boundary you should always be using our "public" API. This is how we handle everything outside of the itwinjs-core repo.

So as an example, with the AppUI packages going to be removed from both the repo and the lockstep versioning of the rest of itwinjs-core, those packages will no longer be able to update the minimum of itwinjs core packages freely and instead be required to follow the "second way".  However, since they are moving into a repo of their own with its own lockstep versioning all of the AppUI packages will be released in lockstep, therefore subject to the "first way".

note: ideally we'd have independence between all of our packages and not require such a tight coupling but with the way we're architected and our extensive use of statics and the objects as parameters to methods it forces us to use peer dependencies quite a bit. We're in essence providing a "set" of packages in some cases rather than individual ones.

pmconne:
TypeScript's flexible, structural type system and its compiler's ability to reason about types imposes pretty strict constraints on what we can call a "non-breaking" change. Union types are a prime example given tsc's default rules about unreachable code (which can only be disabled globally).

Plenty more examples [here](https://www.semver-ts.org/).

pmconne:
@calebmshafer I can't find much info about how VS Code handles breaking changes in its extension API. The API doesn't appear to be versioned. They ["give our best effort to avoid breaking API changes"](https://code.visualstudio.com/api/advanced-topics/using-proposed-api) and ["never want to break existing extensions"](https://code.visualstudio.com/updates/v1_73#_proposed-apis).

They call their beta APIs "proposed APIs" available only in Insider builds and allegedly not permitted to be used in published extensions. These APIs may have breaking changes in new releases.

docs/learning/api-support-policies.md
```diff
@@ -0,0 +1,91 @@
...
+
+Only "public" and "beta" APIs are included in the [published documentation](https://www.itwinjs.org/reference/). An API typically starts out as "beta". It may evolve rapidly in response to feedback before stabilizing and being promoted to "public". On occasion, a "beta" API may be abandoned - not all experiments succeed.
+
+The "public API" of a package comprises the set of all APIs it contains that are marked with the "public" release tag. The package's public API enjoys stability guarantees provided by the [package versioning policy](#package-versioning-policy).
+
+## Package versioning policy
+
+The individual packages that constitute the iTwin.js libraries use [semantic versioning](https://semver.org/) to convey information about how their public APIs change over time.
```
tcobbs-bentley, line 56: I would suggest adding "(symver)" after the symantic versioning link. The term is used later in the document, and without stating that "symantic versioning" is abbreviated "symver", to me that later use of "symver" is confusing. In other words, change this sentence to the following:

> The individual packages that constitute the iTwin.js libraries use [semantic versioning](https://semver.org/) (symver) to convey information about how their public APIs change over time.

pmconne: added.

calebmshafer (approved):
This looks good to me now.

Versioning.md
pmconne: Changed to 4.1.

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -165,22 +50,11 @@ Likewise, *deprecated* API items should include documentation that indicates wha
...
- */
-```
-
-*Beta* and *alpha* API items should include a short reason for that classification.
-
-```ts
-/** Sample description of a beta API item.
- * @beta Comment describing reason API item is beta that will be included in the public SDK documentation.
+ * @public
+ * @deprecated Comment describing reason API item is deprecated and what should be used instead.
```
jbelevicius, line 54: In the "API Categories" it is stated that "deprecated tag is typically accompanied by information about the package version in which it became deprecated" - this should be noted here as well I suppose.

pmconne: True. It's currently an aspirational lie though - nobody has ever actually included version information in a `@deprecated` tag. @aruniverse suggested we start doing that. Perhaps we need to write a script that will add "3.x" to anything currently tagged as deprecated.

PR #4870, opened 2023-01-04
Use WebView instead of MobileFrontend since it has been removed

toddsouthenbentley:
Adapt to changes made in the add-on's java API.

Re-enabled the onResume/onPause handlers as they are working again.

### DIFF
--- a/test-apps/display-test-app/android/imodeljs-test-app/app/src/main/java/com/bentley/imodeljs_test_app/MainActivity.kt
+++ b/test-apps/display-test-app/android/imodeljs-test-app/app/src/main/java/com/bentley/imodeljs_test_app/MainActivity.kt
@@ -21,7 +21,6 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.webkit.WebViewAssetLoader
 import androidx.webkit.WebViewClientCompat
 import com.bentley.itwin.IModelJsHost
-import com.bentley.itwin.MobileFrontend
 import java.io.File
 import java.io.IOException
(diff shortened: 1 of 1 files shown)


test-apps/display-test-app/android/imodeljs-test-app/app/src/main/java/com/bentley/imodeljs_test_app/MainActivity.kt
```diff
@@ -192,9 +185,9 @@ class MainActivity : AppCompatActivity() {
                 }
             }, "DTA_Android")
 
-        host.setFrontend(frontend)
-        setContentView(frontend)
-        frontend.loadEntryPoint()
+        host.webView = webView
+        setContentView(webView)
+        host.loadEntryPoint(env.optStringNotEmpty("IMJS_DEBUG_URL") ?: "https://${WebViewAssetLoader.DEFAULT_DOMAIN}/assets/frontend/index.html", args)
```
tcobbs-bentley, line 190: I think it would be cleaner and less confusing if the creation and initialization of the args variable was moved down to be right above this line. As it is, args gets created an initialized in the spot it was used before. Since it is no longer used before here, I think that should move here.

toddsouthenbentley: Fixed

tcobbs-bentley (approved):

PR #4884, opened 2023-01-06
Fix/ext gen

aruniverse:


pmconne:
Correct me if I am wrong about any of the following:
If you add `@extensions` to some core-frontend API (or core-common, but core-frontend for example purposes), you must build and run extract-api for core-extensions.
Previously, if you did not, then CI would run `rush extract-api` and produce differences but those differences would be ignored.
Now, the differences will be correctly detected, failing the PR.

If this is correct, can you adjust the error message you're producing to indicate what needs to be done to resolve it (run extract-api for core-extensions)?

aruniverse:
> Correct me if I am wrong about any of the following: If you add `@extensions` to some core-frontend API (or core-common, but core-frontend for example purposes), you must build and run extract-api for core-extensions. Previously, if you did not, then CI would run `rush extract-api` and produce differences but those differences would be ignored. Now, the differences will be correctly detected, failing the PR.
> 
> If this is correct, can you adjust the error message you're producing to indicate what needs to be done to resolve it (run extract-api for core-extensions)?

Paul you are correct; updated the message to be more hopeful.

pmconne (approved):

johnnyd710 (approved):

calebmshafer (approved):

PR #4893, opened 2023-01-09
Updates to content of the BIS guide

diegoalexdiaz:


docs/bis/guide/fundamentals/federationGuids.md
```diff
@@ -0,0 +1,25 @@
+# FederationGuids
+
+Every BIS `Element` has an optional 128 bit [Globally Unique Identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier) called `FederationGuid`. It is really identifying the real-world Entity that the Element represents, not the Element itself. Generally it is intended that FederationGuids are assigned by external systems to *federate* Elements to their external meaning.
+
+## Scope of Uniqueness
+
+Assigned FederationGuids are enforced to be unique on a single BIS Repository. However, multiple BIS Repositories under the same iTwin can use the same FederationGuid in order to refer to the same real-world Entity.
```
ColinKerr, line 7: do we need 'under the same iTwin' here ... if by iTwin we mean context then it is feasible that the same entity may exist in iModels spanning multiple contexts.  I think we can just drop this bit of the sentence because scoping it within one iTwin does not add much over saying 'multiple BIS repositories'

docs/bis/leftNav.md
```diff
@@ -118,6 +159,16 @@ $('#guide-list').slideToggle()
 }
 });
 </script>
+<script>
+$("#domain-schemas").append(function () {
```
yashincontrol, line 163: Please change this to `$("#domain\\ schemas")`. Escaping the space will allow jQuery to select the "domain schemas" id and add a chevron.

ben-polinsky: This is not correct.  HTML Element ids cannot contain spaces, and multiple ids are not supported. We should fix this on the bemetalsmith end, if we're creating an invalid id.

yashincontrol: @diegoalexdiaz We will be releasing a new version of BeMetalsmith soon. It fixes the heading ID issues.

ColinKerr (approved):

Issue #3777, opened 2022-06-10
Unable to hilite elements under grouping nodes when a grouping node is selected


grigasp:
**Repro:**

1. Select some grouping node in a tree
2. Request hilite list - it should return all ECInstances under that grouping node, but doesn't

**Investigation notes:**

When doing Node -> ECInstance lookup, our locater uses the 'current' ruleset which is the 'HiliteList' ruleset. The node was created with a different ruleset, so it's not found for the hilite list one. If we don't find the node, then we don't know what instances it groups...

PR #4898, opened 2023-01-10
Presentation: Hilite grouping nodes

BernardasTe:
Resolves https://github.com/iTwin/itwinjs-core/issues/3777

Related to https://github.com/iTwin/imodel-native/pull/101

• Added InstanceKeysSelectQuery to node key;
• Added and updated tests.


presentation/common/src/presentation-common/hierarchy/Key.ts
```diff
@@ -158,6 +158,9 @@ export interface BaseNodeKey {
 
   /** Node hash path from root to the node whose key this is */
   pathFromRoot: string[];
+
+  /** Query that returns all selected instance keys */
+  instanceKeysSelectQuery: InstanceKeysSelectQuery;
```
grigasp, line 163: shouldn't `instanceKeysSelectQuery` be optional?

presentation/common/src/presentation-common/hierarchy/Key.ts
```diff
@@ -274,3 +278,23 @@ export interface LabelGroupingNodeKeyJSON extends GroupingNodeKeyJSON {
  * @public
  */
 export type NodeKeyJSON = BaseNodeKeyJSON | ECInstancesNodeKeyJSON | ECClassGroupingNodeKeyJSON | ECPropertyGroupingNodeKeyJSON | LabelGroupingNodeKeyJSON;
+
+/**
+ * Data structure that describes an instance keys select query
+ * @beta
+ */
+export interface InstanceKeysSelectQuery {
```
grigasp, line 286: While I agree that at the moment instance keys select query is the only type of query we have, I don't see any reason the interface should be named under this use case. Query string + bindings that are stored in this object can be used for any query, thus the interface should have a more generic name.

presentation/common/src/presentation-common/hierarchy/Key.ts
```diff
@@ -274,3 +278,23 @@ export interface LabelGroupingNodeKeyJSON extends GroupingNodeKeyJSON {
...
+  /** The query to select instance keys */
+  query: string;
+  /** The query parameters */
+  bindings: string[];
```
grigasp, line 290: `bindings` aren't storing strings, are they?

presentation/common/src/presentation-common/hierarchy/Key.ts
```diff
@@ -274,3 +278,23 @@ export interface LabelGroupingNodeKeyJSON extends GroupingNodeKeyJSON {
...
+}
+
+/**
+ * Serialized [[InstanceKeysSelectQuery]] JSON representation.
+ * @beta
+ */
+export interface InstanceKeysSelectQueryJSON {
```
grigasp, line 297: is there a reason to have this interface? member types completely match `InstanceKeysSelectQuery`..

presentation/common/src/test/Update.test.ts
```diff
@@ -15,6 +15,7 @@ const testNode: Node = {
     instanceKeys: [],
     pathFromRoot: [],
     type: StandardNodeTypes.ECInstancesNode,
+    instanceKeysSelectQuery: { query: "", bindings: [] },
```
grigasp, line 18: `instanceKeysSelectQuery` should be optional and these should be unnecessary

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
```diff
@@ -185,6 +188,172 @@ describe("Unified Selection", () => {
...
+          code: elemCode,
+        };
+        return builder.insertElement(physicalObjectProps);
+      };
+
+      const objectIdArray = Array<Id64String>();
+      let groupingNodesIModel: IModelConnection;
+      beforeEach(async () => {
+        groupingNodesIModel = await buildTestIModel("GroupByClass", (builder) => {
+          const physicalPartitionId = addPartition("BisCore:PhysicalPartition", builder, "TestDrawingModel");
```
grigasp, line 241: the name `"TestDrawingModel"` doesn't match type of the model

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
```diff
@@ -185,6 +188,172 @@ describe("Unified Selection", () => {
...
+              arePolymorphic: true,
+              instanceFilter: "this.Parent = NULL",
+              groupByClass: true,
+              groupByLabel: false,
+            }],
+          }],
+        };
+
+        await using<RegisteredRuleset, Promise<void>>(await Presentation.presentation.rulesets().add(groupByClassRuleset), async () => {
+          const rootNodes = await Presentation.presentation.getNodes({ imodel: groupingNodesIModel, rulesetOrId: groupByClassRuleset.id });
```
grigasp, line 279: you don't need to call `Presentation.presentation.rulesets().add(groupByClassRuleset)` if you pass ruleset instead of its id to the `getNodes` call

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
```diff
@@ -185,6 +188,172 @@ describe("Unified Selection", () => {
...
+      const addPhysicalObject = (builder: TestIModelBuilder, modelId: string, categoryId: string, elemCode = Code.createEmpty()) => {
+
+        const physicalObjectProps: PhysicalElementProps = {
+          classFullName: "Generic:PhysicalObject",
+          model: modelId,
+          category: categoryId,
+          code: elemCode,
+        };
+        return builder.insertElement(physicalObjectProps);
+      };
```
grigasp, line 235: suggestions:
- move these functions to the root scope
- make `builder` the first arg in all of them

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
```diff
@@ -185,6 +188,172 @@ describe("Unified Selection", () => {
...
+          expect(groupingNodesIModel.hilited.elements.hasId(objectIdArray[1])).to.be.true;
+
+          Presentation.selection.replaceSelection("", groupingNodesIModel, [rootNodes[1].key]);
+          await waitForAllAsyncs([handler, Presentation.selection.getToolSelectionSyncHandler(groupingNodesIModel)!]);
+          expect(groupingNodesIModel.hilited.elements.size).to.be.equal(1);
+          expect(groupingNodesIModel.hilited.elements.hasId(objectIdArray[2])).to.be.true;
+        });
+      });
+
+      it.skip("hilites grouping nodes grouped by label", async () => {
```
grigasp, line 327: please add a comment with a link to GH issue related to this `skip`

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
```diff
@@ -185,6 +188,172 @@ describe("Unified Selection", () => {
...
+          category: categoryId,
+          code: elemCode,
+        };
+        return builder.insertElement(physicalObjectProps);
+      };
+
+      const objectIdArray = Array<Id64String>();
+      let groupingNodesIModel: IModelConnection;
+      beforeEach(async () => {
+        groupingNodesIModel = await buildTestIModel("GroupByClass", (builder) => {
```
grigasp, line 240: I don't think sharing the iModel between all these tests is a good idea - that makes it contain stuff that's needed for all the tests rather than what's needed for each individual test. Instead, I think each test should create a minimal iModel itself. 

Also, make sure it's really **minimal** (at the moment you're inserting some models that don't seem to be relevant in any of the tests).

Ideally, the existing tests would also create their own iModel instead of using "Properties_60InstancesWithUrl2". I don't think we want to refactor them in this PR, but maybe just change them so that the iModel is opened in each test individually would be a good start that could be made here too. That would allow us to avoid having two iModels open at the same time (`imodel` & `groupingNodesIModel`)

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
```diff
@@ -185,6 +188,172 @@ describe("Unified Selection", () => {
...
+      });
+
+      afterEach(async () => {
+        await Presentation.presentation.rulesets().clear();
+        objectIdArray.splice(0);
+        await groupingNodesIModel.close();
+        handler.dispose();
+      });
+
+      it("hilites grouping nodes grouped by class", async () => {
```
grigasp, line 262: the test name is interesting... sounds like grouping nodes are grouped by class. IMO, it should be "hilites class grouping node elements". same for other new tests you added.

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
```diff
@@ -185,6 +188,172 @@ describe("Unified Selection", () => {
...
+            }],
+          }],
+        };
+
+        await using<RegisteredRuleset, Promise<void>>(await Presentation.presentation.rulesets().add(groupByClassRuleset), async () => {
+          const rootNodes = await Presentation.presentation.getNodes({ imodel: groupingNodesIModel, rulesetOrId: groupByClassRuleset.id });
+          Presentation.selection.addToSelection("", groupingNodesIModel, [rootNodes[0].key]);
+          await waitForAllAsyncs([handler, Presentation.selection.getToolSelectionSyncHandler(groupingNodesIModel)!]);
+          expect(groupingNodesIModel.hilited.elements.size).to.be.equal(3);
+          objectIdArray.forEach((id) => expect(groupingNodesIModel.hilited.elements.hasId(id)).to.be.true);
```
grigasp, line 283: please rename `objectIdArray` to something more meaningful. "object" is way too generic, it's not clear what objects you have in mind.

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
```diff
@@ -185,6 +188,172 @@ describe("Unified Selection", () => {
...
+
+        await using<RegisteredRuleset, Promise<void>>(await Presentation.presentation.rulesets().add(groupByLabelRuleset), async () => {
+          const rootNodes = await Presentation.presentation.getNodes({ imodel: groupingNodesIModel, rulesetOrId: groupByLabelRuleset.id });
+          Presentation.selection.addToSelection("", groupingNodesIModel, [rootNodes[0].key]);
+          await waitForAllAsyncs([handler, Presentation.selection.getToolSelectionSyncHandler(groupingNodesIModel)!]);
+          expect(groupingNodesIModel.hilited.elements.size).to.be.equal(2);
+          expect(groupingNodesIModel.hilited.elements.hasId(objectIdArray[0])).to.be.true;
+          expect(groupingNodesIModel.hilited.elements.hasId(objectIdArray[1])).to.be.true;
+
+          Presentation.selection.replaceSelection("", groupingNodesIModel, [rootNodes[1].key]);
```
grigasp, line 350: as far as I can tell `rootNodes[1]` is not a label grouping node, so no point to have a test for it here

grigasp (changes requested):

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
BernardasTe: Moved into `describe("Unified Selection"`, or did you mean to move it outside of that as well?
Placed `builder` first.

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
BernardasTe: Removed a couple of objects from the imodels, but not sure if you had something more in mind?
Made each test create its own imodel.

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
BernardasTe: Renamed to TestPhysicalModel

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
BernardasTe: Passed in ruleset and removed `using`.

presentation/common/src/presentation-common/hierarchy/Key.ts
BernardasTe: Made optional.

presentation/common/src/presentation-common/hierarchy/Key.ts
BernardasTe: Renamed to PresentationQuery

presentation/common/src/presentation-common/hierarchy/Key.ts
BernardasTe: Changed to match native serializer side.

presentation/common/src/presentation-common/hierarchy/Key.ts
BernardasTe: Removed.

presentation/common/src/test/Update.test.ts
BernardasTe: Removed.

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
BernardasTe: Added.

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
BernardasTe: Renamed.

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
BernardasTe: Renamed to `physicalObjectIdArray`. Any better suggestions?

full-stack-tests/presentation/src/frontend/UnifiedSelection.test.ts
BernardasTe: Removed.

presentation/common/src/presentation-common/hierarchy/Key.ts
```diff
@@ -274,3 +278,51 @@ export interface LabelGroupingNodeKeyJSON extends GroupingNodeKeyJSON {
...
+  type: "ValueSet";
+  valueType: string;
+  value: any[];
+}
+
+/**
+ * One of the presentation query binding types
+ * @public
+ */
+export type PresentationQueryBinding = IdBinding | IdSetBinding | ECValueBinding | ECValueSetBinding;
```
saskliutas, line 328: Why is this `public` if all types used in union are `alpha`? I think either they all should be `public` or this one should be `alpha`.

BernardasTe: Made `alpha`.

grigasp (approved):

GerardasB (approved):

grigasp:
Trying to revive the license/cla check

PR #4899, opened 2023-01-10
Replace progress function type in core-frontend

GytisCepk:
Right now `ProgressCallback` is used in both `briefcaseConnection.pullChanges` and `NativeApp.requestDownloadBriefcase`.
`ProgressCallback` is internal and it's definition doesn't match what data we return during a download:

```ts
/** @internal */
export interface ProgressInfo {
  percent?: number;
  total?: number;
  loaded: number;
}

/** @internal */
export type ProgressCallback = (progress: ProgressInfo) => void;
```

This PR suggests replacing it with `OnDownloadProgress` :
```ts
/**
 * Download progress information.
 * @public
 */
export interface DownloadProgressInfo {
  /** Bytes downloaded. */
  loaded: number;
  /** Total size of the download in bytes. */
  total: number;
}

/**
 * Called to show progress during a download.
 * @public
 */
export type OnDownloadProgress = (progress: DownloadProgressInfo) => void;
```


pmconne:
> `(loaded: number, total: number)`

If you make this `({ loaded: number; total: number; })` instead, you will 1) prevent people from mixing up the order of the two numeric args and 2) permit it to be more easily extended in future.

core/frontend/src/BriefcaseConnection.ts
```diff
@@ -20,10 +20,16 @@ import { ProgressCallback } from "./request/Request";
 import { disposeTileTreesForGeometricModels } from "./tile/internal";
 import { Viewport } from "./Viewport";
 
+/**
+ * Called to show progress during a download.
+ * @public
+ */
+export type ProgressFunction = (progress: { loaded: number, total: number }) => void;
```
pmconne, line 27: Why not call this `DownloadProgressCallback` since that's what it's for and then you can tell people what the units of `loaded` and `total` are.

GytisCepk: Technically, it could be used for something else as well (e.g. upload).

pmconne: If you think that's likely, perhaps `TransferProgressCallback`.
I am trying to 1) make it obvious that this is a callback that reacts to progress, not a function that produces progress and 2) make it clear to the user what these `loaded` and `total` numbers actually represent.

kabentley: though frankly types are free. Why not just have another one for upload if that comes up? Calling this `OnDownloadProgress` and leaving it `loaded` seems fine to me.

core/frontend/src/BriefcaseConnection.ts
```diff
@@ -37,12 +54,17 @@ export interface GenericAbortSignal {
  * @public
  */
 export interface PullChangesOptions {
-  /** Function called regularly to report progress of changes download. */
+  /**
+   * Function called regularly to report progress of changes download.
+   * @deprecated Use [[downloadProgressCallback]] instead. Deprecated since 3.6
```
aruniverse, line 59: thx for tagging when the deprecation occurred :)

GytisCepk: I was thinking: wouldn't it be more useful to tag when deprecated item will be removed ? That way user wouldn't have to look up our deprecation policy if he doesn't know it already

GytisCepk: User probably cares more about when it will be removed rather when it was deprecated.

aruniverse: I'm going to defer this to @pmconne and @calebmshafer .

My 2cents:
I think it's ok to leave it as is. Just because an API has been deprecated doesnt mean it will ever be removed, although imo we should. It also might be removed past the original N+2 major we deprecated in (N) so don't want to necessarily give them any false expectations.

pmconne: Agree with @aruniverse. Policy only specifies minimum major release in which it might be removed. Also policy could conceivably change in future.

I do want to update guidelines to standardize the tag like "deprecated in [version]. [Suggestions for adjusting existing code]."
` * @deprecated in 3.6. Use [[downloadProgressCallback]] instead.`

pmconne: #4907.

aruniverse:
```suggestion
   * @deprecated in 3.6. Use [[downloadProgressCallback]] instead.
```

pmconne (approved):

pmconne:
Any reason for this to remain in draft state?

core/frontend/src/BriefcaseConnection.ts
GytisCepk: Renamed it to `OnDownloadProgress`.

PR #4897, opened 2023-01-10
Update NextVersion.md for 3.6.0

pmconne:
Add things that were missing.

@DStradley can you supply any images demonstrating eye-dome lighting for point clouds?

PR #4900, opened 2023-01-10
allow docs ci build to publish artifacts on manual trigger

ben-polinsky:
The ci build for docs relies upon a published artifact produced by the last run of the same build on master. However, we cannot select the last run _which produces an artifact_ (without using ADO's REST api) and so the build fails anytime the docs ci was run manually. This changes the build to produce artifacts any time it is run off master or release.


ben-polinsky:
@skirby1996 - thoughts? @pmconne just ran into this issue

skirby1996 (approved):

skirby1996:
> @skirby1996 - thoughts? @pmconne just ran into this issue in #4897.

This looks good to me, we should backport this to 3.5 just in case we run into the same issue.

ben-polinsky:
@calebmshafer @wgoehrig @ColinKerr quick review?

wgoehrig (approved):

PR #4902, opened 2023-01-10
hard code last docs build with published artifacts

ben-polinsky:
This is a quick-fix to get docs builds working again: Hard coding the last successful build with published artifacts so the next run off master will succeed. Following that, we can revert this change and rethink this workflow.


ben-polinsky:
@skirby1996

calebmshafer (approved):

PR #4912, opened 2023-01-11
Log a warning when successfully downloading a v1 checkpoint after failing to download a v2 checkpoint.

nick4598:


common/changes/@itwin/core-backend/nick-add-querycurrentprecedingv2_2022-12-01-19-37.json
```diff
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/core-backend",
-      "comment": "add queryCurrentOrPrecedingV2Checkpoint as an optional function to backendhubaccess interface",
+      "comment": "add allowPreceding as an optional boolean to CheckpointProps interface",
```
nick4598, line 5: Also modified rush chanhge from a previous PR. After requested changes, the old message: "add queryCurrentOrPrecedingV2Checkpoint as an optional function to backendhubaccess interface" was not true.

nick4598:
> Just curious: is it weird that you log `iTwinId` some places but not others?

I suppose so, I added iTwinId to the other two places. I'm guessing its safer to have more information than less.

PR #4915, opened 2023-01-12
Presentation: Add API to enable hierarchy levels filtering in tree

saskliutas:
Added API for enabling hierarchy level filtering in `ControlledTree` using PresentationTreeDataProvider. To enable filtering `PresentationTreeRenderer` should be used in `ControlledTree.`

Closes https://github.com/iTwin/itwinjs-backlog/issues/497


presentation/components/src/presentation-components/tree/DataProvider.ts
```diff
@@ -197,14 +207,68 @@ export class PresentationTreeDataProvider implements IPresentationTreeDataProvid
...
+
+  // TODO: handle case when requesting children for node with too many children
+
+  return {
+    nodes: [createInfoNode(parentNode, translate("tree.no-filtered-children"))],
+    count: 1,
+  };
+}
+
+function createInfoNode(parentNode: PresentationTreeNodeItem, message: string): PresentationTreeNodeItem {
```
grigasp, line 242: Do we really want this info type of node to be a `PresentationTreeNodeItem`? Being a `PresentationTreeNodeItem` means it needs to have a key, which is fake doesn't make much sense. Seems like it could be an `ImmediatelyLoadedTreeNodeItem` instead.

Also, do we have a plan for "advanced" rendering of the node, e.g. a link in the middle of its text to open a dialog?

presentation/components/src/presentation-components/tree/DataProvider.ts
```diff
@@ -197,14 +207,68 @@ export class PresentationTreeDataProvider implements IPresentationTreeDataProvid
   }
 }
 
+/** @alpha */
+export interface PresentationTreeNodeItem extends DelayLoadedTreeNodeItem {
+  key: NodeKey;
+  filterInfo?: PresentationInstanceFilterInfo;
+  isFilteringDisabled?: boolean;
```
grigasp, line 214: what do you think about grouping all filtering-related attributes together? e.g.
```ts
filtering?: {
  // filtering is disabled if descriptor or its getter is not set
  descriptor?: Descriptor | Promise<Descriptor>;
  // currently applied filter
  active?: PresentationInstanceFilterInfo;
};
```

grigasp (changes requested):

presentation/components/src/presentation-components/tree/controlled/PresentationTreeNodeRenderer.tsx
```diff
@@ -6,70 +6,100 @@
...
+      />
+    </TreeNodeRenderer>
+  );
 }
 
-interface TooManyChildNodeRendererProps {
-  nodeDepth: number;
+interface PresentationTreeNodeActionsProps {
+  onFilter: () => void;
+  onClear: () => void;
```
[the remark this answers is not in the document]
saskliutas: That button opens dialog for filtering so for me it looks like better naming would be `onFilterClick` and `onClearFilterClick`.

presentation/components/src/test/tree/controlled/PresentationTreeRenderer.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+    const closeButton = container.querySelector(".presentation-instance-filter-dialog-close-button");
+    expect(closeButton).to.not.be.null;
+    fireEvent.click(closeButton!);
+
+    const dialog = container.querySelector(".presentation-instance-filter-dialog");
+    expect(dialog).to.be.null;
+  });
+
+  it("renders filter builder dialog when node filter button is clicked", async () => {
```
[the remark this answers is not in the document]
saskliutas: Removed

presentation/components/src/presentation-components/tree/DataProvider.ts
saskliutas: Added separate type for `PresentationInfoTreeNodeItem`.

Regarding advanced rendering my idea is to render it in plain text when used in `PresentationTreeNodeRenderer` and have a better rendering in `PresentationTreeRenderer` as that's where we have all necessary data to open dialog, find parent node, etc.

presentation/components/src/presentation-components/tree/DataProvider.ts
saskliutas: Updated. I like having getter or `Descriptor` itself on `TreeNodeItem`. Just changed `descriptor` property to not be nullable as there is no point of setting `filtering` object if `Descriptor` is `undefined`

grigasp (approved):

grigasp (approved):

ui/components-react/src/test/filter-builder/FilterBuilderRuleGroup.test.tsx
```diff
@@ -34,31 +40,39 @@ describe("PropertyFilterBuilderRuleGroupRenderer", () => {
...
   });
 
+  it("does not render operator selector if only one rule is in group", () => {
+    const { queryByText } = renderWithContext(<PropertyFilterBuilderRuleGroupRenderer
+      {...defaultProps}
+      group={{ id: "id", items: [{ id: "childId", groupId: "id" }], operator: PropertyFilterRuleGroupOperator.And }}
+    />);
+    expect(queryByText(TestUtils.i18n.getLocalizedString("Components:filterBuilder.operators.or"))).to.be.null;
+  });
+
```
raplemie, line 72: I dont see a test that validates that this would ever occur, even if there was more that 1 item in the group at this level.

(If there was multiple items in the group, I'm pretty sure this test would still succeed as we are look for the "or" string rather than the "and", that may be expected, but given that there is no comparison, it look like a simple error.)

saskliutas: Thanks, nice catch. It should look for "and" operator. Fixed it.

raplemie (approved):

PR #4917, opened 2023-01-12
Reset the iModelId on the TS object (SnapshotDb) after calling validateCheckpointGuids

nick4598:
image
from below SEQ link:
https://qa-seq.bentley.com/#/events?filter=iModelId%20%3D%20%228b3ddd69-b6a5-4ad8-80b9-5242fb362215%22%20AND%20Application%20%3D%20%22ChangedElementsAgent%22
iModel: https://qa-connect-imodelhubwebsite.bentley.com/Context/1f96bcc1-65e5-4636-a5e0-471f0afbde85/iModel/8b3ddd69-b6a5-4ad8-80b9-5242fb362215

You can see that the iModelId is outdated( not properly set up) from the screenshot of the SEQ logs. The old iModelId seen as "dbGuid" in the SEQ screenshot returns a 404 when querying the hub. We end up getting a 404 even though we've updated the dbGuid, because the SnapshotDb's iModelId is set before we update the nativeDb's guids. The old iModelId remains stored in the SnapshotDb which is proven by the below screenshot. You can see in the debug console, that the iModelId logged differs from db.iModelId. That gets used to queryChangesets in BriefcaseManager.pullAndApplyChangesets. Thanks @diegopinate for running this agent in the debugger.

image

Note: this is an old iModel.. ~2019


core/backend/src/CheckpointManager.ts
```diff
@@ -330,6 +330,7 @@ export class CheckpointManager {
           nativeDb.resetBriefcaseId(BriefcaseIdValue.Unassigned);
 
         CheckpointManager.validateCheckpointGuids(checkpoint, nativeDb);
+        (db as any).iModelId = nativeDb.getIModelId();
```
nick4598, line 333: Does anyone have better suggestions for how to fix this? We could look at the nativeDb in BriefcaseManager.pullAndApplyChangesets when we're grabbing the iModelId... but then we still have an outdated iModelId stored on the TS object. We can't update it in validateCheckpointGuids because it takes a DgnDb.

kabentley: i don't understand. Is this a V2 checkpoint?

nick4598: Yes its a V2 checkpoint. Initially, I thought we couldn't do that since its a public function... but I guess since its internal we wouldn't be in violation of anything right?

kabentley: yes, internal functions can be changed. But how did we get an invalid V2 checkpoint? That should be where we fix this problem, right?

common/changes/@itwin/core-backend/nick-breficasemanager-bug-fix_2023-01-12-16-54.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "reset the iModelId on the TS object after validatingCheckpointGuids",
```
kabentley, line 5: this is not a public change. Please remove this description.

kabentley (changes requested):

core/backend/src/CheckpointManager.ts
nick4598: Its quite an old iModel used for testing that was made in 2019 I believe. At that time, I don't think the hub was resetting iModelId when copying iModels

nick4598: You're right, I'll make that change in the v2 checkpoint agent.

kabentley: what "agent" this this come up in?

nick4598: changed elements agent

kabentley (approved):

PR #4918, opened 2023-01-12
Fix Viewport waitForSceneCompletion API to wait properly for tiles from external sources

### LINKED ISSUE #4653: Viewport waitForSceneCompletion should wait for Orbit tile trees
**Describe the bug**
Viewport waitForSceneCompletion method does not wait for external tile trees to finish loading, which in testing was exhibited by Orbit tile trees.

**To Reproduce**

1. Open display-test-app.  Attach a point cloud reality dataset in Orbit point cloud format.
2. Navigate so it is in the process of loading points / tiles
3. Quickly keyin "dta save image" while tiles are still loading - this keyin internally uses waitForSceneCompletion
4. Observe that the image snapped contains a partially rendered point cloud scene.  This is because waitForSceneCompletion does not appear to have code to wait for external tile trees to finish loading (it does, but only if they are secondaryViewports).

**Expected behavior**
The point cloud should fully render after the waitForSceneCompletion promise resolves.

**Screenshots**
TBD

- OS: Tested on MacOS and Windows
- Browser: Electron / Chrome
- iTwin.js Version: 3.x

**Additional context**
N/A

markschlosseratbentley:
Fixes #4653 

The `waitForSceneCompletion` API on `Viewport` will now wait properly for external tile sources (particularly OrbitGT tile trees).  Those tile sources internally use a `hasMissingData` flag to indicate that the tree needs to load more data. This has now been exposed internally to `waitForSceneCompletion` so it can properly wait - it is not exposed directly; instead, we bump up the number of reported requested tiles by 1 to indicate we need to keep waiting.  Otherwise, `waitForSceneCompletion` would erroneously resolve its promise before the scene is actually finished loading tiles from that external source.

TODO
- [x] Test locally
- [x] Pass Image Tests


core/frontend/src/tile/OrbitGtTileTree.ts
```diff
@@ -280,7 +280,7 @@ export class OrbitGtTileTree extends TileTree {
     const tileCount = frameData.tilesToRender.size();
 
     // Inform TileAdmin about tiles we are handling ourselves...
-    IModelApp.tileAdmin.addExternalTilesForUser(args.context.viewport, { requested: frameData.tilesToLoad.size(), selected: tileCount, ready: tileCount });
+    IModelApp.tileAdmin.addExternalTilesForUser(args.context.viewport, { requested: frameData.tilesToLoad.size(), selected: tileCount, ready: tileCount, hasMissingData: frameData.hasMissingData() });
```
markschlosseratbentley, line 283: I am realizing now that this could not be tracked properly if there were multiple external sources adding themselves to the external statistics - since this is a boolean flag it does not accumulate.  I should make it a count, or just "lie" about the requested number of tiles (bump it by one if this flag is toggled on).  Marking as draft as I reconsider.

core/frontend/src/tile/TileAdmin.ts
```diff
@@ -30,6 +30,7 @@ export interface ExternalTileStatistics {
   requested: number;
   selected: number;
   ready: number;
+  hasMissingData?: number;
```
pmconne, line 33: This name doesn't make sense for a numeric property.
Did you consider simply having OrbitGT tree add 1 to `requested` if it has missing data?

markschlosseratbentley: Yes that was my first thought, but it felt a bit obscure to sort of lie about the meaning of `requested`.  Do you think that seems better?  We could go that route.

core/frontend/src/Viewport.ts
```diff
@@ -2583,17 +2583,23 @@ export abstract class Viewport implements IDisposable, TileUser {
       haveExternalTexRequests = system.hasExternalTextureRequests;
       haveNewTiles = !this.areAllTileTreesLoaded || this._hasMissingTiles;
       if (!haveNewTiles) {
-        // ViewAttachments and 3d section drawing attachments render to separate off-screen viewports - check those too.
-        for (const vp of this.view.secondaryViewports) {
-          if (vp.numRequestedTiles > 0) {
-            haveNewTiles = true;
-            break;
-          }
+        const tilesThisVp = IModelApp.tileAdmin.getTilesForUser(this);
```
pmconne, line 2586: Would be nice to consolidate code copied between the two implementations of `waitForSceneCompletion` rather than adding more, if possible.

pmconne (approved):

core/frontend/src/tile/TileAdmin.ts
markschlosseratbentley: reworked this to not expose hasMissingData - instead bump up number of requested by 1.

PR #4932, opened 2023-01-17
Automate Local Build of Core Reference Documentation

yashincontrol:
Added copyReferenceFilesToStaging.js in common/scripts and modified gather-docs.yaml to call it.


common/scripts/copyReferenceFilesToStaging.js
```diff
@@ -0,0 +1,57 @@
...
+  return console.error(err);
+}
+
+const basePath = path.resolve(__dirname, "..", "..", "..");
+let dest;
+
+if (process.argv[2]) {
+  dest = process.argv[2];
+} else {
+  dest = "itwinjs-core/staging-docs";
```
pmconne, line 29: You assume the root repo directory is called "itwinjs-core" (not true on my machine), and your `basePath` assumes this script won't be relocated?
Considered using `git rev-parse` instead?

ben-polinsky (approved):

common/scripts/copyReferenceFilesToStaging.js
pmconne: I'm curious - under what circumstances would you `git rev-parse --show-toplevel` produce multiple lines of output?

pmconne (approved):

common/scripts/copyReferenceFilesToStaging.js
yashincontrol: The output from `git rev-parse --show-toplevel` seems to be ending with `"\n"`. That was causing problems with the rest of the code so I got rid of that.

ben-polinsky:
@calebmshafer @wgoehrig @ColinKerr need one more review please.

common/scripts/copyReferenceFilesToStaging.js
```diff
@@ -0,0 +1,69 @@
...
+const childProcess = require("child_process");
+const options = {
+  encoding: "utf8"
+};
+
+let fse;
+try {
+  fse = require("fs-extra");
+} catch (err) {
+  console.log("Please check that fs-extra is installed");
```
aruniverse, line 24: wouldnt it be better to not swallow the error and log it? if its missing the dep it just throw an error anyway

common/scripts/copyReferenceFilesToStaging.js
```diff
@@ -0,0 +1,69 @@
...
+  return console.error(err);
+}
+
+let basePath;
+try {
+  basePath = childProcess.execSync("git rev-parse --show-toplevel", options);
+  basePath = basePath.split("\n").join("");
+} catch (err) {
+  console.log("Could not get the root directory");
```
aruniverse, line 33: should some of these catchs be fatal instead of swallowing/logging the error?

common/scripts/copyReferenceFilesToStaging.js
```diff
@@ -0,0 +1,69 @@
...
+try {
+  fse.ensureDirSync(path.resolve(dest, "extract"));
+} catch (err) {
+  console.log("Could not create staging directory structure");
+  console.error(err);
+}
+
+//copy docs
+try {
+  const docsPath = dest;
```
aruniverse, line 53: dont need an extra var for this

common/scripts/copyReferenceFilesToStaging.js
ben-polinsky: We thought it might be more clear to the user. Could throw.

common/scripts/copyReferenceFilesToStaging.js
ben-polinsky: The `return` effectively exits the program. Again, could throw.

common/scripts/copyReferenceFilesToStaging.js
aruniverse: i defer to you guys, this might be too nit picky. feel free to resolve both comments

ben-polinsky: Nah definitely good to consider.

aruniverse (approved):

PR #4934, opened 2023-01-18
Fail schema import that require data transfer by default

khanaffan:
Change in this PR does not break any existing behavior due to the fact on itwin-core we acquire exclusive lock when doing schema operation by default. Thus, any data transform done during schema upgrade, profile upgrade, or schema import will be done by only one briefcase. 

Future changes to use lockless schema changes will require if a schema change did cause data transform or not. 

[Corresponding imodel-native PR](https://github.com/iTwin/imodel-native/pull/94).


pmconne (approved):

ColinKerr (approved):

aruniverse:
@khanaffan is this required for the 3.6 release?

aruniverse:
> > @khanaffan is this required for the 3.6 release?
> 
> No, it's not required.

Do you intend for this to go in a 3.6 patch or 4.0?

pmconne:
@khanaffan are you waiting on anything other than @kabentley and presentation approvals on the [itwinjs-core PR](https://github.com/iTwin/imodel-native/pull/94)?

khanaffan:
> 

Yes. i am waiting for native PR to merge. It seems to fail on a test. I am looking at memory leak issue so could not investigate why its failing as no test fail for me locally. This should not block 3.6 release. We can push it out as patch. But i  will merge it as soon as native PR is merged.

core/backend/src/IModelDb.ts
```diff
@@ -2388,8 +2388,8 @@ export class BriefcaseDb extends IModelDb {
     // good thing computers are fast. Fortunately upgrading should be rare (and the push time will dominate anyway.) Don't try to optimize any of this away.
     await withBriefcaseDb(briefcase, async (db) => db.acquireSchemaLock()); // may not really acquire lock if iModel uses "noLocks" mode.
     try {
-      await this.doUpgrade(briefcase, { profile: ProfileOptions.Upgrade }, "Upgraded profile");
-      await this.doUpgrade(briefcase, { domain: DomainOptions.Upgrade }, "Upgraded domain schemas");
+      await this.doUpgrade(briefcase, { profile: ProfileOptions.Upgrade, schemaLockHeld: true }, "Upgraded profile");
```
kabentley, line 2391: is all this still necessary (opening/closing/pushing changes in separate changesets?)

kabentley:
The [previous PR](https://github.com/iTwin/imodel-native/pull/94) depends on this PR.

core/backend/src/IModelDb.ts
khanaffan: Yes. We need to refactor this. We have overloaded open() call with too many operations. It should just open the file. Domain/Profile flags are now same as domain will upgrade profile if needed. Which mean we just need two methods

1. Db::Open() 
2. Db::Upgrade().

If the open fails to open, then you call upgrade it just upgrades too latest (profile and schemas).

MichaelBelousov (approved):

Issue #4863, opened 2023-01-03
Hierarchy levels that don't support filtering


grigasp:
There are certain presentation rule combinations that make hierarchy level filtering very inefficient to a point where it doesn't even make sense to support it. At the moment our code assumes that all hierarchy levels can be filtered. Changes needed:

- [ ] Put a flag on root hierarchy level / nodes telling that they don't support filtering. The library should:
  - detect situations when filtering is worthless and set the flag (e.g. usage of `hideExpression` attribute).
  - throw when such hierarchy level is requested with an instance filter. 

- [ ] UI components should detect the flag and make sure filtering is not allowed for these hierarchy levels.

PR #4937, opened 2023-01-19
Presentation: Unfilterable nodes

grigasp:
Part of https://github.com/iTwin/itwinjs-core/issues/4863.

Related addon PR: https://github.com/iTwin/imodel-native/pull/118


common/api/presentation-common.api.md
```diff
@@ -1086,6 +1086,17 @@ export interface HierarchyCompareOptions<TIModel, TNodeKey, TRulesetVariable = R
     resultSetSize?: number;
 }
 
+// @alpha
```
pmconne, line 1089: What's the motivation for tagging as `alpha`, vs `internal` or `beta`?

pmconne (approved):

common/api/presentation-common.api.md
grigasp: We've been adding multiple hierarchy level filtering related pieces over a long period of time and setting them all `@alpha`. The plan is to propagate all of them to `@beta` as soon as all these pieces are glued together and we enable filtering in our test app. I hope that's going to happen early next week.

Then, the plan is to integrate filtering into our consumer trees (Models, Categories, etc.) and finally propagate these APIs to `@public` with 4.0 release.

grigasp: I figured I didn't totally answer your question... The motivation is that we're not sure about the APIs until we make sure all of them can work together.

pmconne: Thanks.

presentation/common/src/test/hierarchy/HierarchyLevel.test.ts
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import { expect } from "chai";
+import { HierarchyLevel } from "../../presentation-common/hierarchy/HierarchyLevel";
+import { Node } from "../../presentation-common/hierarchy/Node";
+import { createTestNode } from "../_helpers/Hierarchy";
+
+describe("HierarchyLevel", () => {
```
saskliutas, line 10: I think we should decide how we format our tests moving forward. Because now we have two kinds of formatting for test:
1. This one in this file with a lot of empty lines before and after closing brackets.
1. [This one](https://github.com/iTwin/itwinjs-core/blob/e98bdf8e927d805c3d053572fb5edd446abebdbe/presentation/components/src/test/tree/Utils.test.ts#L120) without empty lines.

I would prefer the second one.

saskliutas (approved):

presentation/common/src/test/hierarchy/HierarchyLevel.test.ts
grigasp: I personally like the first approach better :) Filed [an issue](https://github.com/iTwin/itwinjs-core/issues/4940) so we could have a discussion, make the decision and update the existing tests.

PR #4938, opened 2023-01-19
Rework when Electron window size and position is saved to the settings file

GytisCepk:
## Problem

We rely on [`resized`](https://www.electronjs.org/docs/latest/api/browser-window#event-resized-macos-windows) and [`moved`](https://www.electronjs.org/docs/latest/api/browser-window#event-moved-macos-windows) events to update Electron window size and position in the settings file. Since `resized` and `moved` events aren't emitted on Linux and are emitted only on manual resize / move on Windows, size and position can get outdated.

## Solution

Switch to using [`resize`](https://www.electronjs.org/docs/latest/api/browser-window#event-resize) and [`move`](https://www.electronjs.org/docs/latest/api/browser-window#event-move) events. These events are emitted more frequently and are consistent across different platforms. 

One potential drawback: when window is maximized, full screen size and position will be saved to settings file. Previously, size and position would remain the same as before maximizing window (not sure if there are any real world cases where this would be useful).


pmconne (approved):

GytisCepk:
@ben-polinsky, @iTwin/itwinjs-core-ui could you review this ?

Not a hard requirement for 3.6, but I would like to get this in due to deprecation.

ben-polinsky (approved):

GerardasB (approved):

PR #4939, opened 2023-01-19
Improved map-layer coordinate systems validation and error reporting

mdastous-bentley:
Some map-layer formats/configuration don't support server-side reprojection.  Those services needs to server data in EPSG:3857 CS to be correctly displayed, if not, UI will now display a specific error message.


pmconne:
/azp run
fyi @mdastous-bentley for whatever reason github doesn't trigger builds when you mark ready for review.

pmconne (approved):

AlainRobertAtBentley:
@mdastous-bentley Just a suggestion. We use many different expression to designate Spatial Reference Systems(SRS), Coordinate Reference System(CRS), Geographic Coordinate System(GCS) and so on yet at Bentley we tend to use GCS the industry most often uses Spatial Reference System or Coordinate Reference System. In the iTwin documentation we tried to use Geographic Coordinate Reference System or Coordinate Reference System(CRS). I suggest we adopt this nomenclature everywhere in iTwinjs, at least for displayable messages, so instead of coordinate system I would suggest the use of coordinate reference system.
Just to clarify coordinate system refers to a cartesian space with two axes and values. A coordinate Reference System defines the relation between a coordinate system (the cartesian space) and an object of reference (usually the Earth)

mdastous-bentley:
> @mdastous-bentley Just a suggestion. We use many different expression to designate Spatial Reference Systems(SRS), Coordinate Reference System(CRS), Geographic Coordinate System(GCS) and so on yet at Bentley we tend to use GCS the industry most often uses Spatial Reference System or Coordinate Reference System. In the iTwin documentation we tried to use Geographic Coordinate Reference System or Coordinate Reference System(CRS). I suggest we adopt this nomenclature everywhere in iTwinjs, at least for displayable messages, so instead of coordinate system I would suggest the use of coordinate reference system. Just to clarify coordinate system refers to a cartesian space with two axes and values. A coordinate Reference System defines the relation between a coordinate system (the cartesian space) and an object of reference (usually the Earth)

@AlainRobertAtBentley that makes sense to me, I will update the error message.

MarcBedard8 (approved):

PR #4948, opened 2023-01-20
RPC deprecations

swbsi:
This deprecation is to allow refactoring (or possible replacement) of the RPC system on 5.0


swbsi:
FYI...I still have lint issues to fix.

pmconne (changes requested):
Please use `@deprecated in 3.6" instead of `@deprecated`.

It doesn't make a lot of sense to deprecate an `@internal` API, especially if 1) we're still using it and 2) there's no replacement for it.

For `@beta` and `@public` APIs, please include either a description of what to use instead of the deprecated API, or a reason why it is deprecated. It seems most of these have no replacement, just might be removed or significantly refactored after 4.0.

If any of the `@beta` APIs are not really intended for use outside of iTwin.js core (e.g., the mobile APIs?), consider making them `@internal` instead of deprecating.

pmconne (approved):

presentation/common/src/presentation-common/PresentationRpcInterface.ts
```diff
@@ -171,7 +171,7 @@ export type ComputeSelectionRpcRequestOptions = PresentationRpcRequestOptions<Co
  * Interface used for communication between Presentation backend and frontend.
  * @public
  */
-export class PresentationRpcInterface extends RpcInterface {
+export class PresentationRpcInterface extends RpcInterface { // eslint-disable-line deprecation/deprecation
```
grigasp, line 174: I suppose this means `PresentationRpcInterface` should also be deprecated with the same comment?

swbsi: Not necessarily...

Unless we switch to a completely different paradigm on 5.0 (like REST), we will still have the concept of an “RPC interface” going forward.

I don’t see the need to deprecate this interface (or the imodel read or tile interfaces, etc) unless their own apis (ie, the set of remote apis that they describe) will be changing...

grigasp: I'm worried `RpcInterface` gets removed or completely changed, which would require the same for `PresentationRpcInterface`. That would force us to break our promise of not breaking APIs for 2 versions since deprecation.

NancyMcCallB (approved):

grigasp (approved):
Approving presentation changes. 

Have to note, though, the deprecation message `The RPC system will be significantly refactored (or replaced) in version 5.0.` would sound very scary to me as an API consumer, as it warns about an API being removed without offering any path forward.

core/backend/src/rpc/tracing.ts
```diff
@@ -14,12 +14,15 @@ import { AsyncLocalStorage } from "async_hooks";
...
 
+/* eslint-disable deprecation/deprecation */
+
 /**
  * Utility for tracing Rpc activity processing. When multiple Rpc requests are being processed asynchronously, this
  * class can be used to correlate the current calltrace with the originating RpcActivity. This is used for automatic appending
  * of RpcActivity to log messages emitted during Rpc processing. It may also be used to retrieve the user accessToken
  * from the RpcActivity.
  * @public
+ * @deprecated in 3.6. The RPC system will be significantly refactored (or replaced) in version 5.0.
```
calebmshafer, line 25: Do we have any alternative today to using this or is it pre-emptively deprecating without replacement?

swbsi: I think this would only get potentially moved to a new package during the refactor. Seems like necessary functionality...

calebmshafer: Had an offline discussion with @swbsi and he is going to remove the deprecation from some of the critical-path public APIs that do not yet have a replacement. In the future once we have a replacement identified, they can be deprecated.

swbsi: I have removed RpcInterface, RpcActivity, RpcTrace, SerializedRpcActivity, and RpcResponseCacheControl from the deprecation list.
Everything else seemed internal to the RPC system (and ideally should never have been public anyway).

We could still do a pretty major refactor (or even replacement) even with these top-level classes undeprecated.
They would just just forward to the new system internally...

calebmshafer: sounds good, thanks!

calebmshafer (approved):

PR #4950, opened 2023-01-20
fix Mobile lifecycle events (frontend)

swbsi:
TS side of fix.
Mobile lifecycle events (suspend and resume) are now properly received on the frontend.

My initial approach (attempting to use the websocket) proved unreliable for suspend. Too much is happening in the app during the suspend/resume sequence to guarantee that the message will be transmitted before the socket goes away.

This approach directly calls the JS event handlers in the frontend webivew from platform-specific code.

### DIFF
--- a/common/api/core-mobile.api.md
+++ b/common/api/core-mobile.api.md
@@ -211,10 +211,6 @@ export interface MobileNotifications {
     // (undocumented)
     notifyAuthAccessTokenChanged: (accessToken: string | undefined, expirationDate: string | undefined) => void;
     // (undocumented)
-    notifyEnterBackground: () => void;
-    // (undocumented)
-    notifyEnterForeground: () => void;
-    // (undocumented)
     notifyMemoryWarning: () => void;
     // (undocumented)
     notifyOrientationChanged: () => void;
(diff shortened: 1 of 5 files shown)


swbsi:
FYI...this will not actually work until a new addon is ready.

aruniverse:
> FYI...this will not actually work until a new addon is ready.

Were the addon changes you need not included in the 3.6.4 build from yesterday? Please link to the addon changes

swbsi:
> > FYI...this will not actually work until a new addon is ready.
> 
> Were the addon changes you need not included in the 3.6.4 build from yesterday? Please link to the addon changes

I didn’t see that...it probably has what I need.

They were in here last week: https://github.com/iTwin/imodel-native-internal/pull/40#pullrequestreview-1242743722

aruniverse:
> They were in here last week: [iTwin/imodel-native-internal#40 (review)](https://github.com/iTwin/imodel-native-internal/pull/40#pullrequestreview-1242743722)

looks like its included in 3.6.1+

aruniverse:
@tcobbs-bentley @toddsouthenbentley can you guys test & verify this with the mobile test apps in this repo?

PR #4953, opened 2023-01-20
Refactor static classes under UiFramework

raplemie:
See NextVersion.md.

fixes iTwin/appui#89

Todo: 
- [ ] Extract interfaces that can be kept beta so we dont expose directly the internal type.

### DIFF
--- a/common/api/summary/appui-react.exports.csv
+++ b/common/api/summary/appui-react.exports.csv
@@ -124,7 +124,8 @@ deprecated;CategoryVisibilityHandler
 alpha;CategoryVisibilityHandlerParams
 deprecated;CategoryVisibilityHandlerParams
 public;ChildWindowLocationProps
-public;ChildWindowManager
+public;ChildWindowManager 
+deprecated;ChildWindowManager 
 beta;ClassGroupingOption
 deprecated;ClassGroupingOption
 beta;ClearEmphasisStatusField(props: ClearEmphasisStatusFieldProps): JSX.Element
@@ -154,7 +155,8 @@ public;class ConfigurableUiControl
 public;ConfigurableUiControlConstructor = new (info: ConfigurableCreateInfo, options: any) => ConfigurableUiElement
 public;ConfigurableUiControlType
 public;ConfigurableUiElement
-public;ConfigurableUiManager
+public;ConfigurableUiManager 
+deprecated;ConfigurableUiManager 
 public;ConfigurableUiReducer(state: ConfigurableUiState | undefined, action: ConfigurableUiActionsUnion): ConfigurableUiState
 public;ConfigurableUiState
 public;connectIModelConnection: (mapStateToProps?: any, mapDispatchToProps?: any) => InferableComponentEnhancerWithProps
@@ -165,7 +167,8 @@ public;ContentControlActivatedEvent
 public;ContentControlActivatedEventArgs
 public;ContentDialog(props: ContentDialogProps): JSX.Element
 public;ContentDialogChangedEvent 
-public;ContentDialogManager
+public;ContentDialogManager 
+deprecated;ContentDialogManager 
 public;ContentDialogProps 
 public;ContentDialogRenderer 
 public;ContentGroup
@@ -176,11 +179,13 @@ public;ContentLayoutActivatedEvent
 public;ContentLayoutActivatedEventArgs
 public;ContentLayoutComponentProps 
 public;ContentLayoutDef
-public;ContentLayoutManager
+public;ContentLayoutManager 
+deprecated;ContentLayoutManager 
 public;ContentProps
 public;ContentToolWidgetComposer(props: ContentToolWidgetComposerProps): JSX.Element
 public;ContentToolWidgetComposerProps
-public;ContentViewManager
+public;ContentViewManager 
+deprecated;ContentViewManager 
 internal;ContentWrapper(props: ContentWrapperProps): JSX.Element
 public;CoreTools
 public;createAction
@@ -292,7 +297,8 @@ public;FrontstageLaunchBackstageItem
 deprecated;FrontstageLaunchBackstageItem 
 public;FrontstageLaunchBackstageItemProps 
 deprecated;FrontstageLaunchBackstageItemProps 
-public;FrontstageManager
+public;FrontstageManager 
+deprecated;FrontstageManager 
 internal;FrontstageNineZoneStateChangedEventArgs 
 public;FrontstageProps 
 deprecated;FrontstageProps 
@@ -360,6 +366,18 @@ public;InputFieldMessageAddedEvent
 public;InputFieldMessageEventArgs
 public;InputFieldMessageRemovedEvent 
 alpha;InputStatus
+internal;InternalChildWindowManager
+internal;InternalConfigurableUiManager
+internal;InternalContentDialogManager
+internal;InternalContentLayoutManager
+internal;InternalContentViewManager
+internal;InternalFrontstageManager
+internal;InternalKeyboardShortcutManager
+internal;InternalModalDialogManager
+internal;InternalModelessDialogManager
+internal;InternalSyncUiEventDispatcher
+internal;InternalToolSettingsManager
+internal;InternalUiShowHideManager
 internal;isCollapsedToPanelState: (isCollapsed: boolean) => StagePanelState.Minimized | StagePanelState.Open
 internal;isFrontstageStateSettingResult(settingsResult: UiStateStorageResult): settingsResult is
 beta;isNoSelectionActive(): boolean
@@ -376,7 +394,8 @@ alpha;IVisibilityHandler
 deprecated;IVisibilityHandler 
 public;KeyboardShortcut 
 public;KeyboardShortcutContainer
-public;KeyboardShortcutManager
+public;KeyboardShortcutManager 
+deprecated;KeyboardShortcutManager 
 public;KeyboardShortcutMenu 
 public;KeyboardShortcutMenuEvent 
 public;KeyboardShortcutMenuState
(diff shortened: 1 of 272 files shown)


ui/appui-react/src/appui-react/UiFramework.ts
```diff
@@ -91,6 +96,96 @@ export interface TrackingTime {
  * @public
  */
 export class UiFramework {
+  /**
```
raplemie, line 99: For reviewers, I suggest starting here which is the root...

From here, you can look at the files exporting the `Internal*` classes, I'll have to extract each API instead of exposing the internal here, but other than that, the idea are there.

The rest of the files are mostly reacting to these changes, and trying not to use the deprecated stuff. (The idea is that eventually the implementation of the UiFramework properties would not be static classes but either instances where it make sense, or simply namespace, but only accessible through this single entry point)

raplemie:
Element in `/framework` folder are the public interfaces of the previous static manager classes.

Every Manager classes are no longer extending the internal class as this exposes the internal type as a public...
So every Manager classes are now directly calling the internal counterpart "manually", tests were added to ensure this, the Manager classes MUST NOT do any work or validation, ALL the logic must remain in the internal class.

That adds a lot of code, but I could not see a good way around it. And, it's not that bad conceptually, I guess.

ui/appui-react/src/appui-react/framework/FrameworkDialogs.ts
```diff
@@ -0,0 +1,101 @@
...
+  parentDocument: Document;
+}
+
+/**
+ * Manages dialog access
+ * @beta
+ */
+export interface FrameworkDialog {
+  /** @internal */
+  readonly dialogManager: DialogManagerBase;
```
GerardasB, line 30: We should avoid internal properties in interfaces, because it makes them impossible to implement?

ui/appui-react/src/appui-react/framework/FrameworkDialogs.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  /** Get the array of modal dialogs */
+  readonly dialogs: DialogInfo[];
+
+  /** Open a modal dialog
+   * @param dialog The Dialog to open
+   * @param id The id of the Dialog. If one is not provided, an id is generated.
+   * @param parentDocument The Document used to determine the owning window.
+   */
+  openDialog(dialog: React.ReactNode, id?: string, parentDocument?: Document): void;
```
GerardasB, line 40: `UiFramework.dialogs.modal.openDialog()` is not too consistent with i.e. `UiFramework.childWindows.open()`
Would this be better?
```suggestion
  open(dialog: React.ReactNode, id?: string, parentDocument?: Document): void;
```

ui/appui-react/src/appui-react/framework/FrameworkDialogs.ts
```diff
@@ -0,0 +1,101 @@
...
+  /** Close a modal dialog
+   * @param dialog The Dialog to open. If one is not specified, the active dialog will be closed.
+   */
+  closeDialog(dialog?: React.ReactNode): void;
+
+  /** @internal */
+  closeAll(): void;
+
+  /** Update the dialogs */
+  update(): void;
```
GerardasB, line 51: Do we plan to refactor these APIs later? I don't think this is needed in any way. Also feels like we should have one dialog manager w/ a `modeless` flag/options instead..

GerardasB (changes requested):

GerardasB:
I didn't comment on each instance separately in the PR, but we should probably:
1. Get rid of internal properties on interfaces
2. Implement new `Framework` interfaces by the internal corresponding class.

mdastous-bentley:
ui-test-app doesnt compile:
`src/frontend/tools/PopupTestPanel.tsx:8:10 - error TS2305: Module '"@itwin/core-react"' has no exported member 'BadgeType'.`

aruniverse:
@mergifyio backport release/3.6.x

ui/appui-react/src/appui-react/framework/FrameworkChildWindows.ts
```diff
@@ -0,0 +1,63 @@
...
+  left: number;
+  top: number;
+}
+
+/**
+ * [[UiFramework.childWindows]] interface.
+ * @beta
+ */
+export interface FrameworkChildWindows {
+  openChildWindows: OpenChildWindowInfo[];
```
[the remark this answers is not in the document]
raplemie: As this was previously public, I've left it but marked it readonly, this whole API is up to be reviewed so I'll leave it as is for now.

ui/appui-react/src/appui-react/framework/FrameworkDialogs.ts
raplemie: I removed all `internal` and `deprecated` elements from the new interfaces, and the places which were calling those are now calling the implementation. We will want to make sure all of this is OK when we change implementation.

ui/appui-react/src/appui-react/framework/FrameworkDialogs.ts
raplemie: Did many renames for consistency.

ui/appui-react/src/appui-react/childwindow/ChildWindowManager.tsx
```diff
@@ -7,207 +7,131 @@
...
+import { InternalChildWindowManager } from "./InternalChildWindowManager";
 
 /** Supports opening a child browser window from the main application window. The child window is managed by the main application
  * and is running in the same security context. The application must deliver the html file iTwinPopup.html along side its index.html.
  * See also: [Child Window Manager]($docs/learning/ui/appui-react/ChildWindows.md)
- * @public */
+ * @public
+ * @deprecated in 3.6. Use `UiFramework.childWindows` property to access.
+ * */
 export class ChildWindowManager {
```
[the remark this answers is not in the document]
raplemie: Since the new API is beta, and this class is public, extract-api sees a conflict so I cant mention that it implement the interface. All others are implemented by static classes so the "implements" clause do not work. We should adress this however when we make the switch to instances implementation (which I think should be the next step in this process)

ui/appui-react/src/appui-react/UiFramework.ts
```diff
@@ -103,12 +204,14 @@ export class UiFramework {
   private static _uiStateStorage: UiStateStorage = new LocalStateStorage();
   private static _settingsManager?: SettingsManager;
   private static _uiSettingsProviderRegistry: Map<string, UserSettingsProvider> = new Map<string, UserSettingsProvider>();
-  private static _PopupWindowManager = new ChildWindowManager();
+  private static _PopupWindowManager = new ChildWindowManager(); // eslint-disable-line deprecation/deprecation
```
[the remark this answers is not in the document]
raplemie: Renamed to `_childWindowsManager`

raplemie (commented):
The latest changes should adress the naming issues and make this PR ready for review.

common/api/summary/appui-react.exports.csv
```diff
@@ -781,13 +808,9 @@ public;ToolWidgetPropsEx
...
 internal;toPanelSide(location: StagePanelLocation_2): PanelSide
 internal;TrackingTime
-internal;UiActivityEvent 
-internal;UiActivityEventArgs
 public;UiDataProvidedDialog({ uiDataProvider, id, isModal, ...dialogProps }: UiDataProvidedDialogProps): JSX.Element
 public;UiDataProvidedDialogProps
 public;UiFramework
-internal;UiIntervalEvent 
-internal;UiIntervalEventArgs
 beta;UiItemsManager
```
GerardasB, line 814: Do we need to apply respective changes to the `UiItemsManager`?

ui/appui-react/src/appui-react/backstage/BackstageManager.tsx
```diff
@@ -7,71 +7,51 @@
...
-    });
+    return UiFramework.backstage.getBackstageToggleCommand(overrideIconSpec);
   }
 }
 
 /** Hook that returns isOpen flag of the backstage.
  * @public
  */
-export const useIsBackstageOpen = (manager: BackstageManager) => {
+export const useIsBackstageOpen = (manager: FrameworkBackstage) => {
```
GerardasB, line 54: This is a breaking change?

ui/appui-react/src/appui-react/backstage/BackstageManager.tsx
```diff
@@ -90,6 +70,6 @@ export const useIsBackstageOpen = (manager: BackstageManager) => {
  * @public
  */
 export const useBackstageManager = () => {
-  const [manager] = React.useState(UiFramework.backstageManager);
+  const [manager] = React.useState(UiFramework.backstage);
```
GerardasB, line 73: This is a breaking change as well (changed return type from `BackstageManager` to `FrameworkBackstage`?

GerardasB (changes requested):

ui/appui-react/src/appui-react/backstage/BackstageManager.tsx
raplemie: The interface is the same, it shouldnt be breaking.

raplemie:
> ui-test-app doesnt compile: `src/frontend/tools/PopupTestPanel.tsx:8:10 - error TS2305: Module '"@itwin/core-react"' has no exported member 'BadgeType'.`

And I forgot about those having a different behavior again, test apps are completely broken, I'll fix today.

raplemie:
All test apps fixed, removed `SyncUiEventDispatcher` from the refactor as the name `events` was clashing and felt wrong, we will revisit this system and see if we want to refactor it later.

2 layout related method on FronstageManager were removed in favor of their clones already existing in ContentLayoutManager:
`FrontstageManager.setActiveLayout` => `UiFramework.content.layouts.setActive`
`FrontstageManager.setActiveContentGroup` => `UiFramework.content.layouts.setActiveContentGroup`

deprecation notices were updated to 3.7 (instead of 3.6)

NextVersion.md was reset because we are now in 4.0 branch, I'll put back the NextVersion.md in the 3.7 branch when it is ported over.

GerardasB (approved):

mdastous-bentley (approved):

aruniverse:

PR #4956, opened 2023-01-23
Fix map-layers WMS request failure on some servers requiring 'Transparent' parameter value to bo upper-case (as specified in the WMS specification)

mdastous-bentley:
Fix issue with a specific WMS server:
[https://api.dataforsyningen.dk/cp_inspire](https://api.dataforsyningen.dk/cp_inspire)

Giving this error message on **GetMap** request:
```
<?xml version="1.0" encoding="UTF-8"?><ServiceExceptionReport xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.3.0" xsi:schemaLocation="http://www.opengis.net/ogc http://schemas.opengis.net/wms/1.3.0/exceptions_1_3_0.xsd">
  <ServiceException>TRANSPARENT must be either TRUE or FALSE</ServiceException>
</ServiceExceptionReport>
```

After revisiting the [WMS specification ](https://www.ogc.org/standards/wms), I could find the following:

> TRANSPARENT can take on two values, "TRUE" or "FALSE". 

And also:

> Parameter names shall not be case sensitive, but parameter values shall be.


pmconne:
Is this not yet ready for review?

mdastous-bentley:
> Is this not yet ready for review?

It's a very simple fix, it's ready for review.

pmconne (approved):

pmconne:
handle leaks reported during tests in map-layers on linux and presentation on windows. Rerunning.

aruniverse:
@mergifyio backport release/3.5.x

PR #4958, opened 2023-01-23
`Element.toJSON` should not include undefined values for userLabel or federationGuid

kabentley:
This code would remove the `federationGuid` from the element:

```ts
const el = <create instance of Element>;
el.insert();
<change properties of el>
el.update()
```
the `federationGuid` member should not be included by `Element.toJSON` if it is undefined.


core/backend/src/IModelDb.ts
```diff
@@ -1701,6 +1701,9 @@ export namespace IModelDb { // eslint-disable-line no-redeclare
      * @param elProps The properties of the new element.
      * @returns The newly inserted element's Id.
      * @throws [[IModelError]] if unable to insert the element.
+     * @note For convenience, the value of `elProps.id` is updated to reflect the resultant element's id.
+     * However when `elProps.federationGuid` is not present or undefined, a new Guid will be generated and stored on the resultant element, but
+     * the value of `elProps.federationGuid` is *not* updated.
```
pmconne, line 1706: Here and `Element.insert` are probably good places to recommend querying for a new `Element` or `ElementProps` after insertion if you intend to continue using the element.

kabentley: see what you think

pmconne (approved):

PR #4962, opened 2023-01-23
Deprecate displayed extents

### LINKED ISSUE #4960: Deprecate IModelConnection.displayedExtents
`IModelConnection.displayedExtents` is used as the basis for the viewed extents of all `SpatialViewState`s. It is expanded every time a reality model is loaded, and never shrinks. The intent was to make sure that reality models outside of the project extents are included in the viewed extents, but it doesn't make sense to apply this expansion to every spatial view. In extreme cases this can lead to graphical artifacts as the extents become far too large.

 Each spatial view should expand (or not) its own extents based on the reality models it is currently displaying.

n/a

**Additional context**
We need to ensure that a `Viewport`'s `invalidateController` method is invoked when the view's displayed extents change. Currently that's done in response to a call to `IModelConnection.expandDisplayedExtents`, which would technically continue working but we want a solution that does not rely upon deprecated APIs.

pmconne:
Fixes #4960.


pmconne (commented):
Awaiting ImageTests results.

markschlosseratbentley (approved):

pmconne:
ImageTests shows 60 differences, all of which look like expected changes in z-fighting due to slightly different frustum.

PR #4963, opened 2023-01-23
add optional `CodeSpecProperties.scopeSpec.relationship` member

kabentley:
Holds the relationship className when `type` is `CodeScopeSpec.Type.RelatedElement`


pmconne:
Does the native code already populate this field and/or use it?

kabentley:
> Does the native code already populate this field and/or use it?

Yes. @PaulPierog can add details. In TypeScript the `CodeSpec.properties`  was declared as `any` until recently when I added this interface. Paul pointed out that I missed one of the optional fields.

core/common/src/Code.ts
```diff
@@ -234,12 +234,19 @@ export namespace CodeScopeSpec {
...
+ */
 export interface CodeSpecProperties {
   scopeSpec: {
+    /** the type of CodeSpec */
     type: CodeScopeSpec.Type;
+    /** If true, the federationGuid of the scope element identifies the scope, for Codes that are globally unique.
+     * Otherwise, the ElementId of the scopeElement is used, for Codes that are scoped only within a single iModel.
+     */
     fGuidRequired?: boolean;
+    /** The relationship className, when `type` is `CodeScopeSpec.Type.RelatedElement` */
```
pmconne, line 248: In format "Schema:Class"?

kabentley: That's an excellent question. The only example I could find in the iModels in my disk just has a className with no schema prefix. @PaulPierog how does that work?

PaulPierog:
> > Does the native code already populate this field and/or use it?
> 
> Yes. @PaulPierog can add details. In TypeScript the `CodeSpec.properties` was declared as `any` until recently when I added this interface. Paul pointed out that I missed one of the optional fields.

The "relationship" property is required when the CodeScopeSpec.Type=RelatedElement.  It is used to define which relationship class to traverse to get the elementId for the CodeScope.

core/common/src/Code.ts
PaulPierog: I agree it should be "Schema:Class" but the current Plant applications only set className.

pmconne: Can someone point me to the native code that uses this property? There's no reference to "relationship" in CodeSpec.cpp except a TODO comment.

PaulPierog: > That's an excellent question. The only example I could find in the iModels in my disk just has a className with no schema prefix. @PaulPierog how does that work?

The Plant applications created the CodeSpecs and they are only used with the ProcessFunctional schema.  I agree it should be Schema:Class.

kabentley: added clarification that the form is "schema:class"

pmconne: Should we document that it might be missing the schema name (probably is in fact, given only Plant appears to set this)?

PaulPierog: If the schema name is missing the fallback should be the schema identified by the class of the element where the CodeSpec is referenced.

pmconne: Is it (or will it be) actually implemented that way somewhere? Where?

PaulPierog: > Is it (or will it be) actually implemented that way somewhere? Where?

PlantSight Aggregation and likely OpenPlant Engineering Data Manager both iTwin.js based.

kabentley: I don't think we should mention the Plant past "misuse". Only they care, and only they (in their code) will look at their CodeSpecs. Nothing in any of the core api's currently look at the value. Someone should write the code that does.

pmconne: I want the documentation of our file format to accurately represent the data that's out there, for our own benefit and that of our users. We don't need to talk about misuse or specific applications, but can we document it as @PaulPierog describes -"schema:class" or "class" with schema implicit? If not go ahead and resolve this.

PR #4964, opened 2023-01-23
Remove inline require from frontend code

kckst8:
Since this is frontend code, we should attempt to avoid inline require statements, which fail in esm-based frontends. Moved this to a top-level import as the savings are not significant as a dynamic import when compared to the cost of changing the function to async


aruniverse (approved):

ui/components-react/src/components-react/converters/TypeConverterManager.ts
```diff
@@ -50,10 +55,11 @@ export class TypeConverterManager {
...
 
-    if (!TypeConverterManager._defaultTypeConverter) {
-      const { StringTypeConverter } = require("../converters/StringTypeConverter"); // eslint-disable-line @typescript-eslint/no-var-requires
-      TypeConverterManager._defaultTypeConverter = new StringTypeConverter();
-    }
     return TypeConverterManager._defaultTypeConverter;
   }
 }
+
+// register string type converters here to avoid circular dependency in the StringTypeConverter module
```
raplemie, line 62: Since we removed the "new StringTypeConverter" line completely (previous line 55), there is no need to move the imports here as the "StringTypeConverter" type is no longer needed in that file, which also removes the circular depenency.

raplemie (changes requested):
I would move back the `registerConverter` calls to their related file, so it would have the same structure as others (if there are any)

ui/components-react/src/components-react/converters/TypeConverterManager.ts
kckst8: meant to move those back after this change. Will do now.

kckst8: done

raplemie (approved):

Issue #3792, opened 2022-06-10
Get rid of enums in rule definitions


grigasp:
Deprecate enums in rule definitions in favor of string unions. That would make it easier to copy/paste the rules between JSON and typescript and would require less imports when defining rules in typescript.

PR #4965, opened 2023-01-24
Presentation: Allow specifying enums values as strings

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/3792.

This is really just a convenience change for consumers - allows specifying enum values in rules as plain strings, while keeping them listed in an enum. This is shorter and doesn't require importing the enum. I updated the rulesets in learning snippets so they appear nicer in our docs, but left other tests as-is to make sure the old way still works.


saskliutas (approved):

presentation/common/src/presentation-common/rules/Rule.ts
```diff
@@ -19,7 +19,7 @@ import { RequiredSchemaSpecification } from "./SchemasSpecification";
  */
 export interface RuleBase {
   /** Used for serializing to JSON. */
-  ruleType: RuleTypes;
+  ruleType: `${RuleTypes}`;
```
pmconne, line 22: In places where you've replaced an `enum` type with a template literal, please include a doc link to the `enum` so people can easily find it when browsing the documentation site, as I assume (@ben-polinsky confirm?) that the docs build will not generate one.
In future consider preferring string union over string enum.

grigasp: > In places where you've replaced an enum type with a template literal, please include a doc link to the enum so people can easily find it when browsing the documentation site

Added. 

> In future consider preferring string union over string enum.

Generally that's what we do. These were leftovers from the initial implementation, so I had to keep the enums to avoid breaking current uses. On the other hand, I'd argue that using them like this instead of a string union makes it easier to document each value, especially if the type is used in more than one place.

pmconne: Did you mean `@see [[RuleTypes]]`?

grigasp: Looks like docs builder  understands the link without the brackets too:
- [source](https://github.com/iTwin/itwinjs-core/blob/8737e4cc516e0f410f829d5918b246d5bad3d655/core/frontend/src/ImageUtil.ts#L244)
- [result](https://www.itwinjs.org/reference/core-frontend/rendering/tryimageelementfromurl/)

pmconne: Yes but what is the scope? What does it do if you provide more than one word after `@see`, or the name of a symbol in a different file, or in a different package? @ben-polinsky might know - I certainly don't.
`[[Symbol]]` and `[Symbol]($package)` are unambiguous.

grigasp: I would guess it's the same as `[[Symbol]]`. I don't mind changing it to that format, but I managed to get 2 of the 4 flaky CI builds to pass on this PR, so would like to avoid having to restart again... I can replace the `@see Symbol` with `@see [[Symbol]]` across the whole repository in a separate PR afterwards.

pmconne (approved):

PR #4966, opened 2023-01-24
Presentation: Deprecate useless `ContentBuilder` methods

grigasp:
Part of https://github.com/iTwin/presentation/issues/35


.vscode/launch.json
```diff
@@ -706,7 +718,10 @@
       ],
       "outFiles": [
         "${workspaceFolder}/{core,clients,ui,presentation}/*/lib/**/*.js"
-      ]
+      ],
+      "env": {
+        "NODE_ENV": "development"
+      }
```
grigasp, line 724: @aruniverse FYI one of the tests was failing for me locally when run in console but not on debugger. Turns out the `assert` function from `@itwin/core-bentley` only has effect with `NODE_ENV = "development"`. The test is not failing CI builds either, so it looks like our CI build machines don't have that env var set up either. I guess they should?

saskliutas (approved):

pmconne (approved):

PR #4977, opened 2023-01-25
PropertyFilterBuilder: Fix rule operator rendering

### LINKED ISSUE #4973: `InstanceFilterBuilder`: Unnecessary character when using `Is TRUE` operator
1. Open `presentation-test-app`, choose the "Properties_60InstancesWithUrl2.ibim" dataset and "VisibilityWidget.ModelsTree-GroupedByClass" ruleset

2. In the hierarchy, click to filter this node:
  Image

3. Create "IsPrivate IS TRUE" filter:
  Image

Notice the extra `0` character next to operator picker.

saskliutas:
Fixed rule rendering as `0` symbol was rendered with `IS TRUE` operator.

Closes #4973


ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx
```diff
@@ -83,8 +83,16 @@ export function PropertyFilterBuilderRuleRenderer(props: PropertyFilterBuilderRu
...
+      {
+        property !== undefined
+          ? operatorRenderer(property)
+          : null
+      }
+      {
+        property && operator && !isUnaryPropertyFilterOperator(operator)
+          ? valueRenderer(property)
+          : null
+      }
```
grigasp, line 95: can you explain why the previous implementation was causing `0` to be rendered (for my own education)?

saskliutas: It was happening because that statement with inlined `&&` (`property && operator && !isUnaryPropertyFilterOperator(operator) && valueRenderer(property)`) was returning value of the condition were evaluation was stopped. In this particular case it was at `operator` because `IsTrue` enum value is `0` and that's falsy value so it returned 0. [TS playground](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBBCcQQ0MwAUVWDEQC0gADVHBQGNJ3KAALygABEAAkYNAMOB4GRonzFBTBPB3AERcwABScSASUAAETqkCVfIATAAGACMAA5Ska9aVjXyAJQBdDPGoMdxahTEVCUTjKvko0r3Yh8-gAb1AeA8ZHy-M47GQlFggdAiEasUjfPs4tQCe4TFT0dj6j5oAAvjabcEVDhmuGvEm+hyODQvm4zMx2AxmJB+LBoOpEvW9CIvHy6QymYJC+L+eAHKgAOTCAByar5QRUAHkvmR3oJO1pIEdUI4FK5pHhQE4kzphlh9CgEIhocJLk2JOLoiIDBxsQsmKwODwgkUFEY1AYgkEgGk81AINFBKEoADEKB+VBWWgdhO1guDiEgCluFQ9DeEwkocDUMg8OdAjFCLYJki4WB6kaExJgUR59CiVAlRtaCiPPfd8SHeC6IYrxamgaBeMEdi+QAbio8lGipK5wIAYUaeFhJ5NAyHQ2SZRoQZelUnh5AaLxeWSZ0d10hSjnAyCY3wlBeTA5B7MQAJEOQxzZOlBS5UgBV4GVWVkCM9TTNAAAySKlNcvNHKimLyA8Fi2OgDjSx8mV-MC5UDO+MKTMY6LYog+KKJQErku1VjJPSzibWkoA) to show this better.

I think in cases were returned result of whole statement with `&&` operators inlined is used it is better to use ternary operator. I think react handles when `false` or `undefined` value is passed to be renderer but for me using ternary operator and explicitly returning `null` when we don't want to renderer component looks safer.

grigasp (approved):

GerardasB (approved):

Issue #4974, opened 2023-01-25
`InstanceFilterBuilder`: Combo box & scrollbar issues


grigasp:
1. Open `presentation-test-app`, choose the "Properties_60InstancesWithUrl2.ibim" dataset and "VisibilityWidget.ModelsTree-GroupedByClass" ruleset

2. In the hierarchy, click to filter this node:
  ![Image](https://user-images.githubusercontent.com/35135765/214559951-b4e3533c-66df-4871-997a-bcda647f9b82.png)

3. Choose the "Model" property and open the "Selected target instance" combo box:
  ![Image](https://user-images.githubusercontent.com/35135765/214561584-540feee9-98d1-4645-afcc-f4b1329322dc.png)

Notice two issues:
1. The combo box is nested inside the component, which makes it nearly unusable.
2. Border of the combo box is obstructed by white chevron background when mouse is over the chevron.

PR #4979, opened 2023-01-25
Presentation: Fixes for PresentationFilterDialog

saskliutas:
Fixed styles of NavigationPropertyEditor to correctly render it in Dialog component.
While testing found that filters with navigation property were incorrectly converted to ECExpression. This was caused because property and field of that property typeName were different. Property itself has `long` type and field typeName is set to `navigation`.

Closes https://github.com/iTwin/itwinjs-core/issues/4974


grigasp:
I see this fixes the menu being nested inside the dialog, but still there are a few styling problems:
image

1. When the popout exceeds available width of the window, we can't see its scrollbar. Should we set max-width? Or maybe the popout can be made to align to the other side?
2. The chevron's white background is still on top of the border of focused combo box.
3. The "Select target instance" text becomes multiline when the combo box is too narrow. IMO it should stay on one line and apply ellipsis on the text (or at least cut it).

saskliutas:
I will take a lot.

Regarding 2 It might be an issue in itwinui as other select components we use in test-app have same behavior
image

grigasp:
> I will take a look.
> 
> Regarding 2 It might be an issue in itwinui as other select components we use in test-app have same behavior

I can't repro it here: https://itwin.github.io/iTwinUI-react/?path=/story/core-combobox--with-custom-message-icon. Maybe we add some kind of style customization?..

saskliutas:
I think this is noticeable only when select size is set to small. https://codesandbox.io/s/itwinui-react-example-forked-llbr13?file=/src/App.tsx

Also it looks like this is fixed in itwinui v2

grigasp:
> Also it looks like this is fixed in itwinui v2

@iTwin/itwinjs-core-ui, any plans to move to itwinui 2?

grigasp (approved):
I think the popout is still not perfect, but I'm approving just so we can get this into 3.6 and we can reiterate afterwards.

PR #4984, opened 2023-01-27
Reuse context reality model tile trees

### LINKED ISSUE #4654: Context reality models don't reuse tile trees
**Describe the bug**
When recalling a saved view containing a context reality model, all tiles completely reload, even when the recalled view is the exact same as the current view.

**To Reproduce**

1. Open display-test-app and attach a point cloud reality dataset.  Create and recall some external saved view of the point cloud.
2. Wait for all tiles to load.
3. Recall the same external saved view.
4. Observe that all tiles completely reload.

**Expected behavior**
When recalling a point cloud saved view which shows the same tiles as are currently being viewed, tiles should not reload.

**Screenshots**
N/A

- OS: MacOS / Windows
- Browser: Chrome / Electron
- Version N/A
- iTwin.js Version: 3.x branch 'eye-dome-lighting'

**Additional context**
The problem is not reproducible with persistent reality models (those associated with a geometric model).
`RealityTreeId` has a `modelId`. Tile trees are compared by Id - if the model Ids don't match, the tile trees are considered "different" so cannot be shared.
The model Id for a context reality model is generated at run-time from `IModelConnection.transientIds` in `RealityModelTileTree.Reference`'s constructor. Therefore every reference to the same tileset gets a unique model Id and unique TileTree.

Ideally the model Id would be generated for the `TileTree`, not the `TileTreeReference`. However, `ContextRealityModelState` has a `modelId` property that is `@public`. It is used to apply symbology overrides - though that's not meaningful until the tile tree is loaded. It is possible some code outside of itwinjs-core is relying on that modelId being defined before the tile tree loads, though. We might consider having the `RealityModelTileTree.Reference` constructor try to find a matching TileTree and use its model Id, instead of generating a new one.

pmconne:
Fixes #4654.

A context reality model has a model Id, which is used to apply feature symbology overrides to it. The Id must be unique within all reality models for a given view.
Every tile tree also has a model Id, which must match the model Id of the corresponding context reality model Id.

Previously, we would allocate a new model Id (and therefore, tile tree) for every context reality model.
This meant, for example, if you opened a view of a reality model, and then reopened the exact same view, we'd create two identical tile trees, download two sets of tiles, etc.
This could get really bad if, for example, you are synchronizing two viewports - this clones the view state each time one of the viewports is manipulated, allocating a new tile tree for each context reality model each time.

To solve the problem, when we instantiate a tile tree reference for a context reality model, we first look for an existing tile tree that is compatible with this reality model. If one exists, and its Id is not already used by another reality model in the same view, we reuse that tree and its model Id for this reality model.

I had to defer population of `DisplayStyleSettings.contextRealityModels` when instantiating a `DisplayStyleState`, because we want to look at `DisplayStyleState.settings.contextRealityModels` while populating them, but `settings` is not initialized until after that happens...


pmconne (commented):
Awaiting ImageTests results.

pmconne:
@DStradley please test locally.

pmconne:
@grigasp @saskliutas macOS and Windows; linux passed.
https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=1924520&view=logs&jobId=47fa0b0a-f6a7-5b52-b86a-09932c451137&j=9f3a3068-7880-5994-433d-b92c1f9d379b
```
--[ FAILURE: presentation-full-stack-tests ]-------[ 6 minutes 16.3 seconds ]--


  335 passing (6m)
  1 pending
  2 failing

  1) ReadWrite
       Handling read-write operations
         handles schema import during nodes request:
     Db is busy: Error importing schema
      at StandaloneDb.importSchemaStrings (D:\vsts_b\25\s\core\backend\src\IModelDb.ts:822:15)
  ...1 line omitted...
      at processImmediate (node:internal/timers:471:21)

  2) ReadWrite
       "after each" hook for "handles schema import during nodes request":
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (D:\vsts_b\25\s\full-stack-tests\presentation\lib\backend\ReadWrite.test.js)
```

DStradley (approved):

DStradley:
> @DStradley please test locally.

@pmconne I have tested this and everything behaves as expected now.  The only issue was the odd point cloud tile failure I showed you, that may not even be related.  I have tried many times since, several ways, and still cannot repro that issue now.

grigasp:
> @grigasp @saskliutas macOS and Windows; linux passed. https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=1924520&view=logs&jobId=47fa0b0a-f6a7-5b52-b86a-09932c451137&j=9f3a3068-7880-5994-433d-b92c1f9d379b
>
> On rerun all three passed.

Thanks, this is a known problem, investigation is in progress: https://github.com/iTwin/itwinjs-backlog/issues/488

PR #4991, opened 2023-01-27
Map-layers should not be displayed past maximum LOD limit advertised by service metadata

### LINKED ISSUE #4916: Map-layers should not be displayed past maximum LOD limit advertised by service metadata
Some map-layers service contains a maximum LOD value in their service's metadata,  indicating the maximum LOD (or zoom level). at which a layer should be displayed. For example the maximum LOD of Bing is 22.  The current implement of iTwin.js  is to continue to display tile of level 22 even though we are past the maximum LOD (i.e. last available tiles are stretched) : 
image

While this might be fine for the base map layer, because displaying stretched tile is better than blank tiles... when it comes to background/overlay layers,  it would be better to honor the maximum LOD and stop displaying map-layers past their maximum LOD value.  Doing so would match the behavior of other GIS viewer, and make sure the layer is displayed in its scale range definition.

So the idea would be to implement a different behavior on how maximum LOD are enforced, depending if the map-layers is displayed as a base map or not.

We might consider introducing a switch that would let applications revert to the legacy behavior if needed.

mdastous-bentley:
Fixes #4916


pmconne (approved):
No way to test any of this behavior?

mdastous-bentley:
> No way to test any of this behavior?

Problem is, the entire reality/map/imagery tiling mechanism is untested (and also undocumented)  and starting from scratch would be a lot of work.

pmconne:
> Problem is, the entire reality/map/imagery tiling mechanism is untested (and also undocumented) and starting from scratch would be a lot of work.

Of course. I just worry that the technical debt that's accumulating as we pile ever more special cases onto it.
Would it be possible to construct some views for ImageTests that will exercise the behavior, so we can detect regressions by comparing daily images?

mdastous-bentley:
@calebmshafer Can this be merged on 3.6 ?

pmconne:
Did you run ImageTests? Any differences?

mdastous-bentley:
Yes dataset was added to test images.

PR #4905, opened 2023-01-11
Fix tests that broke due to WAL mode

kabentley:
Tests that copy an IModel while it is opened for write must call `IModelDb.performCheckpoint`.

PR #4998, opened 2023-01-30
Enable WAL mode for all read-write `StandaloneDb`s

grigasp:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/488.

This is an extension of the following PRs:
- https://github.com/iTwin/itwinjs-core/pull/4853
- https://github.com/iTwin/itwinjs-core/pull/4861
- https://github.com/iTwin/itwinjs-core/pull/4905

In some of our tests we use `StandaloneDb` where we concurrently read and write to it. Need the iModel to use WAL mode or otherwise we may get into a deadlock of:
- main thread holding connection's mutex and waiting for iModel to unlock for write
- worker thread holding a read transaction open and trying to acquire connection's mutex

### DIFF
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -2715,6 +2715,7 @@ export class StandaloneDb extends BriefcaseDb {
   public static createEmpty(filePath: LocalFileName, args: CreateEmptyStandaloneIModelProps): StandaloneDb {
     const nativeDb = new IModelHost.platform.DgnDb();
     nativeDb.createIModel(filePath, args);
+    nativeDb.enableWalMode();
     nativeDb.saveLocalValue(BriefcaseLocalValue.StandaloneEdit, args.allowEdit);
     nativeDb.setITwinId(Guid.empty);
     nativeDb.resetBriefcaseId(BriefcaseIdValue.Unassigned);
@@ -2747,6 +2748,8 @@ export class StandaloneDb extends BriefcaseDb {
     const nativeDb = this.openDgnDb(file, openMode, undefined, options);
 
     try {
+      if (openMode === OpenMode.ReadWrite)
+        nativeDb.enableWalMode();
       const iTwinId = nativeDb.getITwinId();
       if (iTwinId !== Guid.empty) // a "standalone" iModel means it is not associated with an iTwin
         throw new IModelError(IModelStatus.WrongIModel, `${filePath} is not a Standalone iModel. iTwinId=${iTwinId}`);
--- a/core/backend/src/test/changesets/ChangeMerging.test.ts
+++ b/core/backend/src/test/changesets/ChangeMerging.test.ts
@@ -43,6 +43,7 @@ describe("ChangeMerging", () => {
     IModelJsFs.copySync(seedFileName, testFileName);
     const upgradedDb = StandaloneDb.openFile(testFileName, OpenMode.ReadWrite);
     createChangeset(upgradedDb);
+    upgradedDb.performCheckpoint();
 
     // Open copies of the seed file.
     const firstFileName = IModelTestUtils.prepareOutputFile("ChangeMerging", "first.bim");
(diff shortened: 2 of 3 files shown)


pmconne:
Does it seem like maybe the native code ought to be enabling WAL mode instead of having to add it to every call site in TypeScript?

grigasp:
> Does it seem like maybe the native code ought to be enabling WAL mode instead of having to add it to every call site in TypeScript?

It does. I'd wait for @kabentley opinion as he chose the pattern to enable WAL mode from TS. I may be wrong, but from previous PRs it seems like we want to have fine control over which iModels use this mode and we don't have the necessary information to make the decision in the addon (e.g. we don't want the mode for snapshot iModels).

core/backend/src/IModelDb.ts
```diff
@@ -2715,6 +2715,7 @@ export class StandaloneDb extends BriefcaseDb {
   public static createEmpty(filePath: LocalFileName, args: CreateEmptyStandaloneIModelProps): StandaloneDb {
     const nativeDb = new IModelHost.platform.DgnDb();
     nativeDb.createIModel(filePath, args);
+    nativeDb.enableWalMode();
```
kabentley, line 2718: this should not be necessary. We enable it in createIModel.

kabentley:
> It does. I'd wait for @kabentley opinion as he chose the pattern to enable WAL mode from TS. I may be wrong, but from previous PRs it seems like we want to have fine control over which iModels use this mode and we don't have the necessary information to make the decision in the addon (e.g. I think we don't want the mode for snapshot iModels).

We enable WAL mode when we create an iModel or when we create a briefcase of an iModel.

core/backend/src/IModelDb.ts
```diff
@@ -2747,6 +2748,8 @@ export class StandaloneDb extends BriefcaseDb {
     const nativeDb = this.openDgnDb(file, openMode, undefined, options);
 
     try {
+      if (openMode === OpenMode.ReadWrite)
```
kabentley, line 2751: why is this necessary? I don't think we should do it. Perhaps you need to modify an existing test iModel?

core/backend/src/test/changesets/ChangeMerging.test.ts
```diff
@@ -43,6 +43,7 @@ describe("ChangeMerging", () => {
     IModelJsFs.copySync(seedFileName, testFileName);
     const upgradedDb = StandaloneDb.openFile(testFileName, OpenMode.ReadWrite);
     createChangeset(upgradedDb);
+    upgradedDb.performCheckpoint();
```
kabentley, line 46: i'm not sure why this is necessary, but probably it should be done in `createChangeset` *before* it copies?

core/backend/src/IModelDb.ts
grigasp: Our test copies a seed iModel and calls `StandaloneDb.openFile` to modify the copy.

kabentley: what seed iModel? It should have WAL mode enabled. If it's from git, I thought i updated it. Is there another one?

grigasp: The seed is delivered with the tests and doesn't have WAL mode enabled. 

I now see what you mean - if we always enable WAL when creating iModels, then we don't need to do that when opening them. But what about "legacy" ones, still using the DELETE mode - shouldn't we "upgrade" them to WAL too?

kabentley: that's why i added enabling WAL to the "SetToStandalone" tool. For tests, you can just run that once and check the file into git.

grigasp: Our team doesn't build/use the `display-test-app`, I wasn't aware the internal script was intended for general use rather than specific use of the test app. 

I converted the seed to WAL mode and am closing this PR.

PR #5010, opened 2023-01-31
Remove `Localization.getLocalizedStringWithNamespace`

mattbjordan:


pmconne:
Is there a specific motivation to remove this now vs in 5.0?

mattbjordan:
> Is there a specific motivation to remove this now vs in 5.0?

@pmconne I don't believe so. Is there a reason you'd want to delay it? The required code change is pretty simple:
```ts
//before
getLocalizedStringWithNamespace("someNamespace", "someKey");

//after
getLocalizedString("someKey", { ns: "someNamespace" });
//or
getLocalizedString("someNamespace:someKey");
```

pmconne:
> Is there a reason you'd want to delay it? The required code change is pretty simple

Our API support policy claims that if an API is deprecated in major version N, we won't remove it until (at least) major version N+2. Technically, that policy applies only to APIs deprecated in 4.x and onward. But my inclination is to adhere to it for 3.x too where feasible.

Application teams are already justifiably apprehensive about adopting a new major release due their experience upgrading from 2.x to 3.x (which some of them *still* have not done, or only very recently, due to all the breaking changes). I want them to have the opposite experience with 4.x (and 5.x, and 6.x, ...).

calebmshafer:
@pmconne that's fair, I just checked internal repos and it's used quite a few times so would require people to make changes.

@mattbjordan mind closing this down? We can push this removal until 5.x.

mattbjordan:
Waiting until 5.x for removal.

PR #5011, opened 2023-01-31
handle long schema names during transformations

MichaelBelousov:
## Details

linked PRs
imodel-native: https://github.com/iTwin/imodel-native/pull/137

I plan on backporting this to 3.6 as a bugfix for the transformation service to consume

## What

- if a schema name is too long during export from the source, write it with a custom name, remember a mapping of the schema names to custom names. Then when we import, use a custom schema locator that is aware of the mapping to resolve references to the long-named schemas.

## Why

For handling long schema names in the transformer.

linux + mac common file systems usually have a 255 byte limit to path segments and therefore to filenames. This means that exporting a schema with a name of length > `255 - '.ecschema.xml'.length` could fail with an error on those operating systems. Transformation often exports schemas from the source and imports them into the target, which means certain transformations of models with certain schemas fail on linux agents.

This is especially noticeable in certain dgn user-defined schemas if generated from a non-ascii    label, a hex encoded utf16 string with `____` between every wide character is used to construct the schema name. This gets too long for 255 bytes very quickly. That is a summary to the best of my knowledge but @ColinKerr might know more about what/why for this. 

In the accompanying [imodel-native PR](https://github.com/iTwin/imodel-native/pull/137), I added options for exporting schemas with custom names (e.g. shorter ones), and importing schemas with a custom schema locator.


core/transformer/src/IModelTransformer.ts
```diff
@@ -1132,10 +1132,35 @@ export class IModelTransformer extends IModelExportHandler {
     return Semver.gt(`${schemaKey.version.read}.${schemaKey.version.write}.${schemaKey.version.minor}`, Schema.toSemverString(versionInTarget));
   }
 
+  private _longNamedSchemasMap = new Map<string, string>();
+
   /** Override of [IModelExportHandler.onExportSchema]($transformer) that serializes a schema to disk for [[processSchemas]] to import into
    * the target iModel when it is exported from the source iModel. */
   public override async onExportSchema(schema: ECSchemaMetaData.Schema): Promise<void> {
-    this.sourceDb.nativeDb.exportSchema(schema.name, this._schemaExportDir);
+    const ext = ".ecschema.xml";
```
ColinKerr, line 1140: would it be simpler if you always serialized the schemas with the fixed names instead of only doing so if the names were too long.  

Serializing to disk is unnecessary over all with a custom context you could just wrap the call to get as xml string from the target iModel in a call to get the schema by name in the context.  Maybe this is not ideal because it would be harder to dump to disk for debug purposes.

Either way unless there is some real downside to using a single method for all schemas I think that is better.  The code just has fewer alternative paths to do the same thing

MichaelBelousov: I will look into removing the on disk paths entirely, and potentially redraft these PRs. That sounds much better.

MichaelBelousov:
closing to investigate transforming schemas without exporting them to disk and reimporting

core/transformer/src/IModelTransformer.ts
MichaelBelousov: Unfortunately it's not currently possible without breaking the transformation service's  usage of the on-disk schemas. They modify some dynamic schemas while it's on disk and before importing. It is also not the bottleneck in transformation so isn't necessarily worthwhile.

I will however provide a way to access the path of the written file, in case it's name was truncated for being too long, for them to use because of this change where the written name is no longer guaranteed to be `` `${schema.name}.ecschema.xml` ``

ColinKerr (approved):

mindaugasdirg (approved):

mindaugasdirg (approved):

core/transformer/src/IModelExporter.ts
```diff
@@ -114,11 +123,13 @@ export abstract class IModelExportHandler {
...
 
-  /** Called when an schema should be exported.
+  /** Called when a schema should be exported.
    * @param schema The schema to export
    * @note This should be overridden to actually do the export.
+   * @note return an [[ExportSchemaResult]] with a `schemaPath` property to notify overrides that call `super`
+   *       where a schema was written for import.
    */
-  public async onExportSchema(_schema: Schema): Promise<void> { }
+  public async onExportSchema(_schema: Schema): Promise<void | ExportSchemaResult> { }
```
pmconne, line 132: Motivation for using `void | ExportSchemaResult` instead of more idiomatic `ExportSchemaResult | undefined`?

MichaelBelousov: the motivation was so that existing subclasses returning void don't need to add a `return undefined` to their overrides when consuming this update. I will go test what happens if I make it `| undefined`

core/transformer/src/IModelTransformer.ts
```diff
@@ -1132,10 +1133,44 @@ export class IModelTransformer extends IModelExportHandler {
...
+    let schemaFileName = schema.name + ext;
+    // many file systems have a max file-name/path-segment size of 255, so we workaround that on all systems
+    const systemMaxPathSegmentSize = 255;
+    if (schemaFileName.length > systemMaxPathSegmentSize) {
+      // this name should be well under 255 bytes
+      // ( 100 + (Number.MAX_SAFE_INTEGER.toString().length = 16) + (ext.length = 13) ) = 129 which is less than 255
+      // You'd have to be past 2**53-1 (Number.MAX_SAFE_INTEGER) long named schemas in order to hit decimal formatting,
+      // and that's on the scale of at least petabytes. `Map.prototype.size` shouldn't return floating points, and even
+      // if they do they're in scientific notation, size bound and contain no invalid windows path chars
+      schemaFileName = `${schema.name.slice(0, 100)}${this._longNamedSchemasMap.size}${ext}`;
```
pmconne, line 1157: Is it guaranteed here that `schema.name` is not already present in `_longNamedSchemasMap`?

core/transformer/src/test/standalone/IModelTransformer.test.ts
```diff
@@ -2024,6 +2026,91 @@ describe("IModelTransformer", () => {
     noSystemSchemasTransformer.dispose();
   });
 
+  it("handles long schema names and references to them", async function () {
+    const longSchema1Name = `ThisSchemaIs${"Long".repeat(100)}`;
+    assert(Buffer.from(longSchema1Name).byteLength > 255);
+    const longSchema2Name = `${longSchema1Name}ButEndsDifferently`;
+
+    if (process.platform !== "win32") {
+      // windows has no bound on path segment (file name) length, (it does have a bound on total path length),
```
pmconne, line 2035: I think you changed this to limit to 255 chars on all platforms (an improvement IMO) so no need to check here?

core/transformer/src/IModelTransformer.ts
MichaelBelousov: yes, although not directly checked for. It's not possible to have schemas of the same name in the source, and that map is cleared before and after calls to `processSchemas`, (although the transformer is not actually designed for multiple uses anyway)

core/transformer/src/test/standalone/IModelTransformer.test.ts
MichaelBelousov: well this just proves that you can't write >255 byte file names on a platform level, it's not exercising my logic to limit it for schemas during export, so on windows this won't fail (there is no path segment limit), and removing that condition will fail the test on windows.

This assert below is more for asserting evidence of the source issue, I would prefer to keep it.

core/backend/src/IModelDb.ts
```diff
@@ -147,6 +148,20 @@ export interface LockControl {
   releaseAllLocks(): Promise<void>;
 }
 
+/**
+ * Options for the importing of schemas
+ * @public can't make this internal since it's a parameter in a public API,
```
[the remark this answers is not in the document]
MichaelBelousov: fixed

simsta6 (approved):

PR #5022, opened 2023-02-02
Make `webgl-compatibility` a direct dependency in `core-frontend`

GytisCepk:
To decrease number of peer dependencies in `core-frontend` package, `webgl-compatibility` is now a direct dependency rather than a peer dep. 

`webgl-compatibility` is only used for types and enums with a few exceptions:

- `queryRenderCompatibility` - a pure function.
- `Capabilities` - class with some statics:
  - `Capabilities.create` - a pure static method.


core/frontend/package.json
```diff
@@ -99,6 +98,7 @@
     "@itwin/object-storage-core": "~1.4.0",
     "@itwin/core-i18n": "workspace:*",
     "@itwin/core-telemetry": "workspace:*",
+    "@itwin/webgl-compatibility": "workspace:*",
```
aruniverse, line 101: mind removing the direct dep we have on this in display-test-app?

GytisCepk: Removed.

aruniverse: weird, ci is complaining about missing dep on it now, but technically it would be installed anyway as a transitive dep. im unable to repro the build failure locally, but i guess since we do have an import from this pkg we should leave it in. going to drop your last commit

aruniverse (approved):
fyi @pmconne

pmconne (approved):

PR #5000, opened 2023-01-30
Yashasvi/eslint deprecated tags

yashincontrol:
Custom ESLint rule that enforces version and at least a sentence description for @deprecated tags.

PR #5032, opened 2023-02-06
Make no-internal work in workspaces/cospaces/monorepos and add tests

MichaelBelousov:
With these changes, the eslint rule no-internal and the no-internal-report tool can work inside a monorepo (such as on our test apps), or inside a cospace. There were two issues in the way, resolution of the internal formatter module path, and `isLocalFile` being limited to thinking a path with `node_modules` is non-local which doesn't work in linked (e.g. workspaced) dependencies. You do have to turn on the new rule config option `dontAllowWorkspaceInternal: true` for the linter to not allow @internal usage inside the repo (but allow local internal usage).

Added some basic tests for the no-internal rule, deduplicated some test utilities from `no-internal-barrel-imports`, and copied the `.mocharc.js` from #5000

the newer functionality is still really heuristic. An ideal implementation would require knowledge of which packages it may use `@internal` api's from, and check from where references are imported to determine if it can use those apis.


paulius-valiunas (approved):

paulius-valiunas:
I think the first iteration of this rule was implemented using a hardcoded list of imodeljs packages, but that list soon became outdated. There might be more edge cases with different package managers, so I'm debating if a whitelist in the config (instead of blacklist like last time) would be a better choice. So if you're working in a monorepo that's consuming iTwin.js packages, you would still have to add all the package names in your repository to the config, but you would get false positives that remind you to keep it up to date, instead of false negatives that always go unnoticed.

MichaelBelousov:
> So if you're working in a monorepo that's consuming iTwin.js packages, you would still have to add all the package names in your repository to the config, but you would get false positives that remind you to keep it up to date, instead of false negatives that always go unnoticed.

In this PR you just have to add `dontAllowWorkspaceInternal: true` to the eslint config of each package that doesn't participate. Since you can use eslint overrides, you can apply that to any pattern of files and multiple packages at once in one override. Not as explicit though.

I did personally add this so that I could run `pnpm exec no-internal-report` on packages like the transformer and test-apps inside the monorepo itself to judge their usage of @internal APIs

MichaelBelousov:
Also because there is likely some conflict and it got here first, I will wait for #5000 to complete to merge this

PR #5033, opened 2023-02-06
dont publish appui & presentation pkgs

aruniverse:
https://rushjs.io/pages/configs/rush_json/
image

common/config/azure-pipelines/templates/publish.yaml
image

https://rushjs.io/pages/commands/rush_publish/
image

PR #5050, opened 2023-02-09
Fix Popout Widget window title for about:blank

raplemie:
fixes iTwin/itwinjs-backlog#638

When the content is written to the popout, update the title regardless of the url used.


raplemie:
@aruniverse Just want to let you know that #5033 disabled the `rush change` to pick up for AppUi packages. Not sure if we need to add them manually or we just pass for 3.7.

NancyMcCallB (approved):

PR #5052, opened 2023-02-10
rename `CodeService` members to avoid confusion with `Code`

kabentley:
CodeService deals with reserved and proposed codes by CodeSpecName and CodeScopeGuid. Rename the members to avoid confusion with the similar iModel-local values in `Code`.

There is a corresponding PR in the CodeService repo.


pmconne (approved):
Need extract-api and rush change.

PR #5071, opened 2023-02-14
Fix `betools docs` to find correct relative path from root to package source

saskliutas:
Previous implementation to find relative path from root to package source folder was very sensitive to folder layout. It was depending on that root folder name matches `name` value in root package.json file and that path up to root folder does not contain it.

Before changes `packageRoot` was resolved to:
`D:\itwinjs-core\itwin-1\itwinjs-core\core\backend\src` -> `itwin-1\itwinjs-core\core\backend\src`

After changes it is resolved correctly:
`D:\itwinjs-core\itwin-1\itwinjs-core\core\backend\src` -> `core\backend\src`

### DIFF
--- a/tools/build/scripts/utils/addSourceDir.js
+++ b/tools/build/scripts/utils/addSourceDir.js
@@ -17,15 +17,15 @@ if (process.env.RUSHSTACK_FILE_ERROR_BASE_FOLDER) {
   rootPackageJsonPath = path.resolve("../../../../package.json");
 }
 
-const rootPackageJson = require(rootPackageJsonPath);
+// Check if path to root package.json is valid.
+require(rootPackageJsonPath);
 
 // Appends the directory of the package root to the Typedoc JSON output
 function addSourceDir(file, directory) {
   if (FS.existsSync(file) && FS.statSync(file).isFile()) {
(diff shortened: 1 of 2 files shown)


tools/build/scripts/utils/addSourceDir.js
```diff
@@ -17,15 +17,15 @@ if (process.env.RUSHSTACK_FILE_ERROR_BASE_FOLDER) {
   rootPackageJsonPath = path.resolve("../../../../package.json");
 }
 
-const rootPackageJson = require(rootPackageJsonPath);
+// Check if path to root package.json is valid.
+require(rootPackageJsonPath);
```
aruniverse, line 21: does it make sense to maybe use `require.resolve`? 
https://nodejs.org/api/modules.html#requireresolverequest-options?

ben-polinsky (approved):

tools/build/scripts/utils/addSourceDir.js
saskliutas: Changed to use `require.resolve` instead of `path.resolve`

aruniverse (approved):

PR #5076, opened 2023-02-15
Presentation: localization clean up

MartynasStrazdas:
Closes: https://github.com/iTwin/itwinjs-backlog/issues/536
Native changes: https://github.com/iTwin/imodel-native/pull/166
imodel-native: https://github.com/iTwin/imodel-native/tree/Presentation/LocalizationCleanUp


grigasp (changes requested):
should bump major RPC interface version

presentation/backend/src/presentation-backend/Utils.ts
```diff
@@ -13,17 +13,20 @@ import {
...
 /** @internal */
 export function getLocalizedStringEN(key: string) {
+  let result = presentation;
   const [namespace, identifier] = key.split(":", 2);
-  if (namespace === "ECPresentation") {
-    return ecPresentation[identifier];
-  }
-  if (namespace === "RulesEngine") {
-    return rulesEngine[identifier];
+  if (namespace === "Presentation") {
```
saskliutas, line 22: I think it would be nice to invert this `if` statement to return if namespace is not `Presentation`. In that case `for` loop can be indented less.

saskliutas (changes requested):

saskliutas (approved):

PR #4295, opened 2022-09-16
node 18 support

ben-polinsky:
- Add Node 18.x to ci builds

#4308 
Fixes https://github.com/iTwin/itwinjs-core/issues/4135

PR #5077, opened 2023-02-15
Drop Node 16 support

GytisCepk:
### Node 16 will reach End-of-Life later this year ([Node.js release schedule](https://github.com/nodejs/release#release-schedule)).

PR Changes:

- Change ranges to support only Node 18.
- Remove Node 16 runs from integration and regressions pipelines.
- Update documentation.
- Update `object-storage` packages versions to `^1.5.0`.


MichaelBelousov (approved):
transformer changes are fine

aruniverse (approved):

saskliutas:
`@itwin/core-frontend` dependency [`"@itwin/cloud-agnostic-core": "~1.4.0"`](https://github.com/iTwin/itwinjs-core/blob/742071bf231c5ecd029c14d42148b7b91e1cab4a/core/frontend/package.json#L97) does not support Node 18. Shouldn't it be updated before this?

ben-polinsky (approved):

saskliutas (approved):

paulius-valiunas (approved):

common/config/azure-pipelines/jobs/regression-testing.yaml
```diff
@@ -43,22 +43,6 @@ jobs:
           platform: Darwin
           name: $(mac_pool)
           node_version: 18.12.0
-
```
pmconne, line 46: Comments at top of file are out of date, if it matters.

pmconne (approved):

core/backend/package.json
```diff
@@ -6,7 +6,7 @@
   "typings": "lib/cjs/core-backend",
   "license": "MIT",
   "engines": {
-    "node": ">=16.13.0 < 19.0"
+    "node": "^18.0.0"
```
aruniverse, line 9: maybe we should actually set the engine to be ^18.12.0; which is what we're saying is the minimum compatible version in some of the documentation, and the testing pipelines

core/backend/package.json
```diff
@@ -92,9 +92,9 @@
   "dependencies": {
```
aruniverse, line 92: could you also try updating the types/node dev dep to use ^ for all pkgs?

aruniverse (approved):

core/backend/package.json
GytisCepk: Used [this discussion](https://github.com/iTwin/itwinjs-core/pull/4295#discussion_r1005113967) as a reference to what range this should be set. In docs, used `^18.12.0` to encourage using LTS versions. 

`^18.12.0` would make thinks more consistent.

aruniverse: thx for bringing that back up, i guess based on calebs comment we should leave as is then

core/backend/package.json
GytisCepk: Updated

PR #5083, opened 2023-02-16
speed up docs build

aruniverse:
dont build the entire monorepo only the library

### DIFF
--- a/common/config/azure-pipelines/jobs/docs-build.yaml
+++ b/common/config/azure-pipelines/jobs/docs-build.yaml
@@ -53,7 +53,7 @@ jobs:
         displayName: rush audit
         workingDirectory: ${{ parameters.workingDir }}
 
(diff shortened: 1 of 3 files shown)


common/config/azure-pipelines/jobs/docs-build.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
         displayName: rush audit
         workingDirectory: ${{ parameters.workingDir }}
 
-      - script: node common/scripts/install-run-rush.js build:ci
+      - script: node common/scripts/install-run-rush.js build:ci --to-version-policy prerelease-monorepo-lockStep
```
aruniverse, line 56: i think theoretically we could just do build instead of build:ci

skirby1996 (approved):

PR #5084, opened 2023-02-16
Run rush audit before bumping version

skirby1996:
It is possible that between the latest commit on a branch, and when the version bump commit is created, a CVE could have been introduced causing the version bump CI build to fail. This PR updates the pipeline to run `rush audit` before creating a version bump commit as a validation step to avoid this.


aruniverse (approved):

aruniverse:
Could you also update audit.js to remove the last advisory we added, since it was bogus and withdrawn?

PR #5088, opened 2023-02-16
Transformer handle untrimmed whitespace in code values of pre-inserted elements

MichaelBelousov:
Wrapping a `CodeProps` with a `Code` class makes an undefined code value appear as the empty string using its getter, but also trims whitespace which is [allegedly invalid](https://github.com/iTwin/itwinjs-core/blob/bb19163f6cd9b500444adc010d62ef1cfc4b9090/core/common/src/Code.ts#L43).

Don't use the `Code` class when trying to make the Code value not undefined while handling pre-inserted elements (e.g. subcategories) in the transformer, since those elements are queried by code.

In the future if we would like to make this requirement harder, we should consider trimming or rejecting code values with untrimmed white space on the native side so native-only connectors do not produce this kind of data. But these models already exist so the transformer should not mess up their codes. Or we could remove the requirement. Codes are supposed to be human-readable so I can see why we'd want it, but I'm sure us enforcing it without warning/rejection as we do now might cause confusion.

Need to backport this to 3.6 @skirby1996


MichaelBelousov:
this may be because the native addon's definition of `Trim` uses libc's ascii-only `isspace` while javascript's `String.prototype.trim` is unicode aware and trims out things like `\0xa0` (non-breaking space) which is non-ascii. I will add more unicode spaces to the test, but we probably want to update the definition of trim in the native addon to support all unicode whitespace

ViliusRuskys (approved):

mindaugasdirg (approved):

ColinKerr (approved):

PR #5094, opened 2023-02-17
promote apis for map-layers

eringram:
Promote some map layers APIs to public that are required by the map-layers widget which was moved to [viewer-components-react](https://github.com/iTwin/viewer-components-react/tree/master/packages/itwin/map-layers).


core/common/src/Localization.ts
```diff
@@ -88,14 +88,17 @@ export interface Localization {
 
   /** Register a new Namespace and return a Promise that is fulfilled when the content is loaded.
    * If the namespace is already registered, its Promise will be returned.
-   * @param name - the name of the namespace.
+   * @param namespace - the name of the namespace.
```
pmconne, line 91: @ben-polinsky docs build doesn't catch mismatched names between doc tag and function signature?

ben-polinsky: It doesn't seem we've written anything to validate the mismatch. However, there is an eslint rule: https://github.com/gajus/eslint-plugin-jsdoc/blob/main/.README/rules/check-param-names.md

eringram:
@mdastous-bentley As you review this, I would especially appreciate if you have any suggestions for improvement to the documentation for MapLayerImageryProvider and MapLayerFormatRegistry

common/api/core-frontend.api.md
```diff
@@ -14018,10 +14010,10 @@ export abstract class Viewport implements IDisposable, TileUser {
     getFrustum(sys?: CoordSystem, adjustedBox?: boolean, box?: Frustum): Frustum;
     // @alpha (undocumented)
     getMapFeatureInfo(hit: HitDetail): Promise<MapFeatureInfo>;
-    // @internal (undocumented)
     getMapLayerImageryProvider(index: number, isOverlay: boolean): MapLayerImageryProvider | undefined;
     // @internal
     getMapLayerIndexesFromIds(mapTreeId: Id64String, layerTreeId: Id64String): MapLayerIndex[];
+    getMapLayerRange(layerIndex: number, isOverlay: boolean): Promise<MapCartoRectangle | undefined>;
```
pmconne, line 14016: Why don't all these function that take `layerIndex` and `isOverlay` take `MapLayerIndex` instead (besides it being `internal`)?

eringram: I don't see any reason they couldn't take that instead. Although many of the functions would still separate out `layerIndex` and `isOverlay` anyway, so wouldn't be simplified much aside from the parameters. @mdastous-bentley do you think this is worth changing/were you guys planning to use `MapLayerIndex` outside of `MapTiledGraphicsProvider`?

core/frontend/src/Viewport.ts
```diff
@@ -837,11 +843,65 @@ export abstract class Viewport implements IDisposable, TileUser {
...
+
+      return MapCartoRectangle.fromRadians(cartoRange.low.x, cartoRange.low.y, cartoRange.high.x, cartoRange.high.y);
+    }
+
+    const imageryProvider = this.getMapLayerImageryProvider(layerIndex, isOverlay);
+    if (undefined === imageryProvider)
+      return undefined;
 
+    try {
+      await imageryProvider.initialize();
```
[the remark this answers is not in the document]
mdastous-bentley: Yes correct.  
To be safe, we might consider checking the `loadStatus` value of `TileTreeOwner` through the MapLayerTileTreeReference object... or even better we add a new `MapLayerImageryProviderStatus`  value when the `MapLayerImageryProvider` has been (correctly?) initialized.

common/api/core-frontend.api.md
eringram: Just made a commit changing this for most functions that take `isOverlay` and `layerIndex`, excluding just [moveMapLayerToIndex ](https://github.com/iTwin/itwinjs-core/blob/eringram/maplayers-remove-internals/core/frontend/src/DisplayStyleState.ts#L596)which takes 2 indices.

core/frontend/src/tile/map/MapLayerImageryProvider.ts
```diff
@@ -19,29 +19,34 @@ const tileImageSize = 256, untiledImageSize = 256;
...
 export enum MapLayerImageryProviderStatus {
   Valid,
   RequireAuth,
 }
 
-/** Base class for map layer imagery providers.
- * @internal
+/** Abstract class for map layer imagery providers.
+ * Map layer imagery providers request and provide tile images and other data. Each map layer from a separate source needs its own imagery provider object.
+ * @public
```
mdastous-bentley, line 32: I see several publics methods exposed here, do we really need all of them ?  I know we needed at least the status stuff...

eringram: Do you mean some of the getters like lines 51-60? No, I don't think they need to be exposed-- I overlooked the fact that they inherit the public signature. I'll look through if any of them are used in other packages and mark them internal otherwise.

Mark S. brought up a good point which is that if `MapLayerImageryProvder` is made public and meant to be extended to create other imagery providers, eventually people might want to override getters such as `minimumZoomLevel`, etc.

pmconne: > expose only stuff needed by map-layers widget?

Yes.

markschlosseratbentley: Yes that sounds reasonable.

eringram: ### APIs needed by map-layers/other external packages
(all should have been promoted or refactored in this PR by now)

All those listed [here](https://github.com/iTwin/itwinjs-backlog/issues/622#issuecomment-1416385788)
All those in MapLayerImageryProvider.ts that I didn't mark as internal earlier this week:
- Used in map-layers:
    - mutualExclusiveSubLayer 
- Used in map-layers-formats:
    - maximumZoomLevel
    - minimumZoomLevel
    - cartoRange
    - constructUrl
    - loadTile
- Used in display-test-app:
    - tilingScheme

core/common/src/Localization.ts
```diff
@@ -88,20 +88,28 @@ export interface Localization {
...
    * @note - The registerNamespace method starts fetching the appropriate version of the JSON localization file from the server,
    * based on the current locale. To make sure that fetch is complete before performing translations from this namespace, await
    * fulfillment of returned Promise.
    * @see [Localization in iTwin.js]($docs/learning/frontend/Localization.md)
    */
   registerNamespace(namespace: string): Promise<void>;
-  /** @internal */
+
+  /** Unregister a namespace.
+   * @param namespace - the name of the namespace.
```
pmconne, line 100: @mattbjordan any objection to moving this from `internal` to `public`?

pmconne (approved):

core/common/src/Localization.ts
mattbjordan: Nope, sounds fine to me.

core/frontend/src/tile/map/MapLayerImageryProvider.ts
mdastous-bentley: mutualExclusiveSubLayer should be marked as beta, I will probably want to revisit this at some point.

PR #3962, opened 2022-07-25
Aidantran/no render app fix

AidanTran:
Trying to remove the necessity to mock requestAnimationFrame globally when running NoRenderApp in tests outside of certa. 
NOT CURRENTLY TESTED
Having trouble testing this as current RPC tests are run through certa.

PR #5098, opened 2023-02-19
Batch reprojection requests

### LINKED ISSUE #5097: Batch reprojection requests
As described in https://github.com/iTwin/itwinjs-backlog/issues/641, when displaying reality models there can be many small, frequent requests to the backend to reproject tile centers. This appears to negatively impact performance.

Batch requests without negatively affecting performance.

None really.

**Additional context**
For reality models, for each parent tile, `reprojectAndResolveChildren` makes one request to reproject the center of each child tile's volume. Many such requests are observed during a single traversal of the tile tree for tile selection, within one frame. It would be preferable to send at most one request per frame.

For map tiles, there's some pre-loading of reprojected points that are expected to be needed soon; when they are actually needed, they are expected to be extracted directly (and synchronously) from the GeoConverter's cache. This can still involve multiple requests per frame. It's possible that the pre-loading optimization may become obsolete after batching is implemented.

pmconne:
Fixes  #5097. Part of https://github.com/iTwin/itwinjs-backlog/issues/641.

GeoConverter and its associated caches are some of the oldest bits of code in iTwin.js (back when it was still called iModel.js).
I replaced the guts of it with an implementation that ensures:
- There is never more than one reprojection request in flight at a time; and
- It never makes more than one request per frame; and
- It never requests reprojection of points already requested (whether in flight or in cache); and
- It never request the same point more than once.

There remains some awkwardness around the types used in the API:
- `XYZProps` is a `PITA`.
- `PointWithStatus` might make more sense as `XYAndZ & { s: GeoCoordStatus }`.
- The `fromCache` field of the response exists chiefly for tests.

I kept the "pre-load then look up in cache" behavior for `MapTileTree` for now. It's probably no longer relevant. Those previously-per-tile requests will now be batched.

Note, each GeoConverter is independent, and each RealityTileTree has its own GeoConverter.

Incidentally adds a `slice` method to `SortedArray`.

TODO:
- [x] Run ImageTests


pmconne:
@wgoehrig @calebmshafer Maybe you remember why @AidanTran felt we needed to [avoid using requestAnimationFrame](https://github.com/iTwin/itwinjs-core/pull/3962)?

pmconne:
I made three views of the reality model referenced in https://github.com/iTwin/itwinjs-backlog/issues/641 and made `IModelReadRpcInterface.getIModelCoordinatesFromGeoCoordinates` report the following metrics on each call: total number of points requested, min and max number of points in a single request, and total number of requests; then compared master to this branch.
On both branches, the same total number of points are requested, but on master the maximum per request never exceeds 32, vs 300 (the absolute maximum permitted) on this branch.
Number of calls to `getIModelCoordinatesFromGeoCoordiantes:
| View  | Master | Branch |
| ------------- | ------------- | --- |
| Fitted, perspective  | 64  | 23 |
| Zoomed in, perspective | 808  | 53 |
| Zoomed in, orthographic | 916  | 74 |

AlainRobertAtBentley (approved):

MarcBedard8 (approved):

markschlosseratbentley (approved):

core/frontend/src/GeoServices.ts
```diff
@@ -3,240 +3,287 @@
...
-    const result: Array<PointWithStatus | undefined> = [];
-    let missing: XYZProps[] | undefined;
-    for (const geoPoint of geoPoints) {
-      const key = JSON.stringify(geoPoint);
-      const imodelCoord = this._cache[key];
-      result.push(imodelCoord);
-      if (undefined === imodelCoord) {
-        if (undefined === missing)
-          missing = [];
+function cloneXYAndZ(xyz: XYAndZ): XYAndZ {
```
aruniverse, line 32: do we really need to add this? couldnt we just use the spread operator?

aruniverse (approved):

core/frontend/src/GeoServices.ts
pmconne: See comment in description about `XYZProps` is a `PITA`. It can be an array or an object {x, y, z }. Any of the array elements or the object properties can be undefined.

aruniverse: Since its an internal/local function it's ok and no objections to that; i see you're just passing it into cloneFunction of the Arrays so nbd there

pmconne:

pmconne:
@Mergifyio backport release/3.5.x

pmconne:
@Mergifyio backport release/3.6.x

PR #5103, opened 2023-02-21
Deprecate IModelAppOptions.rpcInterfaces

swbsi:


swbsi:

test-apps/presentation-test-app/src/backend/web/BackendServer.ts
```diff
@@ -18,11 +18,12 @@ export default async function initialize(rpcInterfaces: RpcInterfaceDefinition[]
   await IModelHost.startup();
 
   // tell BentleyCloudRpcManager which RPC interfaces to handle
-  const rpcConfig = BentleyCloudRpcManager.initializeImpl({ info: { title: "presentation-test-app", version: "v1.0" } }, rpcInterfaces);
+  const paramsHolder = BentleyCloudRpcParams.wrap({ info: { title: "presentation-test-app", version: "v1.0" } });
```
grigasp, line 21: `BentleyCloudRpcParams` is `@internal` at the moment - shouldn't it be changed to `@public` if you're using it in test apps and example code?

swbsi: Making it public would force most of the internal parts of the RPC system to become public also. We are trying to deprecate the RPC system right now. Making more and more of it public would push that goal further out of reach.

I think my approach (BentleyCloudRpcParamsHolder) is the right way to bridge us to the future, here, especially since BentleyCloudRpc-anything is in fact an internal concept. The next RPC/REST/whatever system should have a more-generic way for apps to connect, that doesn't assume they are using Bentley's cloud infrastructure.

grigasp: I understand the goal, but deprecation should generally by done by first introducing the new approach, then deprecating the old one. But let's ignore that.

At the moment RPC system is partly `@public @deprecated` and partly `@internal`. If we have to choose between these two evils, I believe `@internal` is worse, as it hides the API from the docs. Maybe, at least, the APIs we expect consumers to call could be made `@public @deprecated` rather than `@internal`? I'm specifically referring to `BentleyCloudRpcParams` here. If some of it's attributes should be kept internal, you can tag them `@internal` individually, but I think `info` and `uriPrefix` could be made public, couldn't they?

swbsi: I see your point about info and uriPrefix. Let me try to come up with something for those…

core/common/src/rpc/web/BentleyCloudRpcManager.ts
```diff
@@ -46,18 +66,33 @@ export abstract class BentleyCloudRpcConfiguration extends RpcConfiguration {
...
+
 /** Coordinates usage of RPC interfaces for Bentley cloud deployments.
- * @internal
+ * @public
  */
 export class BentleyCloudRpcManager extends RpcManager {
   /** Initializes BentleyCloudRpcManager for the frontend of an application. */
-  public static initializeClient(params: BentleyCloudRpcParams, interfaces: RpcInterfaceDefinition[], routing: RpcRoutingToken = RpcRoutingToken.default): BentleyCloudRpcConfiguration {
-    return BentleyCloudRpcManager.performInitialization(params, interfaces, routing);
+  public static initializeClient(paramsHolder: BentleyCloudRpcParamsHolder, interfaces: RpcInterfaceDefinition[], routing: RpcRoutingToken = RpcRoutingToken.default): BentleyCloudRpcConfigurationHolder {
```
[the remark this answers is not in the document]
swbsi: My latest changes allow usage of the "necessary" parts of BentleyCloudRpcParams without any changes. The "internal" parts have moved to a derived params type.

swbsi:
> > Making it public would force most of the internal parts of the RPC system to become public also. We are trying to deprecate the RPC system right now. Making more and more of it public would push that goal further out of reach.
> 
> Instead of _"promoting"_ from `internal` to `public`, would it make more sense for some of these things to be _"promoted"_ to `beta`? Especially if the end goal is to deprecate and completely refactor the entire RPC system, which if we make things `public` in version `N`, then deprecate in `N+1`, then can only remove in `N+2` (depending on if it was deprecated at the beginning of the new major) or `N+3`

In an ideal world, I think maybe.

As it is, these new publics are joining the existing RPC public api subset (RpcInterface, etc). All of that will have to be deprecated correctly once the system replacement is ready…

test-apps/presentation-test-app/src/backend/web/BackendServer.ts
grigasp: Thanks, I think with your latest changes it's much better.

ben-polinsky (approved):

core/mobile/src/frontend/MobileApp.ts
```diff
@@ -61,7 +61,7 @@ export class MobileApp {
         authorizationClient.setAccessToken(accessToken, expirationDate);
       });
 
-      const rpcInterfaces = opts?.iModelApp?.rpcInterfaces ?? [IModelReadRpcInterface, IModelTileRpcInterface];
+      const rpcInterfaces = opts?.iModelApp?.rpcInterfaces ?? [IModelReadRpcInterface, IModelTileRpcInterface]; // eslint-disable-line deprecation/deprecation
```
aruniverse, line 64: tangential; i find it weird we have default rpcs to register on Mobile but not on any other platform afaik

cc @itwin/mobilesdkadmins

aruniverse: this shouldnt hold the pr up, just making a comment to keep track of this

core/common/src/rpc/web/BentleyCloudRpcManager.ts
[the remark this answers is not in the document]
aruniverse: But the signature is still different (different return type) and the `.initializeClient` and `.initializeImpl` dont actually initialize the rpc manager on frontend or backend anymore and users will need to do that themselves now.

aruniverse (commented):
@paulius-valiunas mind taking a look from a backend perspective? How much will this pr break gpb?

paulius-valiunas:
> @paulius-valiunas mind taking a look from a backend perspective? How much will this pr break gpb?

GPB uses `BentleyCloudRpcManager.initializeImpl`, which doesn't seem to be deprecated here and its argument types are kept compatible. I don't think it will break anything.

core/common/src/rpc/web/BentleyCloudRpcManager.ts
```diff
@@ -18,20 +18,46 @@ import { RpcRoutingToken } from "../core/RpcRoutingToken";
...
+  protocol?: typeof BentleyCloudRpcProtocol;
   /** Handler for RPC request events. */
   pendingRequestListener?: RpcRequestEventHandler;
 }
 
-/** Operating parameters for Bentley cloud RPC interface deployments.
+/** Holds initialization parameters for BentleyCloudRpcConfiguration.
+ * @public
+ */
+export class BentleyCloudRpcParamsHolder {
```
paulius-valiunas, line 43: what's the purpose of this class?

core/common/src/rpc/web/BentleyCloudRpcManager.ts
```diff
@@ -46,21 +72,44 @@ export abstract class BentleyCloudRpcConfiguration extends RpcConfiguration {
...
-    return BentleyCloudRpcManager.performInitialization(params, interfaces, routing);
+  public static initializeClient(params: BentleyCloudRpcParams | BentleyCloudRpcParamsHolder, interfaces: RpcInterfaceDefinition[], routing: RpcRoutingToken = RpcRoutingToken.default): BentleyCloudRpcConfigurationHolder {
+    const configuration = BentleyCloudRpcManager.performInitialization(this.resolve(params), interfaces, routing);
+    return new BentleyCloudRpcConfigurationHolder(configuration);
   }
 
   /** Initializes BentleyCloudRpcManager for the backend of an application. */
-  public static initializeImpl(params: BentleyCloudRpcParams, interfaces: RpcInterfaceDefinition[]): BentleyCloudRpcConfiguration {
-    return BentleyCloudRpcManager.performInitialization(params, interfaces);
+  public static initializeImpl(params: BentleyCloudRpcParams | BentleyCloudRpcParamsHolder, interfaces: RpcInterfaceDefinition[]): BentleyCloudRpcConfigurationHolder {
```
paulius-valiunas, line 107: I just noticed that this doesn't _actually_ initialize anything anymore...

core/common/src/rpc/web/BentleyCloudRpcManager.ts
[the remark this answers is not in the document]
aruniverse: nvm, i missed you still called performInitialization

core/common/src/rpc/web/BentleyCloudRpcManager.ts
swbsi: To wrap the internal params that a consumer might still be using.

I moved BentleyCloudRpcParam’s two optional params to an internal interface, since publicizing them would have rippled through almost the entire RPC api.

paulius-valiunas: A public class without a public constructor doesn't make sense. You're doing crazy workarounds to avoid (strictly speaking) violating our policies but this only makes the user experience much worse.

swbsi: The alternative is to publicize basically the entire rpc system. So, that’s the value of this workaround.

I think it’s the right trade off.

paulius-valiunas: so you think having a weird workaround that no one will understand is better than just leaving this public and _not_ rippling it through all RPC code?

aruniverse: can we get a way with making this `@beta`?
not a fan of introducing temporary solutions and marking them as `@public`

swbsi: That sounds like a good idea. I will try it…

swbsi: I switched them to beta.

core/common/src/RpcInterface.ts
```diff
@@ -21,10 +21,10 @@ import { RpcNotFoundResponse } from "./rpc/core/RpcControl";
 
 /* eslint-disable deprecation/deprecation */
 
-/** @internal */
+/** @public */
```
calebmshafer, line 24: Can you add documentation to both of these interfaces?

core/common/src/rpc/web/BentleyCloudRpcManager.ts
calebmshafer: If we're going to make it `beta` any reason we can't stick with the current API without the wrapper and then mark the field we don't want to expose as `internal`? Most people won't use it but for those that are doing it today (I know I've used the protocol in a few different cases).

Also, I've seen in a few other files that were changes we need to use the protocol to get the port for setting up the backend. Do we have an alternative for that?

core/mobile/src/frontend/MobileApp.ts
calebmshafer: agreed, if we want to make that change we should do it for 4.0.  Mind creating an issue if there isn't already one?

calebmshafer (commented):
@swbsi should we be marking `BentleyCloudRpcManager` and other related classes as public if we're looking to refactor the whole system in 5.0? I would have expected `beta` tag instead.

Otherwise, this looks good to me.

core/common/src/rpc/web/BentleyCloudRpcManager.ts
```diff
@@ -46,18 +48,34 @@ export abstract class BentleyCloudRpcConfiguration extends RpcConfiguration {
   public abstract override readonly protocol: BentleyCloudRpcProtocol;
 }
 
+/** Operating parameters for Bentley cloud RPC interface deployments.
+ * @beta This is only required to support accessing internal state returned from BentleyCloudRpcManager.
+ * Most use cases can be supported without accessing the internal members of this class.
+ */
+export class BentleyCloudRpcConfigurationHolder {
```
calebmshafer, line 55: @swbsi, if the other classes are all marked `beta` I don't think we need to hide the RpcConfiguration anymore and instead just use the existing class and refactor in 5.0.

swbsi: It’s purpose was to prevent the release level of BentleyCloudRpcManager from rippling through the rest of the api.

swbsi: I realize it doesn’t improve the internal status of RpcConfiguration et al.

paulius-valiunas: I don't really understand what you mean by rippling... could you provide a specific example? Regarding mixing beta with internal, you're already doing it in this class (beta class with internal properties), or did you mean something else?

paulius-valiunas: In that case I would just make `RpcConfiguration` beta, but leave all the properties internal. I would also change everything that was made public in this PR, to beta, because we're probably going to want to remove it all anyway (I'm just guessing here, maybe you have a different plan)

swbsi: Yes, that's basically what Caleb is suggesting. I'm going to try that...

swbsi:
@calebmshafer I think this is ready now :-)

calebmshafer (approved):
LGTM, thanks!

paulius-valiunas (approved):

pmconne (approved):

pmconne:
> All of that will have to be deprecated correctly once the system replacement is ready..

@aruniverse @calebmshafer @swbsi what ever came of these plans for replacing the RPC system?

PR #5137, opened 2023-02-24
add tolerance to RegionOps.regionBooleanXY

dassaf4:
Allow caller to supply (larger) merge tolerance to this method so that imprecise loops with gaps greater than the default micrometer tolerance can be merged successfully, avoiding danglers/spikes that can cause issues with parity analysis downstream.

For example, creating a `UnionRegion` with `RegionOps.regionBooleanXY` by passing in default tolerance and input geometry that has gaps on the order of 1e-5 meters:
image top:image bottom:image
and then passing the output into `RegionOps.constructAllXYRegionLoops` can result in a component with two negative area (outer boundary) loops. However, by passing in a larger tolerance, say 1.0e-4, the same construction sequence results in the expected single boundary loop.

Also: clarify/make consistency changes to angle measurement API.


dassaf4:

saeeedtorabi (approved):

Issue #5029, opened 2023-02-06
Values formatting on the frontend


grigasp:
At the moment all units-related values formatting is done in native addon, which requires us to send both raw and display values from the addon to TS. It seems `itwinjs-core` already supports formatting - we should switch to using TS formatting APIs.

```[tasklist]
- [x] Investigate if `itwinjs-core` values formatting APIs have sufficient support for formatting all our values (https://github.com/iTwin/itwinjs-core/pull/5101)
- [ ] Refactor all our frontend code to format values on the frontend rather than relying on them being formatted on the backend
- [ ] Remove backend values formatting
```

PR #5152, opened 2023-03-01
Presentation: Frontend property formatting

saskliutas:
Added Schema Format lookup based on property KindOfQuantity.

Part of https://github.com/iTwin/itwinjs-core/issues/5029


pmconne (approved):

presentation/common/src/presentation-common/PropertyFormatter.ts
```diff
@@ -79,17 +101,119 @@ export class ContentPropertyValueFormatter extends PropertyValueFormatter {
...
+  return schema.getItem<KindOfQuantity>(propKoqName);
+}
+
+async function getKoqFormat(koq: KindOfQuantity, unitSystem: UnitSystemKey) {
+  const unitSystems = getUnitSystemGroupNames(unitSystem);
+  const presentationFormat = await getKoqPresentationFormat(koq, unitSystems);
+  if (presentationFormat)
+    return presentationFormat;
+
+  return koq.defaultPresentationFormat;
```
grigasp, line 155: should attempt to use persistence unit if it's in requested unit system and only then fall back to default presentation format

presentation/frontend/package.json
```diff
@@ -54,6 +55,7 @@
     "@itwin/core-frontend": "workspace:*",
     "@itwin/core-i18n": "workspace:*",
     "@itwin/core-quantity": "workspace:*",
+    "@itwin/ecschema-metadata": "workspace:*",
```
grigasp, line 58: Do we really need these deps on `presentation-frontend`? Seems like having them on `presentation-common` should be sufficient.

common/changes/@itwin/presentation-common/presentation-frontend_property_formatting_2023-03-01-10-34.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-common",
+      "comment": "Added Format lookup in ECSchema for formatting properties with KindOfQuantity",
```
grigasp, line 5: should probably mention the new peer dep too

grigasp (changes requested):

presentation/common/src/presentation-common/PropertyFormatter.ts
saskliutas: Added format for persistence unit

presentation/frontend/package.json
saskliutas: removed from backend too.

common/changes/@itwin/presentation-common/presentation-frontend_property_formatting_2023-03-01-10-34.json
saskliutas: Added new change file mentioning new peer dep

grigasp (approved):

PR #4845, opened 2022-12-22
API promotions

pmconne:
Many of our APIs have languished in beta or alpha for far too long. Our goal is for any non-internal API to eventually be promoted to public (or abandoned), and for applications to use only `@public` APIs in production. We also have a fair number of non-public APIs defined in one package and used in `@public` APIs within another package - api-extractor does not catch such inconsistencies.

Now is a good time to evaluate and update our API release tags and documentation as we prepare for iTwin.js 4.0.

This PR promotes a batch of no-brainers. Some were de facto part of the public API already by virtue of cross-package usage. See NextVersion.md for a list.

Part of #4851.

PR #5160, opened 2023-03-02
Safeguard readPixels against invalid rects

markschlosseratbentley:
It turns out that `ElementLocateManager.doPick` can create zero-height and/or zero-width `ViewRect`s which get fed to `Viewport.readPixels` if some mouse operation (resize, etc.) occurs too rapidly and the pick point is registered outside the bounds of the viewport. This results in an invalid frustum which causes numerous errors. This used to not cause the system to assert or fall over too obviously. However, the recent redesign of [FrustumPlanes.ts](https://github.com/iTwin/itwinjs-core/pull/4845/files?diff=split&w=0#diff-a5140389b48f54f0d1768cbe5cba9ec5bbe6f1368b38dbb4482f43a2617b8dd8) rightfully made this situation overtly cause an error by returning an empty array of `ClipPlane`s. Previously, a partial array would be returned.

To safeguard against this, a check for a valid view rect has been added to `ElementLocateManager.doPick` - we no longer do a `readPixels` if the view rect is invalid.

Additionally, since `Viewport.readPixels` itself is public API, we will also now check for an invalid view rect there.

This PR also corrects some `ViewRect` documentation.

Fixes https://github.com/iTwin/itwinjs-backlog/issues/640


pmconne (approved):
👍

pmconne:
@mergifyio backport release/3.6.x

aruniverse:

PR #5170, opened 2023-03-07
Batched tiles for spatial views

pmconne:
Working towards https://github.com/iTwin/itwinjs-backlog/issues/657.
Dependent upon [changes to mesh export service](https://dev.azure.com/bentleycs/iModelTechnologies/_git/graphics-exporter-services/pullrequest/318839).

The majority of the code changes are isolated to a new @itwin/frontend-tiles package.
You can initialize it using its only API, `initializeFrontendTiles`. See DisplayTestApp.ts for an example.

It obtains pregenerated tiles in iMdl format produced by the mesh export service which batch all spatial models into a single tileset. The number of tile requests is drastically reduced, and the FPS is quite good once the tiles are decoded and  converted.

Further refinements will be made in subsequent PRs, including:
- Better LOD refinement (pending changes to tile tree structure).
- Support for toggling display of models, subcategories, and geometry classes.


core/frontend/src/tile/GltfReader.ts
```diff
@@ -623,12 +153,12 @@ export interface GltfReaderResult extends TileContent {
  */
 export class GltfReaderProps {
   public readonly version: number;
-  public readonly glTF: Gltf;
+  public readonly glTF: GltfDocument;
   public readonly yAxisUp: boolean;
   public readonly binaryData?: Uint8Array;
-  public readonly baseUrl?: string;
+  public readonly baseUrl?: URL;
```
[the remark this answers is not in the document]
pmconne: Changed it to `string | URL` to avoid breaking existing callers.

extensions/frontend-tiles/src/BatchedTile.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    resolve(children);
+  }
+
+  public override get channel(): TileRequestChannel {
+    return IModelApp.tileAdmin.channels.getForHttp("itwinjs-batched-models");
+  }
+
+  public override async requestContent(_isCanceled: () => boolean): Promise<TileRequest.Response> {
+    const url = new URL(this.contentId, this.batchedTree.reader.baseUrl);
```
wgoehrig, line 105: FYI - This may cause problems if we try to use as-is.  Locally, I've been explicitly appending a ".gz" suffix to the contentId, under the assumption that we'd either update the mesh export service to not append that suffix when uploading or the manifests would be updated to include it in the contentIds...

pmconne: If export service is adding ".gz" to the content URLs then the content URLs embedded in the tileset JSON need to include that suffix too, per the spec.
Preferably, the ".gz" extension will not be present.
Did you confer with @danieliborra?

wgoehrig: We talked a bit about adding the suffix to the tileset JSON, but my preference would also be to just drop the .gz extension before upload.  We just need to make sure we can do that without breaking existing 3DFT consumers.

DStradley (approved):

extensions/frontend-tiles/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "author": {
+    "name": "Bentley Systems, Inc.",
+    "url": "http://www.bentley.com"
+  },
+  "dependencies": {
+    "@itwin/core-bentley": "workspace:*",
+    "@itwin/core-common": "workspace:*",
+    "@itwin/core-frontend": "workspace:*",
+    "@itwin/core-geometry": "workspace:*",
+    "file-saver": "^2.0.2"
```
aruniverse, line 42: this(file-saver) doesnt seem to be used

extensions/frontend-tiles/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "BIM",
+    "iModel",
+    "UI",
+    "Widget"
+  ],
+  "author": {
+    "name": "Bentley Systems, Inc.",
+    "url": "http://www.bentley.com"
+  },
+  "dependencies": {
```
aruniverse, line 37:
```suggestion
  "peerDependencies": {
```

aruniverse: you'll also want to add the rest of these deps to devDeps

extensions/frontend-tiles/src/BatchedTile.ts
danieliborra: Let's talk about it today.
I'd also prefer not using the .gz suffix. To avoid breaking changes, I'd add a property to the mesh export service request with the version of the 3DFT format, and update the UE plugin. That would avoid future problems too.
@diegopinate does it make sense?
We can tell the users to update the plugin asap and deprecate the ".gz" version in a couple of months.

extensions/frontend-tiles/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "main": "lib/cjs/frontend-tiles.js",
+  "module": "lib/esm/frontend-tiles.js",
+  "typings": "lib/cjs/frontend-tiles",
+  "license": "MIT",
+  "scripts": {
+    "build": "npm run -s copy:public && npm run -s build:cjs",
+    "build:ci": "npm run -s build && npm run -s build:esm",
+    "build:cjs": "tsc 1>&2 --outDir lib/cjs",
+    "build:esm": "tsc 1>&2 --module ES2020 --outDir lib/esm",
+    "copy:public": "cpx \"./public/**/*\" ./lib/public",
```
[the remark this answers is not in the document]
pmconne: done, please review.

extensions/frontend-tiles/package.json
pmconne: done, please review.

extensions/frontend-tiles/package.json
pmconne: done, please review.

extensions/frontend-tiles/src/BatchedTile.ts
diegopinate: Yes, we can add a version field and change the blob client to upload non .gz. Do we want to keep compression? We use zlib's deflate to make into a gz archive.

See [this](https://dev.azure.com/bentleycs/iModelTechnologies/_git/graphics-exporter-services?path=packages/export-3dft-agent/src/storage/AzureBlobClient.ts&version=GC5041cd4e0e69a52d072d4017afe77185efc0156f&line=56&lineEnd=92&lineStartColumn=1&lineEndColumn=4&_a=contents)

aruniverse (approved):

extensions/frontend-tiles/src/BatchedTile.ts
wgoehrig: Yes, we should definitely still compress - as long as we're setting the correct `blobContentEncoding` on upload (which we are) then adding the extension is unnecessary.

PR #5172, opened 2023-03-08
Add `ChannelControl` interface to `IModelDb`

kabentley:
Enforce allowed/denied channels for editing elements, aspects, and models.

Requires https://github.com/iTwin/imodel-native/pull/192

imodel-native: https://github.com/iTwin/imodel-native/tree/add-model-to-callback


pmconne (commented):
I'm interested in @swwilson-bsi's perspective.

Does it make sense to enforce channel constraints on `StandaloneDb`s?

core/backend/src/domains/FunctionalSchema.ts
```diff
@@ -27,7 +27,7 @@ export class FunctionalSchema extends Schema {
     }
   }
 
-  /** @deprecated in 2.x. Use [[schemaFilePath]] and IModelDb.importSchemas instead */
```
kabentley, line 30: @ColinKerr @calebmshafer I don't understand why this method was deprecated. The deprecation message was useless to me. If the function `nativeDb.importFunctionalSchema` should be removed, then someone should do that and replace the implementation of this function with the correct code.

We shouldn't be using a deprecated-in-2.x function in 4.0.

pmconne (approved):

Issue #5036, opened 2023-02-07
Rethink our timeouts' strategy


grigasp:
At the moment the `PresentationRpcImpl` returns a "success" response with a `BackendTimeout` status after 90 seconds of not being able to produce a result. The frontend, upon receiving such a response, repeats the request up to 5 times before failing.

I think the frontend should have a set amount of **time** it repeats the requests instead of counting the number of times it got a "timeout" response. Also, we should consider returning a 503 with `Retry-After` rather than a success response with `BackendTimeout` status.

PR #5173, opened 2023-03-08
Presentation: Refactor timeouts' handling

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/5036


pmconne (approved):

presentation/backend/src/presentation-backend/Presentation.ts
```diff
@@ -120,7 +118,7 @@ export class Presentation {
     this._shutdownListener = IModelHost.onBeforeShutdown.addListener(() => Presentation.terminate());
 
     this._rpcImpl = new PresentationRpcImpl({
-      requestTimeout: this._initProps.requestTimeout ?? defaultRequestTimeout,
+      requestTimeout: this._initProps.requestTimeout ?? DEFAULT_REQUEST_TIMEOUT,
```
saskliutas, line 121: Is it necessary to pass `DEFAULT_REQUEST_TIMEOUT` here? It looks like `PresentationRpcImpl` uses `DEFAULT_REQUEST_TIMEOUT` if props are not provided. Maybe `Presentation.initilize` could check if `this._initProps.requestTimeout` is defined otherwise pass undefined to `PresentationRpcImpl`?

saskliutas (approved):

presentation/backend/src/presentation-backend/Presentation.ts
grigasp: Agreed and fixed.

PR #5178, opened 2023-03-09
Fix core build issues for windows VM scale sets

skirby1996:
Fix bugs uncovered while debugging windows scale set VMs. Not a prerequisite to switch to the new VM scale set but fixes the following bugs:

- Repeated re runs of failed windows CI jobs would continuously prepend the build number with duplicates of the itwinjs-core version. Use built in System.JobAttempt variable so that we only prepend the build number on the initial run.
- A commit message containing a ' or " character will cause the CI build to fail due to how predefined variables are inserted into inline YAML pipeline scripts. Switch to inserting the commit message variable into a here-string to allow for any quotation characters in commit message.


common/config/azure-pipelines/templates/publish.yaml
```diff
@@ -5,12 +5,14 @@ parameters:
 
 steps:
 - powershell: |
-    $commitMsg = '$(Build.SourceVersionMessage)'
+    $commitMsg = @"
+    $(Build.SourceVersionMessage)
```
aruniverse, line 9: is this a multi-line msg or something?

aruniverse (approved):

common/config/azure-pipelines/templates/publish.yaml
aruniverse: yaml & powershell are gross...

Issue #5073, opened 2023-02-14
Add an example for property override with ECExpression in `IsDisplayed` attribute


grigasp:
**Current status:** https://www.itwinjs.org/presentation/content/propertyspecification/#attribute-isdisplayed

The docs only have an example for `boolean` type of value. It's not clear what kind of symbols are available for the ECExpression type of value.

PR #5196, opened 2023-03-14
Presentation: Add docs for `PropertySpecification.isDisplayed` ECExpression case

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/5073


grigasp:
Changed jpgs to pngs, whose quality is better and even the size is smaller (`4.4 kB` vs `1.2 kB` and `10.5 kB` vs `5.4 kB`)

saskliutas (approved):

PR #5149, opened 2023-02-28
delete pkgs from core repo which have moved out into their own repos

aruniverse:
- appui pkgs have moved to https://github.com/iTwin/appui
- presentation pkgs have moved to https://github.com/iTwin/presentation
- eslint-plugin has moved to https://github.com/iTwin/eslint-plugin
- map-layers has moved to https://github.com/iTwin/viewer-components-react

PR #5197, opened 2023-03-14
Fix KeepSIngleZero format trait

RohitPtnkr1996:
Formatting on frontend does not handle KeepSingleZero same as native. Native returns `1.0` frontend formats it to `1.`. FormatProps:
```JSON
{
  "type": "Decimal",
  "precision": 6,
  "formatTraits": ["keepSingleZero", "keepDecimalPoint", "showUnitLabel"],
  "composite": { "includeZero": true, "spacer": " ", "units": [{ "label": "", "name": "Units.DELTA_CELSIUS" }] }
}
```
The reason for this behaviour is, for numbers where fractional part is all zeroes, and the format trait "TrailZeroes" is unset, the frontend code trims all the zeroes including the expected single one, and we're left with just the decimal separator at the end.

The fix checks if the fractional part is all zeroes, and if "keepSingleZero" is set. If so, a single "0" is added after the decimal separator.

The behavior is now in line with how native formats decimals.


ui/imodel-components-react/src/test/inputs/QuantityNumberInput.test.tsx
```diff
@@ -472,13 +472,13 @@ describe("<QuantityNumberInput />", () => {
       const wrapper = render(<QuantityNumberInput persistenceValue={value} quantityType={QuantityType.Length} step={.25} onChange={handleChange} />);
       const input = wrapper.container.querySelector("input");
       expect(input).not.to.be.null;
-      expect((input as HTMLInputElement).value).to.eq("1.");
+      expect((input as HTMLInputElement).value).to.eq("1.0");
```
pmconne, line 475: This file is moving to the iTwin/appuii repo (#5149).

RohitPtnkr1996: Thanks for the heads-up !
I've informed Arun and will liaise with him about the conflict when it arises.

pmconne: The fix will be to delete the file in either PR.
@NancyMcCallB @raplemie fyi test failure incoming in your repo.

RohitPtnkr1996:
This PR has been abandoned due to a credential mix up.
Will create a new PR.

PR #5199, opened 2023-03-14
Fix KeepSIngleZero format trait

RohitPtnkr1996:
Formatting on frontend does not handle KeepSingleZero same as native. Native returns `1.0` frontend formats it to `1.`. FormatProps:
```JSON
{
  "type": "Decimal",
  "precision": 6,
  "formatTraits": ["keepSingleZero", "keepDecimalPoint", "showUnitLabel"],
  "composite": { "includeZero": true, "spacer": " ", "units": [{ "label": "", "name": "Units.DELTA_CELSIUS" }] }
}
```
The reason for this behaviour is, for numbers where fractional part is all zeroes, and the format trait "TrailZeroes" is unset, the frontend code trims all the zeroes including the expected single one, and we're left with just the decimal separator at the end.

The fix checks if the fractional part is all zeroes, and if "keepSingleZero" is set. If so, a single "0" is added after the decimal separator.

The behavior is now in line with how native formats decimals.


ui/imodel-components-react/src/test/inputs/QuantityNumberInput.test.tsx
```diff
@@ -472,13 +472,13 @@ describe("<QuantityNumberInput />", () => {
       const wrapper = render(<QuantityNumberInput persistenceValue={value} quantityType={QuantityType.Length} step={.25} onChange={handleChange} />);
       const input = wrapper.container.querySelector("input");
       expect(input).not.to.be.null;
-      expect((input as HTMLInputElement).value).to.eq("1.");
+      expect((input as HTMLInputElement).value).to.eq("1.0");
```
RohitPtnkr1996, line 475: @pmconne mentioned that this file will be moved to iTwin/appuii repo (https://github.com/iTwin/itwinjs-core/pull/5149).
I've had a word with Arun about deleting the file when it conflicts with his PR.
@NancyMcCallB @raplemie this will cause test failures in your repos.

ColinKerr (approved):

aruniverse:
@mergifyio backport release3.7.x

aruniverse:

PR #5202, opened 2023-03-14
Presentation: Lock down `@opentelemetry` dependency versions to `~1.5`

grigasp:
Looks like there're issues building with the latest `1.10` version - lock down to `1.5` which we know is fine. 

Filed https://github.com/iTwin/presentation/issues/100 to look into upgrading to `1.10`.

### DIFF
--- a/presentation/opentelemetry/package.json
+++ b/presentation/opentelemetry/package.json
@@ -43,18 +43,18 @@
   "peerDependencies": {
     "@itwin/presentation-common": "workspace:*",
     "@opentelemetry/api": "^1.0.4",
-    "@opentelemetry/resources": "^1.5.0",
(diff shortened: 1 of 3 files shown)


presentation/opentelemetry/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/presentation-opentelemetry",
-  "version": "4.0.0-dev.45",
+  "version": "4.0.0-dev.39",
```
aruniverse, line 3:
```suggestion
  "version": "4.0.0-dev.45",
```

aruniverse (approved):
could do `~1.9.0` but this should be fine too

presentation/opentelemetry/package.json
grigasp: rebased

PR #5214, opened 2023-03-15
intersect local ranges

dassaf4:
Added `ClipUtilities.doLocalRangesIntersect` to clash-test two local range boxes, each with its own local-to-world transform.

For example, this is useful for detecting clashing elements from data stored in an iModel:

```
// element data from iModel
const localRange0 = Range3d.createXYZXYZ(-0.023434012896785816, -5.00413553129377, -9.650393591767262, 20.02343401289663, 5.03613553129378, 9.68239359176722);
const localToWorld0 = Transform.createRefs(Point3d.create(-108.12092516312605, 80.97820829881688, 67.15651552186583), YawPitchRollAngles.createDegrees(-47.95656913000159, 15.000000000006024, 90).toMatrix3d());
// ... compute localRange1, localToWorld1 similarly
const margin = 0.0;  // or a small positive margin to uniformly expand localRange1 for catching near-adjacent ranges
const isClash = ClipUtilities.doLocalRangesIntersect(localRange0, localToWorld0, localRange1, localToWorld1, margin);
```

Also:
* `PolyfaceBuilder.addTransformedRangeMesh` honors `builder.options.shouldTriangluate` and `builder.reversedFlag`
* add optional pre-allocated `result` args to `Transform.inverse` and `Matrix3d.factorRigidWithSignedScale`
* new test cases for `RegionOps` tests
* doc improvements


core/geometry/src/clipping/ClipUtils.ts
```diff
@@ -490,6 +492,38 @@ export class ClipUtilities {
     }
     return false;
   }
+
+  private static _workTransform?: Transform;
```
saeeedtorabi, line 496: optional: move these 3 private variables to the top of the class (although I dont know what the standard is for where to put private variables; top of the class or close to where they are used)

saeeedtorabi (approved):

core/geometry/src/clipping/ClipUtils.ts
dassaf4: Good idea. There's no standard I'm aware of. But having general purpose statics up there below any instance data increases their visibility and thus encourages their (judicious) re-use in methods. Of course it would have been nicer if we could have static vars _inside_ a method, but that's not happening.

pmconne: > it would have been nicer if we could have static vars inside a method, but that's not happening.

You [can](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAqAWBLAdgc2gbwFDV9ADgK4BGIiw0qApgC4ByVAHjQBQCUAXNMoQLbFUATphx4xAegBU0CDTA1y0SeOgha0YIUGCqyGtAC80AAwBuUWNwIUqAHT5BAexrOAnviq3q9JvqPtDAD4NLR09AGpw80s8HRotZGgaJAgvWgZmdmi8AF8sPKxgR2RZaDBDbioAdzgkNCzC4tLiCuRq2psGrHEVAAVBFBoYACJjABpoAEYJgCYJgGZhxpLHNVsQR1QWMDSfTLYJ4l2M1gPoI+8T9gmdy992NlMgA) but you shouldn't.

pmconne:
Please consider summarizing any notable improvements in NextVersion.md.

core/geometry/src/clipping/ClipUtils.ts
dassaf4: I needed this in my day.

dassaf4:
> Please consider summarizing any notable improvements in NextVersion.md.

Thanks for the reminder! I'm thinking this function is rather esoteric, but if the requestor ends up using it, I will blurb it.

PR #5221, opened 2023-03-16
review how versions are managed and displayed in documentation

yashincontrol:
Core side of the docs version management. Added package metadata, styling for the new version dropdown menu, currentVersion field and v3.6.1 in the docSites.json. Most of the logic is on the bemetalsmith side.


docs/config/layouts/styles/header.css
```diff
@@ -144,6 +144,90 @@ header nav {
...
+  height: 44px;
+  padding-left: 10px;
+  color: #a2a2a2;
+
+}
+
+.version-dropdown-menu {
+  display: none;
+  position: absolute;
+  top: 100%;
```
ben-polinsky, line 197: So this `top:100%` is a bit odd.

yashincontrol: That is to prevent the dropdown menu from vanishing while hovering over the "gap".

ben-polinsky: Eh it's fine.

williamkbentley:
Does this mean the silly LD flag can go away in v4?

ben-polinsky:
> Does this mean the silly LD flag can go away in v4?

Yep I believe so, @yashincontrol let's look to remove that in this PR as well.

ben-polinsky:
@aruniverse @calebmshafer could you take a look at some point today?

tools/build/scripts/utils/addPackageMetadata.js
```diff
@@ -15,8 +15,14 @@ const rootPackageJson = process.env.RUSHSTACK_FILE_ERROR_BASE_FOLDER
 // Check if path to root package.json is valid.
 const rootPackageJsonPath = require.resolve(rootPackageJson);
 
+//Find the package.json of the project and retrieve the version and repository URL
+const packageJsonPath = path.join(process.cwd(), "package.json");
+const packageJson = JSON.parse(FS.readFileSync(packageJsonPath));
+const version = packageJson.version;
+const repositoryUrl = packageJson.repository.url;
```
aruniverse, line 22: might want to add a fallback; repos should have these but just in case they dont

tools/build/scripts/utils/addPackageMetadata.js
```diff
@@ -15,8 +15,14 @@ const rootPackageJson = process.env.RUSHSTACK_FILE_ERROR_BASE_FOLDER
 // Check if path to root package.json is valid.
 const rootPackageJsonPath = require.resolve(rootPackageJson);
 
+//Find the package.json of the project and retrieve the version and repository URL
+const packageJsonPath = path.join(process.cwd(), "package.json");
+const packageJson = JSON.parse(FS.readFileSync(packageJsonPath));
```
aruniverse, line 20: do you need this new var packageJsonPath? wouldnt it work fine w/o it?

yashincontrol: You mean by just using the rootPackageJsonPath variable? I tried that at first, but it was getting the package.json present at the root of core and not for each project.

aruniverse: gotcha; damn monorepos

aruniverse (approved):

tools/build/scripts/utils/addPackageMetadata.js
```diff
@@ -18,8 +18,11 @@ const rootPackageJsonPath = require.resolve(rootPackageJson);
 //Find the package.json of the project and retrieve the version and repository URL
 const packageJsonPath = path.join(process.cwd(), "package.json");
 const packageJson = JSON.parse(FS.readFileSync(packageJsonPath));
-const version = packageJson.version;
-const repositoryUrl = packageJson.repository.url;
+const version = packageJson.version ?? '';
```
aruniverse, line 21: the pkg version should always be defined afaik
if its missing throw an error instead?

aruniverse (approved):

tools/build/scripts/utils/addPackageMetadata.js
ben-polinsky: Internally debated that. Makes sense.

yashincontrol: Only for the version? Like just warning for the url?

ben-polinsky: Both imo

aruniverse: personally not a fan of IIFE, @ben-polinsky do you have an opinion on this?

ben-polinsky: Yes for the most part, agree. I'd just check for the presence and throw if none.

aruniverse: do you really need an extra fn to just call `throw new error` ?

yashincontrol: Ah, I felt like this way was cleaner. Other options were using IIFE or if statements since we can't `throw new error` when using the "??"

ben-polinsky:
@calebmshafer Could you approve?

tools/build/scripts/utils/addPackageMetadata.js
ben-polinsky: Can we fix this once and for all and get this merged?

yashincontrol: Done

aruniverse (approved):

PR #5226, opened 2023-03-16
Multi-model feature tables

pmconne:
#5170 adds support for displaying tilesets that batch many spatial models together. Each tile has a `PackedFeatureTable` describing the features contained within the tile. Previously, a feature was defined as element Id, subCategory Id, and geometry class, with the (previously correct) assumption that all of the features belonged to the same, single model. With batched tilesets, that assumption is no longer accurate. Moreover, with batched tilesets we can't toggle display of individual models simply by adding/removing individual tile trees; instead we need to handle them like categories, using feature overrides.

This PR adds `MultiModelPackedFeatureTable`. It and `PackedFeatureTable` now implement a new `RenderFeatureTable` interface; most usage of `PackedFeatureTable` is replaced with `RenderFeatureTable`. No breaking API changes are introduced.

The new multi-model feature table stores the element Ids, subcategory Ids, and geometry classes the same way that `PackedFeatureTable` does. To this it appends a model table. Crucially, the features are grouped by model, so that the first N features belong to the first model in the model table, the next M features to the second model, and so on. The model table stores, for each model, its Id and the index of the last feature that belongs to it. This enables us to locate a feature's model given only its feature index by performing a binary search on the model table; and eliminates the redundancy that would otherwise be introduced by having to specify the model Id explicitly for each feature.

This PR also introduces iteration over a `RenderFeatureTable`. For the multi-model feature table, this has been optimized to traverse the model table in parallel, so that we don't have to look up the model Id for each feature in sequence.

Incidentally, reduce new object allocations when working with feature tables by permitting or requiring some functions to take an output object. (Feature table traversal occurs on the hot rendering path).

[Corresponding exporter changes](https://dev.azure.com/bentleycs/iModelTechnologies/_git/graphics-exporter-services/pullrequest/322338).


markschlosseratbentley (approved):

aruniverse (approved):
frontend-tiles changes lgtm

PR #4277, opened 2022-09-14
remove imodelhub and other superfluous docs

ben-polinsky:

PR #5230, opened 2023-03-17
Remove docs that are on the developer.bentley.com site

calebmshafer:
This information is covered on developer.bentley.com and we should avoid duplicating and confusing the content of both.

breakdown of larger PR: #4277


calebmshafer:
@Mergifyio backport release/3.6.x release/3.7.x

ben-polinsky:
Nothing is needed in replace of these docs or links?

ben-polinsky (approved):

ben-polinsky:
> > Nothing is needed in replace of these docs or links?
> 
> I don't think so. If anything we need to re-write almost this entire section of the docs to be more clear and helpful. I'll create an issue to put something together.

Right, we may have it actually

PR #5250, opened 2023-03-20
promote OpenTelemetry Tracing to public

paulius-valiunas:


aruniverse (commented):
can you take a look at what happened in the open telemetry update btw 1.9 and 1.10? there was a breaking change in there, that forced presentation to pin it down.

not necessarily something that needs to be done in this pr, but if we're promoting something to public, it might be good to verify that we wont be affected by downstream breaking changes in one of our deps

paulius-valiunas:
> there was a breaking change in there, that forced presentation to pin it down.

hm, thanks for letting me know, I wasn't aware of that (or I forgot). I'll see if we can do anything about that, or maybe we should keep it beta for now.

aruniverse (approved):

Issue #4851, opened 2022-12-27
Update release tags


pmconne:
**Is your feature request related to a problem? Please describe.**
iTwin.js employs [release tags](https://www.itwinjs.org/learning/guidelines/release-tags-guidelines/) to convey API stability to consumers of the library. Notably, we discourage consumers from using `beta` APIs in production and from using `alpha` APIs at all. The goal is for the APIs to evolve based on testing and feedback until they can be promoted to `public`, or abandoned.

Unfortunately, API release tags frequently remain static, leading consumers to either ignore non-`public` APIs assuming they are not ready for broader adoption; or ignore release tags, consuming non-`public` APIs in production and risking stability if the API changes in the future.

**Describe the solution you'd like**
We need to evaluate our non-`public` APIs and update their release tags appropriately to match their current levels of maturity.
Note: any API that is already in broad production use or intended for third-party consumption is a strong candidate for promotion to `public`. For each such API, file a new issue and a PR to adjust the tags. While doing so, ensure documentation is up to date and sufficient.

**Describe alternatives you've considered**
None.

**Additional context**
A related issue is needed for marking APIs as `deprecated` for 3.6 prior to 4.0 release.

Categories with my best guesses at who should review + address them:
- [x] accudraw @bbastings
- [x] appui @NancyMcCallB @raplemie 
- [ ] devtools - @calebmshafer do you know who owns this?
- [X] display @markschlosseratbentley @pmconne
- [x] crash reporting @swwilson-bsi? @nick4598?
- [X] frontend @pmconne
- [X] backend @pmconne
- [x] markup @bbastings?
- [X] ~~ecdb @ColinKerr? @khanaffan? ECJsonTypeMap.~~
- [X] geocoord @pmconne 
- [x] geometry @dassaf4
- [ ] i18n @calebmshafer who owns this now?
- [x] map-layers @eringram @mdastous-bentley 
- [x] presentation @grigasp 
- [ ] quantity @ColinKerr @calebmshafer is there a separate issue for quantity formatting package yet?
- [x] rpc @swbsi 
- [x] telemetry @paulius-valiunas 
- [x] tools @bbastings 

[categorized.txt](https://github.com/iTwin/itwinjs-core/files/11017849/categorized.txt)

pmconne:
@paulius-valiunas @wgoehrig Our OpenTelemetry APIs are intended for use by applications, yet they remain `alpha`. Please evaluate.

pmconne:
[...]

PR #5252, opened 2023-03-20
API promotions - display

pmconne:
Part of #4851.

### DIFF
--- a/core/frontend/src/render/MockRender.ts
+++ b/core/frontend/src/render/MockRender.ts
@@ -39,7 +39,7 @@ import { Scene } from "./Scene";
  *  (2) Call MockRender.App.startup() instead of IModelApp.startup() before tests begin.
  *  (3) Likewise call MockRender.App.shutdown() when finished. This resets the SystemFactory to its default.
  * @note The APIs within this namespace are intended *strictly* for use with unit tests.
- * @public
+ * @internal
  */
 export namespace MockRender {
   /** @internal */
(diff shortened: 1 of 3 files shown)


core/frontend/src/render/MockRender.ts
```diff
@@ -34,7 +34,8 @@ import { Scene } from "./Scene";
  *  (1) If overriding anything in the implementation supplied herein, pass a SystemFactory function to MockRender.App.systemFactory.
  *  (2) Call MockRender.App.startup() instead of IModelApp.startup() before tests begin.
  *  (3) Likewise call MockRender.App.shutdown() when finished. This resets the SystemFactory to its default.
- * @internal
+ * @note The APIs within this namespace are intended *strictly* for use with unit tests.
+ * @public
```
aruniverse, line 38: do we really want to mark testing related apis as public?

pmconne: Everybody's using it all over the place, which I took as evidence of its usefulness.
The APIs it mocks are all `public`.

markschlosseratbentley (approved):

core/frontend/src/render/MockRender.ts
pmconne: That is correct, everybody's using it for tests, and I agree, moving them to a separate testing package would be ideal. Do you have someone in mind who can put together the new package for 4.0?

I disagree about it being ok to use internal APIs in tests outside of the itwinjs-core repo. It doesn't matter how an API is being used, if it's not public then it's subject to breakage at any time.

pmconne: Reverted to `internal`. If you want a new package, file an issue.

aruniverse: Sorry was stuck in a meeting, i will discuss with others and see if we can find someone to create a testing pkg for 4.0

PR #5254, opened 2023-03-21
API promotions - geocoord, frontend, backend

pmconne:
Part of #4851.

### DIFF
--- a/core/frontend/src/DisplayStyleState.ts
+++ b/core/frontend/src/DisplayStyleState.ts
@@ -50,7 +50,7 @@ export interface OsmBuildingDisplayOptions {
  * @extensions
  */
 export abstract class DisplayStyleState extends ElementState implements DisplayStyleProps {
-  /** @internal */
+  /** The name of the BIS class associated with this class. */
   public static override get className() { return "DisplayStyle"; }
   private _scriptReference?: RenderSchedule.ScriptReference;
   private _ellipsoidMapGeometry: BackgroundMapGeometry | undefined;
@@ -851,7 +851,7 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
  * @extensions
  */
 export class DisplayStyle2dState extends DisplayStyleState {
-  /** @internal */
+  /** The name of the BIS class associated with this class. */
   public static override get className() { return "DisplayStyle2d"; }
   private readonly _settings: DisplayStyleSettings;
 
@@ -877,7 +877,7 @@ export class DisplayStyle2dState extends DisplayStyleState {
  * @extensions
  */
 export class DisplayStyle3dState extends DisplayStyleState {
-  /** @internal */
+  /** The name of the BIS class associated with this class. */
   public static override get className() { return "DisplayStyle3d"; }
   private _settings: DisplayStyle3dSettings;
 
--- a/core/frontend/src/DisplayStyleState.ts
+++ b/core/frontend/src/DisplayStyleState.ts
@@ -50,7 +50,6 @@ export interface OsmBuildingDisplayOptions {
  * @extensions
  */
 export abstract class DisplayStyleState extends ElementState implements DisplayStyleProps {
-  /** The name of the BIS class associated with this class. */
   public static override get className() { return "DisplayStyle"; }
   private _scriptReference?: RenderSchedule.ScriptReference;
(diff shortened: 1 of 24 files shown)


pmconne:
@bbastings the TenativePoint class has no documentation. Can you suggest a general description, however brief, of what this class is used for?

core/frontend/src/DisplayStyleState.ts
```diff
@@ -50,7 +50,7 @@ export interface OsmBuildingDisplayOptions {
  * @extensions
  */
 export abstract class DisplayStyleState extends ElementState implements DisplayStyleProps {
-  /** @internal */
+  /** The name of the BIS class associated with this class. */
```
kabentley, line 53: i  don't think you need to copy this comment everywhere.

core/frontend/src/GeoServices.ts
```diff
@@ -266,6 +272,19 @@ export class GeoConverter {
     });
   }
 
+  /** Convert the specified geographic coordinates into iModel coordinates. */
+  public async geoCoordsToIModelCoords(geoPoints: XYZProps[]): Promise<PointWithStatus[]> {
```
kabentley, line 276: `convertToIModelCoords` ?

core/frontend/src/GeoServices.ts
```diff
@@ -266,6 +272,19 @@ export class GeoConverter {
...
+    const result = await this.getIModelCoordinatesFromGeoCoordinates(geoPoints);
+    return result.iModelCoords;
+  }
+
+  /** Convert the specified iModel coordinates into geographic coordinates. */
+  public async iModelCoordsToGeoCoords(iModelCoords: XYZProps[]): Promise<PointWithStatus[]> {
```
kabentley, line 282: `convertFromIModelCoords` ?

core/frontend/src/GeoServices.ts
AlainRobertAtBentley: I would second a change of name if possible since the converter can convert coordinates that are not latitude/longitude now if the user of the GeoConverter sets it otherwise.

core/frontend/src/DisplayStyleState.ts
pmconne: Please provide a reason.
I don't like our documentation to include properties with zero explanation of what they do.
I don't like to assume people know that a `className` is obviously a BIS class name.
These are static properties so I don't think that the docs will produce a link to the documentation for the root type (EntityState) in which the property is initially defined.

core/frontend/src/GeoServices.ts
pmconne: If you want me to rename something, please explain why.
As you know, I hate arguing about names.
I considered how to name this and my inclination was that this name makes it clear what both the source and target of the conversion are.

pmconne: @AlainRobertAtBentley what would you suggest as a replacement for the term "geo coords" then?

pmconne: @AlainRobertAtBentley please suggest changes to terminology of "geographic (lat/long) coordinate", arguments called `geoPoints`, etc.

AlainRobertAtBentley: Initially the class and methods were intended to convert latitude/longitde coordinates to or from the iModel Coordinates. The class was modified so that it allows performing conversion from any Geographic Coordinate Reference System (GCRS) which may be latitude/longitude based or projected (cartesian). The term geoPoint hints to the fact it is latitude/longitude which may not be the case. Simply naming inPoint, outPoint or sourcePoint, targetPoint would be sufficient.
As for the GeoConverter and GeoServices names I do not mind as much their current name. Technically the GeoConverter could be named GeographicCoordinateReferenceSystemConverter but it would be overly verbose. Same reflection holds for GeoServices.

core/frontend/src/DisplayStyleState.ts
kabentley: the "override" keyword means the documentation comes from the superclass.

pmconne: Resolve when satisfied.

pmconne:
@kabentley do you have further objections?

kabentley (approved):

PR #5263, opened 2023-03-21
remove transformer packages in favor of `@itwin/transformer` monorepo

MichaelBelousov:
- The transformer has a new home at [`@itwin/transformer`](https://github.com/iTwin/transformer/)
- The test-app lives there too now, so removing it
- Also adds transformer docs pipeline integration (thanks @skirby1996)
- fixed doc references, kept learning docs in here for now like other externalized repos are doing


docs/changehistory/ChangeLogs.md
```diff
@@ -2,7 +2,7 @@
 
 - [core-backend](../reference/core-backend/changelog)
 
-- [core-transformer](../reference/core-transformer/changelog)
+- [transformer](../reference/transformer/changelog)
```
aruniverse, line 5: NextVersion.md?

pmconne:
I'm curious if the package name change from `core-transformer` to `transformer` was discussed and is set in stone now.
I ask because it's a breaking change for anyone currently consuming the package and if a name change is desired/needed, was something more descriptive considered like `imodel-transformer`?

MichaelBelousov:
> I'm curious if the package name change from `core-transformer` to `transformer` was discussed and is set in stone now. I ask because it's a breaking change for anyone currently consuming the package and if a name change is desired/needed, was something more descriptive considered like `imodel-transformer`?

It was discussed, but it's (kind of?) too late now, @itwin/transformer is already published (but completely unused so theoretically undo-able via imjs-admin). We went with a new name because if we kept the old one, due to internal API usage, separate patch timelines for core and this package could result in a situation where depending upon the same patch version of `@itwin/core-backend` and `@itwin/core-transformer` could be incompatible which we thought was confusing. Now that you mention it, don't the now version-unlocked presentation and appui libraries have the same problem and they definitely of course kept their names (they didn't have core in their names to drop, however)

CC: @calebmshafer

MichaelBelousov:
renaming the new package to `@itwin/imodel-transformer`

Issue #4297, opened 2022-09-16
Support FutureOn texture mapping


pmconne:
**Is your feature request related to a problem? Please describe.**
FutureOn uses a technique for pattern mapping that generates UV coords in the fragment shader to tile the texture and blend it, interpolating between two levels of detail. iTwin.js currently does not support this mode.

**Describe the solution you'd like**
Implement support for this mapping mode.

**Describe alternatives you've considered**
Continue to use UVs pregenerated by the FutureOn connector. This produces a less desirable result. It also prevents us from using different UV coords for normal mapping.

**Additional context**
The technique is described [here](https://github.com/iTwin/itwinjs-core/issues/3989#issuecomment-1228135985).

markschlosseratbentley:
On track for 4.0. (Still need to come up with a name).

markschlosseratbentley:
Still on track.

markschlosseratbentley:
Currently working the backend -- ready to make PRs for these things.

PR #5264, opened 2023-03-21
Implement constant LOD texture mapping mode.

MarcNeely:
Part of #4297.

This texture mapping mode implements a dynamic calculation of texture coordinates such that the texture there is always a constant level of detail for the texture in the rendered view.

There were also a couple of very minor fixes made, one in CopyStencil.ts to remove the specification for a GL extension in a shader which is no longer required after moving to WebGL2 only.  Another to Vertex.ts for some indentation fixes to make the shader code more readable when it is dumped out.

### DIFF
--- a/core/common/src/TextureMapping.ts
+++ b/core/common/src/TextureMapping.ts
@@ -6,7 +6,7 @@
  * @module Rendering
  */
 
-import { IndexedPolyfaceVisitor, Matrix3d, Point2d, Point3d, PolyfaceVisitor, Transform, Vector3d } from "@itwin/core-geometry";
+import { IndexedPolyfaceVisitor, Matrix3d, Point2d, Point3d, PolyfaceVisitor, Transform, Vector3d, XAndY } from "@itwin/core-geometry";
 import { RenderTexture } from "./RenderTexture";
 
 /** Defines normal map parameters.
@@ -21,6 +21,8 @@ export interface NormalMapParams {
   greenUp?: boolean;
   /** Scale factor by which to multiply the components of the normal extracted from [[normalMap]]. */
   scale?: number;
+  /** True if want to use constant LOD texture mapping for the normal map texture. */
+  useConstantLod?: boolean;
 }
 
 /** Describes how to map a [[RenderTexture]]'s image onto a surface as part of a [[RenderMaterial]].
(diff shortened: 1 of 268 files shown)


core/common/src/TextureMapping.ts
```diff
@@ -112,6 +139,10 @@ export namespace TextureMapping { // eslint-disable-line no-redeclare
     mapMode?: TextureMapping.Mode;
     /** @internal */
     worldMapping?: boolean;
+    /** True if want to use constant LOD texture mapping for the surface texture. */
+    useConstantLod?: boolean;
+    /** Parameters for constantLod mapping mode. */
```
pmconne, line 144: You refer to constant LOD as a "mapping mode" but it's not in the `Mode` enum, instead it's controlled by a separate boolean.
What's the motivation for that vs adding it to the enum?

pmconne: I guess you still need a (non-constant) mode specified if there's a normal map, and constant LOD can only be applied to the pattern map - right?

pmconne: Might it make the API a little easier to use if - like we did for `normalMapParams` - you don't need a separate `useConstantLod` property; rather, you'd enable constant LOD by providing the `constantLodParams` property?

core/common/src/TextureMapping.ts
```diff
@@ -96,6 +98,31 @@ export namespace TextureMapping { // eslint-disable-line no-redeclare
...
+    repetitions?: number;
+    /** An offset in world units used to shift the texture. Defaults to (0, 0). */
+    offset?: XAndY;
+    /** The minimum distance (from the eye to the surface) at which to clamp the texture. Defaults to 1.*/
+    minDistClamp?: number;
+    /** The maximum distance (from the eye to the surface) at which to clamp the texture. Defaults to 2^32. */
+    maxDistClamp?: number;
+  }
+
+  /** Parameters to define consant level of detail mapping mode, a mode of texture coordinate generation which dynamically creates
```
pmconne, line 113:
```suggestion
  /** Parameters to define constant level of detail mapping mode, a mode of texture coordinate generation which dynamically creates
```

core/common/src/TextureMapping.ts
MarcNeely: I had thought about adding it to the enum, but then it would have been hard to get the texture to use it without the normal map using it unless we put a mode in the normal map data.  But then that led to the user being able to specify two different mapping modes which are not constantLod which we don't support.

pmconne: Ok. I notice you **can** use constant LOD for normal map. (I think, if no pattern map is defined, only normal map, and `useConstantLod` is `true`?)
Is there a use case for that?

MarcNeely: I thought the normal maps might as well be able to use constant LOD as well.  You can have all 4 combinations of normal map and texture map using constant LOD or not.  In the case of just a normal map and making it use constant LOD, I think that you would have to turn on useConstantLod in the NormalMapParams.

MarcNeely: As for just using the presence of the constantLodParams to specify whether to use them or not, I had thought about that as well, but though the TextureMapping.ParamProps can have unspecified properties, the TextureMapping.Params looked like everything in there was specified so I didn't want to add something to it that was optional like that.  It again made it hard to specify things for a normal map and a texture.

MarcNeely: I'm not too crazy about the whole thing either, but it was the best solution I could come up with after changing my mind on it a few dozen times.  I'm certainly willing to change it if we can come up with something more straight forward.

MarcNeely:
I've done an ImageTest run on them & found some problems with thematic display that I'm currently working out.

core/common/src/TextureMapping.ts
```diff
@@ -96,6 +98,31 @@ export namespace TextureMapping { // eslint-disable-line no-redeclare
     public static readonly identity = new Trans2x3();
   }
 
+  /** Properties used to construct a [[TextureMapping.ConstantLodParams]]. */
```
markschlosseratbentley, line 101: Should we add a NextVersion.md entry?

markschlosseratbentley (changes requested):

core/common/src/TextureMapping.ts
MarcNeely: I don't know.  Do we need to wait until backend stuff is implemented for it (I'm not even sure what that might entail).

pmconne: Makes sense to me @MarcNeely. Hopefully that will end up in the 4.0 release anyway.

markschlosseratbentley: Sounds good.

markschlosseratbentley (approved):

PR #5213, opened 2023-03-15
Remove y-up-to-z-up transform for batched tiles

pmconne:
As a result of [changes to exporter](https://dev.azure.com/bentleycs/iModelTechnologies/_git/graphics-exporter-services/pullrequest/321606).

PR #5269, opened 2023-03-23
Backport frontend-tiles package

pmconne:
The @itwin/frontend-tiles package was introduced on master for inclusion in 4.0. However, the product team most urgently requesting the new features it provides has many dependencies, most of which it does not own, which would all need to be upgraded to 4.0 before the product could integrate+test the new package.

To accelearate things, backport to 3.7 so that (hopefully) only the product itself will need to upgrade.

Backported #5170, #5213, #5226, and #5266.


extensions/frontend-tiles/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@itwin/frontend-tiles",
+  "version": "4.0.0-dev.45",
```
aruniverse, line 3:
```suggestion
  "version": "3.7.0-dev.5",
```

.github/CODEOWNERS
```diff
@@ -104,6 +106,7 @@ rush.json       @iTwin/itwinjs-core-admins
 /core/frontend                  @kabentley @bbastings @iTwin/itwinjs-core-display
 /core/frontend/src/extension    @calebmshafer @wgoehrig @aruniverse @kckst8 @johnnyd710
 /core/frontend-devtools         @iTwin/itwinjs-core-display @Ellord207
+/core/frontend-tiles            @iTwin/itwinjs-core-display @danieliborra
```
aruniverse, line 109:
```suggestion
/extensions/frontend-tiles            @iTwin/itwinjs-core-display @danieliborra
```

pmconne: Thanks for spotting that.

markschlosseratbentley (approved):

.github/CODEOWNERS
aruniverse: might need to move this to LN 139

PR #5274, opened 2023-03-23
Updating Schema Upgrade docs and KOQ docs

ColinKerr:


docs/bis/guide/schema-evolution/schema-versioning-and-generations.md
```diff
@@ -27,18 +27,25 @@ Minor schema upgrades are schema changes that don’t break compatibility with a
 Minor schema upgrades can **ONLY** include:
 
 * The addition of optional classes.
+* The addition of a class in the middle of a class hierarchy
 * The addition of optional properties.
-* The addition of Kinds of Quantities
-* The addition of Categories
-* Changes to items that don’t affect code (or meaning), such as label changes, description changes and other changes that just affect presentation.
+* The addition of Kinds of Quantities.
+* The addition of Categories.
```
diegoalexdiaz, line 33: I got confused for a second while reading "Categories" (especially being an article in the BIS guide, in which a Category is something else). I think we should call these "Property Categories" to avoid confusion.

diegoalexdiaz (approved):

docs/bis/guide/schema-evolution/schema-versioning-and-generations.md
ColinKerr: Good call, fixed

PR #5277, opened 2023-03-24
Allow apps to fall back to default tile generation

pmconne:
Our prototype for batched spatial tile trees currently requires the tiles to be pre-generated by mesh export service. For apps to begin testing, they need to be able to determine if a pre-generated tileset exists for a given iModel. Generally this will require some asynchronous call.

Previously, the function to compute the base tileset URL was synchronous. Now it is asynchronous. It now also can return `undefined` to indicate no tiles are available, in which case the frontend-tiles package will fall back to the default tiles provided by core-frontend. Once the URL is resolved for a given iModel it is cached for reuse until the iModel is closed.

This introduces no changes to core-frontend so no risk of regressions for apps that do not use the frontend-tiles package.

In a follow-up PR I intend to add unit tests to this package.


wgoehrig (approved):

markschlosseratbentley (approved):

aruniverse (approved):

pmconne:
@Mergifyio backport master

PR #5279, opened 2023-03-24
Fix iterator for ProxySpatialTileTreeReferences

pmconne:
If you opened to a non-spatial view, then a spatial view, no tiles would display. If you then opened a different spatial view, the tiles would display.


wgoehrig (approved):

pmconne:
@Mergifyio backport master

PR #5290, opened 2023-03-27
Mark backendHubAccess @internal

nick4598:


aruniverse (changes requested):
why are we marking this internal? 
this will affect / break the `@itwins/imodels-*` pkgs

cc @robertas-kerpys @austeja-bentley

nick4598:
> why are we marking this internal? this will affect / break the `@itwins/imodels-*` pkgs
> 
> cc @robertas-kerpys @austeja-bentley

I think we only break them if we follow through implementing something described in below issue. 
https://github.com/iTwin/itwinjs-backlog/issues/601
Technically they could still use the type even if it was marked internal I believe.

The main idea behind being marked internal is that we don't want 3rd party consumers of itwinjs (besides imodels-*) to call the functions in this interface explicitly. Any calls of this interface should most likely be as part of some other function within itwinjs itself. I agree though that it feels strange, especially if we want to eventually end up somewhere like https://github.com/iTwin/itwinjs-backlog/issues/601 describes.

nick4598:
Another problem: the docs build fails:

Add delta doc links: 4.603s
BWC Markdown: 2.582s
Invalid Link: "[BackendHubAccess.createNewIModel]($backend)" in file "changehistory/3.0.0"
Invalid Link: "[BackendHubAccess.downloadChangeset]($core-backend)" in file "changehistory/3.3.0"
Invalid Link: "[BackendHubAccess.downloadChangesets]($core-backend)" in file "changehistory/3.3.0"
Invalid Link: "[BackendHubAccess.createNewIModel]($backend)" in file "learning/imodelhub/permissions"
Invalid Link: "[BackendHubAccess.deleteIModel]($backend)" in file "learning/imodelhub/permissions"
Invalid Link: "[BackendHubAccess.createNewIModel]($backend)" in file "learning/imodelhub/imodels/createimodel"
Invalid Link: "[IModelHost.hubAccess]($backend)" in file "learning/imodelhub/imodels/createimodel"
One or more invalid links found.
D:\vsts_b\3\itwinjs-core\node_modules\@bentley\bemetalsmith\lib\processDocs.js:309
        throw err
We could probably get rid of the links in some cases, but I would think we'd want to atleast update pages like this: https://www.itwinjs.org/learning/imodelhub/imodels/createimodel/

pmconne:
@nick4598 you can keep the interface itself public while making the instances of it (e.g., `IModelHost.hubAccess`) internal, no?
Then the @itwin/imodel-* packages would have stability guarantees around the interfaces they are implementing.

common/api/core-backend.api.md
```diff
@@ -1393,12 +1393,12 @@ export class DocumentPartition extends InformationPartitionElement {
     static get className(): string;
 }
 
-// @beta
+// @public
```
nick4598, line 1396: As a result of marking backendhubaccess public, also had to mark downloadchangesetarg, downloadchangesetrangearg, and downloadprogress arg public. @GytisCepk any problem with that? We could potentially explicitly mark downloadChangesets and downloadChangeset as @beta if so.

kabentley: wait, i thought the plan was to mark it @internal?

nick4598: Yeah, I started off the PR that way.. but I just don't see how we can mark it internal. imodels-* packages use it, and we expect backends to provide an implementation of BackendHubAccess, which is pretty necessary to do anything useful (open an iModel for example). Additionally marking it internal, means we need to suggest alternatives to some of these docs https://www.itwinjs.org/learning/imodelhub/imodels/createimodel/

kabentley: you should edit the title of this PR

nick4598: Done, thanks

kabentley: I'm not sure whether create/delete iModel are really valid except for tests. Did you ask the iModelHub team?

nick4598: I didn't ask them, but I did get some comfort from the fact that markdown file has this line "To start working with iModelHub, an iModel for an iTwin has to be created. End users should usually create the iModel for an iTwin through iModelHub website." before mentioning createNewIModel.

nick4598: Pushed a couple of small changes per suggestion from hub team.

'create the iModel for an iTwin through iModel manager' and linking to the CreateiModel learning page in the permissions page.

aruniverse (approved):

aruniverse:
might want to make nextversion.md comment?

kabentley (approved):

grigasp:
@nick4598, do we have an example for how to set up `IModelHost` with `hubAccess` now `@internal`? I see GPB is still using this attribute to pass `BackendHubAccess` instance: https://github.com/iTwin/imodel-access-backend/blob/c97b0a856db73a3c4f1215b74224f1a40c6e7ac7/src/GeneralPurposeImodeljsBackend.ts#L166C72-L170.

kabentley:
no, this is correct. GPB should be allowed to use @internal apis.

PR #5296, opened 2023-03-27
Batched tiles: fit to viewed models

pmconne:
Previously fitting the view always fit the the extents of the entire tileset - i.e., all of the spatial models.
Instead, fit only to the union of the extents of those models that are currently in the model selector.


pmconne:
@Mergifyio backport master

markschlosseratbentley (approved):

aruniverse (approved):

PR #5302, opened 2023-03-28
Presentation: Update setup docs

grigasp:


saskliutas (approved):

grigasp:
@ben-polinsky, any idea why this link is not found?
```
Invalid Link: "[ElectronHost.startup]($core-electron)" in file "presentation/setup/index"
```
I can't find `ElectronHost` using search in our doc site either - is `core-electron` not part of the doc site?

ben-polinsky:
> ElectronHost

It doesn't look like it. Not sure why that is the case. @calebmshafer intended?

aruniverse:
see https://github.com/iTwin/itwinjs-core/issues/5403

fyi @yashincontrol

PR #5311, opened 2023-03-29
API Promotions - AccuDraw

bbastings:
Make AccuDraw public so applications can provide the UI.

Part of #4851

### DIFF
--- a/core/frontend/src/AccuDraw.ts
+++ b/core/frontend/src/AccuDraw.ts
@@ -56,13 +56,13 @@ export enum AccuDrawFlags {
   SmartRotation = (1 << 24),
 }
 
-/** @alpha */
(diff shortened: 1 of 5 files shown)


core/frontend/src/AccuDraw.ts
```diff
@@ -56,13 +56,13 @@ export enum AccuDrawFlags {
...
-/** @alpha */
+/** @public */
 export enum CompassMode {
   Polar = 0,
   Rectangular = 1,
 }
 
-/** @alpha */
+/** @public */
 export enum RotationMode {
```
kabentley, line 66: would be nice to at least have a sentence describing what this is for

kabentley (approved):

PR #1542, opened 2021-06-01
Investigate concurrent schema accesses

IvanJK97:
Enable getSchema to deserialize schemas concurrently.

PR #5313, opened 2023-03-30
Fixed DuplicateSchema error with multiple async getSchema calls

RohitPtnkr1996:
The getSchema() in ECSchemaRPCLocater internally calls the Schema.fromJSON().
When reading the schema, a call is made to context.addSchema() so as to be able to locate schemaItems within the context.
However, there is no check when calling addSchema() to see if the schema has already been loaded.

Hence, if multiple async calls are made to getSchema(), every getSchema() call after the first will throw a DuplicateSchema error as it tried to load the schema again.

I've added a check to see if the schema has already been cached.
addSchema() is only called if the schema is not present in the context.

### DIFF
--- a/common/changes/@itwin/ecschema-metadata/fix-schema-locater-multiple-async-calls_2023-03-30-12-42.json
+++ b/common/changes/@itwin/ecschema-metadata/fix-schema-locater-multiple-async-calls_2023-03-30-12-42.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-metadata",
+      "comment": "Fixed scenario where in if the same schema was retrieved through multiple async calls, the schema was being re-added to cache with each call, leading to DuplicateSchema error",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/ecschema-metadata"
+}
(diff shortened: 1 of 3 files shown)


pmconne:
Wouldn't it be more robust and more efficient to ensure that, if another request for the same schema comes into `getSchema` while you awaiting the result of a previous request, the second (and subsequent) requests await the same result rather than duplicating and then discarding the work being done on behalf of the first request?

ColinKerr:
> Wouldn't it be more robust and more efficient to ensure that, if another request for the same schema comes into `getSchema` while you awaiting the result of a previous request, the second (and subsequent) requests await the same result rather than duplicating and then discarding the work being done on behalf of the first request?

@pmconne makes a good point.  if the schema is already in the context do we need to proceed and do all the work we do AFTER we add the schema to the context?  One concern with this is that it is possible that there could be a partially loaded schema already in the cache.  By running through the schema and forcing everything to load we ensure that the schema is fully loaded when the method returns.  If we are getting into this partial load case I suspect that there is a bug but it's worth making sure @AkshayN5 considers this as he is working on async schema loading

Over all I think the right thing to do here is to include all work to load the schema in the check for the schema already existing in the cache.  If it already exists in the cache all the work should be skipped.

RohitPtnkr1996:
I had a look at the PR #1542 and this issue has been resolved in it.
The partially loaded schema check has also been catered to.
I had a call with Akshay and will be adding this test scenario to his branch.

Abandoning this PR.

PR #5320, opened 2023-03-30
Fix CI build conditions

skirby1996:
This PR fixes conditions so we aren't skipping steps on individual CI runs, as well as simplifies some of the step conditions in the CI pipeline.

### DIFF
--- a/common/config/azure-pipelines/jobs/fast-ci.yaml
+++ b/common/config/azure-pipelines/jobs/fast-ci.yaml
@@ -56,7 +56,7 @@ jobs:
       - template: ../templates/core-build.yaml
         parameters:
           nodeVersion: 18.x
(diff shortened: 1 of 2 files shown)


common/config/azure-pipelines/jobs/fast-ci.yaml
```diff
@@ -56,7 +56,7 @@ jobs:
       - template: ../templates/core-build.yaml
         parameters:
           nodeVersion: 18.x
-          buildIos: ${{ eq(variables['Agent.OS'], 'Darwin') }}
```
aruniverse, line 59: you cant build ios on windows/liunux. revert this file

aruniverse (approved):

PR #5321, opened 2023-03-30
Improve batching of coordinate conversion requests

pmconne:
Each `GeoConverter` has its own independent request queue and cache of previously-converted points. Some callers to `GeoServices.getConverter` like `RealityTileTree` obtain a single `GeoConverter` and reuse it throughout their own lifetime. Therefore they benefit from both batching and caching, and the cache gets deleted once the RealityTileTree becomes disused.

Other callers like `IModelConnection.spatialToCartographic` obtain a new `GeoConverter` every time they need one, use it to convert a single point(!), and then discard the converter. This entirely prevents batching - e.g., calling `spatialToCartographic` 20 times in one frame results in 20 http requests.
To address that, we cache each `GeoConverter` returned by `getConverter` by its datum until it has converted at least one point and has no further outstanding conversion requests. If another call to `getConverter` requests the same datum, we reuse the converter from the cache.

In this way, the converter lives for as long as (and no longer than) any caller is awaiting conversion to/from its datum - it and its cache are deleted once it becomes disused. This makes the coordinate caching less useful, but at least bounded - and maximizes batching of requests.

Public APIs like `ViewState.rootToCartographicFromGcs` and `IModelConnection.cartographicToSpatial` that convert one point at a time have new versions that accept and return arrays of points to discourage code that `await`s each of multiple point conversions in sequence. Relatively few instances of such code exist in itwinjs-core. Non-public APIs on BackgroundMapGeometry were replaced with ones that only operate on arrays.

TODO
- [x] Run ImageTests
- [x] Tests
- [x] Fix bug where a request for a point that already had a request in flight would never resolve

@mdastous-bentley


mdastous-bentley:
There is still an await in  `IModelConnection.cartographicToSpatialFromGcs()` :
`const coordResponse = await geoConverter.getIModelCoordinatesFromGeoCoordinates([geoCoord]);`

Is this a problem?

pmconne:
@mdastous-bentley PR is still in draft, please wait.

bbastings (approved):

markschlosseratbentley (approved):

aruniverse (approved):

pmconne:
ImageTests has timed out twice with output truncated as follows:
```
...
2023-04-03 14:30:49  Virago.bim  [EDL 10 stadium zoom]
2023-04-03 14:30:57  Virago.bim  [EDL 11 stadium zoom wht]
2023-04-03 14:31:04  Virago.bim  [EDL 12 multiple]
2023-04-03 14:31:13  Virago.bim  [EDL 13 paddywagon]
2023-04-03 14:31:20  Geolocated/tpbmaps.bim  [EDL 14 riverside wht]
2023-04-03 14:31:32  Geolocated/tpbmaps.bim  [EDL 15 riverside]
```

aruniverse:
@pmconne anything holding this up?

pmconne:
> @pmconne anything holding this up?

Yes, the ImageTests regressions noted above.

aruniverse:
> > @pmconne anything holding this up?
> 
> Yes, the ImageTests regressions noted above.

@pmconne just want to confirm this is still an issue, and you are ok deferring this to after 4.0 ?

pmconne:
@aruniverse yes, still need to investigate the image differences. No reason to delay 4.0 because of it. FYI @mdastous-bentley.

aruniverse:
@pmconne I assume you re-ran image tests and the delta was acceptable?

pmconne:
@aruniverse ImageTests was hanging. That's fixed. No image differences.

core/frontend/src/IModelConnection.ts
```diff
@@ -486,18 +528,57 @@ export abstract class IModelConnection extends IModel {
...
     return (this.noGcsDefined ? this.cartographicToSpatialFromEcef(cartographic, result) : this.cartographicToSpatialFromGcs(cartographic, result));
   }
 
+  /** Convert [Cartographic]($common) coordinates into points in this iModel's spatial coordinate system using a [[GeoConverter]] or the iModel's [EcefLocation]($common).
+   * @param cartographic Coordinates to be converted to the iModel's spatial coordinate system.
+   * @returns The `cartographic` coordinates converted to spatial coordinates, of the same length and order as `cartographic`.
+   * @throws IModelError if [[isGeoLocated]] is false or any point could not be converted.
+   * @see [[cartographicFromSpatial]] to perform the inverse conversion.
+   */
+  public async spatialFromCartographic(cartographic: Cartographic[]): Promise<Point3d[]> {
```
mdastous-bentley, line 549: @pmconne 
If I got an array of array of points, do I need to serialize all points in a single flat array and then make a single call to `spatialFromCartographic`.

Also what would be a good way to verify how many calls to backend service was made if Im using display-test-app ?

pmconne: I already answered this.
There are tests included in PR that you can use as a guide if you're writing new ones.

PR #5323, opened 2023-03-30
Improve Geometry: Part 2 (src - line 600 to end(1200))

saeeedtorabi:
- Improved documentation
- Improved format style
- Improved APIs
- Fixed minor bugs

### DIFF
--- a/core/geometry/src/Geometry.ts
+++ b/core/geometry/src/Geometry.ts
@@ -203,6 +203,29 @@ export type AngleSweepProps =
   [number, number] |
   { degrees: [number, number] } |
   { radians: [number, number] };
+/**
+* Interface for method with a clone operation
+* @public
+*/
+export interface Cloneable<T> {
+  /** Required method to return a deep clone. */
+  clone(): T | undefined;
+}
+/** Options used for methods like [[Vector2d.isPerpendicularTo]] and [[Vector3d.isParallelTo]].
+ * @public
+ */
+export interface PerpParallelOptions {
+  /**
+   * Squared radian tolerance for comparing the angle between two vectors.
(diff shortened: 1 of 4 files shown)


core/geometry/src/Geometry.ts
```diff
@@ -890,159 +936,207 @@ export class Geometry {
...
-   * * Return true if the toUpperCase values match.
+   * * Return `true` if the `toUpperCase` values of `string1` and `string2` match.
    */
   public static equalStringNoCase(string1: string, string2: string): boolean {
     return string1.toUpperCase() === string2.toUpperCase();
   }
-  /** Test for EXACT match of number arrays. */
+  /**
+   * Test for exact match of two number arrays.
+   * Returns `true` if both arrays have the same length and have the same entries (or both are empty arrays).
```
dassaf4, line 1138: Consider this change here and in similar variant methods below.
```suggestion
   * Returns `true` if both arrays have the same length and entries, or if both arrays are empty or undefined.
```

core/geometry/src/Geometry.ts
```diff
@@ -1101,57 +1204,33 @@ export class Geometry {
...
     if (a === undefined && b === undefined)
       return resultIfBothUndefined;
     if (a !== undefined && b !== undefined)
       return a === b;
     return false;
   }
   /**
-   * Clone an array whose members have a clone method.
-   * * undefined return from clone is forced into the output array.
+   * Clone an `array` of type `T` whose members have a clone method.
```
dassaf4, line 1224:
```suggestion
   * Clone an array whose members have type `T`, which implements the clone method.
```

dassaf4 (approved):

PR #5324, opened 2023-03-31
Fix point cloud jitter and quantization issues for additive refinement pnts

DStradley:
Fixes https://github.com/iTwin/itwinjs-core/issues/5333

Investigating problems with a user point cloud showed jitter and gaps in the displayed data, both indicating quantization issues.  The point cloud was PNTS format, but using additive refinement.  Also, the first 13 pnts files were unquantized, but all the rest were quantized.

Found that we were quantizing the unquantized tiles, so first changed that to generate and use unquantized point cloud graphics (for PNTS), but that by itself did not solve the problem.  Investigating further, I found that the RTC_CENTER in these tiles was large enough to cause quantization issues the way it was added into the vertex shader points.  I removed that offset from the vertex shader points, then added it into a transform for the tile instead, and that resolved the problem.

Had to adjust batch range by RTC_CENTER since it is in the transform now (for locate, etc.).  Also had to create a range for unquantized points for the voxelSize calculation, since the qParam range (scale of 1) would not work for that.

TODO
- [x] Run ImageTests


DStradley:
ImageTests ran and the only differences were considered inconsequential ones from now supporting unquantized pnts data.

pmconne (commented):
Run `rush change` and add the files created in common/changes.

common/api/core-frontend.api.md
```diff
@@ -8706,7 +8706,10 @@ export class ReadonlyTileUserSet extends ReadonlySortedArray<TileUser> {
 }
 
 // @internal
-export function readPointCloudTileContent(stream: ByteStream, iModel: IModelConnection, modelId: Id64String, _is3d: boolean, range: ElementAlignedBox3d, system: RenderSystem): Promise<RenderGraphic | undefined>;
+export function readPointCloudTileContent(stream: ByteStream, iModel: IModelConnection, modelId: Id64String, _is3d: boolean, range: ElementAlignedBox3d, system: RenderSystem): Promise<{
+    graphic: RenderGraphic | undefined;
```
pmconne, line 8710: Did you run `lint`? I suspect it's going to want commas in place of semicolons.

pmconne (approved):

common/api/core-frontend.api.md
DStradley: Ran lint and fixed the errors it found.  It said nothing about this though, and I am still getting the extract-api error.

pmconne:
> I am still getting the extract-api error

The error is telling you what you need to do to fix it.

DStradley:
I did "npm run extract-api" in the core\extension folder (step 1), and it generated no modified files anywhere, so can't do step 2.

DStradley:
Don't know what to do about the repeatable MacOS security error.

pmconne:
> Don't know what to do about the repeatable MacOS security error.

As yet there is [no fix available](https://github.com/Leonidas-from-XIV/node-xml2js/issues/663). Nothing to do with your changes.

DStradley:

Issue #5281, opened 2023-03-24
refactor or remove core-frontend http requests apis


aruniverse:
Kill off or refactor [core/frontend/src/request/Request.ts](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/request/Request.ts)

For iTwin.js 3.0 we killed off the bentley/itwin-client pkg which provided some apis to make http requests. There was still some usage left in core-frontend that depended on the request apis so we moved it to be an internal api of core-frontend, and the goal was to eventually get rid of it. Consumers downstream have started to blindly copy whats in the core library, and unnecessarily pulling in deps to make http reqs. 

Our main objective should be to remove the dependency on 
- [superagent](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/package.json#L107), 
- [qs](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/package.json#L105),
- [deep-assign](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/package.json#L103)

In my opinion we could get away with just using the [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch) api

aruniverse:
@iTwin/mobilesdkadmins looks like this was also copied into core-mobile, [see](https://github.com/iTwin/itwinjs-core/blob/master/core/mobile/src/backend/Request.ts), maybe its still needed there, but might be worth you guys looking at removing this too

aruniverse:
fwiw, i started looking at this a few weeks back, and didnt get a chance to finish
i don't think itll be too bad to wrap up, prob better to start from a new branch tho
https://github.com/iTwin/itwinjs-core/compare/dep-cleanup

tcobbs-bentley:
@aruniverse Neither @toddsouthenbentley nor I have any knowledge of the linked Request.ts code. I have no idea what it does or why it's there. So refactoring to get rid of it or have it not use the other three packages is not something I can provide constructive input on. I'm guessing that Affan or Steve would know.

Having said that, SYNCHRO Field depends on @bentley/itwin-client for its own WSG-based SYNCHRO-only communication. It sounds like this is saying that them doing that is bad, but maybe it's not for their use case. They should probably be consulted, though: @SatyakamKhadilkar @vincasrazmabentley.

SatyakamKhadilkar:
@aruniverse (cc @tcobbs-bentley).
[...]

PR #5325, opened 2023-03-31
Refactor internal `request` functions

GytisCepk:
### Changes:
- Refactor internal `request` function in _core-frontend_ to use `fetch` instead of `superagent`.
- Removed unused code from internal `request` function in _core-mobile_.
- _core-frontend_ no longer depends on `qs`, `deep-assign` and `superagent`.
- _core-mobile_ no longer depends on `qs` and `deep-assign`.

### Issue: https://github.com/iTwin/itwinjs-core/issues/5281


core/frontend/src/extension/providers/RemoteExtensionProvider.ts
```diff
@@ -47,14 +47,12 @@ export class RemoteExtensionProvider implements ExtensionProvider {
...
   public async getManifest(): Promise<ExtensionManifest> {
-    const options: RequestOptions = { method: "GET" };
-    const response = await request(this._props.manifestUrl, options);
-    const data = response.body || (() => {
-      if (!response.text)
-        throw new Error("Manifest file was empty.");
-      return JSON.parse(response.text);
-    })();
-    return data;
+    const response = await request(this._props.manifestUrl, "text");
```
aruniverse, line 50: wonder for this one should be "json" not "text", then can avoid having to parse it later

core/frontend/src/RealityDataSourceCesiumIonAssetImpl.ts
```diff
@@ -84,24 +84,14 @@ export class RealityDataSourceCesiumIonAssetImpl implements RealityDataSource {
     else
       this._baseUrl = `${urlParts.join("/")}/`;
   }
-  private async _doRequest(url: string, responseType: string): Promise<any> {
-    let options: RequestOptions = {
-      method: "GET",
-      responseType,
+  private async _doRequest(url: string, responseType: "json" | "arraybuffer"): Promise<any> {
```
aruniverse, line 87: do we need to keep this?

aruniverse (commented):
lgtm; thx for picking this up last min

will just need someone from @itwin/itwinjs-core-display and @mdastous  to test this out since we seem to be lacking e2e tests

pmconne (approved):
ImageTests passed.

core/frontend/src/extension/providers/RemoteExtensionProvider.ts
GytisCepk: I believe `superagent` used to default to `"text"` if `responseType` was not provided.

Changed to `"json"` anyway, because end result should be the same.

core/frontend/src/RealityDataSourceCesiumIonAssetImpl.ts
GytisCepk: Removed it. Without this, code will be a bit more type safe.

aruniverse (approved):

PR #5328, opened 2023-03-31
Harden computeFrustumPlanes() in FrustumPlanes.ts

markschlosseratbentley:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/670

In the case of the above bug, the front rect of a scaled-down frustum for readPixels() can become rather tiny because the source view frustum was already rather small. Deriving the front plane's normals from a small rect resulted in a near-zero plane normal, causing computeFrustumPlanes() to return an empty array of planes. This caused the rendering system to eventually assert.

To fix the problem, we now derive the front plane from the rear plane.

We also generally hardened the code in computeFrustumPlanes() by ensuring that the normals are only ever calculated from points of the frustum box that yield a 90 degree angle (this was not done previously).  We did this by reordering the point used for the cross product function. This reordering also makes it clearer what the values actually mean.

More documentation has been added to this source file in order to make it easier to debug and read in the future.

Furthermore, a test has been added (FrustumPlanes.test.ts) which exercises the particular bad frustum which prompted this work, verifying that it now processes properly. Some good frustums have been added to that test file as well.

- [x] Image Tests

### DIFF
--- a/core/common/src/test/FrustumPlanes.test.ts
+++ b/core/common/src/test/FrustumPlanes.test.ts
@@ -23,7 +23,7 @@ describe("FrustumPlanes", () => {
       Point3d.fromJSON({ x: 22.838134428009578, y: -21.182090505813676, z: 2.248236463773556 }), // left bottom front
       Point3d.fromJSON({ x: 22.826687847036766, y: -21.3106221602675, z: 2.248236463773556 }),   // right bottom front
       Point3d.fromJSON({ x: 22.845184176795684, y: -21.18271833185903, z: 2.332414308846702 }),  // left top front
-      Point3d.fromJSON({ x: 22.833737595822875, y: -21.311249986312852, z: 2.332414308846702 })  // right top front
+      Point3d.fromJSON({ x: 22.833737595822875, y: -21.311249986312852, z: 2.332414308846702 }), // right top front
     ]);
 
     const frustumPlanes = FrustumPlanes.fromFrustum(frustum);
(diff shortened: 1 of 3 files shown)


markschlosseratbentley (commented):
Image Tests are currently running.

core/common/src/test/FrustumPlanes.test.ts
```diff
@@ -0,0 +1,71 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import { expect } from "chai";
+import { FrustumPlanes } from "../geometry/FrustumPlanes";
+import { Frustum } from "../Frustum";
+import { Point3d } from "@itwin/core-geometry";
+
+describe("FrustumPlanes", () => {
```
pmconne, line 10: Also test how it handles a bad frustum (e.g., input points do not define a valid frustum)?

markschlosseratbentley: done

pmconne (approved):

pmconne:
Backporting to 3.6 and 3.7, right? Issue was reported against 3.6.2.

calebmshafer:
@pmconne 3.7 should be the only one needed as 3.7.0 is out and we'll release a 3.7.1 Monday (if we haven't already).

pmconne:
@calebmshafer you're ok with introducing a regression in 3.6 and not fixing it in 3.6?

markschlosseratbentley:
@calebmshafer just 3.7 backport or 3.6 as well per the above comments?

markschlosseratbentley:

markschlosseratbentley:
@pmconne @calebmshafer I completed a backport to 3.7. Let me know if we decide on a 3.6 backport. Any backport needs a bit of manual tweaking due to conflicts in FrustumPlanes.ts.

PR #5332, opened 2023-03-31
Improve Geometry: Part 3 (tests)

saeeedtorabi:
- Add new tests to increase coverage to 99%
- Note: Depricated method isHugeCoordinate is not testable due to lint error


core/geometry/src/test/Geometry.test.ts
```diff
@@ -0,0 +1,133 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import { assert } from "chai";
+
+import { Cloneable, Geometry } from "../Geometry";
+
+describe("Transform.isAlmostEqualOptional", () => {
```
dassaf4, line 9: These `describe` and `it` statements should be named with prefix "Geometry" not "Transform". :)
```suggestion
describe("Geometry.isAlmostEqualOptional", () => {
```

dassaf4 (approved):

core/geometry/src/test/Geometry.test.ts
saeeedtorabi: copy paste issue!

PR #5334, opened 2023-03-31
fix Image Tests quantization regression

dassaf4:
PR #5195 inadvertently loosened the tolerance for zero-vector detection in `Point4d.toPlane3dByOriginAndUnitNormal`, resulting in more undefined planes computed by `BackgroundMapGeometry.getFrustumIntersectionDepthRange`, which affected image quantization and resulted in a regression detected by Image Tests.

This PR addresses the regression by reinstating the previous tighter normalization tolerance. Smaller-magnitude vectors, such as seen in "inside the ellipsoid" views, will now successfully normalize.

Also:
* Standardize 2d/3d/4d vector normalization logic on the tighter `Geometry.smallFraction` tolerance rather than `Geometry.smallMetricDistance`.
* Streamline `Plane3dByOriginalAndUnitNormal` constructors to avoid normalizing twice and allocating unnecessary objects.
* `Vector3d.normalizeWithLength` returns _uncorrected_ original magnitude.
* Add implementation regression testing for `Point4d.toPlane3dByOriginAndUnitNormal`.
* Add `Point4d` missing coverage.


core/geometry/src/Geometry.ts
```diff
@@ -255,6 +255,13 @@ export class Geometry {
     }
     return distance;
   }
+  /** Correct `fraction` to zero if undefined or smaller than fraction tolerance. Otherwise return it unchanged. */
```
saeeedtorabi, line 258: I'm not happy with this doc. I prefer

```
/**
 * Correct `fraction` to `replacement` if fraction is `undefined` or smaller than `Geometry.smallFraction`. Otherwise return it unchanged. 
 * If no `replacement` is passed, use 0.
```

core/geometry/src/geometry3d/Point3dVector3d.ts
```diff
@@ -924,26 +924,32 @@ export class Vector3d extends XYZ {
   public static unitZ(scale: number = 1): Vector3d {
     return new Vector3d(0, 0, scale);
   }
-  /** Divide by denominator, but return undefined if denominator is zero. */
```
saeeedtorabi, line 927: Please fix the formatting:
```
  /** 
   * Scale the instance by 1.0/`denominator`.

**Review by saeeedtorabi — CHANGES_REQUESTED** [review]

File: core/geometry/src/geometry3d/Point3dVector3d.ts [hunk 3]
```diff
@@ -924,26 +924,32 @@ export class Vector3d extends XYZ {
...
+  /** Scale the instance by 1.0/`denominator`.
```
saeeedtorabi, line 927: Please fix the formatting:
```
  /** 
   * Scale the instance by 1.0/`denominator`.

saeeedtorabi (changes requested):

saeeedtorabi (approved):

PR #5337, opened 2023-03-31
Remove core transformer

MichaelBelousov:
- The transformer has a new home at [@itwin/transformer](https://github.com/iTwin/transformer/)
- The test-app lives there too now, so removing it
- Also adds transformer docs pipeline integration (thanks @skirby1996)
- fixed doc references, kept learning docs in here for now like other externalized repos are doing


docs/reference/leftNav.md
```diff
@@ -53,6 +52,13 @@ packageClassification:  [
          "linear-referencing-common",
          "physical-material-backend"
       ]
+   },
+   {
+      "order":"6",
+      "header":"Transformer",
+      "packages":[
+         "transformer"
```
aruniverse, line 60: should this be imodel-transformer now?

docs/config/docSites.json
```diff
@@ -89,7 +89,7 @@
       "presentation-frontend": "presentation-frontend",
       "quantity": "core-quantity",
       "reality-data-client": "reality-data-client",
-      "transformer": "core-transformer",
+      "transformer": "transformer",
```
aruniverse, line 92: imodel-transformer?

docs/changehistory/ChangeLogs.md
```diff
@@ -2,7 +2,7 @@
 
 - [core-backend](../reference/core-backend/changelog)
 
-- [core-transformer](../reference/core-transformer/changelog)
+- [transformer](../reference/transformer/changelog)
```
aruniverse, line 5: imodel-transformer?

aruniverse (approved):

docs/config/docSites.json
MichaelBelousov: I will generate a new doc build to test

MichaelBelousov: actually can we keep the doc artifact name as just `transformer`, I suppose it doesn't matter because of the alias here, and we can always change it in the future. Thoughts?

pmconne: Sounds better to me.

aruniverse: the key can stay transformer, but the value should be the pkg name, imodel-transformer

docs/changehistory/ChangeLogs.md
MichaelBelousov: done

ColinKerr (approved):

PR #5345, opened 2023-04-04
Add OverrideFormat toJSON method

RohitPtnkr1996:
Added getFormatProps function to OverrideFormat class. Added unit tests.
Renamed UnitConversion interface to UnitConversionProps in core-quantity to avoid name conflicts in imports.
Renamed UnitProps interface to SchemaItemUnitProps in ecschema-metadata to avoid name conflicts in imports.


core/ecschema-metadata/src/test/Metadata/OverrideFormat.test.ts
```diff
@@ -132,12 +136,52 @@ describe("OverrideFormat", () => {
...
+    unitListMile.push([mileU!, "mi"]);
+
+    const overridePrecisionAndUnit = new OverrideFormat(format!, FractionalPrecision.Four, unitListMile);
+    assert.isDefined(overridePrecisionAndUnit.units);
+    expect(overridePrecisionAndUnit.units!.length).eq(1);
+    expect(overridePrecisionAndUnit.units![0][0]).eq(mileU);
+    assert.equal(overridePrecisionAndUnit.fullName, "TestSchema.TestFormat(4)[Formats.MILE|mi]");
+    assert.equal(overridePrecisionAndUnit.name, "TestSchema.TestFormat(4)[Formats.MILE|mi]");
+
+    expect(JSON.stringify(overridePrecisionAndUnit.toJSON())).to.be.equal(`{"schemaItemType":"Format","type":"Fractional","precision":4,"composite":{"spacer":"-","includeZero":false,"units":[{"name":"Formats.MILE","label":"mi"}]}}`);
```
ColinKerr, line 185: I don't like comparing the results as strings ... I'd rather compare the json.  you could stringify then parse and compare with expected actual JSON

common/api/ecschema-metadata.api.md
```diff
@@ -7,6 +7,7 @@
 import { BaseFormat } from '@itwin/core-quantity';
 import { BentleyError } from '@itwin/core-bentley';
 import { DecimalPrecision } from '@itwin/core-quantity';
+import { FormatProps as FormatProps_2 } from '@itwin/core-quantity';
```
ColinKerr, line 10: I missed this at first but this indicates a problem we should probably solve.  core-quantity and ecschema-metadata both have a definition for FormatProps.  They are the same except that the one in ecschema-metadata extends SchemaItemProps.

We shouldn't have two things with the same name and we should try to avoid duplicating the definitions.  Maybe we could rename ecschema-metadata FormatSchemaItemProps or SchemaItemFormatProps (I don't like either) and it could extend both SchemaItemProps and FormatProps (from core-quantity)

Please check for other duplicate names between the two packages.

core/ecschema-metadata/src/test/Metadata/OverrideFormat.test.ts
RohitPtnkr1996: Done!

common/api/ecschema-metadata.api.md
RohitPtnkr1996: Done!

pmconne:
> Renamed UnitConversion interface to IUnitConversion in core-quantity to avoid name conflicts in imports.
> Renamed UnitProps interface to IUnitProps in ecschema-metadata to avoid name conflicts in imports.

I hate nitpicking over names but please consult [our coding guidelines](https://www.itwinjs.org/learning/guidelines/typescript-coding-guidelines/#names)

> Do not use I as a prefix for interface names.

common/api/ecschema-metadata.api.md
pmconne: `type SchemaItemFormatProps = FormatProps & SchemaItemProps` would be clearer and less error-prone than duplicating the big list of properties.

core/ecschema-metadata/src/Metadata/OverrideFormat.ts
```diff
@@ -108,7 +109,37 @@ export class OverrideFormat {
...
-    return overrideFormat !== undefined && overrideFormat.name !== undefined && overrideFormat.parent !== undefined &&
-             overrideFormat.parent.schemaItemType === SchemaItemType.Format;
+    return overrideFormat !== undefined && overrideFormat.name !== undefined && overrideFormat.parent !== undefined && overrideFormat.parent.schemaItemType === SchemaItemType.Format;
+  }
+
+  /**
+   * Returns a JSON object that contains the specification for OverrideFormat.
+   */
+  public toJSON(): SchemaItemFormatProps {
+    const formatJson = this.parent.toJSON() as any;
```
pmconne, line 119: Why `any` here and below?

common/api/ecschema-metadata.api.md
RohitPtnkr1996: Done

pmconne: Thanks. You good @ColinKerr?

pmconne:
Please mention renames in NextVersion.md.

core/ecschema-metadata/src/Metadata/OverrideFormat.ts
RohitPtnkr1996: The parent FormatProps has all readonly members.
Without any, I wasn't able to call their toJSON()s and overwrite the JSON object with the OverrideFormat's members.

pmconne: Skipping questions about *why* all the interfaces in this package declare their properties as `readonly`: consider using `as Mutable<FormatProps>` instead. This way you don't lose type safety.
Less efficient alternative would involve spread operator.

core/ecschema-metadata/src/Metadata/OverrideFormat.ts
```diff
@@ -108,7 +110,37 @@ export class OverrideFormat {
...
+   */
+  public toJSON(): SchemaItemFormatProps {
+    const formatJson = this.parent.toJSON() as Mutable<SchemaItemFormatProps>;
+
+    // Update Precision overriden property
+    formatJson.precision = this.precision;
+
+    if (this.units !== undefined) {
+      // Update Units overriden property
+      const composite: any = {};
```
pmconne, line 127: `any`.

pmconne (approved):

core/ecschema-metadata/src/Metadata/OverrideFormat.ts
```diff
@@ -108,7 +110,35 @@ export class OverrideFormat {
...
-    return overrideFormat !== undefined && overrideFormat.name !== undefined && overrideFormat.parent !== undefined &&
-             overrideFormat.parent.schemaItemType === SchemaItemType.Format;
+    return overrideFormat !== undefined && overrideFormat.name !== undefined && overrideFormat.parent !== undefined && overrideFormat.parent.schemaItemType === SchemaItemType.Format;
+  }
+
+  /**
+   * Returns a JSON object that contains the specification for OverrideFormat.
+   */
+  public toJSON(): SchemaItemFormatProps {
+    const formatJson = this.parent.toJSON() as Mutable<SchemaItemFormatProps>;
```
ColinKerr, line 120: I don't know that a Format Override has the same props as a format that is a schema item.  With this implementation it appears that you will return format json that is identical to the parent format just with the overridden values changed.  How is someone supposed to determine if this json is the original format or an override of the format?

One way to implement this would be to serialize the name of the parent format and the overridden values.  This would match how we actually serialize the value in xml and json.  Maybe this is not what @grigasp wants.  Another option is to serialize the entire set of format props but not serialize the schema item portions.  Then add a parent property with the parent formats name.  This would let us identify the json as a format override, it would not let us directly determine which attributes were overridden but it could be determined by comparing with the original format.

I also think the documentation here needs to be updated to be more explicit what it is doing.  The 'specification for the OverrideFormat' is really just the parent name plus the overrides.  Or well at least it is to me but I could see someone having a different opinion.  So we should better describe what the method is doing and how the result should be used.  Normally toJSON() methods are used for serialization, this would not be.

saskliutas: We want to serialize `Format` and `OverrideFormat` into [`core-quantity` FomatProps](https://github.com/iTwin/itwinjs-core/blob/cc29d0d9fba7dbf11b2a87e3b97c8bf80f21c105/core/quantity/src/Formatter/Interfaces.ts#L15). At the moment we need to check what type of format we get from Schema and [convert accordingly](https://github.com/iTwin/itwinjs-core/blob/cc29d0d9fba7dbf11b2a87e3b97c8bf80f21c105/presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts#L128).

From your comment it looks like that this should not be done using `toJSON` method? If that's the case I think a separate function that converts `Fomat` and `OverrideFormat` into `core-quatinty` `FormatProps` should be to introduced and `toJSON` can be implemented in a way that makes the most sense for Schema serialization/deserialization.

docs/changehistory/NextVersion.md
```diff
@@ -248,45 +248,6 @@ In addition to upgrading iTwin.js core dependencies to `4.0`, there are some oth
 - Support for React 18 (keep support of React 17 too).
 - Upgrade [iTwinUI](https://github.com/iTwin/iTwinUI) from v1 to v2.
 
-## Transformation
```
ColinKerr: Should you be removing this?

ColinKerr (approved):

saskliutas:
How this new API should be used to create `FormatterSpec` from schema `Format`? Is it something like this: 
```typescript
function getFormatProps(koq: KindOfQuantity) {
  const format = koq.presentationFormats[0];
  if (OverrideFormat.isOverrideFormat(format))
    return format.getFormatProps();
  return format.toJSON();
}

const formatProps = getFormatProps(koq);
const format = await Format.createFromJSON("", this._unitsProvider, formatProps);
const formatterSpec = FormatterSpec.create("", format, this._unitsProvider, persistenceUnit);
```
Is it possible to update API so that we could avoid `if (OverrideFormat.isOverrideFormat(format))` check on our side? Maybe something like this:

```typescript
const formatProps = koq.presentationFormats[0].getFormatProps(); // or getFormatProps(koq.prsentationFormats[0]);
const format = await Format.createFromJSON("", this._unitsProvider, formatProps);
const formatterSpec = FormatterSpec.create("", format, this._unitsProvider, persistenceUnit);
```

RohitPtnkr1996:
> How this new API should be used to create `FormatterSpec` from schema `Format`? Is it something like this:
> 
> ```ts
> function getFormatProps(koq: KindOfQuantity) {
>   const format = koq.presentationFormats[0];
>   if (OverrideFormat.isOverrideFormat(format))
>     return format.getFormatProps();
>   return format.toJSON();
> }
> 
> const formatProps = getFormatProps(koq);
> const format = await Format.createFromJSON("", this._unitsProvider, formatProps);
> const formatterSpec = FormatterSpec.create("", format, this._unitsProvider, persistenceUnit);
> ```
> 
> Is it possible to update API so that we could avoid `if (OverrideFormat.isOverrideFormat(format))` check on our side? Maybe something like this:
> 
> ```ts
> const formatProps = koq.presentationFormats[0].getFormatProps(); // or getFormatProps(koq.prsentationFormats[0]);
> const format = await Format.createFromJSON("", this._unitsProvider, formatProps);
> const formatterSpec = FormatterSpec.create("", format, this._unitsProvider, persistenceUnit);
> ```

Yes that's how you'll have to use it.
Yeah sure I'll make the change to move the check to my side.

saskliutas (approved):

PR #5240, opened 2023-03-17
Deprecate query, queryRowCount, and restartQuery methods

mattbjordan:
This will need to be backported to 3.7.x. In the backport PR, I will add a description of the changes to NextVersion.md.

@ColinKerr and I want to get this into 3.7.x, so the linter errors which say to replace the use of the now-deprecated methods have been temporarily suppressed. Here is an issue to track that the methods' usages still need to be removed/updated: #5239

PR #5347, opened 2023-04-04
Replace calls to deprecated query methods

### LINKED ISSUE #5239: Replace calls to deprecated query methods
Get rid of calls to deprecrated `query`, `queryRowCount`, and `restartQuery` methods and replace them with updated equivalents. 

The linter was suppressed for lines that use these methods in [this commit (#75fee93)](https://github.com/iTwin/itwinjs-core/commit/75fee93073befac92b2984d81a9298c1e7ee68e2). That commit can be used to see where function calls need to be updated.

mattbjordan:
Resolves #5239  

Replaces calls to `query`, `queryRowCount`, and `restartQuery` with `createQueryReader` as the former were deprecated (see #5240).

Most of the refactoring was relatively straightforward. The more unique changes are to `ExecutingECSQL.ts` which has example code and to `FavoritePropertiesManager` and `RulesetEmbedder` tests which required replacing the mocked `query` method with a mocked `createQueryReader` method. 

I believe the only changes not covered by tests are to display-test-app.


pmconne:
ECSqlReader is entirely undocumented. Will a subsequent PR rectify that?

pmconne:
Was consideration given to providing an API that creates an ECSqlReader internally and iterates over `reader.current`? The new API seems inconveniently verbose vs `query`.

mattbjordan:
> ECSqlReader is entirely undocumented. Will a subsequent PR rectify that?

It is very much on our radar to properly document it early in 4.x. I just created a GitHub issue for it since there was none: #5349

mattbjordan:
> Was consideration given to providing an API that creates an ECSqlReader internally and iterates over `reader.current`? Or at least giving `ECSqlReader` a `[Symbol.iterator]`? The new API seems inconveniently verbose vs `query`.

@khanaffan or @ColinKerr would know more about this

pmconne:
> > Was consideration given to providing an API that creates an ECSqlReader internally and iterates over `reader.current`? Or at least giving `ECSqlReader` a `[Symbol.iterator]`? The new API seems inconveniently verbose vs `query`.
> 
> @khanaffan or @ColinKerr would know more about this

@ColinKerr @khanaffan do you have objections to a better replacement API than the one currently supplied?

pmconne:
@mattbjordan #5240 doesn't explain *why* it deprecated `query`. What was the motivation? Was it just the `any` return type, or something else?

grigasp (approved):

khanaffan (approved):

pmconne (changes requested):
@khanaffan please explain why `query` was deprecated.

> @mattbjordan https://github.com/iTwin/itwinjs-core/pull/5240 doesn't explain why it deprecated query. What was the motivation? Was it just the any return type, or something else?

ColinKerr:
> > > > Was consideration given to providing an API that creates an ECSqlReader internally and iterates over `reader.current`? Or at least giving `ECSqlReader` a `[Symbol.iterator]`? The new API seems inconveniently verbose vs `query`.
> > > 
> > > 
> > > @khanaffan or @ColinKerr would know more about this
> > 
> > 
> > @ColinKerr @khanaffan do you have objections to a better replacement API than the one currently supplied?
> 
> I have no issue with the iterator symbol. Though it will by a `Symbol.asyncIterator`.

I agree that having an async iterator based implementation would provide a better API.  I think I would prefer the iterator on the ECSqlReader rather than another method that returns the iterator instead of the reader, but I'm open to other implementations

pmconne:
My concern is the volume of code that needs to be adjusted to respond to this deprecation - not only within itwinjs-core, but outside of it. We are putting an obstacle in our users' upgrade path, which we want to avoid - especially if we can't show a tangible benefit for doing so.

I'd like to see `query`'s return type change from `AsyncIterableIterator<any>` to `AsyncIterableIterator<QueryRowProxy>`.
If that's not possible, I'd like to see an equivalent method to `query` with a different name, again iterating over `QueryRowProxy`.
If neither of those is acceptable (please explain why), make `ECSqlReader` iterable so people can replace `for await (const row of imodel.query(...))` with `for await (const row of imodel.createQueryReader(...))`.
(Regardless of how we address the `query` deprecation, making the reader iterable seems like a no-brainer).

pmconne:
> just replace .query with .createQueryReader where possible, correct?

Yes please.

pmconne (changes requested):
Please add this change to NextVersion.md with example for adjusting existing code to use `createQueryReady` instead of `query`. Everything else looks good.

mattbjordan:
Well I see that you've noticed that I'm done with the refactor. @pmconne 

The one thing of note is that there is a slight quirk since iterating on ECSqlReader returns a QueryRowProxy object and not a raw JavaScript object. I noticed failing tests, seemingly in places where a comparison was made against the returned row. So basically, if you want to do anything with the returned row other than simply accessing the row values, you can call `.toRow()` on it to get a raw JavaScript object back instead - which is exactly what most of the previous code assumed it was working with. I.e.,

```ts
for await (const queryRow of iModel.createQueryReader(queryString)) {
    jsObjectRow = queryRow.toRow();
    // ...
}
```

So if any issues pop up in the future because of these changes, calling `.toRow()` will probably fix it.

I'll probably note this quirk in what I write in NextVersion.md.

ColinKerr (approved):

calebmshafer (approved):

PR #5356, opened 2023-04-05
Rename ECSqlReader to QueryReader

mattbjordan:
### Changes

- Rename `ECSqlReader` to `QueryReader` (including filenames)
- Rename `QueryRowProxy` to `QueryRow`
- Fix incorrect "getArray" key on `QueryReader` to "toArray"

Typically our names for interfaces and classes that deal with querying are labled as "Query-" something - so `ECSqlReader` is renamed to be consistent with everything else.


aruniverse (dismissed):
plz make sure to update your previous deprecation comments made in 3.7

mattbjordan:
> plz make sure to update your previous deprecation comments made in 3.7

The deprecation comments above the query methods only reference the `createQueryMethod`, not `ECSqlReader`/`QueryReader` itself.

pmconne:
Please consider **not** renaming this if people are using it already (it was introduced quite a while ago, wasn't it), or at the very least defining a type alias. Users get quite annoyed when we make them alter their code just because we came up with a prettier name for something. Note: type aliases do not prevent all cases of users having to change their usage of the type name.

pmconne:
> ECSqlReader is not out of beta yet

I know that. But just because we're allowed to rename it doesn't necessarily justify doing so. i.e., if it imposes more burden on current users than benefit.

If you're pretty confident nobody's relying on it, please consider also renaming `QueryRowProxy` to `QueryRow` or `QueryReaderRow`. Whether or not the iterator exposes a "proxy" is of no interest to users.

mattbjordan (commented):
@ColinKerr Can you review this PR and read conversation happening in the comments? I want to know that you're fully aware of these changes.

ColinKerr (approved):

grigasp:
This change just broke `@itwin/presentation-components`: https://github.com/iTwin/nightman/actions/runs/4624001861/jobs/8178385340. We promised `presentation-components@4.0` will support core packages @ `^3.7 || ^4.0`, so we need an API to run queries that exists in both major versions. We could revert to using the deprecated `query` function, but I'm not a fan of that. Could we add type aliases for the renamed types please?

pmconne:
> Could we add type aliases for the renamed types please?

My vote is to simply revert this PR. Having two names for the same thing is worse (more confusing for users) than having a single slightly-less-perfect name for it.

grigasp:
> > Could we add type aliases for the renamed types please?
> 
> My vote is to simply revert this PR. Having two names for the same thing is worse (more confusing for users) than having a single slightly-less-perfect name for it.

I think having type aliases with `@beta @deprecated`, pointing to the new ones, is also fine.

pmconne:
> I think having type aliases with @beta @deprecated, pointing to the new APIs, is also fine.

Frankly, I think we have already wasted more than enough time on the non-issue of `ECSqlReader` vs `QueryReader`. Unless we can find even a single user who thinks renaming it improves their life in any way, it should be reverted so we can move on to more useful things. We have already established that renaming it will in fact impact existing callers.

mattbjordan:
> > I think having type aliases with @beta @deprecated, pointing to the new APIs, is also fine.
> 
> Frankly, I think we have already wasted more than enough time on the non-issue of `ECSqlReader` vs `QueryReader`. Unless we can find even a single user who thinks renaming it improves their life in any way, it should be reverted so we can move on to more useful things. We have already established that renaming it will in fact impact existing callers.

Hmm, alright since even adding `@beta @deprecated` would require disabling the deprecation warning for the type aliases (in cases where that matters), Paul's swayed me to just revert the name change.

PR #5367, opened 2023-04-07
Concurrent deserialization of schemas

### LINKED ISSUE #5326: Concurrent schema accesses
SchemaContext caches schema before items are loaded. While schema items are loaded all other `SchemaContext.getSchema` callers get cached schema without items. Thus, the getSchema method should be able to deserialize schemas concurrently.

AkshayN5:
Closes #5326


core/ecschema-metadata/src/Context.ts
```diff
@@ -10,9 +10,42 @@ import { SchemaItem } from "./Metadata/SchemaItem";
...
+export class LoadingSchemas extends Array<LoadingSchema> { }
+
+/**
+ * Construct through a function that returns a promise to read the schema.
+ * When loadSchema() is called the first time, it will execute the function to actually begin the promise.
+ * When loadSchema() is called after the first time, it will just return the promise.
+ * This ensures the promise doesn't run until loadSchema() is called, and there's only one loadSchema promise per schema path.
+ * @alpha
+ */
+export class LoadSchema {
```
pmconne, line 36: You have several implementations (this one, `ReadSchemaText`, etc) of a basic pattern. Did you consider defining a generic type instead?

AkshayN5: I'm considering using Generics type with inheritance or using only 1 Generic class instead of LoadSchema and ReadSchemaText classes. What would you suggest me to go ahead with?

AkshayN5: We also have a class DelayedPromise in core\ecschema-metadata\src\DelayedPromise.ts. I'm also looking at the possibility of using that instead of LoadSchema and ReadSchemaText classes.

pmconne: DelayedPromise looks like an improvement.

common/api/ecschema-locaters.api.md
```diff
@@ -19,38 +20,56 @@ export class FileSchemaKey extends SchemaKey {
...
-// @beta
+// @alpha
 export abstract class SchemaFileLocater {
     constructor();
     addSchemaSearchPath(schemaPath: string): void;
     addSchemaSearchPaths(schemaPaths: string[]): void;
+    // (undocumented)
+    addSchemaText(schemaPath: string, readSchemaText: DelayedPromise<string | undefined>): Promise<void>;
+    // (undocumented)
+    addSchemaTextSync(schemaPath: string, readSchemaText: DelayedPromise<string | undefined>): void;
```
ColinKerr, line 31: it seems wrong that a sync method takes a delayed promise

common/api/ecschema-locaters.api.md
```diff
@@ -19,38 +20,56 @@ export class FileSchemaKey extends SchemaKey {
...
+    protected findEligibleSchemaKeys(desiredKey: SchemaKey, matchType: SchemaMatchType, format: string): Promise<FileSchemaKey[]>;
+    protected findEligibleSchemaKeysSync(desiredKey: SchemaKey, matchType: SchemaMatchType, format: string): FileSchemaKey[];
     // (undocumented)
     abstract getSchema<T extends Schema>(key: SchemaKey, matchType: SchemaMatchType, context: SchemaContext): Promise<T | undefined>;
     // (undocumented)
     protected abstract getSchemaKey(data: string): SchemaKey;
     // (undocumented)
+    getSchemaText(schemaPath: string): Promise<string | undefined>;
+    // (undocumented)
+    getSchemaTextSync(schemaPath: string): Promise<string | undefined> | undefined;
```
ColinKerr, line 46: it seems wrong that a sync method returns a promise

common/api/ecschema-metadata.api.md
```diff
@@ -1662,42 +1674,53 @@ export class Schema implements CustomAttributeContainerProps {
...
 }
 
-// @beta (undocumented)
+// @alpha (undocumented)
 export class SchemaCache implements ISchemaLocater {
     constructor();
-    addSchema<T extends Schema>(schema: T): Promise<void>;
-    addSchemaSync<T extends Schema>(schema: T): void;
+    addSchema<T extends Schema>(schema: T, loadSchema?: DelayedPromise<T>): Promise<void>;
+    addSchemaSync<T extends Schema>(schema: T, loadSchema?: DelayedPromise<T>): void;
```
ColinKerr, line 1681: DelayedPromise in a sync method?

core/ecschema-locaters/src/SchemaFileLocater.ts
```diff
@@ -162,7 +251,41 @@ export abstract class SchemaFileLocater {
...
+    const result = new glob.GlobSync(fullPath, { sync: true });
+    for (const match of result.found) {
+      let fileName = path.basename(match, (`.ecschema.${format}`));
+      // TODO: should this be moved or handled elsewhere?
+      // Handles two version file names - SchemaKey.parseString supports only 3 version names.
+      if (/[^\d]\.\d?\d\.\d?\d$/.test(fileName)) {
+        const parts = fileName.split(".");
+        parts.splice(2, 0, "00");
+        fileName = parts.join(".");
+      }
```
ColinKerr, line 266: This code is the same between sync and async methods.  Pull it into a common method to avoid duplication

core/ecschema-locaters/src/SchemaFileLocater.ts
```diff
@@ -193,7 +316,49 @@ export abstract class SchemaFileLocater {
...
+    } else if (matchType === SchemaMatchType.LatestReadCompatible) {
+      twoVersionSuffix = formatString(`.{0}.*.ecschema.${format}`, padStartEx(readVersion, 2, "0"));
+      threeVersionSuffix = formatString(`.{0}.*.*.ecschema.${format}`, padStartEx(readVersion, 2, "0"));
+    } else {
+      twoVersionSuffix = formatString(`.{0}.{1}.ecschema.${format}`, padStartEx(readVersion, 2, "0"), padStartEx(writeVersion, 2, "0"));
+      threeVersionSuffix = formatString(`.{0}.{1}.{2}.ecschema.${format}`, padStartEx(readVersion, 2, "0"), padStartEx(writeVersion, 2, "0"), padStartEx(minorVersion, 2, "0"));
+    }
+
+    const twoVersionExpression = desiredKey.name + twoVersionSuffix;
+    const threeVersionExpression = desiredKey.name + threeVersionSuffix;
```
ColinKerr, line 343: Same, here pull this code out into it's own method to avoid duplication

core/ecschema-locaters/src/test/assets/JSON/AecUnits.ecschema.json
```diff
@@ -0,0 +1,754 @@
+{
```
ColinKerr, line 1: I thought we were going to load the test schemas from the schema packages so we did not need to add so much test data to the repo?

core/ecschema-locaters/src/test/ConcurrentGetSchemaJson.test.ts
```diff
@@ -0,0 +1,135 @@
+/*---------------------------------------------------------------------------------------------
```
ColinKerr, line 1: Should we have added similar tests for xml files?  How about test cases where some schema files are json and some are xml?

core/ecschema-locaters/src/test/assets/JSON/AecUnits.ecschema.json
AkshayN5: I was unable to locate the .json files within the schema packages. Would it be advisable to convert the XML files to JSON for testing purposes, or are there alternative options that may be more suitable?

AkshayN5: Switched to using npm packages for testing now, and as a result, removed the test data.

core/ecschema-locaters/src/test/ConcurrentGetSchemaJson.test.ts
AkshayN5: Done! Added similar tests for XML files now.

common/api/ecschema-locaters.api.md
```diff
@@ -19,38 +20,53 @@ export class FileSchemaKey extends SchemaKey {
     schemaText?: string;
 }
 
-// @beta
+// @alpha
```
ColinKerr, line 23: Is there a reason why you switched from beta to alpha?

AkshayN5: In one of the commits, I had to make the change from beta to alpha because `rush extract-api` was not working without this modification. Nonetheless, I'll roll back this change to beta.

pmconne (approved):

core/ecschema-metadata/src/Context.ts
```diff
@@ -3,16 +3,29 @@
...
+}
+/**
+ * @alpha
+ */
+export class LoadedSchemas extends Array<Schema> { }
+/**
+ * @alpha
  */
-export class SchemaMap extends Array<Schema> { }
+export class LoadingSchemas extends Array<LoadingSchema> { }
```
pmconne, line 28: You're exporting a type that is an array of a non-exported type. Did you intend to export `LoadingSchema`?

pmconne: It looks to me like the only usage of `LoadedSchemas` and `LoadingSchemas` is as `private` fields, in which case I don't see a reason to export either of them. Did I overlook something?

AkshayN5: Your observation is correct.  Currently, I intend to use `LoadedSchemas` and `LoadingSchemas` as private members of the `SchemaCache` class. As such, I'll remove the export modifier from them.

core/ecschema-locaters/src/SchemaFileLocater.ts
AkshayN5: I created two methods named `getTwoVersionExpression` and `getThreeVersionExpression`. However, I think it is possible to merge both methods into a single one that would return an array of VersionExpressions. What's your opinion on this?

common/api/ecschema-locaters.api.md
AkshayN5: I removed the sync method since it was unnecessary, as we simply take a promise from the parameter and append it to an array.

common/api/ecschema-locaters.api.md
AkshayN5: I removed the sync method as the async method was the only one that was utilizing it internally, making the sync method unnecessary.

common/api/ecschema-metadata.api.md
AkshayN5: The `DelayedPromise` is removed from the sync method's parameters. The sync method's behaviour is unchanged, and it still requires that the schema be fully loaded before the method is called.

core/ecschema-metadata/src/Context.ts
AkshayN5: Removed the export modifiers from `LoadedSchemas` and `LoadingSchemas`

common/api/ecschema-locaters.api.md
AkshayN5: Rolled back the testing from alpha to beta

core/ecschema-metadata/src/Metadata/Schema.ts
```diff
@@ -522,6 +523,17 @@ export class Schema implements CustomAttributeContainerProps {
     return schemaJson as SchemaProps;
   }
 
+  /**
+   * Serializes the schema to a string of XML
+   * @alpha
+   */
+  public async toXmlString(): Promise<string> {
```
ColinKerr, line 530: We cannot add this here because xmldom is a dev dependency ... we do not want to make it a full dependency

AkshayN5:
Closing this PR due to the approach still needing dev work to avoid over-complicated logic. Here is another PR #5441 with a different approach which is simpler and exposes less of the logic.

PR #5371, opened 2023-04-10
Make ECSqlReader an AsyncIterableIterator

mattbjordan:
[ECSqlReader](https://www.itwinjs.org/reference/core-common/imodels/ecsqlreader/) now implements [AsyncIterableIterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator) which means that the class/object itself can be iterated on. This makes it function inline with how the - now deprecated - [`.query` ](https://www.itwinjs.org/reference/core-backend/ecdb/ecdb/query/) method worked.

Iterating over an ECSqlReader as an iterable might look like:
```ts
 for await (const row of iModel.createQueryReader(queryString)) {
    id = row[0]; // or 'row.id'
 }
```
<br>

It will be easy to replace existing uses of .query by simply replacing `.query` with `.createQueryReader`:
```ts
// previous calls to .query
for await (const row of iModel.query(queryString)) { ... }
// new call to .createQueryReader
for await (const row of iModel.createQueryReader(queryString)) { ... }
```

<br>

(Link to a previous discussion on this change: https://github.com/iTwin/itwinjs-core/pull/5347#issuecomment-1497612367)


pmconne:
Looks like we gotta upgrade [xml2js](https://github.com/advisories/GHSA-776f-qx25-q3cc).

pmconne (approved):

mattbjordan:
> Looks like we gotta upgrade @azure/storage-blob for [xml2js](https://github.com/advisories/GHSA-776f-qx25-q3cc) (in a separate PR)

I've updated @azure/storage-blob locally and still have the audit error. It looks like [they've updated the xml2js dependency](https://github.com/Azure/azure-sdk-for-js/pull/25502), but it perhaps hasn't gotten published yet. [@azure/storage-blob was last updated 11 days ago](https://www.npmjs.com/package/@azure/storage-blob). I'll check again tomorrow morning.

mattbjordan:
> > Looks like we gotta upgrade @azure/storage-blob for [xml2js](https://github.com/advisories/GHSA-776f-qx25-q3cc) (in a separate PR)
> 
> I've updated @azure/storage-blob locally and still have the audit error. It looks like [they've updated the xml2js dependency](https://github.com/Azure/azure-sdk-for-js/pull/25502), but it perhaps hasn't gotten published yet. [@azure/storage-blob was last updated 11 days ago](https://www.npmjs.com/package/@azure/storage-blob). I'll check again tomorrow morning.

Seems like it's still not updated yet. However, it looks like they are about to merge the change into their stable branch ([link](https://github.com/Azure/azure-sdk-for-js/pull/25518)). Not sure when they'll publish the update though.

pmconne:
We used to only run `rush audit` on 2 out of 3 platforms. Now we apparently run it on all 3, so no way to tell if CI jobs would otherwise succeed.
I'll temporarily ignore it to unblock PRs and file an issue to update once fix becomes available.

PR #5397, opened 2023-04-14
Prune unnecessary deps and update rush audit

aruniverse:
- audit.js
  - remove outdated cves
  - xml2js still stuck on azurite update
- core-backend
  - unnecessarily pulling in azure/storage-blob
- remove dep on bas64-js, can just use native node apis (since we support node18+)
- prune unnecessary deps across all pkgs


core/backend/package.json
```diff
@@ -62,6 +62,7 @@
     "@itwin/core-geometry": "workspace:*",
     "@itwin/core-webpack-tools": "workspace:*",
     "@itwin/eslint-plugin": "^4.0.0-dev.33",
+    "@itwin/object-storage-azure": "^1.5.0",
```
pmconne, line 65: Why has this moved to `devDependencies`? It is used by `IModelHost`.

core/frontend/package.json
```diff
@@ -62,12 +62,10 @@
     "@itwin/core-quantity": "workspace:*",
     "@itwin/certa": "workspace:*",
     "@itwin/eslint-plugin": "^4.0.0-dev.33",
-    "@itwin/webgl-compatibility": "workspace:*",
```
pmconne, line 65: Why is this now only a dev dependency?

aruniverse: we made this a direct dep, since its pure / side effect free, a few weeks back.
it was somehow still listed as both direct and dev dep, just removing the unnecessary dev dep

core/backend/package.json
aruniverse: ah this is a good catch, my bad. thought this was only being used for testing. will revert

core/frontend/package.json
pmconne: Thanks, sorry, I misread which section had changed.

pmconne (approved):

core/backend/package.json
aruniverse: guess im getting old and my eyesight is waning
image

updated

dassaf4 (approved):

common/config/rush/browser-approved-packages.json
```diff
@@ -2,6 +2,18 @@
 {
   "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/approved-packages.schema.json",
   "packages": [
+    {
+      "name": "@azure/core-htp",
```
tcobbs-bentley, line 6: What is this? It looks like a typo, but you have core-http below, so I'm not sure.

aruniverse: Weird. This is autogenerated. Will manually remove

aruniverse: reverted

grigasp (approved):

ColinKerr (approved):

PR #5399, opened 2023-04-14
Added Ray3d.intersectWithTriangle

saeeedtorabi:
- This new method makes ray-triangle intersection speed 2-3 times faster
- Added few other new APIs
- Added lots of accuracy and performance tests


core/geometry/src/geometry3d/BarycentricTriangle.ts
```diff
@@ -560,7 +572,7 @@ export class BarycentricTriangle {
   public snapLocationToEdge(
     location: TriangleLocationDetail,
     distanceTolerance: number = Geometry.smallMetricDistance,
```
saeeedtorabi, line 574: I dont remember why I made this chage. I think you told me to but dont remember why!

dassaf4: It's good. We want that tighter default snap tolerance to correspond to the snap tol in the new method.

dassaf4:
Consider introducing the new method, and contrasting it to similar existing methods, in NextVersion.md. When would someone want to use it over the existing methods, and when not?

dassaf4 (approved):

core/geometry/src/test/geometry3d/Ray3d.test.ts
```diff
@@ -282,3 +288,297 @@ describe("Ray3d", () => {
...
+    triangle = BarycentricTriangle.createXYZXYZXYZ(2, 0, 0, 2, 0, 0, 2, 0, 0);
+    intersectionPoint = ray.intersectionWithTriangle(triangle.points[0], triangle.points[1], triangle.points[2]);
+    ck.testUndefined(
+      intersectionPoint,
+      "expect no intersection when we have a degenerate triangle with three equal vertexes"
+    );
+
+    origin = Point3d.create(0, 0, 0);
+    direction = Vector3d.create(1, 1, 0);
+    ray = Ray3d.create(origin, direction);
```
saeeedtorabi, line 417: I need to redefine triangle here:

triangle = BarycentricTriangle.createXYZXYZXYZ(2, 0, 0, 10, 0, 0, 2, 10, 0);

I added it locally and the test passed.

I also added a test for ray intersects inside a triangle and it passed too. Will added them all after you reviewed this.

dassaf4 (approved):

ben-polinsky (approved):

PR #5402, opened 2023-04-17
Code scope type fix

KyryloVolotovskyi:
- Fixed the issue of Transformer fixing CodeScopes for elements that have CodeSpec with Repository type
- Added warning

### DIFF
--- a/core/backend/src/IModelElementCloneContext.ts
+++ b/core/backend/src/IModelElementCloneContext.ts
@@ -5,14 +5,17 @@
 /** @packageDocumentation
  * @module iModels
  */
-import { Id64, Id64String } from "@itwin/core-bentley";
-import { Code, CodeScopeSpec, CodeSpec, ElementProps, IModel, PropertyMetaData, RelatedElement } from "@itwin/core-common";
+import { Id64, Id64String, Logger } from "@itwin/core-bentley";
+import { Code, CodeScopeSpec, CodeSpec, ElementProps, PropertyMetaData, RelatedElement } from "@itwin/core-common";
 import { IModelJsNative } from "@bentley/imodeljs-native";
 import { SubCategory } from "./Category";
 import { Element } from "./Element";
 import { IModelDb } from "./IModelDb";
 import { IModelHost } from "./IModelHost";
 import { SQLiteDb } from "./SQLiteDb";
+import { BackendLoggerCategory } from "./BackendLoggerCategory";
+
+const loggerCategory = BackendLoggerCategory.IModelElementCloneContext;
 
 /** The context for transforming a *source* Element to a *target* Element and remapping internal identifiers to the target iModel.
  * @beta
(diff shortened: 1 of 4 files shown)


core/backend/src/IModelElementCloneContext.ts
```diff
@@ -153,8 +156,9 @@ export class IModelElementCloneContext {
     if (this.isBetweenIModels) {
       // The native C++ cloneElement strips off federationGuid, want to put it back if transformation is between iModels
       targetElementProps.federationGuid = sourceElement.federationGuid;
-      if (CodeScopeSpec.Type.Repository === this.targetDb.codeSpecs.getById(targetElementProps.code.spec).scopeType) {
-        targetElementProps.code.scope = IModel.rootSubjectId;
+      const targetElementCodeScopeType = this.targetDb.codeSpecs.getById(targetElementProps.code.spec).scopeType;
+      if (CodeScopeSpec.Type.Repository === targetElementCodeScopeType) {
```
kabentley, line 160: i don't understand this change. Can you please explain?

KyryloVolotovskyi: The problem is that elements can become not unique after the change of code scope type which will cause the clash of elements. The real case example that was faced: During IModel transformation,  a couple of elements have CodeSpec which has CodeScopeSpec.Type.Repository, but if CodeScope is set to IModel.rootSubjectId, codes start to clash and the transformation eventually fails. After long discussions, it was decided that those modifications should be done by handlers, which means that transformer shouldn't change anything.

pmconne: If there is an issue filed, please link to it.

kabentley: Also, `IModel.rootSubjectId` is a constant (0x1), so it will always be the same in every iModel. So the "targetElement.code.scope" should always already be 0x1 and i don't see what removing that line will accomplish (though it is useless).

KyryloVolotovskyi: > Also, `IModel.rootSubjectId` is a constant (0x1), so it will always be the same in every iModel. So the "targetElement.code.scope" should always already be 0x1 and i don't see what removing that line will accomplish (though it is useless).

Yes, the expectation is that "targetElement.code.scope" should always already be 0x1 but in reality it is not always like that. I have moved the fix to the transformer and have created the pull request there. Here is the link to it: https://github.com/iTwin/imodel-transformer/pull/25. I have added a test, which causes an error in case the fix is not present.

kabentley: but i think you *should* remove the assignment of ` targetElementProps.code.scope = IModel.rootSubjectId` if you're testing that in your subclass.

KyryloVolotovskyi: I have overridden `cloneElement` in `IModelCloneContext` (in Transformer, in new [Pull Request](https://github.com/iTwin/imodel-transformer/pull/25)) and have removed the assignment there.

PR #5406, opened 2023-04-17
Prune unnecessary deps and update rush audit [release/3.7.x]

aruniverse:
Based on #5397.

PR #5408, opened 2023-04-18
update object-storage dependencies

paulius-valiunas:


paulius-valiunas:
#5406 already covers this

PR #5409, opened 2023-04-18
Replace IModelApp.getAccessToken's usage of 'this' with 'IModelApp'

jason-crow:
In order to prevent needing to add a `.bind(IModelApp)` when using `IModelApp.getAccessToken` like this:
`const getAccessToken = props.getAccessToken ?? IModelApp.getAccessToken` 
Replace the `IModelApp.getAccessToken` implementation's usage of `this` with `IModelApp`
Since this is a static routine there is no reason to complicate its usage with the `this` keyword

### DIFF
--- a/common/changes/@itwin/core-frontend/fix-getaccesstoken_2023-04-18-13-34.json
+++ b/common/changes/@itwin/core-frontend/fix-getaccesstoken_2023-04-18-13-34.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "replace IModelApp.getAccessToken's implementation usage of 'this' with 'IModelApp'",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/core/frontend/src/IModelApp.ts
+++ b/core/frontend/src/IModelApp.ts
@@ -540,7 +540,7 @@ export class IModelApp {
    */
   public static async getAccessToken(): Promise<AccessToken> {
     try {
-      return (await this.authorizationClient?.getAccessToken()) ?? "";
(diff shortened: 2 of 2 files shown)


pmconne (changes requested):
Fix the calling code instead please.
`const getAccessToken = props.getAccessToken ?? (() => IModelApp.getAccessToken)`

jason-crow:
> Fix the calling code instead please. `const getAccessToken = props.getAccessToken ?? (() => IModelApp.getAccessToken())`

There is a consideration of volume here, this kind of mistake that would need your suggested fix or a .bind(IModelApp) is relatively common. Is there an advantage to the usage of `this` that makes changing this once here not better than changing it in many places elsewhere that may even come up again?

pmconne:
> There is a consideration of volume here, this kind of mistake that would need your suggested fix or a .bind(IModelApp) is relatively common.

Let's pretend `IModelApp.getAccessToken` is an instance method rather than a static method.
```
const iModelApp = new IModelApp();
const getAccessToken = props.getAccessToken ?? iModelApp.getAccessToken;
```

The exact same bug. `static` is irrelevant, `this` behaves the same. People writing JavaScript need to understand how `this` works.

pmconne:
Here is the [actual bug](https://github.com/iTwin/eslint-plugin/blob/192b22fb9fc04f207fea303c68f6f19fe10e98df/dist/configs/itwinjs-recommended.js#L231) to fix.

pmconne:
> I don't really see the advantage to pretending static methods should behave as though they could be prototypical

Except classes in JavaScript are objects and have prototypes. We make productive use of this feature all over the place. Lint rule should be configured to complain when you make this mistake.

pmconne:
https://github.com/iTwin/eslint-plugin/pull/15

PR #5418, opened 2023-04-19
update rush and audit steps

aruniverse:
- update rush to the latest, see [changelog](https://github.com/microsoft/rushstack/blob/main/apps/rush/CHANGELOG.md)
- update pnpm to v7, see [release notes](https://github.com/pnpm/pnpm/releases?q=v7.&expanded=true)
- kill off custom internal audit script, use [pnpm audit](https://pnpm.io/7.x/cli/audit) instead via [rush-pnpm](https://rushjs.io/pages/commands/rush-pnpm/) wrapper
- use [pnpm.auditConfig.ignoreCves](https://pnpm.io/7.x/package_json#pnpmauditconfigignorecves) to maintain supression list


common/config/rush/pnpm-lock.yaml
```diff
@@ -52,31 +52,31 @@ importers:
...
+      semver: 7.5.0
       ws: 7.5.9
     devDependencies:
       '@itwin/build-tools': link:../../tools/build
       '@itwin/core-bentley': link:../bentley
       '@itwin/core-common': link:../common
       '@itwin/core-geometry': link:../geometry
       '@itwin/core-webpack-tools': link:../../tools/webpack-core
-      '@itwin/eslint-plugin': 4.0.0-dev.33_eslint@8.36.0+typescript@5.0.2
+      '@itwin/eslint-plugin': 4.0.0-dev.35_voubu7prgxjfsfbgx5d4sqnwiy
```
pmconne, line 72: Do you want to pin this to dev.33 and deal with the peer deps and new recommended config separately, or lump it all in with this PR?
FYI `auto-install-peers` in .npmrc will install the peer deps into common/temp/node_modules but `npm run lint` will try and fail to find them in the specific packages' own node_modules directories.

aruniverse: im running into issues locally even after not changing the npmrc to auto-install-peers and pinning down to dev 33
image

@paulius-valiunas i'd like for us to rethink the stance made in your eslint-plugin pr; looks like the assumption pkg managers installing peers for you doesnt work well in a monorepo

common/config/rush/.pnpmfile.cjs
```diff
@@ -1,8 +1,5 @@
 function readPackage(pkg) {
-  if (
-    (pkg.name == "typedoc") &&
-    pkg.dependencies
-  ) {
+  if (pkg.name == "typedoc" && pkg.dependencies) {
```
paulius-valiunas, line 2: shouldn't we just update typedoc instead of doing this hack?

common/config/rush/pnpm-lock.yaml
paulius-valiunas: by the way, seems like pnpm 6.32 also supports this, so we didn't have to update rush just for this. It's still a good idea to keep it up to date, though.
Source: https://twitter.com/pnpmjs/status/1528300665063321600?lang=en

common/config/rush/.pnpmfile.cjs
ben-polinsky: Yes, Yash and I are about finished with the upgrade. Will publish bemetalsmith@5.0 today which should fix any build issues. You can try it out now if you'd like by using version 5.0.1-dev.2.

pmconne (approved):

paulius-valiunas (approved):

calebmshafer (approved):
This is so much better than my custom script. Thanks!

PR #5437, opened 2023-04-24
CloudSQLite: Add bcvhttplog and clientidentifier support

nick4598:
https://github.com/iTwin/imodel-native/pull/242

### DIFF
--- a/core/backend/src/CloudSqlite.ts
+++ b/core/backend/src/CloudSqlite.ts
@@ -48,6 +48,10 @@ export namespace CloudSqlite {
     writeable?: boolean;
     /** if true, container is attached in "secure" mode (blocks are encrypted). Only supported in daemon mode. */
     secure?: boolean;
+    /** An id which enhances logging provided by CloudSQLite.
+     * This ID will be used to identify, in log messages, all CloudSQLite client connections (also known as database connections) opened using this CloudContainer.
+     */
+    clientIdentifier?: string;
   }
 
   /** Returned from `CloudContainer.queryDatabase` describing one database in the container */
@@ -62,6 +66,39 @@ export namespace CloudSqlite {
     readonly transactions: boolean;
     /** the state of this database. Indicates whether the database is new or deleted since last upload */
     readonly state: "" | "copied" | "deleted";
+  }
+
+  /** Filter options passed to CloudContainer.queryHttpLog */
+  interface BcvHttpLogFilterOptions {
+    /** only return rows whose ID is >= the provided id */
+    startFromId?: number;
+    /** only return rows whose endTime is null OR >= the provided endTime. */
+    finishedAtOrAfterTime?: string;
+    /** only return rows with a non-null end_time. */
+    showOnlyFinished?: boolean;
+  }
+
+  /** Returned from 'CloudContainer.queryHttpLog' describing a row in the bcv_http_log table. */
+  export interface BcvHttpLog {
+    /** Unique, monotonically increasing id value */
+    readonly id: number;
+    /** Time request was made, as iso-8601 */
+    readonly startTime: string;
+    /** Time reply received, as iso-8601 (or NULL) */
+    readonly endTime: string | undefined;
+    /** "PUT", "GET", etc. */
+    readonly method: string;
+    /** Name of the client that caused this request. Name will be "prefetch" if it is a request triggered by a prefetch.
+     *  Name of client can be configured by passing a 'clientIdentifier' to a CloudContainer's ContainerProps.
+     *  Empty string otherwise.
+     */
+    readonly client: string;
+    /** Log message associated with request */
+    readonly logmsg: string;
+    /** URI of request */
+    readonly uri: string;
+    /** HTTP response code (e.g. 200) */
+    readonly httpcode: number;
   }
 
   /** Properties for accessing a CloudContainer */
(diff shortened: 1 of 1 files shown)


core/backend/src/CloudSqlite.ts
```diff
@@ -48,6 +48,10 @@ export namespace CloudSqlite {
     writeable?: boolean;
     /** if true, container is attached in "secure" mode (blocks are encrypted). Only supported in daemon mode. */
     secure?: boolean;
+    /** An id which enhances logging provided by CloudSQLite.
+     * This ID will be used to identify, in log messages, all CloudSQLite client connections (also known as database connections) opened using this CloudContainer.
```
kabentley, line 52: I assume this only applies in daemon mode? BTW, we always use "Id" rather than "ID".

nick4598: Changed it to Id.

A clientIdentifier could be provided in the daemonless case, it probably just wouldn't be as useful. BCV_HTTP_LOG table would still exist in daemonless mode (and uses clientIdentifier), but I believe we only really care about it in the daemon case.

kabentley: i think you should document that as only being relevant for daemon mode. It's confusing to explain why you need to identify yourself to yourself without a daemon.

nick4598: "This Id is mostly only relevant to give more clarity to logs produced running in daemon mode, where there are usually many active CloudContainers and by extension, many ongoing HTTP requests."

What do you think of the above?

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -128,6 +129,135 @@ describe("CloudSqlite", () => {
...
+
+    // endTime to exclude these first 2 entries in later queries.
+    await BeDuration.wait(10);
+    const endTime = new Date().toISOString();
+
+    rows = container.queryHttpLog({startFromId: 2});
+    expect(rows.length).to.equal(1);
+    expect(rows[0].id).to.equal(2);
+
+    container.acquireWriteLock("test");
```
kabentley, line 158: use `CloudSqlite.withWriteLock` (see other tests)

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -128,6 +129,135 @@ describe("CloudSqlite", () => {
...
+    const container = CloudSqlite.createCloudContainer(containerProps);
+    container.initializeContainer();
+    container.connect(caches[1]);
+
+    container.acquireWriteLock("test");
+    await CloudSqlite.uploadDb(container, {localFileName: testBimFileName, dbName: basename(testBimFileName)});
+    container.releaseWriteLock();
+    container.checkForChanges();
+
+    let db: IModelJsNative.DgnDb = IModelDb.openDgnDb({ path: basename(testBimFileName) }, OpenMode.Readonly, undefined, {container});
```
kabentley, line 210: remove type override

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -128,6 +129,135 @@ describe("CloudSqlite", () => {
...
+    const stmt = new SqliteStatement("PRAGMA bcv_client");
```
kabentley, line 211: use `withSqliteStatement`

kabentley: I'm sorry, but I don't see what this test actually tests.

nick4598: It tests that the cloudsqlitelogid defined on the container shows up on databases opened using that container

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
nick4598: Why is that a type override? Thats just the type that IModelDb.openDgnDb returns, right?

kabentley: yes, the type is defined by the method, you should not override it.

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -128,6 +129,135 @@ describe("CloudSqlite", () => {
     await CloudSqliteTest.uploadFile(testContainers[2], caches[0], "testBim", testBimFileName);
   });
 
+  it("should query bcvHttpLog", async () => {
```
kabentley, line 132: @nick4598, let's just leave this test out of this PR, because i've made substantial changes to this stuff on [this PR](https://github.com/iTwin/itwinjs-core/pull/5455) and all this will cause lots of conflicts. After the next native build, we can combine this test with the above tests.

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
nick4598: withSqliteStatement is not defined on IModelJsNative.DgnDb, so I swithced to a SqliteDb. Is that fine? I don't know if I should be using SqliteDb here since I care more that this works for a DgnDb, but they do go through the same code path.

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
nick4598: Is it a bad thing to explicitly mention what type I'm going to get? I don't think I can override in this way, I would get a build error if I put let db: number

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
kabentley: it's on `IModelDb`, but see my other comment.

nick4598: Well I removed should query bcvhttplog, should I remove the cloudsqlitelogid test as well? Will that one also conflict?

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
kabentley: yes it's a bad thing.

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
nick4598: done

kabentley (approved):

PR #5441, opened 2023-04-25
Fix Async schema loading so schemas cannot be used until they are completely loaded

ColinKerr:
Fixes issue #5326 by caching promises to load schemas and returning information about partially loaded schemas instead of returning partially loaded schemas.  This ensures partially loaded schemas are not accessed and allows us to test for circular schema references with just schema keys for schemas and their references


core/ecschema-metadata/src/Context.ts
```diff
@@ -47,23 +48,66 @@ export interface ISchemaItemLocater {
...
+
+  private removeSchemaPromise(schemaKey: Readonly<SchemaKey>) {
+    this._schemaPromises = this._schemaPromises.filter((value: LazyLoadedSchema) => !value.matches(schemaKey, SchemaMatchType.Latest));
+  }
+
+  /**
+   * Returns true if the schema exists in either the schema cache or the promise cache.  SchemaMatchType.Latest used.
+   * @param schemaKey The key to search for.
+   */
+  public schemaExits(schemaKey: Readonly<SchemaKey>): boolean {
```
pmconne, line 80: `s/Exits/Exists`?

ColinKerr: fixed

presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
```diff
@@ -59,7 +59,7 @@ interface FormattingProps {
   persistenceUnitName: string;
 }
 
-async function getFormattingProps(schemaLocater: ISchemaLocater, options: FormatOptions): Promise<FormattingProps | undefined> {
```
ColinKerr, line 62: I think it was a mistake to use ISchemaLocater here ... in all cases a SchemaContext is used and using a locater here would be bad because it would load the schema every time the method was called instead of using a version that was already loaded

docs/changehistory/NextVersion.md
```diff
@@ -306,14 +325,37 @@ In addition to upgrading iTwin.js core dependencies to `4.0`, there are some oth
...
+
+- Added `SchemaInfo` interface with schema keys for a schema and it's references.  `Schema` implicitly supports this interface.
+- Some beta components had breaking changes and were moved to internal:
+  - `SchemaGraph`
+    - Now supports working with a `SchemaInfo` and a `SchemaContext` necessitating the init be made async.
+  - `SchemaMap`
+    - Use `Array<Schema>` in it's place.
+  - `SchemaCache`
+    - Updated to support caching partially loaded schemas, use `SchemaContext` to cache schemas in it's place.
+- Added helper method `SchemaFileUtility.writeSchemaToXmlString` to write schema xml to a string ``
```
grigasp, line 360:
```suggestion
- Added helper method `SchemaFileUtility.writeSchemaToXmlString` to write schema xml to a string
```

presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
grigasp: I see what you mean, but from API standpoint before your changes, I don't see why these methods couldn't have taken `ISchemaLocater`. All these functions care is to get a schema - they don't need to worry about where it comes from (fresh or from cache) - it's up to the caller to make sure they pass an `ISchemaLocater` that doesn't load schemas unnecessarily. And as far as I can tell that was **almost** happening except for passing the third argument to `getSchema`. However, due to what I believe is an API flaw, it was unnecessary.

In my opinion this confusion is caused by `SchemaContext` saying that it implements `ISchemaLocater` when it fact it doesn't. I think with changes in this PR it's even more apparent - `ISchemaLocater.getSchema` takes a required `SchemaContext` as the third argument and uses it to find referenced schemas. `SchemaContext.getSchema` doesn't use the third argument and instead looks for referenced schemas in its own cache. 

Consider this piece of code:
```ts
// add some schemas to `context`
const locater: ISchemaLocater = new SchemaContext();
const schemas = await locater.getSchema(key, matchType, context);
// referenced schemas in `context` won't be found
```
`getSchema` docs say that the third argument is going to be used for loading schema references, but it won't in this case.

To summarize, I think:
1. The change you made to our `getKoq` and `getFormattingProps` is okay.
2. `SchemaContext` should stop saying it implements `ISchemaLocater`.

grigasp (changes requested):

presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
grigasp: I see what you mean, but from API standpoint before your changes, I don't see why these methods couldn't have taken `ISchemaLocater`. All these functions care is to get a schema - they don't need to worry about where it comes from (fresh or from cache) - it's up to the caller to make sure they pass an `ISchemaLocater` that doesn't load schemas unnecessarily. And as far as I can tell that was **almost** happening except for passing the third argument to `getSchema`. However, due to what I believe is an API flaw, it was unnecessary.

In my opinion this confusion is caused by `SchemaContext` saying that it implements `ISchemaLocater` when in fact it doesn't. I think with changes in this PR it's even more apparent - `ISchemaLocater.getSchema` takes a required `SchemaContext` as the third argument and says it's going to use it to find referenced schemas. `SchemaContext.getSchema` doesn't use the third argument and instead looks for referenced schemas in its own cache. 

Consider this piece of code:
```ts
// add some schemas to `context`
const locater: ISchemaLocater = new SchemaContext();
const schemas = await locater.getSchema(key, matchType, context);
// referenced schemas in `context` won't be found
```
`getSchema` docs say that the third argument is going to be used for loading schema references, but it won't in this case.

To summarize, I think:
1. The change you made to our `getKoq` and `getFormattingProps` is okay.
2. `SchemaContext` should stop saying it implements `ISchemaLocater`.

core/ecschema-metadata/src/Metadata/Schema.ts
```diff
@@ -567,6 +567,9 @@ export class Schema implements CustomAttributeContainerProps {
     return schemaXml;
   }
 
+  /**
+   * Loads the schema header (name, version alias, label an description) from the input SchemaProps
```
calebmshafer, line 571:
```suggestion
   * Loads the schema header (name, version alias, label and description) from the input SchemaProps
```

core/ecschema-metadata/src/Metadata/Schema.ts
```diff
@@ -595,10 +598,25 @@ export class Schema implements CustomAttributeContainerProps {
       this._description = schemaProps.description;
   }
 
+  /**
+   * Loads the schema header (name, version alias, label an description) from the input SchemaProps
```
calebmshafer, line 602:
```suggestion
   * Loads the schema header (name, version alias, label and description) from the input SchemaProps
```

presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
ColinKerr: That is a good point, let me review why SchemaContext implements ISchemaLocater and see if I can remove it.

ColinKerr: I removed ISchemaLocater from SchemaContext and updated docs of ISchemaLocater to make its purpose more clear

common/changes/@itwin/ecschema-editing/cgk-async-schema-test_2023-05-10-22-29.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-editing",
+      "comment": "Updated to use modified SchemaGraph",
```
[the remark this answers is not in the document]
ColinKerr: I changed the comment to remove the reference to schema graph

grigasp (approved):

pmconne:
> I did both

You did?
image

docs/changehistory/NextVersion.md
```diff
@@ -157,6 +161,21 @@ for await (const row of iModel.createQueryReader("SELECT * FROM bis.Element")) {
...
+- The interface `UnitConversion` has been renamed to [UnitConversionProps]($quantity).
+
+#### @itwin/ecschema-metadata
+
+- Renamed Interfaces
+  - The interface `FormatProps` has been changed to a type alias [SchemaItemFormatProps]($ecschema-metadata).
+  - The interface `UnitProps` has been renamed to [SchemaItemUnitProps]($ecschema-metadata).
+- `ISchemaLocater`
+  - `getSchema` and `getSchemaSync` methods now take a `Readonly<SchemaKey>` instead of a `SchemaKey` and the `SchemaContext` parameter is required
+  - Added `getSchemaInfo` method to facilitate asynchronous schema loading
```
pmconne, line 177: - The `FormatProps` interface has been replaced with the [SchemaItemFormatProps]($ecschema-metadata) type alias.
- The `UnitProps` interface has been renamed to [SchemaItemUnitProps]($ecschema-metadata).
- [ISchemaLocater.getSchema]($ecschema-metadata) and [ISchemaLocater.getSchemaSync]($ecschema-metadata) now take a `Readonly<SchemaKey>` instead of a [SchemaKey]($ecschema-metadata) and the [SchemaContext]($ecschema-metadata) parameter is no longer optional.
- Added [ISchemaLocater.getSchemaInfo]($ecschema-metadata) to facilitate asynchronous schema loading.

docs/changehistory/NextVersion.md
```diff
@@ -306,14 +325,37 @@ In addition to upgrading iTwin.js core dependencies to `4.0`, there are some oth
...
 
-- The interface 'UnitProps' has been renamed to [SchemaItemUnitProps]($ecschema-metadata).
+### Other minor API changes
+
+- Added `SchemaInfo` interface with schema keys for a schema and it's references.  `Schema` implicitly supports this interface.
+- Some beta components had breaking changes and were moved to internal:
+  - `SchemaGraph`
+    - Now supports working with a `SchemaInfo` and a `SchemaContext` necessitating the init be made async.
+  - `SchemaMap`
+    - Use `Array<Schema>` in it's place.
```
pmconne, line 357:
```suggestion
    - Use `Array<Schema>` in its place.
```

docs/changehistory/NextVersion.md
pmconne, line 177:
```suggestion
- The `FormatProps` interface has been replaced with the [SchemaItemFormatProps]($ecschema-metadata) type alias.
- The `UnitProps` interface has been renamed to [SchemaItemUnitProps]($ecschema-metadata).
- [ISchemaLocater.getSchema]($ecschema-metadata) and [ISchemaLocater.getSchemaSync]($ecschema-metadata) now take a `Readonly<SchemaKey>` instead of a [SchemaKey]($ecschema-metadata) and the [SchemaContext]($ecschema-metadata) parameter is no longer optional.
```

pmconne (approved):

calebmshafer (approved):
LGTM. Definitely need to work out some of the other inconsistencies in the APIs in addition to the ones that @grigasp mentioned but looks good.

PR #5444, opened 2023-04-25
Docs: TypeScript 5 compatibility

ben-polinsky:
Upgrade @itwin/build-tools deps and bump bemetalsmith version in builds. Add setup for auth-* packages


common/config/rush/pnpm-lock.yaml
aruniverse, line 1: revert this file to as is on master, then run plain `rush update` to pull in the right doc tool changes. `update --full` will pull in the broken eslint pkg

aruniverse (approved):

PR #5452, opened 2023-04-27
Presentation: Support Binary/BeGuid type properties

medikornov:
Closes: https://github.com/orgs/iTwin/projects/20/views/1?pane=issue&itemId=22725591
native PR: https://github.com/iTwin/imodel-native/pull/254
imodel-native: https://github.com/iTwin/imodel-native/tree/presentation/support-guid-type-properties


full-stack-tests/presentation/src/frontend/Content.test.ts
```diff
@@ -399,6 +399,44 @@ describe("Content", () => {
...
+              isDisplayed: true,
+              doNotHideOtherPropertiesOnDisplayOverride: true
+            }]
+          }],
+        }],
+      };
+      const content = await Presentation.presentation.getContent({
+        imodel: imodelConnection,
+        rulesetOrId: ruleset,
+        keys: new KeySet([{ className: instanceKey!.className, id: instanceKey!.id }]),
```
grigasp, line 428:
```suggestion
        keys: new KeySet([instanceKey!]),
```

full-stack-tests/presentation/src/frontend/Content.test.ts
```diff
@@ -399,6 +399,44 @@ describe("Content", () => {
...
+      const ruleset: Ruleset = {
+        id: Guid.createValue(),
+        rules: [{
+          ruleType: RuleTypes.Content,
+          specifications: [{
+            specType: ContentSpecificationTypes.SelectedNodeInstances,
+            propertyOverrides: [{
+              name: "FederationGuid",
+              isDisplayed: true,
+              doNotHideOtherPropertiesOnDisplayOverride: true
```
grigasp, line 420: I believe in this test we only care about the `FederationGuid` property, so don't need others to be loaded too. In that case, the `doNotHideOtherPropertiesOnDisplayOverride` attribute is unnecessary.

grigasp (changes requested):

full-stack-tests/presentation/src/frontend/Content.test.ts
medikornov: fixed

full-stack-tests/presentation/src/frontend/Content.test.ts
medikornov: fixed

grigasp (approved):

PR #5455, opened 2023-04-27
Add support for internal Codes to `CodeService`

kabentley:
Add a separate CloudSqlite index for internal codes (see CodeService.md).

Also:
 - add `BlobContainerService` interface for creating blob containers
 - add `CloudSqlite.DbAccess` class for wrapping methods that require write lock
 - add `PropertyDb` implementation to `PropertyStore` interface
 - add `VersionedSqliteDb` base class for SqliteDbs that store an api version enforced at open


core/backend/src/BlobContainerService.ts
```diff
@@ -0,0 +1,142 @@
...
+
+// spell:ignore datacenter
+
+import { AccessToken, Id64String } from "@itwin/core-bentley";
+
+/**
+ * Types and functions for creating, deleting and authorizing access to cloud-based blob containers for an iTwin.
+ * @beta
+ */
+export namespace BlobContainer {
```
kabentley, line 17: @wgoehrig @MichaelSwigerAtBentley This interface is the outline for what the REST service for the BlobContainer service should like like.

kabentley:
@ben-polinsky  do you have any idea what the doc error means on this PR?

ben-polinsky:
> @ben-polinsky do you have any idea what [the doc error](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2114782&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e&t=71c8ff40-78b9-5b98-0fd8-62f51b84014f) means on this PR?

Not exactly, but looking into it now.

docs/learning/backend/CodeService.md
```diff
@@ -2,7 +2,7 @@
 
 ## Codes Overview
 
-[iTwin.js](https://www.itwinjs.org) is a platform for infrastructure digital twins that provides techniques for creating digital models of real world assets such that the physical world and the digital world are *connected*. In iTwin.js, digital models are persisted in [iModels](https://www.itwinjs.org/learning/glossary/#imodel), according to the rules of [BIS](https://www.itwinjs.org/bis/). [Codes](https://www.itwinjs.org/bis/intro/codes/) play an integral role in forming the connection between the physical world and the digital world.
+[iTwin.js](https://www.itwinjs.org) is a platform for infrastructure digital twins that provides techniques for creating digital models of real world assets such that the physical world and the digital world are *connected*. In iTwin.js, digital models are persisted in [iModels](https://www.itwinjs.org/learning/glossary/#imodel), according to the rules of [BIS](https://www.itwinjs.org/bis/). [Codes](https://www.itwinjs.org/bis/guide/fundamentals/codes/) play an integral role in forming the connection between the physical world and the digital world.
```
pmconne, line 5: This is a very informative document. Can you link to it from the API documentation, e.g., in `CodeService.ts`? I could find no existing links to it.

pmconne (approved):

docs/learning/backend/CodeService.md
kabentley: done

core/backend/src/CloudSqlite.ts
```diff
@@ -602,9 +602,224 @@ export namespace CloudSqlite {
...
+    }
+
+    /**
+     * Initialize a cloud container to hold a Cloud SQliteDb. The container must first be created via its storage supplier api (e.g. Azure, or AWS).
+     * A valid sasToken that grants write access must be supplied. This function creates and uploads an empty database into the container.
+     * @note this deletes any existing content in the container.
+     */
+    protected static async _initializeDb(args: { dbType: typeof VersionedSqliteDb, props: ContainerAccessProps, dbName: string, blockSize?: number }) {
+      const container = createCloudContainer({ ...args.props, writeable: true });
+      container.initializeContainer({ blockSize: args.blockSize ?? 64 * 1024 });
```
nick4598, line 705: The comment on this function says 'the container must first be created via its storage supplier api', but initializeContainer will create the container if it doesn't exist I believe. Is that wording intentional to hopefully prevent people from using _initializeDb as a means of initializing the container?

core/backend/src/CloudSqlite.ts
```diff
@@ -570,12 +570,12 @@ export namespace CloudSqlite {
     cacheDir?: string;
   }
 
-  /** The collection of currently extant CloudCaches, by name. */
+  /** The collection of currently extant `CLoudCache`s, by name. */
```
nick4598, line 573: Should be 'CloudCache' instead of 'CLoudCache'

core/backend/src/BlobContainerService.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  /**
+   * Metadata about the use of a container so that:
+   *  - administrators can understand why a container exists for assigning RBAC permissions appropriately
+   *  - usage reports can aggregate types of containers
+   *  - applications can identify their containers
+   *  - applications can store important (to them) properties "on" their containers
+   */
+  export interface Metadata {
```
nick4598, line 57: Which of these properties would be used to serve the purpose of "usage reports can aggregate types of containers"?

core/backend/src/CloudSqlite.ts
kabentley: No, the comment is correct.  `initializeContainer` merely initializes a container for use by CloudSqlite, creating the manifest. It has no way to create a container, which must already exist and have an access token to write to it.

> Is that wording intentional to hopefully prevent people from using _initializeDb as a means of initializing the container

I don't  understand the question. This function *is* the way to initialize a container. It does mean that you can't use this function to store more than one database in a container though.

core/backend/src/BlobContainerService.ts
kabentley: `format` and `application`, for example.

core/backend/src/CloudSqlite.ts
nick4598: initializeContainer calls sqlite3_bcv_create which has many comments in it implying that the container will be created if it does not exist, such as 

" First just try to upload the manifest. If that fails for any reason,
  ** assume the container does not exist and attempt to create it. If that
  ** succeeds, attempt to upload the manifest again. "

kabentley: That can only work with the storage emulator, with the account key. Nobody will have sufficient rights to create containers except the `BlobContainerService` - that's its purpose for existing!

kabentley: BTW, we should remove all mentions of account key from the entire CloudSqlite api, even for storage emulators. Account keys should never be used in any real code, even tests,

nick4598: Currently the BlobContainerService/cloudsqlitedbservice does have the CONNECTION_STRING which contains the account key. This value is part of the process's env.

kabentley: OK, i think @wgoehrig  told me the plan was to eliminate that.

core/backend/src/BlobContainerService.ts
MichaelSwigerAtBentley: These fields line up with what we are tracking currently. In addition, we also have a human readable, non-unique "name" for the container. In the case of "type", type was originally CloudSQLite specific for controlling what would be stored inside the CloudSQLite database, that would be replaced by format + application in this case?

nick4598: I don't see how format + application replaces 'type' unless we infer the type from the application, but I believe that would require us to assume that each application only creates one type. 

How about something like:
export type = 'workspace' | 'codes' | 'propertystore' (These types may be wrong) 

This would also loosely prevent services from creating containers for information that we don't wish to store. They'd have to ask the platform/blobcontainerservice whether we could support a new type or whether their use case fits in with the blobcontainerservice.

core/backend/src/CloudSqlite.ts
kabentley: fixed

core/backend/src/BlobContainerService.ts
kabentley: Do we need both "name" and "description"? I suppose one can be optional. I don't think "name" is the right term though, since it isn't unique. 

Probably we need to think about which of these fields are defined by the application that creates them and which are meant to be free-form for users to recognize them for their own purposes. For example, there will be lots of containers for workspaces. They should have "format=workspace" (@nick4598 i think "format" and "type" are the same concept), so we know they're workspaces. Then, they'll probably have an application that creates them "application=substation". Then they'll have a "description=`templates for electrical components`". Maybe they also need something the user can supply, e.g. "notes=`approved for all Acme projects`". 

@MichaelSwigerAtBentley, don't we also need something like "RBACRole", so the RBAC control panel will know how to associate permissions with a given container. E.g. "rbacRole=project-workspace". Admittedly, i don't know much about your RBAC plan.

nick4598: When I read the comment above format: "a machine-readable string that describes the "format" of the data in this container (e.g. "CloudSqlite")", I interpret that to mean CloudSqlite would be one of the potential options for this property. That would be at odds with a format of workspace, right? Since I thought workspaces would be accessed or stored in CloudSqlite form, i.e. manifest and blocks?

kabentley: maybe i should change the comment to use "workspace" as the example. I think of this string playing the same role as "extension" in filenames.

PR #5460, opened 2023-04-28
get vscode-eslint extension working again

aruniverse:
after updating pnpm ver, and not hoisting any package with eslit in the name, update where plugins are resolved from
its a little hacky, will not work for pkgs that are nested a level deeper, but should only affect domain pkgs, which should be ok for now.
we will be providing a proper fix using eslint flat configs and the rushstack/eslint-plugin in a separate pr


.vscode/cSpell.json
```diff
@@ -1,420 +1,436 @@
...
-        "xyzxyzxyz",
-        "zbuffer"
-    ],
-    "flagWords": [
-        "hte",
-        "minium"
-    ]
+  "version": "0.2",
+  "language": "en",
+  "enableFiletypes": [
```
aruniverse, line 4: just moved this out of the settings.json into here to keep related configs together

.vscode/settings.json
```diff
@@ -1,13 +1,37 @@
...
-    "**/bower_components": true,
-    "**/.nyc_output": true,
-    "**/.rush": true
+  "editor.tabSize": 2,
+  "eslint.enable": true,
+  "eslint.format.enable": true,
+  "eslint.lintTask.enable": true,
+  "eslint.options": {
+    "overrideConfigFile": "./package.json",
+    "resolvePluginsRelativeTo": "../../common/temp/node_modules/.pnpm/@itwin+eslint-plugin*/node_modules",
```
aruniverse, line 16: temporary solution until we adopt eslint flat configs & use the rushstack/eslint-plugin-patch

(basically need to reapply this [commit](https://github.com/iTwin/itwinjs-core/pull/5460/commits/a2b2e6c98d466d4cd8e7384f0f13218fa13f073d) later on)

fyi @Jake-Screen @paulius-valiunas

calebmshafer (approved):
LGTM.

@kabentley fyi, this should fix your VSCode lint issues

pmconne:
Is this going to increase the number of irrelevant formatting/whitespace changes cluttering up PRs?

kabentley:
> Is this going to increase the number of irrelevant formatting/whitespace changes cluttering up PRs?

This fixes eslint running inside VSCode (so you get red squiggles, etc. as you edit.) It broke with some change last week. As far as I know ESLint doesn't do any whitespace or formatting changes. It does "auto fix" errors such as missing semicolons when you save changes, but doesn't change anything that wouldn't fail a CI job.

aruniverse:
> Is this going to increase the number of irrelevant formatting/whitespace changes cluttering up PRs?

No, not necessarily. Most of these whitespace changes you see are probably due to devs using other formatters on their end, such as prettier. There is an eslint rule for spacing, and proper indentation, but thats always been there, and wouldve been flagged by the lint command. This is for "eslint intellisense" to work again inside the editor.

pmconne:
@kabentley I see "auto-format on save", that's why I ask. (I'm only asking - not blocking PR because of it).
I do wish we'd just [switch to prettier](https://prettier.io/docs/en/why-prettier.html) and never have to talk about code formatting again.

PR #5462, opened 2023-04-28
@bentley/imodeljs-native 4.0.6

nick4598:


nick4598:
Had to downgrade @types/node to 18.16.1 from 18.16.2, otherwise I got:

--[ FAILURE: @itwin/core-backend ]---------------------------[ 4.38 seconds ]--

src/rpc/multipart.ts(67,16): error TS2345: Argument of type 'HttpServerRequest' is not assignable to parameter of type 'IncomingMessage'.
  Type 'HttpServerRequest' is missing the following properties from type 'IncomingMessage': headersDistinct, trailersDistinct

common/config/rush/pnpm-lock.yaml
aruniverse, line 1: revert this to as is on master, and run just, `rush update`, instead of `rush update -full`. new itwin/eslint-plugin has a new lint rule which will break

aruniverse:
> Had to downgrade @types/node to 18.16.1 from 18.16.2, otherwise I got:
> 
> --[ FAILURE: @itwin/core-backend ]---------------------------[ 4.38 seconds ]--
> 
> src/rpc/multipart.ts(67,16): error TS2345: Argument of type 'HttpServerRequest' is not assignable to parameter of type 'IncomingMessage'. Type 'HttpServerRequest' is missing the following properties from type 'IncomingMessage': headersDistinct, trailersDistinct

fyi @swbsi

common/config/rush/pnpm-lock.yaml
pmconne: @aruniverse my scripts and [instructions](https://dev.azure.com/bentleycs/iModelTechnologies/_wiki/wikis/iModelTechnologies.wiki/28809/Integrating-a-new-imodel-native-addon?anchor=update-to-the-new-addon-package.#) delete the lockfile (per older instructions) - I should probably update them to not do that.

aruniverse: tbf, it should be ok to use `--full`, but often changes from downstream pkgs (respecting semver) can still "break" us.
afaik, native updates were the only time someone updated all transitive deps with "full", but its probably a good idea, to just isolate the integrate native addon prs to just pull in the addon only, and handle the full update in a separate pr.

image

I will update the docs, and add a note

nick4598: Does it look better now? I grabbed the original pnpm lock and ran rush update, but originally I never ran rush update -full.. or rush update. I only ran Paul's scripts which just run plain rush update in 'update-imodeljs-native-version.sh' from what I see.

aruniverse: yea they look good to me now

tcobbs-bentley (approved):

aruniverse (approved):

PR #5469, opened 2023-05-02
Log metadata when uploading tiles

paulius-valiunas:
This will help with optimizing the performance of tile analytics. Right now it has to query _all_ tiles for each tile tree to get their metadata because the blob storage API doesn't allow selecting only recent blobs. With this change the metadata can be extracted from Docker container logs.

### DIFF
--- a/common/changes/@itwin/core-backend/log-tile-metadata_2023-05-02-10-40.json
+++ b/common/changes/@itwin/core-backend/log-tile-metadata_2023-05-02-10-40.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "log metadata when uploading tiles",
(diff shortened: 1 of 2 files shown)


pmconne (commented):
`Metadata` is just a bunch of key-value pairs. Any potential it will contain something that should not be logged?

paulius-valiunas:
> `Metadata` is just a bunch of key-value pairs. Any potential it will contain something that should not be logged?

That metadata essentially becomes part of the tile in blob storage so anyone who has access to the tile probably has access to the metadata too. I don't think anyone should put anything sensitive there, but just to be safe I'll move this to `IModelTileRpcImpl`.

pmconne (approved):

paulius-valiunas:
https://github.com/Mergifyio backport release/3.7.x

PR #5474, opened 2023-05-02
Added new sandbox links

saeeedtorabi:
- Also increased Point2dArrayCarrier test coverage from 80% to 100%

### DIFF
--- a/core/geometry/src/geometry3d/Ray3d.ts
+++ b/core/geometry/src/geometry3d/Ray3d.ts
@@ -79,7 +79,7 @@ export class Ray3d implements BeJSONFunctions {
   /**
    * Return the dot product of the ray's direction vector with a vector from the ray origin
    * to the `spacePoint`.
-   * * Suppose ray shows the normal vector of a plane. Then this function returns the altitude
+   * * If the instance is the unit normal of a plane, then this method returns the (signed) altitude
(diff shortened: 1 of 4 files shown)


core/geometry/src/geometry3d/Ray3d.ts
```diff
@@ -76,7 +76,13 @@ export class Ray3d implements BeJSONFunctions {
   public isAlmostEqual(other: Ray3d): boolean {
     return this.origin.isAlmostEqual(other.origin) && this.direction.isAlmostEqual(other.direction);
   }
-  /** Return the dot product of the ray's direction vector with a vector from the ray origin to the `spacePoint`. */
+  /**
+   * Return the dot product of the ray's direction vector with a vector from the ray origin
+   * to the `spacePoint`.
+   * * Suppose ray shows the normal vector of a plane. Then this function returns the altitude
```
dassaf4, line 82:
```suggestion
   * * If the instance is the unit normal of a plane, then this method returns the (signed) altitude
```

dassaf4 (approved):

core/geometry/src/geometry3d/Ray3d.ts
saeeedtorabi: Addressed

PR #5475, opened 2023-05-03
[extension-api] remove `@alpha` `registerTool` api

aruniverse:
since UiProviders are an appui-react concept now, need to rethink how to register tools at ui-abstract level. 
until then remove the registerTool api


docs/learning/frontend/Extensions.md
```diff
@@ -109,55 +132,238 @@ The only requirement when creating an Extension is to define a default function.
...
 ```
-The above would print "Hello from Extension!". We can extend functionality by adding, for example, a tool to print to the console whenever it is clicked:
+
+The above would print "Hello from Extension!".
+
+We can extend functionality by adding, for example, adding some decorators representing IoT devices, and when clicked will navigate to a specific view.
+First, we will query for all ceilings, walls, windows, etc, and hide them from the view.
+Then, we will query for all IoT devices, and add a decorator for each one.
+
+```ts
```
pmconne, line 149: Having full-blown multi-file sample code in a markdown file is not ideal...could it be moved to, e.g., example-code so we can confirm it compiles and continues to do so in future?

pmconne (approved):

core/extension/index.js
```diff
@@ -1,14 +1,12 @@
...
- * See LICENSE.md in the project root for license terms and full copyright notice.
- *--------------------------------------------------------------------------------------------*/
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
 
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis
 const globalSymbol = Symbol.for("itwin.core.frontend.globals");
-const ext = globalThis[globalSymbol].getExtensionApi("import.meta.url");
+const ext = globalThis[globalSymbol].getExtensionApi(import.meta.url);
```
calebmshafer, line 8: does this work now with the change to the typings being "module"?

Also with that change to "module" is it required to use esmodules?

aruniverse: yes `import.meta.url` would only work if we deliver esmodules (setting the type to "module")
considering at the moment this pkg is supposed to help augment frontend applications, i think its safe to start pushing only delivering esm for this pkg.

calebmshafer (approved):

PR #5477, opened 2023-05-03
Web workers

pmconne:
Don't look at this yet.

There are several heavyweight operations performed on the client in the main JS thread that we'd like to offload onto web workers.
Constraints:
- Each worker must compile down to a single .js file that can be loaded from the same origin.
- The workers are used by core-frontend so must be delivered with that package or otherwise made accessible to it.
- Each worker should have as few dependencies as possible and tree-shaken, to minimize script size.
- Workers should definitely avoid doing dumb stuff like trying to access `IModelApp` or the DOM.
- Workers should be written in TypeScript.

There are a variety of ways to meet these constraints. Here, I am building the worker code together with the rest of core-frontend, then web-packing each worker's output into a single .js file and copying it to lib/public/scripts. This simplifies many things. However, it makes it rather too easy to accidentally do dumb stuff like accessing `IModelApp` indirectly (e.g., by calling some function that calls some other function that uses `IModelApp.renderSystem` or `IModelApp.tileAdmin`).

It is **not** the goal of this PR to **parallelize** operations like tile decoding. We're not trying to decode multiple tiles simultaneously in different workers, because that will increase memory usage. We're just trying to parallelize these operations relative to the **main** thread.


pmconne:
Split off into #5532; worker support to be introduced subsequently.

pmconne:
Superseded by #5548.

PR #5479, opened 2023-05-03
Mobile RPC queue up pending messages until reconnected in foreground

toddsouthenbentley:
This fixes a problem encountered when cancelling authentication in the Mobile SDK Android app. The native callback would happen before the RPC server had reconnected and the message would be lost.

Note: this was tested in the release/3.7.x branch as we're not able to run mobile on master yet.

### DIFF
--- a/core/mobile/src/backend/MobileRpcServer.ts
+++ b/core/mobile/src/backend/MobileRpcServer.ts
@@ -36,7 +36,7 @@ export class MobileRpcServer {
   private _port: number;
   private _connectionId: number;
   private _pingTimer: NodeJS.Timeout;
-  public constructor() {
+  public constructor(private _pendingMessages: Array<string | Uint8Array> | undefined = undefined) {
     /* _pingTime is a fix for ios/mobile case where when the app moves into foreground from
      * background backend restart ws.Server and then notify frontend to reconnect. But ws.Server
      * listening event is not fired as node yield to kevent and wait for some io event to happen.
@@ -81,6 +81,7 @@ export class MobileRpcServer {
       this._connection = connection;
       this._connection.on("message", (data) => this._onConnectionMessage(data));
       this._createSender();
+      this._sendPending();
       (global as any).__iTwinJsRpcReady = true;
     });
   }
@@ -100,6 +101,20 @@ export class MobileRpcServer {
 
     MobileRpcServer.interop.sendString = sender;
     MobileRpcServer.interop.sendBinary = sender;
+  }
+
+  private _sendPending() {
+    if (this._pendingMessages === undefined)
+      return;
+
+    for (const message of this._pendingMessages) {
+      this._connection!.send(message, (err) => {
+        if (err) {
+          throw err;
+        }
+      });
+    }
+    this._pendingMessages.length = 0;
   }
 
   private _onConnectionMessage(data: ws.Data) {
@@ -152,6 +167,15 @@ export function setupMobileRpc() {
      Thus, we install a temporary timer on suspend to prevent the loop from exiting prematurely.
   */
   let retainUvLoop: NodeJS.Timer | undefined;
+  const pendingMessages: Array<string | Uint8Array> = [];
+
+  function usePendingSender() {
+    const sender = (message: string | Uint8Array, _connectionId: number) => {
+        pendingMessages.push(message);
+    };
+    MobileRpcServer.interop.sendString = sender;
+    MobileRpcServer.interop.sendBinary = sender;
+  }
 
   MobileHost.onEnterBackground.addListener(() => {
     hasSuspended = true;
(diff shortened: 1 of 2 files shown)


core/mobile/src/backend/MobileRpcServer.ts
```diff
@@ -102,6 +103,20 @@ export class MobileRpcServer {
...
   }
 
+  private _sendPending() {
+    if (this._pendingMessages === undefined)
+      return;
+
+    for (const message of this._pendingMessages) {
+      this._connection!.send(message, (err) => {
+        if (err) {
+          throw err;
```
tcobbs-bentley, line 113: Isn't this going to cause problems? I'm not sure what a solution would be, but throwing the error is going to break the construction of this RPC server instead of simply propagating a message failure to wherever it would normally go.

core/mobile/src/backend/MobileRpcServer.ts
```diff
@@ -102,6 +103,20 @@ export class MobileRpcServer {
...
+        }
+      });
+    }
+    this._pendingMessages.length = 0;
```
tcobbs-bentley, line 117: I feel that this would be better:
```ts
delete this._pendingMessages
```

tcobbs-bentley: Does the original clear the outer pendingMessages? If so, then my suggestion above is obviously wrong.

core/mobile/src/backend/MobileRpcServer.ts
```diff
@@ -170,7 +195,7 @@ export function setupMobileRpc() {
       return;
     }
 
-    server = new MobileRpcServer();
+    server = new MobileRpcServer(pendingMessages);
```
tcobbs-bentley, line 198: A basic solution to the error throwing problem would be to do the following:

```ts
try {
  server = new MobileRpcServer(pendingMessages);
} catch (_error) {
  server = new MobileRpcServer();
}
```
That way, at least having an error happen while sending the pending messages wouldn't break everything.

core/mobile/src/backend/MobileRpcServer.ts
tcobbs-bentley: I just realized that the error will be thrown from the callback, which won't propagate up to the constructor. (Having said that, where _will_ it go?)

core/mobile/src/backend/MobileRpcServer.ts
tcobbs-bentley: Given that the error happens in a callback, it seems that it won't get thrown, so this change won't do anything. Sorry for the confusion.

core/mobile/src/backend/MobileRpcServer.ts
toddsouthenbentley: Why is that better? I just want to clear the array as these messages have been processed. The array itself gets reused the next time the app goes to the background.

core/mobile/src/backend/MobileRpcServer.ts
toddsouthenbentley: The constructor isn't blocked by processing the pending messages. That doesn't happen until the websocket is connected.

core/mobile/src/backend/MobileRpcServer.ts
toddsouthenbentley: See my response to the related comment.

core/mobile/src/backend/MobileRpcServer.ts
toddsouthenbentley: Sorry, my browser didn't seem to be updated when I responded. Yes, setting the length clears the outer pendingMessages array. The constructor just has a reference to that array.

core/mobile/src/backend/MobileRpcServer.ts
tcobbs-bentley: Sorry, I should have said that _onConnection wouldn't finish, not the constructor. But since the error happens in a callback, I suspect that the throw just goes into the ether. If that is the case, then simply ignoring the error would be better. If that's not the case, where does the throw end up?

toddsouthenbentley: I don't know where the error goes, I just followed the existing pattern in _createSender.

tcobbs-bentley: I feel that that pattern is wrong in both locations, but given that as a precedent, I'll withdraw my complaint.

pmconne:
Do you need this backported to 3.7? If so please comment `@Mergifyio backport release/3.7.x`.

toddsouthenbentley:

tcobbs-bentley (approved):

aruniverse (approved):

PR #5481, opened 2023-05-04
Add public methods to BriefcaseManager to query changesets

kabentley:
These public methods wrap the internal `BackendHubAccess` interface methods.


aruniverse:
fyi @austeja-bentley @robertas-kerpys 

the imodels-clients will need to react to the change in ChangesetProps (making size required)
image

kabentley:
> the imodels-clients will need to react to the change in ChangesetProps (making size required)

Do you know if this actually breaks their code? I presume it won't, and i hope they really do return the size. It was just a mistake that i had it optional previously.

aruniverse (approved):

PR #5482, opened 2023-05-04
[@itwin/core-common] Removing dependency on semver

Jake-Screen:
Removing semver dependency by implementing own way to compare versions. This is to help resolve an issue introduced when upgrading to semver@7.5.0 that caused failures when installing dependencies for the viewers. Removing our dependency on this should solve that issue.

### DIFF
--- a/core/common/src/RpcInterface.ts
+++ b/core/common/src/RpcInterface.ts
@@ -45,21 +45,8 @@ interface SemverType {
  */
 export abstract class RpcInterface {
 
-  private static findDiff(backend: SemverType, frontend: SemverType): string {
-    const isSame = backend.major === frontend.major && backend.minor === frontend.minor && backend.patch === frontend.patch && backend.prerelease === frontend.prerelease;
-    if (isSame) {
-      return "same";
-    }
-    if (backend.major !== frontend.major) {
-      return "major";
-    }
-    if (backend.minor !== frontend.minor) {
-      return "minor";
-    }
-    if (backend.patch !== frontend.patch) {
-      return "patch";
-    }
-    return "prerelease";
+  private static findDiff(backend: SemverType, frontend: SemverType) {
+    return backend.major !== frontend.major ? "major" : backend.minor !== frontend.minor ? "minor" : backend.patch !== frontend.patch ? "patch" : backend.prerelease !== frontend.prerelease ? "prerelease" : "same";
   }
 
   private static parseVer(version: string): SemverType {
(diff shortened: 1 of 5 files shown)


pmconne (commented):
Write tests for `isVersionCompatible`? Apparently there are none.

core/common/src/RpcInterface.ts
```diff
@@ -34,20 +32,57 @@ export interface RpcInterfaceDefinition<T extends RpcInterface = RpcInterface> {
...
+
 /** An RPC interface is a set of operations exposed by a service that a client can call, using configurable protocols,
  * in a platform-independent way. TheRpcInterface class is the base class for RPC interface definitions and implementations.
  * @public
  */
 export abstract class RpcInterface {
+
+  private static findDiff(backend: SemverType, frontend: SemverType): string {
+    const isSame = backend.major === frontend.major && backend.minor === frontend.minor && backend.patch === frontend.patch && backend.prerelease === frontend.prerelease;
+    if (isSame) {
```
kabentley, line 50: remove : string so this will return the correct type (the set of possible strings);

I don't see why you have the `isSame` test. Why not just:

```ts
private static findDiff(backend: SemverType, frontend: SemverType) {
return 
  backend.major !== frontend.major ? "major" : 
  backend.minor !== frontend.minor ? "minor" : 
  backend.patch !== frontend.patch ? "patch" : 
  backend.prerelease !== frontend.preprelase ? "prerelease" : 
  "same";
}
```

pmconne (approved):

core/common/src/RpcInterface.ts
```diff
@@ -34,20 +32,57 @@ export interface RpcInterfaceDefinition<T extends RpcInterface = RpcInterface> {
...
   public static isVersionCompatible(backend: string, frontend: string): boolean {
-    const difference = semver.diff(backend, frontend);
-    if (semver.prerelease(backend) || semver.prerelease(frontend)) {
-      return difference === null;
-    } else if (semver.major(backend) === 0 || semver.major(frontend) === 0) {
-      return difference === null || (difference === "patch" && semver.patch(frontend) < semver.patch(backend));
+    const backendSemver = this.parseVer(backend);
+    const frontendSemver = this.parseVer(frontend);
+    const difference = this.findDiff(backendSemver, frontendSemver);
+    if ((backendSemver.prerelease !== undefined || frontendSemver.prerelease !== undefined) || difference === "major") {
```
kabentley, line 80: comments explaining this logic would help

core/common/src/RpcInterface.ts
Jake-Screen: modified function to that

core/common/src/RpcInterface.ts
Jake-Screen: added comments explaining the logic

kabentley (approved):

PR #5489, opened 2023-05-08
Mitigate white patches in map tiles

aorzan:
Partially solve the issue of white patches, when parent tiles are loaded. When no tile from the tree is loaded, a white patch is still visible.


pmconne:
@aorzan thank you. Were there any differences? I cannot run Image Diffs on this machine.

aorzan:
Hi Paul,
There were 6 major differences, all in tiles that are far away and/or viewed at an angle. Attached is the result for the biggest difference.

DStradley:
> @DStradley can you please take a look at the image differences?

They all are fine (there are some map changes that happened yesterday and are identical here).

pmconne (approved):

PR #5500, opened 2023-05-10
Make IModeApp.hubAccess public

kabentley:


johnnyd710 (approved):

common/api/core-frontend.api.md
```diff
@@ -5240,7 +5240,6 @@ export class IModelApp {
     static getAccessToken(): Promise<AccessToken>;
     // @internal (undocumented)
     static get hasRenderSystem(): boolean;
-    // @internal
```
aruniverse, line 5243: Do we want to expose this or just have consumers use the `imodels-clients` which implement the FrontendHubAccess interface directly?

aruniverse: @johnnyd710 @kckst8

kabentley: Applications should use this interface rather than depending directly on one specific implementation (e.g. so it can be mocked). That's why it exists.

PR #5511, opened 2023-05-11
Publish latest core docs as artifact

ben-polinsky:
Publish generated core documentation as pipeline artifact to enable iTwin packages extracted from core to build aggregate docs. This allows us to prepare new documentation releases without having to trigger a build manually or wait for core to unnecessarily build.

Two new parameters can be passed to `docs-build.yaml`: 
- `useCoreDocsArtifact` will download the last published generated-docs instead of building from source. 
- `isExternalBuild` permits "external" builds to be tagged for potential release if also publishing


common/config/azure-pipelines/jobs/docs-build.yaml
```diff
@@ -14,6 +14,9 @@ parameters:
   - name: outputDir
     type: string
     default: $(Pipeline.Workspace)/tempDocsBuild
+  - name: useCoreDocsArtifact
```
aruniverse, line 17:
```suggestion
  - name: useCurrentCoreDocsArtifact
```

PR #5523, opened 2023-05-15
Create plain object version of PropertyRecord

mdastous-bentley:


mdastous-bentley:
@raplemie @grigasp Would like to have your feedback about creating this `PropertyRecordJSON` interface.  Why goal is to have a interface only data structure for `MapFeatureInfoRecord`.

ui/appui-abstract/src/appui-abstract/properties/Record.ts
```diff
@@ -107,6 +120,49 @@ export class PropertyRecord {
...
+      result.description = json.description;
+
+    if (json.isDisabled !== undefined)
+      result.isDisabled = json.isDisabled;
+
+    if (json.isMerged !== undefined)
+      result.isMerged = json.isMerged;
+
+    if (json.isReadonly !== undefined)
+      result.isReadonly = json.isReadonly;
```
raplemie, line 162: These could reuse the `assignMemberIfExists` method.

ui/appui-abstract/src/appui-abstract/properties/Record.ts
```diff
@@ -25,6 +25,19 @@ export interface LinkElementsInfo {
   matcher?: (displayValue: string) => Array<{ start: number, end: number }>;
 }
 
+/** JSON representation of a [[PropertyRecord]].
+ * @beta
+ */
+export interface PropertyRecordJSON {
```
raplemie, line 31: appui-abstract follows core with its interface which ends with "props", so this would be called `PropertyRecordProps`, `*JSON` is more in presentation where they have `React` included which have a specific meaning for the "props" term. (appui-abstract is expected to remain framework agnostic, so it should follow core naming conventions)

raplemie (commented):
I dont have a strong feeling about if this should be there or not, do you have a use case where serializing a PropertyRecord is needed ?

grigasp:
I don't mind the addition either, but I'd raise the question if you need everything that `PropertyRecord` stores. A specific use case should answer that.

mdastous-bentley:
Decided to follow @grigasp advice and create dedicate MapFeatureInfo interfaces and keep it independent from PropertyRecord.  Closing this PR...

PR #5528, opened 2023-05-16
Auto-refresh access tokens for CloudSqlite

kabentley:
By default, access tokens are refreshed every hour.


core/backend/src/CloudSqlite.ts
```diff
@@ -34,8 +78,8 @@ export namespace CloudSqlite {
 
   /** Properties of a CloudContainer. */
   export interface ContainerProps {
-    /** blob storage module */
-    storageType: "azure" | "google" | "aws";
+    /** blob storage provider */
+    storageType: "azure" | "google";
```
pmconne, line 82: Motivation for removal of AWS? Is it related to auto-refresh?

core/backend/src/CloudSqlite.ts
```diff
@@ -675,7 +720,7 @@ export namespace CloudSqlite {
       /** The name of the database within the container. */
       dbName: string;
     }) {
-      this._container = createCloudContainer({ ...args.props, writeable: true });
+      this._container = createCloudContainer({ writeable: true, ...args.props });
```
pmconne, line 723: Motivation for switching order of `writeable` and `...args.props`? The new way, if someone passes you an object with a `writeable` property, it will overwrite your `true`.

core/backend/src/CloudSqlite.ts
kabentley: it is not really supported yet. When it is, @wgoehrig  wants to call it "s3".

core/backend/src/CloudSqlite.ts
kabentley: Yes, that's what i wanted. If you don't supply it, the default is `true`. But you can now open a "readonly" container using the `DbAccess` class by passing `writeable: false` (though all writeable methods will fail.)

nick4598 (approved):

nick4598 (approved):

core/backend/src/BlobContainerService.ts
```diff
@@ -45,6 +45,8 @@ export namespace BlobContainer {
     iTwinId: Id64String;
     /** optionally, an iModelId within the iTwin. If present, container is deleted when the iModel is deleted. */
     iModelId?: Id64String;
+    /** optionally, an owner of the container. Owners always have administrator rights for the container. */
+    owner?: string;
```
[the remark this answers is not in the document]
kabentley: this is for the BlobContainer service. I added this so @MichaelSwigerAtBentley will add support for it.

pmconne (approved):

Issue #5247, opened 2023-03-20
Reassemble vertex tables in a web worker


pmconne:
**Is your feature request related to a problem? Please describe.**
As described in #5030, when schedule scripts are applied on the front-end, we disassemble the meshes within each tile and reassemble them into nodes matching those defined by the script. This process can be time-consuming and blocks the main JavaScript thread, resulting in poor UX.

**Describe the solution you'd like**
Move this logic to a web worker.

PR #5532, opened 2023-05-16
iMdl parser (prep for web worker)

pmconne:
Part of #5247. Split off from #5477.

The goal is to move all of the parsing and processing of iMdl tile content into a worker, to unblock the main JavaScript thread. However, we cannot allocate the WebGL resources like graphics and textures in a worker. So, `ImdlReader` has been split into a parser and a graphics creator. Because passing data between workers and/or the main JS thread is always done through [structured cloning](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm), this required defining an intermediate representation of the graphics - no classes, functions, etc. Most of that representation is just pointers into the tile's array buffer, which can be transferred to the worker and back with no copying.

Originally, I intended to introduce workers first in #5477, using the seemingly simple case of upsampling terrain tiles in a worker - but changing the upsample function to `async` turned out to be surprisingly complicated. So instead, this initial PR gets iMdl parsing ready to move to a worker, with a subsequent PR to actually move it to the worker.

Incidental changes:
- certa.json pulls public assets from lib, not src, so that generated assets like workers will be found.
- Don't traverse the tile hierarchy every time an `UpsampledMapTile`'s "loadable tile" is requested.


pmconne:
ImageTests reports 13 major differences, presumably regressions.

MarcNeely (approved):

aruniverse (approved):

DStradley (approved):

PR #5534, opened 2023-05-17
Presentation: Accept `Buffer` responses from addon

grigasp:
imodel-native PR: https://github.com/iTwin/imodel-native/pull/283
imodel-native: https://github.com/iTwin/imodel-native/tree/presentation/avoid-native-string-copy-to-napi-string

### DIFF
--- a/presentation/backend/src/presentation-backend/NativePlatform.ts
+++ b/presentation/backend/src/presentation-backend/NativePlatform.ts
@@ -153,8 +153,7 @@ export const createDefaultNativePlatform = (props: DefaultNativePlatformProps): 
       return this.createSuccessResponse(response);
     }
     private handleConvertedResult<TSource, TTarget>(response: IModelJsNative.ECPresentationManagerResponse<TSource>, conv: (s: TSource) => TTarget): NativePlatformResponse<TTarget> {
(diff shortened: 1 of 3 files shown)


saskliutas (approved):

presentation/backend/src/presentation-backend/NativePlatform.ts
```diff
@@ -194,7 +198,7 @@ export const createDefaultNativePlatform = (props: DefaultNativePlatformProps):
       const response = this._nativeAddon.handleRequest(db, options);
       cancelEvent?.addOnce(() => response.cancel());
       const result = await response.result;
-      return this.handleResult<string>(result);
+      return this.handleConvertedResult<Buffer, string>(result, (buffer) => buffer.toString());
```
pmconne, line 201: If some app upgrades their backend (or we upgrade GPB that they use) without updating their frontend, won't they start receiving `Buffer`s from presentation requests that they won't know what to do with?

grigasp: Our RPC didn't change - we send a string as before. It's just that the addon doesn't do the native string to `Napi::String` copy anymore.

PR #5537, opened 2023-05-17
Revert breaking AppUI removals

GerardasB:
Revert AppUI removals that are breaking already released AppUI 4.0.0 when running with CRA.


raplemie (approved):
LGTM

PR #5544, opened 2023-05-18
Fix finalize-release workflow

DanRod1999:
fix issue with old workflow. Need to test if this works in ubuntu vm


aruniverse:
is this pr still needed / relevant since we now have https://github.com/iTwin/itwinjs-core/pull/5746?

DanRod1999:
> is this pr still needed / relevant since we now have #5746?

No, closing now, and still working on revising that PR.

PR #5548, opened 2023-05-19
Decode iMdl tiles in worker

### LINKED ISSUE #5247: Reassemble vertex tables in a web worker

pmconne:
Replaces #5477.
Fixes #5247. For now this is primarily an improvement when applying schedule animation script to the tiles on the front-end. But we expect to add some other processing (like generating edges) that should be done off the main thread too.

The worker script delivered with @itwin/core-frontend is 900kb uncompressed. (For comparison, @itwin/core-frontend is 31.8mb uncompressed, and @itwin/geometry-core is 25mb uncompressed).

A looong time ago, Barry wrote a `WebWorkerManager` to manage a pool of workers for a given task. It was never used, and it wasn't suitable here where the goal is **not** to run multiple simultaneous tasks, but rather to move individual tasks off the main thread. Its API was also not great - requiring lots of boilerplate (subclasses for every little thing) and providing no type-safety (`any` everywhere). I deleted it.

I added a `WorkerProxy` concept that takes an interface defining the methods to be provided by the worker and splits it into two: the implementation on the worker thread, and a proxy object that can invoke those methods asynchronously from the main thread. Both are type-safe.

There were various ways I could have organized the worker code. The most straightforward is to include it in core/frontend. The main drawback is that doing so makes it easy to accidentally import+use (perhaps indirectly) APIs that shouldn't (or mustn't!) be used in a worker; and to accidentally import tons of code for transitive dependencies that, despite being unused, can spectacularly bloat the size of the worker script (ex: tree-shaking applies to whole classes, not individual methods thereof). To reduce these risks I introduced the following directory structure:
```
core/frontend/
  common/ # APIs usable in workers and main thread; must not import from outside common/
  workers/ # Must import only from common/
  "everything else" # Code that belongs only on the main thread.
```
It would be "cleaner" if "everything else" was in a `main/` subdirectory, but I don't want to touch every single file in this PR. We *should* consider adding a lint rule that will enforce the constraints described above. Note, I only moved code to common/ that I actually needed for the iMdl parser worker. When we implement additional workers, more code will presumably move there.

Tile decoding is handled by an `ImdlDecoder`. There is one unique decoder object per unique schedule script being employed by any tile tree at any given time (including one decoder for all trees that have no schedule script). Each `IModelTileTree` acquires its decoder in its constructor and releases it in its `dispose` method. A decoder makes use of an `ImdlParser` which interfaces with the worker. Parsers (and therefore decoders) are reference-counted. Yes, that seems weird in a garbage-collected language, but alternatives involving `WeakRef` and `WeakMap` were evaluated and found even more awkward. We have one parser per script because the scripts can be huge (10s of MB) and we don't want to have to pass that from the main thread to the worker every time we decode a tile; also, we cache some information on the script for reuse on subsequent decodings, which would be lost when transferring it to the worker or changing the worker's current script.

Files with actual functional changes (all in core/frontend) - everything else is just moving existing code around and adjusting `import` statements accordingly:
```
package.json
src/
  common/
    WorkerProxy.ts
    imdl/ParseImdlDocument.ts
  test/
    tile/ImdlParser.test.ts
    worker/*
  tile/
    ImdlParser.ts
    ImdlDecoder.ts
    ImdlReader.ts
    IModelTile.ts
    IModelTileTree.ts
  workers/*
```

TODO:
- [x] Evaluate ImageTests results.
  - 1 difference in the Zoom3 view that was flaky in previous PR too.
- [x] Enable worker for batched tiles.
- [x] Fix electron full stack tests failing to locate worker script.
- [x] Fix documentation errors
  -  Topic RegisterWorker does not have a @docs-group-description in core-frontend.ts. Please add a description.
  - Topic Worker does not have a @docs-group-description in core-frontend.ts. Please add a description.


pmconne:
Failing core-full-stack-tests:
```
  5 failing

  1) IModelTileRequestChannels
       CloudStorageCacheChannel
         falls back to RPC if content is not found:

      AssertionError: expected 4 to equal 3
      + expected - actual

      -4
      +3

      at Context.<anonymous> (home/paul/c/j2/s/itwinjs/full-stack-tests/core/src/frontend/standalone/tile/TileRequest.test.ts:109:34)

  2) IModelTileRequestChannels
       CloudStorageCacheChannel
         is not used again after cache miss:

      AssertionError: expected 4 to equal 3
      + expected - actual

      -4
      +3

      at Context.<anonymous> (home/paul/c/j2/s/itwinjs/full-stack-tests/core/src/frontend/standalone/tile/TileRequest.test.ts:132:34)

  3) IModelTileRequestChannels
       Metadata cache channel
         is used first:

      AssertionError: expected 4 to equal 3
      + expected - actual

      -4
      +3

      at Context.<anonymous> (home/paul/c/j2/s/itwinjs/full-stack-tests/core/src/frontend/standalone/tile/TileRequest.test.ts:186:34)

  4) IModelTileRequestChannels
       Metadata cache channel
         caches metadata from RPC:
     AssertionError: expected undefined not to be undefined
      at Context.<anonymous> (home/paul/c/j2/s/itwinjs/full-stack-tests/core/src/frontend/standalone/tile/TileRequest.test.ts:247:32)

  5) IModelTileRequestChannels
       Metadata cache channel
         is not used again after cache miss:

      AssertionError: expected 4 to equal 3
      + expected - actual

      -4
      +3

      at Context.<anonymous> (home/paul/c/j2/s/itwinjs/full-stack-tests/core/src/frontend/standalone/tile/TileRequest.test.ts:289:34)
```

pmconne:
<summary>Hanging core full stack tests, electron only.
<details>
```

  1) IModelTileRequestChannels
       CloudStorageCacheChannel
         is used first:
     Error: Timeout of 240000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.


  2) IModelTileRequestChannels
       CloudStorageCacheChannel
         falls back to RPC if content is not found:
     Error: Timeout of 240000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.


  3) IModelTileRequestChannels
       CloudStorageCacheChannel
         is not used again after cache miss:
     Error: Timeout of 240000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.


  4) IModelTileRequestChannels
       Metadata cache channel
         is used first:
     Error: Timeout of 240000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.


  5) IModelTileRequestChannels
       Metadata cache channel
         caches metadata from RPC:
     Error: Timeout of 240000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.


  6) IModelTileRequestChannels
       Metadata cache channel
         is not used again after cache miss:
     Error: Timeout of 240000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```
</details>
</summary>

pmconne:
@wgoehrig do you understand why electron cannot access assets when running electron tests? This is not a new problem - [hypermodeling test](https://github.com/iTwin/itwinjs-core/blob/ae7e969dc643c2b17bd6c6a1310ca86d179e8fbb/full-stack-tests/core/src/frontend/hub/HyperModeling.test.ts#L207) is skipped in electron for the same reason.

pmconne:
> I see a reference to `parse-imdl-worker.js` but can't quite find that file

It is generated from the code in workers/ImdlParser/.
JS filename defined in [webpack config](https://github.com/iTwin/itwinjs-core/blob/b3479052441a398aefe1a6bc880d7fdb56119a78/core/frontend/src/workers/ImdlParser/webpack.config.js#L19).

> I'm probably overlooking something in the large diff.

Ignore most of the changed files; the ones requiring review are listed in the PR description.

markschlosseratbentley (approved):

core/frontend/src/test/tile/ImdlParser.test.ts
```diff
@@ -0,0 +1,276 @@
...
+    const parser = acquireImdlParser({});
+    await expect(parser.parse({
+      data: new Uint8Array(12),
+      batchModelId: "0x123",
+      is3d: true,
+      maxVertexTableSize: 4096,
+    })).to.be.eventually.rejectedWith(RangeError);
+  });
+
+  it("transfers the array buffer to the worker and back", async () => {
```
DStradley, line 153: What does "and back" mean here, since data gets moved and you are checking that its length is 0?

pmconne, line 153:
```suggestion
  it("transfers the array buffer to the worker", async () => {
```

pmconne: It's meaningless. Removed.

DStradley (approved):

wgoehrig (approved):

PR #5553, opened 2023-05-22
4.0.0 changelogs

DanRod1999:
update master changelogs from 4.0.0 release


docs/changehistory/NextVersion.md
```diff
@@ -2,3 +2,7 @@
 publish: false
 ---
 # NextVersion
+<<<<<<< HEAD
```
aruniverse, line 5: conflict

docs/changehistory/index.md
```diff
@@ -1,115 +1,349 @@
...
+const schema = await context.getSchema(schemaKey, SchemaMatchType.Exact);
+// Await the schema promise created by getSchemaInfo or return undefined if not already started
+const schema2 = await context.getCachedSchema(schemaKey, SchemaMatchType.Exact);
 ```
 
-[PresentationTreeRenderer]($presentation-components) renders nodes with action buttons for applying and clearing filters. Some hierarchy levels might not be filterable depending on the presentation rules used to build them. In that case, action buttons for those hierarchy levels are not rendered. If applied filter does not produce any nodes, `There are no child nodes matching current filter` message is rendered in that hierarchy level.
+### Other minor API changes
 
+<<<<<<< HEAD
 ![Filtered Tree](./assets/filtered-tree.jpg)
```
aruniverse, line 347: conflict

aruniverse (approved):

PR #5565, opened 2023-05-25
Improved ArcGIS map-layers validation and removed URL parameters from localization strings.

mdastous-bentley:
A common error is to to mixed 'ArcGIS' and 'ArcGISFeature' format type.  So now, if we detect URL is pointing to a ArcGIS server but the wrong service type has been specified, we give some hints to the end-user.

Also, historically we were adding map-layer's URL in error messages, but it ended up encoded:
image

By default, i18n encodes any special characters to mitigate XSS attacks, [documented here](https://www.i18next.com/translation-function/interpolation).  Even though there are ways to disabled encoding, I believe this would be a bad idea since URLs in our case are inputs from the end-user, and this could be used to inject some html in the hosting application.  So I decided to remove URL parameters from any locale strings, which was not that useful anyway.

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -7498,14 +7498,13 @@ export class MapLayerSources {
 
 // @public
 export enum MapLayerSourceStatus {
-    // @beta
-    IncompatibleFormat = 3,
-    InvalidCoordinateSystem = 7,
+    IncompatibleFormat = 7,
+    InvalidCoordinateSystem = 6,
     InvalidCredentials = 1,
     InvalidFormat = 2,
-    InvalidTileTree = 4,
-    InvalidUrl = 5,
-    RequireAuth = 6,
+    InvalidTileTree = 3,
+    InvalidUrl = 4,
+    RequireAuth = 5,
     Valid = 0
 }
 
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -1277,6 +1277,7 @@ export class ArcGisUtilities {
     };
     static isEpsg3857Compatible(tileInfo: any): boolean;
     static validateSource(url: string, formatId: string, capabilitiesFilter: string[], userName?: string, password?: string, ignoreCache?: boolean): Promise<MapLayerSourceValidation>;
+    static validateUrl(url: string, serviceType: string): MapLayerSourceStatus;
 }
 
 // @internal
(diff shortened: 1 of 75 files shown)


pmconne (commented):
Is test failure surprising?
```
--[ FAILURE: @itwin/map-layers-formats ]---------------------[ 6.78 seconds ]--


  68 passing (534ms)
  1 failing

  1) ArcGisFeaturePBF
       should validateSource:

      AssertionError: expected false to be true
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (src/test/ArcGisFeature/ArcGisFeatureFeatureFormats.test.ts:39:48)
```

mdastous-bentley:
> Is test failure surprising?
> 
> ```
> --[ FAILURE: @itwin/map-layers-formats ]---------------------[ 6.78 seconds ]--
> 
> 
>   68 passing (534ms)
>   1 failing
> 
>   1) ArcGisFeaturePBF
>        should validateSource:
> 
>       AssertionError: expected false to be true
>       + expected - actual
> 
>       -false
>       +true
>       
>       at Context.<anonymous> (src/test/ArcGisFeature/ArcGisFeatureFeatureFormats.test.ts:39:48)
> ```

I missed it, the url used in our test was not close enough of a real ArcGIS service url.

common/api/core-frontend.api.md
```diff
@@ -7497,12 +7498,14 @@ export class MapLayerSources {
...
-    InvalidCoordinateSystem = 6,
+    // @beta
+    IncompatibleFormat = 3,
+    InvalidCoordinateSystem = 7,
     InvalidCredentials = 1,
     InvalidFormat = 2,
-    InvalidTileTree = 3,
-    InvalidUrl = 4,
-    RequireAuth = 5,
+    InvalidTileTree = 4,
```
pmconne, line 7506: Changing the numeric value of an enum member is a breaking API change, which is not ok for `@public` APIs.

core/frontend/src/tile/map/MapLayerSources.ts
```diff
@@ -26,6 +26,10 @@ export enum MapLayerSourceStatus {
   InvalidCredentials,
   /** Provided format id could not be resolved in [[MapLayerFormatRegistry]] */
   InvalidFormat,
+  /** Format is not compatible with the URL provided.
+   * @beta
```
pmconne, line 30: What's the motivation for marking this `@beta`?

mdastous-bentley: No real motivation other than have the possibility to drop it if we change our mind.

pmconne: Do you think that's likely? If so resolve this comment.

pmconne (approved):
Let me know if macOS tests flake out again.

aruniverse (approved):

PR #5567, opened 2023-05-25
Upgrading to sinon 15.0.4

Jake-Screen:
Upgrading core repo to sinon 15.0.4 to bring it in line with presentation


aruniverse:
publishing pr to repro test failures in ci

mdastous-bentley:
@aruniverse I have a [draft PR](https://github.com/iTwin/itwinjs-core/pull/5582) for the map-layers-formats leaks, I guess we no longer need it ?

aruniverse:
> @aruniverse I have a [draft PR](https://github.com/iTwin/itwinjs-core/pull/5582) for the map-layers-formats leaks, I guess we no longer need it ?

if your pr fixes a leak and can merge before this one, why not keep it and get it through?

ben-polinsky (approved):

full-stack-tests/presentation/src/backend/DefaultSupplementalRules.test.ts
```diff
@@ -1,7 +1,7 @@
 /*---------------------------------------------------------------------------------------------
- * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
- * See LICENSE.md in the project root for license terms and full copyright notice.
- *--------------------------------------------------------------------------------------------*/
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
```
aruniverse, line 2: revert file change here

core/backend/package.json
```diff
@@ -84,7 +84,7 @@
     "null-loader": "^4.0.1",
     "nyc": "^15.1.0",
     "rimraf": "^3.0.2",
-    "sinon": "^9.0.2",
+    "sinon": "^15.0.4",
```
aruniverse, line 87: @types/sinon doesnt need to be updated??

anmolshres98: I don't think it is necessary for our usecase, the latest they have is 10 for types. What's a good way to determine if update to types is needed?

aruniverse: Typically when you update a pkg (especially for new major), and they doesnt deliver their own type, and there is a corresponding `@types/` pkg, youll want to update those as well. Typically there should be a new major types ver for every major of the pkg.

Id recommend going through the changelog for sinon to understand the changes: https://github.com/sinonjs/sinon/blob/main/CHANGES.md

Also take a look at the DefinitelyTyped repo to understand how this works
Based off https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sinon/index.d.ts#L1, this ws meant for Sinon 10, but i guess their api hasnt changed much since so it should be safe to use that

Take a look at presentation thats using sinon@15 & types/sinon@10 https://github.com/iTwin/presentation/blob/master/packages/components/package.json

anmolshres98: Sounds good, thanks for the tips!

test-apps/display-test-app/src/frontend/App.ts
```diff
@@ -361,7 +361,7 @@ export class DisplayTestApp {
     await FrontendDevTools.initialize();
     await HyperModeling.initialize();
     await EditTools.initialize();
-    await MapLayersFormats.initialize();
+    MapLayersFormats.initialize();
```
aruniverse, line 364: why was this changed?

anmolshres98: linting error when I ran `rush lint` locally: 
image

anmolshres98: let me know if I should revert and push those 2 lint related changes I made

aruniverse: the lint error is completely irrelevant to the change you made here

aruniverse: i couldve sworn @mdastous-bentley made this async

anmolshres98: It does look to be async, not sure why got lint error. Let me try it again
image

saskliutas:
> Reverting the `@types/sinon` update since this typescript check is failing even though it is executing correctly and tests are passing: image

This was a valid error as previously sinon was inferring `sinon.fake((id: string) => id)` type as `sinon.SinonSpy<any[], any>`. After `@types/sinon` was bumped it started to correctly infer type as `sinon.SinonSpy<[id: string], string>`. 

I fixed that type and bumped `@types/sinon` back up as it looks like you reverted it back only because of this one test.

anmolshres98:
> > Reverting the `@types/sinon` update since this typescript check is failing even though it is executing correctly and tests are passing: image
> 
> This was a valid error as previously sinon was inferring `sinon.fake((id: string) => id)` type as `sinon.SinonSpy<any[], any>`. After `@types/sinon` was bumped it started to correctly infer type as `sinon.SinonSpy<[id: string], string>`.
> 
> I fixed that type and bumped `@types/sinon` back up as it looks like you reverted it back only because of this one test.

Awesome, thanks for pushing the fix!

aruniverse (approved):

PR #5577, opened 2023-05-30
Make minRequests/maxRequests for a prefetch configurable during an attach

nick4598:


nick4598:
@kabentley Did you get a chance to look into making the time interval for checking with the daemon about prefetching (to determine if we should go to minRequests or maxRequests) configurable? I know you said there may be some issues with that.

kabentley:
> @kabentley Did you get a chance to look into making the time interval for checking with the daemon about prefetching (to determine if we should go to minRequests or maxRequests) configurable? I know you said there may be some issues with that.

It already is configurable from TypeScript via the "timeout" parameter. I see that for some reason i didn't add "maxRequests"  to  the interface in TS, though it is supported internally.

kabentley:
> You mentioned that there were some issues with the timeout being either too high or too low(don't remember which specifically).. do you remember any details around that?

see [this comment](https://github.com/iTwin/imodel-native/blob/cf229d25fe2fde080c98549205ad981411d08241/iModelCore/BeSQLite/SQLite/blockcachevfs.h#L639-L660). I don't think changing it is likely to be a good idea.

core/backend/src/CheckpointManager.ts
```diff
@@ -209,14 +209,17 @@ export class V2CheckpointManager {
...
-        const logPrefetch = async (prefetch: CloudSqlite.CloudPrefetch) => {
+        const logPrefetch = async (prefetch: CloudSqlite.CloudPrefetch, minRequests: number, maxRequests: number, timeout: number) => {
           const stopwatch = new StopWatch(`[${container.containerId}/${dbName}]`, true);
-          Logger.logInfo(loggerCategory, `Starting prefetch of ${stopwatch.description}`);
+          Logger.logInfo(loggerCategory, `Starting prefetch of ${stopwatch.description}`, {minRequests, maxRequests, timeout});
           const done = await prefetch.promise;
-          Logger.logInfo(loggerCategory, `Prefetch of ${stopwatch.description} complete=${done} (${stopwatch.elapsedSeconds} seconds)`);
+          Logger.logInfo(loggerCategory, `Prefetch of ${stopwatch.description} complete=${done} (${stopwatch.elapsedSeconds} seconds)`, {minRequests, maxRequests, timeout});
         };
+        const minRequests = IModelHost.appWorkspace.settings.getNumber("Checkpoints/prefetch/minRequests", 3);
```
kabentley, line 218:
```ts
          const getPrefetchConfig = (name: string, defaultVal: number) => IModelHost.appWorkspace.settings.getNumber(`Checkpoints/prefetch/${name}`, defaultVal);
          const minRequests = getPrefetchConfig("minRequests", 3);
          const maxRequests = getPrefetchConfig("maxRequests", 6);
          const timeout = getPrefetchConfig("timeout", 100);
```

nick4598:
> > You mentioned that there were some issues with the timeout being either too high or too low(don't remember which specifically).. do you remember any details around that?
> 
> see [this comment](https://github.com/iTwin/imodel-native/blob/cf229d25fe2fde080c98549205ad981411d08241/iModelCore/BeSQLite/SQLite/blockcachevfs.h#L639-L660). I don't think changing very much it is likely to be a good idea. In particular, setting it to a larger value violates Dan's "timely" comment.

I'm not sure that increasing it does violate Dan's timely comment. Won't our applications always be calling sqlite3_bcvfs_prefetch_run in a timely manner since you call it in a while (true) loop? The only time the timeout we provide comes into play is when m_nOutstanding == 0. m_nOutstanding is "The total number of outstanding requests managed by the
prefetch object." I don't really know when this would be 0 (maybe when minRequests is set to 0 or 1?), since calling sqlite3_bcvfs_prefetch_run "issues http requests for blocks of the database file that are not currently in the local cache until either (a) there are in total nRequest such requests outstanding, or (b) there are requests outstanding for all blocks not currently in the cache. 

If our minRequests is set to 3 for example, I think sqlite3_bcvfs_prefetch_run returns after one of those completes and then we call sqlite3_bcvfs_prefetch_run again to issue another request (potentially up to 6 if there are no on demand requests after one of those prefetches completes). 

It feels like to me we actually need two timeouts. I don't know if I really want sqlite3_bcvfs_prefetch_run to return after 100ms if our block downloads are taking longer than that (I'd prefer to return after one of the blocks has finished). But if our m_nOutstanding is 0, then we probably want a shorter sleep time(BeDuration::FromMilliseconds(m_timeout).Sleep() call). Especially if our minRequests is 1, because we'd always hit that m_nOutstanding is 0 case and take 100 ms per block that we fire off if I'm understanding correctly. 

If our minRequests is 0, then we probably don't want to be too aggressive with our sleeps though, because then that means we are just waiting for nDemand to reach 0 and we'll often be pestering the daemon until there are no demands.

core/backend/src/CheckpointManager.ts
```diff
@@ -209,14 +209,19 @@ export class V2CheckpointManager {
         container.connect(this.cloudCache);
       container.checkForChanges();
       if (IModelHost.appWorkspace.settings.getBoolean("Checkpoints/prefetch", false)) {
-        const logPrefetch = async (prefetch: CloudSqlite.CloudPrefetch) => {
+        // eslint-disable-next-line @typescript-eslint/no-shadow
```
nick4598, line 212: I was getting a no-shadow lint error. I'm assuming its safe to disable that here? I don't see how the outer scope would affect the inner scope of minRequests, maxRequests, timeout.

core/backend/src/CheckpointManager.ts
nick4598: changed it

kabentley (approved):

core/backend/src/CheckpointManager.ts
nick4598: Thanks

nick4598:

PR #5582, opened 2023-05-31
Fix leaks in map-layers-formats tests

mdastous-bentley:
@Jake-Screen was finally able to reproduce locally and we got:
image

So I made sure the tests never reach 'fetch'


pmconne (approved):

mdastous-bentley:
@aruniverse please review

aruniverse (approved):

PR #5590, opened 2023-06-02
correctly set `previous` dist tag on patches

aruniverse:
if the pkg@localVer doesnt exist yet, then it hasnt been published (obviously), so determine the dist tag


aruniverse:

aruniverse:

common/scripts/gather_packages.py
```diff
@@ -121,27 +121,30 @@ def determineDistTag(branchName, currentVer, latestVer, previousVer):
...
     print ("The package does not yet exist.  Copying " + packageName + " to staging area.")
     shutil.copy(artifact, stagingDir)
-    continue
 
   if 0 != len(serverVer):
     print ("The version already exists.  Skipping...")
     continue
 
   packagesToPublish = True
   print ("Local version is newer than on the server.  Copying package " + packageName + " to staging area.")
```
nick4598, line 130: Can we pick one of the two messages? We'll print this one and this print ("The package does not yet exist.  Copying " + packageName + " to staging area.") now, right?

common/scripts/gather_packages.py
```diff
@@ -121,27 +121,30 @@ def determineDistTag(branchName, currentVer, latestVer, previousVer):
...
+  if not os.path.exists(os.path.join(stagingDir, baseName)):
+    shutil.copy(artifact, stagingDir)
 
-  command = "npm view " + packageName + " dist-tags.previous"
+  command = "npm dist-tag ls " + packageName
   proc = subprocess.Popen(command, stdin = subprocess.PIPE, stdout = subprocess.PIPE, shell=True)
-  previousVer = proc.communicate()[0]
-  if len(previousVer) == 0:
-    print("No version found for dist-tag 'previous'")
+  distTags = proc.communicate()[0]
```
nick4598, line 136: We're also figuring out distTags for each package, but we only need to set these values once right?

Should we just do the first package we find that is copied to the staging area? Not sure how much it impacts the runtime

Jake-Screen: added check to see if versions were already set

common/scripts/gather_packages.py
Jake-Screen: chose to use the "local version is newer" message

common/scripts/gather_packages.py
Jake-Screen: now i have

nick4598 (approved):

calebmshafer (approved):

PR #5591, opened 2023-06-03
Two map-layers instances having different formatId or subLayers set should not share the same tile tree.

mdastous-bentley:
Two map layers instance having two different set of sub-layers, such as:
```
{ formatId: "ArcGIS", url: "https://server.com", name: "Layer1", subLayers: [{name: "sub_layer1", visible: true}]};
{ formatId: "ArcGIS", url: "https://server.com", name: "Layer1", subLayers: [{name: "sub_layer2", visible: true}]};
```
should not share the same tile tree since it define different datasets.

Also I think the same apply for two map-layers having two differents formatId, but same URL, although unlikely:
```
{ formatId: "ArcGIS", url: "https://server.com", name: "Layer1",};
{ formatId: "WMS", url: "https://server.com", name: "Layer2"};
```

### DIFF
--- a/common/changes/@itwin/core-frontend/geo-fixImageryTileTreeSupplier_2023-06-03-01-29.json
+++ b/common/changes/@itwin/core-frontend/geo-fixImageryTileTreeSupplier_2023-06-03-01-29.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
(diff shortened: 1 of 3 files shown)


pmconne (approved):

mdastous-bentley:
@pmconne I would like to consume this fix in the next 4.0.x build of core-frontend, do I need to backport ?

pmconne:

PR #5609, opened 2023-06-07
Add support for Electron 25

GytisCepk:
## Electron breaking changes:
- Deprecated: `protocol.registerFileProtocol`:
  - Replacement only available from Electron 25. For that reason, let's not replace deprecated API until API is removed to avoid runtime Electron version checks for as long as possible.

- Electron `Event` no longer extends DOM `Event`, and as a result, is causing type incompatibility with `IpcListener`:
  - This is type only change to fix `IpcRendererEvent` type in Electron side. What this show is what our `IpcListener` type is incorrect when running Electron and has been for a while (`evt` argument in `IpcListener` does not match it's type). This needs to be addressed in separate PR.


pmconne (approved):
Tested display-performance-test-app and display-test-app, including sign-in; everything looks good.

ben-polinsky (approved):

PR #5611, opened 2023-06-07
Support OIDC auth in android display-test-app

tcobbs-bentley:


tcobbs-bentley:
Since the master branch doesn't run at all on mobile right now, I can't submit this to that and then backport to 3.7.x. @aruniverse, @calebmshafer, is it OK to have this PR go directly to release/3.7.x like it does? How do we get it to master in the future?

tcobbs-bentley:
@aruniverse As far as I can tell, the failed pipelines are due to a failing test that has nothing to do with any of the changes I made. (All my changes in the PR are in dta Android, which isn't tested.)

aruniverse:
> @aruniverse As far as I can tell, the failed pipelines are due to a failing test that has nothing to do with any of the changes I made. (All my changes in the PR are in dta Android, which isn't tested.)

did you requeue the test?

tcobbs-bentley:
> did you requeue the test?

No; is that the standard thing to do in this situation?

aruniverse:
> > did you requeue the test?
> 
> No; is that the standard thing to do in this situation?

Yes

PR #5612, opened 2023-06-08
Validate NextVersion.md

yashincontrol:
Part 2 of the solution for https://github.com/iTwin/itwinjs-core/issues/5554. Added logic and a rush command that checks for relative links in NextVersion.md. Also modified the docs build pipeline to make sure that this is being followed. 

The build will fail now since there are relative links in the current NextVersion.md


pmconne:
Why isn't this part of the iTwin.js Docs build instead of the iTwin.js build?

PR #5613, opened 2023-06-08
Edl/sweep sections

EarlinLutz:
This PR implements a new method `CurveFactory.createMiteredSweepSections` to create sections swept along a linestring, and optionally return the `RuledSweep` primitive surface/solid they generate and its stroked `IndexedPolyface` mesh. At each vertex, the previous section projects to the plane with normal that is the average of inbound and outbound edge vectors.

**Supporting changes:**
  * New method `Vector3d.createNormalizedStartEnd` returns (if possible) a unit vector from start to end, with start and end given as `XYAndZ`
  * New method `Matrix3d.createFlattenAlongVectorToPlane` returns a matrix which sweeps a vector along the given sweep direction until it hits the plane through the origin with the given normal.
  * New method `Transform.createFlattenAlongVectorToPlane` returns a transform which projects a point onto a plane, sweeping along a direction which may be different from the plane normal.
  * New method `PolylineOps.createBisectorPlanesForDistinctPoints` constructs at each point a plane which bisects the angle between inbound and outbound edges.
  * Swept surface constructors (`LinearSweep`, `RotationalSweep`, `RuledSweep`) now take `CurvePrimitive` input.

**Changes in test management:**
  * Recent change https://github.com/iTwin/itwinjs-core/commit/2527f48a5904c30204968e41e79e37e05a35916b allowed control of debug outputs for geometry tests by enabler flags in `GeometryCoreTestIO`. With these changes, a test developer has to make changes in `GeometryCoreTestIO` to see even minimal error output, and even error messages at point of failure are suppressed.
  * Revised logic:
    * Essentially all geometry tests call the `Checker` constructor, so now this constructor sets `GeometryCoreTestIO.enableConsole` and `GeometryCoreTestIO.enableSave` flags with values from optional arguments defaulting to false. Hence a developer focusing on a single test (or a few tests) can enable debug output by changing only the invocation of the `Checker` constructor near the `it.only` syntax for the test, so it's very convenient.
    * `Checker.announceError` always produces console output, regardless of the enabler flags.


core/geometry/src/solid/SweepContour.ts
```diff
@@ -41,6 +43,12 @@ export class SweepContour {
   public axis: Ray3d | undefined;
 
   private constructor(contour: CurveCollection, map: Transform, axis: Ray3d | undefined) {
+    // hm it seems that bare CurvePrimitives slip through the type checking but some later steps go blank. Wrap them as Path ...
```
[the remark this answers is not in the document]
EarlinLutz: I shoved a simple LineString3d in the test case.

 const lines = PolyfaceQuery.sweepLinestringToFacetsXYReturnLines(linestringA.packedPoints, mesh);

where linestringA was read from a file.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -353,6 +363,51 @@ export class CurveFactory {
...
+   *   * This plane of the input is commonly (but not necessarily) through the start point itself.
+   *   * If the geometry is not "on a perpendicular plane", the output geometry will still be flattened onto the various planes.
+   * * In the "open path" case (i.e when wrapIfPhysicallyClosed is false or the path does not have matched first and last points)
+   *       the first output plane will be a the base of the first edge and on a perpendicular plane.
+   *   * Hence if the section is already on a plane perpendicular to the
+   * * In the "closed path" case, the output plane for the first and last point is the bisector of the two planes perpendicular to the first and last edges,
+   *    and teh first/last section geometry are different from the input section on its perpendicular plane.
+   * * The centerline path does NOT have to be planar -- twisting effects effects will appear in the various bisector planes.
+   * @param centerline centerline of pipe
+   * @param startSection curve data to be swept.  As noted above, this should be on
```
saeeedtorabi, line 378: startSection does not match with the function arg which is "section"

core/geometry/src/geometry3d/Point3dVector3d.ts
```diff
@@ -980,6 +980,18 @@ export class Vector3d extends XYZ {
       return result;
     return undefined;
   }
+  /**
+   * Create a normalized vector from startPoint to endPoint
+   * @param result optional result
```
saeeedtorabi, line 985: Please add all parameters or mention none of them

core/geometry/src/geometry3d/PolylineOps.ts
```diff
@@ -118,62 +120,62 @@ export class PolylineOps {
...
-   * Return a simplified subset of given points, omitting points on "danglers" that depart and return on a single path.
-   * @param source input points
-   */
-   public static compressDanglers(source: Point3d[], closed: boolean = false, tolerance: number = Geometry.smallMetricDistance): Point3d[] {
-     let n = source.length;
-     const squaredDistanceTolerance = tolerance * tolerance;
-     if (closed)
+  /**
+     * Return a simplified subset of given points, omitting points on "danglers" that depart and return on a single path.
+     * @param source input points
```
saeeedtorabi, line 137: Please add all parameters or mention none of them

core/geometry/src/serialization/GeometrySamples.ts
```diff
@@ -2837,4 +2837,32 @@ export class Sample {
     }
     return builder.claimPolyface(true);
   }
+  /**
+   *  alternately move in directions in the steps array, creating numStroke total strokes.
+   * * if start is a single point, create a new point array with start as its first entry.
+   * * if start is an array, add to it.
+   * * if start is an empty array, create a new one and start at 000
+   * @param start
```
saeeedtorabi, line 2845: Would you mind adding some description like "start of the zig zag" for all aprameters? Otherwise, dont mention them at all because it's clear from API itself

core/geometry/src/test/GeometryCoreTestIO.ts
```diff
@@ -27,9 +27,9 @@ import { prettyPrint } from "./testFunctions";
 // Methods (called from other files in the test suite) for doing I/O of tests files.
 export class GeometryCoreTestIO {
   /** For debugging: set to true to enable output to console via wrapped methods. */
-  public static enableConsole: boolean = false;
+  public static enableConsole: boolean = true;
```
saeeedtorabi, line 30: I think you forgot to put them back to false!

ben-polinsky (approved):

pmconne (approved):

core/geometry/src/solid/SweepContour.ts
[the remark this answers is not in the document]
pmconne: Sounds like when reading from file you got back `any`, which bypasses all type checks.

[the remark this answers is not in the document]
EarlinLutz: Right, and immediately typechecked it as LineString3d, which was correct.  The subtlety that it is going to be used in API that wants Path for sweeping is not evident in the data in the file.

saeeedtorabi (approved):
I only reviewed code style. Looks good now.

dassaf4:
User wants a method that produces a mesh vis-à-vis `PolyfaceBuilder.addMiteredPipes`. I can add this wrapper if you like after I review the rest of this PR.

dassaf4 (approved):
In addition to these suggestions, I will make other innocuous doc changes to the branch, along with expanded `RuledSweep` and `Polyface` output.

dassaf4:
@EarlinLutz I made some additions and comments, please review...

core/geometry/src/solid/SweepContour.ts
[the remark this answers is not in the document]
dassaf4: @pmconne I expanded some input types to explicitly allow CurvePrimitive. See [commit](https://github.com/iTwin/itwinjs-core/pull/5613/commits/6641942b19e1bfc7a7a38a6b11ac29678cafe4e9). The underlying issue was a missing CurvePrimitive case in FrameBuilder.

PR #2881, opened 2021-12-09
Bim map layers

RBBentley:

PR #5616, opened 2023-06-09
Fix bug where background map layers didn't blend with base solid color

### LINKED ISSUE #4980: Background map-layers does not blend with base map when is set to solid color
map-layer imagery tiles are not blended with base map when it's set to a solid color.  Same map-layer blends correctly when base map is a map-layer definition, such as bing. Also moving this cannot be reproduced if map-layer is added to overlay map  instead of background map.

1. Open a geo-referenced imodel and fit the view around the Philadelphia area.
2. Run the following key-in to add a background map-layer:
`fdt attach arcgis maplayer https://hydro.nationalmap.gov/arcgis/rest/services/wbd/MapServer hydro`
3. Use this key-in to set the background to solid red color: 
 `fdt set map base color 255 0 0`

It currently looks like this:
image

Instead it should blend with the red background color:
image

This has been tested with master version.  I can confirm it was working fine with 3.0.2 and it's broken since at least 3.2.3.

eringram:
Fixes #4980.
This bug was introduced by changes to the reality mesh shader from PR #2881. This PR caused a side effect where when a transparent pixel was sampled from the texture being applied in the `applyTexture` function, it was discarded whether or not an opaque base color was present. This fix just adds a check that tells the shader to not discard the pixel if there is an opaque base color which should be visible behind a transparent map layer texture.


pmconne:
If there is an issue please link to it in PR description (e.g., `Fixes #1234`).

pmconne:
Did you run ImageTests? Is there a test case that previously exhibited this bug, or will you add one?

eringram:
>I can also confirm it happens with the same iModel/map layer in #4980 when you zoom out far enough

Amending this to say the problem might not be happening in this case, as the white outside certain extents matches up with bing maps extents. 

hydro_bing

I still think it could be an issue in the first 2 images however.

mdastous-bentley:
@eringram If the base map is set to a solid color, solid colors tiles  should be displayed anywhere you would see bing map tiles.
I just tested your branch by using my own iModel and could not reproduce your 'clipped solid colors tiles' problem.  As a side note, you should also ensure your fix works when the background map is to 'Ellipsoid' globe:
image

eringram:
After this merges I will make a separate issue with steps for reproducing this bug of the base color not covering the same map tiles as Bing/other background maps.

markschlosseratbentley (approved):
I see a bunch of files with just whitespace changes?  is that intentional?

eringram:
> I see a bunch of files with just whitespace changes? is that intentional?

The whitespace changes to `applyTexture()` in glsl/RealityMesh.ts are intentional because the indentation was inconsistent and it was difficult to read. Any others like adding/removing extra spaces were probably automatic by my VScode, but I believe fit the typescript syntax guidelines better-- mostly adding spaces between curly braces and property names, between variables and `|` operators, etc.

PR #5618, opened 2023-06-09
Pass process.env.POD_NAME as logId to V2 checkpoint cloud containers.

nick4598:


aruniverse (approved):
changes look fine, but deferring to @wgoehrig

wgoehrig (approved):

PR #5621, opened 2023-06-12
Add ViewStore for "desktop" views

kabentley:
There are currently 2 forms of *view persistence*:
1. Elements in the iModel
2. cloud view service
(actually there are 2 flavors of cloud-views).

Neither are appropriate for desktop apps or LiveSync: 
 - The element-based api has fundamentally the wrong transaction model since adding/modifying views requires making a changeset. Often views are created by users with readonly access to the iModel.
- the cloud apis do not have the concept of "selectors" or display styles. Instead they store a copy of everything with every view. Also, they do not support local caching, as will be necessary for editing applications (and for offline use).

Both current forms also suffer the severe limitation that their views are specific to a single iModel since they store ElementIds. It is necessary to store Guids if views are to be shared across branches.

This PR introduces a new persistence for views called a ViewStore. A ViewStore is a CloudSqlite database "attached" to an iModel. The same ViewStore may be attached to multiple (branches of) an iModel. See documentation in ViewStore.ts.

ViewStores are intended to become the primary way for storing Views of iModels. They are not a complete replacement for element-based views, since Sheets reference view elements, so there will still need to be some ViewDefinition elements. The cloud view service is useful to store "iTwin views" that don't have an iModel and therefore don't have ElementIds or selectors. Mainly they should store camera position an lists of reality models, etc.

Notes:
 - the existing apis are unchanged, so apps won't use ViewStores unless they are changed to use them, other than "load view" now supports the "viewstore viewId" syntax.
 - LiveSync should use ViewStores and ViewGroups to organize views from desktop applications
 - from the frontend, the Rpc apis for reading from and writing to ViewStores are accessed via `connection.views.viewStoreReader` and `connection.views.viewStoreWriter` respectively
 - ViewStores support `ViewGroups` and `Tags` to help organize and annotate views. ViewGroups may nest and form a tree structure for Views, analogous to directories and files. There is always a "root" ViewGroup, so every View is in a ViewGroup.
 - VeiwGroups each have a "default view" (including the root ViewGroup) so that users or groups of users can choose their default view and one can be set for the root.
 - Model and Category Selectors may either be an Id list, or use an ECQuery, plus lists of "add" and "remove" ids. When they are "loaded" from the frontend they return the result of the query+adds+removes.

For another PR:
- frontend apis that modify displayStyles or selectors should invalidate their ids so they'll be saved if a new view is created
- write learning documentation
- examples


common/api/core-common.api.md
```diff
@@ -10230,15 +10236,23 @@ export class ViewFlags {
...
 export type ViewFlagsProperties = Mutable<NonFunctionPropertiesOf<ViewFlags>>;
 
+// @public
+export type ViewIdString = Id64String;
+
 // @public
 export interface ViewQueryParams extends EntityQueryParams {
     // (undocumented)
-    wantPrivate?: boolean;
+    readonly wantPrivate?: boolean;
```
pmconne, line 10245: Motivation for breaking change?

kabentley: Good catch. I meant to take this out. At one point I changed this interface so you could use it to query ViewStores too. But I then decided that I’d make it so ViewStores are completely separate. This is a vestige of that that should be removed.

core/backend/src/BlobContainerService.ts
```diff
@@ -87,21 +87,29 @@ export namespace BlobContainer {
...
+   */
+  export type RequestAccessLevel = "write" | "read" | "admin" | "writeIfPossible";
+
   /** Information required to request an access token for a container. */
   export interface RequestTokenProps extends AccessContainerProps {
-    /** If true, token should provide write access.
-     * @note if write access is requested and the user is authorized for read but not write, an exception will be thrown (i.e. a read token is *not* returned).
-     */
-    forWriteAccess?: boolean;
+    storageType: Provider;
```
wgoehrig, line 110: Why would we want to let a client choose this?

I'd prefer to just say the service always dictates the provider - that gives us a lot more flexibility and control over our costs...

core/backend/src/BlobContainerService.ts
```diff
@@ -87,21 +87,29 @@ export namespace BlobContainer {
...
 
+  /**
+   * Access level to request for token.
+   * - `"write"`: request a writeable AccessToken.
+   * - `"read"`:  request a read-only AccessToken.
+   * - `"admin"`:  request a admin AccessToken.
+   * - `"writeIfPossible"`: first request a writeable AccessToken. If the user is not authorized for write, request a read-only AccessToken.
+   * If the user is not authorized for the level requested, an exception is thrown.
+   */
+  export type RequestAccessLevel = "write" | "read" | "admin" | "writeIfPossible";
```
wgoehrig, line 106: What does "admin" mean in this context?

For Azure, these "tokens" are really SAS URLs, right?  What rights would an "admin" SAS have that a "write" SAS wouldn't?

core/backend/src/BlobContainerService.ts
kabentley: This is for the request for an access token. This was returned when the container was created, so you have to supply it to get an access token. The choice of supplier is made in the create-new-container call. After that the container can’t change storage type, right?

core/backend/src/BlobContainerService.ts
kabentley: I thought that would be used the delete versions if we enable automatic versioning.

core/backend/src/ViewStore.ts
```diff
@@ -0,0 +1,1397 @@
...
+    parentId: RowId;
+    defaultViewId?: RowId;
+  }
+
+  /** a row in the "thumbnails" table */
+  interface ThumbnailRow {
+    viewId: RowId;
+    data: ThumbnailData;
+    format: ThumbnailFormatProps;
+    owner?: string;
```
diegopinate, line 136: What's the purpose of allowing thumbnails to have a different owner than the owner of the saved view that's related to the thumbnail?

core/backend/src/ViewStore.ts
```diff
@@ -0,0 +1,1397 @@
...
+  type TimelineRow = TableRow;
+
+  /** a row in the "views" table */
+  interface ViewRow extends MarkRequired<TableRow, "name"> {
+    className: string;
+    groupId: RowId;
+    isPrivate?: boolean;
+    modelSel?: RowId;
+    categorySel: RowId;
+    displayStyle: RowId;
```
diegopinate, line 122: Since model selector, category selector and display styles are separate from a view, what will be the procedure to update thumbnails of all affected views when an update occurs?

core/backend/src/test/standalone/ViewStoreDb.test.ts
```diff
@@ -0,0 +1,332 @@
...
+  });
+
+  after(async () => {
+    vs1.closeDb(true);
+    Logger.setLevel("SQLite", LogLevel.Error);
+  });
+
+  it("ViewDb", async () => {
+    const c1 = vs1.addCategorySelectorRow({ name: "first cat", json: "cat1-json" });
+    const ds1 = vs1.addDisplayStyleRow({ name: "first style", json: "style1-json" });
```
diegopinate, line 33: Shouldn't we fail adding non-valid JSON strings?

core/backend/src/ViewStore.ts
kabentley: only that they're added/updated separately. I'm not even sure that `owner` will be used for anything other than views.

core/backend/src/ViewStore.ts
kabentley: Thumbnails are only approximations and can only be updated from a view in the frontend. They can be out-of-date and still be useful. There's no expectation that they'll be updated other than one-by-one manually.

core/backend/src/test/standalone/ViewStoreDb.test.ts
kabentley: This test exists merely to validate that the SQL is working properly.

common/api/core-common.api.md
kabentley: fixed

kabentley:
Can anyone help me understand why the tests fail only on Mac?

pmconne:
> Can anyone help me understand why the tests fail only on Mac?

This has been popping up with some frequency on macOS for the past week or so.
Presumably when #5681 will resolve it by eliminating puppeteer entirely.

```
Invoking: npm -s test 
Timed out after 30000 ms while trying to connect to the browser! Only Chrome at revision r1022525 is guaranteed to work.
TimeoutError: Timed out after 30000 ms while trying to connect to the browser! Only Chrome at revision r1022525 is guaranteed to work.
    at Timeout.onTimeout (/Users/platform-bld/vsts_work_a/1/s/common/temp/node_modules/.pnpm/puppeteer@15.5.0/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserRunner.js:300:20)
    at listOnTimeout (node:internal/timers:569:17)
    at process.processTimers (node:internal/timers:512:7)
```

core/backend/src/ViewStore.ts
diegopinate: Is there a reason why all tables have to contain an owner column then?

kabentley: Not really. I just figured it could be used to see who last changed it. Probably shouldn't be called "owner" though. You are very welcome to change that if you want!

kabentley: I do want to talk to you and @wgoehrig about how to get the "identity" (some unique string that isn't email address) of the person performing an action in the frontend. What do you to today?

kabentley:
I would like to complete this PR today because @khanaffan needs the changes here for his PR . @diegopinate @wgoehrig, please resolve existing comments or reply. @pmconne @aruniverse @calebmshafer please approve.

pmconne (approved):

bbastings (approved):

diegopinate (approved):

core/backend/src/ViewStore.ts
diegopinate: So the token has a subject that we can use as a user id. That's what we do today for saved views api.

kabentley: Yes, we can use the user's GUID for the owner field in the database. Is that what you use today? But, when we want to show the identity to a person (e.g. "who owns this view?"), we need another string that means something to them. It should be a user-chosen string (like your github id) and should **not** be their email address. I'm guessing lots of things can use it with an "@" prefix, like we do here.

full-stack-tests/core/src/backend/AzuriteTest.ts
```diff
@@ -0,0 +1,196 @@
...
+
+export namespace AzuriteTest {
+
+  export const storageType = "azure" as const;
+  export const httpAddr = "127.0.0.1:10000";
+  export const accountName = "devstoreaccount1";
+  export const baseUri = `http://${httpAddr}/${accountName}`;
+
+  export const getContainerUri = (id: string) => `${baseUri}/${id}`;
+  const pipeline = azureBlob.newPipeline(new azureBlob.StorageSharedKeyCredential(accountName, "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="));
```
aruniverse, line 24: sorry, stupid q, but have to ask, this is not an actual valid/real account name & key right? i know this is testing against azurite, but having an actual account key looking string committed to src made me nervous

.vscode/settings.json
```diff
@@ -108,5 +108,10 @@
   "terminal.integrated.env.windows": {
     "NODE_ENV": "development"
   },
-  "typescript.tsdk": "common/temp/node_modules/typescript/lib"
+  "typescript.tsdk": "common/temp/node_modules/typescript/lib",
+  "cSpell.words": [
```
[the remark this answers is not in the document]
kabentley: Yup, good catch.

aruniverse (approved):

PR #5626, opened 2023-06-13
pin tile cache URL expiry timestamp

paulius-valiunas:
This will improve request caching in web browsers. Right now, every time the client requests a tile cache URL, they get a different timestamp in the authorization parameters, which means the web browser can't reuse any locally cached responses from blob storage after the viewer page is refreshed. With this change, the tile cache URL will only change once a week.

Important note: Amazon S3 storage API doesn't allow specifying an exact expiry timestamp, and as a result the response from `getDownloadConfig` will contain only an approximation of the `expiresOn` timestamp we provide. So if using S3 storage, it is recommended to set this to a shorter period of time. Some S3 providers might not support expiry times beyond a few hours.


paulius-valiunas:
keeping this marked as draft for now, while we wait for final object-storage 2.0.0 release.

core/backend/src/IModelHost.ts
```diff
@@ -10,6 +10,7 @@
 import "./IModelDb"; // DO NOT REMOVE OR MOVE THIS LINE!
 
 import * as os from "os";
+import "reflect-metadata"; // this has to be before @itwin/object-storage-* and @itwin/cloud-agnostic-core imports
```
pmconne, line 13: Please explain why in your comment.

pmconne (approved):

paulius-valiunas:
https://github.com/Mergifyio backport release/4.0.x

PR #5627, opened 2023-06-13
Work around content range tile problems

markschlosseratbentley:
https://github.com/iTwin/itwinjs-backlog/issues/761


markschlosseratbentley:
The UseContentBoundingVolumes workaround functions well for this particular dataset.

However, in the case of the ClampByContentBoundingVolumesZ workaround:
- I am not completely sold on the way I implemented it in the tile tree system – I need to first pass through the children, gather the Zs, then load the children tile props a second time.
- It gets rid of the gaps, but some tiles near the camera will “blink out” of existence when zooming in on them.

I theorize the blinking out of tiles near the camera may be because I actually need to pass through the entire JSON (tileset) ahead of time and gather min/max Zs there for every single tile globally, not just when loading children, which appears to be what is happening typically – the entire dataset is not parsed in the beginning – it is parsed here and there as needed to load children. Does this seem right to you, @pmconne ? Not sure yet where / how to parse through everything up front once to gather these Zs.

pmconne:
> Not sure yet where / how to parse through everything up front once to gather these Zs.

I updated your issue with details of the problem and how you are attempting to address it, but yes, the fact that subtrees can be referenced as external JSON files to be fetched asynchronously makes computing min+max Z problematic.

PR #5628, opened 2023-06-13
Ensure Ids are included in render timeline if needed

pmconne:
If timeline is being applied on the frontend (i.e., `TileAdmin.enableFrontendScheduleScripts` is `true`), the element Ids need to be included.
This was the case for scripts stored in the display style's JSON properties, but not for those stored on RenderTimeline elements - they were always unconditionally omitted.


pmconne:

PR #5662, opened 2023-06-21
Fix Entity release tags

pmconne:
Part of https://github.com/iTwin/itwinjs-backlog/issues/768, addressing the following:
- [x] Constructors for `Entity` and most of its subtypes were all flagged as `@internal` making it impossible to implement a subclass using public API. Changed to `protected` to ensure `IModelDb.constructEntity` constructs the most-derived class (vs calling `new` directly), and `@public` so subclass constructors can call `super`.
  - A handful were `public` and `@public`; left intact to avoid breaking API.
- [x] `toJSON` overrides were mostly flagged as `@internal`, presumably to reduce docs clutter. This hides useful information about each override's narrower return type, and makes it a violation of `@itwin/eslint-plugin`'s no-internal rule to call `Entity.toJSON` on a subtype of `Entity`.
- [x] Some overrides of base class methods were marked `@internal` to avoid duplicating documentation and cluttering docs in cases where the fact that the subclass overrides the method is of limited interest (more implementation detail). This produces lint errors if the compiler can see the subtype at the call site.


saskliutas (approved):

aruniverse (approved):

pmconne:
@ben-polinsky I cannot tell from this error where the offending link is in the source code.
```
Invalid Link: "[constructEntity]($core-backend.BriefcaseDb)" in file "reference/core-backend/imodels/standalonedb"
```

ben-polinsky:
> @ben-polinsky I cannot determine from this error where the offending link is in the source code.
> 
> ```
> Invalid Link: "[constructEntity]($core-backend.BriefcaseDb)" in file "reference/core-backend/imodels/standalonedb"
> ```

Looking into...

yashincontrol:
> > @ben-polinsky I cannot determine from this error where the offending link is in the source code.
> > ```
> > Invalid Link: "[constructEntity]($core-backend.BriefcaseDb)" in file "reference/core-backend/imodels/standalonedb"
> > ```
> 
> Looking into...
> 
> @yashincontrol do you think you could try to find a solution? The issue is that the link for `StandaloneDb`'s `constructEntity` method, which is inherited from `BriefcaseDb`, is not resolving. The same method does resolve on BriefcaseDb. And everything works in dev.

Looking into it.

ben-polinsky:
@pmconne docs issue fixed

PR #5666, opened 2023-06-21
Fix issues with ArcGIS feature service not providing its extent in EPSG:3857.  Also make sure default symbology is correctly  applied if none is provided.

mdastous-bentley:
Specific ArcGIS feature service could not be displayed.  

After investigation 2 issues were found:
1. The advertised extent in the service metadata was not available in EPSG:3857
2. A default symbology of ` null `was provided, causing our renderer to ignore the symbology definition.

Key-in used to reproduce:
`fdt attach arcgisFeature maplayer https://services-eu1.arcgis.com/NPIbx47lsIiu2pqz/ArcGIS/rest/services/Neptune_Coastline_Campaign_Open_Data_Land_Use_2014/FeatureServer Neptune_Coastline_Campaign_Open_Data_Land_Use_2014`

### DIFF
--- a/extensions/map-layers-formats/src/ArcGisFeature/ArcGisFeatureProvider.ts
+++ b/extensions/map-layers-formats/src/ArcGisFeature/ArcGisFeatureProvider.ts
@@ -100,6 +100,7 @@ export class ArcGisFeatureProvider extends ArcGISImageryProvider {
           for (const layer of this.serviceJson.layers) {
             if (layer.defaultVisibility) {
               this._layerId = layer.id;
+              break;
             }
           }
         } else {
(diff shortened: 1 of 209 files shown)


extensions/map-layers-formats/src/ArcGisFeature/ArcGisFeatureProvider.ts
```diff
@@ -166,17 +169,24 @@ export class ArcGisFeatureProvider extends ArcGISImageryProvider {
...
-          low: {x: json.fullExtent.xmin, y: json.fullExtent.ymin},
-          high: {x: json.fullExtent.xmax, y: json.fullExtent.ymax} });
-
-        const west = this.getEPSG4326Lon(range3857.xLow);
-        const south = this.getEPSG4326Lat(range3857.yLow);
-        const east = this.getEPSG4326Lon(range3857.xHigh);
-        const north = this.getEPSG4326Lat(range3857.yHigh);
-        this.cartoRange = MapCartoRectangle.fromDegrees(west, south, east, north);
+    if (this._layerMetadata?.extent) {
+      const layerExtent = this._layerMetadata?.extent;
```
aruniverse, line 173: the `?` is redundant here

aruniverse (approved):

extensions/map-layers-formats/src/ArcGisFeature/ArcGisFeatureProvider.ts
mdastous-bentley: fixed

PR #5672, opened 2023-06-22
Update release tags

pmconne:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/768.

Unaddressed:
- [ ] APIs used strictly in tests - IModelDb.openDgnDb, IModelHost.setHubAccess, BackendHubAccess, BriefcaseLocalValue

Address separately:
- [ ] ClashDetectionServer - not in core, contact package maintainer.
- [ ] AccuDraw - usage of internal APIs appears intended to work around input problems; file separate issue to resolve those underlying problems.
- [ ] getAgendaElements - not in core, contact constraint-solver-frontend package maintainer.

UI and presentation APIs:
- [x] Most (all?) addressed. See details in https://github.com/iTwin/itwinjs-backlog/issues/768. File follow-up issues in appropriate repositories if necessary - not part of itwinjs-core.

Remove usage:
- [x] IModelDb.nativeDb.inlineGeometryParts - transformer can do this for you
- [x] ToolAdmin.fillEventFromCursorLocation - it's already filled at the point where you're calling it
- [x] IModelHost.hubAccess - deleteIModel, createNewIModel

Promoted:
- [x] BilinearPatch
- [x] queryLastModifiedTime
- [x] Entity subclass constructors - made `protected` and `@public` so superclasses in other packages can invoke `super`
- [x] Entity subclass toJSON methods
- [x] Relationship constructor
- [x] Schemas.unregisterSchema
- [x] isSubclassOf
- [x] BriefcaseManager.getBriefcaseBasePath
- [x] EditCommandAdmin runCommand and finishCommand
- [x] IModelApp.toolAdmin.toolSettingsState

Use existing alternative public API:
- [x] ClassRegistry.register - use registerModule
- [x] EditableWorkspaceDb, ITwinWorkspaceContainer, WorkspaceContainer.cloudContainer - use PropertyStore
- [x] IpcApp.callIpcChannel - use makeIpcProxy
- [x] ToolAdmin.onPostInstallTool, onInstallTool, startPrimitiveTool - don't override Tool.run, override individual methods instead. See issue comments for details.
- [x] SchemaContext.getCachedSchemaSync - use getSchemaSync

Use new alternative public API:
- [x] CoreTools - added translatePrompt and translateInput
- [x] StandaloneDb.createClassViews
- [x] IModelHost.hubAccess queryChangeset and queryIModelByName - added same methods to BriefcaseManager
- [x] ChannelAdmin.sharedChannel - added ChannelControl.sharedChannelName
- [x] ToolAdmin.sendButtonEvent - use AccuDraw.sendDataPoint


core/backend/src/IModelDb.ts
```diff
@@ -1366,8 +1366,9 @@ export namespace IModelDb { // eslint-disable-line no-redeclare
       return this.tryGetModelJson({ id });
     }
 
-    /** Query for the last modified time of the specified Model.
-     * @internal
```
kabentley, line 1370: Before we make methods like this @public, should we make them take named arguments?

missing “if” in comment.

core/frontend/src/tools/ToolAssistance.ts
```diff
@@ -64,6 +64,48 @@ export enum ToolAssistanceInputMethod {
   Touch,
 }
 
+/** Common Tool Assistance main instruction prompt messages
+ * @public
+ * @extensions
+ */
+export enum ToolAssistancePromptKey {
```
kabentley, line 71: I prefer to use const members of an “as const” object over string enums. It’s more efficient unless you plan to do reverse lookups.

core/backend/src/IModelDb.ts
pmconne: Added "if".
I don't really want to encourage proliferation/expansion of these convenience APIs. If people want to execute a more complicated query for last modified time than "by id", they can write ECSql to do it.

core/frontend/src/tools/ToolAssistance.ts
pmconne: I'll switch it around.

core/backend/src/IModelDb.ts
kabentley: Yeah, that’s probably why it was @internal to begin with. But you only get one chance to set something in stone..

core/frontend/src/tools/ToolAssistance.ts
pmconne: Changed.

kabentley (approved):

bbastings (approved):

aruniverse (approved):

wgoehrig (approved):

PR #5667, opened 2023-06-21
Add support for API keys to DTA Android

tcobbs-bentley:

PR #5674, opened 2023-06-22
use node 18.16.0 on release/3.7.x

DanRod1999:
in relation to this thread
https://github.com/iTwin/itwinjs-core/pull/5667

Cherry picked changes from: https://github.com/iTwin/itwinjs-core/commit/8d1e486fd25042c495d13bdba66e26cc14b8f29e
ignored where 3.7.x used versions not on 18

### DIFF
--- a/common/config/azure-pipelines/ci-ios.yaml
+++ b/common/config/azure-pipelines/ci-ios.yaml
@@ -30,15 +30,15 @@ jobs:
 
     steps:
       - task: InstallAppleProvisioningProfile@1
-        displayName: 'Install an Apple provisioning profile'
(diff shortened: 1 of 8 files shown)


aruniverse (changes requested):
i dont think this is right

DanRod1999:
> i dont think this is right

Cherry picked commit from master to use node 18.16.0
brought over some format/syntax changes as well, so let me know if a manual change is better. Don't know why they aren't the same anyway

DanRod1999:
also added changes to node 16 version to temporarily force LKG version

aruniverse (approved):

PR #5675, opened 2023-06-22
Add ContainerType to BlobContainerService Interface

MichaelSwigerAtBentley:
Adds ContainerType to Metadata interface in BlobContainerService.ts for use by the Storage Container Service Client.
ContainerType describes how a container is used.

Format is not being removed, it is defined by the service itself, and will be returned with metadata from the requestToken() response. It is not required for creating a new container.


kabentley:
@MichaelSwigerAtBentley please see my changes to this interface in [this PR](https://github.com/iTwin/itwinjs-core/pull/5621). They won’t conflict with this change, but you should be aware of them.

MichaelSwigerAtBentley:
> @MichaelSwigerAtBentley please see my changes to this interface in [this PR](https://github.com/iTwin/itwinjs-core/pull/5621). They won’t conflict with this change, but you should be aware of them.

Looks good, I'll add support for the RequestAccessLevel change and "writeIfPossible" into the client and the service.

kabentley:
I think the “if possible” concept and be implemented client-side if it’s difficult server-side (e.g. somehow returning what level was returned.) The client implementation should somehow cache recent requests, because I’m presuming that repeated requests for the same token don’t actually cause any http delay unless necessary (e.g an lru cache, with a validity timeout).

MichaelSwigerAtBentley:
> I think the “if possible” concept and be implemented client-side if it’s difficult server-side (e.g. somehow returning what level was returned.) The client implementation should somehow cache recent requests, because I’m presuming that repeated requests for the same token don’t actually cause any http delay unless necessary (e.g an lru cache, with a validity timeout).

That makes sense. If I'm understanding this right: I'll implement a cache client-side that stores the last few requests and their token's expire time. If a cache hit occurs and the token isn't expired I pass back the old token and never send a request to the service.

core/backend/src/BlobContainerService.ts
```diff
@@ -45,26 +45,21 @@ export namespace BlobContainer {
     iTwinId: Id64String;
     /** optionally, an iModelId within the iTwin. If present, container is deleted when the iModel is deleted. */
     iModelId?: Id64String;
```
kabentley, line 47: I’d leave this. We will need containers owned by individuals.

core/backend/src/BlobContainerService.ts
```diff
@@ -45,26 +45,21 @@ export namespace BlobContainer {
...
-    /** Human readable explanation of the information held in the container. This will be displayed in the administrator RBAC panel, and on usage reports. */
-    description: string;
-    /** a machine-readable string that describes the "format" of the data in this container (e.g. "CloudSqlite") */
-    format: string;
-    /** an identifier of the application that uses this this container */
-    application: string;
-    /** Additional properties stored on the container */
-    [propertyName: string]: string;
+    /** Human-readable name for the container. This will be displayed in the administrator RBAC panel, and on usage reports. Non-unique.*/
+    name: string;
```
kabentley, line 58: `name` is a bad name for this (pun intended). Maybe `rbacLabel` or something. Should it be required?

core/backend/src/BlobContainerService.ts
```diff
@@ -45,26 +45,21 @@ export namespace BlobContainer {
...
+    /** The machine-readable string that describes what the container is being used for (e.g. "workspaces"). Always lowercase and singular. */
```
wgoehrig, line 59:
```suggestion
    /** The machine-readable string that describes what the container is being used for (e.g. "workspace"). Always lowercase and singular. */
```

core/backend/src/BlobContainerService.ts
```diff
@@ -45,26 +45,21 @@ export namespace BlobContainer {
...
    */
   export interface Metadata {
-    /** Human readable explanation of the information held in the container. This will be displayed in the administrator RBAC panel, and on usage reports. */
-    description: string;
-    /** a machine-readable string that describes the "format" of the data in this container (e.g. "CloudSqlite") */
-    format: string;
-    /** an identifier of the application that uses this this container */
-    application: string;
-    /** Additional properties stored on the container */
-    [propertyName: string]: string;
```
kabentley, line 67: Why remove this? That means it won’t allow “unrecognized” members. I thought the idea was that people should be able to store their own Metadata, no?

wgoehrig (approved):

core/backend/src/BlobContainerService.ts
wgoehrig: Yes, but the owner identity needs to come via OIDC - otherwise it could be spoofed.

Wouldn't it be sufficient to say that the user who created the container is always the initial "owner"?

core/backend/src/BlobContainerService.ts
wgoehrig: I'm not against the idea of additional metadata, but I'd like to understand the use cases a bit better before we commit to supporting it in the API...

I mean, you can always store other metadata _within_ your containers.  And if we discover that consumers prefer using the service to track certain things, it should be very quick and easy for us to add additional metadata fields.

core/backend/src/BlobContainerService.ts
MichaelSwigerAtBentley: I agree, will change name to rbacLabel. As for the requirement, I set it as required since I updated description to be optional. I think one of the two should be required for some sort of description, but I could perhaps set a default value for name in the client or just allow it all to be optional.

wgoehrig: What exactly is wrong with `name`?

Every other type of object we manage in RBAC (iTwins, iModels, etc.) just has a `name` property - why shouldn't storage containers?

I could see some cases where the name is often (or always) generated (e.g., "ViewStore for {iModel Name}"), in which case I would concede that the "name" is acting more like a label, but surely there are other uses where container would benefit from having a proper name.  And without requiring one, how would an end-user administrator possibly understand what they're granting access to?

kabentley:
> That makes sense. If I'm understanding this right: I'll implement a cache client-side that stores the last few requests and their token's expire time. If a cache hit occurs and the token isn't expired I pass back the old token and never send a request to the service.
 
Something like that. The cache expiry should be well shorter than the token expiration (maybe 10 minutes?) so you can refresh it before it expires. Also, keep in mind that the user token may vary even for the same backend through RPC, so the user token needs to be part of the cache key.

core/backend/src/BlobContainerService.ts
kabentley: Because it will be confused with [container name](https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names). That’s what I thought it meant.

wgoehrig: I see - that makes sense...

How about just calling this `label` then?

kabentley: I like `label`. There’s no expectation of uniqueness.

core/backend/src/BlobContainerService.ts
kabentley: The idea of metadata is that you can read it without an access token. Here’s an example of something I think would be useful. The concepts of “purging” and “vacuuming” are things that average clients won’t know when to do. I imagine an agent that scans through containers with the appropriate metadata and performs one of those actions as a hygiene action periodically. Probably other more specific tasks that can be performed on specially tagged containers.

wgoehrig: Per our discussion yesterday, the service will continue to dictate the schema metadata for the time being - as new use cases for these fields come up @MichaelSwigerAtBentley will add them to the service as well as the type here.

PR #5461, opened 2023-04-28
@itwin/eslint-plugin 4.0.0-dev.37

pmconne:
Enforces no unbound invocations of static methods, as we do for instance methods.
Don't review yet.

PR #5677, opened 2023-06-22
rush update --full; 

aruniverse:
We're not updating the lockfile often anymore due to running `rush update -p` whenever we update the addon, so manually ran it myself

- fix core-webpack build
  - image
  - caused by changes in https://github.com/webpack/webpack/commit/98942261492c47235c628e047a770dd1915c25d1

- update audit, cleanup cve suppression
- pin types/node ver down 
  - fyi @swbsi 
  - image
- pin itwin/eslint-plugin ver down 
  - fyi @pmconne we should try to get https://github.com/iTwin/itwinjs-core/pull/5461 in so we can then update to the latest eslint-plugin ver which has breaking changes wrt to the config how things are setup


pmconne:
Are you planning to keep doing regular manual full updates, or automate them?

pmconne (approved):

aruniverse:
> Are you planning to keep doing regular manual full updates, or automate them?

For the time being i think manual, [nightman](https://github.com/iTwin/nightman) will always pull in the latests so i usually (or try to) catch reasons to update there

Issue #5507, opened 2023-05-11
PolyfaceQuery.sweepLinestringToFacetsXYReturnChains unexpected output


dassaf4:
User states this method produces two unexpected spikes in z-coordinates on this input, consisting of a linestring and two meshes: [linestring_drape.txt](https://github.com/iTwin/itwinjs-core/files/11456522/linestring_drape.txt)

Output consists of two linestrings below. User states entries 23 and 26 in the first linestring are unexpected:
{
    "0": -8365354.40647523,
    "1": 4852793.72391395,
    "2": -31.755184061248592,
    "3": -8365345.452249724,
    "4": 4852797.780250252,
    "5": -31.838332371431417,
    "6": -8365321.29976544,
    "7": 4852808.72152024,
    "8": -31.887235312354765,
    "9": -8365311.694367033,
    "10": 4852812.264060729,
    "11": -31.904040163776045,
    "12": -8365279.99985307,
    "13": 4852823.95322699,
    "14": -31.710407942126093,
    "15": -8365274.02435133,
    "16": 4852827.20992831,
    "17": -31.67226221355585,
    "18": -8365268.375529689,
    "19": 4852830.2885980345,
    "20": -31.636201889653567,
    "21": -8365268.383727771,
    "22": 4852830.284129989,
    "23": -41.96804154027708,
    "24": -8365268.384860523,
    "25": 4852830.283512628,
    "26": -43.395652929644115,
}
{
    "0": -8365268.375529689,
    "1": 4852830.2885980345,
    "2": -31.636036452102932,
    "3": -8365250.50004448,
    "4": 4852840.03093345,
    "5": -31.64511431897961,
    "6": -8365224.37153723,
    "7": 4852862.03204164,
    "8": -31.646300954448527,
    "9": -8365203.30013189,
    "10": 4852890.80255174,
    "11": -31.63008942901338,
    "12": -8365188.12882827,
    "13": 4852916.18856048,
    "14": -31.611144526394117,
    "15": -8365178.85732679,
    "16": 4852947.49787083,
    "17": -31.574975729665287,
    "18": -8365118.17202385,
    "19": 4853297.82288451,
    "20": -31.111879695897475,
}

EarlinLutz:
Note that each mesh has a small apron going all the way around, including one where they meet.

The zinger is going down one of the aprons, so I think it is basically correct.   

[...]

PR #5684, opened 2023-06-23
PolyfaceQuery.sweepLineStringToFacets: enhancement to (now-deprecated) PolyfaceQuery.sweepLinestringToFacetsXYReturnLines

EarlinLutz:
New method PolyfaceQuery.sweepLineStringToFacets is an enhancement to behavior of PolyfaceQuery.sweepLinestringToFacetsXYReturnChain and PolyfaceQuery.sweepLinestringToFacetsXYReturnChains.  It provides
  * options to select among outputs on forward, side, and rear-facing facets.
  * an angle tolerance for when facets are considered "side"
  * options to select between segment and chain output.
  * Optional input of sweep direction (other than Z direction draping)
 * Two prior m
* https://github.com/iTwin/itwinjs-core/issues/5507 observes unexpected linework returned from PolyfaceQuery.sweepLinestringToFacetsXYReturnChain.
* for the test case given, the output described as zingers is correct.
  * these lines are on "almost vertical" facets which the caller was unaware of or thought were true vertical.
  * the output is correct
  * a second mesh that has an almost-identical facet which truly is vertical.  No output is generated on that facet.  The code is behaving "properly" at that point -- "Works as Designed" but there should be more consistent treatment so callers can reliably receive outputs for front, side and rear facing facets.

The specific changes introduced are:
 * New method PolyfaceQuery.sweepLineStringToFacets which implements all the options
 * New class SweepLineStringToFacetsOptions which conveys input options described above.
 * Angle methods
   * static Angle.createSmallAngle () -- return an angle object with the system-wide small angle.
   * Angle.isMagnitudeLessThanOrEqual (other) -- angle comparison method.
  * File polyface\multiclip\SweepLineStringToFacetContext.ts is an internal context to support drape methods.
     * Existing context SweepLineStringToFacetContext will someday be deprecated.  It is still used by various other drape methods (e.g. asynch methods that still use it and will continue to exhibit the behavior in 5507.)
     * New context (in same file) ClipSweptLineStringContext implements the new feature "sweep in any direction".
       * per-face classification (forward, side, rear) is implemented in the calling code in PolyfaceQuery.
 

An implementation detail:
* the deprecated methods (and their support context) used this logic for a facet and a segment to drape
  * construct a (skewed) coordinate frame with its UV plane in the polygon plane and W vector in the sweep direction (fixed at global Z.
  * transform both the polygon and the segment-to-drape into that frame
  * do polygon clip in that UV plane.
  * This logic fails when the sweep direction is in the plane of the polygon.
  * Correcting this failure is a main goal of this PR
* The new method uses this logic which does forward, side, and rear polygon clips cleanly
  * create a plane whose normal is perpendicular to (a) the drape segment and (b) the sweep direction
  * intersect the polygon with this plane.

Hence there is a reverse of roles:
  * previously had a "line polygon" clip within a plane
  * now have a "polygon-crosses-plane" clip.


docs/changehistory/NextVersion.md
```diff
@@ -69,3 +71,39 @@ Here are those result sections assembled into `RuledSweep` solids and then facet
 ### Swept surface constructions
 
 The constructors for swept surfaces ([LinearSweep]($core-geometry), [RotationalSweep]($core-geometry), [RuledSweep]($core-geometry)) now allow [CurvePrimitive]($core-geometry) input. Internally, the curve is promoted to a `CurveChain` with one member.
+
+### Draping (sweeping) a linestring onto a mesh
```
pmconne, line 75:
```suggestion
### Sweeping a linestring to facets
```

docs/changehistory/NextVersion.md
```diff
@@ -13,9 +13,11 @@ Table of contents:
   - [Renderer, editor and category on calculated properties](#renderer-editor-and-category-on-calculated-properties)
 - [Electron 25 support](#electron-25-support)
 - [Geometry](#geometry)
-  - [Sweeping a section to a sequence of planes](#sweeping-a-section-to-a-sequence-of-planes)
+  - [Sweeping a section to a sequence of planes]
+  - [Sweeping a linestring to facets]
```
pmconne, line 17:
```suggestion
  - [Sweeping a section to a sequence of planes](#sweeping-a-section-to-a-sequence-of-planes)
  - [Sweeping a linestring to facets](#sweeping-a-linestring-to-facets)
```

pmconne: These are supposed to be links. The first one was a link prior to your edit.

pmconne (changes requested):

docs/changehistory/NextVersion.md
pmconne: Unless you object to proposed changes, you can just use the "Commit suggestion" buttons to apply them.

core/geometry/src/geometry3d/Angle.ts
```diff
@@ -61,6 +61,12 @@ export class Angle implements BeJSONFunctions {
   public static createRadians(radians: number): Angle {
     return new Angle(radians);
   }
+  /**
+   * Return a (new) Angle object with the default "small" angle measurement (Geometry.smallAngleRadians)
```
pmconne, line 65:
```suggestion
   * Return a new `Angle` object with the default "small" angle measurement specified by [[Geometry.smallAngleRadians]].
```

core/geometry/src/geometry3d/Angle.ts
```diff
@@ -329,6 +335,14 @@ export class Angle implements BeJSONFunctions {
     const delta1 = delta - numPeriod * period;
     return Math.abs(delta1) <= radianTol;
   }
+  /**
+ * Test if this angle has magnitude less than other.
```
pmconne, line 339:
```suggestion
 * Test if this angle has magnitude no greater than that of `other`.
```

core/geometry/src/polyface/PolyfaceQuery.ts
```diff
@@ -956,35 +1028,75 @@ export class PolyfaceQuery {
...
+        result = chainContext.collectMaximalChains();
+      }
+    }
+    return result;
+  }
+  /** Find segments (within the linestring) which project to facets.
+    * * Return collected line segments
+    * * this calls  [PolyfaceQuery.sweepLineStringToFacets] with options created by
+    *   `const options = SweepLineStringToFacetsOptions.create(Vector3d.unitZ(), Angle.createSmallAngle(),false, true, true, true);`
+    * @deprecated in 4.x. Use [PolyfaceQuery.sweepLineStringToFacets] to get further options.
```
pmconne, line 1078:
```suggestion
    * @deprecated in 4.x. Use [[PolyfaceQuery.sweepLineStringToFacets]] to get further options.
```

core/geometry/src/polyface/PolyfaceQuery.ts
```diff
@@ -956,35 +1028,75 @@ export class PolyfaceQuery {
...
+    const options = SweepLineStringToFacetsOptions.create(Vector3d.unitZ(), Angle.createSmallAngle(),
+      false, true, true, true);
+    const result = PolyfaceQuery.sweepLineStringToFacets(linestringPoints, polyface, options);
+    return result as LineSegment3d[];
   }
 
   /** Find segments (within the linestring) which project to facets.
    * * Return chains.
-   * * Facets are ASSUMED to be convex and planar, and not overlap in the z direction.
+   * * this calls  [PolyfaceQuery.sweepLineStringToFacets] with options created by
```
pmconne, line 1089:
```suggestion
   * * this calls  [[PolyfaceQuery.sweepLineStringToFacets]] with options created by
```

`[[ApiName]]` to link to an API in the same package.
`[ApiName]($package-alias)` to link to an API in a different package.

pmconne (approved):

dassaf4 (changes requested):
Looks great! I pushed some doc changes and cleaned up some stuff. Please review.

dassaf4 (approved):

PR #5689, opened 2023-06-28
Add `watchForChanges` to `OpenBriefcaseArgs`

kabentley:
This allows readonly connections to a briefcase to "watch" changes to it from other processes. When `OpenBriefcaseArgs.watchForChanges === true`, the briefcase is opened readonly with a "watcher" that restarts the defaultTxn and emits the changed elements/models events as if the changes were made by the readonly connnection.

Note: all the iModels stored in the test/assets directory had not been converted to use WAL mode. They are all updated in this PR to enable it.

imodel-native:  https://github.com/iTwin/imodel-native/tree/sync-readonly-briefcases


pmconne:
New test failed on macOS only. Different from the current constant macOS flakiness.
```
--[ FAILURE: @itwin/core-backend ]------------------[ 3 minutes 0.5 seconds ]--


  407 passing (3m)
  1 failing

  1) IModelWriteTest
       WatchForChanges:

      AssertionError: expected '0x100000001' to equal '0x100000000'
      + expected - actual

      -0x100000001
      +0x100000000
      
      at Context.<anonymous> (src/test/standalone/IModelWrite.test.ts:66:43)
```

kabentley:
> New test failed on macOS only. Different from the current constant macOS flakiness.

This uses an OS file watcher. I increased the timeout in hopes that will help, but i don't know how to make this test non-flaky.

pmconne (approved):

PR #5695, opened 2023-06-29
Upgrade @types/node to 20.3.2

swbsi:
(this was previously blocked by an RPC issue)


swbsi:
> are we supporting node20 now?

not intentionally. The version numbers for the types didn’t seem to match the node versions, so I went with the tip.

I will investigate further…

PR #5697, opened 2023-06-29
use fetch API to access tile cache in the frontend

paulius-valiunas:
this would get rid of CORS preflight requests when requesting tiles from the tile cache


paulius-valiunas:
@wgoehrig @pmconne please review

PR #5699, opened 2023-06-30
Fix issues when multiple viewports are closed down to one remaining

DStradley:
For performance and compatibility reasons, when more than one viewport is open, the webgl canvas is handled differently underneath than when only one viewport is used.  The system checks for the transition (either direction) to deal with this.  Currently there are issues when transitioning back down to a single viewport, and this PR attempts to fix those issues.

https://github.com/iTwin/itwinjs-backlog/issues/760 lists some of the issues, but additional testing found more.  Basic steps for repo are:
- Clone the viewport and make one larger than the other
- Delete one of the two and observe the other (which view is selected when deleting does not seem to matter)

That yields 4 cases that can (and do) behave differently.  Current behavior is:
- Delete larger clone -> original view shrinks smaller than viewport (scaled by diff)
- Delete smaller clone -> original turns black sometimes, is fine others
- Delete smaller original -> clone turns black
- Delete larger original -> clone turns black sometimes, otherwise clone view shrinks smaller than viewport (scaled by diff)

So this was actually two different problems that could be caused in this case, both are solved by the PR.


core/frontend/src/Viewport.ts
```diff
@@ -3556,6 +3557,15 @@ export class ScreenViewport extends Viewport {
...
       this.addChildDiv(this.vpDiv, webglCanvas, 5);
+      // if (this._webglCanvas.width !== this.canvas.width)
+      //   this._webglCanvas.width = this.canvas.width;
+      // if (this._webglCanvas.height !== this.canvas.height)
+      //   this._webglCanvas.height = this.canvas.height;
+      if (this._webglCanvas.width !== this.canvas.width || this._webglCanvas.height !== this.canvas.height) {
+        this._webglCanvas.width = this.canvas.width;
+        this._webglCanvas.height = this.canvas.height;
+        resized2 = true;
+      }
```
DStradley, line 3568: @pmconne I first tried testing the canvas size inside of updateDimensions (called from updateViewRect), by adding to the test that was there.  That fixed the "scaling" but introduced some black view cases.  So I removed that and put the test here and still fixed the scaling and got different results of black view (better). 

That is when I investigated further and found that there are 4 different behavior cases (see PR description).  So I tried different things, since it wasn't making sense to me:

Fix scale of canvas width/height:
 - Delete larger clone -> original is fine
 - Delete smaller clone -> original is fine
 - Delete smaller original -> clone turns black
 - Delete larger original -> clone turns black sometimes, otherwise clone is fine

Fix scale of canvas width/height and test for resized2 also:
 - Delete larger clone -> original turns black
 - Delete smaller clone -> original is fine
 - Delete smaller original -> clone turns black sometimes, is fine others
 - Delete larger original -> clone turns black
 - Test passes

Fix scale of canvas width/height and test (resized && !resized2):
 - Delete larger clone -> original is fine
 - Delete smaller clone -> original is fine
 - Delete smaller original -> clone turns black
 - Delete larger original -> clone is fine
 - Test passes
	
Comment out this.target.onResized() - (disposes fbo):
 - Delete larger clone -> original is fine
 - Delete smaller clone -> original is fine
 - Delete smaller original -> clone is fine
 - Delete larger original -> clone is fine
 - Test fails: 1) ViewManager
 - - should resize fbo properly after dropping a recently-resized viewport:  AssertionError: expected false to be true
tried several combinations of testing

Commenting out onResized (below) makes no sense to me why that would fix the black view, but this is where I am at currently, if you have any ideas, and why I just made this a draft.  (I also don't know why some of the black view cases are  not consistent.)

DStradley: Resolved per our discussion.

DStradley:
@pmconne I implemented your fix for the scaling part of the problem, then made a fix for the black view issue.  The problem here is that we are not really changing the size of anything in this case except for the webglCanvas, which gets changed in rendersToScreen and then we are left with a flag that it was resized.  When it gets around to checking to init the framebuffers, the sizes are still all the same so it skipped some things, causing the black view.  In the normal resize case this should not happen, but here (going down to 1 viewport) it is due to grabbing the webglCanvas and resizing just it to what the target and framebuffers already are, and then calling target.onResized which disposes the framebuffers.
 
So I implemented a forceBufferChange function that could be called for this case and tested, and all 4 cases work now, as well as the original test.

core/frontend/src/Viewport.ts
```diff
@@ -3572,6 +3572,7 @@ export class ScreenViewport extends Viewport {
     if (resized) {
       this.target.onResized();
       this.invalidateController();
+      this.target.forceBufferChange();
```
pmconne, line 3575: Can you do this without having to make Viewport reach into RenderTarget's internals?

DStradley: > Can you do this without having to make Viewport reach into RenderTarget's internals?

I did not see a way to do that since the sizes have not changed and rendersToScreen is the only one who knows that webGLCanvas was resized (and that info is lost when it exits).  This does not happen in the normal resize view cases, just from rendersToScreen.

pmconne: Target.setRenderToScreen calls a function on SceneCompositor that sets SceneCompositor._width and _height to -1?

DStradley: Okay, that seems safe, but it is before updateViewRect and determining that it resized.  We only need it for the toScreen case, and I guess we can assume that it is highly likely that it will change size anyway?
Or did you want to just pass a true/false flag to onResized?
Also, in either case, do you prefer the SceneCompositor.forceBufferChange function I had added to set _width and _height rather than a setting a separate flag?

pmconne: > I guess we can assume

Yeah.

> do you prefer the SceneCompositor.forceBufferChange function I had added to set _width and _height rather than a setting a separate flag?

Yes, I prefer to avoid introducing a new flag.

DStradley: Done

pmconne (approved):

PR #5701, opened 2023-06-30
Add `json` field to Blob container metadata

kabentley:
The previous change removed the ability to add additional metadata to containers. 

But there is now no way to add any additional container-specific properties, and that is essential. A few examples:
- for CloudSqlite containers, we need to store the blockSize. This is necessary to attach the container.
- for ViewStores, there will be metadata about the types of views and whether the container applies to branches.
- for CodeService containers we will store the mapping to external services.
- etc.

I imagine that practically every new containerType usage will want to add some properties to it. I added an optional `json` member that can be used for these purposes. 

I also added back the concept of an owner for a "private" container in `Scope`. We need that for user-private property stores.


MichaelSwigerAtBentley:
Here is the repo for the [storage-container-service-client](https://github.com/iTwin/storage-container-service-client/tree/client).

I have an implementation for the cache already. I'm working on testing it out right now.
Let me know what you think, the cache is implemented in [subClient/RequestCache.ts](https://github.com/iTwin/storage-container-service-client/blob/client/core/src/subClient/RequestCache.ts)

I spoke with Arun about this, and Arun was of the opinion that it should be a separate repo within iTwin, so I created the repo based on that.

Currently changing how the responses are handled + testing the cache.

PR #5709, opened 2023-07-05
@bentley/imodeljs-native 3.7.2 [release/3.7.x]

nick4598:
Includes fix for segfault on 18.16.1, 16.20.1.

### DIFF
--- a/core/frontend/src/tile/map/ImageryProviders/ArcGISImageryProvider.ts
+++ b/core/frontend/src/tile/map/ImageryProviders/ArcGISImageryProvider.ts
@@ -109,7 +109,7 @@ export abstract class ArcGISImageryProvider extends MapLayerImageryProvider {
       response = await  fetch(tmpUrl.toString(), options);
     }
 
-    errorCode = await ArcGisUtilities.checkForResponseErrorCode(response.clone());
+    errorCode = await ArcGisUtilities.checkForResponseErrorCode(response);
 
     if (errorCode !== undefined &&
        (errorCode === ArcGisErrorCode.TokenRequired || errorCode === ArcGisErrorCode.InvalidToken) ) {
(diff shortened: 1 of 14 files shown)


core/frontend/src/tile/map/ImageryProviders/ArcGISImageryProvider.ts
```diff
@@ -131,7 +131,7 @@ export abstract class ArcGISImageryProvider extends MapLayerImageryProvider {
 
         // Make a second attempt with refreshed token
         response = await fetch(urlObj2.toString(), options);
-        errorCode  = await ArcGisUtilities.checkForResponseErrorCode(response.clone());
+        errorCode  = await ArcGisUtilities.checkForResponseErrorCode(response);
```
pmconne, line 134: @mdastous-bentley are you cool with this change?

mdastous-bentley: Since its already cloned inside `ArcGisUtilities.checkForResponseErrorCode`, I dont see how this cause a problem. Unfortunately there is no easy way to have automated tests on this.

aruniverse: correct me if im wrong @nick4598 , but this was causing a leak in the tests

nick4598: Yes, I believe so. Its quite confusing as to why this only started showing up now since this code has existed for quite some time, but it makes a lot of sense that if we double clone the response (each of which produces a 'MessagePort' async resource), we'd end up with some extra MessagePort resources preventing the process from shutting down.

Lets hope windows actually passes first, but it did on my 4.0.x PR with the same fix.

PR #5721, opened 2023-07-09
Shift telemetry APIs to internal

calebmshafer:
The telemetry APIs in the `@itwin/core-telemetry` package were marked as `alpha` during the 3.0 release as the usage was sparse and inconsistent. At the time we were undecided on how to expose this API moving forward and marked the use in `IModelHost` and `IModelApp` as internal to discourage usage until we decided what to do.

The usage still existing today is used as feature tracking, specifically a few AppUI features and when an iModel was opened. However, feature tracking is more of an application level responsibility rather than something the library should do explicitly. Instead of having a set of APIs dedicated to it in the core libraries, the libraries focus on raising the appropriate events to enable applications to easily hook up their own feature tracking.

Given the above stance to stop supporting a telemetry API, we should remove the APIs completely since they are only `internal` and `alpha`. Unfortunately, the reliance of AppUI on them as of the 4.0 release would cause a cascading set of breaking changes that would require consumers to react.

Alternatively, we will mark all of the APIs as `internal` for the duration of 4.0 and remove the API entirely in the 5.0 release.

TODO:
- [ ] Create an Issue to track the appropriate removal in 5.0

### DIFF
--- a/core/backend/src/IModelHost.ts
+++ b/core/backend/src/IModelHost.ts
@@ -15,7 +15,6 @@ import { IModelJsNative, NativeLibrary } from "@bentley/imodeljs-native";
 import { DependenciesConfig, Types as ExtensionTypes } from "@itwin/cloud-agnostic-core";
 import { AccessToken, assert, BeEvent, DbResult, Guid, GuidString, IModelStatus, Logger, Mutable, ProcessDetector } from "@itwin/core-bentley";
 import { AuthorizationClient, BentleyStatus, IModelError, LocalDirName, SessionProps } from "@itwin/core-common";
-import { TelemetryManager } from "@itwin/core-telemetry";
 import { AzureServerStorageBindings } from "@itwin/object-storage-azure";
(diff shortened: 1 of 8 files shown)


core/backend/src/IModelHost.ts
```diff
@@ -265,9 +264,6 @@ export class IModelHost {
   /** The AuthorizationClient used to obtain [AccessToken]($bentley)s. */
   public static authorizationClient?: AuthorizationClient;
 
-  /** @alpha */
-  public static readonly telemetry = new TelemetryManager();
```
aruniverse, line 269: did you also want to remove IModelApp.telemetry? Already marked as internal and doesnt seem to be used

calebmshafer: IModelApp unfortunately is the one that is being used by AppUI so if it is removed they would need to update their minimum core supported.

aruniverse (approved):

PR #5724, opened 2023-07-11
Retry concurrent query if timeout occurs

RohitPtnkr1996:
If a timeout occurs during the execution of a concurrent query, the data returned in the response remains undefined.
The current code wasn't validating this and was trying to read the length of the data array returned, which ended up causing a typerror "Cannot read properties of undefined".
I've added a check to continue with the retries if the timeout scenario occurs.

### DIFF
--- a/core/common/src/ECSqlReader.ts
+++ b/core/common/src/ECSqlReader.ts
@@ -366,13 +366,13 @@ export class ECSqlReader implements AsyncIterableIterator<QueryRowProxy> {
   /**
    *
    */
-  private async runWithRetry(request: DbQueryRequest) {
-    const needRetry = (rs: DbQueryResponse) => (rs.status === DbResponseStatus.Partial || rs.status === DbResponseStatus.QueueFull || rs.status === DbResponseStatus.Timeout) && (rs.data.length === 0);
+  protected async runWithRetry(request: DbQueryRequest) {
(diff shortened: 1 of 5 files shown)


core/common/src/ECSqlReader.ts
```diff
@@ -364,15 +364,15 @@ export class ECSqlReader implements AsyncIterableIterator<QueryRowProxy> {
   }
 
   /**
-   *
+   * @internal
    */
-  private async runWithRetry(request: DbQueryRequest) {
-    const needRetry = (rs: DbQueryResponse) => (rs.status === DbResponseStatus.Partial || rs.status === DbResponseStatus.QueueFull || rs.status === DbResponseStatus.Timeout) && (rs.data.length === 0);
+  protected async runWithRetry(request: DbQueryRequest) {
+    const needRetry = (rs: DbQueryResponse) => (rs.status === DbResponseStatus.Partial || rs.status === DbResponseStatus.QueueFull || rs.status === DbResponseStatus.Timeout) && (typeof(rs.data) === "undefined" || rs.data.length === 0);
```
pmconne, line 370:
```suggestion
    const needRetry = (rs: DbQueryResponse) => (rs.status === DbResponseStatus.Partial || rs.status === DbResponseStatus.QueueFull || rs.status === DbResponseStatus.Timeout) && !rs?.data.length;
```
Just a suggestion to simplify long condition.
If you don't like that, consider replacing `typeof(rs.data) === "undefined"` with `rs.data === undefined`.

PR #5728, opened 2023-07-11
With DEBUG_LEAKS on: Print the initialize callstack of active async resources which may be preventing node from shutting down 

nick4598:
Also removes unnecessary double clone from map-layers-format extension.

### DIFF
--- a/tools/build/src/mocha-reporter/index.ts
+++ b/tools/build/src/mocha-reporter/index.ts
@@ -19,23 +19,7 @@ function setupAsyncHooks() {
   const init = (asyncId: number, type: string, triggerAsyncId: number, _resource: any) => {
     const eid = async_hooks.executionAsyncId(); // (An executionAsyncId() of 0 means that it is being executed from C++ with no JavaScript stack above it.)
     const stack = new Error().stack;
-    asyncResourceStats.set(asyncId, {before: 0, after: 0, type, eid, triggerAsyncId, initStack: stack});
-  };
-  const before = (asyncId: number) => {
-    if (asyncResourceStats.get(asyncId) === undefined) {
-      return;
-    }
-    const entry = asyncResourceStats.get(asyncId);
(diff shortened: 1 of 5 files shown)


tools/build/src/mocha-reporter/index.ts
```diff
@@ -6,12 +6,33 @@
...
+      return;
+    }
+    asyncResourceStats.delete(asyncId);
+  };
 
+  const asyncHook = async_hooks.createHook({init, destroy});
+  asyncHook.enable();
+}
 const fs = require("fs-extra");
+import * as path from "path";
```
aruniverse, line 35: feels weird using both import and require, can we consolidate and just use require here?

tools/build/src/mocha-reporter/index.ts
```diff
@@ -76,6 +97,17 @@ class BentleyMochaReporter extends Spec {
           const wtf = require("wtfnode");
           wtf.setLogger("info", console.error);
           wtf.dump();
+
+          let activeResourcesInfo: string[] = (process as any).getActiveResourcesInfo(); // https://nodejs.dev/en/api/v18/process#processgetactiveresourcesinfo (Not added to @types/node yet I suppose)
```
aruniverse, line 101: youve tried updating to the latest @types/node@18.x ver?

aruniverse (approved):
great work!

tools/build/src/mocha-reporter/index.ts
nick4598: No I didn't, but I just checked definitelyTyped repo and I don't see it mentioned anywhere except for a couple of comments in async_hooks types file.

tools/build/src/mocha-reporter/index.ts
nick4598: Changed it

pmconne (approved):

PR #5731, opened 2023-07-12
Switch to eslint flat config system

yashincontrol:
[eslint-plugin](https://github.com/iTwin/eslint-plugin) has migrated to the new [flat config](https://eslint.org/blog/2022/08/new-config-system-part-2/) system in 4.0.0-dev.39. The goal of this PR is to start using the new config system and update @itwin/eslint-plugin to 4.0.0-dev.44 and eslint to 8.44.0


.vscode/settings.json
```diff
@@ -12,6 +12,7 @@
   "eslint.enable": true,
   "eslint.format.enable": true,
   "eslint.lintTask.enable": true,
+  "eslint.experimental.useFlatConfig": true,
   "eslint.options": {
     "overrideConfigFile": "./package.json",
```
aruniverse, line 17: idk if this option is relevant anymore

core/backend/eslint.config.js
```diff
@@ -0,0 +1,20 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+
+module.exports = [
+  {
+    files: ["**/*.{ts,tsx}"],
```
aruniverse, line 5: `*.tsx` extension is for typescript react components, which shouldnt exist at all in this repo, since the core repo should be just a ui framework agnostic and a pure js library

.vscode/settings.json
aruniverse: why remove ln 21?

anmolshres98: deprecated now: 
image

Additionally, I pulled in this branch and tested on VS Code locally and ESLint was showing errors correctly with those lines removed!

yashincontrol: Thanks, will try it out

aruniverse:
https://github.com/iTwin/itwinjs-core/pull/5461

We might need to fold the changes from this pr in, even though they were never completed, or just disable the rule temporarily

fyi @pmconne

mdastous-bentley (approved):

core/backend/eslint.config.js
```diff
@@ -0,0 +1,20 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+
+module.exports = [
+  {
+    files: ["**/*.ts"],
```
aruniverse, line 5: why are some file globs `**/*.ts` and some `src/**/*.ts` ?
shouldnt they be consistent?

core/backend/eslint.config.js
```diff
@@ -0,0 +1,20 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
```
aruniverse, line 1: does the lint script in the package.json need to specify the file glob as an input

aruniverse:
@anmolshres98 @Jake-Screen plz review

core/backend/eslint.config.js
yashincontrol: Yes, I think we can just go with `**/*.ts`

core/backend/eslint.config.js
Jake-Screen: The glob in the config file specifies what rules apply to what files. The glob in the lint script says which files to lint

yashincontrol: Yes, was struggling to phrase that in my head. Thanks Jake

core/ecschema-metadata/eslint.config.js
```diff
@@ -0,0 +1,32 @@
...
+        {
+          "ignored-barrel-modules": [
+            "./src/ECObjects.ts"
+          ]
+        }
+      ]
+    }
+  },
+  {
+    files: ["*.test.ts",
```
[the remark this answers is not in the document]
yashincontrol: fixed

core/frontend/eslint.config.js
```diff
@@ -0,0 +1,40 @@
...
+          "outputApiFile": false
+        }
+      ]
+    }
+  },
+  {
+    files: [
+      "*.test.ts",
+      "*.test.tsx",
+      "**/test/**/*.ts"],
```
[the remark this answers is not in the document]
yashincontrol: fixed

test-apps/display-test-app/eslint.config.js
```diff
@@ -0,0 +1,25 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+
+const customLanguageOptions = {
+  sourceType: "module",
+  parser: require("@typescript-eslint/parser"),
+  parserOptions: {
+    project: [
+      "./tsconfig.json",
+      "./tsconfig.backend.json",
+      "./tsconfig.buildScripts.json"],
```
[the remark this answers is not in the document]
yashincontrol: fixed

core/ecschema-metadata/eslint.config.js
```diff
@@ -0,0 +1,32 @@
...
+        {
+          "ignored-barrel-modules": [
+            "./src/ECObjects.ts"
+          ]
+        }
+      ]
+    }
+  },
+  {
+    files: [
```
Jake-Screen, line 24: The old configuration from the package.json disables the no-internal-barrel-imports rule in the .tsx files as well. A quick change could be to make the file globs end in .ts*

extensions/test-extension/eslint.config.js
```diff
@@ -0,0 +1,8 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+
+module.exports = [
+  {
+    files: ["**/*.ts"],
```
Jake-Screen, line 5: The config this is replacing references @bentley/imodeljs-recommended, which may be a different config, will have to take a look

core/ecschema-metadata/eslint.config.js
yashincontrol: Yes, but there are no .tsx files in this monorepo I believe. So that's why we got rid of it.

anmolshres98 (changes requested):
You should have a base config file similar to [here](https://github.com/iTwin/appui/pull/401) that has the rule 
```js
{
    files: ["**/test/**/*.ts"],
    rules: {
      "@itwin/no-internal-barrel-imports": "off",
    },
  },
```
As mentioned in [this issue](https://github.com/iTwin/appui/issues/413), that linting error for test files should be turned off. This should pass a lot of the linting errors you are seeing currently. Also import the base config for eslint config files that already have this rule (for example, in `core/ecschema-metadata`.

You should be able to fix the `comma-dangle : Missing trailing comma.` error in this PR as well. You could use the eslint cli to do it for you are use VS Code to do it:
1. Hover over a linting error
2. Click on `Quick Fix`
3. Select `Fix all fixable problems` or one of the variations:
image

yashincontrol:
> You should have a base config file similar to [here](https://github.com/iTwin/appui/pull/401) that has the rule
> 
> ```js
> {
>     files: ["**/test/**/*.ts"],
>     rules: {
>       "@itwin/no-internal-barrel-imports": "off",
>     },
>   },
> ```
> 
> As mentioned in [this issue](https://github.com/iTwin/appui/issues/413), that linting error for test files should be turned off. This should pass a lot of the linting errors you are seeing currently. Also import the base config for eslint config files that already have this rule (for example, in `core/ecschema-metadata`.
> 
> You should be able to fix the `comma-dangle : Missing trailing comma.` error in this PR as well. You could use the eslint cli to do it for you are use VS Code to do it:
> 
> 1. Hover over a linting error
> 2. Click on `Quick Fix`
> 3. Select `Fix all auto-fixable problems` or one of the variations:
>    image

Thanks, will make the changes

.vscode/settings.json
```diff
@@ -12,12 +12,7 @@
   "eslint.enable": true,
   "eslint.format.enable": true,
   "eslint.lintTask.enable": true,
-  "eslint.options": {
-    "overrideConfigFile": "./package.json",
-    "resolvePluginsRelativeTo": "../../common/temp/node_modules/.pnpm/@itwin+eslint-plugin*/node_modules",
-    "useEslintrc": false
-  },
-  "eslint.packageManager": "pnpm",
+  "eslint.experimental.useFlatConfig": true,
```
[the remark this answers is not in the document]
yashincontrol: Yes, it's working

core/backend/eslint.config.js
```diff
@@ -0,0 +1,23 @@
...
+  },
+  {
+    files: ["**/*.ts"],
+    rules: {
+      "@typescript-eslint/no-non-null-assertion": "error"
+    }
+  },
+  {
+    files: [
+      "*.test.ts",
```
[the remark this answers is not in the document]
yashincontrol: Yes, nice catch. Fixed

core/ecschema-metadata/eslint.config.js
```diff
@@ -0,0 +1,32 @@
...
+      "*.test.ts",
+      "**/test/**/*.ts"
```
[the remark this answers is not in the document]
yashincontrol: Fixed

core/frontend/eslint.config.js
```diff
@@ -0,0 +1,40 @@
...
+          ],
+          "outputApiFile": false
+        }
+      ]
+    }
+  },
+  {
+    files: [
+      "*.test.ts",
+      "**/test/**/*.ts"
```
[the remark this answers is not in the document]
yashincontrol: fixed

extensions/test-extension/eslint.config.js
Jake-Screen: You were correct in the change to use itwinjs-recommended config, renamed in the past it appears

ui/appui-abstract/eslint.config.js
```diff
@@ -0,0 +1,24 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+
+module.exports = [
+  {
+    files: ["**/*.ts"],
+    ...iTwinPlugin.configs.iTwinjsRecommendedConfig,
+  },
+  {
+    files: ["**/*.ts"],
+    ...iTwinPlugin.configs.jsdocConfig,
```
anmolshres98, line 10: @Jake-Screen Should `uiConfig` also be imported for this package?

anmolshres98:
In my most recent commits I have:
1. Implemented a base eslint config in `common/config/eslint`(has a single rule that turns `@itwin/no-barrel-imports` rule off for files in `test` directory; rationale [here](https://github.com/iTwin/appui/issues/413)) which is imported in other applicable `eslint.config.js` files. This fixes the no-barrel-imports linting error that was being seen
2. Went through the documentation for the other error we were seeing: `@typescript-eslint/unbound-method` (Explanation of error [here](https://typescript-eslint.io/rules/unbound-method/)). I went through all the places the error was being shown then, manually verified that none of the methods being passed were accessing `this` inside the passed methods. This was the case for one location: `core/backend/src/BriefCaseManager.ts`, line 124, but this can also be safely suppressed since the method being passed is a static method in a static class and the the `this._initialized` being accessed in the passed method is also a static field, hence there should be no ambiguity on what the `this` is referring to inside the passed method.

These changes should result in `rush lint` passing safely with no additional work required.

I have chosen to disable each line(s) manually with the `// eslint-disable-next-line @typescript-eslint/unbound-method` comment instead of disabling the rule entirely inside the `eslint.config.js` of each package because we would want the linting error to be disabled on a case-by-case basis after close examination instead of disabling the rule entirely for entire packages, which might very well lead to valid errors being ignored leading to undefined behaviors in the future.

core/backend/src/BriefcaseManager.ts
```diff
@@ -119,6 +119,8 @@ export class BriefcaseManager {
     if (this._initialized)
       return;
     this.setupCacheDir(cacheRootDir);
+    // disable this rule here since even though the method accesses "this", it is a static method accessing a static field of this class
+    // eslint-disable-next-line @typescript-eslint/unbound-method
     IModelHost.onBeforeShutdown.addOnce(this.finalize, this);
```
pmconne, line 124:
```suggestion
    IModelHost.onBeforeShutdown.addOnce(() => this.finalize());
```
> it is a static method accessing a static field

`static` is irrelevant. `static` methods and field are just as vulnerable to unbound method errors as instance methods and fields.
In this case you are actually relying on the fact that `finalize` is currently `private`, so a subclass cannot override it - but that could change, or the code inside `finalize` could change.

core/backend/src/IModelHost.ts
```diff
@@ -491,6 +491,7 @@ export class IModelHost {
     this.configuration = options;
     this.setupTileCache();
 
+    // eslint-disable-next-line @typescript-eslint/unbound-method
     process.once("beforeExit", IModelHost.shutdown);
```
pmconne, line 495:
```suggestion
    process.once("beforeExit", () => IModelHost.shutdown());
```

I don't understand why you would go to the trouble of disabling this rule all over the place instead of just conforming to it.

ui/appui-abstract/eslint.config.js
Jake-Screen: Since this is not a test app i do not believe so

core/backend/src/IModelHost.ts
anmolshres98: I didn't wanna inadvertently break the tests or anything so I just took the disabling route when I started, I can push the confirming version instead. Thanks.

dassaf4 (approved):
```
foo (
  "The linter adds rules",
  "I am drowning in commas",
  "Ev'rything is fine",
);
```
😄 What is the reason to force a dangling comma for multi-line arg lists?

ben-polinsky:
> ```
> foo (
>   "The linter adds rules",
>   "I am drowning in commas",
>   "Ev'rything is fine",
> );
> ```
> 
> 😄 What is the reason to force a dangling comma for multi-line arg lists?

debatable, but: 

> Trailing commas simplify adding and removing items to objects and arrays, since only the lines you are modifying must be touched. Another argument in favor of trailing commas is that it improves the clarity of diffs when an item is added or removed from an object or array:

https://eslint.org/docs/latest/rules/comma-dangle

pmconne:
> debatable

[Prettier](https://prettier.io/docs/en/why-prettier.html) can put an end to that.

core/backend/src/IModelHost.ts
anmolshres98: done

anmolshres98 (approved):

ben-polinsky (approved):

pmconne (approved):
I don't love all the noise introduced by comma pollution, I presume that was somehow unavoidable?
Note: I don't care about dangling comma vs no dangling comma, I just don't like noisy formatting changes.

dassaf4:
> I don't love all the noise introduced by comma pollution, I presume that was somehow unavoidable?

If there were a dangling comma rule setting to avoid functions, the delta in core-geometry would be vastly diminished.

core/backend/eslint.config.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = [
+  {
+    files: ["**/*.ts"],
+    ...iTwinPlugin.configs.iTwinjsRecommendedConfig,
+  },
+  {
+    files: ["**/*.ts"],
+    rules: {
+      "@typescript-eslint/no-non-null-assertion": "error"
```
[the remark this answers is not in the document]
Jake-Screen: It is set to off in the default config

raplemie (approved):
Reviewed Appui-abstract only for core-ui

ColinKerr (approved):

bbastings (approved):

PR #5733, opened 2023-07-12
@bentley/imodeljs-native 4.1.5

nick4598:


aruniverse (approved):
lgtm;

any idea whats going on with the line ending changes?

nick4598:
> lgtm;
> 
> any idea whats going on with the line ending changes?

Not really, but I checked out master's copy again and then kept the 4.1.5 for backend. Should look better now

aruniverse:
> > lgtm;
> > any idea whats going on with the line ending changes?
> 
> Not really, but I checked out master's copy again and then kept the 4.1.5 for backend. Should look better now, atleast for the package.jsons.
> 
> Maybe something different between line endings? These are all files I touched as part of a merge.. mostly "accepting current" changes.

looks much better

nick4598:
@akshay05n In your original PR into imodel02, you marked the functions in SchemaUtils.ts as both @public and @beta, im guessing it should be @beta? Locally running extract-api though we get an error 

"Error: src/SchemaUtils.ts:20:1 - (ae-incompatible-release-tags) The symbol "convertEC2SchemasToEC3Schemas" is marked as @beta, but its signature references "ECSchemaXmlContext" which is marked as @internal
Error: src/SchemaUtils.ts:35:1 - (ae-incompatible-release-tags) The symbol "upgradeCustomAttributesToEC3" is marked as @beta, but its signature references "ECSchemaXmlContext" which is marked as @internal
"

So maybe ECSchemaXmlContext needs to be promoted to beta?

Theres also another error which I'm not sure if its related or not:
 "Topic SchemaUtils does not have a @docs-group-description in core-backend.ts. Please add a description."
 
 I decided to not add a SchemaUtils topic and thought I could just add Module Schema to the top of SchemaUtils.ts... but maybe @ben-polinsky could help out here.

core/backend/src/SchemaUtils.ts
```diff
@@ -3,15 +3,14 @@
...
-/** @packageDocumentation
- * @module Schema
- */
-
 import { IModelJsNative } from "@bentley/imodeljs-native";
 import { BentleyStatus } from "@itwin/core-bentley";
 import { IModelError } from "@itwin/core-common";
 import { ECSchemaXmlContext } from "./ECSchemaXmlContext";
 
+/** @packageDocumentation
```
ben-polinsky, line 11: You were correct before, it should go above imports. Perhaps with your export added to the barrel file, and this reverted, it will work.

nick4598: Ah, I looked at the one other file in this repo which places it below the imports.. Thanks

ben-polinsky: We'll see. I'm running locally as well...

PR #5737, opened 2023-07-14
Support glTF-encoded point clouds

pmconne:
The Orbit team is transitioning to producing standard 3D Tiles. Upgrade GltfReader to deserialize point clouds encoded as glTF (vs deprecated PNTS tile format).
GltfReader is an untestable mess full of special cases and legacy cruft. We need to:
- Separate parsing from graphics creation
  - This will allow decoding in a web worker.
  - It will also allow us to discriminate between querying the geometry (e.g., TileGeometryCollector, draping, etc) and displaying it.
  - It will also make it much more testable.
 - Better conform to the specification.
 - Make it more easily extensible as we expand our spec compliance and as the spec evolves.

I went back and forth on trying to implement some of that and ultimately ended up just shoehorning in the new capability, to revisit the above separately.


pmconne:
> Is this something easy to add a unit test for?

I wanted to (and intend to in the future but:

> GltfReader is an untestable mess

pmconne:

PR #5739, opened 2023-07-14
Fix Viewport.isPixelSelectable following signature change of Viewport.mapLayerFromIds

mdastous-bentley:


mdastous-bentley:
@bbastings can you confirm that solves your tools issues?

PR #5742, opened 2023-07-16
Add support for ECSQL expression tree

khanaffan:
https://github.com/iTwin/itwinjs-backlog/issues/803


pmconne:
Why `alpha`? Is there a reason you don't want anybody to use this?

khanaffan:
> Why `alpha`? Is there a reason you don't want anybody to use this?

This API is at an experimental stage where other teams like to use it to perform their own experimental work. As we get feedback from them, we will improve or expand these APIs and move them up from alpha.

common/api/core-common.api.md
```diff
@@ -666,6 +699,115 @@ export { BentleyError }
...
+    // (undocumented)
+    readonly rhsExpr: ComputedExpr;
+    // (undocumented)
+    static readonly type = ExprType.BinaryBoolean;
+    // (undocumented)
+    writeTo(writer: ECSqlWriter): void;
+}
+
+// @alpha
+export enum BinaryBooleanOp {
```
pmconne, line 747: Did you consider just using a union for readability?

`if (op === BinaryBooleanOp.And || op === BinaryBooleanOp.GreaterThanOrEqualTo)`

vs

`if (op === "AND" || op == ">=")`

pmconne (commented):
This is a big, niche API. Did you consider putting it into a separate package?

khanaffan:
> This is a big, niche API. Did you consider putting it into a separate package?

No, not yet. We have two users
1. Design Review - Named Group filter: This will be a UI, but we like them to standardize how they store their filter and use ECSQL instead of JSON from which they create an ECSQL.
2. ECPresentation: They are interested in it in the context of their initiative where they like to move ECPresentation native logic to typescript.

But we can extract it into a package if needed when API are stable.

pmconne:
> we can extract it into a package if needed when API are stable.

Perhaps it belongs in the existing ecschema-metadata packages?

khanaffan:
@pmconne, so I will take a different approach and move this code change into its package `<root>/misc/ecsql-expr`.  I will remove the RPC and native addon change. And we get the JSON for ECSQL parse tree via ecsql `PRAGMA parse_tree(<ecsql>)`. This way we do not need any peer dependency on the backend or frontend dependency. Instead, any frontend or backend code can reference ecsql-expr and use available api to run ECSQL which is `ECSQLRreader` to populate the parse tree.

pmconne:
> move this code change into its package <root>/misc/ecsql-expr

Ok. Please talk to @aruniverse about repository organization.

aruniverse:
> > move this code change into its package /misc/ecsql-expr
> 
> Ok. Please talk to @aruniverse about repository organization.

anyone oppose: /core/ecsql-ast ?
pkg name being `itwin/core-ecsql-ast`

im indifferent on the `core-` prefix

cc @pmconne @calebmshafer @wgoehrig @ColinKerr  thoughts on naming?

MichaelBelousov:
I vote `@itwin/[core-]ecsql/lib/Parser.ts`. The API seems similar enough to that exposed by other parsers which give you a traversable AST. It will be a good place to put any future non-AST semantic analysis that might ever be exposed, while also already not being too small of a utility to not deserve the generic `@itwin/ecsql` name.

pmconne:
> `@itwin/[core-]ecsql[/lib/Parser.ts]`

I can't tell what you are actually proposing here.

MichaelBelousov:
> I can't tell what you are actually proposing here.
edited for (hopefully) clarity

pmconne:
> I vote @itwin/ecsql or @itwin/core-ecsql, with a class for the parser. The API seems similar enough to that exposed by other parsers packages which give you a traversable AST. It will be a good place to put any future non-AST semantic analysis that might ever be exposed

I'd be fine with `core-ecsql` except, if the API expands in future, it may require backend and frontend packages. @khanaffan's current proposal is for a package that contains APIs usable by frontend **and** backend.
If that's expected I would go with something like `ecsql-common` which can be joined down the road by `ecsql-backend` and `ecsql-frontend`.
Alternatively, if @khanaffan prefers to limit this package to just the AST, I'd name it accordingly.

MichaelBelousov:
> @khanaffan prefers to limit this package to just the AST, I'd name it accordingly, e.g., `ecsql-ast`.

Yes you're right, better to call it `@itwin/ecsql-common` if it works frontend+backend, and eventually as necessary add corresponding `-frontend`/`-backend` specific packages. But I don't want to end up with another (3?) package(s) if we decide to expose something non-ast related in the future, so I would vote against `@itwin/ecsql-ast`. I think this would belong fine in a more general `ecsql` package (set)

aruniverse:
sounds like we are in agreement with `itwin/ecsql-common` then

khanaffan:
> sounds like we are in agreement with `itwin/ecsql-common` then

Where should I place in the folder hierarchy in mono repo? `<root>/ecsql/ecsql-common`

aruniverse:
> > sounds like we are in agreement with `itwin/ecsql-common` then
> 
> Where should I place in the folder hierarchy in mono repo? `<root>/ecsql/ecsql-common`

`~/core/ecsql/common/`

.github/CODEOWNERS
```diff
@@ -36,6 +36,7 @@ rush.json       @iTwin/itwinjs-core-admins
 
 /common/**/*-extension/                       @aruniverse @calebmshafer
 
+/common/**/ecsql-*/                           @iTwin/itwinjs-core-admins @khanaffan @ColinKerr
```
aruniverse, line 39: you will want to add an entry around LN 90 for owning the actual source code
`/core/ecsql/*`

core/ecsql/common/LICENSE.md
```diff
@@ -0,0 +1,9 @@
+# MIT License
+
+Copyright © 2017-2023 Bentley Systems, Incorporated. All rights reserved.
```
aruniverse, line 3:
```suggestion
Copyright © 2023 Bentley Systems, Incorporated. All rights reserved.
```

core/ecsql/common/ThirdPartyNotices.md
aruniverse, line 1: can delete this file, we dont actually use this pkg

core/ecsql/common/extraction.eslint.config.js
aruniverse, line 1: can delete this file; since we're not exposing this pkg via the extension api

core/ecsql/common/eslint.config.js
```diff
@@ -0,0 +1,25 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+const eslintBaseConfig = require("../../../common/config/eslint/eslint.config.base");
+
+module.exports = [
+  {
+    files: ["**/*.ts"],
+    ...iTwinPlugin.configs.iTwinjsRecommendedConfig,
+  },
+  {
+    files: ["**/*.ts"],
```
aruniverse, line 10: can delete this extension config

core/backend/src/ECDb.ts
aruniverse, line 1: can revert changes to this file

full-stack-tests/core/package.json
```diff
@@ -35,6 +35,7 @@
     "@itwin/core-frontend": "workspace:*",
     "@itwin/core-geometry": "workspace:*",
     "@itwin/core-markup": "workspace:*",
+    "@itwin/ecsql-common": "workspace:*",
```
aruniverse, line 38: mind alphabetizing?

core/ecsql/common/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "digital-twin",
+    "iTwin",
+    "ecsql"
+  ],
+  "author": {
+    "name": "Bentley Systems, Inc.",
+    "url": "http://www.bentley.com"
+  },
+  "dependencies": {
+    "flatbuffers": "~1.12.0",
```
aruniverse, line 38: rm dep on flatbuufers, not used anywhere

core/ecsql/common/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "js-base64": "^3.6.1"
```
aruniverse, line 39: rm dep on js-base64, not used anywhere

core/ecsql/common/package.json
```diff
@@ -0,0 +1,61 @@
...
+  },
+  "peerDependencies": {
+    "@itwin/core-bentley": "workspace:^4.2.0-dev.1"
```
aruniverse, line 42: i believe the current usage of core-bentley is pure/safe, so doesnt need to be a dev+peer, can be a direct dep

core/ecsql/common/eslint.config.js
aruniverse: dont forget this;
just delete the second object from this array we are exporting

aruniverse (approved):

aruniverse (approved):

PR #5574, opened 2023-05-29
Convert EC2 XML schemas using native API

akshay05n:
Calls the native API to convert EC 2 schemas to EC 3.2 schemas

imodel-native: https://github.com/iTwin/imodel-native/tree/aks/convert-ec2-schemas

PR #5744, opened 2023-07-17
Update release tag of ECSchemaXmlContext class

akshay05n:


core/backend/src/ECSchemaXmlContext.ts
```diff
@@ -30,18 +32,34 @@ export class ECSchemaXmlContext {
     return this._nativeContext;
   }
```
grigasp, line 33: Just making sure you really want `nativeContext` to be `@public`.

pmconne: Definitely not, it needs to remain `internal`.

pmconne: Is there any reason for the public accessor to exist? Shouldn't the TypeScript `ECSchemaXmlContext` APi provide all the functionality any consumers could need?

akshay05n: The getter method is required by the IModelDb class and SchemaUtils.ts file, so it needs a public accessor. However, I will change the release tag to `internal` because we do not want to expose something from the native package.

core/backend/src/ECSchemaXmlContext.ts
```diff
@@ -17,31 +17,52 @@ export type SchemaKey = IModelJsNative.ECSchemaXmlContext.SchemaKey;
 /** @internal */
 export type SchemaMatchType = IModelJsNative.ECSchemaXmlContext.SchemaMatchType;
 
-/** @internal */
+/** Context for deserializing ECSchemas
```
pmconne, line 20: Can you be more specific about what it's for, how it's used? e.g., link to other APIs that require one of these contexts?

core/backend/src/ECSchemaXmlContext.ts
```diff
@@ -17,31 +17,52 @@ export type SchemaKey = IModelJsNative.ECSchemaXmlContext.SchemaKey;
...
   }
 
+  /** @internal */
   public get nativeContext(): IModelJsNative.ECSchemaXmlContext {
     assert(undefined !== this._nativeContext);
     return this._nativeContext;
   }
 
+  /**
+   * Adds a file path that should be used to search for a matching schema name
```
pmconne, line 37: Please specify in the docs the order in which paths are searched if I call addSchemaPath multiple times. Same for schema locators.

core/backend/src/ECSchemaXmlContext.ts
```diff
@@ -17,31 +17,52 @@ export type SchemaKey = IModelJsNative.ECSchemaXmlContext.SchemaKey;
...
+   * Set the last locater to be used when trying to find a schema
+   * @param locater Locater that should be used as the last locater when trying to find a schema
+   * @internal
+   */
   public setSchemaLocater(locater: IModelJsNative.ECSchemaXmlContext.SchemaLocaterCallback): void {
     this.nativeContext.setSchemaLocater(locater);
   }
 
+  /**
+   * Adds a schema locater as first to the current context
```
pmconne, line 54: I assume "first" means "searched for schema matches before other locaters". Can you rephrase the docs so I don't have to make assumptions?

core/backend/src/ECSchemaXmlContext.ts
pmconne: I did some cleanup on your docs. Your code example used an `internal` API (one of many reasons to put code examples into example-code-snippets package). Resolve this comment if you're okay with my changes.

pmconne (approved):

pmconne (changes requested):
Why is this targeting the `imodel02` branch instead of `master`?

pmconne:
> I created this branch from imodel02 since https://github.com/iTwin/itwinjs-core/pull/5574 started causing PR validation failures.

No validation is happening for your PR because of its target branch and lack of corresponding native changes.
Since all you are changing is documentation and release tags, I think you should target master.

pmconne (approved):

PR #5746, opened 2023-07-17
finalize release by automating changelogs

DanRod1999:
JS script that can be used to automate the updating changelogs process.
This script works locally, and in the workflow we already have.
The script finds any change log versions that don't exist inside master or the current release/X.X.x branch, and brings those changes in from the recently released release/X.X.x branch.


.github/workflows/automation-scripts/update-changelogs.js
```diff
@@ -0,0 +1,91 @@
...
+
+  currentEntries.forEach((entry) => {
+    currVersions.push(entry.version);
+  });
+
+  incomingEntries.forEach((entry) => {
+    incomingVersions.push(entry.version);
+  });
+
+  newEntries = [];
```
aruniverse, line 45: this is supposed to be just a set difference right?

.github/workflows/automation-scripts/update-changelogs.js
```diff
@@ -0,0 +1,91 @@
...
+    if (!compareVersions(currentEntries[i].version, newEntries[j].version)) {
+      currentEntries.splice(i, 0, newEntries[j])
+      j++;
+    }
+    i++;
+  }
+  return currentEntries;
+}
+
+function fixChangeLogs(currentFiles, incomingFiles) {
```
aruniverse, line 67: feel like algorithmically there is an easier way to do this

you have 2 arrays, and you basically want the union of them, where the union contains only unique values and is sorted

.github/workflows/finalize-release.yaml
```diff
@@ -13,48 +13,20 @@ on:
...
-            exit $exitCode
-          fi
-
-          rush change --bulk --message "" --bump-type none
-          git add .
-          git commit --amend --no-edit
-
-          exit $?
+      - name: Run update-changelogs.sh
+        run: ./.github/workflows/automation-scripts/update-changelogs.sh
```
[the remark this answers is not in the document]
DanRod1999: i did this just to make it easier to run the git and awk cmds. I don't know how js runs shell cmds, but if that would be easier I could create a main function in the js script and run everything from there. Does that seem more appropriate?

.github/workflows/automation-scripts/update-changelogs.js
DanRod1999: yes this is all the that exist in the incoming branch, but do not exist in the target branch. Note that in the case of a backported version like release/3.7 this branch is missing all the entries from 4.x.x, so this is why it is only looking for elements that are not in target.

.github/workflows/automation-scripts/update-changelogs.js
DanRod1999: I didn't see a union functionality from JS except for in some other libraries, I don't know JS all that well so i could be wrong. I kind of just implemented my own, and I agree its not very simple, but its fairly efficient, at least compared to the other simpler ways I could think of writing it.

.github/workflows/automation-scripts/update-changelogs.js
```diff
@@ -0,0 +1,91 @@
...
+  })
+  const leadNum = currentDif.filter((num) => { return num !== 0 })[0];
+  return leadNum > 0;
+}
+
+function findNewEntries(currentEntries, incomingEntries) {
+  currVersions = [];
+  incomingVersions = [];
+
+  currentEntries.forEach((entry) => {
```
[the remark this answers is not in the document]
DanRod1999: yeah this is a better implementation thanks.

.github/workflows/automation-scripts/update-changelogs.js
ben-polinsky: Perhaps a merge and then sort? [...currentEntries, ...newEntries].sort(customSortFunction)

ben-polinsky: Ah I see they're objects not version strings...

.github/workflows/automation-scripts/update-changelogs.js
```diff
@@ -0,0 +1,91 @@
...
+      j++;
+    }
+    i++;
+  }
+  return currentEntries;
+}
+
+function fixChangeLogs(currentFiles, incomingFiles) {
+  numFiles = currentFiles.length;
+  for (i = 0; i < numFiles; i++) {
```
ben-polinsky, line 69: Missing a `let` before the `i`.

.github/workflows/automation-scripts/update-changelogs.js
```diff
@@ -0,0 +1,91 @@
...
+    let currentJson = loadJsonFiles(currentFiles[i]);
```
ben-polinsky, line 70: This can be `const` - const object properties can be reassigned, just not the object itself.

.github/workflows/automation-scripts/update-changelogs.js
aruniverse: yea i saw a bunch of these were missing

aruniverse: yea i saw a bunch of these were missing

aruniverse:
enable strict mode for the script, it mightve caught a bunch of mistakes

.github/workflows/finalize-release.yaml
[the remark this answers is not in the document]
ben-polinsky: Was thinking about this as well. I understand the reason for the separation, but at the same time, one script is much more appealing. I'd try to combine them one way or another.

.github/workflows/automation-scripts/update-changelogs.js
DanRod1999: gotcha thanks for the clarification

.github/workflows/automation-scripts/update-changelogs.js
DanRod1999: yeah, I was just avoiding a custom sort function as I didn't know how to implement it off the top of my head, and figured I could write this faster. I'll try this and get something as simple as possible

DanRod1999: Used this method, and tested with the new change logs for 3.7.12, looks good. Still going to clean up some of the syntax, and try to condense into one file

.github/workflows/automation-scripts/update-changelogs.js
```diff
@@ -0,0 +1,83 @@
...
+    for (let i = 0; i < 3; i++) {
+      if (versionA[i] < versionB[i]) return 1;
+      if (versionA[i] > versionB[i]) return -1;
+    }
+
+    return 0;
+  });
+}
+
+function findNewEntries(currentEntries, incomingEntries) {
```
aruniverse, line 36: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
https://javascript.info/map-set#set

aruniverse (changes requested):

.github/workflows/automation-scripts/update-changelogs.js
DanRod1999: using a set seems to not work, don't know why. It doesn't seem to identify the entries as duplicates, and treats them all as unique. I'll keep looking into it

DanRod1999: done

.github/workflows/automation-scripts/update-changelogs.js
DanRod1999: done

.github/workflows/automation-scripts/update-changelogs.js
DanRod1999: i believe i caught the rest

.github/workflows/finalize-release.yaml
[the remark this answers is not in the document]
DanRod1999: after thinking about this some more I personally think it is better to use the shell cmds in the script to run the git cmds and file manipulations. So my secondary solution would be to inline the sh file into the github workflow, then call the js file inside the inlined shell script. This way we only create one js script, and no need for the extra .sh file. Does this sounds like a better solution?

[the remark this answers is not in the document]
ben-polinsky: For me this comes down to does it make sense to write a bunch of shell commands in JS when the shell is right there for us to use. I don't think it does. (And JSON manipulation in the shell doesn't seem particularly appealing.) I'd rather have the extra file (not-inlined) and maybe throw both in a `changelogs/` folder inside of `automation-scripts`? 

I'm easily swayed on matters like this, though.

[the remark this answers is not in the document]
DanRod1999: > Please elaborate on why.

it is probably due to my inexperience with JS, and i am biased because when I need to do any work with file manipulation my first thought is always shell cmds with grep/awk/sed ect. In terms of efficiency and lines of code (like I said, this could be me not knowing a easy way to do this in JS) but finding all changelog files, copying them to a temp dir, changing the file names so they don't overwrite each other, then copying them back to the proper locations, seemed like it would require writing some fairly big functions, where as shell commands could do it in 3, and I also trust the performance of that much more then whatever I would try to do in JS.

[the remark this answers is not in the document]
DanRod1999: Initial tests look good. Even somehow managed to find some duplicates that must've slipped in back in 2.x i think. However, I would like to test it on a real patch locally before i'm confident in its functionality. However, should be good to review. Thanks Arun for that package, works wonders it seems

DanRod1999:
fyi @anmolshres98 @hl662

.github/workflows/finalize-release.yaml
```diff
@@ -13,48 +13,22 @@ on:
...
-            exit $exitCode
-          fi
-
-          rush change --bulk --message "" --bump-type none
-          git add .
-          git commit --amend --no-edit
-
-          exit $?
+          npm i -g zx
+          node ./.github/workflows/automation-scripts/update-changelogs.mjs
```
anmolshres98, line 29: typo? `update-changelogs.mjs` ?

DanRod1999: ah no. the[ zx package](https://github.com/google/zx) calls for .mjs files, not js. So i actually have a typo in the filename. Thanks

.github/workflows/automation-scripts/update-changelogs.mjs
```diff
@@ -0,0 +1,115 @@
...
+* git checkout target branch (master or latest release); git pull
+* git checkout release/X.X.x; git pull
+* uncomment git checkout -b cmd and fix branch name
+* open PR into target branch
+*****************************************************************/
+
+const targetPath = "./temp-target-changelogs"
+const incomingPath = "./temp-incoming-changelogs"
+
+await $`mkdir ${targetPath}`
```
ben-polinsky, line 19: A comment about zx might be helpful. People may be unfamiliar with the tool.

DanRod1999: done

.github/workflows/automation-scripts/update-changelogs.mjs
```diff
@@ -0,0 +1,120 @@
...
+// Uncomment For Manual runs and fix branch name to appropriate version
+// the version should match your incoming branch
+// await $`git checkout -b finalize-release-X.X.X`;
+/*********************************************************************/
+await $`git add .`;
+await $`git commit -m "${commitMessage} Changelogs"`;
+await $`rush change --bulk --message "" --bump-type none`;
+await $`git add .`;
+await $`git commit --amend --no-edit`;
+await $`git push https://$(GITHUBTOKEN)@github.com/iTwin/itwinjs-core HEAD:${targetBranch}`
```
DanRod1999, line 70: These are the only lines i didn't test (67-70). I got a successful run in the workflow, but i had to modify the script to accommodate testing without the correct source and target branches. Assuming this git push command works as intended this should be full functional. I did a regular git push when tested and excluded the two args. I don't fully know if this cmd is correct, but it is how it was initially written for the original workflow, so that is what I'm basing it of off.

.github/workflows/finalize-release.yaml
```diff
@@ -13,48 +13,20 @@ on:
...
-
-          exit $?
+      - name: Run update-changelogs.sh
+        run: ./.github/workflows/automation-scripts/update-changelogs.sh
         env:
-          commitid: ${{ github.sha }}
+          commitId: ${{ github.sha }}
         # turn off the default behavior to fail fast. The script has its own error handling
         shell: bash {0}
       - name: Push to master
```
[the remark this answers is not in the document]
DanRod1999: removed

.github/workflows/finalize-release.yaml
```diff
@@ -13,49 +13,22 @@ on:
 
 jobs:
   finalize:
-    name: Cherry-pick Changelogs to master
+    name: Update master Changelogs
```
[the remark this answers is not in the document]
DanRod1999: also master is not really valid anymore, since technically it may not update master

.github/workflows/finalize-release.yaml
```diff
@@ -13,49 +13,22 @@ on:
...
       - uses: actions/checkout@v2
         with:
-          ref: master
           fetch-depth: 0
       - name: Set Git Config
         run: |
           git config --local user.email 38288322+imodeljs-admin@users.noreply.github.com
           git config --local user.name imodeljs-admin
-      - name: Cherry-pick changelogs
+      - name: Run update-changelogs.sh
```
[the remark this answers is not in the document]
DanRod1999: done

.github/workflows/automation-scripts/update-changelogs.mjs
```diff
@@ -0,0 +1,120 @@
...
+
+targetBranch = String(targetBranch).slice(0, -1);
+currentBranch = String(currentBranch).slice(0, -1);
+commitMessage = String(commitMessage).slice(0, -2);
+
+if (targetBranch === `origin/${currentBranch}`) {
+  console.log("The current branch is the latest release, so the target will be master branch")
+  targetBranch = 'master'
+} else
+  console.log(`The current branch is ${currentBranch}, so the target will be ${targetBranch} branch`)
```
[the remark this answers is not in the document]
DanRod1999: done

.github/workflows/finalize-release.yaml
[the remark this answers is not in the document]
DanRod1999: done

.github/workflows/automation-scripts/update-changelogs.mjs
DanRod1999: I assumed it was a var stored in the workflow, but I can't find a secret vars section inside GH Actions. This was how the workflow was written before, but technically don't know if it ever worked since it would break before reaching this. I'll try to run a quick test

aruniverse (approved):

.github/workflows/automation-scripts/update-changelogs.mjs
DanRod1999: Yeah i still don't really know where that var comes from but the cmd does work. I'll still try to figure that out though. And sorry wasn't thinking and accidentally requested core-admin review when making a dummy push

PR #5755, opened 2023-07-19
Saeed torabi/offset sandbox

saeeedtorabi:
Made changes related to the offset sandbox


core/geometry/src/geometry3d/Plane3dByOriginAndUnitNormal.ts
```diff
@@ -255,52 +281,56 @@ export class Plane3dByOriginAndUnitNormal extends Plane3d implements BeJSONFunct
...
+  /** Return (a clone of) the unit normal. */
   public override getUnitNormal(result?: Vector3d): Vector3d | undefined {
     return this._normal.clone(result);
   }
-  /**
-   * Return (a clone of) the origin.
-   */
+  /** Return (a clone of) the origin. */
   public override getAnyPointOnPlane(result?: Point3d): Point3d {
+    // This function returns origin but in general A Point x is on plane if and only if (x-o).n = 0.
```
dassaf4, line 309:
```suggestion
    // This function returns the plane origin. In general, a point x is on the plane if and only if (x-o).n = 0.
```

core/geometry/src/geometry3d/Plane3dByOriginAndVectors.ts
```diff
@@ -297,6 +297,8 @@ export class Plane3dByOriginAndVectors extends Plane3d implements BeJSONFunction
    * Return some point on the plane.
    */
   public override getAnyPointOnPlane(result?: Point3d): Point3d {
+    // This function returns origin but in general A Point x is on plane if
+    // and only if x = o + a*u + b*v for unique (scalar) parameters a and b.
```
dassaf4, line 301:
```suggestion
    // This function returns the plane origin. In general, a point x is on the plane if
    // and only if x = o + a*u + b*v, where a and b are scalars.
```

dassaf4 (approved):

PR #5757, opened 2023-07-20
Performance improvements and API variants for sweep (drape) linestring on mesh

EarlinLutz:
* For existing method PolyfaceQuery.sweepLineStringToFacets
   * Significant performance improvement using simple range test between facets and range of all swept points.
   * Parameter list accepts a `PolyfaceVisitor | Polyface` (instead of just Polyface) so caller can limit search to a subset of the facets.
 * New method PolyfaceQuery.sweepLineStringToFacetsXY
   * XY-only search, but input includes a (caller-supplied) `Range2dSearchInterface` object to provide more precise range restrictions.


dassaf4 (approved):

dassaf4:
@pmconne could you review the .md files to advance this PR?  Thanks.

pmconne (approved):
Looks like there were some merge messes with presentation in NextVersion.md. Cleaned up now.

PR #5775, opened 2023-07-24
Promote abstract-ui APIs in use by AppUi

NancyMcCallB:
Promote abstract-ui APIs in use by AppUi

### DIFF
--- a/ui/appui-abstract/src/appui-abstract/utils/UiEventDispatcher.ts
+++ b/ui/appui-abstract/src/appui-abstract/utils/UiEventDispatcher.ts
@@ -41,7 +41,7 @@ export class UiEventDispatcher {
     this._secondaryTimeoutPeriod = this._timeoutPeriod / 2;
   }
 
-  /** @internal - used for testing only */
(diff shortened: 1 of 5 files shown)


ui/appui-abstract/src/appui-abstract/utils/UiEventDispatcher.ts
```diff
@@ -41,7 +41,7 @@ export class UiEventDispatcher {
     this._secondaryTimeoutPeriod = this._timeoutPeriod / 2;
   }
 
-  /** @internal - used for testing only */
+  /** @public - used for testing only */
```
calebmshafer, line 44: we should avoid publishing something as public if it's only used for testing purposes.

Does there need to be a test in AppUI that covers this or can that test be in AppUI abstract?

NancyMcCallB: This is used by the unit tests for the SyncUiEventDispatcher class in appui-react.

PR #5776, opened 2023-07-24
Promoted TentativeOrAccuSnap to public

### LINKED ISSUE #5773: Promote TentativeOrAccuSnap
@bbastings The [MeasurePerpendicularTool](https://github.com/iTwin/viewer-components-react/blob/master/packages/itwin/measure-tools/src/tools/MeasurePerpendicularTool.ts) from measure-tools-react uses the `TentativeOrAccuSnap` API, which is marked internal.

Promote this or should the tool use something else?

bsy-nicholasw:
Fixes #5773


bbastings (approved):

bbastings:
Looks like there aren't any docs, don't merge this and I'll update this PR after lunch to add them. Need to run extract-api anyway,

bbastings (changes requested):
Should document.

bbastings (approved):

PR #5787, opened 2023-07-27
Added support to enable experimental feature in concurrent queries

RohitPtnkr1996:
Added a new optional boolean parameter  "enableExperimentalFeatures" to QueryOptions and DbQueryRequest 
Setting this parameter to true from the QueryOptionsBuilder or adding this to the DBQueryRequest will enable experimental features to be used in the concurrent query.
Defaults to false if not specified.

imodel-native: https://github.com/iTwin/imodel-native/tree/enable-experimental-features-concurrent-queries

### DIFF
--- a/core/common/src/ConcurrentQuery.ts
+++ b/core/common/src/ConcurrentQuery.ts
@@ -131,7 +131,7 @@ export interface QueryOptions extends BaseReaderOptions {
   /**
    * Default to false. Enables experimental features to be used in the query
    */
-  experimentalFeatures?: boolean;
+  enableExperimentalFeatures?: boolean;
 }
 /** @beta */
 export type BlobRange = QueryLimit;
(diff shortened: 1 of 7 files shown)


core/common/src/ConcurrentQuery.ts
```diff
@@ -128,6 +128,10 @@ export interface QueryOptions extends BaseReaderOptions {
    * Determine row format.
    */
   rowFormat?: QueryRowFormat;
+  /**
+   * Default to false. Enables experimental features to be used in the query
```
pmconne, line 132: If these features are "experimental", should this property be `beta`?

core/common/src/ConcurrentQuery.ts
```diff
@@ -128,6 +128,10 @@ export interface QueryOptions extends BaseReaderOptions {
...
+   */
```
pmconne, line 133: "experimental features" is extremely vague. Do you document somewhere what that actually means?

PR #5797, opened 2023-07-28
Added setBaseClass method with tests

christophermlawson:


core/ecschema-editing/src/test/Editing/Entities.test.ts
```diff
@@ -127,4 +127,108 @@ describe("Entities tests", () => {
     const testEntity = await testEditor.schemaContext.getSchemaItem<EntityClass>(result.itemKey!);
     expect(await testEntity?.baseClass).to.eql(await testEditor.schemaContext.getSchemaItem(testEntityBaseRes.itemKey!));
   });
+
```
ColinKerr, line 130: How about a test that would create a base class reference cycle.

Should you be able to add mixins via this method or should they have their own method for clarity

christophermlawson: I think for clarity an addMixin method would be better.  I will add the cycle test you mentioned.

christophermlawson: @ColinKerr Since I created a separate task for reference cycle, and, as I just realized, we already have a addMixin method in the editing API, this PR is ready to be looked over again and closed.

ColinKerr (approved):

PR #5799, opened 2023-07-28
Mobile fixes for display-test-app

tcobbs-bentley:
* Update Android minSdk to 28
* More memory for Android build with debug add-on
* Adjust for vite build being in `lib` instead of `build`
* Support vite debug server on mobile
* Fix iOS crash with orientation change event

### DIFF
--- a/common/changes/@itwin/core-mobile/dta-mobile-fixes_2023-07-28-22-25.json
+++ b/common/changes/@itwin/core-mobile/dta-mobile-fixes_2023-07-28-22-25.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-mobile",
+      "comment": "Mobile fixes for display-test-app",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-mobile"
+}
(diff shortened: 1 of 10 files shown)


tcobbs-bentley:
It appears that leftover files from an old build caused it to work, but it doesn't work after a clean build. I will fix.

tcobbs-bentley:
> It appears that leftover files from an old build caused it to work, but it doesn't work after a clean build. I will fix.

The above wasn't quite right, but I fixed the problem.

pmconne (approved):

common/changes/@itwin/core-mobile/dta-mobile-fixes_2023-07-28-22-25.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-mobile",
+      "comment": "Mobile fixes for display-test-app",
```
aruniverse, line 5:
```suggestion
      "comment": "",
```

aruniverse (approved):

hl662 (approved):
LGTM 👍

PR #5803, opened 2023-07-31
Add frontendTilesUrlTemplate to display-performance-test-app config

DStradley:
Added new config frontendTilesUrlTemplate to display-performance-test-app.  When used, it specifies a url template for @itwin/frontend-tiles to obtain tile trees for spatial views, served over localhost.

   * The string can include special tokens: {iModel.key} or {iModel.filename}.
   *   e.g.: http://localhost:8080{iModel.key}-tiles/3dft/ or http://localhost:8080/MshX/{iModel.filename}/
   * These will get replaced by the value of iModel.key or just the filename of that (no path or extension), correspondingly.
   * Note that the contents of iModel.key in DPTA is a GUID.


pmconne:
One thing to bear in mind is that we cache the tileset URL for a given iModel from the time we first compute it until the iModel is closed. So if you have a configuration that opens an iModel and runs one test with tileset URL defined and another with it undefined (or different), the cache will use the URL from the first test.
If that's a problem we'll need to close the iModel when the URL config changes (I think we do the same for other settings like when TileAdmin options change), or provide a way to bypass the cache.

DStradley:
> One thing to bear in mind is that we cache the tileset URL...
For current plan for ImageTests this would not be a problem, but for general usage it could.  So just adding a test for this config var to requiresRestart would solve this?

pmconne:
> just adding a test for this config var to requiresRestart would solve this?

Yes.

DStradley:
I tried changing tileProps in config (while using but not changing frontendTilesUrlTemplate) and that hangs the same way, so this seems a separate problem.  But why does frontendTilesUrlTemplate work without changing requiresRestart?

pmconne:
>  why does frontendTilesUrlTemplate work without changing requiresRestart?

I don't have an answer for you. [Cache is here](https://github.com/iTwin/itwinjs-core/blob/52b38ae1a7983c00f7a41cd98ccf3eed83b3ca10/extensions/frontend-tiles/src/BatchedSpatialTileTreeRefs.ts#L203).

pmconne:
@ben-polinsky fyi looks like BUILDMINION09A is out of disk space.
```
D:\vsts_a\8\s\tools\build\scripts\docs.js:22
  throw err;
  ^

Error: ENOSPC: no space left on device, write
    at Object.writeSync (node:fs:934:3)
    at Object.writeFileSync (node:fs:2255:26)
    at addPackageMetadata (D:\vsts_a\8\s\tools\build\scripts\utils\addPackageMetadata.js:44:8)
    at D:\vsts_a\8\s\tools\build\scripts\docs.js:95:3
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  errno: -4055,
  syscall: 'write',
  code: 'ENOSPC'
}
```

ben-polinsky:
> @ben-polinsky fyi looks like BUILDMINION09A is out of disk space.

And 09B as well.

DStradley:
Okay, frontendTilesUrlTemplate works without changing requiresRestart because the imodel is already being closed and re-opened each time.  The only time it is not, is for wildcard views on the same model, in which case frontendTilesUrlTemplate cannot be changed in between.  So that is not an issue.

The hang in waitForSceneCompletion from _hasMissingTiles is still a potential issue (though not currently in ImageTests runs).  I currently suspect this is related to the same problem with front end tiles drawing black initially until the view is changed.

test-apps/display-performance-test-app/README.md
```diff
@@ -83,7 +87,7 @@ The default configuration file allows you to specify the following:
...
+You can specify which saved view you wish to use for each test with the viewName property.  If external saved views exist for the local file you can specify which external saved view to use with the extViewName property (you can also specify external saved views by name using the viewName property if the name does not clash with a normal saved view).  You can also specify a saved view to use via the viewString property.  The viewString property must include the _viewname_ and _viewStatePropsString_ properties (and optionally the _selectedElements_ and/or _overrideElements_ properties) which you can copy from the external saved view file.
```
pmconne, line 90: Did your editor append underscores to these property names in a mistaken assumption you were trying to underline them?

DStradley:
@pmconne - fixed some problems with detecting when IModelApp needs to restart, but that did not affect the hang in waitForSceneCompletion from _hasMissingTiles when using frontendTilesUrlTemplate.  I propose that that be handled in a separate PR, as it does not affect ImageTests usage.

test-apps/display-performance-test-app/README.md
DStradley: No, I added them since they did not look right in my preview window with the underscores that were there, and I assumed they were attempting to be italicized.

pmconne: The property names begin with an underscore. If you want to make that clearer you can surround them with backticks.

DStradley: Okay, I guess what threw me was the lack of spaces before several of the underscores.
But shouldn't _viewname be _name?

DStradley: Done, assuming I was correct about _name instead of _viewname.

pmconne (approved):

PR #5806, opened 2023-08-01
Set property name

christophermlawson:
@ColinKerr Any others tests that you can think of for this?


core/ecschema-editing/src/Editing/ECClasses.ts
```diff
@@ -283,6 +283,57 @@ export class ECClasses {
...
+    const newName = new ECName(newPropertyName);
+
+    let mutableClass: MutableClass;
+    try {
+      mutableClass = await this.getClass(classKey);
+    } catch (e: any) {
+      return { errorMessage: e.message };
+    }
+
+    // Does this suffice for not allowing property overrides to be renamed?
```
ColinKerr, line 304: what is this comment about?

christophermlawson: Is was a question I had for you, but I can remove it. I think we are good now. I removed it.

ColinKerr (approved):

common/changes/@itwin/ecschema-editing/set-property-name_2023-08-01-16-43.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-editing",
+      "comment": "Added the ability to rename an ECProperty via the setName method.",
```
khanaffan, line 5: This should be the title of your PR.

PR #5817, opened 2023-08-04
Bump auth-clients versions

hl662:
Bump our use of [auth-clients](https://github.com/iTwin/auth-clients) up to the latest version, refactoring for breaking changes along the way.

Refactors were made in display-test-app, display-performance-test-app, and full-stack-tests/core.
Key breaking changes to accommodate for are documented in [this PR](https://github.com/iTwin/auth-clients/pull/144), mainly to do with Electron based authorization.


pmconne:
Please confirm sign-in continues to work in the test-apps in electron and browser. You can test this using the `dta signin` and `dta signout` keyins.

hl662:
> Please confirm sign-in continues to work in the test-apps in electron and browser. You can test this using the `dta signin` and `dta signout` keyins.

Confirmed. They both work, and I also noticed that the Electron error that pops up when getAccessToken is called ([mentioned here](https://github.com/iTwin/itwinjs-core/blob/c42da649ae070eb6838efd695a3de69b4e433a40/test-apps/display-test-app/src/frontend/App.ts#L273-L275)) is more clear. 
Previously the error raised was `TypeError: Cannot read properties of undefined (reading 'Error')`

Now, it's `Error: Not signed In. First call signIn()`

pmconne (approved):

test-apps/display-test-app/src/frontend/App.ts
```diff
@@ -273,7 +274,9 @@ export class DisplayTestApp {
       // It makes debugging with "pause on caught exceptions" infuriating.
       // ###TODO fix that in the client and remove this
       if (!configuration.noElectronAuth)
-        opts.iModelApp!.authorizationClient = new ElectronRendererAuthorization();
+        opts.iModelApp!.authorizationClient = new ElectronRendererAuthorization({
+          clientId: getConfigurationString("oidcClientId") ?? "imodeljs-spa-test",
```
aruniverse, line 278: `dta signin` is broken for me and a few others. 
we cant use a client for a spa for a native (electron) app

how does `getConfigurationString` correctly obtain the client id for electron?

fyi @hl662

hl662: getConfigurationString is an object, and we use [getConfig()](https://github.com/iTwin/itwinjs-core/blob/e579c6d3311c2bd5673a248222bcc13727834d66/test-apps/display-test-app/src/common/DtaConfiguration.ts#L235) in the `common` folder of DTA to assign a value to the client Id. Will open a new PR to make some improvements to the default client id and the getConfig() to differentiate between when to use an electron client id and when to use a browser client id.

PR #5821, opened 2023-08-07
Fix Version Bump pipeline

anmolshres98:


common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -159,7 +159,7 @@ jobs:
     - job: CreateBranch
       displayName: Create branch for next release
       dependsOn: Bump
-      condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
+      condition: and(in(dependencies.CheckPrevCommit.result, 'Succeeded', 'Skipped'), in(dependencies.Bump.result, 'Succeeded'), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
```
aruniverse, line 162: do we need the second 	`in`? wouldnt that just be straight up `eq`?

aruniverse (approved):

PR #5823, opened 2023-08-07
Update gather-docs.yaml

anmolshres98:
Update branch name for new release branch

### DIFF
--- a/common/config/azure-pipelines/templates/gather-docs.yaml
+++ b/common/config/azure-pipelines/templates/gather-docs.yaml
@@ -103,6 +103,6 @@ steps:
       project: 2c48216e-e72f-48b4-a4eb-40ff1c04e8e4
       pipeline: 7436 # iTwin.js/Docs/iTwin.js Docs - YAML
       buildVersionToDownload: latestFromBranch
-      branchName: refs/heads/master
+      branchName: refs/heads/release/4.1.x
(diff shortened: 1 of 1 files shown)


aruniverse (approved):

aruniverse:
@ben-polinsky can you take a look at the docs failure when you are back?

common/config/azure-pipelines/templates/gather-docs.yaml
```diff
@@ -103,6 +103,6 @@ steps:
       project: 2c48216e-e72f-48b4-a4eb-40ff1c04e8e4
       pipeline: 7436 # iTwin.js/Docs/iTwin.js Docs - YAML
       buildVersionToDownload: latestFromBranch
-      branchName: refs/heads/master
+      branchName: refs/heads/release/4.1.x
```
ben-polinsky, line 106: Since this is the first 4.1.x release, we won't have an .updated.json. We should set this to release/4.0.x, then on the next 4.1.x release set it to 4.1.x.

DanRod1999: Now that 4.1.1 was released can we update this to 4.1.x?

ben-polinsky: Yes absolutely.

PR #5824, opened 2023-08-07
Downgrade rush to v5.99.x from v5.100.x

anmolshres98:
Recent upgrade to rush v5.100.x caused the `rush version --override-bump preminor ...` to break since `preminor` is no longer supported. Check related issue [here](https://github.com/microsoft/rushstack/issues/1335). This was breaking our release builds.

Rush maintainers deemed the `preminor` flag to be unused, hence removed it. I added a [comment](https://github.com/microsoft/rushstack/issues/1335#issuecomment-1668417677) in the same issue to see if we can have that change reverted on their end or a better alternative to get the effect we want. For now, downgrading to unblock release


pmconne (approved):
If you haven't already, you may want to notify other teams like appui.

aruniverse:

aruniverse:
fyi @iTwin/itwinjs-core-ui

aruniverse:
sigh; i think the integration tests are failing because the qa projects api has been turned off
we need to update to using the itwins-client

aruniverse (approved):

PR #5825, opened 2023-08-07
Replaces rush env variable with require.main.filename call

Ben-Pusey-Bentley:
This change gives access to the root directory of the project by accessing the main node module. That directory contains the pacakge.json file that we need for this call. require.main documentation: [https://nodejs.org/api/modules.html#requiremain](https://nodejs.org/api/modules.html#requiremain)

### DIFF
--- a/tools/build/scripts/utils/addPackageMetadata.js
+++ b/tools/build/scripts/utils/addPackageMetadata.js
@@ -7,9 +7,9 @@ const FS = require("fs-extra");
 const path = require("path");
 
 // We cannot guarantee the folder structure of a project
-// so find the project root using rush env variable if available.
-const rootPackageJson = process.env.RUSHSTACK_FILE_ERROR_BASE_FOLDER
-  ? path.join(process.env.RUSHSTACK_FILE_ERROR_BASE_FOLDER, "package.json")
+// so find the project root using require.main if available.
+const rootPackageJson = require.main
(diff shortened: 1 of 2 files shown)


tools/build/scripts/utils/addPackageMetadata.js
```diff
@@ -7,9 +7,9 @@ const FS = require("fs-extra");
 const path = require("path");
 
 // We cannot guarantee the folder structure of a project
-// so find the project root using rush env variable if available.
-const rootPackageJson = process.env.RUSHSTACK_FILE_ERROR_BASE_FOLDER
-  ? path.join(process.env.RUSHSTACK_FILE_ERROR_BASE_FOLDER, "package.json")
+// so find the project root using require.main if available.
+const rootPackageJson = require.main
+  ? path.join(path.dirname(require.main.filename), "package.json")
```
aruniverse, line 12: does this work? how did you test this out?

Ben-Pusey-Bentley: So, I tested this locally, and it works as intended there, but it fails on Azure, and looking into this solution a bit more, that seems to be a consistent issue with this kind of code. Seems like a fatal issue, so I'm just going to close this PR and try to investigate some other solutions. May be best to just rename the environment variable after all.

PR #5840, opened 2023-08-09
Add code value trimming "behaviors" for compatibility with old iModels

MichaelBelousov:
native PR: https://github.com/iTwin/imodel-native/pull/404
Description copied verbatim:

In https://github.com/iTwin/imodel-native/pull/180 we started unicode-aware (previously ascii) trimming of code values. Before that, several production iModels stored code values containing non-ascii whitespace. A couple transformer library users that are trying to move up to 4.x have reported failures when transforming these iModels since the codes are not preserved during db loads/stores.

Since the transformer wants to preserve the code values as they are stored in the Db, this PR introduces a per-dgndb DgnCodeValue::Behavior m_codeValueBehavior, and makes initialization of any DgnCodeValue tied to a dgndb context, e.g. FromJson, LoadFromDb. The default behavior is TrimUnicodeWhiteSpace, but now you can also set it to Exact, exposed to javascript as nativeDb.get/setCodeValueBehavior(val: "exact" | "trim-unicode-whitespace"). The intention is for users like the transformer to change the handling during a transformation, and then turn it back.

Misc:
- I intend to backport this as a patch to the current release after this merges to master

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -2905,6 +2905,9 @@ export abstract class IModelDb extends IModel {
     // @internal (undocumented)
     protected _codeService?: CodeService;
     get codeSpecs(): CodeSpecs;
+    // @beta
+    get codeValueBehavior(): "exact" | "trim-unicode-whitespace";
+    set codeValueBehavior(newBehavior: "exact" | "trim-unicode-whitespace");
     computeProjectExtents(options?: ComputeProjectExtentsOptions): ComputedProjectExtents;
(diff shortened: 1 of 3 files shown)


MichaelBelousov:
cherry-picked #5839 to resolve build failures

nick4598 (approved):

MichaelBelousov:
I was asked to merge master into `imodel02` with this PR. done

dassaf4 (approved):

pmconne (changes requested):
> 637 files changed

Please fix.

MichaelBelousov:
going to separate the master merge from this PR, closing

PR #5847, opened 2023-08-11
Restart default txn on v2 checkpoints after 10 minutes of inactivity

nick4598:
This ends a read transaction marking blocks as unused allowing the daemon to better manage its cachefile. 

A 10 minute timer starts when db is opened as a lazy v2 checkpoint, this timer refreshes anytime refreshContainer is called which is called on every RPC Operation. If no refreshContainer call, then the timer will trigger ending the read transaction and marking blocks as unused.


pmconne (approved):

core/backend/src/IModelDb.ts
```diff
@@ -2752,21 +2758,30 @@ export class SnapshotDb extends IModelDb {
       throw err;
     }
 
+    // unref timer, so it doesn't prevent a process from shutting down.
+    snapshot._restartDefaultTxnTimer = setTimeout(() => {
+      snapshot.restartDefaultTxn();
```
kabentley, line 2763: this could benefit from a comment explaining why this is necessary.

nick4598: Thanks, I added a comment to the timer defined on snapshotdb explaining it. Let me know your thoughts.

presentation/backend/src/test/PresentationRpcImpl.test.ts
```diff
@@ -75,6 +82,8 @@ describe("PresentationRpcImpl", () => {
 
     const imodelTokenMock = moq.Mock.ofType<IModelRpcProps>();
     const imodelMock = moq.Mock.ofType<IModelDb>();
+    // need to define .then because mocks aren't 'thenable'. async functions which return IModelDb hang without this.
+    imodelMock.setup((x: any) => x.then).returns(() => undefined);
```
saskliutas, line 86: You could use `configureForPromiseResult` helper function for this. https://github.com/iTwin/itwinjs-core/blob/8ef85dabc4b5665c69aa1b3abeacc594c66cb314/presentation/common/src/test/_helpers/Mocks.ts#L16

saskliutas (changes requested):

presentation/backend/src/test/PresentationRpcImpl.test.ts
nick4598: Thanks, changed to use that.

presentation/backend/src/presentation-backend/PresentationRpcImpl.ts
```diff
@@ -106,10 +106,12 @@ export class PresentationRpcImpl extends PresentationRpcInterface implements IDi
     return Presentation.getManager(clientId);
   }
 
-  private getIModel(token: IModelRpcProps): IModelDb {
+  private async getIModel(accessToken: AccessToken, token: IModelRpcProps): Promise<IModelDb> {
     let imodel: IModelDb;
     try {
       imodel = IModelDb.findByKey(token.key);
+      // call refreshContainer, just in case this is a V2 checkpoint whose sasToken is about to expire, or its default transaction is about to be restarted.
+      await imodel.refreshContainer(accessToken);
```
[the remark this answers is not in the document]
nick4598: Removed it

saskliutas (approved):

nick4598:

PR #5848, opened 2023-08-11
Make DTA iOS simulator test work on Apple Silicon

tcobbs-bentley:
Xcode added the ability to run any simulator in Rosetta 2 mode if the `simctl launch` command is itself run in Rosetta 2. This updates the run script to do that, as well as updating the actual test run to wait for the model to finish rendering before completing, instead of completing immediately after the iModel is opened. It also switches to using JoesHouse.bim as a test iModel, since that has visible 3D geometry.

Note: I copied JoesHouse.bim from appui, so it was already a public iModel, just not one in itwinjs-core.


tcobbs-bentley:
I set @toddsouthenbentley as an explicit reviewer, and I would like to wait for him to return next week to review this PR before letting it complete.

tcobbs-bentley:
It looks like this requires Xcode 14.3 in order to work.

aruniverse:
> It looks like this requires Xcode 14.3 in order to work.

fyi @chuckkir

chuckkir:
@aruniverse So far we don't have any documentation that says that it's limited to 14.3 so this is just a guess. We will have to update one of the boxes to try. Unfortunately changing xcode is a manual process so we don't take it lightly, plus it can break other things. If we're going to have to do that work on each box, I'd like to understand what is going on with the signing stuff (https://github.com/iTwin/imodel-native-internal/pull/119) and tackle it at the same time.

tcobbs-bentley:
> @aruniverse So far we don't have any documentation that says that it's limited to 14.3 so this is just a guess. We will have to update one of the boxes to try. Unfortunately changing xcode is a manual process so we don't take it lightly, plus it can break other things. If we're going to have to do that work on each box, I'd like to understand what is going on with the signing stuff ([iTwin/imodel-native-internal#119](https://github.com/iTwin/imodel-native-internal/pull/119)) and tackle it at the same time.

I found what I believe to be Apple's documentation about why this works in 14.3 when it didn't in 14.2. The Xcode 14.3 Release notes include the following section indicating that Xcode itself can no longer be run in Rosetta mode:

https://developer.apple.com/documentation/xcode-release-notes/xcode-14_3-release-notes#Deprecations

It then links to a document that provides instructions for dealing with processor architectures and there is a (I think new) menu option of "Show Rosetta Destinations". Also, I feel that since everything works for me in 14.3, and doesn't work at all in 14.2 on the same computer, it seems like proof that even if it wasn't widely documented, 14.3 did indeed make changes that allow it to work. I personally am convinced that Xcode 14.3 is required to run Rosetta simulator apps on Apple Silicon Macs.

chuckkir:
Ok, I’ll update one so we can try.

chuckkir:
We have discovered that Xcode 14.3.1 breaks some of our tests due to a change in the way it handles floating point.
https://www.gaige.net/fp-concat-accuracy.html
We will have to address this before we can update the rest of the machines.

aruniverse:
> > We have discovered that Xcode 14.3.1 breaks some of our tests due to a change in the way it handles floating point. [gaige.net/fp-concat-accuracy.html](https://www.gaige.net/fp-concat-accuracy.html) We will have to address this before we can update the rest of the machines.
> 
> The tests have now been fixed so that they pass on Xcode 14.3.1:
> 
> [iTwin/imodel-native#427](https://github.com/iTwin/imodel-native/pull/427)

so what are the next steps needed?
new addon & update xcode on mac agents?

tcobbs-bentley:
Once Xcode is updated on the build agents, this PR should pass the CI build process. This PR enables execution of display-test-app in an Intel X64 Simulator process (which requires Xcode 14.3 or later). That means that every itwinjs-core CI build will verify that iModels can be opened on mobile.

I have other local changes ready to submit in a followup PR to make it so that (with proper pipeline environment variables), the dta test run will also download an iModel from the iModel Hub (instead of opening a local snapshot that is copied onto the simulator by the test script). It uses a service authorization client to make that happen, so needs the service's secret key (along with the iTwin ID and iModel ID) to be present in environment variables at the time that the testing is performed in the CI build environment.

tcobbs-bentley:
@aruniverse The pipeline Macs now have Xcode 14.3.1, so this PR now passes the pipelines.

aruniverse (approved):

toddsouthenbentley (approved):

pmconne (approved):

PR #5849, opened 2023-08-14
Remove hardcoded classid dependency in test

rschili:
Looking at the test it was intended to do something different anyways. I adjusted the test to do what I think the intention was. Will clarify with Matt before doing the final push.

The previous version had a hard coded class id in the test body which breaks on any changes to system schemas.


pmconne (commented):
@khanaffan mind reviewing?

khanaffan (approved):

pmconne (approved):

PR #5852, opened 2023-08-14
Fix OOM in linux CI due to vite memory usage

hl662:
display-test-app and display-performance-test-app uses a lot of memory during chunk rendering. During its peak, it would leave less than 200 MBs free, out of 16 GBs available on our linux agents. That, coupled with parallel builds happening during `rush build`, and DTA and DPTA builds executing one after the other, would cause out of memory errors to happen. The OOM errors were able to be reproduced by disabling the rush build cache, to make sure all modules had to be built and that the max number of parallel processes would be triggered.

Two things were done to address OOM issues:

1. Conditionally disable source map generation in DTA and DPTA for CI builds to reduce memory usage.
2. Lower the max number of processes that can run in parallel during `rush build` by 50%.

To enable memory and cpu monitoring, I modified the script in linux's rush build to use `top`, like so:
`bash -c 'while true; do top -b -n 1 -o +RES >>mem.log | head -n 22; sleep 1s; done' & node common/scripts/install-run-rush.js build && kill %1`

### DIFF
--- a/common/api/presentation-common.api.md
+++ b/common/api/presentation-common.api.md
@@ -534,6 +534,8 @@ export class Descriptor implements DescriptorSource {
     // @deprecated
     filterExpression?: string;
     static fromJSON(json: DescriptorJSON | undefined): Descriptor | undefined;
+    // @beta
+    getFieldByDescriptor(fieldDescriptor: FieldDescriptor, recurse?: boolean): Field | undefined;
     getFieldByName(name: string, recurse?: boolean): Field | undefined;
     readonly inputKeysHash?: string;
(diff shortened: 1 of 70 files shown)


hl662:
Don't review yet, need to simulate build process into main

hl662:
Here's an open issue on vite exploring some of the options that can be done to further reduce memory usage in `vite`: https://github.com/vitejs/vite/issues/2433

aruniverse (approved):

MichaelBelousov:
@iTwin/itwinjs-core-display review?

PR #5854, opened 2023-08-14
Change updateBlob and updateString to insert if resource does not exist

wgoehrig:
This seems much more convenient behavior for something like workspace-editor.

### DIFF
--- a/common/changes/@itwin/core-backend/workspace-replace_2023-08-14-19-03.json
+++ b/common/changes/@itwin/core-backend/workspace-replace_2023-08-14-19-03.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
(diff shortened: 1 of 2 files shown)


aruniverse (approved):
lgtm, but i wonder if adding an explicit `upsert` fn be better?

pmconne:
Are `addString/Blob` redundant now?

wgoehrig:
> lgtm, but i wonder if adding an explicit `upsert` fn be better?

I assume you mean adding `upsert` _in addition to update_?  I avoided that for two reasons:
- My expectation is that this `upsert` behavior is more desirable anyway, so having both `upsert` and `update` could be confusing.
- The existing `update` behavior was already broken - it was not, in fact, throwing an exception if the resource was missing.  So to keep it distinct from `upsert` we'd also have to add another query to `update`, which will make `update` arguably _more_ expensive that `upsert`, even when used exclusively for updating.  That seems counterintuitive to me.

> Are addString/Blob redundant now? Your change assumes callers of update don't care if the resource already exists; do we assume callers of add do want an exception thrown if the resource does already exist?

Yeah, my assumption is that's still useful for inserts - if you are explicitly adding something new you should be able to choose to be more wary of name collisions.  Plus, that's still easy to do in a single SQL statement, whereas throwing an exception when the resource already exists would require us to first run a SELECT, followed by an UPDATE (unless we want to expose the sqlite3_changes function, but that's more work...).

If anyone still really wants the "check if missing and throw" update behavior they can still do that explicitly before calling `update`.

PR #5856, opened 2023-08-14
Instructions to fix broken symlink build error

tcobbs-bentley:


tcobbs-bentley:
@aruniverse @tm-zub @nick4598 @pmconne @calebmshafer @wgoehrig @ben-polinsky Can somebody either provide feedback or approve this?

README.md
```diff
@@ -38,6 +38,14 @@ The `-v` option for `rush` is short for `--verbose` which results in a more verb
 
 The above commands iterate and perform their action against each package in the monorepo.
 
+If you get an error similar to the following, it means that the repo has stopped making use of an npm package that was used in the past.
+
+```
+[Error: ENOENT: no such file or directory, stat '/.../itwinjs-core/test-apps/display-test-app/node_modules/@bentley/react-scripts']
+```
+
+To fix this build error, you should completely remove the node_modules directory that is referenced in the error, and then rerun `rush install`.
```
aruniverse, line 47:
```suggestion
To fix this build error, you should completely remove the node_modules directory that is referenced in the error with `rush purge`, and then rerun `rush install`.
```

aruniverse (approved):
not sure if this is that common or significantly difficult of an issue that this needs to be called out explicitly in the readme

tcobbs-bentley:
> not sure if this is that common or significantly difficult of an issue that this needs to be called out explicitly in the readme

I believe that I created this PR shortly after the last time it happened to me. It's not common, but I feel that the fix is non-obvious.

aruniverse:
> We can close (reject) this PR if you want, but I don't feel it should be left hanging.

Since this will be the only reference to purge command, i dont mind adding it since its a useful command to know.

https://github.com/search?q=repo%3AiTwin%2Fitwinjs-core+%22rush+purge%22&type=code

PR #5861, opened 2023-08-15
Code trim behaviors

MichaelBelousov:
Replaces this one https://github.com/iTwin/itwinjs-core/pull/5840 that originally also merged master. Description copied verbatim:

native PR: https://github.com/iTwin/imodel-native/pull/404
Description copied verbatim:

In https://github.com/iTwin/imodel-native/pull/180 we started unicode-aware (previously ascii) trimming of code values. Before that, several production iModels stored code values containing non-ascii whitespace. A couple transformer library users that are trying to move up to 4.x have reported failures when transforming these iModels since the codes are not preserved during db loads/stores.

Since the transformer wants to preserve the code values as they are stored in the Db, this PR introduces a per-dgndb `DgnCodeValue::Behavior m_codeValueBehavior`, and makes initialization of any DgnCodeValue tied to a dgndb context, e.g. FromJson, LoadFromDb. The default behavior is `TrimUnicodeWhiteSpace`, but now you can also set it to Exact, exposed to javascript as `nativeDb.get/setCodeValueBehavior(val: "exact" | "trim-unicode-whitespace")` and `IModelDb.codeValueBehavior`. The intention is for users like the transformer to change the handling during a transformation, and then turn it back.

Note that this PR does not change the behavior that has always existed of the JavaScript `Code` class, which performs unicode-aware whitespace trimming. The applications that managed to get non-ascii whitespace into the iModel in the first place were going around this class, and so is the transformer. This means `IModelDb.elements.getElement` will still trim codes in returned elements. If someone thinks that means we should remove the `IModelDb.codeValueBehavior` API I am introducing in this PR, I think that is reasonable, but I didn't want to rely on an internal API in the transformer.

Misc:
- I intend to backport this as a patch to the current release after this merges to master
- cherry-picked #5839 from master due to build failures

### DIFF
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -1340,11 +1340,8 @@ export abstract class IModelDb extends IModel {
   }
 
   /**
-   * For cases where you need to preserve exact [[Code]]s between databases,
-   * you can set the codeValueBehavior to "exact". The default is "trim-unicode-whitespace".
-   * Code Values should be trimmed, this is primiarly for compatibility with iModels that
-   * failed to trim their codes.
-   * @beta
+   * Controls how [Code]($common)s are copied from this iModel into another iModel, to work around problems with iModels created by older connectors. The [imodel-transformer](https://github.com/iTwin/imodel-transformer) sets this appropriately on your behalf - you should never need to set or interrogate this property yourself.
+   * @public
    */
   public get codeValueBehavior(): "exact" | "trim-unicode-whitespace" {
     return this.nativeDb.getCodeValueBehavior();
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -1338,6 +1338,21 @@ export abstract class IModelDb extends IModel {
       }
     });
   }
+
+  /**
+   * For cases where you need to preserve exact [[Code]]s between databases,
+   * you can set the codeValueBehavior to "exact". The default is "trim-unicode-whitespace".
+   * Code Values should be trimmed, this is primiarly for compatibility with iModels that
+   * failed to trim their codes.
+   * @beta
+   */
+  public get codeValueBehavior(): "exact" | "trim-unicode-whitespace" {
+    return this.nativeDb.getCodeValueBehavior();
+  }
+
+  public set codeValueBehavior(newBehavior: "exact" | "trim-unicode-whitespace") {
+    this.nativeDb.setCodeValueBehavior(newBehavior);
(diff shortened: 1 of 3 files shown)


core/backend/src/IModelDb.ts
```diff
@@ -1338,6 +1338,21 @@ export abstract class IModelDb extends IModel {
       }
     });
   }
+
+  /**
+   * For cases where you need to preserve exact [[Code]]s between databases,
+   * you can set the codeValueBehavior to "exact". The default is "trim-unicode-whitespace".
+   * Code Values should be trimmed, this is primiarly for compatibility with iModels that
+   * failed to trim their codes.
+   * @beta
```
pmconne, line 1347: I'm concerned that this is `beta` and doesn't make clear that it only affects code values created in native code. Almost nobody using our TS APIs is going to understand the ramifications of that.

MichaelBelousov: Would you prefer I make it internal but note that the transformer packages use it?

pmconne: Do you expect users to know - or realize they should know - that an iModel they're running the transformer on may suffer this bug and need to set this property?
Or will you just make the transformer always set this to "exact"?

MichaelBelousov: transformer always sets this to exact

pmconne:
> A couple transformer library users that are trying to move up to 4.x have reported failures when transforming these iModels since the codes are not preserved during db loads/stores.

Can you be more specific about whether these are real users, what specific failures, etc?

core/backend/src/IModelDb.ts
pmconne, line 1347:
```suggestion
   * Controls how [Code]($common)s are copied from this iModel into another iModel, to work around problems with iModels created by older connectors. The [imodel-transform]($insert-link-here-please) sets this appropriately on your behalf - you should never need to set or interrogate this property yourself.
   * @public
```

pmconne: If anyone takes a code value read out of the native db and passes it to the TS `Code` constructor, new fun bugs will ensue. See suggested change below.

MichaelBelousov: it resolved automatically. Leaving for you to resolve since I changed the link.

pmconne (approved):

MichaelBelousov:

MichaelBelousov:
@merigifyio backport release/4.1.x

PR #5870, opened 2023-08-15
Update vite config, improve vite build memory performance in CI builds

aruniverse:
- run `rush update --full` for dep hygiene
- update vite config to get past OOM flakiness on CI
- test out new [bundler module resolution](https://www.typescriptlang.org/tsconfig#moduleResolution) for ts
- Re-enable parallelism count to max in Linux CI's `rush build` 
- Add `VITE_CI` flag to CI builds, disabling source map generation in CI builds and improving performance
- Bump vite up to `^4.4.9`, which addresses memory consumption

Monitored memory using `bash -c 'while true; do top -b -n 1 -o +RES | head -n 22; sleep 1s; done' & node common/scripts/install-run-rush.js build -v -p max && kill %1`,

After the changes above, peak memory usage when parallelism count set to max is ~6.5 GiB (compared to 16 GiB used previously), 3.5 GiB free to use at that point -> Significant improvement in memory usage on Linux CI builds.


test-apps/display-performance-test-app/vite.config.ts
```diff
@@ -60,21 +66,24 @@ export default defineConfig(() => {
...
           /core\/frontend/, // prevent errors with require in IModelApp
         ],
-        transformMixedEsModules: true // transforms require statements
+        transformMixedEsModules: true, // transforms require statements
       },
       rollupOptions: {
-        input: path.resolve(__dirname,"index.html"),
+        input: path.resolve(__dirname, "index.html"),
         // run `rushx build --stats` to view stats
+        logLevel: process.env.CI ? "silent" : "warn",
```
hl662, line 74: Checking the logs in azure pipelines I noticed the warnings weren't muted for DPTA - the `process.env.CI` also returned undefined. Does azure dev ops not set CI to false automatically? Should we change our rush build command in the yaml file to have CI set to true?

hl662:
I know, that was a mistake 😕 just undid it

test-apps/display-test-app/vite.config.ts
```diff
@@ -15,20 +15,25 @@ import * as packageJson from "./package.json";
 import path from "path";
 
 const mode = process.env.NODE_ENV === "development" ? "development" : "production";
+console.log(process.env.CI);
```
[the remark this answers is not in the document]
hl662: removed

common/config/azure-pipelines/templates/core-build.yaml
```diff
@@ -82,14 +82,16 @@ steps:
     env:
       RUSH_BUILD_CACHE_CREDENTIAL: $(RushBuildCacheSAS)
       RUSH_BUILD_CACHE_ENABLED: ${{parameters.rushBuildCacheEnabled}}
+      VITE_CI: true
     condition: and(succeeded(), ne(variables['Agent.OS'], 'Linux'))
-  # TEMP: build on linux but with 50% CPU saturation to prevent OOM
-  - script: node common/scripts/install-run-rush.js build -v -p '50%'
+
+  - script: node common/scripts/install-run-rush.js build -v -p max
```
aruniverse, line 88: do we need this separate task now? cant we revert the addition form your last pr and just set this new env var?

test-apps/display-performance-test-app/vite.config.ts
hl662: Added a `VITE_CI` flag to address only packages that uses vite for building.

common/config/azure-pipelines/templates/core-build.yaml
hl662: You're right, I've removed the separate task.

pmconne (approved):

ColinKerr (approved):

PR #5871, opened 2023-08-16
CloudSQLite: If user already has write lock and calls acquireWriteLock again, then don't release it.

nick4598:


core/backend/src/CloudSqlite.ts
```diff
@@ -623,13 +623,16 @@ export namespace CloudSqlite {
     await acquireWriteLock(args);
```
kabentley, line 623: I think it's easier to just do:

```ts
    if (args.container.hasWriteLock)
      return operation();
```

nick4598: if (args.container.hasWriteLock && heldBySameUser) ?

nick4598: Also, if I'm not mistaken the current state of the code does not allow you to refresh the write lock if you're the same user in the case of a long-running operation? CloudSqlite.acquireWriteLock returns if its the same user, but the native version of acquireWriteLock has

 "Calling `acquireWriteLock` with the lock already held resets the lock duration from the current time,
     * so long running processes should call this method periodically to ensure their lock doesn't expire (they should also make sure their accessToken is refreshed
     * before it expires.)"

kabentley: there's only one write lock, so if it's already held we should do what the comment says. But the whole "same user" thing is only relevant for RPC calls, which can't be recursive or long-running.

kabentley (approved):

core/backend/src/CloudSqlite.ts
```diff
@@ -623,6 +623,8 @@ export namespace CloudSqlite {
...
     try {
       const containerInternal = args.container as CloudContainerInternal;
+      if (containerInternal.writeLockHeldBy === args.user) // If the user already had the write lock, then don't release it.
```
kabentley, line 626: oops sorry. You need to save this BEFORE calling acquireWriteLock,

kabentley (changes requested):

core/backend/src/CloudSqlite.ts
kabentley: nevermind, it's fine the way you have it (this is all tricky).

kabentley (approved):

aruniverse (approved):

PR #5872, opened 2023-08-16
Fix full-stack-tests/core debugging config and clean up launch.json

wgoehrig:
I took the opportunity to do some light cleanup of the launch.json, including:

- Dropped the [deprecated](https://github.com/microsoft/vscode-js-debug/commit/f314b55c8035fa535c42411c522714124df2dce5) `pwa-` prefix from config types.
- Removed references to deleted directories in `outFiles`
- Replace broken "UI Core Tests" launch config with "AppUI Abstract Tests"

FWIW, I believe that VSCode's sourcemap resolution has improved to the point where we don't need to specify `outFiles` explicitly anymore.  As much as I'd like to go thru and remove all those, we should probably do some more extensive testing first...

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -692,7 +692,7 @@
       "program": "${workspaceFolder}/test-apps/display-test-app/lib/backend/WebMain.js",
       "outFiles": [
         "${workspaceFolder}/test-apps/display-test-app/lib/backend/**/*.js",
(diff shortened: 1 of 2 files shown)


aruniverse (approved):

pmconne (approved):
On Windows, I can now stop at breakpoints in frontend and backend.

PR #5883, opened 2023-08-17
Saeed torabi/curve close approach

saeeedtorabi:
Support curve chain (path and loop) in CurveCurveCloseApproachXY

Also fixed bugs for close approach between segment and lineString/arc.
Added multiple new tests.


core/geometry/src/curve/CurveCurveCloseApproachXY.ts
```diff
@@ -483,49 +509,53 @@ export class CurveCurveCloseApproachXY extends NullGeometryHandler {
...
-    // 2) true intersection
-    // 3) line parallel to arc tangent.
-    this.testAndRecordFractionalPairApproach(cpA, 0, 1, true, arc, 0, 1, false, reversed);
-    // Arc: X = C + cU + sV
-    // Line:  contains points A0,A1
-    // Arc point colinear with line if det (A0, A1, X) = 0
-    // with homogeneous xyw points and vectors.
-    // With equational X:   det (A0, A1, C) + c det (A0, A1,U) + s det (A0, A1, V) = 0.
+    // 1) arc endpoints to segment endpoints or arc endpoints projection to the segment
+    this.testAndRecordFractionalPairApproach(cpA, 0, 1, true, arc, 0, 1, true, reversed);
```
saeeedtorabi, line 513: Should be true, false. Accidentally changed to true,true.

dassaf4: I tried to revert to "true, false", but that caused a test failure. Can you write a test that fails if "true, true" is passed in?

dassaf4:
I added some of the fixes we talked about and moved some files around to keep the internal context classes out of core-geometry.api.md. If you can add a test verifying the new `CurveCurve.closestApproachProjectedXYPair` convenience method, this would be great. I'll add the fix to FrameBuilder tomorrow, then review what's there so far!

dassaf4:
Pushed a typo fix and some other fat fingers revealed by a test (`SingleLineLine2`) that started failing after I added the swap fix in `segmentSegmentBoundedApproach`.

dassaf4 (approved):
Looks good so far!

bbastings (approved):

PR #5884, opened 2023-08-18
Fixes a bug while calling native `SchemaUtility` APIs

akshay05n:
This fixes a bug in SchemaUtils that was causing a call to a native SchemaUtility API to fail with an error message `Cannot read properties of undefined (reading 'SchemaUtility')`


pmconne (approved):
Keep in mind that the target audience for changelogs doesn't care about internal stuff like unit tests and native APIs.

PR #5895, opened 2023-08-21
Promote and document APIs required to implement a MapLayerFormat

### LINKED ISSUE #4479: Promote and document APIs required to implement a MapLayerFormat
Applications can in theory inject their own map layer formats by implementing `MapLayerFormat`, which is `@beta`. But the guts of that interface and the APIs required to properly implement them are largely `@internal`.
Identify the APIs that need to be promoted. Clean them up and simplify where appropriate and document them thoroughly.
Ideally, produce an example implementation of a custom map layer format, using only `beta` and `public` APIs.

@mdastous-bentley indicated he is doing (a subset of?) this as part of his vector map tiles PR.

eringram:
Fixes #4479.

- Added a new test MapLayerFormat.test.ts that contains a simple custom implementation of MapLayerFormat using only public and beta APIs. 
- Promoted `ImageryMapLayerTreeReference` and `MapLayerTileTreeReference` to `beta` to allow new map layer formats to override `MapLayerFormat.createMapLayerTree()`
- Promoted some getters/setters on `MapLayerTileTreeReference` that may be useful to new map layer formats
- Improved documentation in a few classes related to map layer formats.


core/frontend/src/tile/map/MapLayerImageryProvider.ts
```diff
@@ -268,35 +274,48 @@ export abstract class MapLayerImageryProvider {
     return true;
   }
 
-  /** @internal */
-  // calculates the projected x cartesian coordinate in EPSG:3857from the longitude in EPSG:4326 (WGS84)
+  /**
+   * Calculates the projected x cartesian coordinate in EPSG:3857from the longitude in EPSG:4326 (WGS84)
```
ben-polinsky, line 278: nit: typo - space needed in-between `3857` and `from`?

pmconne (commented):
Run `rush change` and commit the resultant files.
More details [here](https://github.com/iTwin/itwinjs-core#source-code-edit-workflow).

core/frontend/src/tile/map/MapLayerImageryProvider.ts
```diff
@@ -268,35 +274,48 @@ export abstract class MapLayerImageryProvider {
...
+   * @param longitude Longitude in EPSG:4326 (WGS84)
+   */
   public getEPSG3857X(longitude: number): number {
     return longitude * 20037508.34 / 180.0;
   }
 
-  /** @internal */
-  // calculates the projected y cartesian coordinate in EPSG:3857from the latitude in EPSG:4326 (WGS84)
+  /**
+   * Calculates the projected y cartesian coordinate in EPSG:3857from the latitude in EPSG:4326 (WGS84)
```
ben-polinsky, line 286: nit: `EPSG:3857from`

ben-polinsky (approved):

common/changes/@itwin/core-frontend/eringram-maplayers-formats-promote_2023-08-22-16-21.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Add MapLayerFormat.test.ts with a simple custom implementation of MapLayerFormat using only public/beta APIs, also improve docs in map layer formats-related classes",
```
pmconne, line 5:
```suggestion
      "comment": "Promote APIs required to implement a MapLayerFormat to beta",
```
Change log messages are written for end users, who don't care about internal stuff like tests.
It's ok to leave the message empty if only internal changes were made.

pmconne (changes requested):
Promoting an interface or class promotes all of its members that don't have their own release tags. Did you review each and make sure that (1) we want to promote it and (2) if so, it is properly documented?
Example: `MapLayerTileTreeReference` has a `layerIndex` property that is supplied to the constructor. Both the constructor and the property are undocumented. I personally have no idea how a caller would determine the correct value to pass to the constructor.

eringram:
> Promoting an interface or class promotes all of its members that don't have their own release tags. Did you review each and make sure that (1) we want to promote it and (2) if so, it is properly documented? Example: `MapLayerTileTreeReference` has a `layerIndex` property that is supplied to the constructor. Both the constructor and the property are undocumented. I personally have no idea how a caller would determine the correct value to pass to the constructor.

I reviewed most of the members of ImageryMapLayerTreeReference and MapLayerTileTreeReference and concluded they were either self-explanatory as getters/setters or inherited the documentation sufficiently from TileTreeReference. However I overlooked the constructors for some reason, and I also see that most of the getters should probably be documented too since they make sense in a map layers context but are clearer if explained explicitly. I'll go back through them

mdastous-bentley:
I think `MapLayerFormat.createImageryProvider()` and `MapLayerFormat.createMapLayerTree()` should also be exposed since it needs to be overridden when creating a new format.

mdastous-bentley:
There are also a lot of methods in `MapLayerImageryProvider`, some of them might need to be exposed.  For example, `MapLayerImageryProvider._generateChildIds` is overridden by  `ArcGISMapLayerImageryProvider` and `WmtsMapLayerImageryProvider`, so we might expect other format needing it too..  
We can also keep it simple and expose it as needed.

eringram:
> There are also a lot of methods in `MapLayerImageryProvider`, some of them might need to be exposed. For example, `MapLayerImageryProvider._generateChildIds` is overridden by `ArcGISMapLayerImageryProvider` and `WmtsMapLayerImageryProvider`, so we might expect other format needing it too.. We can also keep it simple and expose it as needed.

I think that's a good idea. I'm thinking these internal `MapLayerImageryProvider` methods would probably be useful to new providers and should be exposed, as they are used by multiple imagery providers in core-frontend:
`generateChildIds()`
`initialize()`
`getToolTip()`
`addLogoCards()`

And these are used but only by one provider, so we could wait until they have more demand:
`getFeatureInfo()`
`onStatusUpdated()`
`useGeographicTilingScheme()`

eringram:
@mdastous-bentley I'm also wondering why subclasses override `_generateChildIds` and not `generateChildIds`?

core/frontend/src/tile/map/MapLayerTileTreeReference.ts
```diff
@@ -13,22 +13,42 @@ import { IModelApp } from "../../IModelApp";
...
+ * @beta
+ */
 export abstract class MapLayerTileTreeReference extends TileTreeReference {
+  /**
+   * Constructor for a MapLayerTileTreeReference.
+   * @param _layerSettings Map layer settings that are applied to the MapLayerTileTreeReference.
+   * @param _layerIndex The index of the associated map layer.
+   * @param iModel The iModel containing the MapLayerTileTreeReference.
+   */
   constructor(protected _layerSettings: MapLayerSettings, protected _layerIndex: number, public iModel: IModelConnection) {
```
mdastous-bentley, line 27: How does API tags deals with constructor shorthands?  For example, does iModel becomes automatically @beta ?
@pmconne I think we should avoid constructor shorthand?

core/frontend/src/tile/map/MapLayerTileTreeReference.ts
```diff
@@ -13,22 +13,42 @@ import { IModelApp } from "../../IModelApp";
 import { IModelConnection } from "../../IModelConnection";
 import { createModelMapLayerTileTreeReference, MapLayerImageryProvider, TileTreeReference } from "../internal";
 
-/** @internal  */
+/**
+ * A reference to a [[TileTree]] to be used for map layer tiles.
```
mdastous-bentley, line 17: The description is quite thin here.  

I feel like the concept of TileTree/TileTreeReference is not documented enough in general.  We should at least reference the [TileTreeReference ](https://www.itwinjs.org/reference/core-frontend/tiles/tiletreereference/) documentation.

One key concept concept to mentions is that multiple MapLayerTileTreeReference can refer to same TileTree if the layer settings are "equivalent".

core/frontend/src/tile/map/MapLayerTileTreeReference.ts
```diff
@@ -13,22 +13,42 @@ import { IModelApp } from "../../IModelApp";
...
+ * @beta
```
mdastous-bentley, line 18: All members are exposed here, is it needed?  Many of them not documented,,,

core/frontend/src/tile/map/MapLayerImageryProvider.ts
```diff
@@ -268,35 +295,48 @@ export abstract class MapLayerImageryProvider {
     return true;
   }
 
-  /** @internal */
-  // calculates the projected x cartesian coordinate in EPSG:3857from the longitude in EPSG:4326 (WGS84)
+  /**
```
mdastous-bentley, line 298: Do we really want to expose those reprojection methods?

eringram: They are used by ArcGISFeatureProvider in the map-layers-formats package, so I thought they should be promoted due to being used outside of core-frontend and maybe being useful in the future for new imagery providers/formats. But promoting them now might be premature so I can revert until they are needed by more formats

core/frontend/src/tile/map/MapLayerTileTreeReference.ts
eringram: I've gone back over these and reverted some members that aren't needed

core/frontend/src/tile/map/MapLayerTileTreeReference.ts
eringram: Changed to explicitly set the properties

mdastous-bentley (approved):

common/api/core-frontend.api.md
```diff
@@ -5839,7 +5838,7 @@ export class ImageryMapLayerFormat extends MapLayerFormat {
     static createMapLayerTree(layerSettings: ImageMapLayerSettings, layerIndex: number, iModel: IModelConnection): MapLayerTileTreeReference | undefined;
 }
 
-// @internal
+// @beta
```
pmconne, line 5841: You have promoted `resetTreeOwner` to beta, but it is `internal` in the base `TileTreeReference` class.

common/api/core-frontend.api.md
```diff
@@ -5839,7 +5838,7 @@ export class ImageryMapLayerFormat extends MapLayerFormat {
     static createMapLayerTree(layerSettings: ImageMapLayerSettings, layerIndex: number, iModel: IModelConnection): MapLayerTileTreeReference | undefined;
 }
 
-// @internal
+// @beta
 export class ImageryMapLayerTreeReference extends MapLayerTileTreeReference {
     constructor(layerSettings: MapLayerSettings, layerIndex: number, iModel: IModelConnection);
```
pmconne, line 5843: Public constructors are a backwards-compatibility footgun. If a constructor must be public, consider redefining it to take a single argument, so that you can adjust the arguments in the future without breaking the API contract.
Preferring a single `args` object over multiple arguments is a [general guideline](https://www.itwinjs.org/learning/api-support-policies/#breaking-changes) but it's particularly important for constructors because you can't deprecate and replace a constructor.

eringram: Decided to make this constructor beta and switch to a single argument. Because `MapLayerTileTreeReference` is in the signature of `ImageryMapLayerTreeReference`, it also must be beta but its constructor will stay internal. After discussing with Michel I also promoted `MapLayerTileTreeReference` getters/setters that may be useful for public consumers

common/api/core-frontend.api.md
eringram: This was an oversight and I have reverted since it seems like it should stay internal, and the correct publicly accessible way to do this is through the [viewport](https://github.com/iTwin/itwinjs-core/blob/47d893ee815788ab1d6cbad681604b86d19f85da/core/frontend/src/Viewport.ts#L896)

core/frontend/src/test/tile/map/MapLayerFormat.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+class TestMapLayerFormat extends MapLayerFormat {
+  public static override formatId = "TestMapLayerFormat";
+
+  public static override createImageryProvider(settings: TestMapLayerSettings): MapLayerImageryProvider | undefined {
+    return new TestMapLayerImageryProvider(settings, false);
+  }
+
+  public static override createMapLayerTree(layerSettings: TestMapLayerSettings, layerIndex: number, iModel: IModelConnection) {
+    return new ImageryMapLayerTreeReference({ layerSettings, layerIndex, iModel });
```
pmconne, line 24: The constructor you are calling is `internal`.
```
export class ImageryMapLayerTreeReference extends MapLayerTileTreeReference {
    constructor(layerSettings: MapLayerSettings, layerIndex: number, iModel: IModelConnection);
    // @internal
    constructor(args: {
```

eringram: Sorry. Must have forgotten to actually remove the `internal` tag after debating about this...

common/api/core-frontend.api.md
```diff
@@ -7734,9 +7737,9 @@ export class MapLayerFeatureRecord {
 
 // @public
 export class MapLayerFormat {
-    // @internal
+    // @beta
```
pmconne, line 7740: `createImageryProvider` and `createMapLayerTree` should be `abstract`. The default implementations `assert` at run-time presumably to tell the developer that they were supposed to overrde these functions. Far better to express that at compile time.

pmconne: Part of promoting from internal to beta means refactoring APIs to make them more usable.
I'm not sure what your concern is about the methods being called statically. Yes, they are static methods. They still need to be overridden to do anything useful.

pmconne: I think you're telling me that `static` methods cannot also be `abstract`.
That is indeed a problem.

pmconne: Dunno why MapLayerFormat consists of only static methods, but can't change it now without breaking the public API.
Please add a `return undefined` to `createImageryProvider` and add a message to the two `asserts` to make it clear to the developer that they are supposed to override these methods.

PR #5897, opened 2023-08-22
Add tests for query binder

simasjar:
closes iTwin/itwinjs-backlog#849


pmconne (approved):

core/common/src/ConcurrentQuery.ts
```diff
@@ -298,7 +298,7 @@ export class BlobOptionsBuilder {
 }
 
 /** @internal */
-enum QueryParamType {
+export enum QueryParamType {
```
rschili, line 301: @khanaffan is it good to start exporting this type? Looks like it may be intended internal only.
Simonas is using it just for asserting stuff further down in the code.

PR #5900, opened 2023-08-22
Don't prefetch v2 checkpoints twice, check first if one is already ongoing.

nick4598:
imodel-native: https://github.com/iTwin/imodel-native/tree/nick/prefetch-client-counts

### DIFF
--- a/core/backend/src/CheckpointManager.ts
+++ b/core/backend/src/CheckpointManager.ts
@@ -215,7 +215,7 @@ export class V2CheckpointManager {
         const maxRequests = getPrefetchConfig("maxRequests", 6);
         const timeout = getPrefetchConfig("timeout", 100);
         const maxBlocks = getPrefetchConfig("maxBlocks", 500); // default size of 2GB. Assumes a checkpoint block size of 4MB.
-        if (dbStats?.totalBlocks !== undefined && dbStats.totalBlocks <= maxBlocks) {
+        if (dbStats?.totalBlocks !== undefined && dbStats.totalBlocks <= maxBlocks && dbStats.nPrefetch === 0) {
           const logPrefetch = async (prefetch: CloudSqlite.CloudPrefetch) => {
             const stopwatch = new StopWatch(`[${container.containerId}/${dbName}]`, true);
             Logger.logInfo(loggerCategory, `Starting prefetch of ${stopwatch.description}`, { minRequests, maxRequests, timeout });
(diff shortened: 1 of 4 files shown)


khanaffan:
@nick4598 I do not see any unit test with changes.

nick4598:
> @nick4598 I do not see any unit test with changes.

Thanks added a test

core/backend/src/CheckpointManager.ts
```diff
@@ -225,7 +225,7 @@ export class V2CheckpointManager {
           // eslint-disable-next-line @typescript-eslint/no-floating-promises
           logPrefetch(CloudSqlite.startCloudPrefetch(container, dbName, { minRequests, nRequests: maxRequests, timeout }));
         } else {
-          Logger.logInfo(loggerCategory, `Skipping prefetch due to size limits.`, { maxBlocks, totalBlocksInDb: dbStats?.totalBlocks, v2props });
+          Logger.logInfo(loggerCategory, `Skipping prefetch due to size limits or ongoing prefetch.`, { maxBlocks, numPrefetches: dbStats?.nPrefetch, totalBlocksInDb: dbStats?.totalBlocks, v2props });
```
kabentley, line 228: probably helpful to say whether we're skipping due to an active prefetch already in progress or it being disabled.

kabentley:
I frankly don't understand why this comes up. Do we really start more than one backend on the same node for a single iModel?

core/backend/src/CheckpointManager.ts
nick4598: Yeah I thought about having separate log messages for each case, but as it is currently we can deduce which case we are in by looking at what was logged. The maxBlocks limit, totalBlocks in the database, and the number of active prefetches. Between those 3, we can figure out the case. Do you still prefer a separate log message?

nick4598:
> I frankly don't understand why this comes up. Do we really start more than one backend on the same node for a single iModel?

This is fairly new, DR has recently started pointing their frontend traffic for iModelReadRPCInterfaces, PresntationRPCinterface etc to GPB. This would mean two backends for the same iModel. There is some preference in k8s to put pods with the same iModelid together, but its not guaranteed. In the case that they aren't on the same node, DR would also trigger a prefetch on its own.

core/backend/src/CheckpointManager.ts
kabentley: as long as you have something you can search for in the logs to find when multiple prefetches would have happened, that's fine

kabentley (approved):

nick4598:

PR #5908, opened 2023-08-23
fix finalize release errors

DanRod1999:
this is a copy of another PR into master. Might be more beneficial to merge here first though

remove extra parameters to push. The github token appears to not do anything, and removing the "origin/" from the target branch string should allow the push to function correctly. Only question I cannot answer is if this admin account is allowed to push to a protected branch without using some kind of token.

If anyone knows the expected behavior please let me know. I couldn't find anything online so I may be looking in the wrong place.

### DIFF
--- a/.github/workflows/finalize-release.yaml
+++ b/.github/workflows/finalize-release.yaml
@@ -16,9 +16,9 @@ jobs:
     name: Cherry-pick Changelogs
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
-          ref: ${{ github.ref }}
+          ref: ${{ github.ref }} # checkouts the branch that triggered the workflow
           fetch-depth: 0
       - name: Set Git Config
         run: |
--- a/.github/workflows/finalize-release.yaml
+++ b/.github/workflows/finalize-release.yaml
@@ -18,6 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/finalize-release.yaml
```diff
@@ -16,9 +16,9 @@ jobs:
     name: Cherry-pick Changelogs
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
-          ref: ${{ github.ref }}
+          ref: ${{ github.ref }} # checkouts the branch that triggered the workflow
```
aruniverse, line 21: based on the triggers you have (L8-12), doesnt this pipeline get run when a new tag has been pushed with the `release/` prefix? how does it know what branch to use from that?

DanRod1999: Based on the docs I believe its fine:
  | The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by push, this is the branch or tag ref that was pushed.
  
  unless the ref of the tag is different then a ref to the branch? I'm not too familiar with tags, so I'll double check this. Thanks for pointing it out

DanRod1999:
>  Only question I cannot answer is if this admin account is allowed to push to a protected branch without using some kind of token.
> 
> If anyone knows the expected behavior please let me know. I couldn't find anything online so I may be looking in the wrong place.

After talking to Arun, the admin account should be able to push directly to a protected branch since it is the assigned user making the push.

PR #5914, opened 2023-08-25
Remove storageType from RequestTokenProps

MichaelSwigerAtBentley:
Removes storageType from RequestTokenProps as per feedback from Sam. 

Why is storageType required to request a token? The only way to get storageType back is from a token request, but it is also required to make that request.


common/changes/@itwin/core-backend/swiger-blob-container-service-feedback_2023-08-25-17-10.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Removes storageType from TokenRequestProps",
```
pmconne, line 5:
```suggestion
      "comment": "Remove `storageType` from `BlobContainer.RequestTokenProps`.",
```

core/backend/src/BlobContainerService.ts
```diff
@@ -108,7 +108,6 @@ export namespace BlobContainer {
 
   /** Information required to request an access token for a container. */
   export interface RequestTokenProps extends AccessContainerProps {
-    storageType: Provider;
```
kabentley, line 111: No, this change is wrong, please do not make it. When you create a container, it tells you what the storageType is and you must save it. That's what this is all about.

kabentley (changes requested):

core/backend/src/BlobContainerService.ts
kabentley: @swwilson-bsi I don't understand the issue?

MichaelSwigerAtBentley: Wouldn't the service be able to know where the container is without provider? If not, then we would need a 'Provider' for delete() as well, correct?

wgoehrig: @kabentley IMHO, the service must be the ultimate arbiter of which container lives where.  This is why there's no longer the option to specify a provider when creating a container.

I believe the container ID must be globally unique, and it is the container service's responsibility to always know which container is hosted by which provider, and always look that up/return that in each GET response.

I'll defer to @swwilson-bsi to comment on the particulars of his issue with this, but from the POV of the service, including a provider in _any_ request is completely unnecessary.

kabentley: That's wonderful if the service will be able to determine the storageType from baseUri/ContainerId (I didn't presume that). In that case this change is fine and it is not necessary to supply (or return) for `requestToken`,  or `delete`. 

But the point is that while the client can't *request* a specific provider, it has to *know* the provider so it can use the appropriate apis. Therefore 'create' needs to return the `storageType` and the client needs to save it as part of the container's "address".  That change should be part of this PR, right?

kabentley: Actually, i think it will be necessary to allow apps creating containers to specify the types of providers they support, and probably a preferred order of preference. Some applications may only support one provider (e.g. Azure) or may work better with certain providers.

kabentley: @swwilson-bsi why/where are you calling `requestToken`? I don't understand why that's ever necessary for ViewStores.

kabentley (approved):

wgoehrig (approved):

PR #5915, opened 2023-08-25
Added Schema Merger class

StefanApfel-Bentley:
First version of Schema Merger class.

#5879


StefanApfel-Bentley:
Invalid configuration used. That PR is replaced with this: https://github.com/iTwin/itwinjs-core/pull/5982

PR #5918, opened 2023-08-28
refresh write lock expiry time when calling acquireWriteLock as the same user

nick4598:
imodel-native: https://github.com/iTwin/imodel-native/tree/nick/exposebcvkvtime

Removed the writeLockExpires on ContainerInternal object and instead added it as a getter to the container object in imodel-native. This lets us actually read the expiry time from the bcv_kv table which is a more accurate test that the containerinternal timestamp.

### DIFF
--- a/full-stack-tests/backend/src/integration/CloudSqlite.test.ts
+++ b/full-stack-tests/backend/src/integration/CloudSqlite.test.ts
@@ -163,6 +163,24 @@ describe("CloudSqlite", () => {
     db.close();
     container.disconnect({ detach: true });
   });
+  it("Should refresh write lock expiry time if withWriteLock called twice with same user", async () => {
+    const container = testContainers[0];
+    container.connect(caches[1]);
+    let writeLockExpiryTimeNoWriteLock = container.writeLockExpires; // Should be empty string when no write lock.
+    expect(writeLockExpiryTimeNoWriteLock).to.equal("");
+    await CloudSqlite.withWriteLock({user: "testuser", container}, async () => {
+      const firstWriteLockExpiryTime = Date.parse(container.writeLockExpires);
+      await new Promise((resolve) => setTimeout(resolve, 500)); // sleep 500ms so we get a new write lock expiry time.
+      await CloudSqlite.withWriteLock({user: "testuser", container}, async () => {
+        const secondWriteLockExpiryTime = Date.parse(container.writeLockExpires);
+        expect(secondWriteLockExpiryTime).to.be.greaterThan(firstWriteLockExpiryTime);
+        // secondWriteLockExpiryTime should only be a couple thousand ms in the future at most, so subtract 2000 ms and make sure its less than first.
+        expect(secondWriteLockExpiryTime - 2000).to.be.lessThan(firstWriteLockExpiryTime);
+      });
+    });
+    writeLockExpiryTimeNoWriteLock = container.writeLockExpires; // Should be empty string when no write lock.
+    expect(writeLockExpiryTimeNoWriteLock).to.equal("");
+  });
 
   it("cloud containers", async () => {
     expect(undefined !== caches[0]);
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -836,6 +836,8 @@ export namespace CloudSqlite {
     export interface CachedDbProps {
         readonly dirtyBlocks: number;
         readonly localBlocks: number;
+        readonly nClient: number;
+        readonly nPrefetch: number;
         readonly state: "" | "copied" | "deleted";
         readonly totalBlocks: number;
         readonly transactions: boolean;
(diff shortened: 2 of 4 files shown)


full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -163,6 +163,24 @@ describe("CloudSqlite", () => {
...
     container.disconnect({ detach: true });
   });
+  it("Should refresh write lock expiry time if withWriteLock called twice with same user", async () => {
+    const container = testContainers[0];
+    container.connect(caches[1]);
+    let writeLockExpiryTimeNoWriteLock = container.writeLockExpires; // Should be empty string when no write lock.
+    expect(writeLockExpiryTimeNoWriteLock).to.equal("");
+    await CloudSqlite.withWriteLock({user: "testuser", container}, async () => {
+      const firstWriteLockExpiryTime = Date.parse(container.writeLockExpires);
+      await new Promise((resolve) => setTimeout(resolve, 500)); // sleep 500ms so we get a new write lock expiry time.
```
kabentley, line 173:
```suggestion
    await BeDuration.wait(500);
```

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -163,6 +163,24 @@ describe("CloudSqlite", () => {
...
+      await CloudSqlite.withWriteLock({user: "testuser", container}, async () => {
+        const secondWriteLockExpiryTime = Date.parse(container.writeLockExpires);
+        expect(secondWriteLockExpiryTime).to.be.greaterThan(firstWriteLockExpiryTime);
+        // secondWriteLockExpiryTime should only be a couple thousand ms in the future at most, so subtract 2000 ms and make sure its less than first.
+        expect(secondWriteLockExpiryTime - 2000).to.be.lessThan(firstWriteLockExpiryTime);
```
kabentley, line 178: I fear this will be the source of flaky tests. You can't really ensure that the wait won't take much longer. I would just remove this test.

nick4598: What about 20 seconds instead of 2? Atleast if this test ever fails, it will be very obvious that this line was the culprit if it is.

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
nick4598: Changed it

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
kabentley: I'm not sure what the test accomplishes. Comparing times in tests is always tricky and the first test makes sure it changes, which really is sufficient.

nick4598: I suppose it accomplishes that we don't accumulate expiry time, i.e. add the full expiry time of the container's write lock instead of just refreshing it.

kabentley: I'm not sure I understand what you're saying. Refreshing the expiry means to extend the write lock to expire after the timeout duration from the time the "refresh" is called. 

WriteLockExpiration times are tricky because they don't use the local computer time. Instead they use the time of the response from the server (so that they're not misinterpreted by others with clock drift). When running tests using the storage emulator, they *are* correlated (because there is no server). All of this isn't testing what you think it is. All you can tell reliably is that it changed.

nick4598: Assuming that the containers default write lock expiry time is 1 hour, then I am asserting with that testcase that calling withWriteLock multiple times in a short timeframe simply refreshes the time (adds the small gap in time between withWriteLock call 1 and withWriteLock call 2) and doesn't give us a write lock expiry time of 2 hours instead. I would be willing to go even higher than 20 seconds, even 10 minutes, or do you still predict that there is potential for issues there?

kabentley: The expiry time after a refresh doesn't depend in any way on the previous expiry time, so there's really no point in this test.

nick4598: You're right that the expiry time after a refresh doesn't depend in any way on the previous expiry time, atleast the way the code is currently written. The two times though should be fairly reasonably close to each other and what I'm really trying to test is that the client side behavior of adding the container's expiry time (default 1 hour) to the current server time works instead of somehow adding it to the already present expiry time in the writeLock. I agree that it seems unlikely, but it doesn't hurt to have it present just incase this area of the code gets changed at some point in the future.

If you still disagree, I will remove the testcase.

kabentley: tests that compare times are problematic. But you're right if it's on the order of minutes it isn't likely to be tripped up by test machine vagaries. As long as you're confident this is worth testing, do so.

Issue #5536, opened 2023-05-17
Log changeset download time, changeset apply time, checkpoint download time


nick4598:
Alternatively: Some of this logging should possibly go into the imodels-clients repo: https://github.com/iTwin/imodels-clients. Any code to download checkpoints and download changesets will take place in packages provided by the above repo.

iTwin.js Core could/should log changeset apply time.

PR #5930, opened 2023-09-01
Adds logging for changeset application and V2 Checkpoint download times.

Ben-Pusey-Bentley:
Companion PR with [this](https://github.com/iTwin/imodels-clients/pull/191).  Addresses #5536, along with its companion PR. The Imodels clients package will need to be updated when the new version with the linked PR releases.


pmconne (approved):

core/backend/src/CheckpointManager.ts
```diff
@@ -307,7 +307,10 @@ export class CheckpointManager {
   private static async doDownload(request: DownloadRequest): Promise<ChangesetId> {
     try {
       // first see if there's a V2 checkpoint available.
+      const stopwatch = new StopWatch(`[${request.checkpoint.changeset.id}]`, true);
+      Logger.logInfo(loggerCategory, `Starting download of V2 checkpoint with id ${stopwatch.description}`);
       const changesetId = await V2CheckpointManager.downloadCheckpoint(request);
+      Logger.logInfo(loggerCategory, `Applied changeset with id ${stopwatch.description} (${stopwatch.elapsedSeconds} seconds)`);
```
nick4598, line 313: This should say "Finished download of v2 checkpoint" or something along those lines, no?

core/backend/src/CheckpointManager.ts
```diff
@@ -349,7 +352,10 @@ export class CheckpointManager {
           const accessToken = checkpoint.accessToken;
           const toIndex = checkpoint.changeset.index ??
             (await IModelHost.hubAccess.getChangesetFromVersion({ accessToken, iModelId: checkpoint.iModelId, version: IModelVersion.asOfChangeSet(checkpoint.changeset.id) })).index;
+          const stopwatch = new StopWatch(`[${checkpoint.changeset.id}]`, true);
+          Logger.logInfo(loggerCategory, `Starting application of changeset with id ${stopwatch.description}`);
           await BriefcaseManager.pullAndApplyChangesets(db, { accessToken, toIndex });
+          Logger.logInfo(loggerCategory, `Applied changeset with id ${stopwatch.description} (${stopwatch.elapsedSeconds} seconds)`);
```
nick4598, line 358: Since there can be multiple changesets to apply, this timer could be misleading. I think we could apply 100 changesets and we would get the impression that the one singular changeset id that you're logging took the time to apply 100 changesets when it would potentially only account for 1/100th of that time.

Maybe we should log each individual changeset apply time? Which would be inside of pullAndApplychangesets itself ?

core/backend/src/CheckpointManager.ts
kabentley: yes, this message is not correct. At this point you have downloaded a checkpoint and have not applied any changesets. BTW, web backends can NEVER apply changesets (they may only show Named Versoins that must have a checkpoint), so this whole concept of "updating to a version" is only relevant for editable backends.

I see this PR has already merged, but this should be changed.

PR #5933, opened 2023-09-01
Dan/fix finalize release

DanRod1999:
Fixes error wish push command by including `origin` in cmd.

Fixes the  removal of extra newline characters to be more readable.

Upgrades to action/checkoutV3

Add authtoken to checkout step. From testing in a dummy repo, this token also appears to be what authorizes the push later in the pipeline. Also tested pushing to a protected branch with an authorized user and it worked correctly.

Changes trigger from tag to on commits to branches with `release/*` and updates to files with the path `**/CHANGELOG.md`, this is because when the workflow is triggered by a tag, the default branch that gets checkout is in a detached head state, and checking out to a new branch doesn't correct this. I believe using git switch might work, however this is easier because I use the branch name to detect what the target branch for the change logs should be. The branch name when a tag is checked out is not the same name as the source branch, so it would require different logic to get the target branch. Switching to this was also makes it simpler for running the script locally.

### DIFF
--- a/.github/workflows/finalize-release.yaml
+++ b/.github/workflows/finalize-release.yaml
@@ -16,8 +16,10 @@ jobs:
     name: Cherry-pick Changelogs
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
+          token: ${{ secrets.IMJS_ADMIN_GH_TOKEN }}
+          ref: ${{ github.ref }} # checkouts the branch that triggered the workflow
           fetch-depth: 0
       - name: Set Git Config
         run: |
--- a/.github/workflows/finalize-release.yaml
+++ b/.github/workflows/finalize-release.yaml
@@ -8,8 +8,10 @@ name: Finalize Release
 on:
   workflow_dispatch:
   push:
-    tags:
+    branches:
     - 'release/*'
+    paths:
+    - '**/CHANGELOG.md'
 
 jobs:
   finalize:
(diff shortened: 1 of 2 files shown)


.github/workflows/finalize-release.yaml
```diff
@@ -8,16 +8,20 @@ name: Finalize Release
 on:
   workflow_dispatch:
   push:
-    tags:
+    branches:
```
aruniverse, line 11: does this mean anytime a commit is pushed to any release branch itll run this workflow??
or is this only when a commit is pushed to this branch with changes to any chnagelog.md file?

aruniverse:
before this pr merges to master, did the one you merged to the last release branch work?

.github/workflows/finalize-release.yaml
DanRod1999: Commit to branch, and change to a changelog.md file. These two fields are and'd (according to the GH docs), so both are required for this workflow to run. However, doing something like `tags: ... branches: ...` is or'd, so that was not an option.

DanRod1999:
> before this pr merges to master, did the one you merged to the last release branch work?

No, that did not work, which is why I had to make these changes. I Believed I could not directly merge this into the release/4.1.x branch because only the triggers from the default branch are considered. However, I was trying to double check this, and I believe I am wrong. `on push:` appears to be one of the triggers that is not required to be on the default branch. 

Maybe its best to draft this, and open a PR into release/4.1.x with these changes? Either way master will need this merged in eventually, but that can be done after 4.1.x gets is.

aruniverse:
> Either way master will need this merged in eventually, but that can be done after 4.1.x gets is.

i dont want to merge broken pipelines into master
lets make sure it works on 4.1.x first before we do anything on master

aruniverse (approved):

PR #5935, opened 2023-09-01
Add querybcvstats to get some information about state of the cachefile

nick4598:
imodel-native: https://github.com/iTwin/imodel-native/tree/nick/cachefilemetrics

This will be used in the daemon prometheus client.

### DIFF
--- a/core/backend/src/CheckpointManager.ts
+++ b/core/backend/src/CheckpointManager.ts
@@ -168,7 +168,7 @@ export class V2CheckpointManager {
         }
       }
 
-      this._cloudCache = CloudSqlite.CloudCaches.getCache({ cacheName: this.cloudCacheName, cacheDir });
+      this._cloudCache = CloudSqlite.CloudCaches.getCache({ cacheName: this.cloudCacheName, cacheDir, cacheSize: "50G" });
 
       // Its fine if its not a daemon, but lets log an info message
       if (!this._cloudCache.isDaemon)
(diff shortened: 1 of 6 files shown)


khanaffan (approved):

core/backend/src/CheckpointManager.ts
```diff
@@ -168,7 +168,7 @@ export class V2CheckpointManager {
         }
       }
 
-      this._cloudCache = CloudSqlite.CloudCaches.getCache({ cacheName: this.cloudCacheName, cacheDir });
+      this._cloudCache = CloudSqlite.CloudCaches.getCache({ cacheName: this.cloudCacheName, cacheDir, cacheSize: "50G" });
```
aruniverse, line 171: 50 what?

aruniverse (approved):

core/backend/src/CheckpointManager.ts
nick4598: 50 gigabytes, should I add that as a comment?

aruniverse: why is it not `50GB` then?
a comment would help

pmconne: It's documented on `CreateCloudCacheArg.cacheSize`.

```
    /** A string that specifies the maximum size of the CloudCache. It should be a number followed by "K",
     * "M" "G", or "T". Default is "10G". */
    cacheSize?: string;
```

pmconne: I don't see a reason to have to document it everywhere it's used too.

Issue #5587, opened 2023-06-02
ExternalSourceAspect.scope should be RelatedElement | undefined


simsta6:
**Describe the bug**
In `ExternalSourceAspect` class scope is always defined, but real iModels sometimes external source aspects may have it as undefined

**Expected behavior**
`ExternalSourceAspect.scope` should have type of `RelatedElement | undefined`

DanRod1999:
@simsta6, whats the severity of this issue? Do you need a fix for this now, or is this something that can wait until 5.0. Due to this being a breaking change, if this were to be done now it would require implementing a new scope property and deprecating the old one (or something similar), but I cannot directly change `scope` to optional in 4.x.x

aruniverse:
Closing as duplicate of https://github.com/iTwin/itwinjs-core/issues/4631

PR #5941, opened 2023-09-05
allow ExternalSourceAspect scope to be undefined

DanRod1999:
From issue: https://github.com/iTwin/itwinjs-core/issues/5587
Do not know if this change also requires the ExternalSourceAspectProps to be  `RelatedElementProps | undefined`

### DIFF
--- a/core/backend/src/ElementAspect.ts
+++ b/core/backend/src/ElementAspect.ts
@@ -147,7 +147,7 @@ export class ExternalSourceAspect extends ElementMultiAspect {
    * @note Warning: in a future major release the `scope` property will be optional, since the scope is intended to be potentially invalid.
    *       all references should treat it as potentially undefined, but we cannot change the type yet since that is a breaking change.
    */
-  public scope: RelatedElement;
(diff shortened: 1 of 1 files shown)


core/backend/src/ElementAspect.ts
```diff
@@ -147,7 +147,7 @@ export class ExternalSourceAspect extends ElementMultiAspect {
    * @note Warning: in a future major release the `scope` property will be optional, since the scope is intended to be potentially invalid.
    *       all references should treat it as potentially undefined, but we cannot change the type yet since that is a breaking change.
    */
-  public scope: RelatedElement;
+  public scope: RelatedElement | undefined;
```
aruniverse, line 150: this is potentially breaking, please investigate the ramifications

pmconne: No "potential" about it - it violates our [API support policies](https://www.itwinjs.org/learning/api-support-policies/).
Consider deprecating it and adding a new property.

pmconne (changes requested):

PR #5943, opened 2023-09-06
Add taskId and changedFiles to changesetprops

nick4598:


core/common/src/ChangesetProps.ts
```diff
@@ -77,6 +77,10 @@ export interface ChangesetProps {
   userCreated: string;
   /** The size, in bytes, of this changeset */
   size: number;
+  /** id of the synchronization task */
```
nick4598, line 80: If reviewer would prefer, I could also export an interface 'Synchronizationinfo' which contains these two new properties. I thought about doing this, but decided not to because 2/4 of the imodels-clients already define their own synchronizationInfo. The 2/4 which define their own syncInfo do not have a dependency on core-common at the moment, so we wouldn't be able to go to the same type. The other 2 packages (imodels-access-backend specifically) would have to return the syncInfo that we defined which I don't really like, but open to changing it.

I still see a benefit though in defining syncinfo in core-common since it would keep sync related properties grouped.

nick4598: The current interface defined in imodels-clients repo is:

export interface SynchronizationInfo {
  /** Id of the synchronization task. */
  taskId: string;
  /** List of files that were processed by the synchronization. */
  changedFiles: string[] | null;
}

So I guess that wouldn't apply here?

kabentley: Can you add some context? I don't understand why these are properties of a changeset. Where will they be stored. The whole concept of a "synchronization task" presumes there's an external source for changes, which hopefully will become rare. What do you mean by "become required"?

nick4598: I updated the description of the PR to "taskId and changedFiles are useful properties to help determine the cause of potential issues in connector jobs. In iModels where there can be many connections( sources of information for the iModel ), having the taskId as part of the changeset identifies which connection was responsible for creating this changeset. "

The hub stores these properties when applicable in a database. On further thought I think I am wrong about 'becoming required', because as you mention not every changeset will have this information.

Also adding @c-macdonald to provide any further context, if my description is lacking.

c-macdonald: These are set through the BridgePropertiesPtr which is a parameter in the PushChangeSetArguments::Create method.  So, yes, only applicable to the native/batch connectors.  As for the typescript API to retrieve the information, I don't need them to be returned as first class properties.  I (or whomever) would just be looking at the JSON retrieved and would manually find the data.  I wouldn't need it parsed or dumped into a db or anything.

PR #5944, opened 2023-09-06
fix finalize release 4.1.x

DanRod1999:
Fixes error with push command by including `origin` in cmd.

Fixes the removal of extra newline characters to be more readable.

Upgrades to action/checkoutV3

Add authtoken to checkout step. From testing in a dummy repo, this token also appears to be what authorizes the push later in the pipeline. Also tested pushing to a protected branch with an authorized user and it worked correctly.

Changes trigger from tag to on commits to branches with release/* and updates to files with the path **/CHANGELOG.md, this is because when the workflow is triggered by a tag, the default branch that gets checkout is in a detached head state, and checking out to a new branch doesn't correct this. I believe using git switch might work, however this is easier because I use the branch name to detect what the target branch for the change logs should be. The branch name when a tag is checked out is not the same name as the source branch, so it would require different logic to get the target branch. Switching to this was also makes it simpler for running the script locally.

### DIFF
--- a/.github/workflows/automation-scripts/update-changelogs.mjs
+++ b/.github/workflows/automation-scripts/update-changelogs.mjs
@@ -28,9 +28,13 @@ let currentBranch = await $`git branch --show-current`;
 let commitMessage = await $`git log --format=%B -n 1`;
 
 // remove extra null and new line characters from git cmds
-targetBranch = String(targetBranch).slice(0, -1);
-currentBranch = String(currentBranch).slice(0, -1);
-commitMessage = String(commitMessage).slice(0, -2);
+targetBranch = String(targetBranch).replace(/\n/g, '');
+currentBranch = String(currentBranch).replace(/\n/g, '');
+commitMessage = String(commitMessage).replace(/\n/g, '');
+
+console.log(`target branch: ${targetBranch}`);
+console.log(`current branch: ${currentBranch}`);
+console.log(`commit msg: ${commitMessage}`);
 
 if (targetBranch === `origin/${currentBranch}`) {
   console.log("The current branch is the latest release, so the target will be master branch")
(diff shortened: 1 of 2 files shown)


DanRod1999:
ran a final test in the workflow itself, and created this branch to make sure everything is working as intended. This would be an example of 4.1.x changes going into master.
https://github.com/iTwin/itwinjs-core/compare/master...finalize-release-Test

I will delete this branch after this PR is merged.

DanRod1999:
Also based on the changes to the triggers I realized that when this runs in 3.7.x, it will push updated logs to release/4.1.x, and then because 4.1.x update changelog files, it will automatically trigger this workflow again to update master branch.

.github/workflows/automation-scripts/update-changelogs.mjs
```diff
@@ -63,13 +69,12 @@ await $`rush publish --regenerate-changelogs`;
...
 // targetBranch = "finalize-release-X.X.X"
 /*********************************************************************/
-targetBranch = targetBranch.replace("origin/", "");
 await $`git add .`;
 await $`git commit -m "${commitMessage} Changelogs"`;
 await $`rush change --bulk --message "" --bump-type none`;
 await $`git add .`;
 await $`git commit --amend --no-edit`;
-await $`git push HEAD:${targetBranch}`;
+await $`git push origin HEAD:${targetBranch}`;
```
aruniverse, line 77: do you need this `HEAD:` syntax?

aruniverse (approved):

.github/workflows/automation-scripts/update-changelogs.mjs
DanRod1999: i don't think so, I think its usually if your local branch name isn't the same as the upstream branch.
But because the branch is changing during the workflow i thought stating the branch would be best, and I thought this was the most appropriate way to write it.
git push origin HEAD:master
    Push the current branch to the remote ref matching master in the origin repository. This form is convenient to push the current branch without thinking about its local name.

Pretty sure it could be written a few other ways, but i thought this would be the most explicit

PR #5948, opened 2023-09-07
Deprecate default view Id

pmconne:
This property was intended to provide a reasonable default view for an iModel, but in practice each viewing application will have its own definition of "reasonable" and most existing iModels' default views are far from reasonable choices for most applications. Besides, we expect future connectors to avoid embedding views in iModels at all.


pmconne:
```
##[error]Handle leak detected. Node was still running 5 seconds after tests completed.
Try running with the DEBUG_LEAKS env var set to see open handles.
```

@wgoehrig @aruniverse can we either get somebody to find + fix all these alleged transient handle leaks, or disable the check to stop flaking out our builds?

markschlosseratbentley (approved):

nick4598 (approved):

PR #5952, opened 2023-09-07
Reorganize geomlib curve files.

saeeedtorabi:
- Rename CurveChain.ts -> CurveTypes.ts and explain why `CurveChain` lives in CurveCollection.ts rather than its own class (answer: to avoid circular dependency).
- Move `ConsolidateAdjacentCurvePrimitivesOptions` to RegionOps.ts, where it is used.


dassaf4 (approved):

dassaf4:
I just noticed this PR is merging into another one of your branches. Is this intended?

PR #5953, opened 2023-09-07
Fixes issues with debuggers on display-test-app

### LINKED ISSUE #5929: display-test-app 'npm run start' does not pick up frontend changes
**Describe the bug**
If you run display-test-app using `npm run start`, it will not pick up local changes to the frontend code when you rebuild and run it again.

**To Reproduce**

1. make a git clone of iTwin.js core
2. `rush install`
3. `rush build --to display-test-app`
4. `cd test-apps/display-test-app` then `npm run start`
5. Open debugger and observe that Clipping.ts (in frontend) can be opened with ctrl+p.
6. Edit a frontend file like glsl/Clipping.ts
7. Terminate display-test-app, then do: `rush build --to display-test-app`
8. `npm run start`
9. Open debugger and observe that Clipping.ts (in frontend) can no longer be found with ctrl+p.
10. Search for a js chunk file with code from Clipping.ts. That js chunk file will not have the changes we made earlier!
11. Quit the app.
12. `npm run start:electron`
13. Observe that the application is running with the latest changes to Clipping.ts. However, you might still not be able to find Clipping.ts in the debugger with ctrl+p.

**Expected behavior**
`npm run start` should provide .ts files in the debugger for frontend files; it should also pick up local changes to frontend source.

- OS: Windows
- Browser: Electron
- iTwin.js Version: master

**Additional context**

- `rush update --full` then rebuilding makes `npm run start` use the latest changes.
- In the midst of all my experimentation, I noticed at one point, something called frontend-vite getting built, not just regular frontend. May be related.
- When it becomes broken, navigating to the localhost __inspect page shows that no frontend files are found.

Ben-Pusey-Bentley:
Fixes a couple of issues in the vite config file for the display-test-app that were causing the debugger to not display correctly. First, the `build.sourcemap` value has been updated so that changed files can be seen in the debugger source. Second, the `optimizeDeps.force` value has been set, so that changes to those files are properly displayed in the debugger. Note that setting the force value is not a perfect solution, as this change causes the vite cache to be rebuilt entirely on each rebuild, which has a negative effect on performance. However, there is a known issue with vite where `optimizeDeps.include` does not work as intended on monorepos, [here](https://github.com/vitejs/vite/issues/14099). As a result, this is our best solution for known. Once the issue in vite is resolved, we should update to remove the force flag. Closes #5929.

### DIFF
--- a/test-apps/display-test-app/vite.config.ts
+++ b/test-apps/display-test-app/vite.config.ts
@@ -1,7 +1,7 @@
 /*---------------------------------------------------------------------------------------------
-* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
-* See LICENSE.md in the project root for license terms and full copyright notice.
-*--------------------------------------------------------------------------------------------*/
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
 import { defineConfig, loadEnv, searchForWorkspaceRoot } from "vite";
 import envCompatible from "vite-plugin-env-compatible";
 import browserslistToEsbuild from "browserslist-to-esbuild";
@@ -122,6 +122,7 @@ export default defineConfig(() => {
           }),
         ],
       },
+      force: true, // forces cache dumps on each rebuild. should be turned off once the issue in vite with monorepos not being correctly optimized is fixed.
       // overoptimized dependencies in the same monorepo (vite converts all cjs to esm)
       include: [
         "@itwin/core-common", // for opening iModel error
--- a/test-apps/display-test-app/vite.config.ts
+++ b/test-apps/display-test-app/vite.config.ts
@@ -54,7 +54,7 @@ export default defineConfig(() => {
     logLevel: process.env.VITE_CI ? "error" : "warn",
     build: {
       outDir: "./lib",
-      sourcemap: !!process.env.VITE_CI, // append to the resulting output file if not running in CI.
+      sourcemap: !process.env.VITE_CI, // append to the resulting output file if not running in CI.
       minify: false, // disable compaction of source code
       target: browserslistToEsbuild(), // for browserslist in package.json
       commonjsOptions: {
--- a/test-apps/display-test-app/vite.config.ts
+++ b/test-apps/display-test-app/vite.config.ts
@@ -122,7 +122,7 @@ export default defineConfig(() => {
           }),
         ],
       },
-      force: true, // forces cache dumps on each rebuild. should be turned off once the issue in vite with monorepos not being correctly optimized is fixed.
+      force: true, // forces cache dumps on each rebuild. should be turned off once the issue in vite with monorepos not being correctly optimized is fixed. Issue link: https://github.com/vitejs/vite/issues/14099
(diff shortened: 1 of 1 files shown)


aruniverse:
@markschlosseratbentley @DStradley @MarcNeely @AJMigliore 
plz test this out

markschlosseratbentley:
> @markschlosseratbentley @DStradley @MarcNeely @AJMigliore plz test this out

`npm run start` appears to work properly. I can open up the debugger and do `ctrl+p` and find the proper .ts files.  I can edit the code, terminate DTA, do `rush build --to display-test-app`, restart DTA, and it picks up my changes. `ctrl+p` continues to work and find the proper .ts files.

`npm run start:electron` does not appear to allow me to find the proper .ts files using `ctrl+p`.  Instead, it finds the corresponding .js files.

`npm run start`, furthermore, does not allow us to keep DTA running, edit some frontend code (for example), then do `npm run build` in core/frontend, and have DTA hot reload those changes.  This worked at one point.  Not sure if it's a vite problem or something that broke a while ago.

Ben-Pusey-Bentley:
Hi, @markschlosseratbentley, can you explain a bit more about how you are running `npm run start:electron`? I cannot recreate the behavior that you are describing. If I run the command on its own, no files at all appear in cntrl + p, because there is no sourcemap available without the mock web server. If I run it alongside `npm run start:webserver`, then I get the same behavior as `npm run start`, because that command is just a combination of those two.

pmconne:
If resolving @markschlosseratbentley's remaining issues will take more than a day or so, please consider merging this and following up with a subsequent PR. The current behavior is quite bad, and your changes make it much better.

markschlosseratbentley:
> Hi, @markschlosseratbentley, can you explain a bit more about how you are running `npm run start:electron`? I cannot recreate the behavior that you are describing. If I run the command on its own, no files at all appear in cntrl + p, because there is no sourcemap available without the mock web server. If I run it alongside `npm run start:webserver`, then I get the same behavior as `npm run start`, because that command is just a combination of those two.

So maybe `npm run start:electron` is supposed to not be debuggable?  I never used it much in the past, but I know @DStradley and @MarcNeely do.  It's supposed to be running a "prod/release" version of the app.  It definitely is starting a backend, though, because you can open an iModel, etc.  See screenshot at bottom of this comment.  I am able to ctrl+p and find .js files, but not .ts files.  Are you saying you cannot even find .js files with `npm run start:electron`?

I agree with @pmconne that this PR should get merged in ASAP since the current behavior results in not being able to debug at all, or roll back in time before vite.

image

Ben-Pusey-Bentley:
The screenshot bellow shows what I get. The application doesn't show any of the iModel opening features, and using cntrl + p shows no files at all. Also need an approving review by a code owner to merge this PR.
image

markschlosseratbentley (approved):

MarcNeely:
Yes, this is definitely good enough to merge in.  As for not being able to see the .js files with 'run start:electron', maybe you just didn;t wait long enough.  Some times when I go in to try to open up a file it takes a few seconds before it can access them.

DStradley:
With regard to previous behavior, we were working with an old build earlier this week (pre-vite) and using `npm run start:electron`, and it opened and showed the .ts files (not .js) in the debugger and worked fine for debugging.

PR #5954, opened 2023-09-07
export SchemaFileUtility

christophermlawson:
The SchemaFileUtility class allows writing in memory schemas to an xml file. It was requested to expose this class so that users can leverage this function.


calebmshafer (approved):
LGTM

PR #5956, opened 2023-09-08
Add option to premultiply model display transform

pmconne:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/847
To make your `ModelDisplayTransformProvider` produce a transform to be premultiplied with the model's base transform:

```
  public getModelDisplayTransform(_modelId) {
    return { transform: Transform.whatever(), premultiply: true };
  }
```

For post-multiply, omit `premultiply` or set it to false.

I considered less-breaking return types such as `Transform & { premultiply?: boolean }` or `Transform | ModelDisplayTransform` but they are error-prone and awkward to use; since this is a beta API I preferred to change it to be explicit.


core/frontend/src/ViewState.ts
```diff
@@ -55,17 +55,26 @@ export interface ExtentLimits {
...
-/** Interface adopted by an object that wants to apply per-model display transforms.
- * A model's display transform is applied when rendering the model in a [[Viewport]].
+/** A [Transform]($core-geometry) supplied by a [[ModelDisplayTransformProvider]] to be applied to a model when displaying it in a [[Viewport]].
+ * @beta
+ */
+export interface ModelDisplayTransform {
+  /** The transform to be applied to the model. */
+  transform: Transform;
+  /** If `true`, [[transform]] is pre-multiplied with (i.e., appled before) the model's base transform.
+   * Otherwise, the display transform will instead be pre-multiplied with (i.e., applied after) the base transform.
```
bbastings, line 65: Should say post-multiplied for false?

bbastings (approved):

core/frontend/src/ViewState.ts
pmconne: Thanks! Fixed.

GytisCepk (approved):

ben-polinsky (approved):

PR #5960, opened 2023-09-08
Fix schedule script symbology issue.

MarcNeely:
Fixes schedule script issue where element symbology appears wrong when the initial view is loaded and persists during the animation.


ben-polinsky (approved):

pmconne (commented):
Nice find, thanks. Can you run `rush change`?

ben-polinsky:
Perhaps a unit test to cover this?

MarcNeely:
Not sure how easily we could cover this situation with a unit test, but we will certainly add this test case to our nightly ImageTests run.

ben-polinsky:
> Not sure how easily we could cover this situation with a unit test, but we will certainly add this test case to our nightly ImageTests run.

Sounds good

markschlosseratbentley (approved):

PR #5962, opened 2023-09-08
Stabilize watchForChanges

pmconne:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/845 (the geometry synchronization problem, not the project extents synchronization problem).
Previously, a reader process would watch sqlite's WAL file to detect when a writer process modified the briefcase.
This is unreliable - we may be notified of the change to the file before the changes can be read.
Instead:
- When opening a read-only briefcase /path/to/my.bim with `watchForChanges=true`, create /path/to/my.bim-watch if it does not exist, and watch it for changes. I call this the briefcase's "watch file".
- After committing to a briefcase, modify the timestamp on my.bim-watch if it exists, to signal that the changes are ready to be read by a reader process.

I could delete the watch file when closing the read-only briefcase, but that would break a hypothetical second briefcase that is also watching the file.
I could record the number of current watchers into the watch file, and delete it only when closing the only remaining watcher, but that would be prone to all sorts of fun race conditions, and if a watcher process terminated without closing the briefcase the watcher count would become inaccurate.
Instead, I just leave the empty file in perpetuity once created.

### DIFF
--- a/common/changes/@itwin/core-backend/pmc-stabilize-watch-for-changes_2023-09-08-18-32.json
+++ b/common/changes/@itwin/core-backend/pmc-stabilize-watch-for-changes_2023-09-08-18-32.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/core-backend",
-      "comment": "Make watching for briefcase changes more reliably.",
+      "comment": "Make watching for briefcase changes more reliable.",
       "type": "none"
     }
   ],
--- a/common/changes/@itwin/core-backend/pmc-stabilize-watch-for-changes_2023-09-08-18-32.json
+++ b/common/changes/@itwin/core-backend/pmc-stabilize-watch-for-changes_2023-09-08-18-32.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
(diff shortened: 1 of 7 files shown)


common/changes/@itwin/core-backend/pmc-stabilize-watch-for-changes_2023-09-08-18-32.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Make watching for briefcase changes more reliably.",
```
nick4598, line 5: reliable instead of reliably?

nick4598 (approved):

core/backend/src/TxnManager.ts
```diff
@@ -292,6 +303,7 @@ export class TxnManager {
 
   /** @internal */
   protected _onCommitted() {
+    this.touchWatchFile();
```
kabentley, line 306: i realize this PR is already committed, but I think you should only touch the watch file when the briefcase is closed. This will avoid the problem with undo/redo and is all we need. Touching the file on the undo/redo event won't work because we never look for the Txns going backwards.

PR #5964, opened 2023-09-11
Use Instance Query to get Aspects for abstract classes

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/813.

There was a performance issue when calling getAspects() on an abstract class.
The function has been modified to use instance queries in order to get all aspects in a single go.

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -3088,10 +3088,14 @@ export namespace IModelDb {
         insertElement(elProps: ElementProps): Id64String;
         // @internal
         _queryAspects(elementId: Id64String, fromClassFullName: string, excludedClassFullNames?: Set<string>): ElementAspect[];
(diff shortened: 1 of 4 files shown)


RohitPtnkr1996:
This PR depends on https://github.com/iTwin/imodel-native/pull/446, which will cause the builds on this PR to fail until it gets merged.

pmconne:
> This PR depends on [iTwin/imodel-native#446](https://github.com/iTwin/imodel-native/pull/446), which will cause the builds on this PR to fail until it gets merged.

Why not link the two PRs and change this one to target imodel02 instead of master, so we can confirm the builds do indeed succeed with the native PR's changes?

RohitPtnkr1996:
Yeah, good point, will do that.
Thanks Paul !

PR #5966, opened 2023-09-11
Use instance query to get aspects for abstract classes

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/813.

There was a performance issue when calling getAspects() on an abstract class.
The function has been modified to use instance queries in order to get all aspects in a single go.

imodel-native: https://github.com/iTwin/imodel-native/tree/affank/custom-instance-json

### DIFF
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -16,7 +16,7 @@ import {
 import {
   AxisAlignedBox3d, BRepGeometryCreate, BriefcaseId, BriefcaseIdValue, CategorySelectorProps, ChangesetIdWithIndex, ChangesetIndexAndId, Code,
   CodeProps, CreateEmptySnapshotIModelProps, CreateEmptyStandaloneIModelProps, CreateSnapshotIModelProps, DbQueryRequest, DisplayStyleProps,
-  DomainOptions, EcefLocation, ECSchemaProps, ECSqlReader, ElementAspectProps, ElementGeometryRequest, ElementGraphicsRequestProps, ElementLoadProps,
+  DomainOptions, EcefLocation, ECJsNames, ECSchemaProps, ECSqlReader, ElementAspectProps, ElementGeometryRequest, ElementGraphicsRequestProps, ElementLoadProps,
   ElementProps, EntityMetaData, EntityProps, EntityQueryParams, FilePropertyProps, FontId, FontMap, FontType, GeoCoordinatesRequestProps,
   GeoCoordinatesResponseProps, GeometryContainmentRequestProps, GeometryContainmentResponseProps, IModel, IModelCoordinatesRequestProps,
   IModelCoordinatesResponseProps, IModelError, IModelNotFoundResponse, IModelTileTreeProps, LocalFileName, MassPropertiesRequestProps,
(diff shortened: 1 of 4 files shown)


core/backend/src/IModelDb.ts
```diff
@@ -2015,6 +2015,65 @@ export namespace IModelDb { // eslint-disable-line no-redeclare
       return this._iModel.constructEntity<ElementAspect>(aspect);
     }
 
+    private static classMap = new Map<string, string>();
+
+    public _runInstanceQuery(sql: string, elementId: Id64String, excludedClassFullNames?: Set<string>): ElementAspect[] { // eslint-disable-line @typescript-eslint/naming-convention
```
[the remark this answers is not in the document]
pmconne: I see `_runInstanceQuery` can be called from two places. I still don't see a reason for it or `_queryAspectsUsingInstanceQuery` to be public - is there one?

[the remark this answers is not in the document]
RohitPtnkr1996: Nope, I will move the code into getAspects and make runInstanceQuery private.

core/backend/src/IModelDb.ts
```diff
@@ -2035,14 +2059,39 @@ export namespace IModelDb { // eslint-disable-line no-redeclare
      * @param aspectClassFullName Optionally filter ElementAspects polymorphically by this class name
      * @throws [[IModelError]]
      */
-    public getAspects(elementId: Id64String, aspectClassFullName?: string): ElementAspect[] {
-      if (undefined === aspectClassFullName) {
-        const uniqueAspects: ElementAspect[] = this._queryAspects(elementId, ElementUniqueAspect.classFullName);
-        const multiAspects: ElementAspect[] = this._queryAspects(elementId, ElementMultiAspect.classFullName);
-        return uniqueAspects.concat(multiAspects);
+    public getAspects(elementId: Id64String, aspectClassFullName?: string, excludedClassFullNames?: Set<string>): ElementAspect[] {
```
pmconne, line 2062: Document new parameter.

RohitPtnkr1996: Done !

khanaffan (changes requested):
@RohitPtnkr1996  in unit test you shoud compare actual instances instead of count(). Make sure the format matches between old and new implementation.

core/backend/src/IModelDb.ts
```diff
@@ -2033,16 +2057,42 @@ export namespace IModelDb { // eslint-disable-line no-redeclare
...
+      if (aspectClassFullName === undefined)
+        return this.runInstanceQuery(`SELECT $ FROM (
+          SELECT ECInstanceId, ECClassId FROM Bis.ElementMultiAspect WHERE Element.Id = :elementId
+            UNION ALL
+          SELECT ECInstanceId, ECClassId FROM Bis.ElementUniqueAspect WHERE Element.Id = :elementId) OPTIONS USE_JS_PROP_NAMES DO_NOT_TRUNCATE_BLOB ENABLE_EXPERIMENTAL_FEATURES`, elementId, excludedClassFullNames);
+
+      // Check if class is abstract
+      const fullClassName = aspectClassFullName.split(":");
```
GintV, line 2071: This is a breaking change. Previously, it would accept both "Schema:Class" and "Schema.Class". Now, if you pass "Schema.Class", fullClassName[1] is undefined and call to getECClassMetaData fails. I guess this should be changed to `aspectClassFullName.replace(".", ":")split(":")` to avoid breaking change

RohitPtnkr1996: PR raised to fix the breaking change scenario: https://github.com/iTwin/itwinjs-core/pull/6206

PR #5970, opened 2023-09-12
Improve 3D Tiles tileset content type detection

mborjesson-bentley:
Improves content type detection for 3D Tiles tilesets by creating an URL-object and uses the pathname instead.
This fixes loading issues when the tileset URLs contains query- or fragment-parts. I also made it case-insensitive and added a few tests.


pmconne (approved):

ben-polinsky (approved):

pmconne:
@ben-polinsky docs build has twice failed to download appui artifacts.

> ##[error]No builds currently exist in the pipeline definition supplied.

ben-polinsky:
> @ben-polinsky docs build has twice failed to download appui artifacts.
> 
> > ##[error]No builds currently exist in the pipeline definition supplied.

I saw and just raised #5972 to fix.

core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
```diff
@@ -115,7 +115,7 @@ export class RealityDataSourceTilesetUrlImpl implements RealityDataSource {
   }
 
   public getTileContentType(url: string): "tile" | "tileset" {
-    return url.endsWith("json") ? "tileset" : "tile";
+    return new URL(url, "https://localhost/").pathname.toLowerCase().endsWith("json") ? "tileset" : "tile";
```
aruniverse, line 118: @mborjesson-bentley 
why is the base url here localhost and not something like `window.location.origin`?

mborjesson-bentley: in this case we only look at pathname, so baseurl can be anything. the reason i chose localhost is because it's done the same way here:
https://github.com/iTwin/itwinjs-core/blob/1e97a4f18e1c7339863c5bd916b8e75a7dd39422/core/common/src/rpc/web/BentleyCloudRpcProtocol.ts#L55

aruniverse (approved):

mborjesson-bentley:
Replaced with #6064

PR #5971, opened 2023-09-12
Updated getAspects to use instance queries

RohitPtnkr1996:
Updated getAspects() to cache all derived classes of an abstract class irrespective of whether they are instantiated at runtime.
getAspects now logs an error message when no aspects exists for the element Id and aspect class name supplied.

Updated tests to check the data and format of the aspects returned by getAspects() which is now using instance queries.
As instance access is no longer experimental, updated getAspects() and relevant tests to remove ENABLE_EXPERIMENTAL_FEATURES option for instance queries.

imodel-native: https://github.com/iTwin/imodel-native/tree/rohitptnkr/remove-instance-access-from-experimental-features


pmconne (approved):
Awaiting @khanaffan's review.

khanaffan (approved):

PR #5979, opened 2023-09-13
ArcGIS Features and MapFeatureInfo tools bug fixes and improvements

mdastous-bentley:
Several fixes and improvements:
- MapFeatureInfo tool now keep decorations in sync with map layers visibility.
- Added parameter to specify pixel tolerance of `Viewport.getMapFeatureInfo`
- Added missing documentation for `Viewport.getMapFeatureInfo`
- Fixed ArcGIS feature driven symbology bug where properties with null values were preventing symbology to be applied.
- Made ArcGIS feature point symbol appear 25% bigger.
- Improved logging messages.


extensions/map-layers-formats/src/ArcGisFeature/ArcGisFeatureProvider.ts
```diff
@@ -399,7 +399,7 @@ export class ArcGisFeatureProvider extends ArcGISImageryProvider {
 
     const doFeatureInfoQuery = async (format: ArcGisFeatureFormat, outFields?: string, returnGeometry?: boolean) => {
       const infoUrl = this.constructFeatureUrl(quadId.row, quadId.column, quadId.level, format, "standard", queryEnvelope,
-        outFields, undefined, returnGeometry, toleranceWorld);
+        outFields, undefined, returnGeometry, tilePixelSize);
```
eringram, line 402: Why was this changed from `toleranceWorld`?

common/changes/@itwin/core-frontend/geo-mapfeatureInfo_fixes_2023-09-13-13-30.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Added parameter to specify pixel tolerance of Viewport.getMapFeatureInfo",
+      "type": "none"
```
ben-polinsky, line 6: Since you are augmenting a public api, both of these changes should be minor.

mdastous-bentley: Do I need to edit that manually?  `rush change` didn't offer me any option.

aruniverse: we never specify the change type since the repo is in lockstep.
we just need to be careful not to bp this into a patch

ben-polinsky: Ah right... thanks.

extensions/map-layers-formats/src/ArcGisFeature/ArcGisFeatureProvider.ts
mdastous-bentley: That's a good question...

There are two concepts:
1. **Query tolerance**: which is used to increase the area around a given point we want to query.
2. **maxAllowedOffset**: which is used for generalizing geometries returned by the query operation.

Before those changes, we were using the `tolerance` value to specify the `maxAllowedOffset` value, which was wrong.  We want the tolerance to be set around 7 pixels, and maxAllowedOffet to be around 1 pixel (bigger value means coarser geometries will be returned by the server).
Pixels values are converted in world coordinates before sending it to the ArcGIS service.

eringram (approved):

ben-polinsky (approved):

PR #5982, opened 2023-09-14
Added Schema Merger class

StefanApfel-Bentley:


common/api/ecschema-editing.api.md
```diff
@@ -1435,6 +1441,11 @@ export class SchemaItemMissing extends SchemaItemChange {
     toString(): string;
 }
 
+// @internal
+export class SchemaMerger {
```
pmconne, line 1445: Who are the intended/expected consumers of the SchemaMerger class? It's exported, but marked internal, but mentioned in public changelogs, but not imported or used by anything in the core repo.

StefanApfel-Bentley: Currently mainly iModel Transformation services.
I marked it as internal as it's a first version of that class and not complete yet. If I got you right, in that case, we should not export it in the ecschema-editing.api.md ?

pmconne: internal means this API is never intended to be used by anyone outside of the core repo.
It sounds like your new API should be marked alpha as it's eventually intended for use outside the core repo but in very early stages of development.
[How release tags work](https://www.itwinjs.org/learning/guidelines/release-tags-guidelines/).

StefanApfel-Bentley: Alright, I mark it as beta then. It's planned to let them play with it to figure it actually meets their requirements sufficiently.

core/ecschema-editing/src/Merging/SchemaMerger.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @beta
+ */
+export interface SchemaMergeContext {
+  readonly targetSchema: Schema;
+  readonly sourceSchema: Schema;
+}
+
+/**
+ * Class to merge two schemas together.
+ * See @see SchemaMerger.merge for details.
```
pmconne, line 27:
```suggestion
 * @see [[merge]] to merge the schemas.
```

Though it's unclear to me why users of this API must instantiate a stateless instance of a class with one method to merge two schemas.

`const schema = new SchemaMerger().merge(target, src);`
vs
`const schema = mergeSchemas(target, src);`

StefanApfel-Bentley: The class will get more methods or properties in future, to be able to configure the merging workflow.

core/ecschema-editing/src/Merging/CAClassMerger.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @internal
+ */
+export default async function mergeCAClasses(target: CustomAttributeClass, source: CustomAttributeClass, changes: ClassChanges) {
+  const mutableCAClass = target as MutableCAClass;
+  await mergeSchemaItemProperties(mutableCAClass, changes.propertyValueChanges, (item, propertyName, propertyValue) => {
+    switch(propertyName) {
+      case "appliesTo": {
+        const containerType = parseCustomAttributeContainerType(propertyValue);
+        if (containerType !== undefined) {
+          return item.setContainerType(containerType);
```
ColinKerr, line 21: container type is a flags enum so setting it to the new value is not really a 'merge', it would be more correct to add the new flag(s) to the existing ones.  This is OK for now but create a new issue to track proper merging of container types then you can resolve this comment

ColinKerr (approved):

core/ecschema-editing/src/Merging/CAClassMerger.ts
StefanApfel-Bentley: Added Issue: https://github.com/iTwin/itwinjs-core/issues/6014

PR #5998, opened 2023-09-15
Deprecate some appui-abstract utilities

ben-polinsky:
Continuing with the removal of appui-abstract from core, several utility types have been deprecated or removed:

- `UiEvent` is a duplicate of `BeUiEvent` 
- `UiEventDispatcher` is only consumed internally from `SyncUiEventDispatcher` @itwin/appui-react so it has been folded into that class.
- Similarly, `UiSyncEventArgs` and `UiSyncEvent` have been also moved to appui-react.
- `PointProps` was created primarily to avoid a dependency on @itwin/core-geometry, which contains an identical interface named `XAndY`. Similarly, `UiAdmin.createXAndY` has been deprecated.

`isLetter` has been removed since it was @alpha and is unused by core or appui.

The accompanying PR in appui-abstract is here: https://github.com/iTwin/appui/pull/492


ui/appui-abstract/src/appui-abstract/UiAdmin.ts
```diff
@@ -92,9 +92,13 @@ export class UiAdmin {
...
 
   /** Get the cursor X and Y position. */
+  // eslint-disable-next-line deprecation/deprecation
   public get cursorPosition(): PointProps { return { x: 0, y: 0 }; }
 
-  /** Create a PointProps object. */
+  /** Create a PointProps object.
+  * @deprecated in 4.2.x. Please use @core/geometry [[XAndY]] or a custom implementation.
+  */
+  // eslint-disable-next-line deprecation/deprecation
```
pmconne, line 101: If `PointProps` and `XAndY` are identical as your NextVersion.md notes claim, can't you simply replace `PointProps` with `XAndY` here and elsewhere with zero impact on API compatibility, instead of suppressing deprecation warnings?

ui/appui-abstract/src/appui-abstract/utils/PointProps.ts
```diff
@@ -7,7 +7,7 @@
  */
 
 /** Describes 2d points.
- * @public
+ * @public @deprecated in 4.2.x. Use @core/geometry [[Geometry.XAndY]].
  */
 export interface PointProps {
```
pmconne, line 12: `export type PointProps = XAndY`?

ui/appui-abstract/src/appui-abstract/UiAdmin.ts
ben-polinsky: My thinking was to avoid adding core-geometry as a appui-abstract dependency.

pmconne: Then you can't really deprecate it, can you? What is the downside of having core-geometry as a dependency? Do you believe there are apps who will use our appui-abstract library without also having a transitive dependency on core-geometry via, for example, core-frontend?

pmconne: You also have the option of simply spelling out the simple interface at each point of use instead of giving it a name.

raplemie: appui-abstract is a dependency of core-react, which was intended to be used without iTwin.js other packages, hence appui-abstract not depending on anything else than core-bentley.

That being said, with iTwinUI now providing most of the components that core-react was providing when it was the only Bentley option, we are thinking of hauling back the remaining pieces and eventually kill core-react as it no longer appear to be required as a standalone package. The work is not done there yet though.

pmconne (approved):

raplemie (approved):
LGTM!

raplemie (approved):
LGTM!

PR #6004, opened 2023-09-15
Fix ignore material not using default material param

DStradley:
Fix shader problem where the use_material flag was initialized from another variable that hadn't been initialized yet.  This caused the computeMaterialParams function to not return the default material when ignore material is set.
Fixed by just initializing straight from a uniform.


pmconne (approved):

markschlosseratbentley (approved):

DStradley:
Why is iTwin.js Docs - YAML status = Queued for an hour?

PR #6005, opened 2023-09-18
Update executeWritable to handle db busy errors

ImVeryLost:
Follow up to #5980 

Previously, if db was locked, we ended up in corrupted state were `nativeDb` was closed, and any attempt to access it would kill the backend.
As such:
1. Move `closeAndReopen` into the try/finally block, to ensure briefcase is always reopened in read-only mode.
2. Move out the `const fileName = this.pathName;` line, as not to access `nativeDb` when its closed ( `pathName` is a wrapper around`this.nativeDb.getFilePath()`, which crashes the backend if `nativeDb` is closed.

### DIFF
--- a/common/changes/@itwin/core-backend/ImVeryLost-handle-db-busy_2023-09-18-10-25.json
+++ b/common/changes/@itwin/core-backend/ImVeryLost-handle-db-busy_2023-09-18-10-25.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Update executeWritable to handle db busy errors",
+      "type": "none"
+    }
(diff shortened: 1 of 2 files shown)


pmconne (commented):
Where's the test?

pmconne:
What is causing db to be locked? Are you running local sync simultaneously with pulling changes?

ImVeryLost:
> Where's the test?

I can add one, but it would take 30-40s to execute due to "Busy, retrying" logic in openIModel. Is that fine?

> What is causing db to be locked? Are you running local sync simultaneously with pulling changes?

Nothing stopping people from attempting to do so... So we need to gracefully fail in those cases.

pmconne (approved):

PR #6008, opened 2023-09-18
Highlight Edge Cut Geometry

AJMigliore:
Highlighting the edge of geometry cut by a clip plane based on its distance from the plane in window coordinates. Incudes API functionality and 'fdt clip highlight' keyin

@markschlosseratbentley @pmconne @MarcNeely @DStradley


core/frontend/src/render/webgl/glsl/Clipping.ts
```diff
@@ -66,6 +67,43 @@ const applyClipPlanesPostlude = `
...
+      float d = calcClipPlaneDist(v_eyeSpace, plane);
+      vec4 pointOnPlane = vec4((v_eyeSpace - (d * plane.xyz)), 1.0);
+
+      //Converting original point to window coords
+      //Only doing this now to ensure the conversion is correct by later comparing with gl_FragCoord
+      //In final implementation, pointOnPlane will be converted, not v_eyeSpace
+
+      vec4 convertedEyeSpace = vec4(v_eyeSpace, 1.0);
+      convertedEyeSpace = convertedEyeSpace * u_proj;
+      convertedEyeSpace.xyz /= convertedEyeSpace.w;                               //Should now be in NDC
```
AJMigliore, line 92: When testing after this point, it's clear that convertedEyeSpace is not in NDC as its coordinate ranges are not -1 to 1. I believe that u_proj is probably at fault, but I have yet to find how.

core/frontend/src/render/webgl/glsl/Clipping.ts
```diff
@@ -66,6 +67,54 @@ const applyClipPlanesPostlude = `
...
+      //This check will show if it is correctly in NDC
+      //since NDC is in range -1 to 1, we should expect every point to be red
+      // while being darkest at the center, getting brighter as absolute value of x increases
+      if ((abs(convertedEyeSpace.x) <= 1.0) && (abs(convertedEyeSpace.y) <= 1.0) && (!clippedByCurrentPlaneSet)) {
+        g_clipColor = vec3(abs(convertedEyeSpace.x), 0.0, 0.0);
+        return true;
+      }
+
+
+      // A possible workaround if we can't get u_proj to work correctly is to do something like this
```
AJMigliore, line 108: In eye space, the eye is at the origin and we are looking down the negative z axis. As we zoom in, the absolute value of z will decrease, and vice versa. This means if we divide our distance by the z value, the result increases as we zoom in and decreases as we zoom out. This will mean a highlighted area that behaves similarly to a purely pixel based distance calculation. E.g. as we zoom out, the amount of the model highlighted will increase, while the thickness of the highlight in window space will more or less stay the same. 

While this emulates the pixel based measurements and it will significantly cut down on number of calculations, it is not exact and there seems to be an error of 1-2 pixels between being zoomed far in and far out. The other major issue is that the value to determine how much area to highlight is not a well defined number of pixels, but rather an arbitrary eye space distance. I am not convinced this is the correct way forward, but it is a possible work around if we cannot get u_proj to work.

core/frontend/src/render/webgl/glsl/Clipping.ts
```diff
@@ -66,6 +67,54 @@ const applyClipPlanesPostlude = `
...
+      // By dividing by v_eyeSpace.z, we can get a d that changes as we zoom in and out
+      // Which will emulate the behavior of a purely pixel based distance check
+
+      // float d = abs(calcClipPlaneDist(v_eyeSpace, plane) / v_eyeSpace.z);
```
AJMigliore, line 112: https://github.com/iTwin/itwinjs-core/assets/101671244/182498fb-1d13-40e8-a169-945ab273edc1

Here is a clip of what this looks like in action, it generally does emulate the behavior of a true conversion.

core/frontend/src/render/webgl/glsl/Clipping.ts
markschlosseratbentley: Based on later conversations, disregard my previous comment.

core/frontend/src/render/webgl/glsl/Clipping.ts
```diff
@@ -66,8 +69,41 @@ const applyClipPlanesPostlude = `
     } else if (!clippedByCurrentPlaneSet && calcClipPlaneDist(v_eyeSpace, plane) < 0.0) {
       clippedByCurrentPlaneSet = true;
     }
+
+    //avoiding near and far clip planes
+    if ((i <= u_clipParams[1] - 2) && (!clippedByCurrentPlaneSet)) {
```
AJMigliore, line 74: This all works as intended now, the one thing we might want to change is adding some sort of condition to avoid doing these calculations on the majority of frags. Something like:

if calcClipPlaneDist(v_eyeSpace, plane) >= someArbitraryBarrier    -> don't do calculations

common/changes/@itwin/core-frontend/andremig-highlight-clip-edge_2023-10-10-13-28.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Added ClipHighlight API functionality, and 'fdt clip highlight' keyin",
```
pmconne, line 5: Shouldn't there be a changelog for the frontend-devtools package too? That's where the key-in resides, not core-frontend.

AJMigliore:
> What remains to be done before this is ready for review?

The changelogs and getting the extract-api test to pass is all the I still am working on cleaning up, once that is sorted it'll be ready for review

pmconne:
Marking ready for review now because github won't re-trigger the builds after you do so unless you push further changes or manually summon the bot.

pmconne:
> Hilite over Highlight

Personally I've always associated "hilite" specifically with the outline effect applied to elements within a viewport. I'd go with "highlight" here to avoid conflating the two (and because spelling).

markschlosseratbentley (changes requested):
TODO:
- [ ] NextVersion.md
- [ ] Image Tests
- [ ] Perf Testing?

common/changes/@itwin/frontend-devtools/andremig-highlight-clip-edge_2023-10-10-13-28.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/frontend-devtools",
+      "comment": "",
```
pmconne, line 5:
```suggestion
      "comment": "Added 'fdt clip highlight' keyin",
```

common/changes/@itwin/core-frontend/andremig-highlight-clip-edge_2023-10-10-13-28.json
pmconne, line 5:
```suggestion
      "comment": "Added support for highlighting geometry intersecting a clip volume",
```

common/changes/@itwin/core-common/andremig-highlight-clip-edge_2023-10-10-13-28.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
+      "comment": "",
```
pmconne, line 5:
```suggestion
      "comment": "Added support for highlighting geometry intersecting a clip volume",
```

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,62 @@ export class CutStyle {
...
+export interface ClipHighlightProps {
+  /** Color to highlight the edge */
+  color?: RgbColorProps; // default white
+  /** Width of highlight in pixels */
+  width?: number; // in pixels, default 1
+}
+
+/** As part of a [[ClipStyle]], describes if/how edges of cut geometry should be highlighted.
+ * @note Edges are highlighted only if [[ClipStyle.ClipHighlight]] is `true`.
+ * @public
```
pmconne, line 115: This should have @extensions tag too.

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,62 @@ export class CutStyle {
...
+ */
+export class ClipHighlight {
+  /** Color to highlight the edge, default to white */
+  public color?: RgbColor;
```
pmconne, line 119: Make all properties `readonly`. Otherwise people can do things like `viewport.displayStyle.clipStyle.clipHighlight.width = 5` and be confused when their view doesn't update.

pmconne (changes requested):
Add unit tests. See ClipStyle.test.ts for examples, but also make sure that round-tripping a DisplayStyleSettings through JSON and back preserves the clip highlight; that APIs that change the clip highlight on a DisplayStyleSettings object trigger changed events; etc.

core/frontend-devtools/public/locales/en/FrontendDevTools.json
```diff
@@ -174,6 +174,9 @@
     "ClipColorTool": {
       "keyin": "fdt clip color"
     },
+    "ClipHighlightTool": {
+      "keyin": "fdt clip highlight"
```
pmconne, line 178: Add new keyin to README.md.

pmconne:
Ask @diegopinate whether and how this new property will be supported by the old and new saved views services.

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,62 @@ export class CutStyle {
...
 
+/** Wire format describing a [[ClipHighlight]].
+ * @see [[ClipStyleProps.ClipHighlight]].
+ * @public
+ * @extensions
+ */
+export interface ClipHighlightProps {
+  /** Color to highlight the edge */
+  color?: RgbColorProps; // default white
+  /** Width of highlight in pixels */
```
pmconne, line 109: Is this more accurately the distance a pixel can be to be considered to intersect the clip plane and therefore get highlighted instead of clipped?

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,62 @@ export class CutStyle {
...
+  width?: number; // in pixels, default 1
+}
+
+/** As part of a [[ClipStyle]], describes if/how edges of cut geometry should be highlighted.
```
pmconne, line 113:
```suggestion
/** As part of a [[ClipStyle]], describes how to colorize geometry intersecting the clip planes.
```

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,62 @@ export class CutStyle {
...
+    if (!color && !width)
+      return this.defaults;
+
+    return new ClipHighlight(color, width);
+  }
+
+  public static readonly defaults = new ClipHighlight();
+
+  public static fromJSON(props?: ClipHighlightProps): ClipHighlight {
+    const color = props?.color ? RgbColor.fromJSON(props.color) : RgbColor.fromColorDef(ColorDef.white);
```
pmconne, line 138: return `defaults` if input is undefined (it almost always will be).

core/common/src/ClipStyle.ts
```diff
@@ -115,6 +172,8 @@ export interface ClipStyleProps {
   insideColor?: RgbColorProps;
   /** If defined, geometry outside of the clip planes will be drawn in this color instead of being clipped. */
   outsideColor?: RgbColorProps;
+  /** If defined, pixels inside the clip planes within a width will be highlighted */
+  clipHighlight?: ClipHighlightProps;
```
pmconne, line 176:
```suggestion
  highlight?: ClipHighlightProps;
```

We already know it applies to the clip by virtue of being a field of `ClipStyle`.

pmconne (commented):
I hate arguments about names, but this feature is about recoloring geometry that intersects the clip volume.
"highlight" doesn't really convey that, does it?
What alternatives did you consider? Any that include "intersection"?

AJMigliore:
> What alternatives did you consider? Any that include "intersection"?

I was thinking of it more as highlighting the edge of a cut piece of geometry rather than the recoloring the intersection, but something like IntersectStyle would be a better descriptive name in that case

pmconne:
Currently, the only way to turn off this effect is to set `clipStyle.clipHighlight` to `undefined`. This makes it awkward and error-proone to implement, e.g., a toggle button in a UI that allows the user to toggle the effect on and off, because that button will have to remember what the color and width were before toggling it off and restore them when toggling back on.

Most other aspects of a display style (examples: CutStyle, visible edges, skybox, etc) work differently: there is a boolean (e.g., `ViewFlags.visibleEdges`, `ClipStyle.produceCutGeometry`, `Environment.displaySky`) that controls whether the effect is enabled, and the effect's settings are preserved separately, so when you turn the flag back on the effect looks the way it did before you turned it off.

Consider adding a flag to `ClipStyle` to control whether the intersection color is applied or not, and preserve the settings in `toJSON` IFF they don't match the defaults.

pmconne:
> something like IntersectStyle would be a better descriptive name in that case

I agree. I suggest renaming `ClipHighlight` to `ClipIntersectionStyle` and `ClipStyle.clipHighlight` to `intersectionStyle`.

core/common/src/ClipStyle.ts
pmconne: Please leave it to the original commenter to resolve their comments.

AJMigliore: sorry!

docs/changehistory/NextVersion.md
```diff
@@ -7,7 +7,18 @@ Table of contents:
...
+## Display
+
+### Colorizing Clip Intersections
+
+Geometry which intersects clip volumes can now be colorized with [ClipStyle.intersectionStyle]($common). The images below illustrate this effect, first with the intersection style turned off, second with it turned on.
+
+![No Intersection Style](./assets/IntersectionStyle-Off.jpg "No intersection style is applied.") ![Default Intersection Style](./assets/IntersectionStyle-Default.jpg "Geometry determined to intersect the clip plane is recolored white at a width of one pixel.")
+
+You can toggle this colorization on and off using [ClipStyle.colorizeIntersection]. The style of this colorization can be controled using [ClipStyle.intersectionStyle]($common)by defining a [ClipIntersectionStyle]($common). [ClipIntersectionStyle.color]($common) defines the color to apply to the intersecting geometry, and [ClipIntersectionStyle.width]($common) defines the number of pixels considered to be intersecting the clip volume, which will therefore be colorized. The image below illustrates an altered [ClipStyle.intersectionStyle]($common), with [ClipIntersectionStyle.color]($common) set to red, and [ClipIntersectionStyle.width]($common) set to 5.
```
markschlosseratbentley, line 22: API tag on [ClipStyle.colorizeIntersection]

markschlosseratbentley: Also [ClipIntersectionStyle.width]

core/frontend/src/render/webgl/glsl/Clipping.ts
```diff
@@ -124,7 +159,17 @@ export function addClipping(prog: ProgramBuilder) {
...
   });
 
-  addModelViewMatrix(vert);
+  prog.frag.addUniform("u_colorizeIntersection", VariableType.Boolean, (program) => {
+    program.addProgramUniform("u_colorizeIntersection", (uniform, params) => {
+      uniform.setUniform1i(params.target.uniforms.branch.clipStack.colorizeIntersection ? 1 : 0);
+    });
+  });
+
+  prog.addUniform("u_clipIntersection", VariableType.Vec4, (program) => {
```
MarcNeely, line 168: This should probably be prog.frag.addUniform so that the u_clipIntersection uniform is not added to the vertex shader where it isn't used.  It most likely gets compiled out when the shaders compile, but we might as well not put it in there if it doesn't need to be.

AJMigliore:
> [ ] Perf Testing?

No performance issues seemed to be present using the large shell model and monitoring fps like we discussed.

AJMigliore:
> [ ] Perf Testing?

No performance issues were present when using the large model and monitoring fps like we discussed.

AJMigliore:
> Ask @diegopinate whether and how this new property will be supported by the old and new saved views services.

I have sent Diego details of the new props and he says he will create a backlog item for adding support. It should have support in the saved views service hopefully by next week.

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,78 @@ export class CutStyle {
...
+    const color = props?.color ? RgbColor.fromJSON(props.color) : RgbColor.fromColorDef(ColorDef.white);
+    const width = props?.width ? props.width : 1;
+    return new ClipIntersectionStyle(color, width);
+  }
+
+  /** The JSON representation of this style. It is `undefined` if this style matches the defaults. */
+  public toJSON(): ClipIntersectionStyleProps | undefined {
+    const props: ClipIntersectionStyleProps = {};
+
+    if (this === ClipIntersectionStyle.defaults) {
```
pmconne, line 152:
```suggestion
    if (this.matchesDefaults) {
```

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,78 @@ export class CutStyle {
...
+}
+
+/** As part of a [[ClipStyle]], describes how to colorize geometry intersecting the clip planes.
+ * @note Edges are highlighted only if [[ClipStyle.ClipIntersectionStyle]] is `true`.
+ * @public
+ * @extensions
+ */
+export class ClipIntersectionStyle {
+  /** Color to apply to intersection of geometry and clip planes, default white */
+  public readonly color?: RgbColor;
```
pmconne, line 120: `color` and `width` should not be permitted to be `undefined`.

core/common/src/ClipStyle.ts
```diff
@@ -134,23 +211,29 @@ export class ClipStyle {
   public readonly insideColor?: RgbColor;
   /** If defined, geometry outside of the clip planes will be drawn in this color instead of being clipped. */
   public readonly outsideColor?: RgbColor;
+  /** If 'true', intersection of geometry and clip planes will be colorized */
+  public readonly colorizeIntersection?: boolean;
```
pmconne, line 215:
```suggestion
  public readonly colorizeIntersection: boolean;
```

pmconne (changes requested):
Awaiting resolution of performance degradation for non-colorized clips.

core/common/src/ClipStyle.ts
AJMigliore: If they are undefined, the constructor will just return the default values (white, one pixel). I left them able to be undefined to have a quicker/cleaner default constructor, e.g. line 136:

`  public static readonly defaults = new ClipIntersectionStyle();`

If this needs to be changed, it certainly can be.

pmconne: An instance of `ClipIntersectionStyle` has a defined color and width. If the JSON props supplied when instantiating the class do not specify either or both of those properties, it's the class's job to initialize them to defaults. Users of the class should not have to check if those properties are undefined and enforce the defaults themselves (clunky and error-prone).

This same pattern is followed through the various DisplayStyle APIs.

core/common/src/ClipStyle.ts
```diff
@@ -134,23 +211,29 @@ export class ClipStyle {
...
     this.insideColor = inside;
     this.outsideColor = outside;
+    this.colorizeIntersection = colorizeIntersection;
+    this.intersectionStyle = intersectionStyle;
   }
 
   /** Create a style from its components. */
-  public static create(produceCutGeometry: boolean, cutStyle: CutStyle, insideColor?: RgbColor, outsideColor?: RgbColor): ClipStyle {
-    if (!produceCutGeometry && cutStyle.matchesDefaults && !insideColor && !outsideColor)
+  public static create(produceCutGeometry: boolean, cutStyle: CutStyle, insideColor?: RgbColor, outsideColor?: RgbColor, colorizeIntersection?: boolean, intersectionStyle?: ClipIntersectionStyle | undefined): ClipStyle {
```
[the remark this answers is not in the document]
AJMigliore: > Too many arguments to this function now, consider an overload accepting an object.

This should now be addressed with the most recent commits.

AJMigliore:
> Awaiting resolution of performance degradation for non-colorized clips.

The performance degradation issues have been resolved with the most recent commits.

AJMigliore:
TODO: Fix discoloration when colorizing the intersections of transparent geometry

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,102 @@ export class CutStyle {
...
+  public get matchesDefaults(): boolean {
+    if (this === ClipIntersectionStyle.defaults)
+      return true;
+
+    return !this.color && !this.width;
+  }
+}
+
+/** Wire format describing a [[ClipStyle]].
+ * @see [[DisplayStyleSettingsProps.clipStyle]].
```
pmconne, line 174:
```suggestion
/** Arguments supplied to [[ClipStyle.create]].
```

core/common/src/ClipStyle.ts
```diff
@@ -128,39 +229,49 @@ export class ClipStyle {
...
     this.outsideColor = outside;
+    this.intersectionStyle = intersectionStyle;
   }
 
   /** Create a style from its components. */
-  public static create(produceCutGeometry: boolean, cutStyle: CutStyle, insideColor?: RgbColor, outsideColor?: RgbColor): ClipStyle {
-    if (!produceCutGeometry && cutStyle.matchesDefaults && !insideColor && !outsideColor)
+  // public static create(produceCutGeometry: boolean, cutStyle: CutStyle, colorizeIntersection: boolean, insideColor?: RgbColor, outsideColor?: RgbColor, intersectionStyle?: ClipIntersectionStyle): ClipStyle;
+
+  public static create(style: ClipStyleArgs): ClipStyle {
```
pmconne, line 257: You have changed the signature of a `@public` method. That's [not permitted](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/api-support-policies.md). Deprecate the original method (with none of your additional arguments) and add a new overload that takes `ClipStyleArgs`.

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,102 @@ export class CutStyle {
...
+    return !this.color && !this.width;
+  }
+}
+
+/** Wire format describing a [[ClipStyle]].
+ * @see [[DisplayStyleSettingsProps.clipStyle]].
+ * @public
+ * @extensions
+ */
+export interface ClipStyleArgs {
```
pmconne, line 177:
```suggestion
export interface ClipStyleCreateArgs {
```

core/common/src/ClipStyle.ts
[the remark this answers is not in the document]
pmconne: See comment on newest version of code and API support policies linked from there.

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,102 @@ export class CutStyle {
...
+  /** If `true`, geometry will be produced at the clip planes in a 3d view.
+   * - Solids (closed volumes) will produce facets on the clip planes.
+   * - Other surfaces will produce line strings representing the edges of the surface at the clip planes.
+   * @note Cut geometry will only be produced for element geometry - not for, e.g., terrain or reality models.
+   */
+  produceCutGeometry: boolean;
+  /** If 'true', intersection of geometry and clip planes will be colorized */
+  colorizeIntersection: boolean;
+  /** Controls aspects of how the cut geometry is displayed, if [[produceCutGeometry]] is `true`. */
+  cutStyle: CutStyle;
```
pmconne, line 187:
```suggestion
  produceCutGeometry?: boolean;
  /** If `true`, intersection of geometry and clip planes will be colorized */
  colorizeIntersection?: boolean;
  /** Controls aspects of how the cut geometry is displayed, if [[produceCutGeometry]] is `true`. */
  cutStyle?: CutStyle;
```

Callers should only have to supply arguments they care about. Use defaults for anything not supplied.

core/common/src/ClipStyle.ts
```diff
@@ -97,6 +98,101 @@ export class CutStyle {
...
+}
+
+/** As part of a [[ClipStyle]], describes how to colorize geometry intersecting the clip planes.
+ * @note Edges are highlighted only if [[ClipStyle.ClipIntersectionStyle]] is `true`.
+ * @public
+ * @extensions
+ */
+export class ClipIntersectionStyle {
+  /** Color to apply to intersection of geometry and clip planes, default white */
+  public readonly color: Readonly<RgbColor>;
```
pmconne, line 120:
```suggestion
  public readonly color: RgbColor;
```
`RgbColor` is immutable; `Readonly` is redundant.

core/common/src/ClipStyle.ts
markschlosseratbentley: TypeScript cannot overload same-named functions with differing amounts of parameters.
The old create() took a list of individual parameters.
Andre has changed that to take a single arguments object containing properties for each argument.
It seems we might need to do the following:
- deprecate create(); in comment tell user to call a new createClipStyleFromArgs method.
- make a new method createClipStyleFromArgs which takes a the argument object.

Does this seem the right approach?

pmconne: > TypeScript cannot overload same-named functions with differing amounts of parameters.

[Yes, it can](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAQwM6oKYCcoAoIIAmMsCAXIgEZxwA2GyYAlBWpjqogWMaUjJyhYQGRAG8AUImmIYwRLgCE3XvGZSZmqAAsscAO6IwGQwFEserLgBEAQXTY+iYMhi0QWDNaYBuCQF8JCRgwKGwXCFFbLABzTkktbRCYijAQAFtKbD9NVCgQYGAKPKxkv0CJUEgndIBPADFwCFxkWNQKaLiWRBKyyqaahqbcHWTUjKysABoe-MLioWTu3rAYvwB6ACpNxAABELCsMGRaRE31-uq1RDrGyBGk1YB5LE72ownsRAAfRDeZvIFYAAfgWpVWTHEGmk9CgiFGqxyMjkCigtQADhg4PIETEXm9EIoALxExDWNKZbDeKGaRLJRCk3H4toAOlxSNpsyBDPhjzxr1ZgMKHMQ-kQGFomBpnJQDhwuHAhAwwBCGEIhJJXMKvmhtNxPKZAriIoqmk8+SOiAABgASMS4-wzO1C4D+K3lILcVB0DAs2hwGK4W7DMS8saIACMALmRTJCC8oqYOq9Pr9AaDQ3uACYZtYoPo4N5fEA).

markschlosseratbentley: > > TypeScript cannot overload same-named functions with differing amounts of parameters.
> 
> [Yes, it can](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAQwM6oKYCcoAoIIAmMsCAXIgEZxwA2GyYAlBWpjqogWMaUjJyhYQGRAG8AUImmIYwRLgCE3XvGZSZmqAAsscAO6IwGQwFEserLgBEAQXTY+iYMhi0QWDNaYBuCQF8JCRgwKGwXCFFbLABzTkktbRCYijAQAFtKbD9NVCgQYGAKPKxkv0CJUEgndIBPADFwCFxkWNQKaLiWRBKyyqaahqbcHWTUjKysABoe-MLioWTu3rAYvwB6ACpNxAABELCsMGRaRE31-uq1RDrGyBGk1YB5LE72ownsRAAfRDeZvIFYAAfgWpVWTHEGmk9CgiFGqxyMjkCigtQADhg4PIETEXm9EIoALxExDWNKZbDeKGaRLJRCk3H4toAOlxSNpsyBDPhjzxr1ZgMKHMQ-kQGFomBpnJQDhwuHAhAwwBCGEIhJJXMKvmhtNxPKZAriIoqmk8+SOiAABgASMS4-wzO1C4D+K3lILcVB0DAs2hwGK4W7DMS8saIACMALmRTJCC8oqYOq9Pr9AaDQ3uACYZtYoPo4N5fEA).

Ah, ok, thanks for the sample.  @AJMigliore will give it a try.

AJMigliore: This has been addressed with the most recent changes.

core/common/src/ClipStyle.ts
AJMigliore: This has now been addressed, can it be resolved?

core/common/src/ClipStyle.ts
```diff
@@ -251,14 +251,33 @@ export class ClipStyle {
...
   }
 
+  /** @deprecated in 4.x. Use [[create(style: ClipStyleCreateArgs]] */
+  public static create(produceCutGeometry: boolean, cutStyle: CutStyle, insideColor?: RgbColor, outsideColor?: RgbColor): ClipStyle;
+
   /** Create a style from its components. */
-  // public static create(produceCutGeometry: boolean, cutStyle: CutStyle, colorizeIntersection: boolean, insideColor?: RgbColor, outsideColor?: RgbColor, intersectionStyle?: ClipIntersectionStyle): ClipStyle;
+  public static create(style: ClipStyleCreateArgs): ClipStyle;
+
+  public static create(styleOrProduceCutGeometry: ClipStyleCreateArgs | boolean, cutStyle?: CutStyle, insideColor?: RgbColor, outsideColor?: RgbColor): ClipStyle {
```
pmconne, line 260:
```suggestion
  /** @internal */
  public static create(styleOrProduceCutGeometry: ClipStyleCreateArgs | boolean, cutStyle?: CutStyle, insideColor?: RgbColor, outsideColor?: RgbColor): ClipStyle {
```

pmconne (approved):

pmconne:
> TODO:
> 
>     * [x]  NextVersion.md
> 
>     * [x]  Image Tests
> 
>     * [x]  Perf Testing?

@markschlosseratbentley looks like your requested changes have been implemented - re-review?

markschlosseratbentley (approved):

core/frontend/src/render/webgl/glsl/Clipping.ts
markschlosseratbentley: Can this be resolved?

pmconne: It's been addressed.

docs/changehistory/NextVersion.md
```diff
@@ -7,7 +7,18 @@ Table of contents:
 
 - [Electron 27 support](#electron-27-support)
 
-
 ## Electron 27 support
 
 In addition to [already supported Electron versions](../learning/SupportedPlatforms.md#electron), iTwin.js now supports [Electron 27](https://www.electronjs.org/blog/electron-27-0).
+
+## Display
```
aruniverse, line 14: mind adding this to the toc on line 9?

aruniverse (approved):
extension api changes lgtm, nit wrt nextver.md

docs/changehistory/NextVersion.md
```diff
@@ -7,7 +7,18 @@ Table of contents:
 
 - [Electron 27 support](#electron-27-support)
```
aruniverse, line 8:
```suggestion
- [Electron 27 support](#electron-27-support)
- [Display](#display)
```

PR #6025, opened 2023-09-22
rm require in core-frontend

aruniverse:
Remove `require` call in core-frontend preventing pure esm usage of the pkg
Like we do with the cjs require, in esm we import a file (package.json) at runtime which doesnt exist at compile time, so we suppress the ts compiler error.
If we import the file where it actually is, it will copy the json file into the lib folder, and the paths of our output will be different (lib/esm/src instead of just lib/esm)

ToDo later:
investigate what other libraries are doing to expose the version
possibilities:
- [ ] have a txt file that has the version string, and do some pre-processing (like we do w/ imodel-native) (personally not a huge fan of this approach)
- [ ] look into [rootDirs](https://www.typescriptlang.org/tsconfig#rootDirs)

### DIFF
--- a/core/frontend/src/IModelApp.ts
+++ b/core/frontend/src/IModelApp.ts
@@ -11,7 +11,7 @@
 import { version } from "../../package.json";
 /** @public */
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-export const ITWINJS_CORE_VERSION = version;
+export const ITWINJS_CORE_VERSION = version as string;
 const COPYRIGHT_NOTICE = 'Copyright © 2017-2023 <a href="https://www.bentley.com" target="_blank" rel="noopener noreferrer">Bentley Systems, Inc.</a>';
 
 import { UiAdmin } from "@itwin/appui-abstract";
--- a/core/frontend/src/IModelApp.ts
+++ b/core/frontend/src/IModelApp.ts
@@ -6,7 +6,7 @@
  * @module IModelApp
  */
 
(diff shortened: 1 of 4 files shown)


core/frontend/src/IModelApp.ts
```diff
@@ -6,9 +6,12 @@
  * @module IModelApp
  */
 
+// @ts-expect-error resolves from the lib/{cjs,esm} dir
```
aruniverse, line 9: does correctly resolve the pkg.json at runtime and will not copy the package.json into the lib/esm or lib/cjs dir at compile time

pmconne:
@markschlosseratbentley @iTwin/itwinjs-core-display does anyone have some time to verify DTA and DPTA still work?

markschlosseratbentley:
> @markschlosseratbentley @iTwin/itwinjs-core-display does anyone have some time to verify DTA and DPTA still work?

I will verify DTA in a bit.
I propose we kick off an Image Tests on this branch to verify DPTA.

Unless that conflicts with @MarcNeely using the Image Tests system today.

markschlosseratbentley (changes requested):
I will approve once we verify DTA and DPTA.

markschlosseratbentley:
DTA still appears to build + run + rebuild + detect code changes.  Using `npm run start:electron` still only shows .js files, but I think that was a previous problem (`npm run start` shows TypeScript).

markschlosseratbentley (approved):
Image Tests successfully ran with a version of DPTA built off this branch.

PR #6036, opened 2023-09-28
`core-backend`: Add `ECDb.getSchemaProps()`

grigasp:
imodel-native: https://github.com/iTwin/imodel-native/pull/474

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -1668,6 +1668,7 @@ export class ECDb implements IDisposable {
     dispose(): void;
     // @internal
     getCachedStatementCount(): number;
+    getSchemaProps(name: string): ECSchemaProps;
     importSchema(pathName: string): void;
     get isOpen(): boolean;
     // @internal (undocumented)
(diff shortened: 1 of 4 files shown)


pmconne (approved):
@ColinKerr @christophermlawson @rschili `ECSchemaProps`, `ECSchemaItemProps`, and `ECSchemaRefereProps` are public but entirely undocumented. Even I have questions such as: what is the difference between `$schema` and `schema`? What is the format of the `version` string? What are the possible values for `schemaItemType`?

Can we plan to document these types?

ColinKerr (approved):

ColinKerr:
> @ColinKerr @christophermlawson @rschili `ECSchemaProps`, `ECSchemaItemProps`, and `ECSchemaRefereProps` are public but entirely undocumented. Even I have questions such as: what is the difference between `$schema` and `schema`? What is the format of the `version` string? What are the possible values for `schemaItemType`?
> 
> Can we plan to document these types?

These are sadly duplicates to the interfaces defined in ecschema-metadata (which also need documentation) added to avoid adding ecschema-metadata as a dependency of core-backend.  We will document both: https://github.com/iTwin/itwinjs-core/issues/6038

PR #6044, opened 2023-09-28
Expand AuthorizationClient with frequently-used types

ben-polinsky:
Adds a couple of frequently-used and sensible types to the `AuthorizationClient` interface. An optional `isAuthorized` property used to determine if a token is present and not expired, and an optional `onAccessTokenChanged` BeEvent.

The types are optional for backwards compatibility.

### DIFF
--- a/core/common/src/AuthorizationClient.ts
+++ b/core/common/src/AuthorizationClient.ts
@@ -6,7 +6,7 @@
  * @module Authorization
  */
 
-import { AccessToken } from "@itwin/core-bentley";
+import { type AccessToken, BeEvent } from "@itwin/core-bentley";
 
 /** Provides authorization to access APIs.
  * Bentley's iTwin platform APIs [use OAuth 2.0](https://developer.bentley.com/apis/overview/authorization/) for authorization.
@@ -20,4 +20,10 @@ import { AccessToken } from "@itwin/core-bentley";
 export interface AuthorizationClient {
   /** Obtain an [[AccessToken]] for the currently authorized user, or blank string if no token is available. */
   getAccessToken(): Promise<AccessToken>;
+
+  /** [[BeEvent]] which fires when the client's [[AccessToken]] changes. */
+  readonly onAccessTokenChanged?: BeEvent<(token: AccessToken) => void>;
+
+  /** An [[AccessToken]] exists which is not expired */
+  isAuthorized?: boolean;
 }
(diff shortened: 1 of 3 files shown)


ben-polinsky:
@aruniverse thoughts on this? Requested by studio: https://github.com/iTwin/create-itwin-studio-app/pull/115#discussion_r1338811982

core/common/src/AuthorizationClient.ts
```diff
@@ -20,4 +20,10 @@ import { AccessToken } from "@itwin/core-bentley";
 export interface AuthorizationClient {
   /** Obtain an [[AccessToken]] for the currently authorized user, or blank string if no token is available. */
   getAccessToken(): Promise<AccessToken>;
+
+  /** [[BeEvent]] which fires when the client's [[AccessToken]] changes. */
+  readonly onAccessTokenChanged?: BeEvent<(token: AccessToken) => void>;
+
+  /** An [[AccessToken]] exists which is not expired */
+  isAuthorized?: boolean;
```
aruniverse, line 28: i feel like this can be derived from the accessToken itself

i like the event, and think its necessary, and shouldve always been apart of the interface

aruniverse:
> @aruniverse thoughts on this? Requested by studio: [iTwin/create-itwin-studio-app#115 (comment)](https://github.com/iTwin/create-itwin-studio-app/pull/115#discussion_r1338811982)

@calebmshafer thoughts?

core/common/src/AuthorizationClient.ts
calebmshafer: I agree, I don't think this is necessary on the base interface.

MichaelBelousov: fair enough, alternatively you could export a standard utility function for this case...

```ts
export async function isAuthorized(ac: AuthorizationClient) {
  return await ac.getAccessToken() !== "";
}
```

but I agree this is unnecessary, and I don't think studio needs it, we can just recommend inlining of that code. The event however would still be useful.

aruniverse: are we in agreement we can close this pr then?

MichaelBelousov: well no... I'm saying we still would like this PR to add the event, just not add `isAuthorized`

ben-polinsky: Coming right up

ben-polinsky: Created a new PR so as to not have 50 commits merged from master: https://github.com/iTwin/itwinjs-core/pull/6122

aruniverse:
closing in favor of https://github.com/iTwin/itwinjs-core/pull/6122

Issue #6045, opened 2023-09-29
Use Schema Editor instead of Mutables


StefanApfel-Bentley:
In the first version the schema merger was using Mutable Schema Items to set or alter the properties. Since those lack of validation, the schema editor api should be used for that.

PR #6048, opened 2023-09-29
Use Schema Editor instead of using Mutable Schema Items directly

StefanApfel-Bentley:
#6045


core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -72,6 +73,7 @@ export class SchemaContextEditor {
   public readonly unitSystems = new UnitSystems(this);
   public readonly propertyCategories = new PropertyCategories(this);
   public readonly invertedUnits = new InvertedUnits(this);
+  public readonly schemaItems = new SchemaItems(this);
```
ColinKerr, line 76: Why is schemaItems part of the public API?  It is confusing as we have editors for each individual schema item type.  The public API of the SchemaContextEditor is supposed to be user friendly and make it easy for users to do the right thing.

core/ecschema-editing/src/Merging/SchemaMerger.ts
```diff
@@ -64,52 +64,58 @@ export class SchemaMerger {
   public async merge(targetSchema: Schema, sourceSchema: Schema): Promise<Schema> {
```
ColinKerr, line 64: Doing the merging with a generic merger that has all these specializations and resolvers is actually more complex than just writing the code straight out.  It pushes us to write all kinds of wonky things like the SchemaItemFactory and the resolver to merge individual attributes of each schema item type.

I think the abstraction as a whole makes it harder to understand what is happening which would normally be an OK tradeoff if it was making things a lot simpler.  In this case I think it makes is more complex.

StefanApfel-Bentley: I added a new issue for this #6074

ColinKerr (approved):

PR #6056, opened 2023-10-02
fix webpack import of named core version

aruniverse:
image

Fixes the above webpack issue by enabling  [allowSyntheticDefaultImports](https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports) and switching from named to default import

thx @simnorm for reporting the issue


aruniverse:
there should be no need to run the image tests with this change

ben-polinsky (approved):

PR #6070, opened 2023-10-05
Added support to find curve intersection XY and XYZ for curve collections and CurveChainWithDistanceIndex.

saeeedtorabi:
Also added lots of new tests.


dassaf4 (approved):
Looks good.  I dug into some of the bugs you found and fixed them.  Please review my changes.

pmconne (approved):

PR #6082, opened 2023-10-10
Copy target schema before merging

StefanApfel-Bentley:


core/ecschema-editing/src/Merging/SchemaMerger.ts
```diff
@@ -62,10 +62,14 @@ export class SchemaMerger {
    * @returns             The merged target schema.
    */
   public async merge(targetSchema: Schema, sourceSchema: Schema): Promise<Schema> {
-    const schemaChanges = await this.getSchemaChanges(targetSchema, sourceSchema);
+
```
ColinKerr, line 65: You have to think of this in terms of a graph of schemas instead of just a single schema.  I'm not sure the correct way to slice it but at the end of the day we have to be sure that the resulting schema is part of the target graph and doesn't have any pointers to another graph.  When we are done the merging must happen at the context level just like the editing.

I've were going to copy the schema then the context for that schema must be passed in and already have all the references needed to load the schema.  If you load the schema using schemas from either source or target contexts then you've intertwined the new schema with the old.

An alternative is to embrace that the schema merger modifies the target schema and leave the copying, context and order to a higher power ... we may also have to write that code.

PR #6093, opened 2023-10-11
Support downloading iModel in iOS test

tcobbs-bentley:


tcobbs-bentley:
@aruniverse This requires environment variables to be set in order to work. One of those variables is the client secret for service auth, which should be secret, and the others are for the client id, scope, iTwin id, and iModel id. Do you know how I could add those to the pipeline job?

aruniverse:
Can you link to the pipeline that currently runs these tests?

tcobbs-bentley:
> Can you link to the pipeline that currently runs these tests?

The MacOS_Node_18 Job in this pipeline:

https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2453947&view=results

tcobbs-bentley:
> > Can you link to the pipeline that currently runs these tests?
> 
> The MacOS_Node_18 Job in this pipeline:
> 
> https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2453947&view=results

It looks like I have permissions to add variables to that pipeline (including secret ones). I'll do that, then re-run.

tcobbs-bentley:
@aruniverse I added the variables to the pipeline, making the client ID and client secret be secret variables. However, those don't show up at run-time, presumably because of the following:

> **To use a secret variable in a script**, you must explicitly map it as an environment variable.

I'm not sure how to map it as an environment variable, but I assume it's something that has to be done in the yaml.

tcobbs-bentley:
OK, I found out how to do it in the Azure online instructions, and updated the "Build & run iOS display-test-app" step in core-build.yaml to include the two new secret env variables.

tcobbs-bentley:
Variables are configured in the pipeline, and I can confirm by looking at the pipeline run that they are working, and the iOS simulator test is now downloading an iModel from the hub using service auth and then opening that iModel.

aruniverse (approved):

tcobbs-bentley:
@toddsouthenbentley Could you review this from a mobile perspective? I know your review won't allow this PR to close, but if you can verify my changes (or make suggestions), hopefully the non-mobile people that have to review this will be satisfied.

toddsouthenbentley (approved):

PR #6102, opened 2023-10-13
Fixes for iOS backend unit tests

tcobbs-bentley:
Fix assets copy for iOS backend unit tests
iOS test runner:
* Store results xmls in Documents
* Enable file sharing to extract results xmls
* Delete all but the most recent 10 results xmls


nick4598 (approved):

aruniverse (approved):
Missing rush change

PR #6103, opened 2023-10-13
Improved Newton class

saeeedtorabi:
- Improved docs and formatting
- Improved test coverage


dassaf4 (approved):
I found a way to make the 2D Newton docs consistent with currently enshrined method and object names e.g., `plane.vectorU` and `setUV`. So most of my suggestions are connected.

PR #6119, opened 2023-10-19
Improve behavior of Viewport.zoomToElements...

bbastings:
...for lines and points in perspective views by adding the option to impose minimum element-aligned bounding box extents.

Team working on crack detection noticed that zoomToElements was not working well for them in perspective views as cracks would often be clipped by the front plane and the option to specify a margin isn't supported. Cracks are represented by geometric elements with a single line segment in their geometry stream. Added an option to ensure minimum element-aligned bounding box extents so that lines and points have a proper volume.


markschlosseratbentley (approved):

markschlosseratbentley:
@aruniverse fyi

pmconne (approved):

PR #6122, opened 2023-10-19
Add AuthorizationClient onAccessTokenChanged event

ben-polinsky:
Add an optional `onAccessTokenChanged` `BeEvent` to the `AuthorizationClient` interface.
We plan to make this property required in 5.0.

### DIFF
--- a/common/api/core-common.api.md
+++ b/common/api/core-common.api.md
@@ -405,6 +405,7 @@ export namespace Atmosphere {
 // @public
 export interface AuthorizationClient {
     getAccessToken(): Promise<AccessToken>;
+    readonly onAccessTokenChanged?: BeEvent<(token: AccessToken) => void>;
 }
 
 // @public
--- a/common/changes/@itwin/core-common/add-auth-client-access-token-event_2023-10-19-17-57.json
+++ b/common/changes/@itwin/core-common/add-auth-client-access-token-event_2023-10-19-17-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
+      "comment": "Add optional onAccesTokenChanged BeEvent to AuthorizationClient",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-common"
+}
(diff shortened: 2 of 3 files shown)


aruniverse (approved):

MichaelBelousov (approved):

ImVeryLost (approved):

calebmshafer:
@ben-polinsky @aruniverse what is the implementation of this going to look like in the Desktop and Mobile auth? I'm still not clear of the requirement for this.

I'm assuming the reason you would need to get an event is if you're caching the token somewhere and need to make sure it's still valid, plus avoid the overhead of calling the getAccessToken() method all the time. Is that the reasoning here?

This interface, and iModelApp, assume that the implementations passed in have done the most optimal thing on their side and keep the token up-to-date.

There is special cache handling in both the mobile and desktop clients so their specific implementations kept the correct token sync-ed between the backend and frontend processes.

aruniverse:
> what is the implementation of this going to look like in the Desktop and Mobile auth

Desktop already implements this, [see](https://github.com/iTwin/auth-clients/blob/main/packages/electron/src/renderer/Client.ts#L85)

Mobile, im not sure. Maybe @tcobbs-bentley can chime in there

calebmshafer (changes requested):
Going to block this change for now. Let's have a design discussion around this more and look at all 3 implementations (web, mobile, desktop) before moving forward with it.

tcobbs-bentley:
> > what is the implementation of this going to look like in the Desktop and Mobile auth
> 
> Desktop already implements this, [see](https://github.com/iTwin/auth-clients/blob/main/packages/electron/src/renderer/Client.ts#L85)
> 
> Mobile, im not sure. Maybe @tcobbs-bentley can chime in there

MobileHost already has onAuthAccessTokenChanged that gets posted when the native side auth client indicates the change, so moving this to the AuthorizationClient is probably fine (although I'm not sure about the implementation details of the move).

toddsouthenbentley:
This PR would be very for my work with iTwin Studio.

In an effort to get it rolling, I created a draft PR of the changes required for Mobile: #6485

PR #6136, opened 2023-10-24
Fix Crud Perf test

DanRod1999:
Fix perf test, so performance pipeline can run off of master branch. Currently running off of 2.19. Will create a PR for the pipeline once this is merged.


full-stack-tests/backend/src/perftest/ElementCRUD.test.ts
```diff
@@ -130,6 +130,8 @@ describe("PerformanceElementsTests", () => {
         if (IModelJsFs.existsSync(pathname))
           return;
 
+        await IModelHost.startup();
```
nick4598, line 133: No shutdown?

DanRod1999: adding to after.

nick4598 (approved):

PR #6137, opened 2023-10-24
Fix legacy token flow for ArcGIS services not supporting Oauth2

mdastous-bentley:
Added URL validation to check if the Oauth2 endpoint we derived from metadata do exist, if not we fall back to legacy token flow.
Also fixed status code return by `ArcGisUtilities.validateSource`


mdastous-bentley:
@Mergifyio backport release/4.2.x

extensions/map-layers-auth/src/ArcGis/ArcGisAccessClient.ts
```diff
@@ -316,28 +328,39 @@ export class ArcGisAccessClient implements MapLayerAccessClient {
...
+          const regExMatch = mapLayerUrl.match(new RegExp(/([^&\/]+)\/rest\/services\/.*/, "i"));
           if (regExMatch !== null && regExMatch.length >= 2) {
             const subdirectory = regExMatch[1];
             const port = (urlObj.port !== "80" && urlObj.port !== "443") ? `:${urlObj.port}` : "";
             const newUrlObj = new URL(`${urlObj.protocol}//${urlObj.hostname}${port}/${subdirectory}/sharing/rest/oauth2/${endpointStr}`);
 
             // Check again the URL we just composed
-            return await this.createEndpoint(newUrlObj.toString(), endpointType);
+            const isValidUrl = await this.validateEndpointUrl(newUrlObj.toString());
+            if (isValidUrl === undefined || isValidUrl ) {
```
ben-polinsky, line 343: Should this be `undefined || true`?  Wouldn't `isValidUrl === isValidUrl` always be true?

ben-polinsky: Ah sorry I read wrong... should be ok.

ben-polinsky (approved):

pmconne (approved):

eringram (approved):

PR #6139, opened 2023-10-24
Clone attached views when cloning sheet and drawing views

### LINKED ISSUE #6124: ViewportComponent from a sheet view
**Bug:**
We are duplicating the selected view in a ViewportComponent nested in a widget.  The ViewportComponent takes as props the view state.  We are passing in a cloned view state.  The component works for spatial views, but for **sheet views**, we get the exception:

  public attachToViewport(_args: AttachToViewportArgs): void {
    if (this.isAttachedToViewport)
      throw new Error("Attempting to attach a ViewState that is already attached to a Viewport");

**Issue:**
  public override attachToViewport(args: AttachToViewportArgs): void {
    super.attachToViewport(args);  **// VewState.attachToView is called initially**
    assert(undefined === this._attachments);
    this._attachments = this._attachmentsInfo.createAttachments(this); **// ViewState.attachToView is called a second time**
  }

**Callstack:**
image

I will investigate creating a codesandbox.

pmconne:
Fixes #6124.


pmconne:
@jzbuchalski will you need this backported to 4.2.x?

ben-polinsky:
@mergifyio backport release/4.2.x

PR #6144, opened 2023-10-25
compute the range of the intersection of two local ranges

dassaf4:
Implements https://github.com/iTwin/itwinjs-backlog/issues/949


core/geometry/src/clipping/ClipUtils.ts
```diff
@@ -625,34 +627,111 @@ export class ClipUtilities {
...
+        if (point)
+          degeneratePoints.push(point);
+      }
+    }
+    return undefined;
+  }
+  /**
+   * Handle pathological cases of range-range intersection, where one of the ranges defines no area or volume (is a line segment or single point).
+   * @param range local range to intersect with the point/segment
+   * @param points isolated local point, or local segment
```
saeeedtorabi, line 660: so you expect the size of points array is 1 or 2. If it's more than 2 you just ignore extra elements? If so, I suggest this:
```suggestion
   * @param points isolated local point, or local segment's start and end
```

saeeedtorabi (approved):

PR #6153, opened 2023-10-27
Fixed the bug in testAndRecordProjection using Newton method.

saeeedtorabi:


saeeedtorabi:
Combined this into https://github.com/iTwin/itwinjs-core/pull/6175

PR #6159, opened 2023-10-31
Update crypto-js and browserify-sign

pmconne:


ben-polinsky (approved):

nick4598 (approved):

aruniverse (approved):

pmconne:
@ben-polinsky can you please convince the docs build (designated as "required") to run?

pmconne:
@ben-polinsky docs build failed. Don't see any particularly relevant looking updated packages. Still don't see these links (valid or otherwise) it's complaining about.

ben-polinsky:
> @ben-polinsky docs build failed. Don't see any particularly relevant looking updated packages. Still don't see these links (valid or otherwise) it's complaining about.

Same conclusions from my end. Doesn't look like typedoc was updated. Don't see any immediate issue with any of the external docs builds we're pulling in... Continuing on...

ben-polinsky:
Very frustrating. Have not zeroed in on why, but the core-electron generated json (produced from the `docs` script`) is different on this branch vs release/4.2.x. Why does core-electron have _anything_ to do with an issue in core-common if there isn't an explicit link? Because the logic in BeMetalsmith to determine what is internal isn't very precise (if something. is internal, we don't worry about link resolution). If a type is internal, then all of its children are, including (for some reason) types from which it extends. On "working" branches `ElectronRpcManager` is correctly identified as internal, which leads to all of its children (including those of `RpcManager`) being marked as internal. On your branch, `ElectronRpcManager` isn't included at all in the json output, thus it - and none of its children - are marked as internal. 

To put it bluntly, our link resolution/checking stinks and needs to be reworked. However, the immediate question is which package was updated through `rush update --full` that is causing the difference in json schema.

pmconne:
> However, the immediate question is which package was updated through `rush update --full` that is causing the difference in json schema.

Might help narrow things down to compare to the packages updated by #6152 on master, which did not produce this docs error.

PR #6160, opened 2023-10-31
Julio/revert item name comparison flag

julioalvahash:
This PR adds support for comparisons between two schemas with different names, removes the compareItemFullName flag and handles the comparison internally.
For classes and properties comparison between two schemas with different names, it will compare the schema names for references and item names instead of the fullNames.

Supports: 
Entity Class
Mixin Class
Properties
custom attributes class
relationship class

Not supported: 
Complex data types such as Kind of quantity, format unit, unit, constant and inverted units.


core/ecschema-editing/src/Validation/SchemaCompareVisitor.ts
```diff
@@ -20,21 +21,28 @@ import { ISchemaComparer, SchemaComparerOptions } from "./SchemaComparer";
...
     this._schemaB = schemaToCompare;
-    this._lookupFullName = options?.compareItemFullName ?? true;
   }
 
+  /**
+   * Called to look up higher level schema items.
+   * @param item
+   * @returns
+   */
   private async lookupItem<T extends SchemaItem = SchemaItem>(item: SchemaItem) {
```
ColinKerr, line 40: I know you did not name this method but I suggest changing it to something like `lookupItemInSchemaB` or something.  Having the same name as the schema method is confusing and in fact it looks like one case was not updated to call this method in this file.

julioalvahash: Changed it to `lookupItemInSchemaB` and updated the case that was not calling this function.

core/ecschema-editing/src/Validation/SchemaCompareVisitor.ts
```diff
@@ -20,21 +21,28 @@ import { ISchemaComparer, SchemaComparerOptions } from "./SchemaComparer";
...
-    return this._schemaB.lookupItem<T>(this._lookupFullName ? item.key : item.name);
+    let key: SchemaItemKey | string = item.key;
+    if (item.schema.name !== this._schemaB.name){
```
ColinKerr, line 42: I'm walking through the uses of this method and it looks like I was wrong ... it looks like it is only used to look up items directly declared in schema a.  Do you agree?  If so then we can always just use the name.

core/ecschema-editing/src/Validation/SchemaComparer.ts
```diff
@@ -838,4 +859,37 @@ export class SchemaComparer {
 
     await Promise.all(promises);
   }
+
+  /**
+   * Compares properties with different full name that could potentially be the same in the context of comparing different schemas.
+   * @returns flag to indicate whether to report diagnostic or not.
+   */
+  private reportDiagnosticFlag(fullNameA: string, fullNameB: string, itemA: SchemaItem | AnyProperty | undefined, itemB: SchemaItem | AnyProperty | undefined): boolean {
```
ColinKerr: Put yourself in the shoes of someone reading the code who does not know it and tell me what you think this line does:
`const reportFlag = this.reportDiagnosticFlag(structNameA ?? "", structNameB ?? "", propertyA, propertyB);`

Could the code be written in a way that gave the user a better idea of it's function?


Now think about someone who has to write this code, could it be more succinct easier to call?


I believe you could refactor this method to be clearer to read, easier to call and avoid parsing strings.

core/ecschema-editing/src/Validation/SchemaCompareVisitor.ts
julioalvahash: Function has been removed and now we use name get items.

core/ecschema-editing/src/Validation/SchemaComparer.ts
julioalvahash: The function has been removed and replaced with `areItemsSameByName`.

core/ecschema-editing/src/Validation/SchemaComparer.ts
```diff
@@ -838,4 +859,49 @@ export class SchemaComparer {
...
     await Promise.all(promises);
   }
+
+  /**
+   * Compares two item keys.
+   * @param itemKeyA item key A to compare to.
+   * @param itemKeyB item key B to compare to.
+   * @param topLevelSchemaNameA top level schema name in which the item A exists.
+   * @param topLevelSchemaNameB top level schema name in which the item B exists.
+   * @returns true if both names are the same and neither one has a referenced schema.
```
ColinKerr, line 869:
```suggestion
   * @returns true if both names are the same and they come from their respective top level schema.
```

core/ecschema-editing/src/Validation/SchemaComparer.ts
```diff
@@ -838,4 +859,49 @@ export class SchemaComparer {
...
+   * @returns true if a same classA is in containerB, otherwise false.
+   */
+  private containerHasClass(classNameA: string, containerA: CustomAttributeContainerProps, containerB: CustomAttributeContainerProps): boolean {
+    if (containerB && containerB.customAttributes) {
+      for (const caB of containerB.customAttributes) {
+        const classNameB = caB[0];
+        const classItemKeyA = containerA.schema.getSchemaItemKey(classNameA);
+        const classItemKeyB = containerB.schema.getSchemaItemKey(classNameB);
+        const areSameByName = this.areItemsSameByName(classItemKeyA, classItemKeyB, containerA.schema.name, containerB.schema.name);
+        if (classNameA === classNameB || areSameByName) {
```
ColinKerr, line 900: do you need to compare the full class names directly?  Shouldn't that case be covered by areItemsSameByName

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
```diff
@@ -0,0 +1,556 @@
...
+
+  return found;
+}
+
+describe("Schema comparison tests to filter out cases", () => {
+  let reporter: TestSchemaCompareReporter;
+  let contextA: SchemaContext;
+  let contextB: SchemaContext;
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
ColinKerr, line 50: you are using this variable, is this suppression really needed?

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
```diff
@@ -0,0 +1,556 @@
...
+  };
+
+  beforeEach(async () => {
+    reporter = new TestSchemaCompareReporter();
+    contextA = new SchemaContext();
+    contextB = new SchemaContext();
+  });
+
+  describe("Entity Class comparisons cases", () => {
+    it("should not report baseClass delta when comparing testEntityClass", async () => {
```
ColinKerr, line 90:
```suggestion
    it("should not report baseClass delta when base class has same name and is defined in schemas being compared", async () => {
```

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
```diff
@@ -0,0 +1,556 @@
...
+        break;
+      }
+    }
+
+  });
+
+  return found;
+}
+
+describe("Schema comparison tests to filter out cases", () => {
```
ColinKerr, line 45:
```suggestion
describe("Schema comparison tests comparing schemas with different names", () => {
```

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
```diff
@@ -0,0 +1,556 @@
...
+
+      const comparer = new SchemaComparer(reporter);
+      await comparer.compareSchemas(schemaA, schemaB);
+
+      const foundDiag = findDiagnostic(reporter.changes[0].allDiagnostics, "SC-105", "SchemaA.testBaseClass", "SchemaB.testBaseClass");
+      expect(foundDiag).to.equal(false);
+
+    });
+
+    it("should report baseClass delta for testEntityClass", async () => {
```
ColinKerr, line 127:
```suggestion
    it("should report baseClass delta when base class has different name", async () => {
```

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
```diff
@@ -0,0 +1,556 @@
...
+      }, contextB);
+
+      const comparer = new SchemaComparer(reporter);
+      await comparer.compareSchemas(schemaA, schemaB);
+
+      const foundDiag = findDiagnostic(reporter.changes[0].allDiagnostics, "SC-105", "SchemaA.testBaseClassA", "SchemaA.testBaseClassB");
+      expect(foundDiag).to.equal(true);
+    });
+
+    it("should not report baseClass delta when comparing testEntityClass", async () => {
```
ColinKerr, line 165: What is the difference between this test and the first test?

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
```diff
@@ -0,0 +1,556 @@
...
+            schemaItemType: "EntityClass",
+            description: "Test base class B",
+          },
+          testEntityClass: {
+            schemaItemType: "EntityClass",
+            description: "Test entity class",
+            baseClass: "SchemaA.testBaseClassB",
+          },
+        },
+      }, contextB);
```
ColinKerr, line 156: Maybe it would be better if this schema used SchemaB json so we are still testing the case where the schemas we are comparing have different names.  I assume this case is already tested in another part of the test suite (You can check)

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
```diff
@@ -0,0 +1,556 @@
...
+
+      const foundDiag = findDiagnostic(reporter.changes[0].allDiagnostics, "SC-105", "SchemaA.testBaseClass", "SchemaB.testBaseClass");
+      expect(foundDiag).to.equal(false);
+    });
+
+    /**
+     * This case should register testBaseClass differences as usual.
+     * Then baseClass should be reported as difference because the baseClass is in another schema.
+     */
+    it("should report baseClass delta for baseClass in referenced schema", async () => {
```
ColinKerr, line 205:
```suggestion
    it("should report baseClass delta for baseClass with the same name but defined in a referenced schema", async () => {
```

core/ecschema-editing/src/Validation/SchemaComparer.ts
julioalvahash: Looking at it closely and stepping through it, areItemsSameByName can cover a full name comparison. I think I could even take it a step further and remove fullName comparisons for other cases where areItemsSameByName is used like in baseClass comparisons.

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
julioalvahash: No it's not needed.

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
julioalvahash: Done.

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
julioalvahash: Done.

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
julioalvahash: No difference, removed this test.

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
julioalvahash: Yes, it seems unnecessary to use the schemaAJson since the name is also different, changed to schemaBjson and it is testing to make sure the difference is being reported when the name is different.

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
julioalvahash: Done.

core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
julioalvahash: Done.

PR #6161, opened 2023-10-31
rename channel to compartment

swwilson-bsi:


aruniverse:
why the rename?

swwilson-bsi:
> why the rename?

The term "channel" is used both within itwin.js and in the industry to mean a channel of communications, a conduit for IPC. We should not use this term for a data-organizing or access-control mechanism.

pmconne:
This seems like a lot of disruption for very little gain. As I mentioned before, I don't find "compartment" particularly disambiguating, especially given we already have a related subject-hierarchy-organization concept named "partitions".

We now have a BIS class and a bunch of deprecated APIs that refer to this concept as "channel", plus new parallel APIs that refer to it as "compartment". The primary motivator here seems to be the fact that we have a concept of "IPC Channels", but those are literally called `IpcChannel`, which is pretty unambiguous. Couldn't differentiation between these two types of channels in the contexts in which they appear together be achieved by qualifying, e.g., "ownership channel"?

kabentley:
I find "channel" to be a better name for the concept anyway, and fully agree that this flux will be harmful.

swwilson-bsi:
I will abandon. We will try to work around the terminology confusion in connectors some other way

PR #6164, opened 2023-11-01
Add support to walk over SQLite changeset and read it as EC instances

khanaffan:
1. `SqliteChangesetReader` - low level api to read changeset from file or local changes from imodel.
2. `ChangesetECAdaptor` - convert SQLite change to EC change thought its done per table and so change is partial.
3. `PartialECChangeUnifier` - combine partial EC change instances into one. This operation need to read whole changeset and require more memory depending on size of changeset.

This API require a context Db for schema and it should belong to same timeline as changeset. It should be as recent as possible.

imodel-native: https://github.com/iTwin/imodel-native/pull/539


common/api/core-backend.api.md
```diff
@@ -617,6 +620,18 @@ export class CategorySelector extends DefinitionElement {
     toJSON(): CategorySelectorProps;
 }
 
+// @beta (undocumented)
```
pmconne, line 623: Beta APIs need documentation.

core/backend/src/ChangesetECAdaptor.ts
```diff
@@ -0,0 +1,778 @@
...
+        this._cachedClassMaps.set(classId, classMap);
+        return classMap;
+      }
+      return undefined;
+    });
+  }
+}
+
+/** @beta */
+export interface MetaData {
```
pmconne, line 342: Need a more specific name than "MetaData".

khanaffan: I did not intent to export these classes. They are internal i will remove export from it.

core/backend/src/SqliteChangesetReader.ts
```diff
@@ -0,0 +1,279 @@
...
+  readonly db?: AnyDb; /** db from which schema will be read. It should be close to changeset.*/
+  readonly invert?: true; /** invert the changeset operations */
+  readonly disableSchemaCheck?: true; /** do not check if column of change match db schema instead ignore addition columns */
+}
+
+/**
+ * @beta
+ */
+export interface SqliteChange {
+  $table?: string;
```
pmconne, line 56: What motivates prepending `$` to these property names?

pmconne: Do you expect `[key: string]: any` to contain `op`, `table`, and/or `stage`?

core/backend/src/ChangesetECAdaptor.ts
khanaffan: Ah wait this is something i need to export I will rename it.

pmconne (changes requested):
Please declare function return types explicitly.

core/backend/src/SqliteChangesetReader.ts
khanaffan: to avoid conflict with column names.

common/api/core-backend.api.md
khanaffan: done

core/backend/src/ChangesetECAdaptor.ts
khanaffan: renamed

pmconne:
Your corresponding imodel-native PR already merged. Please figure out (perhaps with @DanRod1999's assistance) how to get this PR validated - all the CI jobs appear to have been skipped.

pmconne (approved):

DanRod1999:
https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2502906&view=results
This build should validate that this current branch builds with imodel-native main. So if it it is necessary to validate again please wait for this build to succeed.

 I recently changed how our imodel-native pipeline works, and an unintended byproduct was now we have the ability to run the imodel-native pipeline against any desired itwinjs-core branch. However, this is the first real opportunity I've had to test this since the change was made, so I will also keep an eye on it as well.

PR #6170, opened 2023-11-02
Run iOS backend unit tests in simulator

tcobbs-bentley:
Set things up so that the backend unit tests can run in the iOS Simulator, and configure the integration tests pipeline to do that.


tcobbs-bentley:
@aruniverse Once this works, we should probably disable the App Center pipeline.

tcobbs-bentley:
@aruniverse @calebmshafer @nick4598 I see that my changes to integration-test-steps.yaml to run the iOS backend unit tests in an iOS Simulator will only execute from integration-validation.yaml (which runs it in Windows, Linux, and macOS), not integration-pr-validation.yaml or integration-client-regression-pr-validation.yaml, which only run it in Linux. Is this OK? When does integration-validation.yaml get run? If not very often, should I split out this one test and have it run (in macOS) alongside the existing integration-test-steps.yaml?

nick4598:
@tcobbs-bentley If these tests are unit tests only, then maybe it would be better to modify core-build.yaml and publish-test-results.yaml?

Both of those are called by fast-ci.yaml.

tcobbs-bentley:
> @tcobbs-bentley If these tests are unit tests only, then maybe it would be better to modify core-build.yaml and publish-test-results.yaml?
> 
> Both of those are called by fast-ci.yaml.

Thanks. I have removed my original yaml changes and added the test run to core-build.yaml, and the publishing to publish-test-results.yaml. The iOS Simulator backend unit tests now run right after the iOS Simulator frontend (display-test-app based) test.

nick4598 (approved):

tcobbs-bentley:
@toddsouthenbentley You're not a required reviewer here, but it would probably be good for you to review anyway.

tcobbs-bentley:
@aruniverse @calebmshafer I just merged from master and now one of the backend unit tests [fails](https://dev.azure.com/bentleycs/iModelTechnologies/_TestManagement/Runs?runId=41963315&_a=runCharts). The sole purpose of this PR is to prevent that kind of thing from happening, although I have no idea why it is failing. The failure is on test "[ECSqlStatement should restart query](https://dev.azure.com/bentleycs/iModelTechnologies/_TestManagement/Runs?runId=41963315&_a=resultSummary&resultId=100099)".

tcobbs-bentley:
When I run the test locally after pulling the update, it passes. I'm rerunning the failed pipeline. If that solves the problem, it is worrisome that the test is inconsistent. If it doesn't solve the problem, I'll have to try to figure out what is different between the CI run and my local run.

tcobbs-bentley:
The rerun passed, so it seems that particular test might not be consistent.

aruniverse (approved):
@toddsouthenbentley plz review

toddsouthenbentley (approved):

PR #6173, opened 2023-11-03
remove tile gen test and util file

DanRod1999:
This TilesGen.test.ts file requires a specific json config file, however this file is not included in the repo like outer similar ones, so I don't really know how it works. If this config file is not included when trying to run any of our perf tests an error occurs. While working on the CRUD perf tests I was trying to find a way to avoid this.

 I was given the impression that display might manage Tile Gen performance now. @markschlosseratbentley I was wondering if you had any info about this, and if its safe to remove these tests from core. I also don't really know if we run these tests anywhere anyway. It looks like atm the pipeline that used to run this on 2.19 is broken, and as I mentioned, only ran on 2.x generation anyway. So if we shouldn't remove these, maybe we can do something to fix the [pipeline](https://dev.azure.com/bentleycs/iModelTechnologies/_build?definitionId=2715)


markschlosseratbentley:
> I was given the impression that display might manage Tile Gen performance now. @markschlosseratbentley I was wondering if you had any info about this, and if its safe to remove these tests from core. I also don't really know if we run these tests anywhere anyway. It looks like atm the pipeline that used to run this on 2.19 is broken, and as I mentioned, only ran on 2.x generation anyway. So if we shouldn't remove these, maybe we can do something to fix the [pipeline](https://dev.azure.com/bentleycs/iModelTechnologies/_build?definitionId=2715)

@DanRod1999 I am not sure about these tests. It seems they were originally committed by @khanaffan 
https://github.com/iTwin/itwinjs-core/commit/4016c9937c019a23d0975a556ec5e3a4923a9f9f

As far as display team goes, I am not aware that we ever monitored these tests. @pmconne please correct me if I am wrong.

DanRod1999:
> @DanRod1999 I am not sure about these tests. It seems they were originally committed by @khanaffan here [4016c99](https://github.com/iTwin/itwinjs-core/commit/4016c9937c019a23d0975a556ec5e3a4923a9f9f)
> 
> As far as display team goes, I am not aware that we ever monitored these tests. @pmconne please correct me if I am wrong.

Good to know, thanks. If we still think these tests are beneficial then maybe they could be fixed and used once again, but if you guys have anything that do the job of this test, and/or a better job, then I can remove them. Affan was the one that mentioned to me that display manages performance, so my understanding was you guys might have something already. I don't really know what this test entails atm since I don't know much about tile gen myself.

markschlosseratbentley:
> > @DanRod1999 I am not sure about these tests. It seems they were originally committed by @khanaffan here [4016c99](https://github.com/iTwin/itwinjs-core/commit/4016c9937c019a23d0975a556ec5e3a4923a9f9f)
> > As far as display team goes, I am not aware that we ever monitored these tests. @pmconne please correct me if I am wrong.
> 
> Good to know, thanks. If we still think these tests are beneficial then maybe they could be fixed and used once again, but if you guys have anything that do the job of this test, and/or a better job, then I can remove them. Affan was the one that mentioned to me that display manages performance, so my understanding was you guys might have something already. I don't really know what this test entails atm since I don't know much about tile gen myself.

We typically monitor performance via other means than full-stack tests like this.

pmconne (approved):

markschlosseratbentley (approved):

PR #6175, opened 2023-11-06
Improvement on close approach for arcs using Newton

saeeedtorabi:
- Fixed the bug in testAndRecordProjection using Newton method.
- Implemented arc-arc intersection using Newton method.
- Extend close approach code to handle non-circular arcs using Newton


dassaf4 (approved):
Looks good! I made some comment changes, and suggested one optimization.

PR #6176, opened 2023-11-06
Add 'pointString' on the GeometryProps interface. It's been missing for a while

a-gagnon:
Add the (missing) 'pointString' property when iterating the entries of a GeometryStream.
We've been casting as 'any' on the civil side to pick this one up, but we shouldn't need to.

Add proper type information when using PointString3d/LineString3d' toJSON function.


dassaf4 (approved):
Looks good to me!  Thanks

PR #6179, opened 2023-11-06
 deprecate appui-abstract utils: UiError, getClassName, KeyboardKey types; IconSpecUtilities; 

ben-polinsky:
Deprecations:

- `UiError` - This class extends BentleyError with a category, however, that category is never used in any internal code.
- `getClassName` - This confusingly named function gets the class name of an instance or static class. It's used internally, but doesn't have a good reason to be exposed.
- `KeyboardKey.SpecialKey`, `KeyboardKey.FunctionKey`, `isArrowKey` - convenience enums and a method which are no longer used in core. These are used in service of HTML keyboard events, which are not typed. The recommendation is therefore to use the [`ts-key-enum`](https://www.npmjs.com/package/ts-key-enum) package or string literals.
- `IconSpecUtilities` - Utilities in this class appended a prefix to attempt to distinguish svg's as a source from data uri's or itwinui classes. However, we can detect when a svg is source file is used from the .svg extension.

Several modules have been removed:
- `charCode.ts`, `filters.ts`, `map.ts`, and `string.ts` - These types were taken from a VSCode implementation for matching and filtering strings. The only non-internal types have been deprecated (and in-lined in AppUI), while the remaining internal types which are no longer used have been removed.

Companion PR in [AppUI](https://github.com/iTwin/appui/pull/516)


ui/appui-abstract/src/appui-abstract/utils/IconSpecUtilities.ts
```diff
@@ -6,8 +6,10 @@
  * @module Utilities
  */
 
+/* eslint-disable deprecation/deprecation */
+
 /** Creates an IconSpec with an SVG source and gets the SVG source from an IconSpec.
- * @public
+ * @public @deprecated in 4.3. AppUI libraries > 4.6.x support loading SVGs sources without prefixes, eliminating the need for these utilities.
```
raplemie, line 12: Will be at least 4.7.x

raplemie: It should be changed in the 2 other deprecations too :)

ui/appui-abstract/src/appui-abstract/utils/UiError.ts
```diff
@@ -9,7 +9,7 @@
 import { BentleyError, BentleyStatus, GetMetaDataFunction } from "@itwin/core-bentley";
 
 /** iTwin.js UI UiError class is a subclass of BentleyError. Errors are logged.
- * @public
+ * @public @deprecated in 4.3. Use [[Backend.BentleyError]] instead.
```
raplemie, line 12: Frontend packages dont typically expect to depend on `Backend` package, is this a typo (core?)

ui/appui-abstract/src/appui-abstract/UiItemsManager.ts
```diff
@@ -13,8 +13,14 @@ import { CommonStatusBarItem } from "./statusbar/StatusBarItem";
...
 import { AbstractZoneLocation, StagePanelLocation, StagePanelSection } from "./widget/StagePanel";
-import { loggerCategory } from "./utils/misc";
 import { UiItemsProvider } from "./UiItemsProvider";
+import { getClassName } from "./utils/misc";
+
+const loggerCategory = (obj: any): string => {
+  const className = getClassName(obj);
+  const category = `appui-abstract${(className ? `.${className}` : "")}`;
+  return category;
+};
```
raplemie, line 23: This keeps using the newly deprecated `getClassName` function, and is used only to generate the string `appui-abstract.UiItemsManager`, might as well just replace the string :)

raplemie (changes requested):
Typos, and possible cleaning, but seems good to go otherwise.

raplemie (approved):

ui/appui-abstract/src/appui-abstract/UiItemsManager.ts
ben-polinsky: Thanks - that makes sense to me.

ui/appui-abstract/src/appui-abstract/utils/IconSpecUtilities.ts
ben-polinsky: That's a forehead slapper. Thanks. Fixed.

PR #6184, opened 2023-11-08
Fix automatic save behavior for Standalone .bim files

### LINKED ISSUE #6041: StandaloneDb.upgradeStandaloneSchemas does not save changes if the .bim file is a true standalone file instead of a briefcase
**Describe the bug**
Closing a briefcase will automatically save changes, closing a standalone .bim will not, upgradeStandaloneSchemas closes the iModel without explicitly saving changes.  So the method works as expected on a briefcase being opened as a standalone file but not on a true standalone.  

**To Reproduce**

1. Call StandaloneDb.upgradeStandaloneSchemas on the example-code/snippets/src/backend/assets/test.bim file
2. See that it is not upgraded

**Expected behavior**
It should save changes

- iTwin.js Version 4.x (the tip)

**Additional context**

It is worth considering that the inconsistent behavior of closeIModel may be confusing to folks.  I could not find documentation different behavior for briefcases and standalone .bims.  Considering that StandaloneDb extends BriefcaseDb the different behavior for the same methods is particularly unexpected

akshay05n:
Closes #6041 

Closing a standalone .bim file does not trigger an automated save operation, and the `upgradeStandaloneSchemas` method, in its current state, closes the iModel without saving changes. Consequently, the method does not behave as anticipated when applied to a genuine standalone file.


core/backend/src/test/imodel/IModel.test.ts
```diff
@@ -2391,7 +2391,28 @@ describe("iModel", () => {
 
     iModel = StandaloneDb.openFile(testFileName, OpenMode.ReadWrite);
     const afterVersion = iModel.querySchemaVersion("BisCore");
-    assert.isTrue(semver.satisfies(afterVersion!, ">= 1.0.10"));
+    assert.isTrue(semver.satisfies(afterVersion!, ">= 1.0.16"));
+    iModel.close();
+  });
+
+  it("upgrade the domain schema in a true StandaloneDb", async () => {
```
pmconne, line 2398: What is a "true StandaloneDb" and how does it differ from the (presumably false?) "StandaloneDb" in the test above this one?

kabentley: No clue. A StandaloneDb is a file with no iTwinId. The static `StandaloneDb.openFIle` method won't open a file if it has an iTwinId.

kabentley: The issue this is attempting to address is that the `closeIModel` method tests whether there are any local _tracked_ changes before saving. For standalone files we don't track changes (since they can't generate changesets) so SaveChanges isn't called. I don't understand why these two tests give different results - there's no such thing as a "StandaloneDb that isn't a StandaloneDb". 

But why write this test instead of just fixing the real problem?

kabentley: i now recall that years ago we created a "standalone edit flag" to allow undo when performing local editing of a standalone file. The test file "testImodel.bim" must have that flag saved on.

The real fix is to simply remove [this line](https://github.com/iTwin/imodel-native/blob/a6e7306944d58d79c74628120f8e25076976337f/iModelJsNodeAddon/IModelJsNative.cpp#L1059) and _always_ call SaveChanges on close since that's our policy. I don't know who added that test or why (I presume it was me, but I can't go back that far in Git history). It is innocuous to call SaveChanges on a readonly file or with no changes. But skipping it means that we won't save changes if you don't have change tracking enabled (as you've found) or if the changes are to non-transactable tables. Both are bad.

kabentley: > You can open a briefcase as a standalone or you can open a locally created .bim as standalone

No.

By definition, a "briefcase" is a file with an iTwinId and a briefcaseId (it must also have an iModelId to obtain a briefcaseId). You **cannot** open one with the StandaloneDb class. 

A "standalone" file is one with no iTwinId (that is, it "stands alone, unassociated with a timeline" [see explanation here](https://github.com/iTwin/itwinjs-core/blob/796333cfac9b25b1cfb1559ccecc4555375b89a5/core/backend/src/IModelDb.ts#L2953-L2966). Standalone files should always have an unassigned briefcaseId.

ColinKerr: You're right Keith, the difference between the files is that one has the StandaloneEdit[txns] flag set to true.  Sorry about that mistake

I talked to @khanaffan previously about saving all the time on close and he was concerned that this could result in unintentional changes being saved with no way to rollback.  One idea was to make the close call fail if there were unsaved changes, so they had to be explicitly cleared before close.  I'm ok with saving the changes automatically on close so long as it isn't opening us up to corrupting the file with partial changes when we close after an error.

kabentley: I will talk to @khanaffan, but we can't have our policy be "we usually save changes when you close". We should either always save changes or never save them. If someone wants to abandon their local unsaved changes, they should do so before calling close. As it stands, they'd have to do that "most of the time" which is nonsense.

ColinKerr: I agree the current inconsistent behavior is bad

kabentley: I tracked back to see why we call SaveChanges on Close. I see that Raman added that way back in 2017. I don't know why. I wish we could change the policy and **not** call it, but i think it's too late for that. Certainly the test for "HasChanges" was always wrong.

akshay05n: As per my discussion with @khanaffan, I will modify the `CloseDgnDb()` function to remove [this code](https://github.com/iTwin/imodel-native/blob/main/iModelJsNodeAddon/IModelJsNative.cpp#L1059-L1060) that handles saving changes. Additionally, I will ensure that changes are saved before closing the database on the TypeScript side and create new pull requests for these changes.
These changes will ensure that `closeIModel` always closes the imodel without saving any local changes. Users will need to explicitly call `saveChanges` to preserve local changes. Please provide any feedback or suggestions you may have regarding this approach.

kabentley: The change to the native code is correct. @khanaffan and I had a lengthy discussion on what should happen on the TypeScript side. Please make sure you coordinate with him.

akshay05n: Thanks for the details! I'll cooperate with @khanaffan for this. Also, I'll close the current pull request and initiate new ones to incorporate the discussed changes.

PR #6187, opened 2023-11-09
Run Android integration test in Emulator during CI

tcobbs-bentley:


tcobbs-bentley:
I suspect that the Azure Artifacts-based upack install will fail, but since I'm not sure it will fail, I submitted the PR now in case it doesn't fail.

tcobbs-bentley:
@toddsouthenbentley This runs the same display-test-app test on Android as is already being run on iOS. In order to do that, it has to install three upacks:
* Android SDK
* Open JDK
* Android Emulator

The python script builds display-test-app in order to have control over the `JAVA_HOME` and `ANDROID_HOME` environment variables during the build (pointing to the upacks).

toddsouthenbentley (approved):

pmconne (approved):

common/config/azure-pipelines/jobs/fast-ci.yaml
```diff
@@ -6,6 +6,7 @@ trigger:
 
 variables:
   - group: Rush Build Cache SAS Token
+  - group: Travis Cobbs - GitHub PAT
```
aruniverse, line 9: whats this for?

common/config/azure-pipelines/templates/core-build.yaml
```diff
@@ -84,18 +84,28 @@ steps:
...
       VITE_CI: true
 
+  - script: npm run android:all
+    workingDirectory: test-apps/display-test-app
+    displayName: Build & run Android display-test-app
+    env:
+      IMJS_OIDC_CLIENT_ID: $(IMJS_OIDC_CLIENT_ID)
+      IMJS_OIDC_CLIENT_SECRET: $(IMJS_OIDC_CLIENT_SECRET)
+      AZURE_DEVOPS_EXT_PAT: $(System.AccessToken)
+      TOKEN: $(GitHubPAT)
```
aruniverse, line 94: what is this used for?

common/config/azure-pipelines/jobs/fast-ci.yaml
tcobbs-bentley: In order to download the android native library from GitHub, a personal access token is required. This is completely stupid due to the thing being downloaded being totally public, but that's how GitHub works, and all indications are that they have no intention of changing.

common/config/azure-pipelines/templates/core-build.yaml
tcobbs-bentley: The first two are the client ID and client secret needed for service authorization to download the test iModel. The Azure PAT is needed to download the Android upacks. The GitHubPAT is needed to download the mobile-native-android library.

common/config/azure-pipelines/jobs/fast-ci.yaml
aruniverse: instead of creating a new var group for your pat, you couldnt just reuse an exisiting one?
is there elevated priveleges it needs or something like that?

tcobbs-bentley: The only one I can see that _might_ work is `PerfTesting-Github-AccessToken`. I don't know that it will work, and even if it does, it seems inappropriate to use it based on its name. Using it does require permission from its creator (you). Variable groups have to have unique names, so I included my own name as part of it to make it unique (similar to `Caleb - iTwin GitHub Token`, which based on its description is **not** a token that will work here). If `PerfTesting-Github-AccessToken` is a GitHub PAT with read:packages permission and you really want me to switch to it, I can do so. However, IMO if the desire is for one token to be shared by multiple pipelines, then the token name should reflect what it contains, not the first pipeline to use it.

tcobbs-bentley: Oh, and the description of `PerfTesting-Github-AccessToken` states that it is already expired. I'm not sure if the description is out of date or if the token is really expired.

aruniverse: `common/config/azure-pipelines/jobs/version-bump.yaml` uses `GITHUBTOKEN` which i believe is generated by our admin service account, `imodeljs-admin@users.noreply.github.com`. i would try to use that if possible

aruniverse: nvm, that one looks hardcoded to the pipeline

aruniverse (approved):

PR #6191, opened 2023-11-10
Allow custom parameters to be passed on MapLayerSource and MapLayerSettings objects.

mdastous-bentley:
Added the following properties to `MapLayerSource` and `ImageMapLayerSettings` :
`savedQueryParams`
`unsavedQueryParams`
`queryParams` -> readonly
This allow injecting custom query parameters in any requests made by the provider to the underlying service.  The **saved** variant will get saved part of the view object serialization, and the **unsaved** one will not.

*Added several tests to ensure requests URL are being built as before, and also can now include custom query parameters.


core/frontend/src/tile/map/MapLayerFormatRegistry.ts
```diff
@@ -30,7 +30,7 @@ export class MapLayerFormat {
 
   /** Allow a source of a specific to be validated before being attached as a map-layer.
   */
-  public static async validateSource(_url: string, _userName?: string, _password?: string, _ignoreCache?: boolean): Promise<MapLayerSourceValidation> { return { status: MapLayerSourceStatus.Valid }; }
+  public static async validateSource(_url: string, _userName?: string, _password?: string, _ignoreCache?: boolean, _accesKey?: MapLayerKey): Promise<MapLayerSourceValidation> { return { status: MapLayerSourceStatus.Valid }; }
```
mdastous-bentley, line 33: @pmconne Not sure why we initially decided to pass deconstructed MapLayerSource properties here, but adding one more parameter feels wrong.
Any suggestion to adjust the API without making it too confusing?

core/common/src/MapLayerSettings.ts
```diff
@@ -198,6 +199,16 @@ export interface MapLayerKey {
...
 }
 
+/**
+   * stores key-value pair to be added to all requests made involving map layer.
+   * @beta
+   */
+export interface MapLayerUrlParam {
+  key: string;
+  value: string;
+  secret?: boolean;
```
mdastous-bentley, line 209: @pmconne @calebmshafer  MapLayerUrlParam represent one `key=value` pair that should be apprend to map layer base URL.  This can contain non-secret information such as: `apiVersion=2`.  But it might contains some nasty information such as `password=12345`.

What do you think of having a `secret `property, when turned on, the param would not get serialized part of the JSON representation.  That we leave the decision on how to handle the secret to the API caller.
Do you foresee potential issues?  Should we have a more brute approach and strip all custom parameters from serialized json?

pmconne: Is there any reason for us to know/care about individual query parameters?
Doesn't the problem boil down to: we need a URL which may contain query parameters, and some of those query parameters should not be persisted?
If so, can't you just accept a URL to persist, plus optional additional query params to append but not persist?
If not, can you clarify (in your PR description or an issue linked therefrom) what problem you are trying to solve?

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
mdastous-bentley: The "url' parameter we see everywhere is in fact the **service endpoint** URL which means any query parameters get discarded no matter what, so if you pass:

https://server/service?customParam=value => https://server/service?<Service_official_parameters_appended_here>

I think this behavior is ok (although it is somewhat unexpected given the actual documentation).
That's why I suggest adding a `customParameters` property where the user would be free to pass any custom parameters (which of course should not conflict with the service standard parameters).  The idea of having a `isSecret` prop, is for case where user want to use custom parameter like username=password=123.

core/common/src/MapLayerSettings.ts
mdastous-bentley: The "url' parameter we see everywhere is in fact the **service endpoint** URL which means any query parameters get discarded no matter what, so if you pass:
@pmconne 
https://server/service?customParam=value => https://server/service?<Service_official_parameters_appended_here>

I think this behavior is ok (although it is somewhat unexpected given the actual documentation).
That's why I suggest adding a `customParameters` property where the user would be free to pass any custom parameters (which of course should not conflict with the service standard parameters).  The idea of having a `isSecret` prop, is for case where user want to use custom parameter like username=password=123.

pmconne: Yeah that's confusing that we automatically strip out all query parameters. We can't change the behavior now but we should fix the documentation.

`isSecret` isn't an accurate name for a property that really means "don't store this parameter". And because we currently don't save any query parameters, that should be default for the new parameters too.

Did you consider the following instead of a single `customParameters` array?
```
{
  savedQueryParams?: { [key: string]: string },
  unsavedQueryParams?: { [key: string]: string },
}
```

mdastous-bentley: That could work yes.. what is the benefit of using an Index signature here?  Do you want to express the uniqueness of param keys ?  If so, that could still happen be because we have two independents list.

pmconne: > benefit of using an Index signature

Because these are key-value pairs of strings, rather than inventing a new type `MapLayerUrlParam` for such a simple concept.

pmconne: Please publish it so we can see CI results.

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
pmconne: It's `@public` so your options are limited. What changes did you want to make?

core/common/src/MapLayerSettings.ts
```diff
@@ -280,6 +287,16 @@ export class ImageMapLayerSettings extends MapLayerSettings {
   public userName?: string;
   public password?: string;
   public accessKey?: MapLayerKey;
+
+  /** List of query parameters that will get appended to the settings URL that should be be persisted part of the JSON representation.
```
pmconne, line 291:
```suggestion
  /** List of query parameters to appended to the settings URL and persisted as part of the JSON representation.
   * @note Sensitive information like user credentials should be provided in [[unsavedQueryParams]] to ensure it is never persisted.
```

core/common/src/MapLayerSettings.ts
```diff
@@ -280,6 +287,16 @@ export class ImageMapLayerSettings extends MapLayerSettings {
   public userName?: string;
```
pmconne, line 287: `userName`, `password`, and `accessKey` (?) probably all really belong in `unsavedQueryParams`, no?
Several APIs now take `userName`, `password`, and `queryParams` which is awkward.
Would it make sense to convert these public data members into getter/setters that read to and write from `unsavedQueryParams`?

mdastous-bentley: `userName`, `password ` are used for basic authentication, which result in a header being added to the request, not query paramters.

`accessKey` is overlapping with unsaved query parameters.  Strangely it's being overridden internally to pass the MapLayerRegistry's key inside [createImageryProvider](https://github.com/iTwin/itwinjs-core/blob/f12f5738b0739b9ccc7053d402cf5776c7ff6b08/core/frontend/src/tile/map/MapLayerFormatRegistry.ts#L166C30-L166C31).  Such a mess...

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
mdastous-bentley: @pmconne  I created `validateSourceObj`, does that make sense to you ?

pmconne: Your default implementation of `validateSourceObj` does not forward to potentially-overridden implementation of `validateSource`. This will break any existing `MapLayerFormat` not controlled by us.

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
```diff
@@ -28,42 +28,35 @@ export class MapLayerFormat {
...
-  /**
-   * Allow a source of a specific format to be validated before being attached as a map-layer.
-   * @param _url The URL of the source.
-   * @param _userName The username to access the source if needed.
-   * @param _password The password to access the source if needed.
-   * @param _ignoreCache Flag to skip cache lookup (i.e. force a new server request).
-   * @returns Validation Status. If successful, a list of available sub-layers may also be returned.
-   */
-  public static async validateSource(_url: string, _userName?: string, _password?: string, _ignoreCache?: boolean): Promise<MapLayerSourceValidation> { return { status: MapLayerSourceStatus.Valid }; }
+  /** @deprecated in 4.2. Use the overload that takes a [[MapL]]. */
```
pmconne, line 31:
```suggestion
  /** @deprecated in 4.3. Use [[validateSourceObj]]. */
```

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
```diff
@@ -28,42 +28,35 @@ export class MapLayerFormat {
...
-   * @param _settings The map layer settings to be applied to the imagery provider.
-   * @returns Returns the new imagery provider.
-   * @beta
-   */
-  public static createImageryProvider(_settings: MapLayerSettings): MapLayerImageryProvider | undefined {
-    assert(false, "Subclasses must override this method.");
-    return undefined;
-  }
+  /** Allow a source object to be validated before being attached as a map-layer.
+    * @beta
```
pmconne, line 35: You have prescribed this as the replacement for the deprecated, `@public` method `validateSource`, but this replacement is `@beta`. (Why is everything you've added marked beta?)

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
mdastous-bentley: > Did you encounter a problem adding an overload of `validateSource`, motivating the `Obj` suffix?

I'm getting: `Duplicate function implementation.ts(2393)`
If there's a way to create an override, I will happily remove the `Obj` suffix.

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
mdastous-bentley: I wanted to get it adopted in the UI before freezing the API.

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
mdastous-bentley: > > Did you encounter a problem adding an overload of `validateSource`, motivating the `Obj` suffix?
> 
> I'm getting: `Duplicate function implementation.ts(2393)` If there's a way to create an override, I will happily remove the `Obj` suffix.

ok I think I found how to achieve it.

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
pmconne: Ok, but we can't deprecate a public API if the only replacement is a beta (unstable) API. That will have to be done when you promote the replacement to public.

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
pmconne: Something like that, yes. Mark the one that has an implementation and ugly signature is `@internal`. Make the new one take a single argument, not two (e.g., add `source: MapLayerSource` to `ValidateSourceOptions`). Nitpickery, but our convention is to use the suffix `Args` for function argument interface types, not `Options`.

mdastous-bentley: Problem with overrides is that it invalidate sub-classe implementations, forcing them to implements all overrides of the base class.

I can certainly apply the overrides strategy to `MapLayerFormatRegistry.validateSource` but dont see how I can achieve it for `MapLayerFormat.validateSource` without breaking the API.

pmconne: > Problem with overrides is that it invalidate sub-class

Ugh. Fair.
I'd still prefer the new method to take a single `Args` argument - that's how we avoid similar problems in future when we want to evolve an API without breaking/deprecating it. If you can tolerate more extreme nitpickery, I'd ditch the `Obj` from `validateObj` - that non-word suffix isn't adding any useful information.

core/frontend/src/tile/map/MapLayerSources.ts
```diff
@@ -112,12 +128,33 @@ export class MapLayerSource {
...
   }
 
   private getCredentials(): RequestBasicCredentials | undefined {
     return this.userName && this.password ? { user: this.userName, password: this.password } : undefined;
   }
+
+  /** Get all query parameters
+ * @beta
+ */
+  public get queryParameters() {
```
pmconne, line 148: This should probably be a method like `collectQueryParams()`. Getter makes it look like a simple property access which has performance implications but more importantly implies that repeated calls will return the same object. Some bundlers make that latter assumption to optimize code, which will produce incorrect results in this case.

core/frontend/src/tile/map/MapLayerImageryFormats.ts
```diff
@@ -45,11 +47,26 @@ class WmsMapLayerFormat extends ImageryMapLayerFormat {
...
+      return { status: MapLayerSourceStatus.InvalidFormat };
+    source.userName = userName;
+    source.password = password;
+    return WmsMapLayerFormat.validateSourceObj(source, {ignoreCache});
+  }
+
+  public static override async validateSourceObj(source: MapLayerSource, opts?: ValidateSourceOptions): Promise<MapLayerSourceValidation> {
+    const { url, userName, password, savedQueryParams, unsavedQueryParams } = source;
+    const queryParams: {[key: string]: string} = {};
+    savedQueryParams && Object.keys(savedQueryParams).forEach((key) => queryParams[key] = savedQueryParams[key]);
```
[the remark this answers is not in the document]
mdastous-bentley: thats a left over, fixed now.

core/frontend/src/tile/map/MapLayerSources.ts
mdastous-bentley: I agree, done.

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
```diff
@@ -28,42 +28,35 @@ export class MapLayerFormat {
...
-  /**
-   * Creates a MapLayerTileTreeReference for this map layer format.
-   * @param _layerSettings Map layer settings that are applied to the MapLayerTileTreeReference.
-   * @param _layerIndex The index of the associated map layer.
-   * @param _iModel The iModel containing the MapLayerTileTreeReference.
-   * @returns Returns the new tile tree reference.
-   * @beta
-   */
+  /** Create a [[MapLayerImageryProvider]] that will be used to feed data in a map-layer tile Tree.
+   * @internal
```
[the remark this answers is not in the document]
mdastous-bentley: I think this is a merge error, reverted.

pmconne (approved):

aruniverse (approved):

extensions/map-layers-formats/src/ArcGisFeature/ArcGisFeatureProvider.ts
```diff
@@ -300,7 +300,7 @@ export class ArcGisFeatureProvider extends ArcGISImageryProvider {
     try {
       const url = new URL(this._settings.url);
       url.pathname = `${url.pathname}/${layerId}`;
-      metadata = await ArcGisUtilities.getServiceJson(url.toString(), this._settings.formatId, this._settings.userName, this._settings.password, this._accessTokenRequired);
+      metadata = await ArcGisUtilities.getServiceJson(url.toString(), this._settings.formatId, this._settings.userName, this._settings.password, this._settings.collectQueryParams(), this._accessTokenRequired);
```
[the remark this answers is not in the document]
aruniverse: plz resolve this comment once youve read it

eringram (approved):

extensions/map-layers-formats/src/ArcGisFeature/ArcGisFeatureProvider.ts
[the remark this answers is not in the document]
pmconne: Nor am I, but I ignored it because it's `@internal`.

PR #6206, opened 2023-11-16
Modified getAspects to handle dot separator in full classname

RohitPtnkr1996:
In the change made in the PR #5966, the function getAspects was modified to use instance queries.
As part of the change in which we are checking if the input class is abstract, I had assumed the input aspect class full name to only ever be colon-separated (as per the unit tests present at the time).
This constitutes a breaking change as we used to accept class full names that were dot-separated as well as colon-separated before the PR #5966 
This PR fixes the change to handle both inputs and makes the change made in #5966 non-breaking.

Fixes iTwin/itwinjs-backlog/issues/981


pmconne:
@khanaffan @ColinKerr is this a desirable change?
@RohitPtnkr1996 please provide context in your PR description.

RohitPtnkr1996:
PR description updated.

pmconne (approved):

PR #6211, opened 2023-11-17
Fix Map Feature Info for ArcGIS map services when reprojection is needed

mdastous-bentley:
Very small fix to correctly set the spatial reference id in the 'Identify' query parameters.  
Also use the default 'top' mode instead of all `visible` layers.


aruniverse (approved):

markschlosseratbentley (approved):

mdastous-bentley:
@aruniverse Do we still plan to make 4.2.x builds ?If so, I think this fix should be included.

aruniverse:
Yes we are still doing 4.2 patches

@mergifyio backport release/4.2.x

pmconne:
@Mergifyio backport release/4.2.x

PR #6218, opened 2023-11-21
Fix schedule simulation for unquantized tiles

pmconne:
Fixes https://github.com/iTwin/imodel-native-internal/pull/251

The bytes representing the positions and feature indices are transposed in unquantized tiles for more efficient access on the GPU. When splitting vertex tables, we need to invert the transpose when extracting the feature index.


pmconne:
Awaiting ImageTests results.

pmconne:
Closing in favor of https://github.com/iTwin/itwinjs-core/pull/6220

PR #6219, opened 2023-11-21
Presentation: Support default formats when formatting in ts

grigasp:
Note: This PR doesn't have a native counterpart. The reason it targets `imodel02` is that it builds on top of what's not in `master` yet.


saskliutas (approved):

pmconne:
> This PR doesn't have a native counterpart. The reason it targets imodel02 is that it builds on top of what's not in master yet.

Still would like the CI jobs to validate your changes.

grigasp:
> Still would like the CI jobs to validate your changes.

Apparently not as simple as it sounds 🤔

pmconne (approved):
Please publish a new addon as soon as this PR completes.

PR #6220, opened 2023-11-21
Support unquantized batched tiles

pmconne:
[Corresponding imodel-native-internal PR](https://github.com/iTwin/imodel-native-internal/pull/251)

### Fix schedule animations

The bytes representing the positions and feature indices are transposed in unquantized tiles for more efficient access on the GPU. When splitting vertex tables, we need to invert the transpose when extracting the feature index.

### Support nested per-tile transforms

https://github.com/iTwin/imodel-native-internal/pull/274 will enable single-precision non-leaf tiles with local transforms where previously we asserted only leaf tiles could have local transforms.


aruniverse:
is this pr supposed to target the `imodel02` branch?

pmconne:
> is this pr supposed to target the imodel02 branch?

No. The corresponding native changes are not part of the addon.

aruniverse (approved):

markschlosseratbentley (approved):

PR #6231, opened 2023-11-23
Fix extract api summary script to work with additional '.' symbols in path

JonasDov:
There was a bug in extract-api-summary.js it did not create valid fileName for <fileName>.export.csv file. This was happening when there were additional '.' symbols in the 'apiSignature' path.


tools/build/scripts/extract-api-summary.js
```diff
@@ -33,8 +33,10 @@ if (undefined !== argv.gatherFullReport)
   shouldGenerateFullReport = true;
 
 // create output file
-const trimmedApiSignature = (argv.apiSignature.split('.'))[0];
-const sigFileName = path.basename(path.normalize(trimmedApiSignature));
+const apiSignatureNormalized = path.normalize(argv.apiSignature);
+const apiSignatureFileName = apiSignatureNormalized.split('\\').slice(-1)[0];
```
grigasp, line 37: `split('\\')` will probably not work on systems other than Windows as they use `/`. Maybe use [`path.parse`](https://nodejs.org/api/path.html#pathparsepath) for this?

tools/build/scripts/extract-api-summary.js
```diff
@@ -33,8 +33,10 @@ if (undefined !== argv.gatherFullReport)
...
+const trimmedApiSignatureFileName = (apiSignatureFileName.split('.'))[0];
+const sigFileName = path.basename(path.normalize(trimmedApiSignatureFileName));
```
grigasp, line 39: I don't think the file name needs to be normalized, especially since it's taken from an already normalized path.

grigasp (changes requested):

tools/build/scripts/extract-api-summary.js
JonasDov: Changed to: 
```ts
const apiSignatureFileName = path.parse(argv.apiSignature).name;
```

tools/build/scripts/extract-api-summary.js
JonasDov: Removed

grigasp (approved):
LGTM

aruniverse (approved):

PR #6232, opened 2023-11-24
Save changes before closing ReadWrite IModels

akshay05n:
With this pull request, changes made to the iModel will now be automatically saved before closing.

imodel-native: https://github.com/iTwin/imodel-native/pull/558

### DIFF
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -364,7 +364,7 @@ export abstract class IModelDb extends IModel {
     }
   }
 
-  /** Close this IModel, if it is currently open. */
+  /** Close this IModel, if it is currently open, and save changes if it was opened in ReadWrite mode. */
   public close(description?: string): void {
     if (!this.isOpen)
       return; // don't continue if already closed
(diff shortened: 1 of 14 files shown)


core/backend/src/IModelDb.ts
```diff
@@ -365,7 +365,7 @@ export abstract class IModelDb extends IModel {
   }
 
   /** Close this IModel, if it is currently open. */
```
kabentley, line 367: please add a comment that this function automatically saves changes.

akshay05n: Updated comment to reflect the current behaviour.

core/backend/src/IModelDb.ts
```diff
@@ -364,8 +364,8 @@ export abstract class IModelDb extends IModel {
     }
   }
 
-  /** Close this IModel, if it is currently open. */
-  public close(): void {
+  /** Close this IModel, if it is currently open, and save changes if it was opened in ReadWrite mode. */
+  public close(description?: string): void {
```
pmconne, line 368: It's unclear to me who would ever call `close` with a description - the only reason to do that would be if you know the file contains unsaved changes and what those changes are, in which case, why aren't you calling `saveChanges`?

pmconne (approved):

PR #6233, opened 2023-11-24
`@itwin/map-layers-formats`: Add missing `core-frontend` peer dep

saskliutas:
`@itwin/map-layers-formats` package exposes types from `core-frontend` but does not have it as peer dependency.

https://github.com/iTwin/itwinjs-core/blob/ce0577203cc5ac67767871025e756566f937c933/common/api/map-layers-formats.api.md?plain=1#L62

https://github.com/iTwin/itwinjs-core/blob/ce0577203cc5ac67767871025e756566f937c933/common/api/map-layers-formats.api.md?plain=1#L94

This might lead to build failures in packages consuming `@itwin/map-layers-formats` if different `core-frontend` is resolved by package manager.

### DIFF
--- a/common/changes/@itwin/map-layers-formats/missing_peer_dep_2023-11-24-07-53.json
+++ b/common/changes/@itwin/map-layers-formats/missing_peer_dep_2023-11-24-07-53.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/map-layers-formats",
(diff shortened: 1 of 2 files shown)


aruniverse (approved):
sigh...
this pkg heavily relies on core-frontend, and shouldve always had a peer on it (it uses IModelApp)

this is technically breaking, however, i cant imagine anyone being able to use this pkg without already treating this as a peer

PR #6234, opened 2023-11-24
change `pnpm` `resolution-mode` to `lowest-direct`

saskliutas:
Changed `pnpm` resultion mode to `lowest-direct`. This makes `pnpm` to resolve dependencies to the lowest matching version instead of the highest. This should help to avoid mistakes where new APIs are used but minimal dependency version is not bumped. 

E.g: `@itwin/express-server` has dependency on `express`@^4.16.3.
https://github.com/iTwin/itwinjs-core/blob/ce0577203cc5ac67767871025e756566f937c933/core/express-server/package.json#L63
But uses `text` function that is only available in `4.17.0` https://github.com/expressjs/express/blob/2a00da2067b7017f769c9100205a2a5f267a884b/History.md?plain=1#L130
https://github.com/iTwin/itwinjs-core/blob/ce0577203cc5ac67767871025e756566f937c933/core/express-server/src/ExpressServer.ts#L45

Some other problems encountered when building `itwinjs-core` with minimal dependency versions:
* `@itwin/core-webpack-tools` fails to build because `@types/chai-jest-snapshot`@1.3.0 does not have `setFilename` and `setTestName` functions
https://github.com/iTwin/itwinjs-core/blob/ce0577203cc5ac67767871025e756566f937c933/tools/webpack-core/src/test/setup.test.ts#L18

* `@itwin/certa` does not build with `@types/lodash`@4.14.0. This might be because types are too outdated for building with TypeScript 5.0

* `@itwin/core-backend` fails to build because `form-data`@2.3.2 does not deliver types. Types are delivered from `2.5.0` version.

Some additional dependency updates:
* Bumped `mocha` and `@types/mocha` to the latest version.
* Bumped `sinon` to the latest version because `sinon`@15.0.4 is deprecated.
* Changed `lolex` to `@sinonjs/fake-timers` as it is deprecated.


grigasp (approved):

grigasp:
Looks like the change revealed another issue - the audit fails due to a number dependencies on vulnerable versions.

paulius-valiunas (approved):

dassaf4 (approved):

nick4598 (approved):

naveedkhan8067 (approved):

pmconne (approved):

core/geometry/package.json
```diff
@@ -40,11 +40,12 @@
     "@itwin/eslint-plugin": "4.0.0-dev.44",
     "@types/chai": "4.3.1",
     "@types/flatbuffers": "~1.10.0",
-    "@types/mocha": "^8.2.2",
+    "@types/mocha": "^10.0.6",
+    "@types/node": "~18.16.20",
```
aruniverse, line 44: do we need to re-add types/node here?

aruniverse (commented):
i see value in switching the resolution-mode to lowest, especially for libraries like ourselves.

but i wonder does it make more sense for us to not use ^ or ~ and just lock it down to the exact version then?

core/markup/package.json
```diff
@@ -56,14 +56,15 @@
     "@itwin/certa": "workspace:*",
     "@itwin/eslint-plugin": "4.0.0-dev.44",
     "@types/chai": "4.3.1",
-    "@types/mocha": "^8.2.2",
+    "@types/mocha": "^10.0.6",
+    "@types/node": "~18.16.20",
```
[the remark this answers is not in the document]
saskliutas: Removed. Missed this when merging.

core/geometry/package.json
saskliutas: Removed. Missed this when merging.

saskliutas:
> > i see value in switching the resolution-mode to lowest, especially for libraries like ourselves.
> > but i wonder does it make more sense for us to not use ^ or ~ and just lock it down to the exact version then?
> 
> I don't think we should move away from `^` in version specifiers. Locking `dependencies` versions would not allow our consumers to get latest versions of our dependencies when consuming our packages. I guess we could lock `devDependencies` versions but that might result in duplicate packages being installed because `pnpm` cannot resolve them to the lowest common version.

aruniverse (approved):
as a library maintainer, i see value in this setting forcing us to be honest about our dependencies. 

as long as this setting applies to our direct & (dev) deps, and not transitive/sub dependencies this is fine as we should be able to pick up security fixes outside our direct deps.

wgoehrig (approved):

calebmshafer (approved):
Approving with a caveat and a hopeful follow up here. This to me seems to be trading one problem for a different one. Both of which are rare.

I agree that this is a good flag to use as it is deterministic rather than our current in-determinate state. However, I think we need to be pushing forward either our dependency minimum more quickly (through automation), while avoiding breaking changes and/or using locally the sort of newest versions of our dependencies to find the other end of issues.

ColinKerr (approved):

paulius-valiunas:
> Approving with a caveat and a hopeful follow up here. This to me seems to be trading one problem for a different one. Both of which are rare.
> 
> I agree that this is a good flag to use as it is deterministic rather than our current in-determinate state. However, I think we need to be pushing forward either our dependency minimum more quickly (through automation), while avoiding breaking changes and/or using locally the sort of newest versions of our dependencies to find the other end of issues.

Not sure what "the other end of issues" means. If you mean someone publishing a breaking change as a minor version of one of our dependencies, then:
1. it's an extremely rare problem, I wouldn't even compare it to how many times we've broken semver by inadvertently using features not available in our declared dependency versions without even knowing.
2. even if it happens, it's not our problem, there will be lots of complaints from all their users, and the library authors will have to fix it one way or another (usually rolling back the breaking change).
3. someone else not following semver is not a reason for us to stop following it too.
4. Using the old pnpm setting does not prevent the issue, as "older" versions can still be pulled in if you have multiple different dependencies on the same library anywhere in the monorepo - pnpm will try to dedupe them and pick the highest _common_ version. In other words, no package manager was ever _guaranteed_ to get you the latest version, and this does not change that.

PR #6242, opened 2023-11-29
Multiply glyph alpha with surface alpha

### LINKED ISSUE #6239: Raster text ignores surface transparency
**Describe the bug**
The defaultAppearance we set when calling `EmphasizeElements.emphasizeElements` does not always get applied to 3D text graphic. 
It works fine when we're zoomed in close, but as we move out, the text goes back to its original style.

Please refer to the attached video below

**To Reproduce**
1. Open an iModel that has 3D text graphics. (The dataset I'm using is in QA if need be)
2. Emphasize some elements of the model with a default apperance.

**Expected behavior**
Default appearance should be applied regardless of zoom level

- OS: Windows 11
- Browser: Chrome
- Version: 119.0.6045.160 (Official Build) (64-bit) (cohort: Stable) 
- iTwin.js Version: 4.2.1

**Additional context**
Here's a video of the behavior:

https://github.com/iTwin/itwinjs-core/assets/3418768/b2a648a2-2ed2-46bc-83cb-22235f5d4273

pmconne:
Fixes #6239.

- [x] Run ImageTests

### DIFF
--- a/common/changes/@itwin/core-frontend/pmc-raster-glyph-transparency_2023-11-29-20-30.json
+++ b/common/changes/@itwin/core-frontend/pmc-raster-glyph-transparency_2023-11-29-20-30.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Fix incorrect transparency for text rendered as raster glyphs.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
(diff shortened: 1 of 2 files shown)


aruniverse (approved):

markschlosseratbentley (approved):

pmconne:
@a-gagnon do you need this backported to 4.2.x?

pmconne:
ImageTests showed up two views of the same iModel where text that is supposed to be semi-transparent now renders as such when drawn as raster glyphs.

a-gagnon:
> @a-gagnon do you need this backported to 4.2.x?

No, I think it's fine if we pick it up in the next minor release.
Thanks!

PR #6244, opened 2023-11-30
Allow RC on older Maj version

DanRod1999:
Add new parameter called `releaseCandidatePrevMaj` which runs the the create branch step job.

Skips other RC tasks as I believe they are unnecessary/unwanted in this circumstance.

Once branch is created I believe we can run `minor` bump once the new "RC" clears all its checks

Need to figure out what steps are needed in the updateMaster job because I believe some of them are necessary but not sure about all of them.

### DIFF
--- a/common/config/azure-pipelines/jobs/version-bump.yaml
+++ b/common/config/azure-pipelines/jobs/version-bump.yaml
@@ -26,6 +26,7 @@ parameters:
     - minor
     - patch
     - releaseCandidate
+    - releaseCandidatePrevMaj
 
 trigger: none
 pr: none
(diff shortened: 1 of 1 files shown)


aruniverse:
what happens if i run this pipeline with `minor` on a previous `release/X.x` branch ?

DanRod1999:
> what happens if i run this pipeline with `minor` on a previous `release/X.x` branch ?

`node common/scripts/install-run-rush version --override-bump ${{ parameters.BumpType }} --version-policy prerelease-monorepo-lockStep --bump`
`# When creating a minor release, the NextVersion.md need to be cleared and the contents placed into a {Version Number}.md file`
`git tag -a release/$(getVersion.version) -m "v$(getVersion.version)"`

these three steps occur when you run minor bump
No RC is created (don't know if thats important)
  this also implies no 3.8.0-dev.x version are published to npm
release branch isn't created (this happens when we run the pipeline as `releaseCandidate`


just running `minor` on this would work if we create the release branch for the new minor manually I think, but this is just an educated guess. I don't know how:
`node common/scripts/install-run-rush version --override-bump ${{ parameters.BumpType }} --version-policy prerelease-monorepo-lockStep --bump`
behaves when the current version is already `--version-policy prerelease-monorepo-lockStep`

PR #6250, opened 2023-11-30
Thematic transparency

### LINKED ISSUE #6241: Enable Transparent Scientific Visualization
### Discussed in https://github.com/iTwin/itwinjs-core/discussions/6212

See this comment for a potential way to fix this: https://github.com/iTwin/itwinjs-core/discussions/6212#discussioncomment-7705644

<div type='discussions-op-text'>

<sup>Originally posted by **FelixGirard** November 17, 2023</sup>
I've been able to use the scientific visualization to display what I want (basically a 3d heatmap). 
But I would like to be able to modify its transparency. 
I've tried using custom transparent colors for the [ThematicGradientSettings](https://www.itwinjs.org/reference/core-common/symbology/thematicgradientsettings/), but that didn't work (still appears solid).

It's not graphics on an iModel element, just a [Polyface](https://www.itwinjs.org/reference/core-geometry/polyface/polyface/) living in a [decorator](https://www.itwinjs.org/reference/core-frontend/views/decorator/).

Is there a way to achieve this? 
If not, could a new transparency parameter be added to the [ThematicGradientSettings](https://www.itwinjs.org/reference/core-common/symbology/thematicgradientsettings/)?
 
 Thanks!

</div>

pmconne:
Fixes #6241.

`ThematicGradientSettings` permits the use of custom colors with alpha channels, but the alpha channel is not properly handled by thematic display nor analysis styles ("scientific visualization"), in different ways.

Depending on the combination of transparency values in the gradient and the surface itself, we may need to render the surface in both the opaque and translucent passes, discarding pixels based on alpha in one pass or the other - the same way we do for surfaces with textures that contain a mix of opaque and transparent texels.

### Thematic display

Thematic display always ignored the gradient's alpha channel, using only the alpha of the underlying surface color.

I added a `transparencyMode` specifying either to ignore the gradient alpha (default), or to multiply it with the surface's alpha (new).
The latter behaves like material alpha - it can make a surface more transparent, but never more opaque.

You can test this using the Thematic Display controls in display-test-app's View Attributes dialog. A new checkbox enables the new transparency mode. Two new custom color schemes are available in the combo box - one containing all transparent colors, another containing a mix of opaque and transparent colors. Both of these contain one fully-transparent color.

### Analysis style

The analysis style creates a texture from the gradient and intends to completely ignore the surface's own color and alpha in favor of the texture's color and alpha. However, `SurfaceGeometry.getPass` did not take this texture into account when computing which pass(es) to render in - instead it used the surface's base color and alpha. You could define a gradient with transparency, and that would work correctly only if your surface already had transparency. If the surface was opaque, it'd render in the opaque pass and no transparency would be applied.

I updated `getPass` to base the render pass(es) on the gradient's alpha channel, ignoring that of the surface.

You can test this using the "Open Analysis Example" option when starting up display-test-app. I modified the styles it creates for height channels to use a mix of opaque, semi-transparent, and fully-transparent colors. The default view has transparency disabled due to solid fill render mode, so the gradient alpha channel is correctly ignored. If you switch to smooth shade render mode, the transparency will be correctly applied.

NOTE: the new `transparencyMode` has no effect on analysis styles - they fully override the surface color+alpha.

### TODO

- [x] Evaluate ImageTests results.
- [x] Document that `transparencyMode` has no effect for analysis styles.
- [x] NextVersion.md


markschlosseratbentley (approved):
I took it for a spin, and it worked great.  Nice stuff!

PR #6254, opened 2023-12-01
Update gather-docs.yaml for 3.8.0 release

DanRod1999:
Looks like we never properly switched this to the correct branch on 3.7.0, so this was still labeled as master.
Once first patch for 3.8.0 is out we will set this to 3.8.x. 

Don't know if its worth changing 3.7.x branch since it will no longer be in use after next week. Thoughts @ben-polinsky @aruniverse ? It is a simple change, so for continuity purpose might as well right?


ben-polinsky:
> Looks like we never properly switched this to the correct branch on 3.7.0, so this was still labeled as master. Once first patch for 3.8.0 is out we will set this to 3.8.x.
> 
> Don't know if its worth changing 3.7.x branch since it will no longer be in use after next week. Thoughts @ben-polinsky @aruniverse ? It is a simple change, so for continuity purpose might as well right?

This isn't too important, IMO.

ben-polinsky (approved):

PR #6280, opened 2023-12-12
Raise an event when mobile backend connects to frontend

tcobbs-bentley:
This is a complement to a future PR that will cause IpcHost.send to throw an exception when the backend and frontend are not connected. (Right now it crashes the mobile app.) Any code that catches that exception can then record the fact somewhere and add an onConnect handler.

### DIFF
--- a/core/mobile/src/backend/MobileHost.ts
+++ b/core/mobile/src/backend/MobileHost.ts
@@ -102,7 +102,7 @@ export class MobileHost {
   /**
    * Raised when the mobile OS informs a mobile app that it is running low on memory.
    *
-   * __Note:__ iOS and iPadOS send this warning so often as to make it not very useful.
+   * @note iOS and iPadOS send this warning so often as to make it not very useful.
    */
   public static readonly onMemoryWarning = new BeEvent();
   /**
@@ -120,7 +120,7 @@ export class MobileHost {
   /**
    * Raised after a mobile backend connects to the mobile frontend.
    *
-   * __Note:__ this will be raised at startup, and it will also be raised every time the app returns
+   * @note this will be raised at startup, and it will also be raised every time the app returns
    * to the foreground from the background.
    */
   public static readonly onConnected = new BeEvent();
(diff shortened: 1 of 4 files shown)


tcobbs-bentley:
@swbsi FYI

nick4598:
> @aruniverse: I ran `rush extract-api`, and yet that check is failing. I just ran it again locally, and it produced no changed files.

Have you ran rush clean && rush build locally before running extract-api?

core/mobile/src/backend/MobileHost.ts
```diff
@@ -99,11 +99,38 @@ export interface MobileHostOpts extends NativeHostOpts {
...
+  /**
+   * Raised when a mobile app is about to enter the background.
+   */
   public static readonly onEnterBackground = new BeEvent();
+  /**
+   * Raised after a mobile backend connects to the mobile frontend.
+   *
+   * __Note:__ this will be raised at startup, and it will also be raised every time the app returns
+   * to the foreground from the background.
+   */
```
pmconne, line 125:
```suggestion
  /**
   * Raised when the mobile OS informs a mobile app that it is running low on memory.
   *
   * @note iOS and iPadOS send this warning so often as to make it not very useful.
   */
  public static readonly onMemoryWarning = new BeEvent();
  /**
   * Raised when the device orientation changes on a device running a mobile app.
   */
  public static readonly onOrientationChanged = new BeEvent();
  /**
   * Raised after a mobile app returns to the foreground.
   */
  public static readonly onEnterForeground = new BeEvent();
  /**
   * Raised when a mobile app is about to enter the background.
   */
  public static readonly onEnterBackground = new BeEvent();
  /**
   * Raised after a mobile backend connects to the mobile frontend.
   *
   * @note this will be raised at startup, and it will also be raised every time the app returns
   * to the foreground from the background.
   */
```

pmconne (approved):

tcobbs-bentley:
@Mergifyio backport release/4.3.x

PR #6289, opened 2023-12-12
Fix mobile backend to frontend communication

tcobbs-bentley:
The main fix is to make MobileIpc.sendToFrontend not be async so that if an error happens it does not crash the whole app. This required RpcMarshaling.serialize to also not be async.


tcobbs-bentley:
@swbsi FYI

toddsouthenbentley (approved):

GytisCepk (approved):

swbsi (approved):

aruniverse (approved):

PR #6291, opened 2023-12-13
Add glyph data to TextString

fjch1997:
There is the ability to include text layout data in the flatbuffer of TextString rather than computing them at display time. Batch connector (in C++) uses this to preserve visual fidelity between MicroStation and IModels. This PR adds the same data in the TypeScript library so that the live sync connector can do this as well.

### DIFF
--- a/common/api/core-common.api.md
+++ b/common/api/core-common.api.md
@@ -9225,6 +9225,8 @@ export class TextString {
     constructor(props: TextStringProps);
     bold?: boolean;
     font: FontId;
+    glyphIds?: number[];
+    glyphOrigins?: Point2d[];
     // (undocumented)
     height: number;
     italic?: boolean;
@@ -9254,6 +9256,8 @@ export interface TextStringPrimitive {
 export interface TextStringProps {
     bold?: boolean;
     font: FontId;
+    glyphIds?: number[];
+    glyphOrigins?: Point2d[];
     // (undocumented)
     height: number;
     italic?: boolean;
--- a/common/api/core-common.api.md
+++ b/common/api/core-common.api.md
@@ -2637,7 +2637,7 @@ export namespace ElementGeometry {
     export function fromGeometryQuery(geom: GeometryQuery, worldToLocal?: Transform): ElementGeometryDataEntry | undefined;
     export function fromImageGraphic(image: ImageGraphicProps, worldToLocal?: Transform): ElementGeometryDataEntry | undefined;
     export function fromSubGraphicRange(bbox: ElementAlignedBox3d): ElementGeometryDataEntry | undefined;
-    export function fromTextString(text: TextStringProps, worldToLocal?: Transform): ElementGeometryDataEntry | undefined;
+    export function fromTextString(text: TextStringProps, worldToLocal?: Transform, glyphs?: TextStringGlyphData): ElementGeometryDataEntry | undefined;
     export function getBRepEntityType(entry: ElementGeometryDataEntry): BRepEntity.Type | undefined;
     export function isAppearanceEntry(entry: ElementGeometryDataEntry): boolean;
     export function isCurve(entry: ElementGeometryDataEntry): boolean;
@@ -2690,6 +2690,7 @@ export namespace ElementGeometry {
     export function toImageGraphic(entry: ElementGeometryDataEntry, localToWorld?: Transform): ImageGraphicProps | undefined;
     export function toSubGraphicRange(entry: ElementGeometryDataEntry): ElementAlignedBox3d | undefined;
     export function toTextString(entry: ElementGeometryDataEntry, localToWorld?: Transform): TextStringProps | undefined;
+    export function toTextStringGlyphData(entry: ElementGeometryDataEntry): TextStringGlyphData | undefined;
     export function toTransform(sourceToWorld: Float64Array): Transform | undefined;
     export function transformBRep(entry: ElementGeometryDataEntry, inputTransform: Transform): boolean;
     export function updateGeometryParams(entry: ElementGeometryDataEntry, geomParams: GeometryParams, localToWorld?: Transform): boolean;
@@ -9225,8 +9226,6 @@ export class TextString {
     constructor(props: TextStringProps);
     bold?: boolean;
     font: FontId;
-    glyphIds?: number[];
-    glyphOrigins?: Point2d[];
     // (undocumented)
     height: number;
     italic?: boolean;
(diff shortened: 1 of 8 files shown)


core/common/src/geometry/TextString.ts
```diff
@@ -59,6 +63,10 @@ export class TextString {
   public readonly origin: Point3d;
   /** Rotation relative to element's placement */
   public readonly rotation: YawPitchRollAngles;
+  /** Optional cached text layout information. If undefined, it will be computed at display time. Default is undefined. */
```
pmconne, line 66: Can you be more specific than "cached text layout information"?

bbastings: > Can you be more specific than "cached text layout information"?

Maybe add something like "used to support legacy proxy graphics" or something like that. Nothing should use this except the V8 connector, so it doesn't need to be that detailed.

common/api/core-common.api.md
```diff
@@ -9254,6 +9256,8 @@ export interface TextStringPrimitive {
 export interface TextStringProps {
     bold?: boolean;
     font: FontId;
+    glyphIds?: number[];
+    glyphOrigins?: Point2d[];
```
bbastings, line 9260: This should be XYProps and not Point2d on TextStringProps...

core/common/src/geometry/TextString.ts
fjch1997: I've added the comments as you suggested. Also, I assume no changelog entry is needed for this? I just committed ones with empty comments.

core/common/src/geometry/TextString.ts
```diff
@@ -59,6 +63,10 @@ export class TextString {
   public readonly origin: Point3d;
   /** Rotation relative to element's placement */
   public readonly rotation: YawPitchRollAngles;
+  /** Optional cached text layout information used to support legacy proxy graphics. If undefined, it will be computed at display time. Default is undefined. */
+  glyphIds?: number[];
+  /** Optional cached text layout information used to support legacy proxy graphics. If undefined, it will be computed at display time. Default is undefined. */
+  glyphOrigins?: Point2d[];
   public get width() { return this.height * (this.widthFactor ? this.widthFactor : 1.0); }
 
   public constructor(props: TextStringProps) {
```
bbastings, line 72: constructor doesn't do anything with glyph info from TextStringProps?

core/common/src/geometry/ElementGeometry.ts
```diff
@@ -1093,6 +1112,12 @@ export namespace ElementGeometry {
       builder.addTransform(fbb, transformOffset);
     }
 
+    if (glyphs) {
+      builder.addRange(fbb, EGFBAccessors.TextStringRange.createTextStringRange(fbb, 0, 0, 20741, 2500));
```
bbastings, line 1116: This hard-coded range can't be right...probably should ask Jeff Marker about this/whether it's needed since we can't compute it in typescript.

fjch1997: Sorry, it was something I did for testing. Will correct it

fjch1997: Looks like it has to be passed as a property as well.

fjch1997: The C++ code populates the range together with glyphs, so I assume it's needed

bbastings: I still think it would be better to go with my initial suggestion and just to add an optional "glyph data" arg to ElementGeometry.Builder.appendTextString/ElementGeometry.fromTextString instead of adding things to TextString/TextStringProps that aren't generally useful and are going to be ignored by the json GeometryStreamBuilder now. Any opinion @pmconne ?

pmconne: Lacking some context, but I take it there is no desire/need to retrieve the glyph data from a persistent geometry stream, only to write it to a new geometry stream?

bbastings: Correct, this is strictly to preserve the V8 glyph layout when it uses options that are no longer supported bt itwin. Only the native code needs to look at it for display and it's lost on edit.

fjch1997: The only reason I went with this was for unit tests. But I can create new tests in `core-common` rather than modifying the ones in `core-backend`

bbastings: You will still have a class/props for the glyph data in common just like TextString/TextStringProps that can be used by the backend. Not sure why you think it affects unit tests?

fjch1997: Yes, I can create tests in `core-common`. The current code modifies existing unit tests in `core-backend`

bbastings: I don't understand why you think you can't continue to create the tests in core-backend. All the geometry stream tests are in core-backend...there is no reason this one needs to be in common, please explain???

fjch1997: Right. `core-backend` does not import `flatbuffer` and `EGFBAccessors` which is internal to `core-common`. I needed them to read the flatbuffer.

bbastings: NOTE: The current "create GeometricElement3d from world coordinate text using a newly added font" test is using the json builder, which is not going to persist the glyph data (because the backend native fromJSON code implementation)...you will need to write a new text test for the flatbuffer builder.

bbastings: It doesn't need to, you are calling ElementGeometry.Builder/Iterator methods.

fjch1997: It looks to me that the current "create GeometricElement3d from world coordinate text using a newly added font" test is calling `fromTextString` which builds the flatbuffer with the glyphs and passing it to `insertElement`. So it does persist the glyph data....?
But anyways. I can revert the changes in `GeometryStream.test.ts` since it doesn't have a way to read the flat buffer if the reading code is removed from `toTextString`. Then, I can add the tests in the `core-common` package

bbastings: > So it does persist the glyph data....?

Yes, the flatbuffer test should persist the data, the json test/builder does not...which is one of the reasons I don't think this should be added to TextStringProps.

Feel free to call me if this still isn't clear...

core/common/src/geometry/TextString.ts
jffmarker: I agree that only a V8 Connector could reasonably set these. `glyphOrigins` is pretty straightforward, but `glyphIds` is more of an internal implementation detail.

pmconne (approved):

PR #6300, opened 2023-12-14
Expose `ParticleCollectionBuilder` for extensions

GytisCepk:
`ParticleCollectionBuilder` namespace is not available in iTwin.js extensions and, as a result, something like [snow and rain particle effects](https://www.itwinjs.org/sandboxes/iTwinPlatform/Snow%20and%20Rain%20Particle%20Effect) can't be implemented in extension.


ben-polinsky (approved):

markschlosseratbentley:
@AJMigliore fyi

markschlosseratbentley (approved):

PR #6307, opened 2023-12-15
We should be able to create an EcefLocation directly from a transformation matrix.

MarcBedard8:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/1009

We should be able to create an EcefLocation directly from a transformation matrix and have the ECEFLocation keep that matrix when passed as props in the workflow.


MarcBedard8:
c.f. this issue:  https://github.com/iTwin/itwinjs-backlog/issues/1009

pmconne:
> We should be able to

Why, vs just passing the `origin` and `orientation` that define the transform?
I can understand the motivation for a `createFromTransform` method. I don't understand the motivation for adding an optional `Transform` to the JSON representation.

core/common/src/IModel.ts
```diff
@@ -68,6 +68,9 @@ export interface EcefLocationProps {
   readonly xVector?: XYZProps;
   /** Optional Y column vector used with [[xVector]] to calculate potentially non-rigid transform if a projection is present. */
   readonly yVector?: XYZProps;
+  /** Optional transform from which all other props will be computed - potentially non-rigid transform if a projection is present. */
+  /**          when provided, will have priority over others props */
```
pmconne, line 72:
```suggestion
  /** Optional potentially non-rigid transform defining the ECEF location.
   * @note If this property is supplied, the other properties are ignored.
   */
```

core/common/src/IModel.ts
```diff
@@ -68,6 +68,9 @@ export interface EcefLocationProps {
   readonly xVector?: XYZProps;
   /** Optional Y column vector used with [[xVector]] to calculate potentially non-rigid transform if a projection is present. */
   readonly yVector?: XYZProps;
+  /** Optional transform from which all other props will be computed - potentially non-rigid transform if a projection is present. */
```
pmconne: Update the interface documentation above "If the xVector and yVector properties are define then they take precedence..."

core/common/src/IModel.ts
```diff
@@ -285,6 +294,18 @@ export class EcefLocation implements EcefLocationProps {
     return new EcefLocation({ origin: ecefOrigin, orientation: YawPitchRollAngles.createFromMatrix3d(matrix)!, cartographicOrigin: origin });
   }
 
+  /** Construct ECEF Location from transform with optional position on the earth used to establish the ECEF origin and orientation. */
+  public static createFromTransform(transform: Transform): EcefLocation {
+    const ecefOrigin = transform.getOrigin();
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const angleFromInput = YawPitchRollAngles.createDegrees(0,0,0);
```
pmconne, line 301:
```suggestion
    const angleFromInput = YawPitchRollAngles.createDegrees(0,0,0);
```
`angleFromInput` isn't unused.

core/common/src/IModel.ts
```diff
@@ -285,6 +294,18 @@ export class EcefLocation implements EcefLocationProps {
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const locationOrientationFromInputT = YawPitchRollAngles.createFromMatrix3d(transform.getMatrix(), angleFromInput);
```
pmconne, line 303:
```suggestion
    YawPitchRollAngles.createFromMatrix3d(transform.getMatrix(), angleFromInput);
```
Address lint rule instead of suppressing it.

core/frontend/src/tile/ThreeDTileFormatInterpreter.ts
```diff
@@ -82,14 +82,17 @@ export class ThreeDTileFormatInterpreter  {
...
-          if (!worldToEcefTransform.matrix.isIdentity && locationOrientation !== undefined && locationOrientation.angles !== undefined)
-            ecefLocation = new EcefLocation({ origin: locationOrientation.origin, orientation: locationOrientation.angles.toJSON() });
-          else
-            ecefLocation = EcefLocation.createFromCartographicOrigin(cartoCenter!);
+          if (worldToEcefTransform.matrix.isIdentity) {
+            // For georeferenced Reality Meshes, its origin is translated to model origin (0,0,0).
+            // Apply range center to translate it back to its original position.
+            const worldCenter = !worldToEcefTransform.matrix.isIdentity ? range.center : undefined;
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            ecefLocation = EcefLocation.createFromCartographicOrigin(cartoCenter!, worldCenter);
```
pmconne, line 93: If you are confident that `cartoCenter` can't possibly ever be null or undefined, `assert` it. Otherwise, handle it.

core/common/src/IModel.ts
MarcBedard8: Fix pushed

core/common/src/IModel.ts
MarcBedard8: Fixed

core/frontend/src/tile/ThreeDTileFormatInterpreter.ts
MarcBedard8: Fixed

pmconne (approved):

PR #6309, opened 2023-12-19
Schema merging/custom attribute merging

StefanApfel-Bentley:


core/ecschema-editing/src/Merging/ClassMerger.ts
```diff
@@ -124,6 +124,38 @@ export class ClassMerger<TClass extends ECClass> {
...
+            ...sourceCustomAttribute,
+            className: targetCustomAttribute.fullName,
+          };
+
+          const results = await this.context.editor.entities.addCustomAttribute(classKey, customAttribute);
+          if (results.errorMessage !== undefined) {
+            return {  errorMessage: results.errorMessage };
+          }
+        } else {
+          return { errorMessage: `Changes of Custom Attribute ${customAttributeContainerChange.ecTypeName} on ${classKey.name} merge is not implemented.`};
```
ColinKerr, line 152: This might be a case where we have to be a bit more flexible and not error out when we hit a difference in a custom attribute ... I guess we should see how it goes for now, but keep it in mind

core/ecschema-editing/src/Merging/SchemaMerger.ts
```diff
@@ -123,3 +125,41 @@ function * filterChangesByItemType<TChange extends SchemaItemChanges>(changes: M
...
   }
 }
+
+/**
+ * Merges the custom attribute changes on schema level to the target schema.
+ * @param mergeContext  The current schema merging context.
+ * @param changes       An iterable with custom attribute changes on the schema.
+ * @returns             A SchemaEditResults object.
+ */
+async function mergeCustomAttributes(mergeContext: SchemaMergeContext, changes: Iterable<CustomAttributeContainerChanges>): Promise<SchemaEditResults> {
```
ColinKerr, line 135: It seems wrong that we have essentially the same method 3 times.

core/ecschema-editing/src/test/Merging/CustomAttributeMerger.test.ts
```diff
@@ -0,0 +1,309 @@
...
+          },
+          {
+            StringPrimitiveArrayProp: [
+              "SchemaCustomAttribute",
+            ],
+            className: "TestSchema.TestCA",
+          },
+        ],
+      );
+    });
```
ColinKerr, line 307: you need a  test where the CA is defined in a schema referenced by the source but not by the target.

core/ecschema-editing/src/Merging/ClassMerger.ts
StefanApfel-Bentley: I added a backlog item #6311

core/ecschema-editing/src/Merging/ClassMerger.ts
```diff
@@ -124,6 +125,38 @@ export class ClassMerger<TClass extends ECClass> {
...
+  //           return {  errorMessage: results.errorMessage };
+  //         }
+  //       } else {
+  //         return { errorMessage: `Changes of Custom Attribute ${customAttributeContainerChange.ecTypeName} on ${classKey.name} merge is not implemented.`};
+  //       }
+  //     }
+  //   }
+  //   return { itemKey: classKey };
+  // }
+
```
ColinKerr, line 159: don't leave commented out code

ColinKerr (approved):

PR #6310, opened 2023-12-19
Support civil use case that requires update be able to clear the geometry stream.

bbastings:
Added test for clearing the geometry stream.

imodel-native: https://github.com/iTwin/imodel-native/pull/580


pmconne (approved):

dassaf4 (approved):

pmconne (changes requested):
You merged master. Please roll that back.

PR #6313, opened 2023-12-20
Merging schemas kind of quantities 

AureMonke:


core/ecschema-editing/src/test/Merging/KindOfQuantityMerger.test.ts
```diff
@@ -0,0 +1,381 @@
...
+      },
+      TestPhenomenon: {
+        schemaItemType: "Phenomenon",
+        definition: "Test Phenomenon",
+      },
+      TU_PER_TU: {
+        schemaItemType: "Unit",
+        unitSystem: "ReferenceSchema.TestUnitSystem",
+        phenomenon: "ReferenceSchema.TestPhenomenon",
+        definition: "TU",
```
ColinKerr, line 44: it may not actually matter here but I'd like to keep the units we use in testing valid.  So the definition should be TU*TU(-1) and TU should be defined as a unit.  The unit TU can have itself as its definition because in this fictional unit system TU is a base unit.  

See https://github.com/iTwin/bis-schemas/blob/master/Standard/Units.ecschema.xml for an example of real units

core/ecschema-editing/src/test/Merging/KindOfQuantityMerger.test.ts
```diff
@@ -0,0 +1,381 @@
...
+          },
+          TestKoq: {
+            schemaItemType: "KindOfQuantity",
+            label: "Test",
+            description: "Description of koq",
+            relativeError: 0.0030480000000000004,
+            persistenceUnit: "ReferenceSchema.TU_PER_TU",
+            presentationUnits: [
+              "ReferenceSchema.TestDecimal(4)[ReferenceSchema.TU_PER_TU|undefined]",
+              "ReferenceSchema.TestDecimal(5)[ReferenceSchema.TU_HORIZONTAL_PER_TU_VERTICAL|kJ/K]",
```
ColinKerr, line 230: avoid using an incorrect display label like kJ/K for clarity sake

core/ecschema-editing/src/test/Merging/KindOfQuantityMerger.test.ts
```diff
@@ -0,0 +1,381 @@
...
+    name: "ReferenceSchema",
+    version: "1.2.0",
+    alias: "reference",
+    items: {
+      TestUnitSystem: {
+        schemaItemType: "UnitSystem",
+      },
+      TestPhenomenon: {
+        schemaItemType: "Phenomenon",
+        definition: "Test Phenomenon",
```
ColinKerr, line 38: The definition here should be a valid definition.  if this is a 'base' phenomenon then the definition should be the same as the name of the phenomenon.

core/ecschema-editing/src/test/Merging/KindOfQuantityMerger.test.ts
```diff
@@ -0,0 +1,381 @@
...
+            name: "ReferenceSchema",
+            version: "1.2.0",
+          },
+        ],
+        items: {
+          TU_PER_HR: {
+            schemaItemType: "Unit",
+            unitSystem: "ReferenceSchema.TestUnitSystem",
+            phenomenon: "ReferenceSchema.TestPhenomenon",
+            definition: "TU*HR(-1)",
```
ColinKerr, line 118: This unit is not dimensionally compatible with the TestPhenomenon so this is not technically a valid unit definition.

ColinKerr (approved):

PR #6315, opened 2023-12-20
Support civil use case that requires update be able to clear the geometry stream.

bbastings:
imodel-native: https://github.com/iTwin/imodel-native/pull/580


pmconne (approved):
Need to [fix the insert test](https://github.com/iTwin/itwinjs-core/pull/6310/commits/eca40d5147cf277db1357f985bf7e51df45aa454) again.

PR #6318, opened 2023-12-20
Update ChangesetReader API and relaxing requirement for ECClassId in update change 

khanaffan:
This change include two fixes reported by developers

1. Make `$meta.ClassName` compatible with other APIs (classFullName).
2. Change unifier failed with update change was processed as the update does not have ECClassId. Now the check is more flexible and Api fallback to root class for the table and does not fail.


MichaelBelousov (approved):

pmconne (approved):

nick4598:
@mergifyio backport release/4.3.x

PR #6319, opened 2023-12-21
SchemaDifference API

StefanApfel-Bentley:
Translates the schema comparison diagnostics into a JSON prop inspired object model to easily see the differences of the compared schema to the target schema. 
- Almost all non-primitive types got enhanced by a property $changeType which reflects whether an item has been changed (like different label) or if it had been missing in the target schema it compared with.


core/ecschema-editing/src/Differencing/SchemaDiagnosticVisitor.ts
```diff
@@ -0,0 +1,594 @@
...
+ */
+export class SchemaDiagnosticVisitor {
+
+  private readonly _differenceReport: SchemaDifferences;
+
+  constructor(differenceReport: SchemaDifferences) {
+    this._differenceReport = differenceReport;
+  }
+
+  private addEntry<T extends AnySchemaDifference>(entry: T): T {
```
christophermlawson, line 68: Maybe move the private methods to the bottom of the class?

christophermlawson (commented):
Looks good. Just that one comment.

christophermlawson (approved):

PR #6321, opened 2023-12-21
Give selectDecoration a chance to handle terrain/map picks.

bbastings:
This change should help with a request for showing location dependent map layer information when using the default SelectTool. Add release tag for TextStringGlyphData.

### DIFF
--- a/common/api/core-common.api.md
+++ b/common/api/core-common.api.md
@@ -9245,7 +9245,7 @@ export class TextString {
     widthFactor?: number;
 }
 
-// @public
+// @beta
 export interface TextStringGlyphData {
     // (undocumented)
     glyphIds: number[];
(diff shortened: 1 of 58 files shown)


common/api/core-common.api.md
```diff
@@ -9245,7 +9245,7 @@ export class TextString {
     widthFactor?: number;
 }
 
-// @public
+// @beta
```
aruniverse, line 9248: this is breaking

bbastings: Not breaking, this was just added, and it's only used by the ElementGeometry api which is all beta...

aruniverse (approved):

common/api/core-common.api.md
pmconne: Added in #6291, merged 3 days ago. Not yet in any release.

PR #6323, opened 2023-12-21
Fix comparison of schemas with same name version

julioalvahash:
This PR removes the keyMismatch comparison when comparing two schemas, this will allow the API to create two reports every time regardless of the schema names and version.

### DIFF
--- a/core/ecschema-editing/src/test/Validation/SchemaComparisonSameNameVersion.test.ts
+++ b/core/ecschema-editing/src/test/Validation/SchemaComparisonSameNameVersion.test.ts
@@ -2,8 +2,8 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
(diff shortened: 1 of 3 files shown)


core/ecschema-editing/src/test/Validation/SchemaComparisonSameNameVersion.test.ts
```diff
@@ -0,0 +1,325 @@
...
+import { expect } from "chai";
+
+class SchemaCompareReporter implements ISchemaCompareReporter {
+  public changes: SchemaChanges[] = [];
+  public report(schemaChanges: ISchemaChanges): void {
+    this.changes.push(schemaChanges as SchemaChanges);
+  }
+}
+
+function findDiagnostic(diagnostics: AnyDiagnostic[], code: string, fullNameA: string, _fullNameB?: string, _propertyType?: string) {
```
ColinKerr, line 16: why have _fullNameB and _propertyType if you never use them?

julioalvahash: There was no need for them for those diagnostics, removed.

ColinKerr (approved):

PR #6324, opened 2023-12-21
Fixed an issue preventing tiles published by some specific ArcGIS MapServer to be correctly displayed in the view.

mdastous-bentley:
This fix is intended to mitigate a limitation of the [ArcGIS Tile Map REST API](https://developers.arcgis.com/rest/services-reference/enterprise/tile-map.htm).

As documented, the service returns error code 422 when no tiles are available for the request tilemap area.  Unfortunately, this is **only** true within the data bundle boundaries in which the top-left corner fall into:
_The response will only contain the part of the requested area that is in the same bundle as the top-left corner._

Which means, when you receive error 422 you can't tell for sure if all tiles for the requested tile map are all non-visible, or if only a subset of it is non-visible (i.e. clipped by the data bundle boundaries). 

By studying the requests made by ESRI's own web viewer, and by testing several datasets, we concluded that lowering the requested tile map area from 32x32 tiles to 8x8 tiles was solving the issue.

A bug has been filled at ESRI at the beginning of this year, but has been addressed : BUG-000158087


mdastous-bentley:
@khanaffan
```
Changeset Reader API
       Changeset reader / EC adaptor:

      AssertionError: expected '0x3d' to equal '0x3f'
      + expected - actual

      -0x3d
      +0x3f
      
      at Context.<anonymous> (src\test\standalone\ChangesetReader.test.ts:235:14)
```

khanaffan:
> @khanaffan
> 
> ```
> Changeset Reader API
>        Changeset reader / EC adaptor:
> 
>       AssertionError: expected '0x3d' to equal '0x3f'
>       + expected - actual
> 
>       -0x3d
>       +0x3f
>       
>       at Context.<anonymous> (src\test\standalone\ChangesetReader.test.ts:235:14)
> ```
I recently fixed it. Just merge master into your branch.
https://github.com/iTwin/itwinjs-core/pull/6334#event-11372986567

pmconne (approved):

mdastous-bentley:
@Mergifyio backport release/4.3.x

PR #6341, opened 2024-01-04
classIdPresentInChange should be true when ecClassId is present

nick4598:
Without this change I get a ChangedECInstance which has no ECClassId or fallbackClassId

image

With the change a classId is present:
image

### DIFF
--- a/common/changes/@itwin/core-backend/nick-changesetecadaptor-fixmaybe_2024-01-04-21-23.json
+++ b/common/changes/@itwin/core-backend/nick-changesetecadaptor-fixmaybe_2024-01-04-21-23.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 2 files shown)


nick4598:
@khanaffan Could you review this change?

khanaffan (approved):

pmconne (approved):
Keep an eye on apparent test flakery. I requeud the builds - all three had failed, each differently, one in a previously-flaky test that was recently fixed.

```
  1) Changeset Reader API
       Changeset reader / EC adaptor:

      AssertionError: expected 6 to equal 4
      + expected - actual

      -6
      +4
      
      at Context.<anonymous> (src\test\standalone\ChangesetReader.test.ts:229:14)
```

nick4598:
This test isn't flaky and is a direct result of my change it seems

PR #6342, opened 2024-01-05
Add article providing guidelines regarding Calculated Properties

diegoalexdiaz:


docs/bis/guide/other-topics/calculated-properties.md
```diff
@@ -0,0 +1,47 @@
...
+
+However, the process of the actual calculation of property values needs careful analysis in order to make sure that the cached values do not become out-of-date. Typically it involves capturing the calculation logic in a location that every data-writer that can target the schema uses to write to it. The calculation logic will typically be part of a domain library if the involved schema is at one of the Standardized layers in BIS, and thus, it is expected to be shared. Or it will be part of the Application's logic if the calculated property is part of an Application schema.
+
+This approach is most appropriate for Calculated properties that meet the following criteria:
+
+1. Calculation is expected to be more costly either because it requires joining data from separate Elements or due to the complexity of the computation itself
+1. Calculated properties are expected to be displayed in Standard iTwin.js UX controls such as Property Pane and Grids
+1. Calculated values need to be queried via ECSQL
```
kabentley, line 37: Won't this approach has lots of problems for editing applications? When the calculated properties are stored in the database, they are updated by every user (when?). Then there will always be merge conflicts. Seems unworkable to me.

diegoalexdiaz: We've been discussing this point lately too. There seems to be two cases under that category:

1) Calculated properties that are needed early by the user creating the data (e.g. "Length" property on a Pipe / Pipeline, computed as a straight distance between its two "Port" elements, or along its "centerline") - typically to run queries against them later by the same user (e.g. "Query pipes longer than X").

2) Calculated properties that are needed later in the workflow - typically "summary / aggregation" properties - the ones we've discussed as belonging to "Shared" iModels, computed by Transformations.

This draft article started addressing #1 (and needs to be much clearer about it) assuming the calculation itself is handled by code (either by a shared-domain or app library depending on the case) that all data-writers use (e.g. Pipeline.setCenterline(), which would also compute and cache its "Length" in the iModel). 

If an app allows merge conflicts to happen (e.g. by not using Pessimistic locking before every edit), there needs to be some discipline-specific merge-conflict logic (e.g. Piping-aware) that makes sure the Length value on a "merged Pipe" is still correct even after addressing a conflict. However, it seems our decisions with editing apps have been around minimizing conflicts to happen in the first place, so if apps pessimistically lock elements before editing them, that should prevent conflicts involving Calculated Properties of the #1 kind.

The article needs to addressed the #2 case as well, but that one is not problematic as it just happens in a transform between WIP and Shared iModels.

kabentley: i'm not worried about calculated properties "of" an element (e.g. length). I'm worried about calculated properties that are comprised from (potentially) many elements (e.g. average length). In that case _every_ editor must update the value and there will always be conflicts, since nobody can "lock" the value. If there is some way to specify that the calculated value may *only* appear in a Shared iModel, then that's fine since it may only be updated by a transformation.

diegoalexdiaz: Thanks for your feedback Keith. I'll incorporate it into this article.

PR #6354, opened 2024-01-10
Support per-model display settings for batched tiles

### LINKED ISSUE #6340: Support per-model settings for batched tiles
A view can customize the way a particular model displays in several ways:
- [Apply different clip volumes to groups of models](https://www.itwinjs.org/reference/core-common/views/viewdetails3d/modelclipgroups/)
- [Apply a display transform to individual models](https://www.itwinjs.org/reference/core-frontend/views/viewstate/modeldisplaytransformprovider/)
- Use [PlanProjectionSettings](https://www.itwinjs.org/reference/core-common/displaystyles/displaystyle3dsettings/planprojectionsettings/) to apply an elevation transform and/or transparency to a plan projection model; and/or specify that the model should draw as an overlay.

These work well when each model is associated with its own independent tile tree. They do not currently work at all for batched tile trees, which combine geometry from many models into a single tile tree.

The features above should work as expected with batched tiles.

None.

**Additional context**
PlanProjectionSettings can *also* specify that the display priorities associated with subcategories should be used to control the order in which geometry lying in the same XY plane is drawn. That also does not work with batched tiles, but it is separate from these per-model display settings.

Schedule animations are closely related in that they also can apply a transform on a per-model basis. They do currently work with batched tiles, but their implementation will need to be replaced (generalized) as part of the work required to implement the above features.

pmconne:
Fixes #6340.

Group models that have compatible display settings, then produce a tile tree reference for each group.
When decoding tiles, split the contents based on those groupings.
In typical cases, only one group exists so no tile splitting occurs.
When the display settings change, re-evaluate the groupings and, if they've changed, produce new tile tree references.

The `BENTLEY_BatchedTileSet` extension is now required; it provides the Id, extents, and view flag overrides for every model included in the tileset. Implemented in https://github.com/iTwin/imodel-native-internal/pull/355 which should be merged simultaneously with this PR.

After a viewer upgrades to this version of the @itwin/frontend-tiles package, any tilesets published by an older version of TilesetPublisher will fail to load due to the absence of the metadata described above.

TODO:
- [ ] Evaluate ImageTests results
- [x] Add to display-performance-test-app support for applying display transforms stored in external saved views
- [ ] Add test cases to ImageTests
- [x] NextVersion.md


dassaf4 (approved):

markschlosseratbentley (approved):

test-apps/display-test-app/src/frontend/App.ts
```diff
@@ -25,7 +25,8 @@ import { dtaChannel, DtaIpcInterface } from "../common/DtaIpcInterface";
 import { DtaRpcInterface } from "../common/DtaRpcInterface";
 import { ToggleAspectRatioSkewDecoratorTool } from "./AspectRatioSkewDecorator";
 import { ApplyModelDisplayScaleTool } from "./DisplayScale";
-import { ApplyModelTransformTool } from "./DisplayTransform";
+import { ApplyModelTransformTool, ClearModelTransformsTool, DisableModelTransformsTool } from "./DisplayTransform";
+import { ApplyModelClipTool } from "./ModelClipTools.ts";
```
aruniverse, line 29: surprised this worked

```suggestion
import { ApplyModelClipTool } from "./ModelClipTools";
```

ben-polinsky (approved):

aruniverse (approved):

pmconne:
ImageTests results: some slight z-fighting changes in AnimatedCastle and fixes to plan projections in GeoBridgeDemo and FloridaGIS. @DStradley fyi.

PR #6355, opened 2024-01-10
Fix `RenderMaterialElement.onCloned` not remapping map texture ids

MichaelBelousov:
### Why

Transformations can lose data or cause strange behavior if they do not remap ids in json properties, and the ids are attempted to be reused by another application. In this case, transformations can just lose textured materials.

### What

- fix not remapping texture ids in `RenderMaterial` json properties
- add a test over a mock clone context
  - move some test utilities up in the file to reuse in a new `describe` test block
- add optional types for possible maps in iModel data, see below for questions

### Questions

I am seeking review of the documentation I added for possible maps as observed in the house model sample. I see a potential list of really supported maps [here in the native code](https://github.com/iTwin/imodel-native/blob/116fc61ff5d19c08c0d5296295b44a71c5e2cf71/iModelCore/iModelPlatform/PublicAPI/DgnPlatform/RenderMaterial.h#L89), but the house model had some not listed there iirc.

It seems connectors are generating others not listed. Would it be preferable to specify an index signature in the type instead of adding these new optional explicit ones as I did?

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -4567,6 +4567,8 @@ export class RenderMaterialElement extends DefinitionElement {
     static createCode(iModel: IModelDb, scopeModelId: CodeScopeProps, name: string): Code;
     description?: string;
     static insert(iModelDb: IModelDb, definitionModelId: Id64String, materialName: string, params: RenderMaterialElementParams): Id64String;
+    // @internal (undocumented)
+    protected static onCloned(context: IModelElementCloneContext, sourceProps: ElementProps, targetProps: ElementProps): void;
     paletteName: string;
     // (undocumented)
     toJSON(): RenderMaterialProps;
--- a/common/api/core-common.api.md
+++ b/common/api/core-common.api.md
@@ -7327,8 +7327,17 @@ export namespace RenderMaterial {
 
 // @public
 export interface RenderMaterialAssetMapsProps {
+    Bump?: TextureMapProps;
+    Diffuse?: TextureMapProps;
+    Displacement?: TextureMapProps;
+    Finish?: TextureMapProps;
+    GlowColor?: TextureMapProps;
     Normal?: NormalMapProps;
     Pattern?: TextureMapProps;
+    Reflect?: TextureMapProps;
+    Specular?: TextureMapProps;
+    TranslucencyColor?: TextureMapProps;
+    TransparentColor?: TextureMapProps;
 }
 
 // @public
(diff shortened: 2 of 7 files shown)


pmconne:
> It seems connectors are generating others not listed.

Can you provide examples, including if possible what data is associated with them? (i.e., do they all contain a `TextureId` property?) Do they have "PBR" in their names?

> Would it be preferable to specify an index signature in the type instead of adding these new optional explicit ones as I did?

This JSON blob is not really intended to be extensible to the point where the core code can't know what if anything inside it needs to be remapped.

MichaelBelousov:
> This JSON blob is not really intended to be extensible to the point where the core code can't know what if anything inside it needs to be remapped.

The implementation I provided will however remap unknown maps so long as they conform to the `TextureId` property. Either way is fine with me though, I can enumerate the new map types explicitly.

@Dominykas-Rapalis can you provide non-exact data on which maps, with what properties, were present in the transformation source that lost data?

For myself, I checked the house model sample.
I can only find a [public DGN file](https://github.com/iTwin/Course_1-Smart-House-Sample/raw/main/House_Model.dgn). You can reach out privately to me if you want the bim, I can't find a public link for it but I have a private one.

I queried out the list of unique `RenderMaterial.JsonProperties.materialAssets.renderMaterial.Map` keys and they are:

```
  'Bump',
  'AnisotropicDirection',
  'Geometry',
  'GlowColor',
  'Pattern',
  'TranslucencyColor',
  'Diffuse',
  'TransparentColor',
  'Displacement',
  'Finish',
  'Reflect',
  'Specular'
```

The list of unique properties on those maps is:
<details><pre>
  'BumpFlags',
  'TextureId',
  'antialias_strength',
  'bump_map_scale',
  'enable_antialiasing',
  'high_value',
  'image_gamma',
  'layer',
  'low_value',
  'map_link',
  'offset_z',
  'pattern_angle',
  'pattern_cylindrical_projection_capped',
  'pattern_flip',
  'pattern_lock_projection_scale',
  'pattern_lock_size',
  'pattern_mapping',
  'pattern_off',
  'pattern_offset',
  'pattern_opacity',
  'pattern_proj_angles',
  'pattern_proj_offset',
  'pattern_proj_scale',
  'pattern_scale',
  'pattern_scalemode',
  'pattern_tile_decal_u',
  'pattern_tile_decal_v',
  'pattern_tile_mirror_u',
  'pattern_tile_mirror_v',
  'pattern_u_flip',
  'scale_z',
  'snappable',
  'texture_filter_type',
  'use_cell_colors',
  'AnisotropicDirectionFlags',
  'Filename',
  'PatternFlags',
  'pattern_weight',
  'DiffuseFlags',
  'DisplacementFlags',
  'LxoProcedure',
  'FinishFlags',
  'ReflectFlags',
  'SpecularFlags'
</pre></details>

pmconne (approved):

PR #6359, opened 2024-01-11
Include model ranges in batched tilesets

pmconne:
In reaction to https://github.com/iTwin/imodel-native-internal/pull/351.
Prerequisite for #6354.

The BENTLEY_BatchedTileset extension is now required, and must contain the parallel arrays of model Ids and model extents. Tilesets published prior to this change to tileset.json will fail to load. I do not want to have to maintain multiple alternate code paths to support pre-release versions of the tileset schema.

### DIFF
--- a/common/changes/@itwin/frontend-tiles/pmc-include-model-ranges_2024-01-11-18-41.json
+++ b/common/changes/@itwin/frontend-tiles/pmc-include-model-ranges_2024-01-11-18-41.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/frontend-tiles",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/frontend-tiles"
+}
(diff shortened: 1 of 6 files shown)


danieliborra:
Paul, is this a breaking change for the viewer and the new tiles?
If so, how do we synchronise the viewer and the exporter version?

pmconne:
> Paul, is this a breaking change for the viewer and the new tiles?

This is covered in the PR descriptions.

https://github.com/iTwin/imodel-native-internal/pull/351
> applications using older versions of @itwin/frontend-tiles can continue to decode and use tilesets published after this change.

In other words, viewer using older version of @itwin/frontend-tiles will have no problem displaying tiles produced by new version of TilesetPublisher.

https://github.com/iTwin/itwinjs-core/pull/6359
> The BENTLEY_BatchedTileset extension is now required, and must contain the parallel arrays of model Ids and model extents. Tilesets published prior to this change to tileset.json will fail to load. 

In other words, once you upgrade to this new version of @itwin/frontend-tiles, tilesets produced by older versions of TilesetPublisher lacking these two arrays will stop working.

danieliborra:
Thank you for the clarification!
Perfect, I think we can manage this transition.

PR #6360, opened 2024-01-11
Add core-common to the list of packages for electron-auth and browser-auth to be marked as workspace:*

nick4598:
After running rush update --full in the release branch a couple of days ago and a failing docs build: https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2612229&view=logs&j=bf01bec3-f6e2-5983-bdff-93019feebaa1&t=8d110379-66a4-5f89-b46c-1a9b11ce4c18

I decided to debug into typedoc/typescript code and poke around. In a broken docs build (after running rush update --full), we were missing some exports from RpcManager. This was due to RpcManager now resolving to a path like `"D:/itwinjs-core2/common/temp/node_modules/.pnpm/@itwin+corecommon@4.3.3_67wltvhdskk2oee2c3z2o4tfly/node_modules/@itwin/core-common/lib/cjs/RpcManager.d.ts"` instead of `"D:/itwinjs-core/core/common/lib/cjs/RpcManager.d.ts"`. 

When running typedoc we intentionally exclude paths with node_modules (by passing an exclude regex) in the name which makes sense. Even when hovering over imports in vscode "import from @itwin/core-common" these incorrect paths would be shown which suggested that typescript module resolution was giving these paths to typedoc. I later confirmed that while stepping through typedoc and ending up in typescript code, but I didn't quite know why TS seemed to be getting confused.

The reason rush update --full caused this is due to the below
image
Because browser-auth was pulling in a version of @itwin/core-common... Since locally our package.json had 4.3.3 in it as well as in the pnpm-lock after rush update --full, this was confusing TS and it was no longer looking at the local copy of core-common but the installed version of core-common which was also 4.3.3. This lines up with why master works just fine, locally we had a dev version and installed we had some other officially released version. That also supports why release/4.3.x worked just fine before we had run rush update --full. Locally we had 4.3.3 and installed we had a 4.2.1 version.

Thanks Bill for noticing that we had an installed copy of core-common in the common/temp/.pnpm/node_modules and thanks Ben for general help with this.

**Open Questions**
1) Its confusing to me that I only had to change browser-auth in pnpmfile.cjs to get this to work again, when electron-auth should've been introducing the same problem (Note I did change both in the PR anyways).
1) Should this have caused more problems with docs than it did? Or does core-electron just have the perfect conditions to expose the problem. It feels like all that needs to happen is an import from core/common for it to have broken elsewhere.. but I don't know.

### DIFF
--- a/common/config/rush/.pnpmfile.cjs
+++ b/common/config/rush/.pnpmfile.cjs
@@ -28,6 +28,9 @@ function readPackage(pkg) {
     ].includes(pkg.name)
   ) {
     pkg.dependencies["@itwin/core-bentley"] = "workspace:*";
+    if (pkg.name === "@itwin/browser-authorization" || pkg.name === "@itwin/electron-authorization") {
+      pkg.dependencies["@itwin/core-common"] = "workspace:*";
+    }
   }
 
   // https://github.com/iTwin/reality-data-client
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -1731,7 +1731,7 @@ importers:
       webpack-cli: ^5.0.1
     dependencies:
       '@azure/storage-blob': 12.7.0
-      '@itwin/browser-authorization': 1.0.1_67wltvhdskk2oee2c3z2o4tfly
+      '@itwin/browser-authorization': 1.0.1_mdtbcqczpmeuv6yjzfaigjndwi
       '@itwin/core-backend': link:../../core/backend
       '@itwin/core-bentley': link:../../core/bentley
       '@itwin/core-common': link:../../core/common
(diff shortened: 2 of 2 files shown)


nick4598:
@mergifyio backport release/4.3.x

common/config/rush/.pnpmfile.cjs
```diff
@@ -28,6 +28,9 @@ function readPackage(pkg) {
     ].includes(pkg.name)
   ) {
     pkg.dependencies["@itwin/core-bentley"] = "workspace:*";
+    if (pkg.name === "@itwin/browser-authorization" || pkg.name === "@itwin/electron-authorization") {
+      pkg.dependencies["@itwin/core-common"] = "workspace:*";
```
aruniverse, line 32: @ben-polinsky makes me wonder if our deps in these pkgs are right

ben-polinsky: I was thinking the same. The justification for the direct dependency was that we're using types from core-common in our public api. 

Practically, would anyone auth packages without core-common? Perhaps some, but it's only a slight inconvenience to install.

aruniverse: very unlikely, might as well roll your own auth. resolved comment to prevent blocking any future rush update fulls

PR #6361, opened 2024-01-11
Loosen tolerance of Matrix3d.createRigidFromMatrix3d to allow near-rigid matrices

saeeedtorabi:
Now if a near-rigid matrix is passed to Matrix3d.createRigidFromMatrix3d, the function does not return undefined.


dassaf4 (approved):
Looks good. I clarified doc and test, and added test matrices that get progressively less rigid.

PR #6365, opened 2024-01-12
custom attribute error handle when creating schema fromJson

julioalvahash:
This PR adds error handle when attempting to create a schema with referenced custom attributes and no reference defined using fromJson.


core/ecschema-metadata/src/Deserialization/Helper.ts
```diff
@@ -1010,11 +1010,18 @@ export class SchemaReadHelper<T = unknown> {
   private async loadCustomAttributes(container: AnyCAContainer, caProviders: Iterable<CAProviderTuple>): Promise<void> {
     for (const providerTuple of caProviders) {
       // First tuple entry is the CA class name.
-      const caClass = await this.findSchemaItem(providerTuple[0]) as CustomAttributeClass;
+      const caClass = await this.findSchemaItem(providerTuple[0]);
+
+      // If custom attribute exist within the context and is referenced, validate the reference is defined in the container's schema
+      if (caClass && caClass.key.schemaName !== container.schema.name &&
+        !container.schema.references.find((reference) => reference.name === caClass.key.schemaName)) {
```
ColinKerr, line 1017: you should use Schema.getReferenceSync here instead of using find directly.  if you look at the implementation you'll notice that it is doing a lower case comparison while you are comparing by case

core/ecschema-metadata/src/test/Metadata/SchemaFromJson.test.ts
```diff
@@ -0,0 +1,334 @@
...
+          },
+          {
+            className: "DummyReferenceTwo.customAttributeOneRefTwo",
+            showClasses: true,
+          },
+        ],
+      };
+
+      await expect(Schema.fromJson(schemaA, context)).to.be.rejectedWith(ECObjectsError, "Unable to load custom attribute DummyReferenceTwo.customAttributeOneRefTwo from container SchemaA, DummyReferenceTwo reference not defined");
+    });
```
ColinKerr, line 146: You don't need both of these tests if you are using the getReferenceSync method because that method handles accessing references directly and should already have tests to ensure it works with no references

core/ecschema-metadata/src/test/Metadata/SchemaFromJson.test.ts
```diff
@@ -0,0 +1,334 @@
...
+              },
+            ],
+          },
+        },
+      };
+
+      await expect(Schema.fromJson(schemaA, context)).to.be.rejectedWith(ECObjectsError, "Unable to load custom attribute DummyReference.customAttributeOne from container testClassOne.Offset, DummyReference reference not defined");
+    });
+
+    it("should throw an error and not allow the creation of the schema with custom attribute in an array property container", async () => {
```
ColinKerr, line 283: I don't think you need this test.  There is no meaningful difference between types of property definitions when it comes to custom attribute behavior.

core/ecschema-metadata/src/Deserialization/Helper.ts
julioalvahash: Changed to use getReferenceSync.

core/ecschema-metadata/src/test/Metadata/SchemaFromJson.test.ts
julioalvahash: Removed.

core/ecschema-metadata/src/test/Metadata/SchemaFromJson.test.ts
julioalvahash: Removed.

ColinKerr (approved):

PR #6369, opened 2024-01-16
Change config handling for display test app

hl662:
- Change default client id for electron renderer auth to a more appropriately named client id (electron client ids should follow `native-XX`)
- Fixes `dta signin` callback hanging after successful log in on the electron DTA, when `npm run start` script is used. Specifically, a different redirect url has to be used in this case, because the default, localhost:3000 is already used by the webserver. Added an if statement in DtaConfiguration to allow redirect urls to branch depending on the process.


hl662:
@DStradley can you run the image tests before I turn this PR active? Want to check for regressions

aruniverse (commented):
im assuming display-performance-test-app needs to be updated too

aruniverse (approved):

hl662:
> im assuming display-performance-test-app needs to be updated too

No change needed, dpta sign in works as intended

markschlosseratbentley (approved):

PR #6373, opened 2024-01-17
Fail on data conflicts instead of suppressing and pushing a potential bad changeset to hub

khanaffan:
For detail read NextVersion.md

imodel-native:  https://github.com/iTwin/imodel-native/pull/544


docs/changehistory/NextVersion.md
```diff
@@ -2,3 +2,23 @@
...
+
+`BriefcaseDb.pushChanges()` will fail with following error, if their was data conflict detected.
+
+> _`UPDATE/DELETE before value do not match with one in db or CASCADE action was triggered`_
+
+### What is data conflict?
+
+Data conflict come in two flavours,
+
+1. Dirty reads happen when current briefcase update value of a property `P` from `X` to `Y`. But incoming changeset has property `P` with value `K` instead of `X`.
```
pmconne, line 18: I don't understand this sentence, could you rephrase it please?

khanaffan: I will fix it. But it means the old values do not match for the changed row. The change is of type update, so it has old and new values. When the current db does not have the old values for a row then it causes a data conflict. 

Let's say the user reads x=1 and changes it to 2.  But when he does pull/merge another user has already changed x from 1 to 9. When applying the change for the current briefcase x=2 does not match the old value of x=1 and so it fails.

core/backend/src/test/changesets/ChangeMerging.test.ts
```diff
@@ -1,142 +0,0 @@
-/*---------------------------------------------------------------------------------------------
-* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
-* See LICENSE.md in the project root for license terms and full copyright notice.
-*--------------------------------------------------------------------------------------------*/
-
-import { assert } from "chai";
-import * as path from "path";
-import { Id64String, OpenMode } from "@itwin/core-bentley";
```
khanaffan, line 8: @pmconne after consulting with @swwilson-bsi  we found ChangeMerging is a very old test used to test no-lock merge which was something on the table very early in itwin.js development. My new test and this test do the same thing both use no-lock to change elements and expect them to either merge or fail. 

This test failed with the expected error but we decided to remove it as its redundant to what I have.

docs/changehistory/NextVersion.md
pmconne: Am I misunderstanding or could this simply be stated as "two briefcases attempt to modify the same property"?

khanaffan: yes that is correct.

pmconne: I have edited NextVersion.md for clarity. Please resolve this comment if you're ok with those changes.

pmconne (approved):

PR #6376, opened 2024-01-19
Enable Kerberos authentication for map-layers

mdastous-bentley:
Some of our users have on-promise ArcGIS Enterprise servers protected by Kerberos (i.e Windows Authentication), and want to consume their GIS data as map-layers. 

Kerberos authentication is supported natively by all modern web browsers, however in CORS scenarios, you have to manually turn ON a switch to when using the Fetch API (i.e. `credentials: "include"`), to inform the browser to pass credentials of the active user.

Additionally, in order to activate this browser security mechanism, the server response  **MUST** include the following headers:
- The `Access-Control-Allow-Credentials` header must be set to `true`
- The `Access-Control-Allow-Origin` header must set to the **exact** origin value of the actual requesting web page (i.e. `*` value will be rejected)
- The `Access-Control-Expose-Headers` header must include the `Www-Authenticate` value (This is a custom itwin.js requirement; we want to include user credentials only for very specific authentication methods.  Without this header, the browser prevent us from reading the `Www-Authenticate` header from the response headers.)


core/frontend/src/request/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Check whether or not one of the requested authentication method is listed in the HTTP 'WWW-Authenticate' response header
+ * @param headers Headers object
+ * @param query List of authentication method to lookup
+ * @note For CORS requests, the 'Access-Control-Expose-Headers' header from the server must make the 'WWW-Authenticate' available to the browser, otherwise this won't work.
+ * @internal
+ */
+export function headersIncludeAuthMethod(headers: Headers, query: string[]): boolean {
+  const headersArray = Array.from(headers);
+  const foundAuth = headersArray.filter((pair)=> pair[0].toLowerCase() === "www-authenticate");
```
aruniverse, line 18: why not just use [Headers.has](https://developer.mozilla.org/en-US/docs/Web/API/Headers/has) and [Headers.get](https://developer.mozilla.org/en-US/docs/Web/API/Headers/get)

core/frontend/src/tile/map/WmsUtilities.ts
```diff
@@ -13,4 +16,28 @@ export class WmsUtilities {
...
+
+    let response = await fetch(url, { method: "GET", headers });
+    if (!credentials && response.status === 401 && headersIncludeAuthMethod(response.headers, ["ntlm", "negotiate"])) {
+    // We got a http 401 challenge, lets try SSO (i.e. Windows Authentication)
+      response = await fetch(url, { method: "GET", credentials: "include" });
+    }
+
+    if (response.status !== 200)
+      throw new HttpResponseError(response.status, await response.text());
+    return response.text();
```
aruniverse, line 41: is this supposed to be awaited?

core/frontend/src/request/utils.ts
mdastous-bentley: For some reasons I assumed has/get were case sensitive and I wanted to avoid any case related issues .  
After testing, looks like those methods are already [case-insensitive](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAFgUwIYBMECcIwLwzAgdxgAlk1MAKASgG4AoRVDCAOggSgoHIB1PgWgCCAVyiIwUAJbAkUBFwA0MLgDkAKgBkAsl1p1QkEABsELIyADmFAAbEkWXgJFiEE6bPkAuGABIA3ozkrHD2NgTh-Eii4lIyctZUAL4JNEA), I will change the code.

core/frontend/src/tile/map/WmsUtilities.ts
mdastous-bentley: `WmsUtilities.fetchXml` already returns a Promise<string>, so does it matter to await inside fetchXml if anyway the caller of `WmsUtilities.fetchXml`  is expected to await?

core/frontend/src/tile/map/MapLayerImageryProvider.ts
```diff
@@ -45,6 +46,15 @@ export abstract class MapLayerImageryProvider {
   /** @internal */
   private _status = MapLayerImageryProviderStatus.Valid;
 
+  /** @internal */
+  protected _includeUserCredentials = false;
+
+  /** @internal */
+  public readonly onFirstRequestCompleted = new BeEvent<() => void>();
```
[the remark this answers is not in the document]
mdastous-bentley: totally agree.

core/frontend/src/request/utils.ts
mdastous-bentley: should be better now.

aruniverse (approved):

PR #6377, opened 2024-01-19
split up ecsql reference into indiviual files, remove iframed iModel …

ColinKerr:
…Console from tutorial and remove tutorial pages are just reference docs


docs/bis/ec/ec-custom-attributes.md
```diff
@@ -43,4 +43,17 @@ Custom attributes allow any custom metadata to be applied because they are defin
     </ECCustomAttributeClass>
 ```
 
+It is also possible to query custom attributes:
```
ColinKerr, line 46: this doesn't belong here ... this is just EC reference we don't have anything talking about ECDb.  Maybe we need a meta query section inn the ECSql Reference.  In there we need some examples that include getting the values of custom attributes

docs/bis/ec/ec-views.md
```diff
@@ -0,0 +1,78 @@
+# ECViews
+
+A view is a special kind of [ECEntityClass](ec-entity-class.md). The concept is realized through an [ECCustomAttribute](ec-custom-attributes.md) which is defined in the `ECDbMap` schema.
+
+The custom attribute allows defining an abstract entity class which is not backed by actual data but rather by an ECSQL query which is executed to obtain instances of the class.
```
ColinKerr, line 5: This section should start with something that better describes what ECViews are.  For example if we take part o the third sentence that might make a good start.  'Allows defining an abstract entity class which is not backed by actual data but rather by an ECSql query which is executed to obtain instances of the class.'  Maybe start with the first sentence put some of the third second and then explain that it is implemented via a mapping custom attribute.

docs/bis/ec/ec-views.md
```diff
@@ -0,0 +1,78 @@
...
+
+```
+Length   |Diameter |Material
+----------------------------------------
+100      |10       |steel
+200      |15       |copper
+150      |20       |plastic
+```
+
+A view may contain navigation properties, for these, the query can use the `navigation_value`function provided by ECSQL [ECSqlFunctions](./../../learning/ECSqlReference/ECSqlFunctions.md)
```
ColinKerr, line 78: it would be good to have another example that shows navigation properties

docs/bis/ec/ec-custom-attributes.md
rschili: done

docs/bis/ec/ec-views.md
rschili: I added a new example, check ECSqlReference/View.md

docs/bis/ec/ec-views.md
rschili: Rewrote the section, check ECSqlReference/View.md

grigasp (approved):
approving the presentation docs change

rschili (approved):

docs/changehistory/4.3.0.md
```diff
@@ -31,11 +31,11 @@ The new [ClipUtilities.clipAnyCurve]($core-geometry) clips any `CurvePrimitive`,
 
 ECSQL supports querying instance properties, which are any property in a class selected in ECSql or its derived classes.
 
-[**ECSQL Instance Properties Documentation**](../learning/ECSQLTutorial/ECSqlSyntax.md#instance-query)
+[**ECSQL Instance Properties Documentation**](../learning/ECSqlReference/InstanceQuery.md)
 
 ### Window functions
 
-ECSQL now supports [window functions](../learning/ECSQLTutorial/ECSqlSyntax.md#window-functions). **This functionality add following new keywords that might collide with alias, parameter, property or class name and may break existing queries.** Application need to [escape](../learning/ECSQLTutorial/ECSqlSyntax.md#escaping-keywords) those name or try escaping all names in their queries to be on safe side in future.
+ECSQL now supports [window functions](../learning/ECSqlReference/WindowFunctions.md). **This functionality add following new keywords that might collide with alias, parameter, property or class name and may break existing queries.** Application need to [escape](../learning/ECSqlReference/ECSqlKeywords.md#escaping-keywords) those name or try escaping all names in their queries to be on safe side in future.
```
ben-polinsky, line 38:
```suggestion
ECSQL now supports [window functions](../learning/ECSqlReference/WindowFunctions.md). **This functionality adds the following new keywords that might collide with alias, parameter, property or class names and may break existing queries.** Applications should [escape](../learning/ECSqlReference/ECSqlKeywords.md#escaping-keywords) those names (or try escaping all names in their queries) to be on the safe side in the future.
```

docs/learning/ECSQLTutorial/ClassFilter.md
```diff
@@ -46,8 +44,6 @@ ECSQL allows filters by class
 > SELECT * FROM bis.Element WHERE ECClassId IS NOT (ONLY Generic.PhysicalObject, ONLY BisCore.LightLocation)
 > ```
 
-<iframe class="embedded-console" src="https://imodelconsole.bentley.com/?embedded=true&nosignin=true&imodel=HouseSample&query=SELECT * FROM bis.Element WHERE ECClassId IS NOT (ONLY Generic.PhysicalObject, ONLY Biscore.LightLocation)"></iframe>
-
 ---
 
-[**< Previous**](./ChangeSummaryQueries.md) &nbsp; | &nbsp; [**Next >**](./ConditionalExpr.md)
+[**< Previous**](./ChangeSummaryQueries.md) &nbsp;
```
ben-polinsky, line 49: Not a huge fan of the bold links:
image

Looks fine without and using unicon chevrons `‹` and `›` I think
image

docs/learning/ECSQLTutorial/index.md
```diff
@@ -25,6 +25,9 @@ The ECSQL examples in the tutorial work with the "House Sample" iModel. You can
 
 The schemas for iModels are based on [BIS](../../bis/index.md). Therefore, the examples throughout the tutorial use the BIS schemas. While not required to learn ECSQL, familiarity with BIS is a good idea to get more from this tutorial.
 
+For a video overview of BIS checkout this Youtube videos:
```
ben-polinsky, line 28:
```suggestion
For a video overview of BIS:
```

docs/learning/ECSQLTutorial/index.md
```diff
@@ -74,9 +63,16 @@ Or simply use the provided sample below:
...
 - [Lesson 9: Class Filter](./ClassFilter.md)
-- [Lesson 10: Conditional Expressions](./ConditionalExpr.md)
-- [Lesson 11: Built-In functions](./BuiltInFunctions.md)
-- [Lesson 12: Instance Props](./InstanceProps.md)
+
+## Further Learning and Resources
+
+- [ECSql Overview](../ECSQL.md)
+- [ECSql Reference Documentation](../ECSqlReference/index.md)
+- Youtube videos about ECSql
```
ben-polinsky, line 71:
```suggestion
- YouTube videos about ECSql
```

docs/learning/ECSqlReference/CASE.md
```diff
@@ -0,0 +1,50 @@
+# CASE-WHEN-THEN-ELSE
+
+ECSQL supports only searched CASE expressions:
```
ben-polinsky, line 3: Not sure I understand. Might need re-wording?

docs/learning/ECSqlReference/CASE.md
```diff
@@ -0,0 +1,50 @@
...
+CASE
+      WHEN <expr> THEN <expr>
+    [ WHEN <expr> THEN <expr> ...]
+    [ ELSE <expr> ]
+END
+```
+
+## Limitations
+
+Only primitive type can be used with WHEN, THEN and ELSE. Primitive does not include p2d, p3d, IGeometery and NavigationProperties. You can still use sub-queries that return single column and pretty much any SQL expressions.
```
ben-polinsky, line 15:
```suggestion
Only primitive types can be used with WHEN, THEN and ELSE. This does not include p2d, p3d, IGeometery and NavigationProperties. You can still use sub-queries that return single column and pretty much any SQL expressions.
```

docs/learning/ECSqlReference/CompoundSELECT.md
```diff
@@ -0,0 +1,62 @@
+# Compound SELECT
+
+Result of `SELECT` statement can be combined with other select statements using one of following operator.
+
+1. `UNION` - take a union of result of two queries such that there is no duplicate results.
+1. `UNION ALL` - take a union of results of two queries.
+1. `INTERSECT` - take only rows that are common in both queries.
+1. `EXCEPT` - take rows from first queries that are not present in second query.
```
ben-polinsky, line 8:
```suggestion
1. `EXCEPT` - take rows from first query that are not present in second query.
```

docs/learning/ECSqlReference/DateAndTime.md
```diff
@@ -0,0 +1,72 @@
...
+## `TIMESTAMP`
+
+Describe a literal timestamp value. Timestamp is provided according to
+[iso-8601](https://en.wikipedia.org/wiki/ISO_8601) standard.
+
+```bnf
+TIMESTAMP <iso-8601-timestamp>
+```
+
+Following example use [strftime()](https://www.sqlite.org/lang_datefunc.html) to extract date/time component from a timestamp value.
```
ben-polinsky, line 12:
```suggestion
Following example uses [strftime()](https://www.sqlite.org/lang_datefunc.html) to extract date/time component from a timestamp value.
```

docs/learning/ECSqlReference/DateAndTime.md
```diff
@@ -0,0 +1,72 @@
...
+
+```sql
+SELECT strftime('%m/%d/%Y %H:%M:%S', TIMESTAMP '2013-02-09T12:01:22') AS [output]
+```
+
+Output look like following
```
ben-polinsky, line 18:
```suggestion
Output:
```

ben-polinsky: There's more than this one, btw

docs/learning/ECSqlReference/ECSqlFunctions.md
```diff
@@ -0,0 +1,188 @@
...
+1. [`navigation_value()`](#navigation_value-ecnavigationproperty-path-id--RelECClassId) - Constructs an ECNavigation value, given ECNavigation property, Id and optionaly, RelECClassId.
+
+## ec_classname( _ecclassId_ [, _format-string_ | _format-id_] )
+
+For the specified ecClassId, returns the class name as a string formatted according to the specified format-string
+
+### Parameters
+
+`ecclassId`: An integer which could be a constant, column or a parameter.
+`format-string | format-id`: Optional format specifier and could be one of the following values. `NULL` is also valid value -- this is the same as not specifying the second parameter at all
```
ben-polinsky, line 22: `this is the same as not specifying the second parameter at all`

As if the second parameter were not specified? Not sure what the intent is...

docs/learning/ECSqlReference/ECSqlFunctions.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Parameters
+
+Can take either one or two parameters:
+`schema-name-or-alias`: Schema name or alias e.g. bis (alias) or BisCore (name)
+`class-name`: Name of the class e.g. Element
+
+### Returns
+
+The function return a integer `ECClassId` or `NULL` if the name could not be resolved.
```
ben-polinsky, line 74:
```suggestion
An integer `ECClassId` or `NULL` if the name could not be resolved.
```

docs/learning/ECSqlReference/ECSqlFunctions.md
```diff
@@ -0,0 +1,188 @@
...
+Island,Terrain
+Reference,Terrain
+Source,Terrain
+Spot,Terrain
+Void,Terrain
+```
+
+## StrToGuid( _guid-string_ )
+
+When `GUID` is stored a binary, it need to be converted for comparison purpose.
```
ben-polinsky, line 144:
```suggestion
When `GUID` is stored as a binary, it needs to be converted for comparison purposes.
```

docs/learning/ECSqlReference/ECSqlFunctions.md
```diff
@@ -0,0 +1,188 @@
...
+
+```sql
+SELECT * FROM [BisCore].[Element] WHERE FederationGuid = StrToGuid('407bfa18-944d-11ee-b9d1-0242ac120002')
+```
+
+## GuidToString( _binary-guid_ )
+
+When `GUID` is stored a binary, it need to be converted for comparison purpose.
```
ben-polinsky, line 152:
```suggestion
When `GUID` is stored as a binary, it needs to be converted for comparison purpose.
```

docs/learning/ECSqlReference/ECSqlKeywords.md
```diff
@@ -0,0 +1,35 @@
...
+| S   | `SELECT`, `SET`, `SOME`, `STRING`, `SUM`                                                                                                |
+| T   | `THEN`, `TIES`, `TIME`, `TIMESTAMP`, `TOTAL`, `TRUE`                                                                                    |
+| U   | `UNBOUNDED`, `UNION`, `UNIQUE`, `UNKNOWN`, `UPDATE`, `USING`                                                                            |
+| V   | `VALUE`, `VALUES` ,`VARCHAR`                                                                                                            |
+| W   | `WHEN`, `WHERE`, `WINDOW`, `WITH`                                                                                                       |
+
+## Escaping keywords
+
+ECSQL has large set of [keywords](#ecsql-keywords). Keywords sometime appear in query as class name, property name or parameter name, cte block name or aliases and will result in query to fail. To fix it the keyword need to be quoted or escaped. Following is different ways keywords can be escaped. In ECSQL is preferred to escape using square brackets e.g. `[keyword]`.
```
ben-polinsky, line 27:
```suggestion
ECSQL has a large set of [keywords](#ecsql-keywords). Keywords sometime appear in queries as class names, property names or parameter names, cte block names or aliases and will cause the query to fail. To fix it, the keyword needs to be quoted or escaped. The following are different ways keywords can be escaped. In ECSQL it is preferred to escape using square brackets e.g. `[keyword]`.
```

docs/learning/ECSqlReference/ECSqlOptions.md
```diff
@@ -0,0 +1,42 @@
+# ECSQLOPTIONS or OPTIONS clause
+
+`ECSQLOPTIONS` which can also be written as just `OPTIONS` use to specify flags thats will effect processing of ECSQL statement.
```
ben-polinsky, line 3:
```suggestion
`ECSQLOPTIONS` (or `OPTIONS` for short) are used to specify flags thats will effect processing of ECSQL statements.
```

docs/learning/ECSqlReference/ECSqlParameters.md
```diff
@@ -0,0 +1,25 @@
+# ECSQL Parameters
+
+ECSQL support named and positional parameters.
```
ben-polinsky, line 3:
```suggestion
ECSQL supports named and positional parameters.
```

docs/learning/ECSqlReference/ECSqlParameters.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Named parameters
+
+Name parameter can be use to bind parameter by name.
```
ben-polinsky, line 7:
```suggestion
Named parameters can be used to bind a parameter by name.
```

docs/learning/ECSqlReference/ECSqlParameters.md
```diff
@@ -0,0 +1,25 @@
...
+
+Syntax: `:<parameter-name>`
+
+```sql
+SELECT * FROM [meta].[ECClassDef] WHERE [Name] = :className
+```
+
+## Positional parameters
+
+Positional parameter are bind by position from left to right.
```
ben-polinsky, line 17:
```suggestion
Positional parameters are bound by position from left to right.
```

docs/learning/ECSqlReference/IIF.md
```diff
@@ -0,0 +1,23 @@
+# IIF (_condition-expr_, _true-expr_, _false-expr_)
+
+ECSQL supports IIF(), which is really shorthand for `CASE WHEN <condition-expr> THEN <true-expr> ELSE <false-expr> END`
```
ben-polinsky, line 3:
```suggestion
ECSQL supports IIF(), which is shorthand for `CASE WHEN <condition-expr> THEN <true-expr> ELSE <false-expr> END`
```

docs/learning/ECSqlReference/IIF.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Parameters
+
+`condition-expr`: A condition expression that resolve into a boolean value. e.g. Length > 1.0.
```
ben-polinsky, line 7:
```suggestion
`condition-expr`: A condition expression that resolves into a boolean value. e.g. Length > 1.0.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
+# Instance query
+
+Instance queries allow you to select properties defined in derived classes when selecting a base class or multiple properties with the same name defined in different derived classes.
+
+## What is instance property?
```
ben-polinsky, line 5:
```suggestion
## What is an instance property?
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+Instance property is any property in a class selected in ECSql or its derived classes accessed via the instance query syntax.
```
ben-polinsky, line 7:
```suggestion
An instance property is any property in a class selected in ECSql or its derived classes accessed via the instance query syntax.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+## How to access instance property?
+
+In ECSQL instance property can be accessed by using the `$->` operator.
```
ben-polinsky, line 11:
```suggestion
In ECSQL instance properties can be accessed by using the `$->` operator.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+```sql
+SELECT $->[CodeValue] FROM [BisCore].[Element] WHERE $->[CodeValue] IS NOT NULL LIMIT 1;
+--
+SELECT e.$->[CodeValue] FROM [BisCore].[Element] e LIMIT 1;
+```
+
+## How it works?
+
+Instance property allows relaxed access to any property within a hierarchy or selected class. It allows full access to the underlying instance of a class using its base class. We can think of it as if `$` represent the full instance not just properties of the selected class.
```
ben-polinsky, line 21:
```suggestion
An instance property allows relaxed access to any property within a hierarchy or selected class. It allows full access to the underlying instance of a class using its base class. We can think of it as if `$` represents the full instance not just properties of the selected class.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+Following ECSQL will return only properties declared in `BisCore.Element`
```
ben-polinsky, line 23:
```suggestion
The following ECSQL will return only properties declared in `BisCore.Element`
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+```sql
+    SELECT * FROM [BisCore].[Element] WHERE ECInstanceId = 0xc000000014c
+```
+
+| ECInstanceId    | ECClassId | Model                                  | Last Modified              | Code Specification           | Code Scope                   | Code   | User Label | Parent | Federation GUID | JSON Properties |
+| --------------- | --------- | -------------------------------------- | -------------------------- | ---------------------------- | ---------------------------- | ------ | ---------- | ------ | --------------- | --------------- |
+| `0x8000000014c` | `0x710`   | `{Id:0x80000000003,RelECClassId:0x51}` | `2020-09-13T21:03:39.281Z` | `{Id:0x1,RelECClassId:0x59}` | `{Id:0x1,RelECClassId:0x5b}` | `NULL` | `Computer` | `NULL` | `NULL`          | `NULL`          |
+
+While following return all properties of respective derived class of `BisCore.Element`
```
ben-polinsky, line 33:
```suggestion
While the following will return all properties of respective derived class of `BisCore.Element`
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+```sql
+    SELECT $ FROM [BisCore].[Element] WHERE ECInstanceId = 0xc000000014c
+```
+
+above return one column and it contain serialized json instance with all properties
```
ben-polinsky, line 39:
```suggestion
the above returns one column containing a serialized json instance with all properties
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+```sql
+    SELECT $->[RevitId], $->[LastModifier]  FROM [BisCore].[Element] WHERE $->[Asset_Tag] ='COMPUTER 005'
+```
+
+| $ -> RevitId | $ -> LastModifier |
+| ------------ | ----------------- |
+| 381840       | kiran.patkar      |
+
+ECSql will apply a property filter on selected rows such that those instances which has at least one property out of set of instance property must exists. This improve performance.
```
ben-polinsky, line 118:
```suggestion
ECSql will apply a property filter on selected rows such that those instances which have at least one property out of set of instance properties must exist. This improves performance.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+```sql
+    SELECT $->[ThisPropertyDoesNotExists] from [BisCore].[Element];
+```
+
+If `ThisPropertyDoesNotExists` does not exists in `Bis.Element` derived hierarchy then no row will be returned. ECSql filter only include rows that must have at least one instance property. If any instance does not have any instance property requested then it will will be skipped.
```
ben-polinsky, line 124:
```suggestion
If `ThisPropertyDoesNotExists` does not exists in the `Bis.Element` derived hierarchy then no row will be returned. ECSql filter only includes rows that must have at least one instance property. If any instance does not have any instance properties requested, then it will will be skipped.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Accessing composite properties like `NavigationProperty`, `Point2d`, `Point3d` or `Struct`s
+
+Only top level instance property can be accessed via `$-><prop>` syntax. Doing something like `$->Model.Id` will not not work as of now. It might be supported in future but as of now any access-string within a composite property is not supported, if its the only property selected then zero row will be returned.
```
ben-polinsky, line 128:
```suggestion
Only top level instance properties can be accessed via `$-><prop>` syntax. Doing something like `$->Model.Id` will not work. It might be supported in future but as of now any access-string within a composite property is not supported: if it is the only property selected then zero row will be returned.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+Following type of properties can directly be use in filters and return strong type value.
```
ben-polinsky, line 130:
```suggestion
The following type of properties can be used directly in filters to return strong type values.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+- DateTime
+- Integer
+- Long
+- Binary
+- String
+- Double
+
+Here is example of `RevitId` use with `IN()` clause.
```
ben-polinsky, line 139:
```suggestion
Here is example of `RevitId` used with `IN()` clause.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+| ModelId         |
+| --------------- |
+| `0x80000000003` |
+
+## Optional and non-optional instance properties
+
+By default, all properties accessed via instance accessor i.e. `$->prop` must exist in the class identifying the row for that row to qualify for output.
+
+If the user uses `?` after a property accessor e.g. `$->prop?` then it will be considered optional, and the row class will not be checked to see if the `prop` exists or not.
```
ben-polinsky, line 179:
```suggestion
If the user uses `?` after a property accessor e.g. `$->prop?` then it will be considered optional, and the row class will not be checked to see if the `prop` exists.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+On the other hand, the following query makes `Foo` optional by adding `?` at the end like `$->Foo?`. This will exclude this property from the list of instance properties that must exist in the class of a row for it to qualify for output.
+
+```sql
+  SELECT ECClassId, ECInstanceId
+  FROM [BisCore].[Element]
+      WHERE $->CodeValue = 'Profiling' OR $->Foo? = 'Hello'
+  LIMIT 1
+```
+
+> Note: Optional property may slow down performance while non-optional properties will improve the performance of instance query.
```
ben-polinsky, line 199:
```suggestion
> Note: Optional properties may slow down performance while non-optional properties will improve the performance of an instance query.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Accessing composite properties
+
+Only top level instance properties can be accessed using instance property accessor syntax `$-><prop>`.\
```
ben-polinsky, line 203: Seems like the same information as above in `Accessing composite properties like `NavigationProperty`, `Point2d`, `Point3d` or `Struct`s`

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+ECSql will apply a property filter on selected rows such that those instances which has at least one property out of set of instance property must exists. This improve performance.
+
+```sql
+    SELECT $->[ThisPropertyDoesNotExists] from [BisCore].[Element];
+```
+
+If `ThisPropertyDoesNotExists` does not exists in `Bis.Element` derived hierarchy then no row will be returned. ECSql filter only include rows that must have at least one instance property. If any instance does not have any instance property requested then it will will be skipped.
+
+## Accessing composite properties like `NavigationProperty`, `Point2d`, `Point3d` or `Struct`s
```
ben-polinsky, line 126:
```suggestion
## Accessing composite properties like `NavigationProperty`, `Point2d`, `Point3d` or `Struct`
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+## Limitation
+
+1. Only top level property is allowed.
+2. Only primitive type values can be accessed in the filter directly. Any composite type will require `JSON_EXTRACT()` to extract child value before it can be used in a query. Refer [Accessing Composite Properties](#accessing-composite-properties)
+3. Currently indexes are not supported on instance properties.
+4. MetaData a.k.a `ColumnInfo` is dynamically updated only for primitive properties selected for output. All other properties will get generic `ColumnInfo` with a string property and `extendType=JSON`.
+
+## Performance
+
+Generally speaking the performance of instance prop is pretty good though it involve overhead of extracting either property value or complete instance.
```
ben-polinsky, line 279:
```suggestion
Generally speaking the performance of instance prop is pretty good, though it involves overhead of extracting either property value or complete instance.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+
+- Try use regular properties accessor where possible.
```
ben-polinsky, line 281:
```suggestion
- Use regular properties accessors where possible.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+- Do not use instance property access for local properties of class been selected.
```
ben-polinsky, line 282:
```suggestion
- Do not use instance property access for local properties of class been selected.
```
```suggestion
- Do not use instance property access for local properties of class selected.
```

docs/learning/ECSqlReference/InstanceQuery.md
```diff
@@ -0,0 +1,285 @@
...
+- Try avoiding filtering queries by instance properties. Though it fast be without a index it could be slow depending on number of rows to which filter will be applied.
```
ben-polinsky, line 283: Could use rewording, I'm not exactly sure of the intent.

docs/learning/ECSqlReference/JSON1VirtualClasses.md
```diff
@@ -0,0 +1,62 @@
...
+| orbital_period    | 687 days                                                                                                                                                                                | text   | 687 days      | 0      | $."orbital_period"    | $       |
+| moons             | ["Phobos","Deimos"]                                                                                                                                                                     | array  | NULL          | 0      | $.moons               | $       |
+| 0                 | Phobos                                                                                                                                                                                  | text   | Phobos        | 14     | $.moons[0]            | $.moons |
+| 1                 | Deimos                                                                                                                                                                                  | text   | Deimos        | 14     | $.moons[1]            | $.moons |
+
+## json_each()
+
+Iterate top level json and return each entry as row.
```
ben-polinsky, line 36:
```suggestion
Iterate top level json and return each entry as a row.
```

docs/learning/ECSqlReference/LIKE.md
```diff
@@ -0,0 +1,51 @@
+# LIKE operator
+
+Match value to a pattern.
+
+Syntax: `<expr> [NOT] LIKE <pattern> [ESCAPE '<char>']`
+
+- The percent sign `%` represents zero, one, or multiple characters
+- The underscore sign `_` represents one, single character
```
ben-polinsky, line 8:
```suggestion
- The underscore sign `_` represents one single character
```

docs/learning/ECSqlReference/LIKE.md
```diff
@@ -0,0 +1,51 @@
...
+
+Find classes with name start with `IL`.
```
ben-polinsky, line 10:
```suggestion
Find classes with names starting with `IL`.
```

docs/learning/ECSqlReference/LIMIT.md
```diff
@@ -0,0 +1,15 @@
+# LIMIT clause
+
+Limit the number of rows returned by query. The clause also set offset from which the limit on rows is applied. [Read sqlite docs](https://www.sqlite.org/lang_select.html#limitoffset)
```
ben-polinsky, line 3:
```suggestion
Limit the number of rows returned by query. The clause also sets offset from which the limit on rows is applied. [Read sqlite docs](https://www.sqlite.org/lang_select.html#limitoffset)
```

docs/learning/ECSqlReference/Literals.md
```diff
@@ -0,0 +1,33 @@
+# NULL, NUMBER, STRING & BOOLEAN Literals
+
+ECSQL support following primitive types but not all can be declared as literal in ECSQL though can be inserted/updated and queried in ECSQL.
```
ben-polinsky, line 3:
```suggestion
ECSQL supports the following primitive types. Not all can be declared as literals in ECSQL, though they can be inserted/updated and queried in ECSQL.
```

docs/learning/ECSqlReference/ORDERBY.md
```diff
@@ -0,0 +1,24 @@
+# ORDER BY clause
+
+Sort result by set of expressions in ascending or descending order. It is also use to order nulls in result set by putting them in front or last of results.
```
ben-polinsky, line 3:
```suggestion
Sort result by set of expressions in ascending or descending order. It is also used to order nulls in result set by putting them at the start or end of results.
```

docs/learning/ECSqlReference/Pragmas.md
```diff
@@ -0,0 +1,88 @@
...
+PRAGMA experimental_features_enabled
+
+// experimental_features_enabled
+// -----------------------------
+// False
+```
+
+## `PRAGMA integrity_check` (experimental)
+
+1. `check_ec_profile` - checks if the profile table, indexes, and triggers are present. Does not check of be\_\* tables. Issues are returned as a list of tables/indexes/triggers which was not found or have different DDL.
```
ben-polinsky, line 60:
```suggestion
1. `check_ec_profile` - checks if the profile table, indexes, and triggers are present. Does not check be\_\* tables. Issues are returned as a list of tables/indexes/triggers which were not found or have different DDL.
```

docs/learning/ECSqlReference/Pragmas.md
```diff
@@ -0,0 +1,88 @@
...
+2. `check_data_schema` - checks if all the required data tables and indexes exist for mapped classes. Issues are returned as a list of tables/columns which was not found or have different DDL.
```
ben-polinsky, line 61:
```suggestion
2. `check_data_schema` - checks if all the required data tables and indexes exist for mapped classes. Issues are returned as a list of tables/columns which were not found or have different DDL.
```

docs/learning/ECSqlReference/Pragmas.md
```diff
@@ -0,0 +1,88 @@
...
+3. `check_data_columns` - checks if all the required columns exist in data tables. Issues are returned as a list of those tables/columns.
+4. `check_nav_class_ids` - checks if `RelClassId` of a Navigation property is a valid ECClassId. It does not check the value to match the relationship class.
+5. `check_nav_ids` - checks if `Id` of a Navigation property matches a valid row primary class.
+6. `check_linktable_fk_class_ids` - checks if `SourceECClassId` or `TargetECClassId` of a link table matches a valid ECClassId.
+7. `check_linktable_fk_ids`- checks if `SourceECInstanceId` or `TargetECInstanceId` of a link table matches a valid row in primary class.
+8. `check_class_ids`- checks persisted `ECClassId` in all data tables and make sure they are valid.
```
ben-polinsky, line 67:
```suggestion
8. `check_class_ids`- checks persisted `ECClassId` in all data tables and makes sure they are valid.
```

docs/learning/ECSqlReference/Pragmas.md
```diff
@@ -0,0 +1,88 @@
...
+9. `check_schema_load` - checks if all schemas can be loaded into memory.
+
+```sql
+PRAGMA integrity_check ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES;
+```
+
+output of above will look like listing all check with result and time took to run the test.
```
ben-polinsky, line 74: Perhaps this can be reworded, but I am not sure what is intended.

docs/learning/ECSqlReference/SQLiteScalarFunctions.md
```diff
@@ -0,0 +1,38 @@
+# Scalar SQLite built-in functions
+
+Following is list of SQLite built-in scalar functions exposed in ECSql, see [SQLite documentation](https://www.sqlite.org/lang_corefunc.html) for more details
```
ben-polinsky, line 3:
```suggestion
The following is a list of SQLite built-in scalar functions exposed in ECSql, see [SQLite documentation](https://www.sqlite.org/lang_corefunc.html) for more details
```

docs/learning/ECSqlReference/Views.md
```diff
@@ -0,0 +1,123 @@
...
+
+- The custom attribute `ECDbMap.View` must be applied to a standalone abstract entity class.
+- View cannot be derived from another class
+- View class definition cannot have derived classes.
+- All columns selected in the ECSQL query except for ECInstanceId and ECClassId must have corresponding ECProperties defined on the class.
+- Query column type and property type have to match (int for integers, string for strings etc.).
+- The query has to return an ECInstanceId and an ECClassId column
+- All types of properties and computed expressions can be returned by view query if the class definition defines those properties and their types correctly.
+- Views can be applied only an ECEntityClass.
+- For relationship classes, a different custom attribute `ECDbMap.ForeignKeyBasedView` can be used instead, which will make the runtime go and check both sides for a navigation property which matches the relationship. If one is found, a view for the relationship is automatically generated.
```
ben-polinsky, line 16:
```suggestion
- For relationship classes, a different custom attribute `ECDbMap.ForeignKeyBasedView` can be used instead, which will make the runtime check both sides for a navigation property matching the relationship. If one is found, a view for the relationship is automatically generated.
```

docs/learning/ECSqlReference/Views.md
```diff
@@ -0,0 +1,123 @@
...
+- Any metadata like property category or kind of quantity is taken from the property definitions. Metadata that may come from the view query is overridden by these.
```
ben-polinsky, line 17:
```suggestion
- Any metadata-like property category or kind of quantity is taken from the property definitions. Metadata that may come from the view query is overridden by these.
```

docs/learning/ECSqlReference/Views.md
```diff
@@ -0,0 +1,123 @@
...
+
+```
+Length   |Diameter |Material
+----------------------------------------
+100      |10       |steel
+200      |15       |copper
+150      |20       |plastic
+```
+
+A view may contain navigation properties, for these, the query can use the `navigation_value`function provided by ECSQL [ECSqlFunctions](./ECSqlFunctions.md)
```
ben-polinsky, line 76:
```suggestion
A view may contain navigation properties. For these, the query can use the `navigation_value`function provided by ECSQL [ECSqlFunctions](./ECSqlFunctions.md)
```

docs/learning/ECSqlReference/WindowFunctions.md
```diff
@@ -0,0 +1,138 @@
...
+
+```sql
+SELECT row_number() OVER (ORDER BY a) AS row_number FROM test.Foo;
+```
+
+[Read more.](https://www.sqlite.org/windowfunctions.html#introduction_to_window_functions)
+
+## Window name
+
+Named window definition clauses may also bee added to a `SELECT` statement using a `WINDOW` clause and then referred to by name within window function invocations.
```
ben-polinsky, line 17:
```suggestion
Named window definition clauses may also be added to a `SELECT` statement using a `WINDOW` clause and then referred to by name within window function invocations.
```

docs/learning/ECSqlReference/WindowFunctions.md
```diff
@@ -0,0 +1,138 @@
...
+
+```sql
+SELECT group_concat(b, '.') OVER (
+  win ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
+)
+FROM t1
+WINDOW win AS (PARTITION BY a ORDER BY c);
+```
+
+[Read more.](https://www.sqlite.org/windowfunctions.html#window_chaining)
```
ben-polinsky, line 38:
```suggestion
[Read more](https://www.sqlite.org/windowfunctions.html#window_chaining)
```

docs/learning/ECSqlReference/WindowFunctions.md
```diff
@@ -0,0 +1,138 @@
...
+- An `EXCLUDE` clause.
+
+Ending frame boundary and `EXCLUDE` clause are `optional`.
+
+### Frame type
+
+There are three frame types: `ROWS`, `GROUPS`, and `RANGE`. The frame type determines how the starting and ending boundaries of the frame are measured.
+
+- `ROWS`: The `ROWS` frame type means that the starting and ending boundaries for the frame are determined by counting individual rows relative to the current row.
+- `GROUPS`: The `GROUPS` frame type means that the starting and ending boundaries are determine by counting "groups" relative to the current group. A "group" is a set of rows that all have equivalent values for all all terms of the window ORDER BY clause. ("Equivalent" means that the IS operator is true when comparing the two values.) In other words, a group consists of all peers of a row.
```
ben-polinsky, line 68:
```suggestion
- `GROUPS`: The `GROUPS` frame type means that the starting and ending boundaries are determined by counting "groups" relative to the current group. A "group" is a set of rows all having equivalent values for all terms of the window ORDER BY clause. ("Equivalent" means the IS operator is true when comparing the two values.) In other words, a group consists of all peers of a row.
```

docs/learning/ECSqlReference/WindowFunctions.md
```diff
@@ -0,0 +1,138 @@
...
+- `RANGE`: The `RANGE` frame type requires that the `ORDER BY` clause of the window have exactly one term. Call that term `X`. With the `RANGE` frame type, the elements of the frame are determined by computing the value of expression `X` for all rows in the partition and framing those rows for which the value of `X` is within a certain range of the value of `X` for the current row.
```
ben-polinsky, line 69:
```suggestion
- `RANGE`: The `RANGE` frame type requires that the `ORDER BY` clause of the window has exactly one term. Call that term `X`. With the `RANGE` frame type, the elements of the frame are determined by computing the value of expression `X` for all rows in the partition and framing those rows for which the value of `X` is within a certain range of the value of `X` for the current row.
```

docs/learning/ECSqlReference/WindowFunctions.md
```diff
@@ -0,0 +1,138 @@
...
+
+[Read more.](https://www.sqlite.org/windowfunctions.html#frame_type)
```
ben-polinsky, line 71:
```suggestion
[Read more](https://www.sqlite.org/windowfunctions.html#frame_type)
```

docs/learning/ECSqlReference/WindowFunctions.md
```diff
@@ -0,0 +1,138 @@
...
+  - `GROUPS`: A "group" is a set of peer rows - rows that all have the same values for every term in the `ORDER BY` clause. The frame boundary is the group that is `<expr>` groups before the group containing the current row, or the first group of the partition if there are fewer than `<expr>` groups before the current row.
+  - `RANGE`: For this form, the `ORDER BY` clause of the window definition must have a single term. Call that `ORDER BY` term `X`. Let `Xi` be the value of the `X` expression for the i-th row in the partition and let `Xc` be the value of `X` for the current row. Informally, a `RANGE` bound is the first row for which Xi is within the <expr> of Xc.
+- `CURRENT ROW`: The current row. For `RANGE` and `GROUPS` frame types, peers of the current row are also included in the frame, unless specifically excluded by the `EXCLUDE` clause.
+- `<expr> FOLLOWING`: This is the same as `<expr> PRECEDING` except that the boundary is `<expr>` units after the current rather than before the current row.
+- `UNBOUNDED FOLLOWING`: The frame boundary is the last row in the partition.
+
+[Read more.](https://www.sqlite.org/windowfunctions.html#frame_boundaries)
```
ben-polinsky, line 86:
```suggestion
[Read more](https://www.sqlite.org/windowfunctions.html#frame_boundaries)
```

docs/learning/ECSqlReference/WindowFunctions.md
```diff
@@ -0,0 +1,138 @@
...
+### The `EXCLUDE` clause
+
+The optional `EXCLUDE` clause may take any of the following four forms:
+
+- `EXCLUDE NO OTHERS`: This is the default. In this case no rows are excluded from the window frame as defined by its starting and ending frame boundaries.
+- `EXCLUDE CURRENT ROW`: In this case the current row is excluded from the window frame. Peers of the current row remain in the frame for the `GROUPS` and `RANGE` frame types.
+- `EXCLUDE GROUP`: In this case the current row and all other rows that are peers of the current row are excluded from the frame. When processing an `EXCLUDE` clause, all rows with the same `ORDER BY` values, or all rows in the partition if there is no `ORDER BY` clause, are considered peers, even if the frame type is `ROWS`.
+- `EXCLUDE TIES`: In this case the current row is part of the frame, but peers of the current row are excluded.
+
+[Read more.](https://www.sqlite.org/windowfunctions.html#the_exclude_clause)
```
ben-polinsky, line 97:
```suggestion
[Read more](https://www.sqlite.org/windowfunctions.html#the_exclude_clause)
```

docs/learning/ECSqlReference/WindowFunctions.md
```diff
@@ -0,0 +1,138 @@
...
+),
+group_concat(b, '.') OVER (
+    ORDER BY c ROWS UNBOUNDED PRECEDING
+)
+FROM test.Foo;
+```
+
+## The `FILTER` clause
+
+If a `FILTER` clause is provided, then only rows for which the expr is true are included in the window frame. The aggregate window still returns a value for every row, but those for which the FILTER expression evaluates to other than true are not included in the window frame for any row. [More info.](https://www.sqlite.org/windowfunctions.html#the_filter_clause)
```
ben-polinsky, line 114:
```suggestion
If a `FILTER` clause is provided, then only rows for which the expr is true are included in the window frame. The aggregate window still returns a value for every row, but those for which the FILTER expression evaluates to other than true are not included in the window frame for any row. [More info](https://www.sqlite.org/windowfunctions.html#the_filter_clause)
```

docs/learning/ECSqlReference/ECSqlOptions.md
ColinKerr:
```suggestion
`ECSQLOPTIONS` (or `OPTIONS` for short) are used to specify flags thats will effect processing of the ECSQL statement.
```

docs/learning/ECSqlReference/Views.md
ColinKerr: I refuse to use '-' like this.  It's either one word or two

docs/learning/ECSqlReference/CASE.md
ColinKerr: This is actually what this flavor of CASE statement is called: https://learn.microsoft.com/en-us/sql/t-sql/language-elements/case-transact-sql?view=sql-server-ver16#syntax

docs/learning/ECSqlReference/InstanceQuery.md
ColinKerr: Removed

docs/learning/ECSqlReference/CASE.md
ben-polinsky: Thanks

docs/learning/ECSqlReference/Views.md
ben-polinsky: Alright, well at least here's the justification: 

With `like`, usually, common phrases are one word: _lifelike_, for instance. I'd argue that metadatalike is not a common word. It's also never two words. Otherwise the sentence reads as if you're making a comparison.

Issue #6362, opened 2024-01-12
Locale support in Quantity Formatting


calebmshafer:
Spawned out of #6322.

The [Quantity Formatting](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/quantity-formatting/QuantityFormatter.ts) code in the frontend is missing ability to change the formatting based on different locales being set in the browsers settings. This functionality exists in the [C++ code](https://github.com/iTwin/imodel-native/blob/main/iModelCore/Units/src/Formatting/FormatUtils.cpp#L83) and handles the desktop having a different region set but this was never carried forward to the TS implementation.

Proposal:

- Update quantity formatting API to make the default decimal and thousands separators configurable.  If no separators are explicitly set then the defaults are used.
- Update the frontend code to configure quantity formatting to use the correct separators based on locale.

Two important notes:
- We need to make sure the TypeScript code behaves like the C++ code.  For example we round trip unset separators and when we load a format we do not set the separators to the defaults, rather we leave them blank.
- TypeScript implementation should differ from C++ in one area.  In C++ the formatting code looks directly for locale.  In TypeScript it should be the layer above that sets the config for the formatting.

PR #6387, opened 2024-01-25
Add different decimal separator support to core-quantity's Parser

hl662:
Addresses #6362 

Adds logic, plus test cases to core quantity's Parser, which supports parsing numeric strings that have decimal separators that don't use a period.

### DIFF
--- a/core/quantity/src/test/Parsing.test.ts
+++ b/core/quantity/src/test/Parsing.test.ts
@@ -108,29 +108,17 @@ describe("Parsing tests:", () => {
   });
 
   it("Generate Parse Tokens given different decimal and thousand separators", async () => {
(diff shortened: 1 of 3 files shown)


core/quantity/src/test/Parsing.test.ts
```diff
@@ -107,6 +107,55 @@ describe("Parsing tests:", () => {
     }
   });
 
+  it("Generate Parse Tokens given different decimal and thousand separators", async () => {
+    const testStrings = ["", "0/1", "1/0", "1/2", "-1/2", "+1/2", "2 /", "1,616252eggs", "1,616252E-35eggs", "-1,616252E-35eggs", "756345,345", "12.345,345", "3,6252e3 Miles", "-1 1/2 FT", "+1 1/2 FT", "-135°11'30,5\"", "-2FT 6IN", "135°11'30,5\"", "2FT 6IN", "1 1/2 FT"];
```
ColinKerr, line 111: I would probably narrow down the scenarios to only those that use different decimal and thousands separators to keep the tests more compact or make room for additional new test cases specific to the separator problem

hl662: got it, I've taken a look through the tests and removed ones that don't apply separators at all

ColinKerr (approved):

aruniverse (approved):

PR #6388, opened 2024-01-26
Classes merging order

julioalvahash:
This PR implements multi pass merging for schema classes, the first pass is to create and merge the items, second pass is to complete the merge with properties, baseClass, mixins and custom attributes.


core/ecschema-editing/src/Merging/ClassMerger.ts
```diff
@@ -125,12 +125,14 @@ export class ClassMerger<TClass extends ECClass> {
     }
   }
 
+  // First pass to create missing changes
   public static async mergeChanges(context: SchemaMergeContext, classChanges: Iterable<ClassChanges>) {
```
ColinKerr, line 129: Consider naming these methods better so someone reading them can understand their function without reading your comments.  For example mergeStubs and mergeContents

core/ecschema-editing/src/test/Merging/ClassMergerOrder.test.ts
```diff
@@ -0,0 +1,367 @@
...
+    it("should merge missing mixins regardless of items order", async () => {
+      const sourceSchema = await Schema.fromJson({
+        ...sourceJson,
+        items: {
+          testClass: {
+            schemaItemType: "EntityClass",
+            description: "Test class",
+            mixins: ["SourceSchema.mixinA", "SourceSchema.mixinB"],
+          },
+          mixinA: {
```
ColinKerr, line 39: Since mixins were merged before entity classes this should have worked before your change.  It's important to confirm that a test fails before your change to ensure you're testing the right thing.

The thing your change may have fixed is if a mixing has a base class.  Mixins may have 0 or 1 base class: https://www.itwinjs.org/bis/ec/ec-mixin-class/

I would like to see better test names.  For example 'should merge missing mixins when derived mixin before base mixin'

core/ecschema-editing/src/test/Merging/ClassMergerOrder.test.ts
```diff
@@ -0,0 +1,367 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import {  EntityClass, Schema, SchemaContext, StructClass } from "@itwin/ecschema-metadata";
+import { SchemaMerger } from "../../ecschema-editing";
+import { expect } from "chai";
+
+describe("Struct class merger tests", () => {
```
ColinKerr, line 9: Should all these tests be within the 'Struct class merger tests' group?  You also have a lot of sub describes that just have one test.  I think the type of thing the test applies to should be clear even if it was a flat list of tests under the top level describe

core/ecschema-editing/src/Merging/ClassMerger.ts
```diff
@@ -125,12 +125,14 @@ export class ClassMerger<TClass extends ECClass> {
...
     const merger = new this(context);
+    const changes = Array.from(classChanges);
```
ColinKerr, line 131: Why do we need to create an array from the iterable? can't we just iterate it?

core/ecschema-editing/src/Merging/ClassMerger.ts
```diff
@@ -126,9 +126,9 @@ export class ClassMerger<TClass extends ECClass> {
   }
 
   // First pass to create missing changes
-  public static async mergeChanges(context: SchemaMergeContext, classChanges: Iterable<ClassChanges>) {
+  public static async mergeItemStubChanges(context: SchemaMergeContext, classChanges: Iterable<ClassChanges>) {
     const merger = new this(context);
-    const changes = Array.from(classChanges);
+    const changes = classChanges;
```
ColinKerr, line 131: there is no point in assigning this to a new variable now.  Using the input parameter directly makes the code clearer because the reader does not have to notice the assignment.  As they read the code they will see that the loop is over `classChanges` which is the input parameter.

core/ecschema-editing/src/test/Merging/ClassMergerOrder.test.ts
```diff
@@ -0,0 +1,398 @@
...
+    version: "1.0.0",
+    alias: "target",
+  };
+
+  beforeEach(async () => {
+    context = new SchemaContext();
+  });
+
+  describe("Entity and Mixing class merging order tests", () => {
+    it("should merge the missing entity class with derived base class before the base class", async () => {
```
ColinKerr, line 31: 'derived base class' doesn't make sense, just 'base class' seems to make sense for this tests case

core/ecschema-editing/src/test/Merging/ClassMergerOrder.test.ts
```diff
@@ -0,0 +1,398 @@
...
+    });
+
+    it("should merge missing mixin with derived baseClass and appliesTo before the base class and base appliesTo", async () => {
+      const sourceSchema = await Schema.fromJson({
+        ...sourceJson,
+        items: {
+          mixinA: {
+            schemaItemType: "Mixin",
+            description: "Mixin A",
+            baseClass: "SourceSchema.testBaseClass",
```
ColinKerr, line 111: This is technically wrong.  a mixin may only have another mixing as it's base class ... I suspect schema validation would fail if your an it no this schema.  the 'testBaseClass' should be a mixin.  The test description also have 'derived baseClass' which doesn't make sense

ColinKerr (approved):

PR #6399, opened 2024-02-02
Fix Reality Mesh not logging its memory usage for textures

DStradley:
Reality meshes were not logging their texture memory usage, which is the bulk of their memory usage.  This problem was causing the gpuMemoryLimit functionality to not work properly for reality data, and as a result some datasets would experience WebGL rendering context lost from running out of memory.

See https://github.com/iTwin/itwinjs-backlog/issues/919 and https://github.com/iTwin/itwinjs-backlog/issues/896.

Note that any earlier workarounds used for this problem, that set gpuMemoryLimit to very low values, will now cause excessive tile flushing with the fix, and should be modified to a more reasonable value now.

Also note that large enough reality datasets can still cause rendering context lost when the desktop default of none is used for GpuMemoryLimit.  In those cases, setting a number of 1/4 of the total GPU memory available is a good starting point for the GpuMemoryLimit, and should now prevent the problem.

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -1716,7 +1716,8 @@ export abstract class BaseUnitFormattingSettingsProvider implements UnitFormatti
 export class BatchedTileIdMap {
     constructor(iModel: IModelConnection);
     getBatchId(properties: any): Id64String;
-    getBatchProperties(id: Id64String): any;
+    // (undocumented)
+    getFeatureProperties(id: Id64String): Record<string, any> | undefined;
 }
 
 // @public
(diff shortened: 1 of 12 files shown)


pmconne (approved):

markschlosseratbentley:
fyi @danieliborra

danieliborra:
👏 👏 👏 

That was fast! Thanks David!

danieliborra:
Is this fix going to be included in 4.4 ?

markschlosseratbentley:
> Is this fix going to be included in 4.4 ?

I don't think it quite made it in time for that, but maybe we can backport it if need be.
@aruniverse thoughts?

aruniverse:

PR #6408, opened 2024-02-05
Open checkpoint directly from a cloud container via IPC

kabentley:
For opening checkpoints directly from desktop apps.

### DIFF
--- a/common/config/rush/version-policies.json
+++ b/common/config/rush/version-policies.json
@@ -2,7 +2,7 @@
   {
     "policyName": "prerelease-monorepo-lockStep",
     "definitionName": "lockStepVersion",
-    "version": "4.5.0-dev.7",
+    "version": "4.5.0-dev.8",
     "nextBump": "prerelease"
   }
 ]
--- a/core/backend/package.json
+++ b/core/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-backend",
-  "version": "4.5.0-dev.7",
+  "version": "4.5.0-dev.8",
   "description": "iTwin.js backend components",
   "main": "lib/cjs/core-backend.js",
   "typings": "lib/cjs/core-backend",
@@ -44,9 +44,9 @@
     "url": "http://www.bentley.com"
   },
   "peerDependencies": {
-    "@itwin/core-bentley": "workspace:^4.5.0-dev.7",
-    "@itwin/core-common": "workspace:^4.5.0-dev.7",
-    "@itwin/core-geometry": "workspace:^4.5.0-dev.7",
+    "@itwin/core-bentley": "workspace:^4.5.0-dev.8",
+    "@itwin/core-common": "workspace:^4.5.0-dev.8",
+    "@itwin/core-geometry": "workspace:^4.5.0-dev.8",
     "@opentelemetry/api": "^1.0.4"
   },
   "peerDependenciesMeta": {
--- a/core/bentley/package.json
+++ b/core/bentley/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-bentley",
-  "version": "4.5.0-dev.7",
+  "version": "4.5.0-dev.8",
   "description": "Bentley JavaScript core components",
   "main": "lib/cjs/core-bentley.js",
   "module": "lib/esm/core-bentley.js",
--- a/core/common/package.json
+++ b/core/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-common",
-  "version": "4.5.0-dev.7",
+  "version": "4.5.0-dev.8",
   "description": "iTwin.js components common to frontend and backend",
   "main": "lib/cjs/core-common.js",
   "module": "lib/esm/core-common.js",
@@ -40,8 +40,8 @@
     "js-base64": "^3.6.1"
   },
   "peerDependencies": {
-    "@itwin/core-bentley": "workspace:^4.5.0-dev.7",
-    "@itwin/core-geometry": "workspace:^4.5.0-dev.7"
+    "@itwin/core-bentley": "workspace:^4.5.0-dev.8",
+    "@itwin/core-geometry": "workspace:^4.5.0-dev.8"
   },
   "devDependencies": {
     "@itwin/build-tools": "workspace:*",
--- a/core/ecschema-editing/package.json
+++ b/core/ecschema-editing/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-editing",
-  "version": "4.5.0-dev.7",
+  "version": "4.5.0-dev.8",
   "description": "ECSchema editing and validation API",
   "license": "MIT",
   "main": "lib/cjs/ecschema-editing.js",
@@ -62,9 +62,9 @@
     "typescript": "~5.0.2"
   },
   "peerDependencies": {
-    "@itwin/core-bentley": "workspace:^4.5.0-dev.7",
-    "@itwin/core-quantity": "workspace:^4.5.0-dev.7",
-    "@itwin/ecschema-metadata": "workspace:^4.5.0-dev.7"
+    "@itwin/core-bentley": "workspace:^4.5.0-dev.8",
+    "@itwin/core-quantity": "workspace:^4.5.0-dev.8",
+    "@itwin/ecschema-metadata": "workspace:^4.5.0-dev.8"
   },
   "nyc": {
     "extends": "./node_modules/@itwin/build-tools/.nycrc"
--- a/core/ecschema-locaters/package.json
+++ b/core/ecschema-locaters/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-locaters",
-  "version": "4.5.0-dev.7",
+  "version": "4.5.0-dev.8",
   "description": "EC Schema file locaters",
   "license": "MIT",
   "main": "lib/cjs/ecschema-locaters.js",
@@ -75,7 +75,7 @@
     "@xmldom/xmldom": "~0.8.5"
   },
   "peerDependencies": {
-    "@itwin/ecschema-metadata": "workspace:^4.5.0-dev.7"
+    "@itwin/ecschema-metadata": "workspace:^4.5.0-dev.8"
   },
   "nyc": {
     "extends": "./node_modules/@itwin/build-tools/.nycrc",
--- a/core/ecschema-metadata/package.json
+++ b/core/ecschema-metadata/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-metadata",
-  "version": "4.5.0-dev.7",
+  "version": "4.5.0-dev.8",
   "description": "ECObjects core concepts in typescript",
   "license": "MIT",
   "main": "lib/cjs/ecschema-metadata.js",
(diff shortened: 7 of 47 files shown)


full-stack-tests/backend/src/integration/Checkpoints.test.ts
```diff
@@ -153,36 +152,26 @@ describe("Checkpoints", () => {
...
-    });
-    sinon.stub(CheckpointManager, "validateCheckpointGuids").callsFake(() => { });
-
-    await SnapshotDb.openCheckpointV2({
-      accessToken,
-      iTwinId: testITwinId,
-      iModelId: testIModelId,
-      changeset: testChangeSet,
-    });
-    expect(prefetchSpy.callCount).to.equal(1);
```
nick4598, line 176: A test which makes sure we don't double prefetch seems valuable to me. Do we need to remove this? 
This confirms that two backends which open the same iModels don't trigger a prefetch more than once ( thus giving that prefetch more weight ).

core/backend/src/IModelDb.ts
```diff
@@ -2931,16 +2917,23 @@ export class SnapshotDb extends IModelDb {
...
-    }, (10 * 60) * 1000).unref(); // 10 * 60 is 10 minutes in seconds, then converted to milliseconds (* 1000);
-    snapshot._refreshSas = new RefreshV2CheckpointSas(container.accessToken, checkpoint.reattachSafetySeconds);
+    }, (10 * 60) * 1000).unref(); // 10 minutes
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    snapshot._refreshSas = new RefreshV2CheckpointSas(snapshot.nativeDb.cloudContainer!.accessToken, checkpoint.reattachSafetySeconds);
     return snapshot;
   }
 
+  // Open a Checkpoint directly from a cloud container.
+  public static async openCheckpoint(checkpoint: CheckpointProps): Promise<SnapshotDb> {
```
nick4598, line 2927: Should this be marked @beta? 

Should we also consider making prefetching more visible? I imagine an IPC app may want to be able to prefetch / know how to trigger one.

nick4598 (commented):
Does this PR actually expose this as part of IPC? Or just make it easier to expose it as part of IPC? I imagine there would have to be a change to IpcAppProps, right? I might be confused about this point though.

full-stack-tests/backend/src/integration/Checkpoints.test.ts
kabentley: It is illegal to open the same iModel more than once from a backend. You stubbed out "openFile" and "validate.." to avoid that rule, so this "issue" really only applies to daemons. We never open the same iModel/changeset from two backends on the same vm. Even if we did, prefetching an already perfectched block doesn't do anything.

nick4598: You're right this issue only applies to daemons, but why shouldn't we test it (even if its a simulated version of a test where I had to stub some calls out?).

>We never open the same iModel/changeset from two backends on the same vm

DR has started deferring rpc calls to certain interfaces(tile, read, presentation) to GPB. So that would mean two backends reading the same imodel/changeset. Prefetching an already prefetched block does do nothing, you're right. But there will be two prefetch objects prefetching the same iModel if we were to allow two prefetches on the same iModel.

core/backend/src/CheckpointManager.ts
```diff
@@ -149,30 +149,14 @@ export class V2CheckpointManager {
...
-      } else {
-        // Make sure the checkpoint_cache_dir has an iTwinDaemon specific file in it, otherwise fall back to other directory.
-        if (!(IModelJsFs.existsSync(path.join(cacheDir, "portnumber.bcv")))) {
-          cacheDir = this.getFolder();
-          Logger.logWarning(loggerCategory, `No evidence of the iTwinDaemon in provided CHECKPOINT_CACHE_DIR: ${process.env.CHECKPOINT_CACHE_DIR}, using ${cacheDir} instead.`);
-        }
-      }
+      let cacheDir: string | undefined = process.env.CHECKPOINT_CACHE_DIR ?? this.getFolder();
+      // See if there is a daemon running, otherwise use profile directory for cloudCache
+      if (!(IModelJsFs.existsSync(path.join(cacheDir, "portnumber.bcv"))))
```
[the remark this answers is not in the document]
kabentley: > for the daemonless case and why you want it to live under a profile directory

yes, that's why profiles exist. No two processes may share a cloud cache. 

> Would we be better off just encouraging downloading checkpoints to a shared directory if they're only going to be opened readonly anyways?

No, you're *way* better off opening a checkpoints from their container rather than downloading them, because: 
 - startup is faster. Often people will only need a few blocks. Prefetching works if you need more.
 - disk space management. This is nearly impossible without cloud caches.
 - blocks are shared across checkpoints. This will be the biggest benefit for many desktop use cases where people want to compare versions.
 
In the (rare) case where people use the same checkpoint with multiple profiles, it will use a little more disk space. That is not an issue.

kabentley:
> Does this PR actually expose this as part of IPC? Or just make it easier to expose it as part of IPC? I imagine there would have to be a change to IpcAppProps, right? I might be confused about this point though.

I'm still working on that. I just made this PR so I could get the CI jobs to run on other platforms. I changed it to Draft.

core/backend/src/CheckpointManager.ts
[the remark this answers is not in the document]
kabentley: It turns out that it is not really possible for two processes to share the same downloaded checkpoint at the same time anyway, even if they both open it for read. That's because we create and write to "cache" files (e.g. ".Tiles") during the session. I'm not sure how we can stop that problem

full-stack-tests/backend/src/integration/Checkpoints.test.ts
kabentley: i re-added that test, but i'm not really even sure about the policy. If we really do have two backends on the same VM sharing a daemon and they both are using the same checkpoint, why *not* start two prefetches. I'm not sure two prefetch requests will actually do anything different than one, but certainly if they did somehow bump the priority for that checkpoint over others, isn't that a good thing (given there are two clients waiting for it)?

nick4598: From my limited testing, it did do something different in that more prefetches occurred for that checkpoint. You have a good point, but in this particular scenario I guess I view this as an artificial inflation of 1 client to 2 clients, since GPB is really only there to serve the DR backend's requests and not because someone else opened the same checkpoint in some other frontend application.

core/backend/src/IModelDb.ts
kabentley: Yes, i do think we should add documentation about how prefetching works (and of how you indicate that a checkpoint should be prefetched in particular). I leave it to you whether you want to add that to this PR or another one.

Along those lines, I'm not sure I understand the logic of skipping prefetches for big iModels. It seems to me that's where it will help the most (although I recognize the cachefile will fill/spill quicker). How did we arrive at the default 2G size? That seems way too small to me, but perhaps a better approach is to limit the time a prefetch can take, or maybe some other way to determine when to stop prefetching. For desktops, the default should be different. This topic should be researched/tested and I don't volunteer for that.

nick4598: I will address that(documentation + prefetching desktop) in another PR then. I agree that in the desktop case we definitely need to reconsider things and should prefetch probably everything (as long as user requests it and they understand the ramifications, which all comes with the documentation part)

2G default was based on poorer performance across a node (not just the imodel thats being prefetched) when prefetching XL iModels. I think we can probably bump up that 2G in our cluster. Additionally I think just prefetching one XL iMdoel slowed down the time it took for its requests to finish in the stresstest we were running, but I'd have to look back at my notes.

I also agree we could probably use a better approach in the daemon case. I still like the idea of having the daemon maximize its throughput and prefetching everything but backing off (especially on the larger iModels) when it needs to in the case of higher activity. Feel free to resolve this comment unless you want to continue the discussion.

core/backend/src/IModelDb.ts
```diff
@@ -2931,16 +2921,22 @@ export class SnapshotDb extends IModelDb {
...
-    snapshot._refreshSas = new RefreshV2CheckpointSas(container.accessToken, checkpoint.reattachSafetySeconds);
+    }, (10 * 60) * 1000).unref(); // 10 minutes
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    snapshot._refreshSas = new RefreshV2CheckpointSas(snapshot.nativeDb.cloudContainer!.accessToken, checkpoint.reattachSafetySeconds);
     return snapshot;
   }
 
+  // Open a Checkpoint directly from a cloud container.
+  public static async openCheckpoint(args: OpenCheckpointArgs): Promise<SnapshotDb> {
+    return this.attachAndOpenCheckpoint(await CheckpointManager.toCheckPointProps(args));
```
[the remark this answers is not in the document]
kabentley: done

core/bentley/src/BentleyError.ts
```diff
@@ -668,7 +668,6 @@ export class BentleyError extends Error {
       case GeoServiceStatus.CSMapError: return "CSMap error";
       case GeoServiceStatus.Pending: return "Pending";
       case RealityDataStatus.InvalidData: return "Invalid or unknown data";
-      case IModelStatus.Success:
```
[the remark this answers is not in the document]
kabentley: The problem is that both `BentleyStatus.SUCCESS` and `IModelStatus.Success` are 0, so having them both here results in 2 redundant case statements. I was tired of seeing the linter complaining every build.

full-stack-tests/backend/src/integration/Checkpoints.test.ts
```diff
@@ -265,95 +256,87 @@ describe("Checkpoints", () => {
...
 
       const container = (await containerSpy.returnValues[0]).container;
-      let stats = container.queryBcvStats({addClientInformation: true});
+      let stats = container.queryBcvStats({ addClientInformation: true });
       const populatedCacheslots = stats.populatedCacheslots;
       // Opening the database causes some blocks to be downloaded.
       expect(stats.populatedCacheslots).to.be.greaterThan(0);
       // Only one database and this db has a default txn open so all it's local blocks should be locked.
-      expect(stats.populatedCacheslots).to.equal(stats.lockedCacheslots);
+      expect(stats.lockedCacheslots).greaterThan(0);
```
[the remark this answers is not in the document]
kabentley: yes, it was a different random value every run, probably based on the progress of the prefetch from a previous step.

docs/learning/frontend/IModelConnection.md
```diff
@@ -10,7 +10,17 @@ Various subclasses of [IModelConnection]($frontend) are used depending on the ty
...
+[CheckpointConnection]($frontend) is used to create a **readonly** connection to a Checkpoint of an iModel. A `CheckpointConnection` is initiated via [CheckpointConnection.openRemote]($frontend).
+
+If the current process is a a browser frontend, the connection will to a remote backend over the internet using RPC. In that case, due to the nature of RPC requests, the backend servicing this connection may be servicing many simultaneous frontends, and may even change over time.
+
+If the current process is connected to a private backend over IPC (e.g. on a desktop via Electron), the connection will be to Checkpoint from that backend.
```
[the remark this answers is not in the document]
kabentley: fixed typo

full-stack-tests/backend/src/integration/Checkpoints.test.ts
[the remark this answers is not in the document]
nick4598: Hm thats concerning. I wonder if another read of the database + a queryBcvStats would resolve it, but doesn't seem worth it.

nick4598 (approved):

pmconne (approved):

PR #6379, opened 2024-01-20
glTF instancing, plus bounding box fixes

pmconne:
Fixes #6378 by adding support for the glTF 2.0 instancing extension described in the issue. It successfully renders the model referenced in the issue.

Fixes #5937. Fixes https://github.com/iTwin/itwinjs-backlog/issues/969. Both relate to incorrect bounding boxes computed for glTF models. For unclear reasons we were applying inverse transforms to the bounding boxes.

TODO:
- [x] Investigate and fix model drawing black instead of white
  - Good ol' white-on-white reversal...
- [x] Run ImageTests and evaluate results
  - Similar inconcsistent pixel diffs as occur on master
- [x] Remove debug bounding box graphic

PR #6409, opened 2024-02-05
Fix 3d tiles bounding box regression

pmconne:
Introduced in #6379.
A Batch's content range needs to be in the local coordinate system of the batch.
Most batches are tiles. In that case the caller supplies a content range in the coordinate space of the tile tree, and we must apply the inverse transform to put it into the tile's coordinate space.

On the other hand, when decoding glTF graphics outside of the context of a tile tree, no content range is supplied - we must compute it from the geometry as we decode it.
In that case, the computed content range is in tile space, so no transform is required.
By "fixing" this case by removing the transform unconditionally, #6379 broke tiles.

This PR fixes the tile regression while preserving the fix for glTF.


aruniverse (approved):

aruniverse:

pmconne:

PR #6411, opened 2024-02-06
Increase log level of a single RPC log

GytisCepk:
To better identify backend errors, we need a way to filter out non 5XX HTTP errors as well as see error messages in logs.

### DIFF
--- a/core/backend/src/rpc/web/logging.ts
+++ b/core/backend/src/rpc/web/logging.ts
@@ -6,7 +6,7 @@
  * @module RpcInterface
  */
 
-import { Logger } from "@itwin/core-bentley";
+import { BentleyError, Logger, StatusCategory } from "@itwin/core-bentley";
 import { CommonLoggerCategory, RpcInvocation, RpcProtocolEvent, WebAppRpcLogging } from "@itwin/core-common";
 import * as os from "os";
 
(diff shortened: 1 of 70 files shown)


paulius-valiunas (approved):

core/backend/src/rpc/web/logging.ts
```diff
@@ -31,15 +31,19 @@ export class WebAppRpcLoggingBackend extends WebAppRpcLogging {
...
       errorMessage,
+      error: BentleyError.getErrorProps(result),
       // Alert! The following properties are required by Bentley DevOps standards. Do not change their names!
       ActivityId: invocation.request.id, // eslint-disable-line @typescript-eslint/naming-convention
       MachineName: this.getHostname(), // eslint-disable-line @typescript-eslint/naming-convention
       ...pathIds,
-    }));
+    };
+
+    Logger.logError(CommonLoggerCategory.RpcInterfaceBackend, `${message}.${operationDescriptor}`, metadata);
```
aruniverse, line 47: This supposed to be logError?

aruniverse (commented):
Good first pass

core/backend/src/rpc/web/logging.ts
GytisCepk: Yes.

It was `logInfo`, but this is only logged on error so I'm not sure why it was only `info` (+ this is inside function called `logErrorBackend`).

pmconne (approved):

aruniverse:

aruniverse:

PR #6415, opened 2024-02-06
Skip PR Val if native PR is linked

DanRod1999:
When a native PR is linked, itwinjs-core builds and tests can be skipped. The native PR will run the core tests for us (on win, mac, linux) and post the success or failure if needed. 

This change is needed to cut out extra build time for when addon releases are automated. If we don't do something like this than when we push the updated version of imodeljs-native:X.X.X it will spend at least 20 minutes building and in that time people could potentially merge PRs that could break the new addon.

This would also mean we should make the 'itwinjs-core PR validation' status a required check.


nick4598:
Can you explain this 'This would also mean we should make the 'itwinjs-core PR validation' status a required check.' a bit more?

common/config/azure-pipelines/jobs/fast-ci.yaml
```diff
@@ -71,3 +115,28 @@ jobs:
           nodeVersion: 20.x
       # The publish script identifies any new packages not previously published and tags the build
       - template: ../templates/publish.yaml
+
+  - job: PostResult
+    dependsOn:
+    - CheckLinkedPR
```
nick4598, line 121: So theres a job in the imodel-native pipeline which runs nearly the same bash script, correct?

I also thought that if there was a linked PR we skipped itwinjs core build anyways. Or I guess this was never necessary because its PRs into imodel02 which would typically have a linked PR (and therefore no build reqs to merge into imodel02 on this repo). Is that the reason?

DanRod1999: Yeah we never skipped builds before because imodel02 PRs skipped by default.

And yes in imodel-native theres a similar bash script however much more complicated because that pipeline needs a lot more info. This just needs to know a pr was linked.

DanRod1999:
> Can you explain this 'This would also mean we should make the 'itwinjs-core PR validation' status a required check.' a bit more?

Right now if we link PR between imodel-native and itwinjs-core, when a imodel-native PR completes its validation it will post a status check call `itwinjs-core PR validation` on this PR. However, we do not have it as a required check. So the user can merge even if it fails. I'm saying we should make it required so it prevents merging on failed imodel-native runs.

Because it would be required I've added some logic into this PR so that if no imodel-native PR is linked we still post a success message on a completed build.

nick4598:
> > Can you explain this 'This would also mean we should make the 'itwinjs-core PR validation' status a required check.' a bit more?
> 
> Right now if we link PR between imodel-native and itwinjs-core, when a imodel-native PR completes its validation it will post a status check call `itwinjs-core PR validation` on this PR. However, we do not have it as a required check. So the user can merge even if it fails. I'm saying we should make it required so it prevents merging on failed imodel-native runs.
> 
> Because it would be required I've added some logic into this PR so that if no imodel-native PR is linked we still post a success message on a completed build.

That makes sense. making something a required check is probably github administrative stuff? Not a yaml change?

nick4598 (approved):

nick4598:
> Yeah i believe its a setting change in itwinjs-core PRs

If we require it will PRs which aren't linked to a native PR still be okay? That check would be required but wouldn't be run in those cases, correct?

aruniverse (approved):

PR #6417, opened 2024-02-06
Fix case in Typescript code for pattern_useconstantlod parameter to match backend code.

MarcNeely:
There is an inconsistency between the name of the pattern_useconstantlod parameter in TextureMapProps between the Typescript code and the backend code.  This PR makes it match so that modifying the property in Typescript will actually have an effect.

Fixes https://github.com/orgs/iTwin/discussions/23


pmconne (approved):

pmconne:

pmconne:
@aruniverse @DanRod1999 fyi

aruniverse:
> @aruniverse @DanRod1999 fyi

fyi @anmolshres98

core/common/src/MaterialProps.ts
```diff
@@ -63,14 +63,14 @@ export interface TextureMapProps {
   /** Weight at which to combine diffuse image and color; if undefined, defaults to 1.0 */
   pattern_weight?: number;
   /** If true, override the mapping mode with constant LOD mapping for the normal map, defaults to false. */
-  pattern_useConstantLod?: boolean;
-  /** The number of times the texture is repeated if pattern_useConstantLod is true.  Increasing this will make the texture pattern appear smaller, decreasing it will make it larger. Defaults to 1.*/
+  pattern_useconstantlod?: boolean;
```
pmconne, line 66: The camel-cased one is `@public` which means you can't delete it without breaking people's code, which we can't do in a minor release.
Mark it `@deprecated in 4.4. It never functioned properly - use [[pattern_useconstantlod]] instead`.

MarcNeely: done.

pmconne (approved):

PR #6420, opened 2024-02-07
Support Cesium bathymetry

pmconne:
By enabling terrain providers to specify a specific terrain data source (in this case, Cesium ION asset).
The alternative would have been to add a new CesiumWorldBathymetry provider, which would be far less flexible/future-proof.

See NextVersion.md for details.


core/common/src/TerrainSettings.ts
```diff
@@ -80,11 +105,31 @@ export class TerrainSettings {
     return this._nonLocatable;
   }
 
-  constructor(providerName: string = "CesiumWorldTerrain", exaggeration: number = 1.0, applyLighting = false, heightOrigin = 0.0, heightOriginMode = TerrainHeightOriginMode.Geodetic) {
-    this.providerName = providerName;
-    this.exaggeration = Math.min(100, Math.max(0.1, exaggeration));
-    this.applyLighting = applyLighting;
-    this.heightOrigin = heightOrigin;
+  /** @deprecated in 4.x. Use the overload that takes [[TerrainProps]]. */
```
aruniverse, line 108:
```suggestion
  /** @deprecated in 4.5.x. Use the overload that takes [[TerrainProps]]. */
```

core/frontend/src/RealityDataSourceCesiumIonAssetImpl.ts
```diff
@@ -112,7 +112,7 @@ export class RealityDataSourceCesiumIonAssetImpl implements RealityDataSource {
     // The following is only if the reality data is not stored on PW Context Share.
     const cesiumAsset = CesiumIonAssetProvider.parseCesiumUrl(url);
     if (cesiumAsset) {
-      const tokenAndUrl = await getCesiumAccessTokenAndEndpointUrl(cesiumAsset.id, cesiumAsset.key);
+      const tokenAndUrl = await getCesiumAccessTokenAndEndpointUrl(`${cesiumAsset.id}`, cesiumAsset.key);
```
aruniverse, line 115: just trying to cast to a string or?

pmconne: `cesiumAsset.id` is a number, `getCesiumAccessTokenAndEndpointUrl` expects a string.

core/common/src/TerrainSettings.ts
aruniverse: i think adding the minor is useful. i _believe_ if an api was deprecated in 4.0, you could remove it in 5.0, where as if it was deprecated in 4.1 it cant be removed til 6.0

aruniverse (approved):

markschlosseratbentley (approved):

PR #6421, opened 2024-02-07
Add `GenericInstanceFilter` definition

saskliutas:
Moved `GenericInstanceFilter` definition from `presentation-components` to `core-common`. It was introduced as an intermediate format when converting `PresentationInstanceFilter` into other formats (e.g. ECSQL) and does not depend on `presentation`. Having it in `core-common` allows to use it without having dependency on `presentation-components` and on `AppUI` as a result.

### DIFF
--- a/core/common/src/GenericInstanceFilter.ts
+++ b/core/common/src/GenericInstanceFilter.ts
@@ -72,12 +72,15 @@ export namespace GenericInstanceFilterRuleValue {
     id: string;
     className: string;
   }
+  /** Checks if supplied value is `Point2d` like. Returns `true` for `Point2d` and `Point3d`. */
   export function isPoint2d(value: GenericInstanceFilterRuleValue.Values): value is GenericInstanceFilterRuleValue.Point2d {
     return (value as GenericInstanceFilterRuleValue.Point2d).x !== undefined && (value as GenericInstanceFilterRuleValue.Point2d).y !== undefined;
   }
+  /** Checks if supplied value is `Point3d` like. */
   export function isPoint3d(value: GenericInstanceFilterRuleValue.Values): value is GenericInstanceFilterRuleValue.Point3d {
     return isPoint2d(value) && (value as GenericInstanceFilterRuleValue.Point3d).z !== undefined;
   }
+  /** Checks if supplied value is `InstanceKey` like. */
   export function isInstanceKey(value: GenericInstanceFilterRuleValue.Values): value is GenericInstanceFilterRuleValue.InstanceKey {
     return (value as GenericInstanceFilterRuleValue.InstanceKey) !== undefined && (value as GenericInstanceFilterRuleValue.InstanceKey).className !== undefined;
   }
--- a/core/common/src/GenericInstanceFilter.ts
+++ b/core/common/src/GenericInstanceFilter.ts
@@ -110,7 +110,7 @@ export interface GenericInstanceFilterRule {
    */
   operator: GenericInstanceFilterRuleOperator;
   /**
-   * Value to which property values is compared to. For unary operators value is 'undefined'.
+   * Value to which property values is compared to. For unary operators value is `undefined`.
    */
   value?: GenericInstanceFilterRuleValue;
   /**
@@ -161,16 +161,16 @@ export interface GenericInstanceFilterRelatedInstanceDescription {
  * @beta
  */
 export interface GenericInstanceFilterRelationshipStep {
-  /** Class name of the source class. */
+  /** Full class name of the source class, e.g. `BisCore.Element`. */
   sourceClassName: string;
   /** Class name of the target class. */
   targetClassName: string;
   /** Class name of the relationship class that should be used to move from source to target. */
   relationshipClassName: string;
   /**
-   * Flag that describes if this step follows relationship class in `Forward` or `Backward` direction.
-   * If step follows relationship class in `Forward` direction than `sourceClassName` matches relationship source class and `targetClassName` matches relationship target class.
-   * Otherwise `sourceClassName` matches relationship target class and `targetClassName` matches relationship source class.
+   * A flag that describes if this step follows relationship class in forward or backward direction.
+   * If the step follows relationship in forward direction then `sourceClassName` matches relationship's source class and `targetClassName` matches relationship's target class.
+   * Otherwise, `sourceClassName` matches relationship's target class and `targetClassName` matches relationship's source class.
    */
   isForwardRelationship: boolean;
 }
--- a/core/common/src/GenericInstanceFilter.ts
+++ b/core/common/src/GenericInstanceFilter.ts
@@ -166,9 +166,9 @@ export interface GenericInstanceFilterRelatedInstanceDescription {
 export interface GenericInstanceFilterRelationshipStep {
   /** Full class name of the source class, e.g. `BisCore.Element`. */
   sourceClassName: string;
-  /** Class name of the target class. */
+  /** Full class name of the target class, e.g. `BisCore.Element`. */
   targetClassName: string;
-  /** Class name of the relationship class that should be used to move from source to target. */
+  /** Full class name of the relationship class that should be used to move from source to target, e.g. `BisCore.ElementOwnsChildElements`. */
   relationshipClassName: string;
   /**
    * A flag that describes if this step follows relationship class in forward or backward direction.
--- a/core/common/src/GenericInstanceFilter.ts
+++ b/core/common/src/GenericInstanceFilter.ts
@@ -16,7 +16,7 @@ export interface GenericInstanceFilter {
   /**
    * Information about related instances that has access to the properties used in filter.
    * These can be used to create `JOIN` clause when building `ECSQL` query. Each related property
-   * used in rule will have `sourceAlias` that matches `GenericInstanceFilterRelatedInstanceDescription.alias`.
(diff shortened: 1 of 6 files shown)


core/common/src/GenericInstanceFilter.ts
```diff
@@ -0,0 +1,187 @@
...
+  alias: string;
+}
+
+/**
+ * Describes single step between source class and target class.
+ * @beta
+ */
+export interface GenericInstanceFilterRelationshipStep {
+  /** Class name of the source class. */
+  sourceClassName: string;
```
grigasp, line 165: the comment for class name properties should be more specific about the format

```suggestion
  /** Full class name of the source class, e.g. `BisCore.Element`. */
  sourceClassName: string;
```

core/common/src/GenericInstanceFilter.ts
```diff
@@ -0,0 +1,187 @@
...
+  /**
+   * Property name for accessing property value.
+   */
+  propertyName: string;
+  /**
+   * Comparison operator that should be used to compare property value.
+   */
+  operator: GenericInstanceFilterRuleOperator;
+  /**
+   * Value to which property values is compared to. For unary operators value is 'undefined'.
```
grigasp, line 113:
```suggestion
   * Value to which property values is compared to. For unary operators value is `undefined`.
```

core/common/src/GenericInstanceFilter.ts
```diff
@@ -0,0 +1,187 @@
...
+  /** Class name of the target class. */
+  targetClassName: string;
+  /** Class name of the relationship class that should be used to move from source to target. */
+  relationshipClassName: string;
+  /**
+   * Flag that describes if this step follows relationship class in `Forward` or `Backward` direction.
+   * If step follows relationship class in `Forward` direction than `sourceClassName` matches relationship source class and `targetClassName` matches relationship target class.
+   * Otherwise `sourceClassName` matches relationship target class and `targetClassName` matches relationship source class.
+   */
+  isForwardRelationship: boolean;
```
grigasp, line 175: I don't think "forward" and "backward" should be specified in backticks, since that's not an enum value or something like that.

```suggestion
  /**
   * A flag that describes if this step follows relationship class in forward or backward direction.
   * If the step follows relationship in forward direction then `sourceClassName` matches relationship's source class and `targetClassName` matches relationship's target class.
   * Otherwise, `sourceClassName` matches relationship's target class and `targetClassName` matches relationship's source class.
   */
  isForwardRelationship: boolean;
```

core/common/src/test/GenericInstanceFilter.test.ts
```diff
@@ -0,0 +1,52 @@
...
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { expect } from "chai";
+import { GenericInstanceFilter, GenericInstanceFilterRule, GenericInstanceFilterRuleValue } from "../GenericInstanceFilter";
+
+describe("GenericInstanceFilterRuleValue", () => {
+  it("'isPoint2d' returns correct result", () => {
+    expect(GenericInstanceFilterRuleValue.isPoint2d({ x: 1, y: 2})).to.be.true;
```
grigasp, line 11:
```suggestion
    expect(GenericInstanceFilterRuleValue.isPoint2d({ x: 1, y: 2 })).to.be.true;
```

core/common/src/test/GenericInstanceFilter.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    expect(GenericInstanceFilterRuleValue.isPoint2d(1)).to.be.false;
+    expect(GenericInstanceFilterRuleValue.isPoint2d(false)).to.be.false;
+    expect(GenericInstanceFilterRuleValue.isPoint2d("text")).to.be.false;
+    expect(GenericInstanceFilterRuleValue.isPoint2d(new Date())).to.be.false;
+    expect(GenericInstanceFilterRuleValue.isPoint2d({ id: "0x1", className: "TestClass"})).to.be.false;
+  });
+
+  it("'isPoint3d' returns correct result", () => {
+    expect(GenericInstanceFilterRuleValue.isPoint3d({ x: 1, y: 2, z: 3})).to.be.true;
+    expect(GenericInstanceFilterRuleValue.isPoint3d({ x: 1, y: 2})).to.be.false;
```
grigasp, line 21:
```suggestion
    expect(GenericInstanceFilterRuleValue.isPoint3d({ x: 1, y: 2, z: 3 })).to.be.true;
    expect(GenericInstanceFilterRuleValue.isPoint3d({ x: 1, y: 2 })).to.be.false;
```

core/common/src/test/GenericInstanceFilter.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    expect(GenericInstanceFilterRuleValue.isPoint3d(false)).to.be.false;
+    expect(GenericInstanceFilterRuleValue.isPoint3d("text")).to.be.false;
+    expect(GenericInstanceFilterRuleValue.isPoint3d(new Date())).to.be.false;
+    expect(GenericInstanceFilterRuleValue.isPoint3d({ id: "0x1", className: "TestClass"})).to.be.false;
+  });
+
+  it("'isInstanceKey' returns correct result", () => {
+    expect(GenericInstanceFilterRuleValue.isInstanceKey({ id: "0x1", className: "TestClass"})).to.be.true;
+    expect(GenericInstanceFilterRuleValue.isInstanceKey({ x: 1, y: 2, z: 3})).to.be.false;
+    expect(GenericInstanceFilterRuleValue.isInstanceKey({ x: 1, y: 2})).to.be.false;
```
grigasp, line 32:
```suggestion
    expect(GenericInstanceFilterRuleValue.isInstanceKey({ id: "0x1", className: "TestClass" })).to.be.true;
    expect(GenericInstanceFilterRuleValue.isInstanceKey({ x: 1, y: 2, z: 3 })).to.be.false;
    expect(GenericInstanceFilterRuleValue.isInstanceKey({ x: 1, y: 2 })).to.be.false;
```

core/common/src/test/GenericInstanceFilter.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  it("'isFilterRuleGroup' returns correct result", () => {
+    const rule: GenericInstanceFilterRule = {
+      operator: "is-equal",
+      propertyName: "PropName",
+      propertyTypeName: "string",
+      sourceAlias: "alias",
+    };
+
+    expect(GenericInstanceFilter.isFilterRuleGroup(rule)).to.be.false;
+    expect(GenericInstanceFilter.isFilterRuleGroup({ operator: "and", rules: [rule]})).to.be.true;
```
grigasp, line 50:
```suggestion
    expect(GenericInstanceFilter.isFilterRuleGroup({ operator: "and", rules: [rule] })).to.be.true;
```

core/common/src/test/GenericInstanceFilter.test.ts
```diff
@@ -0,0 +1,52 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { expect } from "chai";
+import { GenericInstanceFilter, GenericInstanceFilterRule, GenericInstanceFilterRuleValue } from "../GenericInstanceFilter";
+
+describe("GenericInstanceFilterRuleValue", () => {
+  it("'isPoint2d' returns correct result", () => {
```
grigasp, line 10: We're not testing `isPoint2d` against a `Point3d` value, so it's not entirely clear what to expect from it. Looking at the implementation, I see it returns `true` for `Point3d` values - is that intended? If yes, please mention that in `isPoint2d` comment. If no, then the function needs to be fixed. An additional check for `Point3d` value in this test is needed either way, IMO.

core/common/src/GenericInstanceFilter.ts
saskliutas: Updated comments on other props.

core/common/src/test/GenericInstanceFilter.test.ts
saskliutas: It wasn't intentional I just wanted to keep implementation simple. I don't see a problem with returning `true` for `Point3d` so updated test and added comment on `isPoint2d`.

core/common/src/GenericInstanceFilter.ts
grigasp: Just to confirm, do we accept only `.`, or `:` is also fine?

saskliutas: Actually we accept only `:` as we use class names that we get from content descriptor.

core/common/src/GenericInstanceFilter.ts
```diff
@@ -0,0 +1,190 @@
...
+   */
+  propertyClassNames: string[];
+  /**
+   * List of class names which will be used for additionally only querying instances of specific classes.
+   */
+  filteredClassNames?: string[];
+}
+
+/**
+ * Type definition that describes operators supported by `GenericInstanceFilterRule`.
```
pmconne, line 35: Please use doc links throughout (`[[Thing]]` and `[Thing]($package)` instead of `` `Thing` ``).

pmconne (approved):

saskliutas:

PR #6427, opened 2024-02-08
Fix an issue where the background map would become blank after adding a new map-layer

mdastous-bentley:
It was recently discovered that if any newly added map-layers get stuck during the “initialize” phase, all map-layers are not drawn until all tile tree are marked as loaded. So if for some reason a server is slow to respond during this phase, we lose all map-layers tiles (even the previously loaded tiles).


pmconne (approved):

mdastous-bentley:
@aruniverse Is this the only PR getting those tests failure?

aruniverse (approved):

PR #6434, opened 2024-02-12
As/update-gather-docs-after-patch

anmolshres98:
Update gather-docs.yaml branchName value to `release/4.4.x` after first successful patch release


aruniverse (approved):
what happened to the goal for automating this?

anmolshres98:
> what happened to the goal for automating this?

the previous gather docs yaml update is different than this one and happens in 2 separate phases of the release process, meaning the automation steps also need to go in different automation pipelines. This one in particular should be part of finalize-release automation. I still plan to incorporate both this cycle, just haven't gotten around to it yet 😄

PR #6437, opened 2024-02-13
Add changeset conflict handler to IModelDb

khanaffan:
1. Add typescript changeset and apply conflict hander to make it easy to update it.
2. This PR also fixes a reported issue where indirect change was causing data conflict.

### DIFF
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -11,8 +11,8 @@ import { join } from "path";
 import * as touch from "touch";
 import { IModelJsNative } from "@bentley/imodeljs-native";
 import {
-  AccessToken, assert, BeEvent, BentleyStatus, ChangeSetStatus, DbResult, Guid, GuidString, Id64, Id64Arg, Id64Array, Id64Set, Id64String,
-  IModelStatus, JsonUtils, Logger, OpenMode, UnexpectedErrors,
+  AccessToken, assert, BeEvent, BentleyStatus, ChangeSetStatus, DbConflictCause, DbConflictResolution, DbOpcode, DbResult, Guid, GuidString, Id64, Id64Arg, Id64Array, Id64Set, Id64String,
+  IModelStatus, JsonUtils, Logger, LogLevel, OpenMode, UnexpectedErrors,
 } from "@itwin/core-bentley";
 import {
   AxisAlignedBox3d, BRepGeometryCreate, BriefcaseId, BriefcaseIdValue, CategorySelectorProps, ChangesetIdWithIndex, ChangesetIndexAndId, Code,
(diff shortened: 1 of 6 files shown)


core/bentley/src/BeSQLite.ts
```diff
@@ -26,6 +26,29 @@ export enum DbOpcode {
   Update = 23,
 }
 
+/** Cause of conflict when applying a changeset
+ * @public
+ */
+export enum DbConflictCause {
```
pmconne, line 32: These two enums are only used in the private `onChangesetConflict` method. What's the motivation for making them `@public`?

pmconne (commented):
Some tests seem warranted.

core/backend/src/IModelDb.ts
```diff
@@ -1082,6 +1094,148 @@ export abstract class IModelDb extends IModel {
...
+        // Caused by CASCADE DELETE on a foreign key, and is usually not a problem.
+        return DbConflictResolution.Skip;
+      }
+
+      if (args.opcode === DbOpcode.Update && args.tableName.startsWith("ec_")) {
+        // Caused by a ON DELETE SET NULL constraint on a foreign key - this is known to happen with "ec_" tables, but needs investigation if it happens otherwise
+        return DbConflictResolution.Skip;
+      }
+
+      // Refer to comment below
```
nick4598, line 1194: Can you explain this more?

If the specific change in this changeset is an Update to some row, then we skip the incoming change? Does that correspond to the point about 'briefcase with local changes' in the farther down comment?

Not sure I understand

khanaffan: No, it means there is a conflict where the parent row is missing. e.g. we update a row that has fk and fk point to a row that does not exist. then we have to decide what to do. It could be an insert/update/delete operation that can use NotFound conflict.

Generally, we will skip on not found since inserting it will later cause an FK violation as we attempt to insert/update an orphan row. Though replacing for insert will cause an FK violation later.

core/bentley/src/BeSQLite.ts
khanaffan: yes, that was a mistake. I have made them internal.
Also just made the following changes.
1. Move `onChangesetConflict()` to `BriefcaseDb` as it's the one that can apply changesets.
2. Update the existing test to make sure the `BriefcaseDb.onChangesetConflict` is called.

core/backend/src/IModelDb.ts
nick4598: So even though one of your comments says "// Caused by a ON DELETE SET NULL constraint on a foreign key - this is known to happen with "ec_" tables, but needs investigation if it happens otherwise"

You still return the same DbConflictResolution.Skip `if args.tableName.startsWith("ec_")` evaluates to false. How will this investigation take place?

Within the same larger if statement `(if args.cause === DbConflictcause.NotFound) ` There is also this comment:

"/*
       * Note: If DbConflictCause = NotFound, the primary key was not found, and returning DbConflictResolution::Replace is
       * not an option at all - this will cause a BE_SQLITE_MISUSE error.
       */"
       
There is also this comment which says returning DbConflictResolution::Replace is not an option at all, but then you return replace in `
      return args.opcode === DbOpcode.Update ? DbConflictResolution.Skip : DbConflictResolution.Replace;`
Does that comment only apply to the opcode of delete?

khanaffan: My mistake, I have to say that a better source is the following. For NotFound replace is not an option. That will make it insert so sqlite will return Misuse which is correct. But NotFound is only ever going to happen for Update/Delete. So i guess though check is there it will never happen.

I copied the conflict resolver from the native side as is. I didn't want to make any changes to it. 

https://www.sqlite.org/session/sqlite3changeset_apply.html

khanaffan: So after adding test i have update the conflict handler on `NotFound` the only thing we can do is `Skip`

ColinKerr (approved):

pmconne (approved):

nick4598:

PR #6438, opened 2024-02-13
Invalidate PRs on Change to core/backend/package.json

DanRod1999:
Invalidate PR builds when core-backend package.json is changed. This is so existing PRs are forced to rebuild against master when a new version of imodeljs-native is integrated into master branch. Currently this will also invalidate all PRs when the new nightly version of itwinjs-core is released, but our current 3hr validation rules would also invalidate these PRs left over night anyway.


DanRod1999:
> do you have an example manual run of this?

[Here](https://github.com/iTwin/itwinjs-core/actions/runs/7889237380/job/21528623754) is the last test I ran manually, but only on a single PR

My code was the exact same, but I used an if stmt to only add the status check on one of my own PRs like this:
`for (let i = 0; i < pull_requests.data.length; i++) {
  console.log(pull_requests.data[i].head.ref)
  let pr_sha = pull_requests.data[i].head.sha;

  if(pr_sha === "my sha")
  await octokit.request('POST /repos/{owner}/{repo}/statuses/{sha}', {
    ...
    }
  })
}`
I did not run it on all open PRs because I didn't want to invalidate them for no reason. I can trigger it tonight to test, and avoid messing up anyone currently trying to get stuff merged?

.github/workflows/automation-scripts/invalidate-status-checks.mjs
```diff
@@ -0,0 +1,37 @@
...
+import { Octokit, App } from "octokit"
+import dotenv from 'dotenv'
+
+dotenv.config();
+
+const octokit = new Octokit({
+  auth: `${process.env.GITHUB_TOKEN}`
+});
+
+let pull_requests = await octokit.request('GET /repos/{owner}/{repo}/pulls', {
```
aruniverse, line 14: does this get draft prs too? we shouldnt care about those

DanRod1999: According to the [docs](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests), the only options are open, closed, and all. Drafts are considered opened. I didn't realize this before, so I'll skip drafts.

DanRod1999: fixed on line 26

.github/workflows/invalidate-open-prs.yaml
```diff
@@ -0,0 +1,36 @@
...
+          git config --local user.email 38288322+imodeljs-admin@users.noreply.github.com
+          git config --local user.name imodeljs-admin
+      - name: Run invalidate-status-checks.mjs
+        run: |
+          npm install octokit
+          npm install dotenv
+          node ./.github/workflows/automation-scripts/invalidate-status-checks.mjs
+        env:
+          GITHUB_TOKEN: ${{ secrets.IMJS_ADMIN_GH_TOKEN }}
+        # turn off the default behavior to fail fast. The script has its own error handling
```
nick4598, line 35: What happened when you just did shell: bash ? Or did you go straight to bash {0}? I don't understand the fail fast concern / error handling comment.

DanRod1999: This was in a previous workflow, so when I copied the format I brought it over without thinking. Shouldn't be necessary for this workflow as I believe its for helping with scripts where the workflow might error out with non 0 exit codes

nick4598 (approved):

aruniverse (approved):

PR #6440, opened 2024-02-13
Update to TypeScript 5.3.3

wgoehrig:
WIP


pmconne:
Does upgrading to TypeScript 5.3.3 necessarily also involve upgrading the eslint plugin? Most of these changes appear related to changes to lint rules, not compiler.

anmolshres98:
> Does upgrading to TypeScript 5.3.3 necessarily also involve upgrading the eslint plugin? Most of these changes appear related to changes to lint rules, not compiler.

For some more context: https://github.com/iTwin/eslint-plugin/issues/49

wgoehrig:
Also worth noting that TypeScript updates almost always include a lot of lint fixes, even if no eslint plugin update is required, because most new TS releases these days significantly improve type inferencing, which allows the same typescript-eslint rules to find more violations.

You can see an example of that in https://github.com/iTwin/itwinjs-core/commit/7b693b7a87d0ed2e54377877612a7921fb26655c - those are additional no-unnecessary-type-assertion failures that only appear with TypeScript 5.4.0-beta.

common/api/appui-abstract.api.md
```diff
@@ -1224,9 +1224,9 @@ export enum StandardTypeNames {
     // (undocumented)
     Bool = "bool",
     // (undocumented)
-    Boolean = "boolean",
+    Boolean = "boolean",// locale specific
```
pmconne, line 1227: What are these "locale specific" comments about?

pmconne: I guess just changes to api-extractor.

wgoehrig: yep

pmconne (approved):
I reviewed the .api.md files. Reviewing all 240 changed files seems infeasible.

wgoehrig:
> I reviewed the .api.md files. Reviewing all 240 changed files seems infeasible.

@pmconne FWIW, most of the lint fixes were trivial no-unnecessary-type-assertions fixes, so looking at these changes in isolation should cut out most of the noise: https://github.com/iTwin/itwinjs-core/pull/6440/files/971c618dfbd9792f9b53955bed45358ec7234cdb..706863714949a6b4da70747f57fab8910a2dbeb1

tools/build/package.json
```diff
@@ -39,15 +39,15 @@
     "mocha-junit-reporter": "^2.0.2",
     "rimraf": "^3.0.2",
     "tree-kill": "^1.2.2",
-    "typedoc": "^0.23.28",
+    "typedoc": "^0.25.8",
```
aruniverse, line 42: fyi @ben-polinsky

aruniverse (approved):

dassaf4 (approved):
core-geometry looks good

mdastous-bentley (approved):
map-layers looks good.

saeeedtorabi (approved):

grigasp (approved):
reviewed changes in presentation code

paulius-valiunas:
I think [Stable Support `resolution-mode` in Import Types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html#stable-support-resolution-mode-in-import-types) is pretty useful too. This should allow us to use ESM-only dependencies while still being able to compile to CJS. I was able to do it before this too, but it was a hacky mess and type resolution was still kinda broken.

GerardasB (approved):

tools/build/package.json
wgoehrig: docs validation is now passing.  thanks @ben-polinsky!!

calebmshafer (approved):

PR #6442, opened 2024-02-15
Fix dev container & bump node and debian versions

anmolshres98:
VS Code dev container working fix and bump debian to 12 (current stable) version and node to 20.x in Dockerfile


.devcontainer/devcontainer.json
```diff
@@ -1,14 +1,17 @@
 // See https://aka.ms/vscode-remote/devcontainer.json for format details.
 {
   "name": "itwinjs-core-linux",
-  "dockerFile": "./Dockerfile",
-  "context": ".",
```
aruniverse, line 5: what happened to context, workspaceMount and workspaceFolder?

anmolshres98: Don't think those are needed. The dev container is working fine without them set. zsh starts off at `/workspaces/itwinjs-core` correctly

aruniverse (approved):

PR #6450, opened 2024-02-20
Update tests to release Units v01.00.08 schema

akshay05n:
Units v01.00.08 schema is about to be released. To ensure that PR build pipelines can work with the latest schema, some tests and test data need to be modified. This PR introduces the required test changes.

imodel-native: https://github.com/iTwin/imodel-native/pull/627


naveedkhan8067:
I think you have to commit the change descriptions as well.

core/ecschema-editing/package.json
```diff
@@ -35,7 +35,7 @@
     "url": "http://www.bentley.com"
   },
   "devDependencies": {
-    "@bentley/units-schema": "^1.0.7",
+    "@bentley/units-schema": "^1.0.8",
```
naveedkhan8067, line 38: `rush install` might also fail in the build. You can check it locally and may have to do `rush update` to fix it.

akshay05n: The devDependencies are working fine as we've published the `Units v1.0.8` package. Also, I've updated the change description using `rush change`.

naveedkhan8067: yes, I can see now `pnpm-lock.yaml` has updated so the build should be fine.

naveedkhan8067 (approved):

PR #6455, opened 2024-02-21
Ignore 409 errors when initializing tile storage

paulius-valiunas:
A race condition can occur when multiple people try to open the same iModel at the same time. Since `@itwin/object-storage` doesn't have good error handling, this PR simply ignores 409 errors that come from Azure blob storage.

### DIFF
--- a/core/backend/src/TileStorage.ts
+++ b/core/backend/src/TileStorage.ts
@@ -47,6 +47,8 @@ export class TileStorage {
       try {
         await this.storage.createBaseDirectory({ baseDirectory: iModelId });
       } catch (e: any) {
+        // Ignore 409 errors. This is what Azure blob storage returns when the container already exists.
+        // Usually this means multiple backends tried to initialize tile storage at the same time.
         if(e.statusCode !== 409)
           throw e;
       }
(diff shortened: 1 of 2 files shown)


core/backend/src/TileStorage.ts
```diff
@@ -44,7 +44,12 @@ export class TileStorage {
     if (this._initializedIModels.has(iModelId))
       return;
     if (!(await this.storage.baseDirectoryExists({ baseDirectory: iModelId }))) {
-      await this.storage.createBaseDirectory({ baseDirectory: iModelId });
+      try {
+        await this.storage.createBaseDirectory({ baseDirectory: iModelId });
+      } catch (e: any) {
+        if(e.statusCode !== 409)
```
nick4598, line 50: Is it worth documenting what a 409 is and why it might occur?

pmconne: Yes.

paulius-valiunas: It's a standard HTTP status code, but I added a comment explaining why this happens

pmconne (approved):

paulius-valiunas:

PR #6463, opened 2024-02-23
Change gpuMemoryLimit setting used when undefined

DStradley:
For non-mobile devices, the gpuMemoryLimit used a setting of "none" (no limit) when it was not defined.  This changes that setting to TileAdmin.nonMobileUndefinedGpuMemoryLimit (6,000 MB) instead.

Some very large reality data sets and imodels could on occasion cause a WebGL context loss with no limit imposed, so this attempts to set a generous limit that should prevent most of these problems for the common case of the gpuMemoryLimit being undefined.  If desired, the limit can still be set to "none".

See https://github.com/iTwin/itwinjs-backlog/issues/1047

### DIFF
--- a/core/frontend/src/tile/TileAdmin.ts
+++ b/core/frontend/src/tile/TileAdmin.ts
@@ -249,11 +249,8 @@ export class TileAdmin {
     else
       gpuMemoryLimit = gpuMemoryLimits;
 
-    if (undefined === gpuMemoryLimit && isMobile)
-      gpuMemoryLimit = "default";
-
-    if (undefined === gpuMemoryLimit && !isMobile)
-      gpuMemoryLimit = TileAdmin.nonMobileUndefinedGpuMemoryLimit;
+    if (undefined === gpuMemoryLimit)
+      gpuMemoryLimit = isMobile ? "default" : TileAdmin.nonMobileUndefinedGpuMemoryLimit;
 
     if (undefined !== gpuMemoryLimit)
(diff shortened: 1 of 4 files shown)


pmconne (changes requested):
Documentation on line 95 is now a lie.
`  /** Limits applied to clients running on non-mobile devices. Defaults to "none" if undefined. */`

core/frontend/src/tile/TileAdmin.ts
```diff
@@ -1256,6 +1259,8 @@ export namespace TileAdmin { // eslint-disable-line no-redeclare
     relaxed: 2.5 * 1024 * 1024 * 1024, // 2.5 GB
   };
 
+  export const nonMobileUndefinedGpuMemoryLimit = 6000 * 1024 * 1024; // 6,000 MB - used when nonMobile limit is undefined
```
pmconne, line 1262: This is implicitly `@public`. If you didn't intend that, add `@internal`. If you did intend that, give it a better name and document it.

pmconne: Fixed to use `/** doc comment */`.
You'll need to `npm run build` core-frontend and then `npm run extract-api`.

pmconne (approved):

ben-polinsky (approved):

markschlosseratbentley:
@aruniverse can we backport this to 4.4?
cc @danieliborra

aruniverse:

PR #6467, opened 2024-02-26
Tile Caching Using IndexedDB

andremig-bentley:
WIP

Cache Tiles using IndexedDB. Create a new Tile Request Channel which connects to an IDBDatabase. Through the channel, search for tiles in the db. If a match is found, check if the tile has expired (based on time of storage). If expired, delete the tile from the db and move to the next request channel. If no match is found, move to the next request channel. Once a tile has been successfully requested, store it in the db.

### DIFF
--- a/core/frontend/src/tile/IModelTileRequestChannels.ts
+++ b/core/frontend/src/tile/IModelTileRequestChannels.ts
@@ -30,6 +30,160 @@ class CloudStorageCacheChannel extends TileRequestChannel {
   }
 }
 
+class LocalStorageCacheChannel extends TileRequestChannel {
+
+  private _db: any;
+
+  public constructor(cacheConcurrency: number) {
+    console.log("CREATING LOCALSTORAGECACHECHANNEL");
+    super("itwinjs-local-cache", cacheConcurrency);
+
+    // Now set up IndexedDB
+    const requestDB = window.indexedDB.open("IDB", 1);
+
+    requestDB.onerror = () => {
+      console.log("Error opening up IDBL");
+    };
+
+    requestDB.onsuccess = (event) => {
+      console.log("Success opening up IDB");
+
+      const target: any = event.target;
+      if (target) {
+        this._db = target.result;
+        console.log(this._db);
+      }
+    };
+
+    // This will get called when a new version is needed - including going from no database to first version
+    // So this is how we set up the specifics of the db structure
+    requestDB.onupgradeneeded = (event) => {
+      console.log("ON UPGRADE NEEDED");
+      const target: any = event.target;
+
+      if (target)
+        this._db = target.result;
+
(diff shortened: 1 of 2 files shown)


andremig-bentley:
Things still needing to be confirmed/investigated further:

Opening/Closing the DB - when/where should the db be opened? As of now, it's opened as a part of the constructor for the local storage request channel. Should that be changed to part of the IModelTileMetadataCacheChannel, or put somewhere else entirely? When should we close/clear the db? Maybe based on a total size of the stored content? As of now, there is no logic implemented for closing or clearing.

Testing the content storage and return - is the tile content actually being successfully stored in and returned from the db, saving from regenerating the tile? Is there a better way to obtain a successfully requested tile for storage besides using the rpc channel's content callback to store the tile in the db?

Built-in browser caching - Is the browser caching any of these tiles already, meaning we are doing unnecessary work by adding them to a second cache? Should we somehow check to see what the browser has cached and avoid caching those tiles in the db?

TileAdmin toggle flag - Eventually, we will most likely want this functionality to be on by default. But for now, should there be a flag similar to TileAdmin.Props.cacheTileMetadata to toggle this functionality?

pmconne:
When I asked, you told me you are caching all kinds of tiles, but this only caches iModel tiles. Did you change your mind?

danieliborra:
> > When I asked, you told me you are caching all kinds of tiles, but this only caches iModel tiles. Did you change your mind?
> 
> My original thought was that all types of tiles were going through these channels, but seeing as these are IModelTileRequestChannels, that seems sort of obviously wrong in retrospect... My bad. I believe the final goal is to cache all types of tiles, so that is something still to do.

As a first step, caching iModel tiles is enough, and we can focus on the rest of tile types later.

pmconne:
How do you plan to measure the effectiveness of these changes?

danieliborra:
For now, I'd just check if tiles are being correctly stored, even with SaS tokens. That can be checked manually in the debug console.
The end game is to use this with "incremental tiles", in which we re-use old tiles. There is where the big wins will come. Until then, this is more to test the viability of the approach than for having quick performance wins (if local cache doesn't work, incremental tiles don't make that much sense)

pmconne:
Sure, I just don't want us to enable it for users until it's providing actual value.

pmconne:
The tile request channels you are modifying are used for tiles that are produced on demand by the imodel backend.
Tiles produced by mesh export service go through a [different channel](https://github.com/iTwin/itwinjs-core/blob/74610de97a26fc395842b34adf63dbfd7c9432e4/extensions/frontend-tiles/src/BatchedTile.ts#L129).

danieliborra:
Thanks for mentioning it!
I didn't expect us to activate this now either.
Does it make sense to integrate this functionality behind an option disabled by default?

PR #6468, opened 2024-02-27
imodeljs-native@4.5.4

DanRod1999:


nick4598 (approved):

core/backend/src/test/standalone/ChangeConflictHandler.test.ts
```diff
@@ -0,0 +1,369 @@
...
+        expect(spy.alwaysReturned(DbConflictResolution.Skip)).true;
+        const arg = spy.args[0][0];
+        expect(arg.cause).eq(DbConflictCause.NotFound);
+        expect(arg.opcode).eq(DbOpcode.Delete);
+        expect(arg.indirect).true;
+        expect(arg.tableName).eq("bis_GeometricElement3d");
+      },
+    );
+  });
+  it.skip("DbConflictCause.Constraint - duplicate userLabels", async () => {
```
pmconne, line 325: @khanaffan I failed to notice in the original PR - why is this test skipped?

aruniverse (approved):
approving changes for mobile

core/backend/src/test/standalone/ChangeConflictHandler.test.ts
khanaffan: Yes, that needs to be skipped as I did not find any solution when a constraint violation happens. When it happens the saveChanges/abandonChanges() both fail. I did not dig into it but we technically cannot close the db.

I will write a issue for it as it need more investigation.

khanaffan: Also, this is not a known issue. My effort was to test all cases in a conflict handler and so this test simulated a case. Its not any new issue its just not covered by any test up till now.

wgoehrig (approved):

PR #6472, opened 2024-02-27
Presentation: Add functions that return async iterators for paged responses

### LINKED ISSUE #6314: Presentation: Stream result sets as the requests are fulfilled
At the moment `PresentationManager` on the frontend, in case of a large result set, requests data in batches in a serial manner, combines everything and only then returns the result. We should provide APIs that:
- request batched results in parallel,
- stream the results as we receive them from the backend.

See `HiliteSetProvider.getHiliteSetIterator` for an example.

Yato333:
Closes #6314
- Add `StreamedResponseGenerator` class which sends paged requests and returns a response containing an async iterator. It  uses RXJS internally to fetch pages in parallel.
- Use this class for all existing API functions that send paged requests.
- Deprecate all existing functions in `PresentationManager`, that return responses for paged requests as arrays and add alternatives that return async iterators.


presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
```diff
@@ -0,0 +1,144 @@
...
+          batchSize = receivedItemsLength;
+        }
+
+        const remainingBatches = Math.ceil(itemsToFetch / batchSize);
+
+        // Return the first page and then stream the remaining ones.
+        return concat(
+          of(response.items),
+          range(1, remainingBatches).pipe(
+            mergeMap(async (idx) => {
```
grigasp, line 123: I believe `mergeMap` may return results in order different than the requests are made, e.g. response for request 3 may return earlier than for request 2, and in that case the 3rd batch is going to be emitted before the 2nd. Need a test for this.

presentation/frontend/src/test/StreamedResponseGenerator.test.ts
```diff
@@ -0,0 +1,183 @@
...
+        return {
+          total: items.length,
+          items: items.slice(page.start, page.start + 10),
+        };
+      },
+    };
+
+    const generator = new StreamedResponseGenerator(props);
+    await generator.getItems();
+  }).timeout(100);
```
grigasp, line 25: This may fail sporadically, needs to be rewritten in a way that doesn't rely on timeouts. Suggestion:

1. make `getBatch` a `sinon.fake` that returns `ResolvablePromise` of a batch.
2. call `generator.getItems()`, don't await.
3. confirm the fake created in step 1 has been called expected number of times.
4. resolve all resolvable promises.
5. await on `getItems()` result, verify it.

grigasp (changes requested):

presentation/frontend/src/presentation-frontend/PresentationManager.ts
```diff
@@ -607,8 +670,18 @@ export class PresentationManager implements IDisposable {
...
+
+    await generator.fetchFirstBatch();
+    const iterator = generator.itemsIterator;
+    return {
+      total: generator.total,
+      async *items() {
+        for await (const item of iterator) {
+          yield item;
+        }
+      },
```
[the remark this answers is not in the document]
Yato333: This function returns an `AsyncGenerator` which is almost the exact same thing as `AsyncIterableIterator` but they are totally incompatible 😄

grigasp (commented):
Also, could we have async iterator versions of the following, too?
- `getDisplayLabelDefinitionsIterator` - should be simple.
- `getContentIterator` - could return a type `{ descriptor: Descriptor, total: number, items: AsyncIterableIterator<Item> }`.

common/api/presentation-frontend.api.md
```diff
@@ -277,6 +282,9 @@ export interface OfflineCachingFavoritePropertiesStorageProps {
     impl: IFavoritePropertiesStorage;
 }
 
+// @public
+export type PagedRequestOptions<TOptions extends {}> = Paged<TOptions> & StreamingOptions;
```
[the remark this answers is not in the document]
Yato333: Removed StreamingOptions

presentation/frontend/src/presentation-frontend/PresentationManager.ts
```diff
@@ -86,6 +87,32 @@ export interface IModelContentChangeEventArgs {
   imodelKey: string;
 }
 
+/**
+ * Options for requests that can return a page of a certain size.
+ * @public
+ */
+export type PagedRequestOptions<TOptions extends {}> = Paged<TOptions> & StreamingOptions;
```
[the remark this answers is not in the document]
Yato333: Did as you've recommended

presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
```diff
@@ -0,0 +1,144 @@
...
+import { eachValueFrom } from "rxjs-for-await";
+import { Paged, PagedResponse, PageOptions } from "@itwin/presentation-common";
+
+/**
+ * Options for requests that send smaller requests in batches.
+ */
+export interface StreamingOptions {
+  /** Number of parallel requests. Default: unlimited. */
+  parallelism?: number;
+}
```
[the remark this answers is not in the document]
Yato333: Made it as you've recommended

presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
```diff
@@ -0,0 +1,144 @@
...
+  public get total(): number {
+    return this._total;
+  }
+
+  /**
+   * Fetches and caches the first batch.
+   * This function can be called in order to retrieve the total items count.
+   * @returns response for the first page.
+   */
+  public async fetchFirstBatch(): Promise<PagedResponse<TPagedResponseItem>> {
```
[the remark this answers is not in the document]
Yato333: Thanks, that seems cleaner

presentation/frontend/src/test/StreamedResponseGenerator.test.ts
Yato333: Make sense, done

presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
Yato333: That's right. Added the logic for aggregating the results in order and an a test

presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
```diff
@@ -0,0 +1,144 @@
...
+  public get itemsIterator(): AsyncIterableIterator<TPagedResponseItem> {
+    return eachValueFrom(this.items);
+  }
+
+  /**
+   * RXJS observable of items.
+   * Items count will be limited to the page size, if one is specified in the configuration.
+   */
+  public get items(): Observable<TPagedResponseItem> {
+    return this._batches.pipe(mergeAll());
```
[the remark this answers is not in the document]
Yato333: That's right, fixed it

presentation/frontend/src/test/StreamedResponseGenerator.test.ts
```diff
@@ -0,0 +1,225 @@
+import { StreamedResponseGenerator, StreamedResponseGeneratorProps } from "../presentation-frontend/StreamedResponseGenerator";
```
grigasp, line 1: missing file header

presentation/frontend/src/test/StreamedResponseGenerator.test.ts
```diff
@@ -0,0 +1,225 @@
...
+import { ResolvablePromise } from "@itwin/presentation-common/lib/cjs/test";
+
+describe("StreamedResponseGenerator", () => {
+  it("should run requests concurrently", async () => {
+    const total = 10;
+    const firstBatchPromise = new ResolvablePromise<PagedResponse<number>>();
+    const restBatchesPromise = new ResolvablePromise<PagedResponse<number>>();
+    const fakeGetBatch = sinon.fake(async (_, idx: number) => {
+      const result = idx ? restBatchesPromise : firstBatchPromise;
+      return result as unknown as Promise<PagedResponse<number>>;
```
grigasp, line 14: is `as unknown as Promise<PagedResponse<number>>` necessary?

presentation/frontend/src/test/StreamedResponseGenerator.test.ts
```diff
@@ -0,0 +1,225 @@
...
+
+      const itemsPromise = generator.createItemsResponse();
+      for (const idx of ordering) {
+        await fakePromises[idx].resolve({ total, items: [idx + 1] });
+      }
+
+      const { items } = await itemsPromise;
+      expect(items).to.deep.eq([...new Array(total).keys()]);
+    }
+  });
```
grigasp, line 68: I don't think we need to check every possible combination of promise resolution order. For the same of simplicity, I'd:
- Rename the test to `returns values in correct order when requests resolve when requests resolve in different order than being made`.
- Use 2 resolvable promises for 2nd and 3rd pages, resolve 3rd first and then 2nd.
- Check the result.

presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
```diff
@@ -0,0 +1,156 @@
...
+            if (idx - 1 !== lastEmitted) {
+              accumulatedBatches.insert(value);
+              return { lastEmitted, accumulatedBatches, itemsToEmit: [] };
+            }
+
+            // Collect all batches to emit in order.
+            lastEmitted = idx;
+            const batchesToEmit = [value];
+            for (const batch of accumulatedBatches) {
+              if (batch.idx - 1 === lastEmitted) {
```
grigasp, line 125: Based on the fact that `accumulatedBatches` is sorted by batch index, I think we could `break` on `else`, no?

presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * This class allows loading values in multiple parallel batches and return them either as an array or an async iterator.
+ * Pages are prefetched in advanced according to the `parallelism` argument.
+ * @internal
+ */
+export class StreamedResponseGenerator<TPagedResponseItem> {
+  constructor(private readonly _props: StreamedResponseGeneratorProps<TPagedResponseItem>) {}
+
+  /** Creates a response with the total item count and an array of items for the requested page. */
+  public async createItemsResponse(): Promise<{ total: number; items: TPagedResponseItem[] }> {
```
grigasp, line 28: Looks like this is now used only in tests. Maybe the function should be moved there?

presentation/frontend/src/presentation-frontend/PresentationManager.ts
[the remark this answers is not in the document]
grigasp: Let's return `AsyncIterator<T>` in all new APIs. I feel like we should've done that for the ones that currently return `AsyncGenerator` or `AsyncIterableIterator`, but will have to wait for a major release to fix that.

grigasp (commented):
- Let's deprecate the functions that return arrays and you added iterator replacements for.
- I think this change is worth mentioning in `NextVersion.md`

presentation/frontend/src/test/StreamedResponseGenerator.test.ts
Yato333:
```
Type 'ResolvablePromise<PagedResponse<number>>' is missing the following properties from type 'Promise<PagedResponse<number>>': catch, finally, [Symbol.toStringTag]
```

presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
Yato333: I figured, this can be useful. I didn't use this method in `PresentationManager` so that I wouldn't have to write repeating tests for functions that return iterators.

presentation/frontend/src/test/StreamedResponseGenerator.test.ts
grigasp: Let's fix it like this: https://github.com/iTwin/presentation/blob/350aef3ec6f283c2d902096e343fc6a52a14a464/packages/components/src/test/_helpers/Promises.ts#L31

presentation/frontend/src/test/StreamedResponseGenerator.test.ts
Yato333: That's not enough. Apparently the edge case when there is a difference of more than 1 between batch indices is left uncovered (`batch.idx - 1 !== lastEmitted`). I still have to use 3 promises, but I'll only leave 2 orderings, which will cover all the edge cases.

Yato333:
All the changes in the full-stack-tests are due to the deprecations in the PresentationManager

presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
Yato333: Removed

presentation/frontend/src/presentation-frontend/StreamedResponseGenerator.ts
Yato333: That's right. I was looking for a sorted collection for the exact same point 😆

presentation/frontend/src/test/StreamedResponseGenerator.test.ts
Yato333: Done

docs/changehistory/NextVersion.md
```diff
@@ -34,3 +36,9 @@ The new [TerrainSettings.dataSource]($common) property can be used by custom [Te
 ## Electron 29 support
 
 In addition to [already supported Electron versions](../learning/SupportedPlatforms.md#electron), iTwin.js now supports [Electron 29](https://www.electronjs.org/blog/electron-29-0).
+
+## Presentation
+
+### Deprecation of array batches in favor of async iterators
+
+Some of the methods in frontend's `PresentationManager` class (e.g. `getNodes`) return arrays of items. These methods may take a lot of time to complete which prevents the consumer from continuing. For each such method, alternative methods have been added, which return async iterators. Old methods which return arrays have been deprecated.
```
grigasp, line 44: ### Deprecation of async array results in favor of async iterators

[PresentationManager]($presentation-frontend) contains a number of methods to retrieve sets of results like nodes, content, etc. All of these methods have been deprecated in favor of new ones that return an async iterator instead of an array:

- Use `getContentIterator` instead of `getContent` and `getContentAndSize`.
- Use `getDisplayLabelDefinitionsIterator` instead of `getDisplayLabelDefinitions`.
- Use `getDistinctValuesIterator` instead of `getPagedDistinctValues`.
- Use `getNodesIterator` instead of `getNodes` and `getNodesAndCount`.

All of the above methods, including the deprecated ones, received ability to load large sets of results concurrently. Previously, when requesting a large set (page size > 1000), the result was created by sending a number of requests sequentially by requesting the first page, then the second, and so on, until the whole requested set was retrieved. Now, we send a request for the first page to determine total number of items and backend's page size limit, together with the first page of results, and then other requests are made all at once. At attribute `maxParallelRequests` was added to these methods to control how many parallel requests should be sent at a time.

While performance-wise deprecated methods should be in line with the newly added async iterator ones, the latter have two advantages:

1. Caller can start iterating over results as soon as we receive the first page, instead of having to wait for the whole set.
2. The iterator version stops sending requests as soon as the caller stops iterating over the async iterator.

> **Example:** Showing display labels for a large set of elements in a React component.
> 
> The deprecated approach would be to use [PresentationManager.getDisplayLabelDefinitions]($presentation-frontend) to retrieve labels of all elements. Because the API has to load all the data before returning it to the widget, user has to wait a long time to start seeing the results. Moreover, if he decides to close the widget (the component is unmounted), the `getDisplayLabelDefinitions` keeps building the array by sending requests to the backend - there's no way to cancel that.
> 
> The new approach is to use [PresentationManager.getDisplayLabelDefinitionsIterator]($presentation-frontend) to get the labels. The > labels get streamed to the called as soon as the first results' page is retrieved, so user gets to see initial results quickly, while > additional pages keep loading in the background. In addition, if the component is unmounted, iteration can be stopped, thus cancelling > all further requests to the backend:
> 
> ```ts
> for await (const label of manager.getDisplayLabelDefinitionsIterator(requestProps).items) {
>   if (isComponentUnmounted) {
>     break;
>   }
>   // update component's model to render the loaded label
> }
> ```

presentation/frontend/src/presentation-frontend/PresentationManager.ts
```diff
@@ -580,38 +585,60 @@ export class PresentationManager implements IDisposable {
...
-      ...(await generator.createAsyncIteratorResponse()),
+      ...(await generator.createObservableResponse()),
       descriptor,
     };
   }
 
-  /** Retrieves a content descriptor, item count and async generator for the items themselves. */
-  public async getContentIterator(
+  /** @internal */
+  public async getContentObservable(
```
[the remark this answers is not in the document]
Yato333: Yes, I just don't feel like this workflow is efficient or neccessary: 
1. We convert the observable to an iterator in `PresentationManager`
2. In `HiliteSetProvider` we convert it back to observable so that we could pipe it to `shareReplay`
3. Then we convert it back to an iterator so that we could return it to the consumer.

With `getContentObservable` we could eliminate the first and second conversions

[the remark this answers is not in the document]
grigasp: Agree. What I don't like about having a public method returning an `Observable`, though, is that it adds `rxjs` to our API summary. From the API correctness standpoint that's not a problem, because the method is `@internal`, however it makes it easier to miss `rxjs` APIs being exposed through our public API (which we strictly don't want) in the future.

Maybe the method could be moved out of `PresentationManager` class as a standalone function instead?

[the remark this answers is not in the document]
Yato333: What about `ensureIModelInitialized` that we have to call before returning the result `Item`? Also, if we moved it to a function it would still have to be in `PresentationManager.ts` and marked `export`, so I don't see how it would help. I think the easiest solution would be to stick with `getContentIterator`, although I don't know how it would affect the performance.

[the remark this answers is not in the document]
grigasp: > What about `this.ensureIModelInitialized` that we have to call before returning the result?

We only need to await on it when the result contains a descriptor (in case caller wants to check if one of the descriptor  fields is favorited by a user). For hilite sets we don't care about that, if that's you're worried about.

> Also, if we moved it to a function it would still have to be in `PresentationManager.ts` and marked `export`, so I don't see how it would help.

Right... Unless the function is some other file that's not exported through the barrel :)

full-stack-tests/presentation/src/learning-snippets/content/rules/ContentModifier.test.ts
```diff
@@ -89,8 +88,10 @@ describe("Learning Snippets", () => {
           ])
           .and.to.have.lengthOf(6);
         const calculatedField = tryGetFieldByLabel(content!.descriptor.fields, "Calculated");
-        expect(content!.contentSet[0].displayValues[calculatedField!.name]).to.be.undefined;
-        expect(content!.contentSet[1].displayValues[calculatedField!.name]).to.eq("PREFIX_Uncategorized");
+
+        const items = [await content!.items.next(), await content!.items.next()];
```
grigasp, line 92: why not use `collect` for this?

presentation/frontend/src/presentation-frontend/PresentationManager.ts
```diff
@@ -241,6 +242,10 @@ export class PresentationManager implements IDisposable {
       this._clearEventListener = IpcApp.addListener(PresentationIpcEvents.Update, this.onUpdate);
       this._ipcRequestsHandler = props?.ipcRequestsHandler ?? new IpcRequestsHandler(this._requestsHandler.clientId);
     }
+
+    PresentationManagerExtensions.provider = {
+      getContentObservable: this.getContentObservable.bind(this),
+    };
```
grigasp, line 248: In theory you could have multiple `PresentationManager` instances set up with different attributes (active unit system, locale, etc.). In that case, all those instances would use `PresentationManagerExtensions` set up with the most recently created `PresentationManager` attributes. Also, it's not great that developers need to know that `PresentationManager` must be created before `PresentationManagerExtensions` can be used at all.

My suggestion:
1. Change `PresentationManagerExtensions` to accept implementation through constructor.
2. Change `PresentationManagerExtensions.getContentObservable` to be non static.
3. Create `PresentationManagerExtensions` object as a member when creating a `PresentationManager`. Add an internal public getter for it.
4. `HiliteSetProvider` can then use it like `Presentation.presentation.internal.getContentObservable(...)`.
5. I'd also rename `PresentationManagerExtensions` to `PresentationManagerInternal` and `PresentationManagerProvider` to `PresentationManagerInternalImpl`.

full-stack-tests/presentation/src/learning-snippets/content/rules/ContentModifier.test.ts
Yato333: Makes sense

presentation/frontend/src/presentation-frontend/PresentationManager.ts
Yato333: Made it so that HiliteSetProvider uses an iterator as we've discussed

grigasp (approved):

aruniverse (approved):

PR #6473, opened 2024-02-27
Fix RPC ignoring errors thrown in native side

GytisCepk:
### Changes:
- Fix correct HTTP status code not being set when error is thrown in the native side. This was due native part throwing `Error` and not `BentleyError` which caused `errorNumber` property not being checked. Since native side can't throw custom errors (at least to my knowledge), we need to check if `errorNumber` exists on provided `Error` not just `BentleyError`.
- Introduced new `IModelStatus.Aborted` status. For now it will only be used when snap requests are aborted.
- Improve logging. Previously HTTP status code was logged too early and didn't match actual code return by backend in some cases.


core/bentley/src/StatusCategory.ts
```diff
@@ -133,8 +137,10 @@ class UnknownError extends HTTP.InternalServerError { }
 
 class NotImplemented extends HTTP.NotImplemented { }
 
-function lookupCategory(error: BentleyError): StatusCategory {
-  switch (error.errorNumber) {
+class Aborted extends HTTP.BadRequest { }
```
GytisCepk, line 140: Most accurate HTTP status code for `Aborted` errors would be 204, but it would require a breaking change on `IModelReadRpcInterface`, so for now we treat `Aborted` as HTTP 400 "Bad Request".

GitHub issue will be open once this PR is merged to switch to 204 for iTwin.js 5.0.

paulius-valiunas (approved):

aruniverse (approved):

core/bentley/src/StatusCategory.ts
aruniverse: https://github.com/iTwin/itwinjs-core/issues/6475

PR #6476, opened 2024-02-28
Address memory leak regarding markers and decorators

hl662:
When ImodelApp.shutdown() is called, ViewManager's onShutdown() is called. Problem is, the  viewports that are a part of IModelApp do not have their dispose() functions called, which holds up WebGL resources. This results in a memory leak even after shutdown() is called: thousands of featureAppearance overrides are being retained every refresh, which are coming from ScreenViewport's OnScreenTarget.

Solution is to drop all viewports in IModelApp's viewManager when shutdown is called.

### DIFF
--- a/common/changes/@itwin/core-frontend/nam-mem-leak-resolution_2024-02-28-21-45.json
+++ b/common/changes/@itwin/core-frontend/nam-mem-leak-resolution_2024-02-28-21-45.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Dispose of all viewports when IModelApp.shutdown() is called to clear up WebGL resources, and avoid memory leak",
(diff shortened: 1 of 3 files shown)


hl662:
@pmconne @markschlosseratbentley @Maxime-Brassard can you check this? This is a problem Maxime found after he put IModelApp.shutdown() into WaterSight after the meeting. The viewport still remained, with thousands of decorators remaining. 
To note, this is different from the 2 problems we discussed in the meeting.

markschlosseratbentley (approved):
Approving, but I agree with Ben's comment about a unit test.

hl662:
> Approving, but I agree with Ben's comment about a unit test.

Agreed, just added one unit test, had to accommodate for errors when iterating through already disposed viewports

ben-polinsky (approved):

aruniverse:

PR #6481, opened 2024-03-01
Fix waitForSceneCompletion hang when tiles have been freed by GpuMemoryLimit

DStradley:
Fixes some issues where waitForSceneCompletion never exits the loop because _hasMissingTiles is always true.  These occur when a GpuMemoryLimit is used and has caused tiles to discard from the limit being exceeded.

There were actually three different of these issues found while investigating this.  The first (preloading) was dealt with in https://github.com/iTwin/itwinjs-core/pull/6512 while the investigation continued.

The second issue was resolved by changing LRUTileList.add to move the tile after the sentinel to close a window that could cause some tiles to get deleted before they were selected.

The third issue was resolved by preventing freeing of a realityTile if it has any AdditiveRefinementStepChildren. That case still depends on the parent tile to draw, but the parent is not selected and could be freed under pressure.

Merging after final ImageTests complete.


pmconne:
The culprit is `RealityTile.preloadProtectedTiles`.

We should disable preloading when under memory pressure. Proposed solution is to add a property like `TileAdmin.allowPreloading` that always returns false on mobile (RealityTileTree.ts checks this before preloading, it belongs here instead), and otherwise computes % utilization based on LRUTileList.totalBytesUsed / max memory limit (or 0%, if no memory limit) and returns false if it's above some threshold (80% seems reasonable).

core/frontend/src/tile/TileAdmin.ts
```diff
@@ -1170,6 +1186,24 @@ export namespace TileAdmin { // eslint-disable-line no-redeclare
      */
     disableMagnification?: boolean;
 
+    /** Allow preloading for context (reality and map tiles).
+     * This will be ignored and treated as false for mobile devices.
+     *
+     * Default value: true
+     * @alpha
+     */
+    allowPreloading?: boolean;
```
pmconne, line 1195: This separate flag seems redundant - preloading could be disabled by setting the memory threshold to 0% (if you slightly adjust your `isPreloadingAllowed` implementation to compute % mem usage as 0 if no limit defined).

DStradley: Removed `isPreloadingAllowed` flag

DStradley:
Unfortunately, while this does work correctly now for disabling preloading when under memory pressure, it does not fix the hang problem after all, or at least not completely.
If I just bring up the debugger window while display-performance-test-app is hung on a view, then that view completes.  The following view then will hang.  But if I set breakpoints and step through a little of the code and then continue, then that view works and the test finishes.

pmconne:
@DStradley is this only reproducible with reality models?

DStradley:
> @DStradley is this only reproducible with reality models?

As far as I know, it is only happening on a fairly large reality model, with shadows on, from my test suite.  However, it is not likely that many of the other models are under memory pressure (which is key here, I was a little surprised to find this one was), so I am not sure.

> It would be surprising if the tile tree asks us to load content for a tile, but after the content loads the tile tree does not select it for display.

If am understanding the flow I am seeing, which I may not be, it wants to select it but doesn't get the chance to before it is freed, at least for this reality data case I am studying.

pmconne:
> it is not likely that many of the other models are under memory pressure (which is key here

Have you performed the experiment of setting the memory threshold very low and seeing if the problem occurs with other views, and if so, if they all have something in common, such as containing only reality models, or only shadows, etc?

DStradley:
> Have you performed the experiment of setting the memory threshold very low and seeing if the problem occurs with other views...

No I have not, I will try that.

Also, on the view it is failing on currently, there are two views drawn back to back that are the same view, except the first is "masked" and the second is "without mask".  Either view run by itself will succeed, but back to back, the second one hangs.  So it is a complex failure case, though repeatable.  And if I remove the GPUmemoryLimit then of course the pair succeeds.

DStradley:
Initial test results with a lower memory threshold:  I set the limit to 1000 MiB and ran the full test suite.  This produced a lot more hangs, all reality models.  35 hung views now across 9 different models, only 4 of which have shadows.  So the commonality so far seems to be reality models.

During the run with 35 hangs, I was able to manually keep the test going through these techniques, the middle one is the only one I think I understand why it worked:
+ Just bringing up the debugger (Ctrl-Shift-I) will clear the hang a lot of times, but not always.
+ Otherwise, I set a breakpoint in `disposeContents` and after `this._state` gets set to NotReady, I set it back to Ready (3) in the debugger, dismiss the debugger to continue, and that clears the hang.
+ Near the end I accidentally discovered that collapsing DPTA and restoring it would also clear the hang, at least for the few times I tried it.

I then launched a new run with a 100 MiB limit and PlanarClipMasks, RealityTests, and Thematic-Reality groups turned off.  This is to see if this low a limit shows up anything else (already tested reality models, a few more hangs there doesn't tell us anything).
I am only 20 views into the test and have already had 5 different views from 4 different imodels hang on this run (all plant models).  For some reason, I can't get the debugger to come up on this run, but collapsing the window and restoring it will free the hang.  I am going to kill this run, since I don't have time now to babysit all the hangs.

But apparently it is not just reality models, if the limit is set low enough.

DStradley:
Okay, I had another imodel hang (a building this time) before I could get the last test killed, so that made 6 hangs in a fairly short time, of all imodel tiles.  So I didn't see the point of continuing that test.  Instead, I tried moving processQueue to the end of process to see if that would fix these.  I first just ran that locally with the 6 known failing (after verifying they also failed locally).

However, what I found was that this would not reliably fix the problem.  Most would not hang now, but one of the files (varying, apparently random after 4 tries) in the test would still hang.  When it did, I found I could still do my scheme of forcing Ready back on in the debugger to clear the hang.  Also, it was just one tile that was "in a loop", just like reality tiles.

Not sure yet what to try next.

DStradley:
@pmconne I have been adding various debug prints in to attempt to figure out the flow of the problem cases.
I added a probtile field to tile and initialized it to 0.  I set it to 1 in renderFrame if missingTiles.size === 1, tile is not in use, and probtile is 0.  I set it to 0 again in markUsed.  Also, I increment it (if > 0) in freeMemory and go ahead and disposeContents.
All of this was an effort to narrow the scope of the problem tiles, and prevent a flood of prints.  When things work properly, probTile gets marked, but later markUsed clears it and everything is fine, otherwise, a single tile just churns.
I do print TileAdmin.process begin and end every time, and at the beginning I search for any probTile in the LRUlist.  If none found, it prints a -2.

Here is a sequence of prints (the [] portion contains timestamps) from a hung case, indicating the flow.  // comments were added afterwards.  It seems this tile repeatedly is getting freed before it can be selected. What would you suggest?
```
[51808:0311/174849.564:INFO:CONSOLE(128801)] "renderFrame: 1 missing: id -b-5-0-1-4-1 probtile 67  loadstatus 0  inUse false  selected false"  // renderFrame called that had 1 missing tile
[51808:0311/174849.568:INFO:CONSOLE(130056)] "TileAdmin.process begin - probtile -2" // -2 means probtile not in LRUlist
[51808:0311/174849.568:INFO:CONSOLE(130060)] "TileAdmin.process end - probtile -2"   // -2 means probtile not in LRUlist
[51808:0311/174849.570:INFO:CONSOLE(130056)] "TileAdmin.process begin - probtile -2" // -2 means probtile not in LRUlist
[51808:0311/174849.570:INFO:CONSOLE(130060)] "TileAdmin.process end - probtile -2"   // -2 means probtile not in LRUlist
[51808:0311/174849.583:INFO:CONSOLE(110838)] "setIsReady - probTile: id -b-5-0-1-4-1  probtile 67  state 3  inUse false  selected false"
                                                // setIsReady then calls onTileContentLoaded after this print, which calls drop and add, add puts the tile in the LRUlist as a non-selected tile
[51808:0311/174849.598:INFO:CONSOLE(130056)] "TileAdmin.process begin - probtile 67" // This shows that probTile is in the LRUlist at the beggining of process
[51808:0311/174849.598:INFO:CONSOLE(110770)] "freeMemory - free probTile: id -b-5-0-1-4-1  probtile 67  state 3  inUse false  selected false"  // so it gets freed, and I increment probTile afterwards
[51808:0311/174849.599:INFO:CONSOLE(130060)] "TileAdmin.process end - probtile 67"
[51808:0311/174849.679:INFO:CONSOLE(145270)] "selectTiles: id -b-5-0-1-4-1  probtile 68  loadStatus 0  inUse false  vis true  hasGraphics false  selected false"  // then the select for this probTile happens, but now it has no content

                                             // repeat of all the above, with incremented probTile - keeps doing this forever
[51808:0311/174849.680:INFO:CONSOLE(128801)] "renderFrame: 1 missing: id -b-5-0-1-4-1 probtile 68  loadstatus 0  inUse false  selected false"
[51808:0311/174849.686:INFO:CONSOLE(130056)] "TileAdmin.process begin - probtile -2"
[51808:0311/174849.687:INFO:CONSOLE(130060)] "TileAdmin.process end - probtile -2"
[51808:0311/174849.687:INFO:CONSOLE(130056)] "TileAdmin.process begin - probtile -2"
[51808:0311/174849.687:INFO:CONSOLE(130060)] "TileAdmin.process end - probtile -2"
[51808:0311/174849.705:INFO:CONSOLE(110838)] "setIsReady - probTile: id -b-5-0-1-4-1  probtile 68  state 3  inUse false  selected false"
[51808:0311/174849.716:INFO:CONSOLE(130056)] "TileAdmin.process begin - probtile 68"
[51808:0311/174849.716:INFO:CONSOLE(110770)] "freeMemory - free probTile: id -b-5-0-1-4-1  probtile 68  state 3  inUse false  selected false"
[51808:0311/174849.716:INFO:CONSOLE(130060)] "TileAdmin.process end - probtile 68"
[51808:0311/174849.797:INFO:CONSOLE(145270)] "selectTiles: id -b-5-0-1-4-1  probtile 69  loadStatus 0  inUse false  vis true  hasGraphics false  selected false"
```

DStradley:
There appears to be a timing window that allows this case to occur where it can get freed before it is selected (under memory pressure).  setIsReady calls onTileContentLoaded, which calls _lruList.drop and _lruList.add.  The add puts in back in the lruList and calls moveBeforeSentinel, which means it can be freed if freeMemory happens to be called before selectTiles (which would end up causing it to be moved after the sentinel and therefore safe from free).

After some discussion of this in a group meeting, we decided to it should be okay to just have _lruList.add move it after the sentinel instead of before, and thus protect it.

I implemented this and tested it on a full suite of ImageTests views, but I found there were still hangs.  After digging into this, I found that it did indeed fix the problem I had been looking at, however, there is still a 3rd remaining hang problem related to models with Open Street Map buildings.

DStradley:
The flow I found this time is:

1. setIsReady calls _lruList drop and add on the tile.  (add now calls moveAfterSentinel, which makes it "safe" from freeMemory.)
2. Then, renderFrame happens and it calls clearTilesForUser, which calls _lruList.clearUsed, which calls moveBeforeSentinel, so then it is no longer "safe" from freeMemory.
3. Then, when TileAdmin.process gets called, it calls freeMemory, and a set of tiles which were needed and used, but not marked as such yet, get freed.
4. They then get put in the missing list, and so get requested again, and eventually are ready, and so setIsReady is called, and the cycle repeats.

DStradley:
While studying the code flow on all this, I noticed that part of the comment on markTileUsed says “A tile will not be discarded while it is in use by any TileUser.”  This appears to be the case for pruneAndPurge, which checks getIsTileInUse(), but it is not the case for freeMemory (which is only looking at the position in lruList with regard to sentinel). I also noticed that the tiles getting freed were considered inUse.  So I tried adding that check to freeMemory as well, and found that it fixes the OSM building hang.

I have started a full ImageTests run with a GPUMemLimit of 50 MB, to see how that goes.

@pmconne, is this a viable fix?

pmconne:
`pruneAndPurge` are the normal mechanisms for freeing up memory consumed by tiles that are no longer in use. `freeMemory` is the mechanism by which we free memory consumed by tiles that might still be in use. If you change `freeMemory` to only free memory for tiles that are not in use, you will largely defeat the whole purpose. It would be equivalent to setting the tile expiration time to zero. It would cause us to never free undisplayed ancestor tiles of displayed tiles, for example.

DStradley:
> `pruneAndPurge` is the normal mechanism by which we free up memory consumed by tiles that are no longer in use... 

So what is the definition of "in use", and how is it different than "selected" for instance?  I assume we still don't want to free tiles that are actually being displayed (or trying to be)?

pmconne:
> I assume we still don't want to free tiles that are actually being displayed (or trying to be)?

Right, that is why `freeMemory` only frees selected tiles (based on its partition in the LRU list).

> So what is the definition of "in use", and how is it different than "selected" for instance? 

See [TileDrawArgs.markUsed](https://github.com/iTwin/itwinjs-core/blob/ef99f23573f5f0edce5fb7b08906e4d94d25cf65/core/frontend/src/tile/TileDrawArgs.ts#L410):

> Indicate that the specified tile is being used for some purpose by the [[SceneContext]]'s [[Viewport]]. Typically "used" means "displayed", but the exact meaning is up to the [[TileTree]] - for example, "used" might also mean that the tile's children are being used. A tile that is "in use" by any [[Viewport]] will not be discarded.

DStradley:
Updates:
 * I made a test case that only has OSM buildings (no reality data or map tiles) and that still fails the same way.
* I noticed AdditiveRefinementStepchildren are in play for some of these tiles, so I tried forcing that disabled in the code and that "fixes" the hang.

DStradley:
@pmconne as I think I currently understand things, AdditiveRefinementStepChildren mark themselves and their loadableTile as used.  Both are needed to draw since the StepChildren only contain transforms, but only the StepChildren are getting selected, I assume because we don't want the loadableTile to draw directly.  So the actual RealityTile is getting freed even though it is needed to draw.
So my current attempt at a fix is to check and not free a RealityTile if it has any StepChildren.  This prevents hanging on my test cases, is this a viable fix?

core/frontend/src/tile/LRUTileList.ts
```diff
@@ -323,7 +323,8 @@ export class LRUTileList {
     while (prev && prev !== this._sentinel && this.totalBytesUsed > maxBytes) {
       const tile = prev as Tile;
       prev = tile.next;
-      if (!tile.usageMarker.getIsTileInUse())
+      // prevent freeing if AdditiveRefinementStepChildren are in play, as they depend on the parent tile to draw
+      if (!tile.children || !tile.children.some((child) => (child as RealityTile).isStepChild))
```
pmconne, line 327: The effect of this is that we will assume at least one of the step children is currently selected, which is not necessarily the case. Eventually the normal periodic pruning of expired tiles will clean them up, but you'll be holding them in memory for longer than they are needed.

Since I don't think we have an efficient way to check if a tile *is* in the selected partition, I'm ok with this approach for now to get this resolved. But don't do it here with the ugly cast. Make RealityTile override `freeMemory` and put the step-child check in that method.

DStradley: Makes sense, done

pmconne (approved):

markschlosseratbentley (approved):

PR #6485, opened 2024-03-02
Generalize duplicated frontend and backend code into common MobileAuhorizationClient

toddsouthenbentley:
-Add onAccessTokenChanged event which is raised in setAccessToken


core/mobile/src/common/MobileAuthorizationClient.ts
```diff
@@ -47,8 +48,11 @@ export class MobileAuthorizationBackend implements AuthorizationClient {
     }
   }
 
+  public readonly onAccessTokenChanged = new BeEvent<(token: AccessToken) => void>();
+
   public setAccessToken(accessToken?: string, expirationDate?: string) {
     this._accessToken = accessToken ?? "";
     this._expirationDate = expirationDate ? new Date(expirationDate) : undefined;
+    this.onAccessTokenChanged.raiseEvent(this._accessToken);
```
tcobbs-bentley, line 56: I could be wrong, but I don't think `setAccessToken` ever gets called (in either the frontend or the backend). If the code changes so that it _does_ get called in the backend, I suspect it won't be called in the frontend.

tcobbs-bentley: Never mind; I just missed where it gets called (both in frontend and backend).

PR #6493, opened 2024-03-06
Allow itwin.js to change busy timeout for editing applications

khanaffan:
The editing application can hang when trying to open a write connection to a shared briefcase. Currently, it takes 20 sec for it to fail iModel open call to fail with `SQLITE_BUSY`. Which hangs the application. The application can set a shorter timeout and can retry so UI is not hung.

The goal of this PR is to allow applications to specify a `busyTimeout` in milliseconds. It can be zero. It will override any other busy handler present including the default one.

imodel-native: https://github.com/iTwin/imodel-native/pull/657


core/common/src/IModel.ts
```diff
@@ -135,6 +135,15 @@ export interface IModelEncryptionProps {
   readonly password?: string;
 }
 
+/**
+ * Sqlite options.
+ * @public
+ */
+export interface OpenSqliteArgs {
+  /** For more information https://www.sqlite.org/c3ref/busy_timeout.html. */
```
pmconne, line 143: Prefer to document what this does rather than redirecting.
At a minimum, document that the units are milliseconds.

kabentley:
This change is fine, but I presume this is related to LocalSync holding the iModel open? 

@swwilson-bsi as we discussed, it should run to completion and exit, while MS waits.

pmconne (approved):

aruniverse:
fyi @DanRod1999 

plz hold off on merging this pr; dan just completed some prs for addon automation, which should hopefully kill off this branch

aruniverse (approved):

PR #6498, opened 2024-03-07
Support for interactive editing tools that were using internal apis.

bbastings:


core/frontend/src/tools/ToolAdmin.ts
```diff
@@ -1807,6 +1807,11 @@ export class ToolAdmin {
     return this.onMotion(ev.viewport!, ev.viewPoint, InputSource.Touch);
   }
 
+  /** Can be called by tools to invoke their [[InterativeTool.onDynamicFrame] method without requiring a motion event. */
```
ben-polinsky, line 1810:
```suggestion
  /** Can be called by tools to invoke their [[InterativeTool.onDynamicFrame]] method without requiring a motion event. */
```

bbastings: fixed

core/frontend/src/AccuDraw.ts
```diff
@@ -3366,6 +3371,14 @@ export class AccuDrawHintBuilder {
     return (undefined !== current ? current.inverse() : undefined);
   }
 
+  /** Return a [[Matrix3d]] from a [[SnapDetail]].
+   * Uses [[SnapDetail.normal]] and [SnapDetail.primitive]] when available to create the most well defined rotation for the given snap location.
```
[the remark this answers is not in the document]
bbastings: fixed

ben-polinsky (approved):

PR #6501, opened 2024-03-08
Bug Fix: Resolve `ECSqlReader.toRow()` crash while duplicate property names are selected

### LINKED ISSUE #6343: QueryReader toRow crashes when QueryRowFormat is UseECSqlPropertyNames and there are duplicate property names selected
The query `SELECT * FROM bis.Element c JOIN bis.Element p ON p.ECInstanceId = c.ECInstanceId LIMIT 10` will crash with the error `Cannot redefine property: Model`:

```ts
const reader = db.CreateQueryReader("SELECT * FROM bis.Element c JOIN bis.Element p ON p.ECInstanceId = c.ECInstanceId LIMIT 10", undefined, {rowFormat: QueryRowFormat.UseECSqlPropertyNames});
while (await reader.step()) {
  console.log(reader.current.toRow());
}
```

The reason for the crash is that conflicting names are not made unique like they are with the `UseJsPropertyNames` option.

To fix the issue duplicate names should have _1 .._2 etc matching the UseJsPropertyNames behavior.

Add test cases covering like property names, aliases and constants

akshay05n:
This Pull Request addresses a critical bug causing the `ECSqlReader.toRow()` method to crash when the `QueryRowFormat` is set to `UseECSqlPropertyNames` or `UseECSqlPropertyIndexes` and duplicate property names are selected.
Resolved the crash issue by updating the row formatting logic to prevent the use of the same property name multiple times.

Closes #6343

### DIFF
--- a/common/changes/@itwin/core-backend/aks-queryreader-crash-fix_2024-03-08-09-55.json
+++ b/common/changes/@itwin/core-backend/aks-queryreader-crash-fix_2024-03-08-09-55.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Added new tests for selecting duplicate property names using UseECSqlPropertyNames QueryRowFormat.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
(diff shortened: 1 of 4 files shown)


pmconne (commented):
Awaiting review from @ColinKerr or @khanaffan.

khanaffan (approved):

pmconne (approved):

PR #6506, opened 2024-03-11
Changes to Channels

diegoalexdiaz:
+ Remove hasChannels API.
+ Add IModelHost option to remove sharedChannel from allowedChannels list. In 5.0, we'll make that the default behavior.
+ Introduce queryChannelRoot API.
+ Enforce one ChannelRoot per ChannelKey in an iModel.
+ Update tests and documentation accordingly.


kabentley:
I suggest you get rid of the concept of "hasChannels" entirely and presume there's always channels. All the "legacy channel" stuff can be eliminated and just say that's all part of the (non-editable by default) shared channel. No need to propagate that at all since nothing from the old connectors should be editable by iTwin.js.

core/backend/src/IModelDb.ts
```diff
@@ -980,7 +980,7 @@ export abstract class IModelDb extends IModel {
         IModelJsFs.recursiveMkDirSync(baseDir);
         props = { ...props, tempFileBase: join(baseDir, file.path) };
       }
-      nativeDb.openIModel(file.path, openMode, upgradeOptions, props, props?.container, props);
```
kabentley, line 983: @khanaffan just added this a few days ago???

core/backend/src/test/categories/Category.test.ts
```diff
@@ -20,6 +20,7 @@ describe("Category", () => {
       projectExtents: { low: { x: -500, y: -500, z: -50 }, high: { x: 500, y: 500, z: 50 } },
       guid: Guid.createValue(),
     });
+    imodel.channels.addAllowedChannel("shared");
```
kabentley, line 23: I think you should add the shared channel to `StandaloneDb.createEmpty`. That function creates a new empty file that can only be edited in one session. There's no problem with editing the shared channel in that case. Then none of these tests will need to be changed.

core/backend/src/ChannelControl.ts
```diff
@@ -136,13 +131,10 @@ export class ChannelAdmin implements ChannelControl {
     if (ChannelControl.sharedChannelName !== this.getChannelKey(args.elementId))
       throw new Error("channels may not nest");
 
+    if (!this._bisCoreSupportsChannelRootAspect)
```
kabentley, line 134: remove.

core/backend/src/ChannelControl.ts
```diff
@@ -70,13 +53,13 @@ export namespace ChannelControl {
 export class ChannelAdmin implements ChannelControl {
   public static readonly sharedChannel = "shared";
   public static readonly channelClassName = "bis:ChannelRootAspect";
+  public static readonly subjectClassName = "bis:Subject";
```
kabentley, line 56: this is not used.

core/backend/src/test/TestChangeSetUtility.ts
```diff
@@ -52,6 +52,7 @@ export class TestChangeSetUtility {
     this.iModelId = await HubWrappers.recreateIModel({ accessToken: this._accessToken, iTwinId: this.iTwinId, iModelName: this._iModelName, noLocks: true });
 
     this._iModel = await HubWrappers.downloadAndOpenBriefcase({ accessToken: this._accessToken, iTwinId: this.iTwinId, iModelId: this.iModelId });
+    this._iModel.channels.addAllowedChannel("shared");
```
kabentley, line 55: use `ChannelControl.sharedChannelName` instead of "shared"

common/api/editor-common.api.md
```diff
@@ -19,6 +19,8 @@ import { TransformProps } from '@itwin/core-geometry';
 
 // @beta
 export interface BasicManipulationCommandIpc extends EditCommandIpc {
+    // (undocumented)
+    addAllowedChannels(channelKeys: string[]): Promise<IModelStatus>;
```
kabentley, line 23: i'm never sure where people will want their logic for initializing an editing session, since they'll always have a backend and a frontend. I suppose there may be cases where it could be on the frontend. In that case, you should add `removeAllowedChannels` too.

kabentley: I think these functions should only allow a single channel name.

core/backend/src/ChannelControl.ts
```diff
@@ -68,15 +51,13 @@ export namespace ChannelControl {
 
 /** @internal */
 export class ChannelAdmin implements ChannelControl {
-  public static readonly sharedChannel = "shared";
   public static readonly channelClassName = "bis:ChannelRootAspect";
   private _allowedChannels = new Set<ChannelKey>();
   private _allowedModels = new Set<Id64String>();
   private _deniedModels = new Map<Id64String, ChannelKey>();
-  private _hasChannels?: boolean;
+  private _channelRootAspectClassExists?: boolean;
```
kabentley, line 58: no longer needed

core/backend/src/ChannelControl.ts
```diff
@@ -86,27 +67,20 @@ export class ChannelAdmin implements ChannelControl {
...
-      return channel;
+    try {
+      const channel = this._iModel.withPreparedStatement(`SELECT Owner FROM ${ChannelAdmin.channelClassName} WHERE Element.Id=?`, (stmt) => {
+        stmt.bindId(1, elementId);
+        return DbResult.BE_SQLITE_ROW === stmt.step() ? stmt.getValue(0).getString() : undefined;
+      });
+      if (channel !== undefined)
+        return channel;
+    } catch {
+      return ChannelControl.sharedChannelName;
```
kabentley, line 82: add a comment that this happens if the db is too old and the channel root aspect class isn't present. In that case we just say everything is in the shared channel.

core/backend/src/IModelDb.ts
```diff
@@ -3196,7 +3199,10 @@ export class StandaloneDb extends BriefcaseDb {
       if (iTwinId !== Guid.empty) // a "standalone" iModel means it is not associated with an iTwin
         throw new IModelError(IModelStatus.WrongIModel, `${filePath} is not a Standalone iModel. iTwinId=${iTwinId}`);
       assert(undefined !== file.key);
-      return new StandaloneDb({ nativeDb, key: file.key, openMode, briefcaseId: BriefcaseIdValue.Unassigned });
+      const db = new StandaloneDb({ nativeDb, key: file.key, openMode, briefcaseId: BriefcaseIdValue.Unassigned });
```
kabentley, line 3202: I don't think this is appropriate. Standalone iModels can still have channels. It's OK to do it for createEmpty because we know there's nothing there, but not when we open an existing file.

core/backend/src/test/standalone/IModelWrite.test.ts
```diff
@@ -53,7 +54,7 @@ describe("IModelWriteTest", () => {
   });
   after(() => HubMock.shutdown());
 
-  it("Check busyTimeout option", async () => {
+  it.only("Check busyTimeout option", async () => {
```
kabentley, line 57: remove .only

editor/backend/src/EditBuiltInCommand.ts
```diff
@@ -23,6 +23,16 @@ export class BasicManipulationCommand extends EditCommand implements BasicManipu
 
   public override async onStart() { return BasicManipulationCommand.commandId; }
 
+  public async addAllowedChannel(channelKey: string): Promise<IModelStatus> {
```
kabentley, line 26: no point in returning a status

core/backend/src/ChannelControl.ts
diegoalexdiaz: I don't know how I missed adding that comment. Sorry!

core/backend/src/IModelDb.ts
diegoalexdiaz: Makes sense now. Thanks!

core/backend/src/test/standalone/IModelWrite.test.ts
diegoalexdiaz: This test is failing for me locally but I can't figure out why. The failure doesn't seem to be related to the changes to Channels. I'll make the PR officially available for review so the tests are run in the cloud. At this point, I'm hoping it is just a fluke in my machine.

full-stack-tests/presentation/src/backend/IModelUpdates.test.ts
```diff
@@ -27,9 +27,9 @@ describe("Reacting to IModel data changes", () => {
     }
     const socketStub = {
       send: sinon.stub(),
-      addListener: sinon.stub().returns(() => {}),
+      addListener: sinon.stub().returns(() => { }),
       removeListener: sinon.stub(),
-      handle: sinon.stub().returns(() => {}),
+      handle: sinon.stub().returns(() => { }),
```
grigasp, line 32: please revert these formatting changes

full-stack-tests/presentation/src/IModelSetupUtils.ts
```diff
@@ -93,11 +94,11 @@ export function insertPhysicalElement(db: IModelDb, label: string, modelId: Id64
...
-          parent: {
-            id: parentId,
-            relClassName: "BisCore:PhysicalElementAssemblesElements",
-          },
-        }
+        parent: {
+          id: parentId,
+          relClassName: "BisCore:PhysicalElementAssemblesElements",
+        },
+      }
```
grigasp, line 101: please revert these formatting changes

full-stack-tests/presentation/src/IModelSetupUtils.ts
```diff
@@ -19,6 +19,7 @@ export function createValidIModelFileName(imodelName: string) {
 export async function buildTestIModelDb(name: string, cb: (db: IModelDb) => Promise<void>) {
   const outputFile = setupOutputFileLocation(createValidIModelFileName(name));
   const db = SnapshotDb.createEmpty(outputFile, { rootSubject: { name } });
+  db.channels.addAllowedChannel(ChannelControl.sharedChannelName);
```
grigasp, line 22: Looking at changes in `IModelDb.ts` I see that this is being called in `SnapshotDb.createEmpty` - do we need to repeat it here?

core/backend/src/test/standalone/IModelWrite.test.ts
kabentley: i assume this has something to do with the change you removed from @khanaffan. I don't understand why you needed to do that (or why he made it necessary to pass `props` twice).

full-stack-tests/presentation/src/IModelSetupUtils.ts
diegoalexdiaz: Missed that one. Thanks!

core/backend/src/test/standalone/IModelWrite.test.ts
diegoalexdiaz: The change from my side was certainly by mistake. I don't remember what exactly led me to change that line. Changing it back addressed the remaining test failure. Thanks!

kabentley:
> 2. Is there any Channel Root Element in an iModel for a particular "Channel key"?

Yes, this is what i'm asking. Suppose you want to create a new model "in a channel". How can i find the channel root element, if it exists?

diegoalexdiaz:
What we have discussed is that we need multiple Channel Roots per Channel Key because an authoring app can store its data across multiple Partitions in the same iModel. That is:

1) A user (admin) configures the expected Subject hierarchy for their iModels.
2) An authoring app needs to be told where in the Subject hierarchy to store its data (we haven't finished the discussion on how that happens). 
3) An authoring app may have been designed to create its data across multiple partitions. E.g. a PhysicalPartition, a FunctionalPartition, a DefinitionPartition and a DocumentPartition. Each one of those 4 partitions are created under the correct Subject(s) from step 2), and all thsoe four need to be made ChannelRoot elements with the same ChannelKey (e.g. "myAppKey").

A "Channel" is then the union of those 4 model-branches, each one with a "ChannelRoot".

kabentley:
> What we have discussed is that we need multiple Channel Roots per Channel Key because an authoring app can store its data across multiple Partitions in the same iModel. That is:
> 
> 1. A user (admin) configures the expected Subject hierarchy for their iModels.
> 2. An authoring app needs to be told where in the Subject hierarchy to store its data (we haven't finished the discussion on how that happens).
> 3. An authoring app may have been designed to create its data across multiple partitions. E.g. a PhysicalPartition, a FunctionalPartition, a DefinitionPartition and a DocumentPartition. Each one of those 4 partitions are created under the correct Subject(s) from step 2), and all thsoe four need to be made ChannelRoot elements with the same ChannelKey (e.g. "myAppKey").
> 
> A "Channel" is then the union of those 4 model-branches, each one with a "ChannelRoot".

I don't like this, it violates the theory that all channels are singularly rooted. That is important if we try to categorize changesets by the channels they modify (that will be important, i think). Why not have 4 channels?

diegoalexdiaz:
That works for authoring apps. They will just need to use a different key per Partition (e.g. "myAppKey-Physical", "myAppKey-Documents", etc). If that's the rule, I'll enforce it in makeChannelRoot (i.e. not allowing duplicates) and update the md article on the topic.

BTW, the reusage of a ChannelKey across multiple ChannelRoots was discussed in light of LocalSync last year. I believe @swwilson-bsi is already reusing a ChannelKey on each Partition that gets created (mapped to an external DgnModel). That logic will have to change so unique ChannelKeys get assigned (e.g. "MstnLocalSyncKey-<dgnfilename-dgnmodelname>").

diegoalexdiaz:
Legacy channels (created by Batch Connectors as Json on certain Subjects) can certainly have Subjects below them. But that approach prevents users from laying out a Subject hierarchy that works for them (ala Folder-structure in Projectwise), which was the original premise used when such concept was introduced in BIS during its early days. 

Thus, for LocalSync and authoring apps, we're working with that premise in mind. So, if Subjects can be defined/rearranged by users, Subjects can't be coupled to software-based expectations. That brings us to Channels being rooted at Partitions instead (i.e. No Subjects under Partitions) or, in special cases, rooted at certain DefinitionContainers (when a Channel is needed at the repository-global/DictionaryModel level).

docs/learning/backend/Channel.md
```diff
@@ -58,19 +65,19 @@ Later, to disallow editing of that channel call:
...
 
 E.g.:
 
 ```ts
-  imodel.channels.insertChannelSubject({ subjectName: "Chester", channelKey: "surface-stubs" });
+  imodel.channels.makeChannelRoot({ elementId: partitionId, channelKey: "surface-stubs" });
 ```
 
-Generally, Channel Subject elements are created as an child of the Root Subject. However,  `insertChannelSubject` accepts an optional `parentSubjectId` argument so that Channel Subjects can appear elsewhere in the Subject hierarchy. However, channels may not nest. Attempts to create a Channel Subject within an exiting channel will throw an exception.
+Note that channels may not nest. Attempts to create a Channel root element within an exiting channel will throw an exception.
```
jchick-bentley, line 80: should be 'existing' not 'exiting'

docs/learning/backend/Channel.md
```diff
@@ -1,44 +1,44 @@
-# Working With Channels in iModels
+# Editing Channels in iModels
```
kabentley, line 1: i think it should be `Working with "Editing Channels" in iModels` with the quotes. Otherwise "Editing" looks like a verb.

docs/learning/backend/Channel.md
```diff
@@ -1,44 +1,44 @@
...
 
-A "channel" is a tree of models and elements below a *Channel Subject* element. Channels segregate the contents of an iModel into *sections* to provide access control over which applications may change which data. The concept is *cooperative* in that applications indicate the channels to which they pertain, and any attempt to modify data outside one of those channels is denied with a *channel constraint* exception at runtime.
+A "Channel" is a tree of models and elements below one *Channel Root* `Subject` element. Channels segregate the contents of an iModel into *sections* to provide access control over which applications may change which data. The concept is *cooperative* in that applications indicate the channels to which they pertain, and any attempt to modify data outside one of those channels is denied with a *channel constraint* exception at runtime.
```
kabentley, line 3:
```suggestion
An "Editing Channel" (hereinafter just "Channel") is a tree of models and elements below one *Channel Root* `Subject` element. Channels segregate the contents of an iModel into *sections* to provide access control over which applications may change which data. The concept is *cooperative* in that applications indicate the channels to which they pertain, and any attempt to modify data outside one of those channels is denied with a *channel constraint* exception at runtime.
```

kabentley:
> This introduces a breaking change of the worst kind: one that does not produce a compiler error, but causes your code to no longer work properly. At a minimum, people who upgrade to this new version will need to add the following line to make it work: `iModel.channels.addAllowedChannel("shared")` Better, they would start using channels properly.
> 
> Besides our own apps, we know of third party connectors using the TypeScript APIs that will be impacted by this. @calebmshafer how do you want to proceed?

I think the universe of affected applications is small, but definitely not zero. I suppose one approach to this is to add a new IModelHostOption field to disable channel checking altogether (that's the previous behavior). It would have to default to true, but iTS (and our tests) would set it to false. 

Transformers will need to be reworked to become channel-aware, because what they're doing now (preserving channels) is definitely wrong.

kabentley:
> I think the universe of affected applications is small, but definitely not zero. I suppose one approach to this is to add a new IModelHostOption field to disable channel checking altogether (that's the previous behavior). It would have to default to true, but iTS (and our tests) would set it to false.

I added something for this. It means that we'll have to tell the iTS team to add it to their startup options. Probably warrants an entry in nextversion too (which I didn't write). I made our backend tests override the default so they test the new behavior. Probably should add a way to test that the default allows editing the shared channel.

pmconne (approved):

jchick-bentley:
> @jchick-bentley fyi as this will affect the connector framework.
> 
> @aruniverse

Thanks Caleb - I've been monitoring.

bbastings (approved):

jchick-bentley (approved):

PR #6512, opened 2024-03-12
Disable tile preloading when under GPU memory pressure

DStradley:
When GpuMemoryLimit is used, and the model being viewed is using at or over the limit due to selected tiles, then the system gets into a cycle where the preloading keeps requesting tiles, and the memory manager keeps freeing them, because they are not necessary to the view.  This causes it to just churn repeatedly and tile loading never completes.  If waitForSceneCompletion is used during this, it will effectively hang  (see https://github.com/iTwin/itwinjs-core/pull/6481).

This PR disables preloading when under memory pressure.  TileAdmin.isPreloadingAllowed determines this dynamically, and is checked before preloading.  Basically, it computes % memory utilization based on LRUTileList.totalBytesUsed / max memory limit (or 0%, if no memory limit) and returns false if it's above some threshold (default and max is 80%).  The threshold can be changed with TileAdmin.Props.percentGPUMemDisablePreload, and if that is set to 0 then preloading is disabled.


DStradley:

PR #6519, opened 2024-03-14
Cut Plane Graphics for Batched Tiles

MarcNeely:
This modification allows cut plane graphics to be generated for batched tiles.

It also enhances the fdt test clip style key-in command so that you can optionally specify a CutStyleProps to use for the cut plane geometry instead of the hard-coded one.


MarcNeely:
@pmconne Paul, I've put this up as a draft PR so you can look at it.  It's probably not what you were expecting, but it seems to work just fine.

markschlosseratbentley (approved):

markschlosseratbentley:
@MarcNeely plans to run ImageTests to check for any unintended consequences.

Issue #6518, opened 2024-03-14
`CloudSqlite.requestToken` returns empty token if `BlobContainer.service` is not defined


johnnyd710:
If `BlobContainer.service` is `undefined`, any call to `CloudSqlite.requestToken` will return `""` which is usually interpreted as a sign-in error (see [IModelApp.getAccessToken](https://www.itwinjs.org/reference/core-frontend/imodelapp/imodelapp/getaccesstokenstatic/#getaccesstoken)). So its impossible to tell the difference between an authentication error (e.g., user does not have RBAC permissions for the cloud sqlite container) and `BlobContainer.service` not being set. I believe the problematic line is [here](https://github.com/iTwin/itwinjs-core/blob/a0339bebfab82065f771c162bc9fa24cd2ae577c/core/backend/src/CloudSqlite.ts#L42):
```ts
// `@itwin/core-backend` CloudSqlite.ts
  export async function requestToken(args: RequestTokenArgs): Promise<AccessToken> {
    // allow the userToken to be supplied via Rpc. If not supplied, or blank, use the backend's accessToken.
    const userToken = args.userToken ? args.userToken : (await IModelHost.getAccessToken());
    const response = await BlobContainer.service?.requestToken({ ...args, userToken });
    return response?.token ?? "";
  }
```

**Expected behavior**

I would expect if `BlobContainer.service` is `undefined`, I'd get a clear error message `BlobContainer.service is not defined` or something like that so I know I forgot to initialize it.

PR #6521, opened 2024-03-14
Add undefined Error to CloudSqlite.requestToken

MichaelSwigerAtBentley:
CloudSqlite.requestToken now throws an error if BlobContainer.service is undefined.

This prevents confusion in cases where the BlobContainer.service might be defined but authentication is unsuccessful. 

#6518


core/backend/src/CloudSqlite.ts
```diff
@@ -39,7 +39,11 @@ export namespace CloudSqlite {
   export async function requestToken(args: RequestTokenArgs): Promise<AccessToken> {
     // allow the userToken to be supplied via Rpc. If not supplied, or blank, use the backend's accessToken.
     const userToken = args.userToken ? args.userToken : (await IModelHost.getAccessToken());
-    const response = await BlobContainer.service?.requestToken({ ...args, userToken });
+    if (BlobContainer.service === undefined) {
+      logError(`BlobContainer.service is not defined`);
```
kabentley, line 43: Do not log errors on throw, they can be logged where they're caught.

pmconne (approved):

PR #6522, opened 2024-03-14
Fix bbox JSON types

pmconne:
`Placement3d.bbox` was typed as `LowAndHighXYZ` but it's serialized on the backend by `BeJsGeomUtils::Range3dToJson` which produces an object with `low` and `high`, each of which is an array of 3 numbers. The native deserialization permits `low` and `high` to each be either the array of 3 numbers, or an object with `{x, y, z}`.

`Range3dProps` permits a single array of 6 numbers, so that's not appropriate here (produces compiler errors in TypeScript).

Introduced `LowAndHighXYZProps` which supports these 2 different representations. Made analogous changes for `Placement2d` and `GeometryPartProps.bbox`.


bbastings (approved):

aruniverse (approved):

dassaf4 (commented):
Good fix. Thanks @pmconne.

Issue #6520, opened 2024-03-14
Deferred cesiumIonKey initialization


FelixGirard:
**Is your feature request related to a problem? Please describe.**
Since we use a service (PSS) to get the key, we need to do it after we are authenticated.

**Describe the solution you'd like**
I would like to have a deferred way to set the cesiumIonKey.

**Describe alternatives you've considered**

**Additional context**
Seems like you embrace the deferred initialization for other properties: https://github.com/iTwin/itwinjs-core/pull/6449
So I think it would make sense to apply that to others.

ben-polinsky:
This makes sense. We'll get someone on it.

pmconne:
Most of the context for this issue is in the issue linked by @FelixGirard, so let's discuss it there.

tcobbs-bentley:
Based on https://github.com/iTwin/itwinjs-core/pull/6526, should this issue going to be closed as not planned?

PR #6526, opened 2024-03-15
Allow cesiumIonKey to use a provider function

tcobbs-bentley:


tcobbs-bentley:
@aruniverse @ben-polinsky I set this up as a draft PR for a few reasons:

1. Is my change to the cesiumIonKey types allowed in a point release? The new types are a superset of the old ones, so they should be fully backwards compatible. However, I'm not sure if my changes are allowed.
2. The [issue](https://github.com/iTwin/itwinjs-core/issues/6520) requested something similar to a Presentation PR, and that PR did things very differently to what I do here. I don't think it's possible to make things async here, so users will have to return undefined if their callback is called before they have fetched a key. I'm not sure if that is acceptable or not.
3. I have not yet tested my solution. However, if 1 or 2 make the whole solution a no-go, there's no point spending the time to test.
4. Related to the above: I have not created a unit test. I tried running the existing RealityDataAccess tests that use the key, and they didn't work, and I haven't yet figured out why.

pmconne:
Doesn't the function that obtains the key need to be async if - as described in the issue - it needs to query the product settings service?

ben-polinsky:
> 1. Is my change to the cesiumIonKey types allowed in a point release? The new types are a superset of the old ones, so they should be fully backwards compatible. However, I'm not sure if my changes are allowed.

It is unfortunately a breaking change in that any code reading the `TileAdmin.Props.cesiumIonKey` would not be checking if the value is function.

pmconne:
> It is unfortunately a breaking change in that any code reading the TileAdmin.Props.cesiumIonKey would not be checking if the value is function.

`TileAdmin.Props` is only used during initialization (`IModelApp.startup`). It doesn't have to be implemented the way @tcobbs-bentley did it here. He could have added a new `getCesiumIonKey` function. But if that function is not `async`, there's no point - it doesn't solve the problem.

`TileAdmin.cesiumIonKey` is `beta` so technically could be removed, though I'd prefer to deprecate it rather than break anyone who might be using it. It could be replaced with an `async` `getCesiumIonKey` function, except we have two `public` APIs in itwinjs-core - `DisplayStyleState.get/setOSMBuildingDisplay` that would have to be made `async` - a breaking change. We don't know what code outside of the core repo is depending on synchronous access to the key.

I'm not convinced it's worth it. The motivation is that the app calls `IModelApp.startup` before signing in, but they need to be signed in to get the Cesium ION key, so can't supply it to `IModelApp.startup`. Simplest solution is to change the app to sign in first. It would be *nice* if we didn't waste time waiting to fetch the key before the app startup process can proceed, since we don't need it until we open a viewport containing a background map - but I don't think it's nice *enough* to warrant breaking things.

ben-polinsky:
>  Simplest solution is to change the app to sign in first. 
I agree... @FelixGirard did you consider this?

FelixGirard:
We don't control the authenticate flow, but I think that moving the authentication before the startup is a good solution. You might have a look at the private linked issue for more discussion on the topic. Thanks!

tcobbs-bentley:
My expectation here was the an app would fetch the key asynchronously after the startup call and then store it somewhere that the function could immediately return. I acknowledge that this is not ideal, but didn't see any other choice due to the inability to make the code that makes use of the key async. Since this is apparently not considered useful, should this draft PR simply be closed?

pmconne:
> an app would fetch the key asynchronously after the startup call and then store it somewhere that the function could immediately return.

This would introduce a race condition - if the app doesn't finish fetching the key before TileAdmin needs it, the function would not be able to return the key.

PR #6527, opened 2024-03-18
Replace schema changes with differences api in Schema Merger

StefanApfel-Bentley:
In it's first implementation the Schema Merger used the schema changes api from schema validation to discover what changes between two schemas needed to be merged. This is no replaced with the differencing api that produces a flat list of schema alternations and possible conflicts which allows the clients to alter the differences before they start to merge.


core/ecschema-editing/src/Differencing/SchemaDiagnosticVisitor.ts
```diff
@@ -59,19 +56,21 @@ interface LookupArgs {
...
-  constructor(differenceReport: SchemaDifferences) {
-    this._differenceReport = differenceReport;
+  constructor(changes: Array<AnySchemaDifference>, conflicts: Array<SchemaDifferenceConflict>) {
+    this._changes = changes;
+    this._conflicts = conflicts;
   }
 
-  private addEntry<T extends AnySchemaDifference>(entry: T): T {
-    this._differenceReport.changes.push(entry);
+  private addEntry<T extends AnySchemaDifference, R=Extract<AnySchemaDifference, Omit<T, "difference">>>(entry: R): R {
```
ColinKerr, line 67: If there are partial differences which do not have a 'difference' property set  then that should be included in the difference types.  Here we're just lying about the contract.  Folks will be able to get a schema difference with an unset difference property even though the contract is that it will always exist.

core/ecschema-editing/src/Differencing/SchemaDiagnosticVisitor.ts
```diff
@@ -59,19 +56,21 @@ interface LookupArgs {
...
+    this._changes.push(entry as AnySchemaDifference);
     return entry;
   }
 
   private lookupEntry(changeType: DifferenceType, args: LookupArgs) {
-    return this._differenceReport.changes && this._differenceReport.changes.find((change) => {
+    return this._changes && this._changes.find((change: any) => {
```
ColinKerr, line 73: why have change be of type any?

core/ecschema-editing/src/Differencing/SchemaDiagnosticVisitor.ts
```diff
@@ -254,20 +253,20 @@ export class SchemaDiagnosticVisitor {
...
       itemName: enumeration.name,
       path: "$enumerators",
-      json: enumerator,
+      difference: enumerator,
     });
   }
 
   private lookupEnumeratorEntry(changeType: DifferenceType, item: string, enumeratorName: string) {
-    return this._differenceReport.changes && this._differenceReport.changes.find((change) => {
+    return this._changes && this._changes.find((change: any) => {
```
ColinKerr, line 264: same here why make type explicitly any?

ColinKerr: btw I get that the union type here does not have itemName but I think that's an indication that there is something wrong rather than an indication that you should cast away the type

core/ecschema-editing/src/Differencing/SchemaDifference.ts
```diff
@@ -74,26 +76,41 @@ export namespace SchemaDifference {
...
+  [P in keyof Editable<Omit<T, keyof Omit<SchemaItemProps, "label" | "description" | "customAttributes">>>]: T[P]
 };
 
+type GroupByDifferenceType<T extends  { changeType: DifferenceType, difference: unknown }> =
+  T extends { changeType: infer R extends DifferenceType }
+    ? R extends "modify"
+      ? Omit<T, "difference"> & { difference: Partial<T["difference"]> }
+      : T
+    : never;
+
```
ColinKerr, line 113: I don't think the combination of all of this is going to be maintainable.  Having the 'Any' types were always designed to be discriminated unions: https://basarat.gitbook.io/typescript/type-system/discriminated-unions. This just makes the whole thing way complex

core/ecschema-editing/src/Differencing/SchemaDifference.ts
```diff
@@ -102,119 +119,260 @@ export interface SchemaDifferences {
...
-  readonly changes: AnySchemaDifference[];
-  readonly conflicts: SchemaDifferenceConflict[];
+  readonly changes?: AnySchemaDifference[];
+  readonly conflicts?: SchemaDifferenceConflict[];
 }
 
 /**
  * @internal
  */
 export type AnySchemaDifference =
```
ColinKerr, line 129: how much value do we get by treating absolutely every difference polymorphicly?  Do we need to do such complex type magic to get that value?

core/ecschema-editing/src/Merging/ClassMerger.ts
```diff
@@ -2,209 +2,187 @@
...
-      // merge custom attributes
-      await merger.handleError(mergeCustomAttributes(merger.context, change.customAttributeChanges.values(), async (ca) => {
-        return merger.context.editor.entities.addCustomAttribute(targetItemKey, ca);
-      }));
+function getBaseClassSetter(context: SchemaMergeContext, item: ECClass) {
+  return async (itemKey: SchemaItemKey, baseClassKey: SchemaItemKey) => {
+    switch(item.schemaItemType) {
+      case SchemaItemType.EntityClass: return context.editor.entities.setBaseClass(itemKey, baseClassKey);
+      case SchemaItemType.Mixin: return context.editor.mixins.setMixinBaseClass(itemKey, baseClassKey);
+      // TODO: verify; structs and relationship classes can't have base classes?
```
ColinKerr, line 174: Relationships very often have base classes.  Technically structs can have base classes but it is discouraged.  So we should probably support both here.

core/ecschema-editing/src/Merging/EntityClassMerger.ts
```diff
@@ -2,40 +2,57 @@
...
+ * @internal
+ */
+export const entityClassMixinMerger: SchemaItemMergerHandler<EntityClassMixinDifference> = {
+  async add(context, change) {
+    for(const mixinFullName of change.difference) {
+      const mixinKey = await updateSchemaItemKey(context, mixinFullName);
+      const entityKey = new SchemaItemKey(change.itemName, context.targetSchemaKey);
+      const result = await context.editor.entities.addMixin(entityKey, mixinKey);
+      if(result.errorMessage === undefined) {
+        throw new Error(result.errorMessage);
```
ColinKerr, line 50: shouldn't it throw if error message is defined?   Do we have a test that hits this code?

core/ecschema-editing/src/Merging/ClassMerger.ts
StefanApfel-Bentley: I've created a backklog item to cover that #6582

ColinKerr (approved):

common/changes/@itwin/presentation-common/schema-editing-replaceschemachangeswithdiffapi_2024-04-03-15-16.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-common",
+      "comment": "Removed deprecated calls of schemaItemTypeToString",
```
grigasp, line 5: Change files should only mention changes that are important to package consumers. I don't think they care about us removing some deprecated calls.

```suggestion
      "comment": "",
```

grigasp (approved):
Approving changes under `presentation` dir.

common/changes/@itwin/presentation-common/schema-editing-replaceschemachangeswithdiffapi_2024-04-03-15-16.json
StefanApfel-Bentley: I added a note to the nextversion.md file

grigasp: That's great, but I think the change comment should still be removed.

PR #6530, opened 2024-03-19
SchemComparer should not create diagnosis entries for every property on added items

AureMonke:


christophermlawson (approved):
Nice job.  Looks good.

christophermlawson (approved):

PR #6541, opened 2024-03-21
Promote editable workspace classes from @internal to @beta

williamkbentley:
Classes were previously @internal, but should be exposed for workspace management apps.
These classes should only be used in a dedicated administer-facing settings manager application. Never in an end-user application.


pmconne:
Each of the classes you've promoted from `internal` to `beta` (except for `EditableWorkspaceDb`) is an implementation of a `beta` interface. Instead of exposing the implementation details, did you consider adding an interface for editing, and functions for instantiating them?

core/backend/src/workspace/Workspace.ts
```diff
@@ -657,10 +657,11 @@ export class ITwinWorkspaceDb implements WorkspaceDb {
 }
 
 /**
- * An editable [[WorkspaceDb]]. This is used by administrators for creating and modifying `WorkspaceDb`s.
+ * An editable [[WorkspaceDb]]. This is used by administrators for creating and modifying `WorkspaceDb`s and should
+ * only be used in a dedicated "administrative settings manager" application - never in a user facing application.
```
kabentley, line 661:
```suggestion
 * only be used in a dedicated "administrative workspace editor" application - never in a user facing application.
```

core/backend/src/workspace/Workspace.ts
```diff
@@ -657,10 +657,11 @@ export class ITwinWorkspaceDb implements WorkspaceDb {
...
  * For cloud-backed containers, the write token must be obtained before this class may be used. Only one user at at time
  * may be editing.
- * @internal
+ * @beta
```
kabentley, line 664: It seems to me that only this class needs to become public, right?

pmconne: extract-api is going to fail. You can't give a derived class a higher release tag than its base class.

kabentley: reworked

pmconne (approved):

PR #6543, opened 2024-03-21
`IModelDb`: Fix `IModelDb.views.accessViewStore` ignoring supplied `userToken` and `accessLevel` values

roluk:
I cannot open a ViewStore from my test frontend application. It turns out `accessViewStore` function isn't forwarding user token to `CloudSqlite.requestToken` invocation when `ContainerProps` are not supplied.

### DIFF
--- a/common/changes/@itwin/core-backend/master_2024-03-21-12-51.json
+++ b/common/changes/@itwin/core-backend/master_2024-03-21-12-51.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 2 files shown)


ben-polinsky (approved):

kabentley:
I see this already merged, but shouldn't we add a test to make sure this works?

ben-polinsky:
> I see this already merged, but shouldn't we add a test to make sure this works?

Should have asked for one, yes. @roluk can you follow up with one?

PR #6549, opened 2024-03-25
Struct property null update handling test

6ar8nas:
Related PR on iModel-native side iTwin/imodel-native#688 aiming to solve issue iTwin/itwinjs-backlog#1057

This changes how we check for expectedValues in the test fixture, where null equals to setting to null and undefined means no change.

### DIFF
--- a/core/backend/src/test/element/ElementRoundTrip.test.ts
+++ b/core/backend/src/test/element/ElementRoundTrip.test.ts
@@ -59,45 +59,64 @@ interface TestElementAspect extends IPrimitive, IPrimitiveArray, ElementAspectPr
 interface TestElementRefersToElements extends IPrimitive, IPrimitiveArray, ElementRefersToElements { }
 
 function verifyPrimitiveBase(actualValue: IPrimitiveBase, expectedValue: IPrimitiveBase) {
-  if (expectedValue.i)
+  if (expectedValue.i !== undefined)
     assert.equal(actualValue.i, expectedValue.i, "'integer' type property did not roundtrip as expected");
-  if (expectedValue.l)
+  if (expectedValue.l !== undefined)
     assert.equal(actualValue.l, expectedValue.l, "'long' type property did not roundtrip as expected");
-  if (expectedValue.d)
+  if (expectedValue.d !== undefined)
     assert.equal(actualValue.d, expectedValue.d, "'double' type property did not roundtrip as expected");
-  if (expectedValue.b)
+  if (expectedValue.b !== undefined)
     assert.equal(actualValue.b, expectedValue.b, "'boolean' type property did not roundtrip as expected");
-  if (expectedValue.dt)
+  if (expectedValue.dt !== undefined)
     assert.equal(actualValue.dt, expectedValue.dt, "'dateTime' type property did not roundtrip as expected");
-  if (expectedValue.s)
+  if (expectedValue.s !== undefined)
     assert.equal(actualValue.s, expectedValue.s, "'string' type property did not roundtrip as expected");
   if (expectedValue.p2d) {
-    assert.equal(actualValue.p2d?.x, expectedValue.p2d?.x, "'Point2d.x' type property did not roundtrip as expected");
-    assert.equal(actualValue.p2d?.y, expectedValue.p2d?.y, "'Point2d.y' type property did not roundtrip as expected");
+    assert.equal(actualValue.p2d?.x, expectedValue.p2d.x, "'Point2d.x' type property did not roundtrip as expected");
+    assert.equal(actualValue.p2d?.y, expectedValue.p2d.y, "'Point2d.y' type property did not roundtrip as expected");
+  } else if(expectedValue.p2d === null) {
+    assert.equal(actualValue.p2d, expectedValue.p2d, "'Point2d' type property did not roundtrip as expected.");
   }
   if (expectedValue.p3d) {
-    assert.equal(actualValue.p3d?.x, expectedValue.p3d?.x, "'Point3d.x' type property did not roundtrip as expected");
-    assert.equal(actualValue.p3d?.y, expectedValue.p3d?.y, "'Point3d.y' type property did not roundtrip as expected");
-    assert.equal(actualValue.p3d?.z, expectedValue.p3d?.z, "'Point3d.z' type property did not roundtrip as expected");
+    assert.equal(actualValue.p3d?.x, expectedValue.p3d.x, "'Point3d.x' type property did not roundtrip as expected");
+    assert.equal(actualValue.p3d?.y, expectedValue.p3d.y, "'Point3d.y' type property did not roundtrip as expected");
+    assert.equal(actualValue.p3d?.z, expectedValue.p3d.z, "'Point3d.z' type property did not roundtrip as expected");
+  } else if(expectedValue.p3d === null) {
+    assert.equal(actualValue.p3d, expectedValue.p3d, "'Point3d' type property did not roundtrip as expected.");
   }
-  if (expectedValue.bin)
+  if (expectedValue.bin) {
     assert.isTrue(blobEqual(actualValue.bin, expectedValue.bin), "'binary' type property did not roundtrip as expected");
-  if (expectedValue.g)
+  } else if(expectedValue.bin === null) {
+    assert.equal(actualValue.bin, expectedValue.bin, "'binary' type property did not roundtrip as expected.");
+  }
+  if (expectedValue.g) {
     expect(actualValue.g, "'geometry' type property did not roundtrip as expected.").to.deep.equal(expectedValue.g);
+  } else if(expectedValue.g === null) {
+    assert.equal(actualValue.g, expectedValue.g, "'geometry' type property did not roundtrip as expected.");
+  }
 }
 
 function verifyPrimitiveArrayBase(actualValue: IPrimitiveArrayBase, expectedValue: IPrimitiveArrayBase) {
   if (expectedValue.array_bin) {
     assert.equal(actualValue.array_bin!.length, expectedValue.array_bin.length, "'binary[].length' array length mismatch");
     expectedValue.array_bin.forEach((value, index) => {
-      assert.isTrue(blobEqual(actualValue.array_bin![index], value), "'binary[]' type property did not roundtrip as expected");
+      if(value) {
+        assert.isTrue(blobEqual(actualValue.array_bin![index], value), "'binary[]' type property did not roundtrip as expected");
+      } else if(value === null) {
+        assert.equal(actualValue.array_bin![index], value, "'binary[]' type property did not roundtrip as expected");
+      }
     });
+  } else if(expectedValue.array_bin === null) {
+    assert.equal(actualValue.array_bin, expectedValue.array_bin, "'binary[]' type property did not roundtrip as expected.");
   }
+
   if (expectedValue.array_i) {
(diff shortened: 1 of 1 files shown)


core/common/src/Code.ts
```diff
@@ -42,14 +42,14 @@ export class Code implements CodeProps {
...
-  public set value(val: string) { this._value = val?.trim(); }
+  public get value(): string { return this._value ?? ""; }
+  public set value(val: string | undefined) { this._value = val?.trim(); }
   private _value?: string;
 
   constructor(codeProps: CodeProps) {
     this.spec = Id64.fromJSON(codeProps.spec);
     this.scope = JsonUtils.asString(codeProps.scope);
-    this.value = JsonUtils.asString(codeProps.value);
+    this.value = codeProps.value;
```
6ar8nas, line 52: We had inconsistencies between using a setter and ctor here. Previously for undefined values we used to store an empty string through the constructor, but have it as a undefined through setter. This matches the behavior for both.

Should have no effect whatsoever.

core/common/src/ElementProps.ts
```diff
@@ -47,8 +47,6 @@ export interface ElementProps extends EntityProps {
   federationGuid?: GuidString;
   /** A [user-assigned label]($docs/bis/guide/fundamentals/element-fundamentals.md#userlabel) for this element. */
   userLabel?: string;
-  /** Optional [json properties]($docs/bis/guide/fundamentals/element-fundamentals.md#jsonproperties) of this element. */
```
6ar8nas, line 50: EntityProps define it as `{ [key: string]: any }`, which is more type-safe, so there's no need to override it here with any.

kabentley:
Have you actually determined that this is faster? Is that the only motivation for this change?

kabentley:
Can you please explain the motivation behind this change?

6ar8nas:
> Can you please explain the motivation behind this change?

Sorry, didn't expect much input on a draft PR yet, so didn't notice the question.

To my knowledge, the motivation is not only it being faster. We already have an interface how we can get data as a stringified JSON without some intermittent layers using the dollar syntax. This is why the expected outcome is that it should be faster, though some testing is soon to be carried out comparing with the current benchmark.

This change would result in a more accurate representation of actual data stored in DB. To illustrate, previously in some cases it seemed to not portray null struct information correctly [#itwinjs-backlog/1057](https://github.com/iTwin/itwinjs-backlog/issues/1057); even though code value is unset, previously it was always returned as an empty string. The output is using EC instance JSON adapters and it seems like it's not really that well extendable and we've had some discussions that ideally we would like to move away from it.

Applications seldom had their own reasons to do the mappings themselves (e.g. [iTwin/imodel-transformer](https://github.com/iTwin/imodel-transformer/blob/2a3d8074fe816e4e87faf85e8ac69c1bfc59d70c/packages/transformer/src/IModelExporter.ts#L295) in the past). This would streamline mapping by delivering a new mapping function.

I might be missing something, but these are some of the key points. If anything else raises any concerns, please let me know.

kabentley:
> To my knowledge, the motivation is not only it being faster.

I am very skeptical that this change will result in appreciable performance difference - and it's not clear to me which way will be faster. The DgnEelement "cost" you endeavor to eliminate adds caching, which for editing applications (where the same set of elements is read repeatedly) helps tremendously. I'm not sure eliminating it is a good idea at all.

> This change would result in a more accurate representation of actual data stored in DB. 

If there are bugs in the native code, we should fix them.

> Applications seldom had their own reasons to do the mappings themselves (e.g. [iTwin/imodel-transformer](https://github.com/iTwin/imodel-transformer/blob/2a3d8074fe816e4e87faf85e8ac69c1bfc59d70c/packages/transformer/src/IModelExporter.ts#L295) in the past). This would streamline mapping by delivering a new mapping function.
> 

My strong suggestion is that this change should be implemented in native code, not here.

core/backend/src/IModelDb.ts
```diff
@@ -1739,11 +1739,81 @@ export namespace IModelDb { // eslint-disable-line no-redeclare
...
+          return undefined;
+
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+        return mapNativeElementProps(JSON.parse(statement.getValue(0).getString()), loadProps) as T;
+      });
+
+      if (!elementProps)
         return undefined;
+
+      if (loadProps.wantGeometry || loadProps.wantBRepData) {
```
bbastings, line 1782: This should just be if (loadProps.wantGeometry)...wantBRepData is only applicable when wantGeometry is true.

core/common/src/geometry/GeometryStream.ts
```diff
@@ -435,7 +435,7 @@ export class GeometryStreamBuilder {
     return undefined !== hdr && GeometryStreamFlags.None !== (hdr.flags & GeometryStreamFlags.ViewIndependent);
   }
   public set isViewIndependent(viewIndependent: boolean) {
-    if (viewIndependent === this.isViewIndependent)
+    if (viewIndependent === this.isViewIndependent && this.getHeader() !== undefined)
```
bbastings, line 438: obtainHeader creates the header and inserts it at the head. The header is never present by default...with this change viewIndependent will always be ignored...

core/common/src/geometry/GeometryStream.ts
```diff
@@ -229,8 +229,8 @@ export class GeometryStreamBuilder {
    * containing more than one [[GeometryQuery]] differentiable by range. Not useful for a single [[GeometryQuery]] as its range and that of the [[GeometricElement]] are the same.
    * Ignored when defining a [[GeometryPart]] and not needed when only appending [[GeometryPart]] instances to a [[GeometricElement]] as these store their own range.
    */
-  public appendGeometryRanges() {
-    this.geometryStream.push({ subRange: Range3d.createNull() });
+  public appendGeometryRanges(subRange?: Range3d) {
```
bbastings, line 232: Please add a comment that subRange should not be supplied and will be ignored when inserting/updating an element as it's calculated for you...

core/common/src/geometry/GeometryStream.ts
kabentley: issues like this is why i'm very hesitant to encourage this whole change. Why not just rework the EC part of the native code? @khanaffan?

pmconne:
> more accurate DB data representation

Can you be more specific?

> considerably better Elements API performance

Have you measured this?

core/backend/src/IModelDb.ts
```diff
@@ -1865,6 +1936,82 @@ export namespace IModelDb { // eslint-disable-line no-redeclare
       });
     }
 
+    /** Get GometryStreamProps of an Element by Id.
```
bbastings, line 1939: Gometry

core/common/src/geometry/GeometryStream.ts
6ar8nas: The goal of this change was to ensure compatibility with current getElement output. I think the default behavior could remain unchanged if we added a separate method that would explicitly request appending header information

bbastings: The current output always includes the header. Now you will never include the header. For compatibility obtainHeader will need to be called followed by a call to set isViewIndependent (original version!) if needed.

```
void GeometryCollection::ToJson(BeJsValue output, BeJsConst opts) const
    {
    output.SetNull();
    GeometryStreamIO::Collection collection(m_data, m_dataSize);
    GeometryStreamIO::Reader reader(m_state.m_dgnDb);
    DgnSubCategoryId lastSubCategory = m_state.m_geomParams.GetSubCategoryId();
    bool wantBRepData = opts["wantBRepData"].asBool();

    for (auto const& egOp : collection)
        {
        switch (egOp.m_opCode)
            {
            case GeometryStreamIO::OpCode::Header:
                {
                auto header = output.appendValue()["header"];
                header["flags"] = static_cast<uint32_t>(reader.GetHeader(egOp)->m_flags);
                break;
                }
...
```

6ar8nas:
Waiting for imodel-native#782 to be completed before advancing further.

6ar8nas:
Currently blocked by iTwin/itwinjs-backlog#1162

full-stack-tests/backend/src/perftest/ElementCRUD.test.ts
```diff
@@ -82,11 +82,12 @@ function createElemProps(className: string, _iModelName: IModelDb, modId: Id64St
     elementProps.sub1Long = values.sub1Long;
     elementProps.sub1Double = values.sub1Double;
   }
-  if (className.includes("PerfElement2d"))
```
6ar8nas, line 85: This was a faulty behavior of our previous tests implementation. Fixed it, though it might slightly alter the tests output.

core/frontend/src/SubCategoriesCache.ts
```diff
@@ -31,7 +31,6 @@ export class SubCategoriesCache {
   private readonly _byCategoryId = new Map<string, Id64Set>();
   private readonly _appearances = new Map<string, SubCategoryAppearance>();
   private readonly _imodel: IModelConnection;
-  private _missingAtTimeOfPreload: Id64Set | undefined;
```
6ar8nas, line 34: Removed, as this private variable was unused.

6ar8nas:
> > more accurate DB data representation
> 
> Can you be more specific?

We would be returning data as stored in our database without layers of further mapping on imodel-native side. This would move the business/itwinjs-core logic from imodel-native to itwinjs-core itself and get the data straight from the DB without all of those adaptors.

This would only be the first part of similar efforts, with insert/update to follow thereafter, ultimately aiming to use the imodel-native data objects as the DTOs without the need to always map all of the data stored in the database to an itwinjs-core interface on imodel-native side already.
 
> > considerably better Elements API performance
> 
> Have you measured this?

iTwin/imodel-native#782 has measured it with the improvement in big actually used datasets varying between -8% and +91%.

pmconne:
> It allows bypassing the creation of a DgnElement on the native side and various json mapping adapters, which should result in both more accurate DB data representation

Have you verified the "more accurate representation" claim? Are you taking into account data fix-up/normalization that occurs during element loading such as https://github.com/iTwin/imodel-native/pull/820 or [this](https://github.com/iTwin/imodel-native/blob/b652775fbe778f2e411fe36435cf8ed4f2de6f36/iModelCore/iModelPlatform/DgnCore/ViewDefinition.cpp#L1568)?

6ar8nas:
> > It allows bypassing the creation of a DgnElement on the native side and various json mapping adapters, which should result in both more accurate DB data representation
> 
> Have you verified the "more accurate representation" claim? Are you taking into account data fix-up/normalization that occurs during element loading such as [iTwin/imodel-native#820](https://github.com/iTwin/imodel-native/pull/820) or [this](https://github.com/iTwin/imodel-native/blob/b652775fbe778f2e411fe36435cf8ed4f2de6f36/iModelCore/iModelPlatform/DgnCore/ViewDefinition.cpp#L1568)?

We are taking that into account. This just changes the mapping place from C++ backend, to an itwinjs-core/common function, as per the second link, that's needed in a restricted format only in frontend.

MichaelSwigerAtBentley:
@ColinKerr @khanaffan The new tryGetElementJson() logic does not seem to be returning null valued iModel Element Addresses. Is this expected behavior or a regression? There is a test failing that expects a null valued address to be returned.

MichaelSwigerAtBentley:
Related issue: https://github.com/iTwin/itwinjs-backlog/issues/1403

PR #6550, opened 2024-03-25
Use the last docs build before GenericInstanceFilter was added otherwise docs build fails.

nick4598:


aruniverse (approved):
fyi @ben-polinsky

ben-polinsky:
> fyi @ben-polinsky

Yep we discussed. We'll need to think of a long-term solution. Have it on my radar.

PR #6552, opened 2024-03-25
Add documentation for openCheckpoint and do prefetching in IPC/desktop case

jiaruiz717:
As https://github.com/iTwin/itwinjs-backlog/issues/1036

Add documentation for openCheckpoint and do prefetching in IPC/desktop case. Note that this is a draft PR. As discussed with @nick4598, we will do perf test to figure out what is an appropriate window size and prefetching threshold. Documentation will be adjusted accordingly and new tests will be added.

### DIFF
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -3081,8 +3081,9 @@ export class SnapshotDb extends IModelDb {
 
   /**
    * Open a Checkpoint directly from its cloud container. Data for the checkpoint is loaded from the container on demand as it is accessed.
-   * This method may optionally start a "prefetch" (see [[CloudSqlite.startCloudPrefetch]]) operation to asynchronously download all blocks in the checkpoint.
-   * Prefetching can improve performance when a large portion of the database will be needed, or to allow offline usage.
+   * This method starts a "prefetch" (see [[CloudSqlite.startCloudPrefetch]]) operation to asynchronously download all blocks in the checkpoint by
+   * default. Prefetching can improve performance when a large portion of the database will be needed, or to allow offline usage.
+   * To disable prefetching, set `args.doPrefetch = false`;
    * @beta
    */
   public static async openCheckpoint(args: OpenCheckpointArgs): Promise<SnapshotDb> {
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -3048,6 +3048,11 @@ export class SnapshotDb extends IModelDb {
     return db;
   }
 
+  /**
+   * Attach to a checkpoint and open it. This involves fetching the database name and container from V2CheckpointManager, and opening the file.
+   * @param checkpoint The checkpoint properties
+   * @returns A promise that resolves with an instance of SnapshotDb.
+   */
   private static async attachAndOpenCheckpoint(checkpoint: CheckpointProps): Promise<SnapshotDb> {
     const { dbName, container } = await V2CheckpointManager.attach(checkpoint);
     const key = CheckpointManager.getKey(checkpoint);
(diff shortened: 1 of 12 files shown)


core/backend/src/IModelDb.ts
```diff
@@ -3075,10 +3080,19 @@ export class SnapshotDb extends IModelDb {
   }
 
   /**
-   * Open a Checkpoint directly from its cloud container.
+   * Open a Checkpoint directly from its cloud container in a "lazy" manner, allowing data to be fetched as needed.
```
kabentley, line 3083:
```suggestion
   * Open a Checkpoint directly from its cloud container. Data for the checkpoint is loaded from the cloud as it is accessed.
```

core/backend/src/IModelDb.ts
```diff
@@ -3075,10 +3080,19 @@ export class SnapshotDb extends IModelDb {
...
+   * This method can optionally be combined with a prefetch to download all blocks of an iModel if desired.
+   * Prefetching can improve performance when a large portion of the database will be needed, performance would be better the more you needed to read if you happened to prefetch.
+   * But use it smartly to avoid unnecessary data transfer and storage.
+   * If only a small portion of data is needed, it's more efficient to rely solely on lazy opening.
```
kabentley, line 3087:
```suggestion
   * This method may optionally start a [prefetch](CloudSqlite.startCloudPrefetch) operation to download all blocks in the checkpoint.
   * Prefetching can improve performance when a large portion of the database will be needed, or to allow offline usage,      
```

core/backend/src/IModelDb.ts
```diff
@@ -3075,10 +3080,19 @@ export class SnapshotDb extends IModelDb {
...
+   * @param args The arguments required to open the checkpoint.
+   * @param returns A promise that resolves with an instance of SnapshotDb.
```
kabentley, line 3089: neither of these comments are helpful.

core/backend/src/IModelDb.ts
```diff
@@ -3075,10 +3080,19 @@ export class SnapshotDb extends IModelDb {
...
    * @beta
    */
   public static async openCheckpoint(args: OpenCheckpointArgs): Promise<SnapshotDb> {
+    if (args.prefetch) {
+      IModelHost.appWorkspace.settings.addDictionary("prefetch", SettingsPriority.application, {"Checkpoints/prefetch": true});
+    }
```
kabentley, line 3095: No, you shouldn't set create a setting. Add something to the argument to `V2CheckpointManager.attach`

kabentley:
The default should be to start a prefetch for IPC, but there should be a way to tell it not to do so.

core/backend/src/IModelDb.ts
nick4598: I was unsure about this, part of me says we should activate a prefetch in the same way that RPC does (through settings at the moment). What do you think about starting a prefetch manually in the `private static async attachAndOpenCheckpoint` function? We'd add a line CloudSqlite.startCloudPrefetch, we already have the dbName and container in that function after it returns from the attach so we have all that we need.

kabentley: yes, but then you'd have to reproduce all the minRequest/maxrequest and logging logic. I will pull this branch and add an outline of how i think it should work.

nick4598: We should also allow for CheckpointConnection.ts: openRemote to pass a prefetch property, no? That is the main way that a desktop app would actually be able to use this?

kabentley: that would be nice, but stupidly that function's signature can't be changed (because the last argument is optional.) I guess we could deprecate the current signature and add a new one with named arguments.

kabentley: actually, that method is only used from RPC where the backend decides whether or not to prefetch, so it shouldn't change.

kabentley: nope, wrong again. That function is also used for IPC.  I'm not sure what to do. I think it's probably fine to just have it use the default (doPrefetch=true).

nick4598: > that would be nice, but stupidly that function's signature can't be changed (because the last argument is optional.) I guess we could deprecate the current signature and add a new one with named arguments.

Yes I also was struggling with that, feels like openRemote should take a single arg which is an interface as opposed to a list of args. Is it really a breaking change to add a 4th optional arg? 

Current usage of openRemote would be 
openRemote(itwinid, imodelid)
openRemote(itwinid, imodelid, imodelversion.latest())

After the change I believe all of the following would also be acceptable:
openRemote(itwinid,imodelid, imodelversion.latest())  
openRemote(itwinid,imodelid) 
openRemote(itwinid, imodelid, imodelversion.latest(), false)

nick4598: That works

core/frontend/src/CheckpointConnection.ts
```diff
@@ -54,13 +54,55 @@ export class CheckpointConnection extends IModelConnection {
...
+   * @note this function is equivalent to [[openRemote]] but allows more flexibility specifying the changeset version,
+   * and also optionally starting a prefetch operation.
+   */
+  public static async openFromIpc(args: OpenCheckpointArgs): Promise<CheckpointConnection> {
+    const connection = new this(await IpcApp.appFunctionIpc.openCheckpoint(args), true);
+    IModelConnection.onOpen.raiseEvent(connection);
+    return connection;
+  }
+
+  /**
```
kabentley, line 82: there must have been a bad merge.

jiaruiz717: Sorry I was trying to git commit amend the email associated with two previous commits. Accidentally made this bad merge, now it's resolved.

PR #6557, opened 2024-03-26
Expand messages in errors thrown in `IModelDb.Elements`

GytisCepk:
Changes to errors thrown by `insertElement`, `updateElement` and `deleteElement` in `IModelDb`:
- ~`IModelError` is now thrown instead of `Error`.~
- `message` and `metaData` now contains info relevant to the error.
- `message` format is now more inline with other `iModelDb` errors.


core/backend/src/IModelDb.ts
```diff
@@ -1847,8 +1847,8 @@ export namespace IModelDb { // eslint-disable-line no-redeclare
       try {
         return elProps.id = this._iModel.nativeDb.insertElement(elProps);
       } catch (err: any) {
-        err.message = `error inserting element: ${err.message}`;
-        throw err;
+        const message = `Error inserting element [${err.message}], modelId: ${elProps.model}, classFullName: ${elProps.classFullName}, code: ${JSON.stringify(elProps.code)}`;
+        throw new IModelError(err.errorNumber, message, { elProps });
```
kabentley, line 1851:
```suggestion
        throw new IModelError(err.errorNumber,  `error inserting element: ${err.message}`,  elProps);
```

c-macdonald (approved):

kabentley:
In general, I think our subclasses of `Error` are problematic and we should work towards eliminating them rather than propagating them. For one thing, the "hierarchy" of them is confusing and `typeof` tests are the root of lots of problems. Also, they can't be thrown from native code, which is why I purposely left these of type `Error` rather than `IModelError` with the hope of eventually eliminating IModelError entirely.

In particular the `errorNumber` field of IModelError is not helpful. It is vestige of the C++ world where passing strings is inefficient. It does serve the purpose of providing a key that programmers can use to determine if the exception is expected so they can recover, but uniqueness is vague and problematic. Further, the "message" field is only useful to programmers in debuggers (or potentially in logs) because it can't be localized. Metadata is fine, but can become voluminous in when people just blindly log exceptions. For debugging, there's a calltrace so metadata isn't explicitly necessary for context.

I know you didn't want to open up a whole discussion of exception handling, but I wanted to explain why what you're changing is the way it is. If we had a magic wand and could rewind time, i would suggest we create an `ITwinError` interface:

```ts
export interface ITwinError {
    namespace: string;   // Localization namespace
    errorKey: string; // unique key for error, within namespace
    message: string; // explanation of what went wrong. (Should provide some format for localization, like ITwinLocalization.getLocalizedKeys but the opposite way)
    metadata?: LoggingMetaData;
}
```

and a set of static functions in an `ITwinError` namespace for dealing with them. We'd eliminate all of our classes derived from `Error`.

Or something like that.

kabentley:
> @kabentley I agree with you on the subclasses of `Error` and would like to try and avoid them as much as possible. @GytisCepk can you use `Error` directly?

The existing code uses `Error`, this changes it to use `IModelError`. Certainly we can add metadata without changing the type, but i presume @GytisCepk thought that what was there was inconsistent with other code.

> As for introducing a new type Keith, do you see a way for us to migrate to just use `Error` instead? I'd prefer if possible to not introduce a new version that we'd need to shift everyone to over time.

I think there's probably a way to do it without too much pain, but it would be a breaking change to existing code, so not possible without a major version. I think someone could define the `ITwinError` interface we want and we could start using it for new code (particularly in new iTS code).

calebmshafer:
@aruniverse let's get this on the backlog, I think this is worth cleaning up due to the confusion around IModelError and other statuses (which most are really no longer valid after we split out all of the client libraries in 3.0).

aruniverse (approved):

aruniverse:
@mergifyio backport release/4.5.x

PR #6567, opened 2024-03-27
Support EXT_meshopt_compression

pmconne:
This is the only type of compression supported by glTF for points. It significantly reduces the file size, so we want to use it in our point cloud exporters.

The [meshoptimizer decoder library](https://www.npmjs.com/package/meshoptimizer/v/0.20.0) is the de facto standard (provided by the same people who maintain the C++ encoder+decoder). I'm doing the decoding in a single web worker to avoid blocking the main event loop.

I'm testing against the [pirate sample](https://github.com/zeux/meshoptimizer/blob/master/demo/pirate.glb); awaiting a point cloud sample.
The pirate sample uses a couple of other glTF features that we lack, so I'm adding them too:
- [x] KHR_mesh_quantization
- [ ] signed, normalized, 8-bit normals

### DIFF
--- a/common/config/rush/browser-approved-packages.json
+++ b/common/config/rush/browser-approved-packages.json
@@ -659,6 +659,10 @@
       "allowedCategories": [ "tools" ]
     },
     {
+      "name": "meshoptimizer",
+      "allowedCategories": [ "frontend" ]
+    },
+    {
       "name": "mkdirp",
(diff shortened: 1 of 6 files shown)


pmconne:
Awaiting ImageTests results.

aruniverse (approved):

DStradley (approved):

PR #6572, opened 2024-03-29
Text annotation APIs

pmconne:
imodel-native: https://github.com/iTwin/imodel-native/pull/718

This PR introduces support for creating and manipulating blocks of text, laying them out, and converting them to geometric primitives. It also provides straightforward APIs for querying and updating the text blocks stored on TextAnnotation2d/3d elements and updating their geometry streams.

It introduces the following APIs:

### TextStyle (core-common)

Represents a text style obtained from a workspace. The TextStyle elements created by batch connectors are ignored. Text styles are identified by name.

### TextBlock (core-common)

In-memory representation of a block of text, with content and styling. The wire format and persistence format are the same - JSON.

TextBlocks refer to fonts by name, not Id.

### produceTextAnnotationGeometry (core-backend)

Processes a TextAnnotation and splits it into lines with word-wrapping. It then outputs primitive commands, namely: color changes, TextStrings, and fraction separators. This can be used to append a TextBlock to a geometry stream e.g. via `GeometryStreamBuilder.appendTextBlock`.

This must be in core-backend because 1) it needs access to fonts and text styles stored in workspaces and font Ids stored in the iModel's font table; 2) it needs the native freetype library to layout glyphs; and 3) it needs the `Intl.Segmenter` API that despite being a Stage 4 ECMAScript proposal implemented in most other browsers and Node since 2020, is still not generally available in Firefox.

### TextAnnotation (core-common)

A TextBlock positioned in 2d or 3d space, typically attached to an element, though you can also use it to produce decoration graphics.

The BisCore schema and native code define an optional frame and leaders as part of the annotation. No one has ever used them, so they're omitted here, but they could be introduced in the future if needed without breaking the API. I expect dimensions will have their own representations of these, though.

### TextAnnotation elements (core-backend)

These elements used to store their annotation in a separate aspect as a binary flatbuffer-encoded blob. Now, they store it directly as JSON in their `jsonProperties` field. A `setAnnotation` method updates this property and recomputes the geometry stream accordingly. 

### `dta text` keyin

A lazy, quick-and-dirty, append-only text editing keyin. You can invoke the keyin with different commands+arguments in succession to build up a TextBlock displayed as a decoration graphic in the current viewport.

---

### TODO

- [x] Fix numerator drawing underneath horizontal fraction separator
- [x] Fix numerator and denominator not aligned on center of horizontal separator
- [x] Add a very simple key-in to place a block of text
- [x] Test against real fonts

### Follow-up

- Implement lookup of text styles from workspaces.
- Deprecate BisCore classes like TextAnnotationData.
- Delete native annotation APIs and the callback that recomputes an annotation element's geometry stream when the annotation aspect is modified.


core/common/src/annotation/TextBlock.ts
```diff
@@ -0,0 +1,301 @@
...
+  export function fromJSON(props: RunProps): Run {
+    switch (props.type) {
+      case "text": return TextRun.create(props);
+      case "fraction": return FractionRun.create(props);
+      case "linebreak": return LineBreakRun.create(props);
+    }
+  }
+}
+
+export type TextRunShiftMode = "subscript" | "superscript" | "baseline";
```
pmconne, line 93: There's probably a better name than "shift mode" that eludes me.

jffmarker: I can't think of a unifying term off the top of my head either. One variation might be `BaselineShift = "none" | "subscript" | "superscript";`.

kabentley: vertical alignment?

kabentley:
re font names: 
- for styles there's no Id, since that is a table in an iModel. Therefore, i think we should migrate to using font name only in new apis.
- font names aren't quite unique if we support different font types. My suggestion is to ignore this and resolve by name using TT, then SHX, then RSC (if we even support SHX/RSC). In other words, if you have two fonts with the same name, the TT one will be used.
- font names should be case insensitive (right @jffmarker )
- some fonts can be found by more than one name. I don't see why that's a problem as long as you always find the same font for a given name
- lots more font name/family quirks when we attempt to resolve a "missing" font, but i don't think they're worth worrying about here.

kabentley:
even though existing connectors create Text Style Elements and TextAnnotationData and even refer to them by Id in other elements, nothing anywhere, including the connectors themselves can be using them for anything, right? I don't think we need to maintain any compatibility with them and can declare their types "obsolete".

pmconne:
> font names should be case insensitive

They are not treated as such by [FontMap](https://github.com/iTwin/itwinjs-core/blob/ef99f23573f5f0edce5fb7b08906e4d94d25cf65/core/common/src/Fonts.ts#L67).

kabentley:
> They are not treated as such by [FontMap](https://github.com/iTwin/itwinjs-core/blob/ef99f23573f5f0edce5fb7b08906e4d94d25cf65/core/common/src/Fonts.ts#L67).

Yup, that's a bug. They're required to be [unique case insensitively](
https://github.com/iTwin/imodel-native/blob/1fddc1bd52215e764014c0792a0a39360cda1c79/iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp#L328).

pmconne:
> [unique case insensitively](https://github.com/iTwin/imodel-native/blob/1fddc1bd52215e764014c0792a0a39360cda1c79/iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp#L328).

Eww, using SQLite's definition of "case-insensitive"?

> the 26 upper case characters of ASCII are folded to their lower case equivalents before the comparison is performed. Note that only ASCII characters are case folded. SQLite does not attempt to do full UTF case folding

pmconne:
@aruniverse our [release tags guidelines](https://www.itwinjs.org/learning/guidelines/release-tags-guidelines/) make conflicting statements about the `@preview` tag:

> A lint rule enforces that the "extensions" tag may only be applied to APIs tagged as "public" or "preview".

> "The "preview" tag is supposed to be the equivalent of "beta" for extension APIs. It is currently unused, and may never be.

> Preview API tag should be on its own line before the @extensions tag:

`@preview` is used nowhere in core-common nor core-frontend. My APIs are tagged `@beta` and `@extensions`, and they produce the following error:

```
/home/paul/c/js/s/itwinjs/core/common/src/annotation/TextBlock.ts
   15:1  error  interface "ApplyTextStyleOptions" without one of the release tags "public,preview"      @itwin/public-extension-exports
```

However, if I add `@preview`, the error is resolved. So the comment about "preview" being currently unused seems inaccurate.

I still don't understand why we need both "beta" and "preview".

kabentley:
> Eww, using SQLite's definition of "case-insensitive"?

We are supposed to [use the ICU extension](https://github.com/iTwin/imodel-native/blob/1fddc1bd52215e764014c0792a0a39360cda1c79/iModelCore/BeSQLite/BeSQLite.cpp#L5479) for that. We ship it, but I don't see where it gets hooked up. Looks to me like that got lost somewhere. @khanaffan @chuckkir

core/common/src/annotation/TextStyle.ts
```diff
@@ -0,0 +1,262 @@
...
+   * The offset is computed in meters as superScriptOffsetFactor * [[lineHeight]].
+   * Default: -0.5.
+   */
+  superScriptOffsetFactor?: number;
+  /** Multiplier used to compute the height of a superscript [[TextRun]].
+   * The height is computed as superScriptScale * [[lineHeight]].
+   * Default: 2/3
+   */
+  superScriptScale?: number;
+  /** Multiplier used to compute the width of each glyph.
```
pmconne, line 96: @jffmarker it doesn't seem right to say "width of each glyph", because glyphs can of course vary in width. What should it say instead?

chuckkir:
> We are supposed to [use the ICU extension] for that. We ship it, but I don't see where it gets hooked up. Looks to me like that got lost somewhere. @khanaffan @chuckkir @jffmarker
> 
It gets linked into itwinbentley.dll in bentley.mke:

     LINKER_LIBRARIES            = $(BuildContext)SubParts/Libs/$(libprefix)iTwinIcu4c$(libext)

It is also used by the itwinlibxml2 partfile, although I'm not sure if it is used in the DLL itself.

jffmarker:
> > Eww, using SQLite's definition of "case-insensitive"?
> 
> We are supposed to [use the ICU extension](https://github.com/iTwin/imodel-native/blob/1fddc1bd52215e764014c0792a0a39360cda1c79/iModelCore/BeSQLite/BeSQLite.cpp#L5479) for that. We ship it, but I don't see where it gets hooked up. Looks to me like that got lost somewhere. @khanaffan @chuckkir @jffmarker
> 
> This isn't that important for font names, but is definitely required for code values.

Font names can be localized though... and I agree they are case-insensitive. I think it tends not to come up in practice, however, because we're the ones that manage the font table, which is based on the names we see in the fonts... without any free-handing, the names tend to retain their case. As such... it may not be worth many hurdles to make it that way at the DB level.

Back in the day, I think we made it possible to integrate ICU with sqlite, however, it was opt-in by the application, because case folding is not universal, so the App had to pick a locale it wanted to fold against (e.g. the Turkish i scenario).

core/common/src/annotation/TextStyle.ts
jffmarker: Comment is probably over-specific with no benefit... I might go with "Scale factor applied to the width of the glyphs."

core/backend/src/TextAnnotationElement.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (!builder.appendGeometryParamsChange(params)) {
+    return false;
+  }
+
+  const props = produceTextAnnotationGeometry({ iModel: element.iModel, annotation });
+  if (!builder.appendTextBlock(props)) {
+    return false;
+  }
+
+  // ###TODO will placement bounding box be computed for me on insert/update?
```
pmconne, line 28: @bbastings questions for you.

bbastings: > @bbastings questions for you.

Yup, native code will update placement bbox.

dassaf4 (approved):
Approved core-geometry. Neat use of `this` in `Range2d.cloneTranslated`!

core/common/src/geometry/GeometryStream.ts
```diff
@@ -333,6 +334,35 @@ export class GeometryStreamBuilder {
     return true;
   }
 
+  /** Append a series of entries representing a [[TextBlock]] to the [[GeometryStreamProps]] array.
+   * @beta
+   * @extensions
+   * @preview
+   */
+  public appendTextBlock(block: TextBlockGeometryProps): boolean {
```
bbastings, line 342: Going to want ElementGeometry.Builder.appendTextBlock too (flatbuffer api).

bbastings (approved):
Looks good

core/common/src/geometry/GeometryStream.ts
pmconne: How do I change the color? It wants a GeometryParams, which wants a category Id, which I don't necessarily know.

bbastings: Backend native will handle an invalid category/sub-category and get them from the element (resets to sub-category appearance). 

For this:
  result = this.appendSubCategoryChange(Id64.invalid);
You should be able to do this:
  const params = new GeometryParams(Id64.invalid);
  builder.appendGeometryParamsChange(params);

For this:
  this.geometryStream.push({ appearance: { color: entry.color } });
You should be able to do this:
  const params = new GeometryParams(Id64.invalid);
  params.lineColor = entry.color;
  builder.appendGeometryParamsChange(params);

pmconne: added.

Josh-Schifter:
I don't see anything like an insertion cursor.  How will an editor insert runs within an existing paragraph?

pmconne:
> I don't see anything like an insertion cursor.

That's the responsibility of the editor package.

> How will an editor insert runs within an existing paragraph?

`paragraph.runs.splice(insertionIndex, 0, runToInsert)`.

markschlosseratbentley (approved):

Josh-Schifter:
> That's the responsibility of the editor package.

Where is the editor code?  
 
> `paragraph.runs.splice(insertionIndex, 0, runToInsert)`.

Fair enough but I was thinking of more generic editing.  Like inserting characters within a run, or changing properties (like underline) of a span.

pmconne:
> Where is the editor code?

This PR does not include a text editing tool. It is a necessary prerequisite for building one.

DanRod1999 (approved):

GytisCepk (approved):

core/backend/src/TextAnnotationElement.ts
```diff
@@ -0,0 +1,106 @@
...
+export class TextAnnotation2d extends AnnotationElement2d {
+  /** @internal */
+  public static override get className(): string { return "TextAnnotation2d"; }
+  protected constructor(props: TextAnnotation2dProps, iModel: IModelDb) { super(props, iModel); }
+
+  public static fromJSON(props: TextAnnotation2dProps, iModel: IModelDb): TextAnnotation2d {
+    return new TextAnnotation2d(props, iModel);
+  }
+
+  public override toJSON(): TextAnnotation2dProps {
```
aruniverse, line 48: do we need this override if all we're doing is just calling super?

aruniverse (approved):

core/backend/src/TextAnnotationElement.ts
pmconne: For the return type.

ben-polinsky:
@pmconne Looking into the documentation link issues which don't seem like typos:

`GeometryStreamBuilder.appendTextAnnotation` method does not exist?
Invalid Link: "[GeometryStreamBuilder.appendTextAnnotation]($common)" in file "reference/core-backend/elementgeometry/producetextannotationgeometry"

**This may be a bug...** 

Invalid Link: "[Placement]($geometry)" in file "reference/core-backend/elements/textannotation2d"
Invalid Link: "[Placement]($geometry)" in file "reference/core-backend/elements/textannotation3d"

Word/camel hump order (`ApplyTextStyleOptions`):

TypeDoc Reference Link Error: Reference to 'TextStyleApplyOptions' was not found at File: core-common/core/common/src/annotation/TextBlock.ts Line: 81

Spelling (`Workspace`):

Invalid Link: "[Worksapce]($backend)" in file "reference/core-common/annotation/fractionrunprops"
Invalid Link: "[Worksapce]($backend)" in file "reference/core-common/annotation/linebreakrunprops"
Invalid Link: "[Worksapce]($backend)" in file "reference/core-common/annotation/paragraphprops"
Invalid Link: "[Worksapce]($backend)" in file "reference/core-common/annotation/textblockcomponentprops"
Invalid Link: "[Worksapce]($backend)" in file "reference/core-common/annotation/textblockprops"
Invalid Link: "[Worksapce]($backend)" in file "reference/core-common/annotation/textrunprops"

($backend)?:

Invalid Link: "[TextAnnotation2d]($common)" in file "reference/core-common/entities/textannotation2dprops"
Invalid Link: "[TextAnnotation3d]($common)" in file "reference/core-common/entities/textannotation3dprops"
Invalid Link: "[TextAnnotation2d]($common)" in file "reference/core-common/entities/"
Invalid Link: "[TextAnnotation3d]($common)" in file "reference/core-common/entities/"
Invalid Link: "[TextAnnotation2d]($common)" in file "reference/core-common/all/"
Invalid Link: "[TextAnnotation3d]($common)" in file "reference/core-common/all/"

PR #6580, opened 2024-04-03
Update leftNav.md after minor release

anmolshres98:
This step has been automated here: https://github.com/iTwin/itwinjs-core/blob/17656436e63ba98e1d413acd251676f18caf4002/.github/workflows/automation-scripts/update-changelogs.mjs#L131

However, this automation is done when the finalize release script runs after a minor release. The scripts checks for the most recent commit message to verify it is a minor release then does this change. During this minor release, this check failed because when the finalize release step ran, there was already a commit made after the minor release commit. Hence, the part of the script that does this edit also did not run and now needs to be done manually for this minor.

Minors releases in the future shouldn't need this

### DIFF
--- a/docs/changehistory/4.5.0.md
+++ b/docs/changehistory/4.5.0.md
@@ -1,0 +1,226 @@
+---
+deltaDoc: true
+version: '4.5.0'
+---
+# 4.5.0 Change Notes
+
+Table of contents:
+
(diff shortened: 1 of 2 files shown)


anmolshres98:
@hl662 @DanRod1999

hl662 (approved):

DanRod1999 (approved):

PR #6587, opened 2024-04-05
Upgrade `glob` to v10.3.12

### LINKED ISSUE #6149: KnownLocations.nativeAssetsDir returns a path not compatible with the SchemaXmlFileLocater on Windows
**Describe the bug**
This method returns a path prefixed with \\?\ like `\\?\C:\banana\`. This seems to interfere with `glob.GlobSync` used to match candidate schema files

**To Reproduce**
Run this test on windows:

```typescript
  it.only("locate core custom attributes", async () => {
    await IModelHost.startup();
    const cont = new SchemaContext();
    const loc = new SchemaXmlFileLocater();
    cont.addLocater(loc);
    loc.addSchemaSearchPath(path.join(KnownLocations.nativeAssetsDir, "ECSchemas", "Standard"));

    const schemaKey = new SchemaKey("Units", 1, 0, 0);
    const schema = await cont.getSchema(schemaKey, SchemaMatchType.Latest);

    assert.isDefined(schema);
    assert.strictEqual(schema?.name, "Units");
    await IModelHost.shutdown();
  });
```
> NOTE: If you strip the first 4 characters of the string returned by nativeAssetsDir the test works

**Expected behavior**
The Units schema should be loaded correctly

- OS: Windows
- Version: 11
- iTwin.js Version: 4.x tip

akshay05n:
**Description:**
This pull request addresses critical issues encountered on Windows due to buggy behaviour in `glob` v7 and before related to path handling.

**Changes Made:**

- Updated `glob` package from version 7.1.2 to 10.3.12
- With the new version of `glob`, the previously used `glob.sync` function is no longer supported. References to `glob.sync` have been updated to `globSync`.
- All backslashes (`\\`) in glob patterns have been replaced with forward slashes (`/`) by setting the `windowsPathsNoEscape` option.

Closes #6149


pmconne:
> Further investigation is needed to determine if there is something we can do or if there is a bug in glob. If there is a bug in glob we will have to report it and put a temporary work around in place until it is fixed

Did this investigation and potential reporting happen?

akshay05n:
> > Further investigation is needed to determine if there is something we can do or if there is a bug in glob. If there is a bug in glob we will have to report it and put a temporary work around in place until it is fixed
> 
> Did this investigation and potential reporting happen?

I looked into the problem and discovered it was related to the way we were using file paths with the `glob.GlobSync` function. This function needs a search pattern, and it can be the actual file path. The issue was that on Windows, we were adding `\\?\` prefix before the path, but normal file paths don't have that prefix. This difference caused the function to not find any matching files.

example-code/snippets/src/backend/SchemaXmlFileLocater.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    assert.isTrue(schemaPath.length < 260);
+    const schemaKey = new SchemaKey("Units", 1, 0, 0);
+    const schema = await cont.getSchema(schemaKey, SchemaMatchType.Latest);
+
+    assert.isDefined(schema);
+    assert.strictEqual(schema?.name, "Units");
+  });
+
+  it("Schema path is between 260 and 1024 character long", async () => {
```
pmconne, line 29: What happens if path is > 1024 characters? Is it documented somewhere?

pmconne (approved):

example-code/snippets/src/backend/SchemaXmlFileLocater.test.ts
akshay05n: According to this thread https://discussions.apple.com/thread/250275651?sortBy=best, the maximum file path length on macOS is 1024 characters, while Linux allows for paths up to 4096 characters.

core/ecschema-locaters/src/SchemaFileLocater.ts
```diff
@@ -162,7 +162,11 @@ export abstract class SchemaFileLocater {
    * @param format The type of file that the schema key refers to. json or xml
    */
   private addCandidateSchemaKeys(foundFiles: FileSchemaKey[], schemaPath: string, fileFilter: string, desiredKey: Readonly<SchemaKey>, matchType: SchemaMatchType, format: string) {
-    const fullPath = path.join(schemaPath, fileFilter);
+    let fullPath = path.join(schemaPath, fileFilter);
+    // `glob.GlobSync` doesn't work correctly if the Glob pattern has the long path prefix for Windows systems
+    if (fullPath.startsWith("\\\\?\\")) {
```
ColinKerr, line 167: I took a look at the Glob docs and they make pretty clear that they do not support \ in paths (unless you set an option to force replace \\ with /)

They also seem to say they support UNC paths so long as you use //? not \\?

https://www.npmjs.com/package/glob#Windows

Further more we are using glob 7.1.2 which was released 7 years ago, the newest 7.x version was released 3 years ago (7.2.0).  Currently the latest is 10.x in the docs there is a note that v7 has buggy windows path support, search for `windowsPathsNoEscape`.  So I think the correct thing to do is to investigate updating iTwin.js to 10.x and investigating the various flags (like windowsPathsNoEscape) to support windows paths.

Stripping out the \\? really feels like a hack

akshay05n: In my previous attempt, I replaced all backslashes `\\` with forward slashes `/` because glob patterns use forward slashes for path separation. This didn't work. I also tried using the setting `windowsPathsNoEscape: true` with glob v7.1.2, but it wasn't effective either. To move forward, I'll update glob and test various options.

akshay05n: The glob package has been upgraded to its newest version (10.3.12). Since the updated API no longer supports `glob.sync`, references are changed to `globSync`. Additionally, all `\\` in glob patterns has been replaced with `/` as glob patterns should always use `/` as a path separator, even on Windows systems.

References:
https://github.com/isaacs/node-glob/issues/493
https://stackoverflow.com/questions/55512907/node-glob-sync-returns-empty-array

pmconne (approved):
In future, please provide an empty `rush change` message for changes like this that are uninteresting to end users.

ColinKerr (approved):

tools/internal/ios/ios.webpack.config.js
```diff
@@ -4,13 +4,16 @@
...
+const { globSync } = require("glob");
 
 module.exports = {
   mode: "development",
   entry: [
     path.resolve(__dirname, "scripts/configureMocha.js"),
-    ...glob.sync(path.resolve(__dirname, path.relative(__dirname, process.env.TESTS_GLOB))),
+    ...globSync(
+      path.resolve(__dirname, path.relative(__dirname, process.env.TESTS_GLOB)),
+      { windowsPathsNoEscape: true }
```
aruniverse, line 15: not required, as this should only be used on macs, but its ok to have

aruniverse (approved):

tools/internal/ios/ios.webpack.config.js
akshay05n: Removed the `windowsPathsNoEscape` option from the internal iOS tool

ben-polinsky (approved):

christophermlawson (approved):

wgoehrig (approved):

FlyersPh9 (approved):

PR #6593, opened 2024-04-08
make ClassRegistry.register api public

### LINKED ISSUE #6589: should ClassRegistry.register() be a public instead of internal ?
**Describe the bug**
If I want to register individual entityClass, there is no way.  Only way is to register that class as module using ClassRegistry.registerModule().

**Expected behavior**
Make a ClassRegistry.register() api public.

YashodipD:
Fixes #6589 plus various related documentation improvements.


pmconne:
Thanks, I'll add proper docs + run extract-api etc shortly.

PR #6603, opened 2024-04-10
Add extra cleanup functionality in PrimitiveTool

hl662:
A continuation of the memory leak we see on refreshes and navigation away from a viewer. 

In [ToolAdmin](https://github.com/iTwin/itwinjs-core/blob/be0a9dca5d53a34b10bc04a9253c7195049d7a8c/core/frontend/src/tools/ToolAdmin.ts#L339), we see PrimitiveTool objects still carrying references to [targetView](https://github.com/iTwin/itwinjs-core/blob/be0a9dca5d53a34b10bc04a9253c7195049d7a8c/core/frontend/src/tools/PrimitiveTool.ts#L27C10-L27C20), a Viewport object, preventing Viewport objects to be garbage collected, even after the ViewManager calls [onShutdown()](https://github.com/iTwin/itwinjs-core/blob/be0a9dca5d53a34b10bc04a9253c7195049d7a8c/core/frontend/src/ViewManager.ts#L139).

This PR removes that reference to the viewport object when the PrimitiveTool's cleanup function is called, and also adds this cleanup functionality to the ToolAdmin's onShutdown().

Also, refactored the onCleanup() of subclasses extending from PrimitiveTool, to call super.cleanup() after custom cleanup functionality - Some custom cleanup functionality might rely on the iModel the tool is associated with, which is accessed through the renderTarget's [reference](https://github.com/iTwin/itwinjs-core/blob/be0a9dca5d53a34b10bc04a9253c7195049d7a8c/core/frontend/src/tools/PrimitiveTool.ts#L38). We don't want to remove that reference until after custom cleanup functionality from the subclasses have ran.

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -10780,6 +10780,8 @@ export abstract class PrimitiveTool extends InteractiveTool {
     get iModel(): IModelConnection;
     isCompatibleViewport(vp: Viewport | undefined, isSelectedViewChange: boolean): boolean;
     isValidLocation(ev: BeButtonEvent, isButtonEvent: boolean): boolean;
+    // @internal (undocumented)
+    onCleanup(): Promise<void>;
     onRedoPreviousStep(): Promise<boolean>;
     onReinitialize(): Promise<void>;
     abstract onRestartTool(): Promise<void>;
@@ -17482,6 +17484,8 @@ export abstract class ViewTool extends InteractiveTool {
     exitTool(): Promise<void>;
     // (undocumented)
     inDynamicUpdate: boolean;
+    // (undocumented)
+    onCleanup(): Promise<void>;
     // (undocumented)
     onResetButtonUp(_ev: BeButtonEvent): Promise<EventHandled>;
     // (undocumented)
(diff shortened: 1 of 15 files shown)


pmconne:
This feels like whack-a-mole. Isn't the root problem that you've got IModelApp inside some React state that's repeatedly getting recreated?

hl662:
> This feels like whack-a-mole. Isn't the root problem that you've got IModelApp inside some React state that's repeatedly getting recreated?

You're right that it is whack-a-mole, but we can't dismiss the presence of moles, and the need to whack when we've found them. 

The root problem isn't the IModelApp being recreated - that's not the problem we're seeing anymore now that we are explictly calling onShutdown() in our viewer components. 

The problem is the instances where a viewport is unmounted and we want to dispose of the resources that the allocated ScreenViewport uses. The unmounting event comes from dropViewport(), which can can be called outside of IModelApp being shutdown. This particular case being ViewportComponent's [handleDisconnectFromViewManager() ](https://github.com/iTwin/appui/blob/master/ui/imodel-components-react/src/imodel-components-react/viewport/ViewportComponent.tsx).

After your comment, I went back to look at how ViewManager calls ToolAdmin's forgetViewport, and added a line where we call the current tool's onCleanup(), since the tools themselves also carry references to the viewport in use. I didn't want to set the tool's themselves to be undefined, in case the tools are to be reused in between viewports.

bbastings:
> This feels like whack-a-mole. Isn't the root problem that you've got IModelApp inside some React state that's repeatedly getting recreated?

Agree. You can't now require that sub-class call super when it was previously a no-op.

PR #6609, opened 2024-04-11
Property editing additions

christophermlawson:


core/ecschema-editing/src/Editing/Properties.ts
```diff
@@ -0,0 +1,271 @@
...
+   */
+  protected async getProperty<T extends MutablePropertyType>(classKey: SchemaItemKey, propertyName: string): Promise<T> {
+    let mutableClass: MutableClass;
+    try {
+      mutableClass = await this.getClass(classKey);
+    } catch (e: any) {
+      // TODO: Update error
+      throw new Error(e.message);
+    }
+
```
StefanApfel-Bentley, line 120: Instead of the any type hint, I'd propose to use the BentleyError helper function here:
```ts
} catch (e: unknown) {
  throw new Error(BentleyError.getErrorMessage(e));
}

File: core/ecschema-editing/src/Editing/Properties.ts [hunk 2]
```diff
@@ -0,0 +1,271 @@
...
+    const schemaItems = this._schemaEditor.schemaContext.getSchemaItems();
+    let { value, done } = schemaItems.next();
+    while (!done) {
+      if (await value.is(mutableClass)) {
+        if (!mutableClass.key.matches(value.key)) {
+          derivedClasses.push(value);
+        }
+      }
+      ({ value, done } = schemaItems.next());
+    }
```
StefanApfel-Bentley, line 102: Wouldn't a async for-loop work too here?
Also value.is only exists for classes, but getSchemaItems could return any kind of schema items, so we might to add a typecheck here too.
```ts
for await (const schemaItem of this._schemaEditor.schemaContext.getSchemaItems()) {
  if(ECClass.isECClass(schemaItem) && await schemaItem.is(mutableClass)) {
    // ...
  }
}
```

core/ecschema-editing/src/Editing/Properties.ts
```diff
@@ -0,0 +1,271 @@
...
+type MutablePropertyType = MutableProperty | MutableArrayProperty | MutablePrimitiveOrEnumPropertyBase;
+
+/**
+ * @alpha
+ * A class allowing editing of attributes of the base Property class.
+ */
+export class Properties {
+  public constructor(protected _schemaEditor: SchemaContextEditor) { }
+
+  public async setName(classKey: SchemaItemKey, existingPropertyName: string, newPropertyName: string) {
```
StefanApfel-Bentley, line 17: For consistency reasons, I'd rename `existingPropertyName` to `propertyName` to match the other signatures.
Also, since it's public, doesn't it require a function comment or isn't that necessary as the class is alpha?

core/ecschema-editing/src/Editing/Properties.ts
```diff
@@ -0,0 +1,271 @@
...
+
+/**
+ * @alpha
+ * A class extending Properties allowing editing of EnumerationProperty attributes.
+ */
+export class EnumerationProperties extends PrimitiveOrEnumProperties {
+  public constructor(_schemaEditor: SchemaContextEditor) {
+    super(_schemaEditor);
+  }
+}
```
StefanApfel-Bentley, line 270: Not sure we really need to have own editors if we don't add any new values to it.
But if we keep them, we should then also have an explicit editor for Struct- and Navigation properties.

StefanApfel-Bentley: I interpreted the code that we didn't add an editor for Navigation- and Struct-Properties because we don't want to allow to change what makes those two unique (struct/relationship reference), so the remain properties can be set through the common properties editor.

core/ecschema-editing/src/Editing/Properties.ts
```diff
@@ -0,0 +1,271 @@
...
+    switch (ecClass.schemaItemType) {
+      case SchemaItemType.EntityClass:
+      case SchemaItemType.Mixin:
+      case SchemaItemType.StructClass:
+      case SchemaItemType.CustomAttributeClass:
+      case SchemaItemType.RelationshipClass:
+        break;
+      default:
+        throw new ECObjectsError(ECObjectsStatus.InvalidSchemaItemType, `Schema item type not supported`);
+    }
```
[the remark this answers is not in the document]
christophermlawson: Yeah, I copied this code from ECClasses. Not sure why we did it this way.  I will update.

core/ecschema-editing/src/Editing/Properties.ts
christophermlawson: yes, much cleaner. thanks.

core/ecschema-editing/src/Editing/Properties.ts
christophermlawson: There more just place holders for now.  We'll need them in the future, especially based on our conversation with Colin.  Nav and Struct will be added in the future too.

core/ecschema-editing/src/Editing/Properties.ts
christophermlawson: done.

core/ecschema-editing/src/Editing/Properties.ts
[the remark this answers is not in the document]
christophermlawson: done

core/ecschema-editing/src/Editing/Properties.ts
christophermlawson: done

core/ecschema-editing/src/Editing/Properties.ts
```diff
@@ -0,0 +1,298 @@
...
+import { MutablePrimitiveOrEnumPropertyBase } from "./Mutable/MutablePrimitiveOrEnumProperty";
+import { MutableClass } from "./Mutable/MutableClass";
+
+type MutablePropertyType = MutableProperty | MutableArrayProperty | MutablePrimitiveOrEnumPropertyBase;
+
+/**
+ * @alpha
+ * A class allowing editing of attributes of the base Property class.
+ */
+export class Properties {
```
StefanApfel-Bentley, line 14: With introducing the Properties editor, we now have the several places where callers can edit certain property aspects. For instance the _PropertyCategories_ can be set in _ECClasse_s only, the property name can be set here and in ECClasses as well, ...

In my opinion, we should probably move all property related stuff from the ECClass editor into the properties editor as a new central place for everything property related as long all of this is still _alpha_.

I'm aware that this might break the merging API callers, but that should be easy to adopt.

core/ecschema-editing/src/Editing/ECClasses.ts
```diff
@@ -28,6 +29,23 @@ export class ECClasses {
 
   protected constructor(protected _schemaEditor: SchemaContextEditor) { }
 
+  /**
+   * Allows access for editing of base Property attributes.
+   */
+  public readonly properties = new Properties(this._schemaEditor);
```
StefanApfel-Bentley, line 35: Since enumerationProperties and primitiveProperties also rather act as placeholder. I wonder if we should not also add placeholders for the missing navigation- and struct- properties. Then we would not need to make the base Properties public, as this is all covered through inheritance. 

And, when overriding the getProperty method, we can even ensure we get the correct property type in the editor call.

core/ecschema-editing/src/Editing/Properties.ts
christophermlawson: Yeah, I thought the same.  Wasn't sure if it should be in this PR, but I will go ahead and move the API.

core/ecschema-editing/src/Editing/ECClasses.ts
christophermlawson: OK. I will add them.

StefanApfel-Bentley: If all property editor classes derive from Properties editor, there is no need to keep line 33, right?

christophermlawson: based on our conversation, can you resolve this?

core/ecschema-editing/src/Editing/Properties.ts
christophermlawson: My latest push has these changes.

StefanApfel-Bentley (approved):

PR #6611, opened 2024-04-11
use previously working node version until rush issue is fixed

ben-polinsky:
Latest nodejs security patch breaks rush integration on Windows: https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2/

This can get builds moving again until the issue is fixed. We can also hold if we think it's worth waiting for.

See:
- https://github.com/microsoft/rushstack/issues/4636
- https://github.com/node-red/node-red/pull/4652
- https://github.com/microsoft/rushstack/pull/4637


aruniverse (approved):

ben-polinsky:
4.6.6 snuck in, but is prevented from releasing due to the error I am fixing in this PR. Reverting that so we can get green once again.

ben-polinsky:
@mergifyio backport release/4.5.x

PR #6612, opened 2024-04-12
`@itwin/express-server` missing peer dep on `@itwin/core-common`

saskliutas:
`@itwin/express-server` uses `WebAppRpcProtocol` from `@itwin/core-common` in public API but does not have `@itwin/core-common` listed as peer dependency. 

https://github.com/iTwin/itwinjs-core/blob/fb405140fcff8bfb09a67b04d29491d48f0437cc/common/api/express-server.api.md?plain=1#L16

### DIFF
--- a/common/changes/@itwin/express-server/express_server_missing_peer_2024-04-12-07-57.json
+++ b/common/changes/@itwin/express-server/express_server_missing_peer_2024-04-12-07-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/express-server",
+      "comment": "Added missing peer dependency on `@itwin/core-common`.",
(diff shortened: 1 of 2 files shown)


paulius-valiunas (approved):
looks good, probably needs a rush update too

saskliutas:
> looks good, probably needs a rush update too

`rush updated` did not produce any changes. Windows and MacOS builds are failing for unrelated issues.

aruniverse:
The use of `WebAppRpcProtocol` as a `type` in express-server imo should be ok. It's the use of `BentleyCloudRpcConfiguration` that requires this to be a peer. Note, this is technically breaking, but I don't see how anyone could've used this package before without already having core-common installed as a peer

PR #6613, opened 2024-04-12
Upgrade core-backend to ES2022

pmconne:
**Don't merge - for demonstration purposes only**.

This produces a slew of test failures, occurring when inserting a model, because `modeledElement` is undefined.

Model constructor calls Entity constructor. When Entity constructor exits, `this.modeledElement` is valid. When execution returns to Model constructor, `this.modeledElement` is undefined.

The property gets assigned via black magic in [Entity.forEachProperty](https://github.com/iTwin/itwinjs-core/blob/fb405140fcff8bfb09a67b04d29491d48f0437cc/core/backend/src/Entity.ts#L58). Something in ES2022 changes the behavior there.

I had upgraded core-backend to target ES2022 in #6572 because TypeScript doesn't provide type declarations for `Intl.Segmenter` for older targets.

### DIFF
--- a/common/changes/@itwin/core-backend/pmc-es2022_2024-04-12-13-47.json
+++ b/common/changes/@itwin/core-backend/pmc-es2022_2024-04-12-13-47.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/core/backend/tsconfig.json
+++ b/core/backend/tsconfig.json
@@ -1,5 +1,8 @@
 {
   "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
+  "compilerOptions": {
+    "target": "ES2022"
+  },
   "include": [
     "./src/**/*.ts"
   ]


kabentley:
maybe the [constructor for Entity](https://github.com/iTwin/itwinjs-core/blob/fb405140fcff8bfb09a67b04d29491d48f0437cc/core/backend/src/Entity.ts#L54-L59) should be:
```ts
  protected constructor(props: EntityProps, iModel: IModelDb) {
    this.iModel = iModel;
    this.id = Id64.fromJSON(props.id);
    // copy all auto-handled properties from input to the object being constructed
    iModel.forEachMetaData(this.classFullName, true, (propName: string, meta: PropertyMetaData) => (this as any)[propName] = meta.createProperty((props as any)[propName]), false);
  }
```
I don't know if that will help but we won't be permuting `this` while we're calling a method on it.

GytisCepk:
TS flag `useDefineForClassFields` default value switches to `true` with ES2022 or higher ([`useDefineForClassFields` TypeScript docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier)). This causes values assigned with `Entity.forEachProperty` to get overridden with `undefined`.

Currently, `useDefineForClassFields` is explicitly set to `false` in config, but should be set to `true` to make our TS code more consistent with JS standard.

Issue: https://github.com/iTwin/itwinjs-core/issues/7474
PR: https://github.com/iTwin/itwinjs-core/pull/7563

aruniverse:
@pmconne is this draft pr / branch still needed with #7563 merging?

pmconne:
> @pmconne is this draft pr / branch still needed with #7563 merging?

Shouldn't be.

PR #6627, opened 2024-04-15
Update mysql2 to 3.9.4 for CVE-2024-21508

nick4598:


aruniverse (approved):
approving, but any reason why we are updating 4.4.x branch?

PR #6631, opened 2024-04-16
Drawing view fixes

Alfonso-Martello:
The "fit volume" for a section drawing view now includes the extents of the section drawing attachment.

The maximum extents of a drawing view were previously defined as 2x the extents of the drawing model. This is an empty range if the model is empty, and counter-productive when you're trying to add content to the model (you can't zoom out far enough to, e.g., place a line of the desired length; instead you get a "Maximum Window" error). Removed this restriction.

Fixed computed extents for rendering a spatial view into a section drawing view to resolve front/back plane clipping.

Added a keyin to display-test-app to create a section drawing from the current spatial view, for illustration and testing purposes.


test-apps/display-test-app/src/frontend/SectionDrawingApi.ts
```diff
@@ -0,0 +1,344 @@
...
+      jsonProperties: {
+        styles: {
+          backgroundColor: ColorDef.white.tbgr,
+        },
+      },
+    };
+
+    // const rotation = spatialViewState.rotation;
+
+    // const transform = Transform.createOriginAndMatrix(spatialViewState.origin, rotation).inverse();
```
pmconne, line 314: @Alfonso-Martello it took me far too long to realize you were recomputing an incorrect drawing-to-spatial transform here instead of using the one passed in.

I have fixed  the view extents such that the spatial view will always display in the center of the drawing view. There's still a bit of front plane clipping in some cases, I will circle back on figuring that out in a bit.

Alfonso-Martello: Thanks @pmconne,

I think we added that early on while trying something out, and then forgot to delete it (thereby messing up all our future attempts).

@Ellord207 and I reviewed your fix and understand what it is doing.

Regarding the plane clipping, there is another .bim file I wanted to share with you as I think it shows back plane clipping occurring as well. I sent you an email titled "Back clipping example .bim" with the .bim file and more details.

pmconne: Was it created using the current version of the code in this PR?

Alfonso-Martello: Yes

pmconne:
The viewport expands the ViewState's extents (and moves its origin) via ViewingSpace. I added the relevant logic to do the same before pulling the frustum out of your ViewState. This fixes the clipping issues you're observing and preserves the centering in the drawing view.

However, if you zoom in very closely in a view that's not a top or bottom view, you'll still get clipping. I haven't figured out why yet. It computes the same expanded frustum as the viewport does.

pmconne:
Awaiting ImageTests results.

markschlosseratbentley (approved):

pmconne:
ImageTests results: this fixes some previously-unnoticed back plane clipping in Revit section drawings. @DStradley we will want to update reference images.

PR #6634, opened 2024-04-17
Documentation to Remove/Replace KindOfQuantity from an ECProperty

akshay05n:


docs/bis/ec/update-koq-on-property.md
```diff
@@ -0,0 +1,186 @@
...
+	</ECCustomAttributes>
+</ECProperty>
+```
+
+## Replacing KindOfQuantity of a property
+
+Changing the KindOfQuantity of an ECProperty to another KindOfQuantity with a different persistence unit is allowed by using the `AllowUnitChange` [Custom Attribute](./ec-custom-attributes.md) on the property. However, `AllowUnitChange` [Custom Attribute](./ec-custom-attributes.md) must have non-empty From and To properties.
+
+Additionally, the 'From' value must match the persistence unit from the old KindOfQuantity and the 'To' value must match the persistence unit from the new KindOfQuantity.
+
```
diegoalexdiaz, line 82: This article should also clarify that existing values on the property will not be changed automatically upon a KOQ update.

docs/bis/ec/update-koq-on-property.md
```diff
@@ -0,0 +1,186 @@
...
+```xml
+<ECProperty propertyName='Length' typeName='double'>
+	<ECCustomAttributes>
+		<AllowUnitChange xmlns="SchemaUpgradeCustomAttributes.01.00.00">
+			<From>u:M</From>
+			<To></To>
+		</AllowUnitChange>
+	</ECCustomAttributes>
+</ECProperty>
+```
```
ColinKerr, line 75: I don't think you need all these explicit examples.  Just briefly explain that From must be set to the source unit and To must be there but empty.

docs/bis/ec/update-koq-on-property.md
```diff
@@ -0,0 +1,186 @@
...
+	<ECProperty propertyName='Length' typeName='double' kindOfQuantity='KindOfQuantity2' >
+		<ECCustomAttributes>
+			<AllowUnitChange xmlns="SchemaUpgradeCustomAttributes.01.00.00">
+				<From>u:CM</From>
+				<To></To>
+			</AllowUnitChange>
+		</ECCustomAttributes>
+	</ECProperty>
+</ECEntityClass>
+```
```
ColinKerr, line 186: Same here, no need to explicitly show all invalid options ... just list what the requirements are

docs/bis/ec/update-koq-on-property.md
ColinKerr: I agree

docs/bis/ec/update-koq-on-property.md
```diff
@@ -0,0 +1,186 @@
+# KindOfQuantity
```
ColinKerr, line 1: This file should be linked to from the main KOQ page

docs/bis/ec/update-koq-on-property.md
```diff
@@ -0,0 +1,81 @@
+# Updating KindOfQuantity on an ECProperty
+
+This document provides instructions for iModel users on how to remove or replace a [KindOfQuantity (KoQ)](./kindofquantity.md) associated with an [ECProperty](./ec-property.md) in an iModel schema.
```
ColinKerr, line 3:
```suggestion
This document provides instructions for iModel users on how to change a properties persistence unit by removing or replacing the [KindOfQuantity (KoQ)](./kindofquantity.md) associated with an [ECProperty](./ec-property.md) in an iModel schema.
```

docs/bis/ec/kindofquantity.md
```diff
@@ -108,4 +108,10 @@ The below represents the string literal syntax. All italicized values are to be
   ```
   - If the unit label override is kept unset `TestFormat[TestUnitMajor]`, the final label will default to the unit's display label `TestUnitMajorLabel`.
   - If the unit label override is set to an empty string `TestFormat[TestUnitMajor|]`, the final label will set to an empty string.
-  - If the unit label override is set to a valid `TestFormat[TestUnitMajor|centimetre]`, the final label will be the overriden string `centimetre`.
\ No newline at end of file
+  - If the unit label override is set to a valid `TestFormat[TestUnitMajor|centimetre]`, the final label will be the overriden string `centimetre`.
+
+### Updating KindOfQuantity on ECProperty
+
+In some scenarios, you may need to remove or replace the KindOfQuantity assigned to an [ECProperty](./ec-property.md) within your iModel schema. Such modifications can help correct errors in the initial schema design.
```
ColinKerr, line 115:
```suggestion
The KindOfQuantity assigned to a property can be switch to any other KindOfQuantity that has the same persistence unit.  In some scenarios, you may need to remove or change the persistence unit of an [ECProperty](./ec-property.md) within your iModel schema.
```

docs/bis/ec/update-koq-on-property.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Note**: Removing or Replacing KindOfQuantity from a property does not alter other attributes or the fundamental data type of the property.
```
ColinKerr, line 5:
```suggestion
## Overview

Normally changing the persistence unit of a property is not supported unless you change the major schema version.  So you can change the KOQ applied to a property so long as it does not change the persistence unit.  Applying the `AllowUnitChange` custom attribute to a property allows you to change the persistence unit with only a minor version change.  

| Change | AllowUnitChange required|
|-|-|
| Add KOQ to property that has none | not required |
| Change KOQ to KOQ with same persistence unit | not required |
| Remove KOQ from property that has one | required |
| Change KOQ to KOQ with different persistence unit | required | 

The `AllowUnitChange` custom attribute is intended for cases where the persistence unit is incorrect but the persisted value is correct.  Changing the persistence unit only updates the metadata, it **does not transform existing values**.  If conversion is needed you must do it yourself and consider updating the write (middle) schema version to make it clear that the persisted data format has changed.
```

docs/bis/ec/update-koq-on-property.md
ColinKerr: now covered in my overview suggestion

ColinKerr (approved):

PR #6622, opened 2024-04-15
temporarily disable mobile builds

aruniverse:
temporarily disable mobile builds until we are able to release native-android/ios again

PR #6642, opened 2024-04-17
Re-enable mobile builds

toddsouthenbentley:
Enable mobile builds again now that mobile-native-android/ios are fixed.

Previous PR to disable them: #6622


aruniverse (approved):

toddsouthenbentley:
@Mergifyio backport release/4.5.x

core/backend/src/test/annotations/TextAnnotation.test.ts
```diff
@@ -156,173 +156,177 @@ describe("layoutTextBlock", () => {
...
-  });
-
-  function expectLines(input: string, width: number, expectedLines: string[]): TextBlockLayout {
-    const textBlock = TextBlock.create({ styleName: "" });
-    textBlock.width = width;
-    const run = makeTextRun(input);
-    textBlock.appendRun(run);
+  // Node in the mobile add-on does not include Intl, so these tests fail. Right now, mobile users
+  // are not expected to do any editing, but long term we will attempt to find a better solution.
+  if (!ProcessDetector.isMobileAppFrontend) {
```
ben-polinsky, line 161: Looks like you brought along a bunch of formatting changes as well, maybe revert those so we keep our author commit history/blame in tact?

pmconne: Should be able to do something like the following inside a test:
```
  if (ProcessDetector.isMobileAppFrontend) {
    this.skip();
  }
```

tcobbs-bentley: @pmconne, that's not working here, and I have no idea why. For some reason, `this` may be undefined. If I uses `this?.skip()` I get a different error: Property 'skip does not exist on type 'never'.

tcobbs-bentley: All the other places that call this.skip() use async function () instead of () =>. I guess I could try that.

PR #6646, opened 2024-04-18
Rework Workspace and Settings apis

kabentley:
Reviewers: please first consult Workspace.md and WorkspaceExamples.test.ts.

---

Make the apis for loading Settings and WorkspaceResources easier to use. It makes it possible to store SettingsDictionaries in cloud workspaces and load them automatically when you open an iModel.

This includes breaking changes to the (beta) apis for:
 - Settings Schemas
 - Settings Dictionaries
 - Workspaces
 - CloudSqlite

It also introduces the Workspace.Editor namespace that can be used to modify Workspaces.

@johnnyd710 this will require work for the Settings Editor. It enforces that all Settings defined by a SettingSchema must all have the same prefix. That will undoubtedly break existing code in iTS, since settings will have a different name. You also need to add the concept of storing SettingsDictionaries in cloud containers, as we discussed.

Also, @johnnyd710, @wgoehrig, @MichaelSwigerAtBentley  we should talk about how the settings editor will allow picking cloud-based WorkspaceDbs. We will probably need some kind of queries from the BlobContainer service.


core/backend/src/assets/Settings/Schemas/Workspace.Schema.json
```diff
@@ -1,53 +1,93 @@
 {
-  "$id": "http://itwinjs.org/json-schema/workspace",
-  "$schema": "http://json-schema.org/draft-07/schema",
-  "title": "Workspace settings",
-  "description": "settings for accessing data from workspaces",
-  "type": "object",
-  "groupName": "workspace",
-  "properties": {
-    "workspace/databases": {
+  "$schema": "itwin-settings/schema/1.1",
```
johnnyd710, line 2: Keith, what is `itwin-settings/schema/1.1` and does it even exist somewhere? If it does not exist, is intellisense (auto-complete) not available in your IDE when creating a new schema?

core/backend/src/assets/Settings/Schemas/Gcs.schema.json
```diff
@@ -1,32 +1,28 @@
 {
-  "$id": "http://itwinjs.org/json-schema/gcs",
-  "$schema": "http://json-schema.org/draft-07/schema",
```
johnnyd710, line 3: Using the JSON spec was nice because we were [generating types from our settings schemas](https://github.com/iTwin/studio/blob/main/packages/framework/src/common/settings/SettingsSchema.d.ts) automatically. It was useful, for example, if my setting was an enum, while developing my app I'll see the enum values in my editor. Likewise if my setting is an object, I'll know the keys.

Now with this custom schema (with settingDefs and typeDefs), [we won't be able to do that automatically](https://github.com/iTwin/studio/blob/main/packages/framework/scripts/create-settings-schema-type.js).

core/backend/src/workspace/Workspace.ts
```diff
@@ -766,9 +990,144 @@ class WorkspaceDbImpl implements WorkspaceDb {
       throw new Error("no cloud container to prefetch");
     return CloudSqlite.startCloudPrefetch(cloudContainer, this.dbFileName, opts);
   }
+
+  public queryResource(search: WorkspaceResource.Search, callback: (result: WorkspaceResource.SearchResult) => void | "stop"): void {
+    this.withOpenDb((db) => {
+      db.withSqliteStatement(`SELECT value from ${search.resourceType}s WHERE id ${search.nameCompare ?? "="} ?`, (stmt) => {
+        stmt.bindString(1, search.nameSearch);
+        while (DbResult.BE_SQLITE_ROW === stmt.step()) {
+          if (callback({ workspaceDb: this, rscName: stmt.getValueString(0) }) === "stop")
```
williamkbentley, line 999: what if i want the name of the resource as well?

core/backend/src/workspace/Workspace.ts
```diff
@@ -766,9 +990,144 @@ class WorkspaceDbImpl implements WorkspaceDb {
...
+      public static async initializeWorkspace(args: { props: CloudSqlite.ContainerProps, dbName: WorkspaceDb.DbName, manifest: WorkspaceDb.Manifest }) {
+        const dbFullName = WorkspaceContainer.makeDbFileName(args.dbName, "1.0.0");
+        return super._initializeDb({ ...args, dbName: dbFullName, dbType: WorkspaceSqliteDb, blockSize: "4M" });
+      }
+    }
+    return CloudAccess.initializeWorkspace(args);
+  }
+
+  public async createNewCloudContainer(args: Workspace.Editor.CreateNewContainerProps): Promise<Workspace.Editor.Container> {
+    const service = BlobContainer.service;
```
williamkbentley, line 1027: again, where is this instantiated?

core/backend/src/assets/Settings/Schemas/Gcs.schema.json
```diff
@@ -1,32 +1,28 @@
...
+      "description": "array of GCS WorkspaceDbs",
       "items": {
-        "type": "string"
-      },
-      "default": [
-        "gcs/base",
-        "gcs/entire-earth"
-      ]
+        "type": "object",
+        "extends": "itwin/core/workspace/workspaceDb",
```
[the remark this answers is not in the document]
williamkbentley: nevermind. i see in SettingsSchema. but i dont see this typedef defined anywhere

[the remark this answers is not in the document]
williamkbentley: and now i see that in Workspace.Schema.json. Does that imply that that particular json file must be parsed first?

core/backend/src/CloudSqlite.ts
```diff
@@ -37,8 +37,10 @@ export namespace CloudSqlite {
    * If the service is unavailable or returns an error, an empty token is returned.
    */
   export async function requestToken(args: RequestTokenArgs): Promise<AccessToken> {
-    // allow the userToken to be supplied via Rpc. If not supplied, or blank, use the backend's accessToken.
-    const userToken = args.userToken ? args.userToken : (await IModelHost.getAccessToken());
+    // allow the userToken to be supplied via args. If not supplied, or blank, use the backend's accessToken. If that fails, use the value from the current RPC request
+    let userToken = args.userToken ? args.userToken : await IModelHost.getAccessToken();
+    if (userToken === "")
+      userToken = RpcTrace.currentActivity?.accessToken ?? "";
     if (BlobContainer.service === undefined) {
```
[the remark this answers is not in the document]
kabentley: It must be supplied by the application that calls IModelHost.startup. It will come from iTS, usually.

core/backend/src/assets/Settings/Schemas/Gcs.schema.json
kabentley: we should talk. I don't know if you can do what i did with the "extends" keyword in JSONSchema. Maybe you can, that whole JSONSchema thing seems like it was invented by amateurs haphazardly, so I gave up.

core/backend/src/assets/Settings/Schemas/Workspace.Schema.json
```diff
@@ -1,53 +1,93 @@
...
+        "baseUri"
+      ],
+      "properties": {
+        "dbName": {
+          "type": "string",
+          "description": "the name of the database within its cloud container. If not present, defaults to 'workspace-db'."
+        },
+        "baseUri": {
+          "type": "string",
+          "description": "the baseUri for the container, without trailing slash (e.g., https://containers.itwinjs.org)"
```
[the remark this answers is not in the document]
kabentley: That's an excellent question. I meant to ask @wgoehrig about that. I wasn't sure that there is a single URL or one for each datacenter.

core/backend/src/assets/Settings/Schemas/Gcs.schema.json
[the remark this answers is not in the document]
kabentley: The definitions aren't parsed except by a "Settings Editor" or during validation checks at runtime. In both cases all schema definitions must be loaded before they start.

core/backend/src/workspace/Workspace.ts
kabentley: good point, you will need that for anything other than "=".

core/backend/src/workspace/Workspace.ts
```diff
@@ -706,15 +908,33 @@ class WorkspaceDbImpl implements WorkspaceDb {
   public close() {
     if (this.isOpen) {
       this.onClose.raiseEvent();
-      this.sqliteDb.closeDb();
+      this.sqliteDb.closeDb(true); // save changes, if present
```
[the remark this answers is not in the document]
kabentley: not really, but it should have been in EditableDb anyway. I moved it there.

core/backend/src/assets/Settings/Schemas/Workspace.Schema.json
```diff
@@ -1,53 +1,93 @@
...
-            "description": "the container name"
-          },
-          "version": {
-            "type": "string",
-            "description": "the (semver) range of acceptable versions"
-          },
-          "includePrerelease": {
-            "type": "boolean",
-            "description": "include prerelease version as acceptable versions"
+            "description": "the name of the string resource holing the settings dictionary in this WorkspaceDb",
```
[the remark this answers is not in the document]
kabentley: fixed

core/backend/src/CloudSqlite.ts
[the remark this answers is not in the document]
johnnyd710: Should iTS itself initialize it if the startup app doesn't?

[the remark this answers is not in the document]
kabentley: Yes, it should be delivered and installed (much like the CodeService and IModelHub backend packages) by iTS. There should not be any startup app involvement (in fact, it shouldn't be possible to set it from the startup app).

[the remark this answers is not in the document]
johnnyd710: https://github.com/iTwin/studio/pull/1705

core/backend/src/assets/Settings/Schemas/Workspace.Schema.json
johnnyd710: I think we should at least replace `itwin-settings/schema/1.1` with an actual "meta" or "base" JSON Schema that defines what "settingDef", "typeDef", and "groupName" are, and host that JSON schema somewhere so you can reference it like "http://somewhere/itwin-settings/schema/1.1"

E.g.,
```json5
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "additionalProperties": false,
  "properties": {
    "settingDefs": {
      "type": "object",
      "description": "A setting definition"
    },
    "typeDefs": {
      "type": "object",
      "description": "A type definition that can be used in the settingDefs, e.g., 'extends': 'itwin/core/workspace/workspaceDb'"
    },
    "groupName": { // or change this to "schemaId"
      "type": "string",
      "description": "The unique id of the schema"
    }
  }
}
```

core/backend/src/assets/Settings/Schemas/Gcs.schema.json
johnnyd710: Keith, while I did find a way to make it work with JSON Schema, its probably not worth the trade-off. Yes, you'll be able to use JSON-to-Typescript and have the types, but now you have to deal with JSON Schema and all the extra baggage and properties that we don't use or need.

core/backend/src/assets/Settings/Schemas/Workspace.Schema.json
kabentley: OK, but should that also be in a local meta-meta-schema file? 

Shouldn't the descriptions of `settingDefs` and `typeDefs` be plural (e.g. "A group of Setting definitions", "A group of Type definitions that each can be used...")?

How 'bout `SchemaPrefix` instead of `groupName` (Id sounds like a number)?

johnnyd710: > OK, but should that also be in a local meta-meta-schema file?

As long as its accessible to everyone. One thing I've seen ppl do is reference it directly from Github that way you don't have to go through the pain of hosting it somewhere else:

```json5
// my app's settings.schema.json
{
  "$schema": "https://raw.githubusercontent.com/iTwin/itwinjs-core/master/core/backend/src/assets/Settings/Base.Schema.json",
  // ... now I get intelli-sense for the rest of my file!
  "settingsDef": {
    // ... etc 
  }
}
```

> Shouldn't the descriptions of `settingDefs` and `typeDefs` be plural (e.g. "A group of Setting definitions", "A group of Type definitions that each can be used...")?

Yes you're right

> How 'bout `SchemaPrefix` instead of `groupName` (Id sounds like a number)?

That sounds right

kabentley: Will you make these changes? I'm looking for volunteers to take over this PR...

johnnyd710: Yeah, I can

full-stack-tests/backend/src/integration/CloudWorkspace.test.ts
```diff
@@ -7,156 +7,278 @@ import "./StartupShutdown"; // calls startup/shutdown IModelHost before/after al
...
+    await withPatchVersion(styles2, (editDb) => {
+      defineStyles(editDb, "batch2", 200);
+    });
+
+    const dbs: WorkspaceDb.CloudProps[] = [
+      { ...styles1.cloudProps!, loadingHelp: "if problems with styles1", prefetch: true },
+      { ...styles1.cloudProps!, containerId: "styles3", loadingHelp: "if problems with styles3" },
+      { ...styles2.cloudProps!, loadingHelp: "if problems with styles2" },
+      { ...styles1.cloudProps!, loadingHelp: "if problems with styles1", prefetch: true },
+    ];
```
williamkbentley, line 234: In practice, were would this list of Cloud WorkspaceDbs come from? iModelSettings?

core/backend/src/assets/Settings/Schemas/Workspace.Schema.json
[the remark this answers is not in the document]
williamkbentley: container service appears to be an internal-only API, so I guess it wouldn't be appropriate to go in the open source stuff. Also, the baseUri the container service returns does appear to be region specific.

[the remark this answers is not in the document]
williamkbentley: so, seems to me like no default is correct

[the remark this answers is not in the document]
kabentley: ok, that makes sense. Please resolve this conversation.

full-stack-tests/backend/src/integration/CloudWorkspace.test.ts
kabentley: please see latest

kabentley:
This still needs more documentation in Workspace.md and lots of examples, but the api is now ready for review.

core/backend/src/workspace/Workspace.ts
kabentley: see tests

full-stack-tests/backend/src/integration/CloudWorkspace.test.ts
```diff
@@ -7,156 +7,162 @@ import "./StartupShutdown"; // calls startup/shutdown IModelHost before/after al
...
 import * as fs from "fs-extra";
 import { join } from "path";
-import { BaseSettings, CloudSqlite, EditableWorkspaceDb, IModelHost, IModelJsFs, SettingsPriority, Workspace } from "@itwin/core-backend";
-import { assert } from "@itwin/core-bentley";
+import { IModelHost, IModelJsFs, SettingObject, Settings, StandaloneDb, Workspace, WorkspaceContainer, WorkspaceDb, WorkspaceSettings } from "@itwin/core-backend";
+import { assert, Guid } from "@itwin/core-bentley";
 import { AzuriteTest } from "./AzuriteTest";
 
-describe("Cloud workspace containers", () => {
+describe.only("Cloud workspace containers", () => {
```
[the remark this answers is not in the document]
kabentley: removed

full-stack-tests/presentation/src/frontend/Content.test.snap
grigasp, line 1: I believe the two new `*.test.snap` files under `full-stack-tests/presentation/src/frontend` are not necessary. Probably were generated before merging with `master`.

grigasp (changes requested):

full-stack-tests/presentation/src/frontend/Content.test.snap
kabentley: I deleted the two that were created by this PR. There is still "SelectionScopes.test.snap" there. Is that one OK?

grigasp: yes, that one is okay

pmconne:
I'll be taking over this PR. I am currently familiarizing myself with it. It could help to adjust the PR description to describe what specific changes were made for what specific purposes, as many of the changes appear to be cosmetic. e.g., in what specific ways does this make settings "easier to use".

kabentley:
> ways does this make settings "easier to use".

It doesn't make "settings" easier to use, it makes "loading Settings and WorkspaceResources easier to use". The previous approach had a setting called "cloud/containers" that was an array of container properties with an "alias". Then, there was a setting called "workspace/databases" that was an array of workspaceDb properties that referenced entries in the "cloud/container" array through their alias. This eliminates both of those in favor of the new setting type "itwin/core/workspace/workspaceDbList", making the api much simpler.

I will describe the other changes separately.

docs/learning/backend/Workspace.md
```diff
@@ -26,20 +26,68 @@ In the list above, later entries tend to change more frequently and, in the case
   1. [Settings](#settings)
   2. [WorkspaceDb](#workspaceDbs)
 
-`Settings` are *named parameters* that an application defines and whose values are supplied at runtime. `WorkspaceDb`s hold *named resources* (i.e. data) that the application uses. `Settings` and `WorkspaceDb`s are often related in application logic, e.g.:
+`Settings` are *named parameters* that an application defines via `SettingSchema`s and whose values are determined via a "Settings Editor" and supplied at runtime via one or more [Settings.Dictionary]($backend). `WorkspaceDb`s hold *named resources* (i.e. data) that the application uses. `Settings` and `WorkspaceDb`s are often related in application logic, e.g.:
 
-- a Setting may contain the "formula" to find a resource
-- a `WorkspaceDb`s may hold a resource that defines a group of `Settings`
+- Settings may contain the properties necessary to find a resource, including a list of `WorkspaceDb` to search
+- `WorkspaceDb`s may hold a [Settings.Dictionary]($backend) resource that defines a group of `Settings`
```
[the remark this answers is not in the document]
kabentley: Yes. See [here](https://github.com/iTwin/itwinjs-core/blob/aec5cdb3a898f6b11d9b3538c6f67e1b2f0b2ed4/core/backend/src/workspace/Workspace.ts#L991-L992)

core/backend/src/CloudSqlite.ts
```diff
@@ -37,8 +37,10 @@ export namespace CloudSqlite {
    * If the service is unavailable or returns an error, an empty token is returned.
    */
   export async function requestToken(args: RequestTokenArgs): Promise<AccessToken> {
-    // allow the userToken to be supplied via Rpc. If not supplied, or blank, use the backend's accessToken.
-    const userToken = args.userToken ? args.userToken : (await IModelHost.getAccessToken());
+    // allow the userToken to be supplied via args. If not supplied, or blank, use the backend's accessToken. If that fails, use the value from the current RPC request
+    let userToken = args.userToken ? args.userToken : await IModelHost.getAccessToken();
```
kabentley, line 41: This change allows CloudSqlite files to be opened by web backends using the credentials supplied by a single web client (vs. Ipc backends where the credentials are stored in IModelHost) to obtain an access token. The only previous use of this was for ViewStores, but workspaces also will need user credentials. See corresponding change in IModelReadRpcImpl.ts.

(Note: GCS workspaces are public, hence didn't need an access token)

core/backend/src/CloudSqlite.ts
```diff
@@ -103,31 +105,33 @@ export namespace CloudSqlite {
   }
   export interface ContainerProps {
     /** The type of storage provider. */
-    storageType: "azure" | "google";
+    readonly storageType: "azure" | "google";
```
kabentley, line 108: these member should always have been readonly. But we now store ContainerProps in memory to identify containers. This change makes it so they're immutable and therefore reliable identifiers.

core/backend/src/CloudSqlite.ts
```diff
@@ -824,23 +828,41 @@ export namespace CloudSqlite {
...
       });
       container.disconnect({ detach: true });
     }
 
+    /**
+     * Create a new BlobContainer from the BlobContainer service to hold one or more VersionedSqliteDbs.
+     * @returns A ContainerProps that describes the newly created container.
+     * @note the current user must have administrator rights to create containers.
+     */
+    protected static async createBlobContainer(args: Omit<BlobContainer.CreateNewContainerProps, "userToken">): Promise<CloudSqlite.ContainerProps> {
```
kabentley, line 853: There was previously never a way to create a blob container and also initialize it for a CloudSqlite use. Previously you had to create the container first, and then call "initialize" for the type of CloudSqlite client. This new capability is important for WorkspaceDbs, but i also added it for ViewStores and PropertyStores (though admittedly that has nothing to do with this PR).

core/backend/src/IModelDb.ts
```diff
@@ -1429,25 +1431,55 @@ export abstract class IModelDb extends IModel {
...
         } catch (e) {
-          UnexpectedErrors.handle(e);
+          Workspace.exceptionDiagnosticFn(e as WorkspaceDb.LoadError);
         }
       }
     });
   }
 
+  /** @internal */
+  protected async loadWorkspaceSettings() {
```
kabentley, line 1452: Whenever we open an IModel, we now see if any of its internal setting dictionaries has a setting called "settingsWorkspaces". If so we attempt to load settings from those WorkspaceDbs. Since this happens on "open", we don't throw any exceptions if there are problems and instead report those would-be exceptions to the host application (e.g. iTS) via the global variable `Workspace.exceptionDiagnosticFn`. I expect them to be shown in a diagnostics panel?

core/backend/src/workspace/Settings.ts
```diff
@@ -261,62 +287,71 @@ export class BaseSettings implements Settings {
...
+        all.push({ value, dictionary });
+        return undefined;
+      },
     });
 
     return all;
   }
 
+  // get the setting and verify the result is either undefined or the correct type. If so, return it. Otherwise throw an exception.
+  private getResult<T extends SettingType>(name: SettingName, expectedType: string) {
```
kabentley, line 320: This is a change to the error handling of the "getXXX" methods for Settings. Previously settings were not validated against the SettingSchema when their values were returned. If you called `getString`, and the found value was a string it was returned, regardless of whether the schema indicated it should be a string. Also, if you called `getString` and the result was a type other than `string` it returned the default value. Now, settings are always validated against the schema (if one exists) and an exception is thrown if the value is not the expected type. This change affected several tests.

core/backend/src/workspace/Workspace.ts
```diff
@@ -102,22 +88,27 @@ export namespace WorkspaceContainer {
...
 }
 
 /** @beta */
 export namespace WorkspaceDb {
-  /** The name of a WorkspaceDb in a "workspace/databases" setting. */
-  export type Name = string;
+  /** complete a WorkspaceDb name, replacing undefined with the default value "workspace-db"
+   * @internal
+   */
+  export const dbNameWithDefault = (dbName?: WorkspaceDb.DbName) => dbName ?? "workspace-db";
```
kabentley, line 111: Most likely WorkspaceContainers will hold one and only one WorkspaceDb (i.e. they are 1:1). The previous api had "workspace db name" as a required field in many places. To make the most common scenario easier, there is now a "default name" (workspace-db) that is used if no name is provided.

pmconne:
To investigate: error when starting display-test-app electron:
```
(node:16904) UnhandledPromiseRejectionWarning: Error: parsing SettingSchema file "/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/assets/Settings/Schemas/Cloud.Schema.json": settings group has no "schemaPrefix" member"
    at SettingsSchemas.addFile (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/workspace/SettingsSchemas.js:144:19)
    at SettingsSchemas.addDirectory (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/workspace/SettingsSchemas.js:152:22)
    at IModelHost.initializeWorkspace (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/IModelHost.js:207:43)
    at IModelHost.startup (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/IModelHost.js:242:14)
    at IpcHost.startup (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/IpcHost.js:89:39)
    at NativeHost.startup (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/NativeHost.js:129:33)
    at ElectronHost.startup (/home/paul/c/j2/s/itwinjs/core/electron/lib/cjs/backend/ElectronHost.js:211:41)
    at initializeDtaBackend (/home/paul/c/j2/s/itwinjs/test-apps/display-test-app/lib/backend/Backend.js:207:46)
    at async dtaElectronMain (/home/paul/c/j2/s/itwinjs/test-apps/display-test-app/lib/backend/DtaElectronMain.js:55:5)
(Use `electron --trace-warnings ...` to show where the warning was created)
(node:16904) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
```

kabentley:
> To investigate: error when starting display-test-app electron:
> 
> ```
> (node:16904) UnhandledPromiseRejectionWarning: Error: parsing SettingSchema file "/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/assets/Settings/Schemas/Cloud.Schema.json": settings group has no "schemaPrefix" member"
>     at SettingsSchemas.addFile (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/workspace/SettingsSchemas.js:144:19)
>     at SettingsSchemas.addDirectory (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/workspace/SettingsSchemas.js:152:22)
>     at IModelHost.initializeWorkspace (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/IModelHost.js:207:43)
>     at IModelHost.startup (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/IModelHost.js:242:14)
>     at IpcHost.startup (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/IpcHost.js:89:39)
>     at NativeHost.startup (/home/paul/c/j2/s/itwinjs/core/backend/lib/cjs/NativeHost.js:129:33)
>     at ElectronHost.startup (/home/paul/c/j2/s/itwinjs/core/electron/lib/cjs/backend/ElectronHost.js:211:41)
>     at initializeDtaBackend (/home/paul/c/j2/s/itwinjs/test-apps/display-test-app/lib/backend/Backend.js:207:46)
>     at async dtaElectronMain (/home/paul/c/j2/s/itwinjs/test-apps/display-test-app/lib/backend/DtaElectronMain.js:55:5)
> (Use `electron --trace-warnings ...` to show where the warning was created)
> (node:16904) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
> ```

That file was deleted. You probably need to do an "npm clean"

core/backend/src/IModelDb.ts
```diff
@@ -223,17 +226,17 @@ const withBriefcaseDb = async (briefcase: OpenBriefcaseArgs, fn: (_db: Briefcase
...
+  protected override verifyPriority(priority: Settings.Priority) {
+    if (priority <= Settings.Priority.application)
       throw new Error("Use IModelHost.appSettings");
   }
 
-  // attempt to resolve a setting from this iModel's settings, otherwise use appWorkspace's settings, otherwise defaultValue.
-  public override resolveSetting<T extends SettingType>(name: SettingName, resolver: SettingResolver<T>, defaultValue?: T): T | undefined {
-    return super.resolveSetting(name, resolver) ?? IModelHost.appWorkspace.settings.resolveSetting(name, resolver, defaultValue);
+  // attempt to resolve a setting from this iModel's settings, otherwise optionally use appWorkspace's settings, otherwise defaultValue.
+  public override resolveSetting<T extends SettingType>(arg: { settingName: SettingName, resolver: Settings.Resolver<T>, skipAppWorkspace?: boolean }, defaultValue?: T): T | undefined {
```
pmconne, line 238: I don't see any way that a user of `IModelDb.workspace.settings` could pass the `skipAppWorkspace` flag, which exist only here on this non-exported type. Is it useful and if so does it belong in the base interface as something like `noCascade`?

docs/learning/backend/Workspace.md
```diff
@@ -444,32 +257,34 @@ Possible resource types are:
...
 - `file` resources that hold arbitrary files. They may be extracted to local files with [WorkspaceDb.getFile]($backend)
 
-> Note: files may be compressed as they are stored in `WorkspaceContainer`s.
+> Note: files may be compressed as they are stored in `WorkspaceContainer`s. In general, files should only be used for backwards compatibility with code that reads files from disk. They should be avoided if possible.
 
 ### WorkspaceResource.Names
 
 [WorkspaceResource.Name]($backend)s identify resources within a `WorkspaceDb`. There are no restrictions on the format of a `WorkspaceResource.Name`, other than they are limited to 1024 characters and may not start or end with a blank character.
 
 > Note: `WorkspaceResource.Name`s must be unique for each resource type. But, it is possible to have a `string`, a `blob`, and a `file` resource in the same `WorkspaceDb` with the same `WorkspaceResource.Name`.
```
pmconne, line 266: > it is possible to have a `string`, a `blob`, and a `file` resource in the same `WorkspaceDb` with the same `WorkspaceResource.Name`

Do you have use cases for this? It seems confusing and error-prone.

kabentley: No, I don't really have a use case. But they are stored in 3 different tables, so there's no way to enforce uniqueness across them without a schema change adding some other "name" table". I don't really think its a big deal, people have to know what type their resource is.

core/backend/src/assets/Settings/Schemas/Gcs.schema.json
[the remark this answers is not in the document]
pmconne: @williamkbentley please resolve if satisfied, or follow up if unsatisfied.

core/backend/src/workspace/Workspace.ts
pmconne: @williamkbentley

core/backend/src/IModelDb.ts
```diff
@@ -1405,7 +1407,7 @@ export abstract class IModelDb extends IModel {
    * @note All saved `SettingDictionary`s are loaded into [[workspace.settings]] every time an iModel is opened.
    * @beta
    */
-  public saveSettingDictionary(name: string, dict: SettingDictionary) {
+  public saveSettingDictionary(name: string, dict: SettingObject) {
```
pmconne, line 1410: Shouldn't `saveSettingDictionary` try to obtain the schema (file) lock?

pmconne:
In a subsequent PR we need to address discrepancies with font dbs, including:
- They don't follow the persistence and retrieval mechanisms described by the WorkspaceDb interface.
- No one ever calls `iModelJsNative.addFontWorkspace`, so native code cannot find workspace fonts.
- `FontManager` maintains a static list of font workspaces, which poses at least two problems
  - Race conditions - worker threads (e.g., for tile generation) will want stable access to the list of font dbs while the main thread is trying to change the list's contents. Currently zero synchronization.
  - The static list does not account for font dbs specified in one particular iModel's workspace.
- Adding geometry that uses a font requires updating the font name<->Id mapping in the iModel, which requires an exclusive lock on the entire iModel.

The issues with the static list are also relevant to GCS workspace dbs.

pmconne:
@kabentley is it assumed that each WorkspaceDb holds exactly one type of resource? If I wanted to write the logic to, say, look up all the text styles available to me, it'd be something like:

- Get the value of the setting that tells me the list of workspace dbs containing text styles.
- Walk through the list of workspace dbs.
- Assume every row in the `strings` table is valid stringified JSON representing a text style.
- Decode the JSON into a TextStyle.

If somebody messes up and puts a dimension style workspace db into my text style db list, or adds a dimension style into one of my text style dbs, difficult-to-diagnose errors will occur at run-time. There is currently no mechanism to prohibit either of those occurrences. I can of course validate the JSON and ignore things that aren't text styles, but that just hides the configuration problem from the user.

If someone is inspecting a .itwin-workspace file to determine what kind of resources it contains, they'll have to hope someone filled in the optional description in the manifest sufficiently, or that it's obvious from the names or contents of the data present.

If we want to permit dimension styles and text styles and whatever else to be stored in the same workspace db, we need some mechanism for distinguishing between types of resources and preventing name collisions, whether a naming scheme, a type column, etc. To iterate through text styles, I should not have to look at each string in the strings table and determine that 1) it is JSON and 2) the JSON represents a text style.

(Of course, a single workspace db *is* able to contain up to 3 types of resources without these problems, if each has a different persistence (file, blob, string) - 4 if you count fonts, which are stored differently. This seems awkward to me though).

If we want to enforce that a workspace db contains only resources of a single type, we again need some notion of "resource type" beyond "file, string, or blob (or font)".

People will presumably want tools to view the contents of a workspace db, e.g., to see what text styles it contains and the properties of the styles. They will also want tools to add new resources to workspaces. I don't want us (and every app developer who invents their own kinds of resources) to have to write a separate UI for viewing/editing each type of resource. It might be useful for each resource in the workspace db to specify its type, and for a schema to be provided for each resource type, so both people and programs can understand the contents of the workspace db.

core/backend/src/IModelDb.ts
kabentley: Yes, i suppose that's true. Changing the setting dictionary should definitely be an administrator task.

core/backend/src/workspace/SettingsSchemas.ts
```diff
@@ -51,35 +55,125 @@ export interface SettingSchemaGroup {
 export class SettingsSchemas {
   private constructor() { } // singleton
```
pmconne, line 56: Singleton seems wrong. SettingsSchemas is a big ball of mutable state that persists after `IModelHost.shutdown`, seems like IModelHost ought to own it.

core/backend/src/workspace/SettingsSchemas.ts
```diff
@@ -21,28 +21,32 @@ import { IModelJsFs } from "../IModelJsFs";
...
+  readonly items?: SettingSchema;
+  /** type is required for settings */
+  readonly type: JSONSchemaTypeName;
+  /** name of typeDef for objects or arrays to inherit */
+  readonly extends?: string;
+  /** for objects, a list of named properties and their definitions */
   readonly properties?: { [name: string]: SettingSchema };
-
   /** whether the setting replaces lower priority entries with the same name or combines with them. */
   readonly cumulative?: true;
```
pmconne, line 33: Looks like this is the property that's supposed to control whether querying an array property returns the highest-priority match or a concatenation of all the unique values in all matching settings. Presumably it is meaningless for non-arrays (including objects, though you could make a case for a higher-priority setting overriding some properties of a lower-priority one).

This is not used anywhere currently.

Uniqueness is straightforward for things that aren't objects or arrays (assuming strings are compared case-sensitively).

kabentley: I added this for Settings editors, so they can indicate whether the values supplied are meant to override or combine with other values. It is only useful for arrays, but there's no way to have a schema member that only applies to arrays.

core/backend/src/workspace/SettingsSchemas.ts
kabentley: I agree.

core/backend/src/workspace/SettingsSchemas.ts
pmconne: Expectation is that `getArray` will look up the setting def, check this flag, and combine or override as specified, no? That's the missing piece.

kabentley: That sounds like a good idea. I guess the default should be `false` if there's no schema?

core/backend/src/workspace/Settings.ts
```diff
@@ -86,237 +97,91 @@ export interface Settings {
...
-  * @param defaultValue value returned if settingName is not present in any SettingDictionary, or if the highest priority setting is not an object.
+  * @param defaultValue value returned if settingName is not present in any Settings.Dictionary, or if the highest priority setting is not an object.
   */
   getObject<T extends object>(settingName: SettingName, defaultValue: T): T;
   getObject<T extends object>(settingName: SettingName): T | undefined;
 
   /** Get an array setting by SettingName.
   * @param settingName The name of the setting
-  * @param defaultValue value returned if settingName is not present in any SettingDictionary, or if the highest priority setting is not an array.
+  * @param defaultValue value returned if settingName is not present in any Settings.Dictionary, or if the highest priority setting is not an array.
```
kabentley, line 183: I just noticed that these comments are now wrong. It throws if the value doesn't match.

core/backend/src/internal/workspace/SettingsImpl.ts
```diff
@@ -0,0 +1,161 @@
...
+const dictionaryMatches = (d1: Settings.Dictionary.Source, d2: Settings.Dictionary.Source): boolean => {
+  return (d1.workspaceDb === d2.workspaceDb) && (d1.name === d2.name);
+};
+
+class SettingsDictionaryImpl implements Settings.Dictionary {
+  public readonly props: Settings.Dictionary.Props;
+  public readonly settings: SettingObject;
+  
+  public constructor(props: Settings.Dictionary.Props, settings: SettingObject) {
+    this.props = { ...props }; // make a copy so it can't be changed by caller
```
kabentley, line 27: I think this copy is no longer necessary since all the members are `readonly`.

core/backend/src/CloudSqlite.ts
kabentley: @pmconne feel free to resolve these comments whenever you like

core/backend/src/internal/workspace/SettingsImpl.ts
pmconne: If you're concerned about caller modifying the input, `readonly` won't save you.
```ts
const props = { name: "a", priority: 1 };
settings.addDictionary(props, {});
props.name = "b";
```

kabentley:
I meant to reply to this comment: 

> Presumably it is meaningless for non-arrays (including objects, though you could make a case for a higher-priority setting overriding some properties of a lower-priority one).

VS Code does that for objects, but I strongly suggest we don't allow it. It creates a big complicated, ambiguous mess for queries. I don't see it being worthwhile or necessary.

pmconne:
> VS Code does that for objects, but I strongly suggest we don't allow it. It creates a big complicated, ambiguous mess for queries. I don't see it being worthwhile or necessary.

Fine by me. I will rename `cumulative` to `combineArrays`.

pmconne:
Some questions from @Josh-Schifter below. I am starting on rewriting Workspace.md now, with an eye toward making sure it provides answers to all of these (among others).

> What is the admin workflow for managing workspaces?
> - At present, an admin must create the cloud containers, create the styles (and decide versioning), and configure specific workspace versions to iModels.
> - Where should all that happen?
> 
> What is the unit of versioning for workspace information?
> - Should all ‘Drawing Production’ data be in one block, versioned together?
> - Should users (admins) be able to create and manage the blocks and decide on their contents?
> 
> What is the admin workflow for creating new versions?
> - I assume that admins will make changes over multiple sessions before publishing a new version.  Is that correct?
> - How can they test their changes?  Can they point their session to the unpublished version?
> - Should the UI for changing data and publishing versions be part of our studio app?
> 
> How are new versions consumed by iTwins / iModels?
> - Who will decide when to consume a new version?
> - I assume that consuming a new version will imply a changeset is created in the target iModel.  Is that right?
> - Where will that UI be?
> - Specific code will be needed to apply the new version (ex. dimension style).  How will we assure that code is present when applying?
  > - This is related to the earlier question about which data is versioned together.
> 
> How iModels consume data from the application default workspaces?
> - They will also be versioned – by us. But what version will the iModels get?
> - Will there be a way to update the version for a particular iModel?
> - Is this handled the same or differently from user managed workspaces?
> 
> How will admins organize “Org level workspaces”?
> - Will every style in an Org workspace by available for every iModel?
> - What if they want a subset of those styles for a particular project?
> - Will they need to create multiple workspacedbs for each possible subset?
> - Wouldn’t that force them to either a) create a separate db for each style or b) duplicate some styles between dbs?

core/backend/src/assets/Settings/Base.Schema.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$comment": "This schema describes the meta-structure of an iTwin Settings Schema file. It is used to provide a guide for the schema files themselves.",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "additionalProperties": false,
+  "required": [
+    "schemaPrefix",
+    "description"
```
pmconne, line 7: `description` is declared as optional in `SettingSchemaGroup`. Which one is correct?

core/backend/src/workspace/SettingsSchemas.ts
```diff
@@ -21,208 +18,52 @@ import { IModelJsFs } from "../IModelJsFs";
...
+ * and are identified by their (required) `schemaPrefix` member
  * @beta
  */
 export interface SettingSchemaGroup {
-  readonly groupName: string;
-  readonly properties: { [name: string]: SettingSchema };
+  readonly schemaPrefix: string;
+  readonly settingDefs?: { [name: string]: SettingSchema | undefined };
+  readonly typeDefs?: { [name: string]: SettingSchema | undefined };
   readonly order?: number;
```
pmconne, line 46: `order` does not appear in Base.Schema.json, is undocumented here, and appears to be unused anywhere. I infer that it is used for ordering groups in a UI, with lower values being ordered before higher values. Correct?

kabentley: Yes, that's its only purpose.

core/backend/src/assets/Settings/Base.Schema.json
kabentley: I'd say lets make it required.

pmconne: Localization strategy TBD by application?

core/backend/src/assets/Settings/Schemas/Gcs.schema.json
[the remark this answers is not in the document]
pmconne: file deleted

core/backend/src/assets/Settings/Base.Schema.json
```diff
@@ -0,0 +1,27 @@
+{
```
pmconne, line 1: Base.Schema.json ought to be in Settings/Schemas, no? Currently it gets loaded as a settings dictionary instead of as a schema.

pmconne: Moved it.

core/backend/src/assets/Settings/Base.Schema.json
pmconne: Made it required.

core/backend/src/workspace/SettingsSchemas.ts
pmconne: Added to schema and documented.

core/backend/src/workspace/SettingsSchemas.ts
pmconne: Yes. Implemented it.

core/backend/src/assets/Settings/Base.Schema.json
kabentley: Yikes! Maybe we should make a different extension for schema files? The downside is that VSCode, etc. won't treat it as JSON.

docs/learning/backend/Workspace.md
```diff
@@ -54,15 +102,13 @@ For example:
 "vibration-map/filters/prefabricated"
 ```
 
-`SettingName`s must be valid [JavaScript property names](https://developer.mozilla.org/en-US/docs/Glossary/property/JavaScript), but should not contain periods or spaces.
-
-### SettingSchema
+Are SettingNames from the `energyAnalysis`, `iot-scan-visualization`, and `vibration-map` SettingSchemas.
 
-A single `Setting` is defined according to the rules of [JSON Schema](https://json-schema.org/) via a [SettingSchema]($backend). The primary objective of creating a [SettingSchema]($backend) is to advertise the existence, meaning, and "form" of a Setting. Users supply values for setting using a settings editor, and are presented with the information from `SettingsSchema`s to guide their choices. Also, `SettingSchema`s may also supply a default value so users can understand what happens if they don't provide a value for a Setting.
+`SettingName`s must be valid [JavaScript property names](https://developer.mozilla.org/en-US/docs/Glossary/property/JavaScript), but may not contain periods or spaces.
```
pmconne, line 107: When you say setting names must be "valid JS property names" do you really mean JS identifiers (i.e., legal variable names)? Any valid JS string is a valid JS property name. JS identifiers can't start with a number, contain whitespace, contain characters like "-", "+", etc. If that is your intent, what's the motivation for those restrictions?

core/backend/src/workspace/Settings.ts
pmconne: Updated

core/backend/src/workspace/Workspace.ts
```diff
@@ -15,95 +15,133 @@ import type { IModelJsNative } from "@bentley/imodeljs-native";
...
 
-/** The base name of a WorkspaceDb within a WorkspaceContainer (without any version identifier) */
+/** The base name of a [[WorkspaceDb]], without any version information.
+ * @see [[WorkspaceDbFullName]] for the fully-specified name, including version information.
+ * @beta
+ */
 export type WorkspaceDbName = string;
 
-/** The  name of a WorkspaceDb within a WorkspaceContainer, including the version identifier */
+/** The fully-specified name of a [[WorkspaceDb]], combing its [[WorkspaceDbName]] and [[WorkspaceDbVersion]] in the format "name:version".
```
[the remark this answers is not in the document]
pmconne: fixed.

docs/learning/backend/Workspace.md
kabentley: I think setting names should be "simple".

pmconne: I don't see any code enforcing these restrictions, or a clear definition of what the restrictions actually are.

core/backend/src/internal/workspace/WorkspaceImpl.ts
```diff
@@ -0,0 +1,794 @@
...
+export function constructWorkspaceEditor(): WorkspaceEditor {
+  return new EditorImpl();
+}
+
+export function noLeadingOrTrailingSpaces(name: string, msg: string) {
+  if (name.trim() !== name)
+    throw new Error(`${msg} [${name}] may not have leading or trailing spaces`);
+}
+
+export function validateWorkspaceDbName(dbName: WorkspaceDbName) {
```
pmconne, line 729: Can I summarize this as "must be a valid Windows and Unix filename no longer than 255 characters"? It's not currently documented anywhere.

pmconne: I added documention on `WorkspaceDbName`.

kabentley: It came from [here](https://www.npmjs.com/package/filename-reserved-regex)

example-code/snippets/src/backend/WorkspaceExamples.test.ts
```diff
@@ -0,0 +1,303 @@
...
+        light: "full" | "shade" | "partial";
+      }
+
+      function addTree(treeDb: EditableWorkspaceDb, species: string, tree: TreeResource): void {
+        treeDb.addString(species, JSON.stringify(tree));
+      }
+
+      const cornusDb = container.getEditableDb({ dbName: "cornus" });
+
+      addTree(cornusDb, "alternifolia", {
```
pmconne, line 285: @kabentley what exactly can you do with the default workspace in a newly-created container? I get `Error: cornus:1.0.0 has been published and is not editable. Make a new version first.`. I understand why that happens, but I don't understand why we create an empty default workspace that can never have content added to it unless we create a new version.

I suppose if we made pre-release dbs editable that would solve the problem. Am I overlooking something here?

pmconne: See my next commit, where I switch to creating a new db and trying to populate it, receiving the same error. It seems the only way to edit a db is to make a new version of an existing one.

pmconne: Switched to creating and populating a new version of the empty db for now.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
```diff
@@ -0,0 +1,363 @@
...
+      cornusDb = container.getEditableDb(cornusPatchProps);
+      cornusDb.open();
+      addTree(cornusDb, "racemosa", {
+        commonName: "Northern Swamp Dogwood",
+        hardiness: { minimum: 4, maximum: 9 },
+        light: "full",
+      });
+      cornusDb.close();
+
+      await container.createDb({
```
pmconne, line 327: @kabentley `WorkspaceEditor.createDb` appears to be unusable. It will upload a new, empty db, and then throw an error trying to retrieve it for editing.

```
Error: abies:1.0.0 has been published and is not editable. Make a new version first.
      at EditorContainerImpl.getEditableDb (/home/paul/c/j2/s/itwinjs/core/backend/src/internal/workspace/WorkspaceImpl.ts:553:13)
      at EditorContainerImpl.getWorkspaceDb (/home/paul/c/j2/s/itwinjs/core/backend/src/internal/workspace/WorkspaceImpl.ts:548:17)
      at EditorContainerImpl.createDb (/home/paul/c/j2/s/itwinjs/core/backend/src/internal/workspace/WorkspaceImpl.ts:591:17)
      at Context.<anonymous> (src/backend/WorkspaceExamples.test.ts:327:7)
```

pmconne: Working around it by making 1 db per container (the default db), which is the pattern I want to encourage anyway, but we shuold `createDb`.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
pmconne: #6811

docs/learning/backend/Workspace.md
```diff
@@ -26,20 +26,68 @@ In the list above, later entries tend to change more frequently and, in the case
...
 
 ## Settings
 
-Settings are named parameters defined by applications but supplied at runtime so that their values may vary according to circumstances across and even within sessions. At runtime Settings are just JavaScript primitives and may be accessed via [Settings.getSetting]($backend) by supplying a `SettingName`. Setting lookup is generally very efficient, so settings should *not* be cached in application code and should instead be retrieved as needed. That way they do not get out of sync as they change.
+Settings are named parameters defined by applications but supplied at runtime. Their values may be supplied by administrators and may vary according to circumstances across and even within sessions. At runtime, Settings are just JavaScript primitives and may be accessed via [Settings.getSetting]($backend), [Settings.resolveSetting]($backend) (and the related type-specific functions), by supplying a `SettingName`. Setting lookup is generally very efficient, so settings should *not* be cached in application code and should instead be retrieved as needed. That way they do not get out of sync as they change.
```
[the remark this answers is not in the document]
pmconne: @williamkbentley completely rewritten, please resolve or revisit.

pmconne:
@ben-polinsky @williamkbentley docs build errors are so tedious to reverse-engineer...can you tell what source code file+line this is referring to?
```
Invalid Link: "[manifest]($core-backend.WorkspaceDb)" in file reference/core-backend/workspace/editableworkspacedb/index.html"
```

pmconne:
> The [manifest] one does not make sense to me. 

@kabentley had `get manifest(): WorkspaceDbManifest` on the `WorkspaceDb` interface. I switched it to `readonly manifest: WorkspaceDbManifest`, maybe the docs generator likes that better.

williamkbentley:
> EDIT: Yeah, it's happy now.

I cringe when docs generation dictates code design. @ben-polinsky , maybe something to look into

example-code/snippets/src/backend/WorkspaceExamples.test.ts
kabentley: As we discussed, `getEditableDb` should take an optional parameter that skips the "copied" check for amending previously published WorkspaceDbs (the opposite of the NPM paradigm.) Then, `createDb` would pass `true` for that.

docs/learning/backend/Workspace.md
```diff
@@ -1,475 +1,222 @@
...
+"Resources" are bits of data that an application depends on at run-time to perform its functions. The kinds of resources can vary widely from one application to another, but some common examples include:
+- [TextStyle]($common)s and fonts used when placing [TextAnnotation]($common)s.
+- [GeographicCRS]($common)es used to specify an iModel's spatial coordinate system.
+- Images that can be used as pattern maps for [Texture]($backend)s.
+- [SettingsDictionary]($backend)s defining reusable settings.
 
-Conceptually, you can picture the hierarchy like this:
+It might be technically possible to store resources in [Setting]($backend)s, but doing so would present significant disadvantages:
+- Some resources, like images and fonts, may be defined in a binary format that is inefficient to represent using JSON.
+- Some resources, like geographic coordinate system definitions, must be extracted to files on the local file system before they can be used.
```
kabentley, line 142: CRS's (and fonts) are not extracted to files. They are read directly from blobs.

pmconne: I thought the whole thing with files was that CSMap needs to do fopen. If that's not the case, do we have any real-world examples of file resources?

example-code/snippets/src/backend/WorkspaceExamples.test.ts
```diff
@@ -0,0 +1,424 @@
...
+
+      // __PUBLISH_EXTRACT_START__ WorkspaceExamples.AddTrees
+      interface TreeResource {
+        commonName: string;
+        hardiness: HardinessRange;
+        light: "full" | "shade" | "partial";
+      }
+
+      function addTree(treeDb: EditableWorkspaceDb, species: string, tree: TreeResource): void {
+        // We use a prefix to distinguish trees from other kinds of resources that might be present in the same WorkspaceDb.
```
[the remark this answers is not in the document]
pmconne: Elaborate please. Setting names, yes. Resources names? We talked about how there *isn't* any such convention and how it would be useful, but nothing enforces it.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
```diff
@@ -0,0 +1,424 @@
...
+      });
+
+      addTree(cornusDb, "asperifolia", {
+        commonName: "Roughleaf Dogwood",
+        hardiness: { minimum: 9, maximum: 9 },
+        light: "full",
+      });
+
+      cornusDb.close();
+      cornusDb.container.releaseWriteLock();
```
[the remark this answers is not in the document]
pmconne: Added.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
```diff
@@ -0,0 +1,424 @@
...
+      // __PUBLISH_EXTRACT_START__ WorkspaceExamples.CreateWorkspaceDb
+      const editor = WorkspaceEditor.construct();
+
+      async function createTreeDb(genus: string): Promise<EditableWorkspaceDb> {
+        const label = `Trees ${genus}`;
+        const description = `Trees of the genus ${genus}`;
+        const container: EditableWorkspaceContainer = await editor.createNewCloudContainer({
+          // A description of the new WorkspaceContainer.
+          metadata: {
+            label,
```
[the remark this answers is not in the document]
pmconne: Adjusted.

Josh-Schifter:
* I think workspace.md could be improved by expanding a bit on the opening paragraph.  For example, a small block diagram showing how Resources are stored inside containers and referred to by Settings.  As it is now, it is easy for the reader (of the first paragraph) to gloss over the distinction between settings and resources.

* As far as I can see, workspace.md does not attempt to address any of the [questions I raised previously](https://github.com/iTwin/itwinjs-core/pull/6646#issuecomment-2127684175).  Has there been an attempt to answer those somewhere else?

docs/learning/backend/Workspace.md
kabentley: No, I added virtual methods to CSMap to read data from SQLite. There are no uses of files in core. I expect apps will need them though for backwards compatibility, unfortunately. Maybe we don't publish them until someone asks?

example-code/snippets/src/backend/WorkspaceExamples.test.ts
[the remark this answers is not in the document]
kabentley: You're right I was mixing resource names and settings names about the "required" part. But, the prefix shouldn't be optional, otherwise users won't be able to store a mix of resources in the same WorkspaceDb. 

I don't know how/where to enforce it though, since I didn't prefix the CRS resource names (they are the CSMap file names). Those WorkspaceDbs are special though.

pmconne:
> I think workspace.md could be improved by expanding a bit on the opening paragraph. For example, a small block diagram showing how Resources are stored inside containers and referred to by Settings. As it is now, it is easy for the reader (of the first paragraph) to gloss over the distinction between settings and resources.

The article is designed to gradually introduce concepts rather than overwhelm by front-loading everything. But a diagram in the intro that shows the organization of a worksapce without delving immediately into the specifics is a good idea.

> As far as I can see, workspace.md does not attempt to address any of the https://github.com/iTwin/itwinjs-core/pull/6646#issuecomment-2127684175.

I find this a little baffling. I attempted to specifically address these questions. The exceptions are those about workflows and UI for admins to manage workspaces, because those will be in a Mango app which is not yet available. I refer to it obliquely here:

> Note: Creating and managing data in workspaces is a task for administrators, not end-users. Administrators will typically use a specialized application with a user interface designed for this task. For the purposes of illustration, the following examples will use the `WorkspaceEditor` API directly.

The other exception is this:

> What is the admin workflow for creating new versions?
> - I assume that admins will make changes over multiple sessions before publishing a new version.  Is that correct?
> - How can they test their changes?  Can they point their session to the unpublished version?

We discussed this via email. Currently it's not possible to edit a new version over multiple sessions. I proposed a solution for a follow-up PR. I can't document it here because it doesn't exist yet.

Unless you literally mean you didn't find answers to **any** of your questions, please be more specific about which questions remain unanswered.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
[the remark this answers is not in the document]
pmconne: Same for fonts - they are not prefixed, though they are namespaced in the be_Props table (and no other resources are stored in that table anyway).

docs/learning/backend/Workspace.md
pmconne: Adjusted.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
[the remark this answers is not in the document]
pmconne: I can rewrite to be more emphatic that you should definitely use a prefix unless you are creating a WorkspaceDb that will only contain resources of a specific type.

docs/learning/backend/Workspace.md
```diff
@@ -1,475 +1,222 @@
...
-            "description": "if true, pre-fetch all of the data from the cloud in the background"
-          }
-        }
-      }
-    },
+[[include:WorkspaceExamples.CreateWorkspaceDb]]
 ```
 
-For example:
+Since every [WorkspaceDb]($backend) must reside inside a [WorkspaceContainer]($backend), we must first create a container. Creating a container also creates a default `WorkspaceDb`. In the example above, we set up the default `WorkspaceDb` to be our as-yet empty "cornus" tree database.
```
johnnyd710, line 171: nit: `Since every [WorkspaceDb]($backend) must reside inside a [WorkspaceContainer]($backend), we must first create a container.` should probably appear before the example.

docs/learning/backend/Workspace.md
```diff
@@ -1,475 +1,222 @@
...
 
-```ts
-[[include:Settings.containerAlias]]
-```
+Now, let's define what a "tree" resource looks like, and add some to the new `WorkspaceDb`. To do so, we'll need to make a new version of the empty "cornus" `WorkspaceDb` we created above. `WorkspaceDb`s use [semantic versioning](https://semver.org/), and each version of a given `WorkspaceDb` becomes immutable once it is published to cloud storage. So, the process for creating a new version of a `WorkspaceDb` is as follows:
```
johnnyd710, line 173: nit: `To do so, we'll need to make a new version of the empty "cornus" `WorkspaceDb` we created above` this is a tad confusing to me because we didn't create a new version of the db, we only defined a function to do so.

docs/learning/backend/Workspace.md
```diff
@@ -1,475 +1,222 @@
...
-4. current iModel
-
-In the list above, later entries tend to change more frequently and, in the case of duplicate values, later entries override earlier ones.
-
-[Workspace]($backend)s expresses the current state of the session in two forms:
+## Settings
 
-  1. [Settings](#settings)
-  2. [WorkspaceDb](#workspaceDbs)
+[Settings]($backend) are how administrators of an application or project configure the workspace for end-users. Be careful to avoid confusing them with "user preferences", which can be configured by individual users. For example, an application might provide a check box to toggle "dark mode" on or off. Each individual user can make their own choice as to whether they want to use this mode - it is a user preference, not a setting. But an administrator may define a setting that controls whether users can see that check box in the first place.
```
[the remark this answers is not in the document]
pmconne: Only an administrator persona should be able to access a settings editor. Maybe under the hood it uses the same basic editor UI as a preferences editor - I don't think that matters.

docs/learning/backend/Workspace.md
```diff
@@ -1,475 +1,222 @@
...
 
 ```ts
-[[include:Settings.addDictionary]]
+[[include:WorkspaceExamples.GetSettings]]
 ```
 
-Values in `SettingDictionary`s with a higher `SettingsPriority` override values in dictionaries with a lower priority.
+Note that `getString` returns `undefined` for "landscapePro/preferredStyle" because our dictionary didn't provide a value for it. The overload of that function (and similar functions like [Settings.getBoolean]($backend) and [Settings.getObject]($backend)) allows you to specify a default value to use if the value is not defined.
+
+> Note: In general, avoid caching the values of individual settings - just query them each time you need them, because they can change at any time. If you must cache (for example, if you are populating a user interface from the setting values), listen for and react to the [Settings.onSettingsChanged]($backend) event.
```
williamkbentley, line 80: I understand not caching settings. What about caching Workspace resources? 

In the example, would you expect the `getAvailableTrees` function to be called every time someone ran/installed the `place-dogwood` or `place-shrub` tool?

pmconne: Yes, those resources are accessed via settings that point to the relevant workspace dbs.

docs/learning/backend/Workspace.md
pmconne: Moved it above.

docs/learning/backend/Workspace.md
pmconne: Reorganized a bit.

johnnyd710 (approved):
Workspace.md and the example code lgtm. Much easier to understand.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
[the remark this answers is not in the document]
kabentley: Yes. It should be obvious from your example that uses a prefix for querying. The expectation is that administrators will organize WorkspaceDbs according to project type, clients, etc. rather than by application (to answer @Josh-Schifter 's question about WorkspaceDb granularity/organization.) The prefix is necessary to avoid collision with other apps.

kabentley:
> Currently it's not possible to edit a new version over multiple sessions.

I don't understand why not. It will be up to the editor app, but a "session" may end before the write lock is released.

pmconne:
> I don't understand why not. It will be up to the editor app, but a "session" may end before the write lock is released.

Are you suggesting the admin user request the write lock, overriding the default expiration time of 1 hour with something like 1 week, and make sure they remember to release the write lock before the end of the week?

That is awkward. It doesn't allow the user to switch to a different device, or for them to release the lock so a different user can contribute their own changes to the same version. It doesn't permit sharing the WIP new version with other (potentially non-admin) users to test.

We talked about the idea that pre-release versions should be editable even after the changes are "published", to address these problems.

pmconne:
> a small block diagram showing how Resources are stored inside containers and referred to by Settings

@Josh-Schifter I am bad at diagramming. Does this come anywhere close to capturing what you want to see?

image

docs/learning/backend/Workspace.md
pmconne: Added a paragraph to the end after the `getAvailableTrees` example.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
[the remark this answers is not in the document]
pmconne: Adjusted.

Josh-Schifter:
> I find this a little baffling.

Ok.  My statement was overly broad.  I'll try to narrow down.

If the answer to my questions on workflows is 'we don't know yet'... that is valid.  It doesn't change the fact that those questions were never addressed.

> The exceptions are those about workflows and UI for admins to manage workspaces, because those will be in a Mango app which is not yet available.

Is the intention that Bentley will develop a single app which admins will use to manage workspaces?

The answer regarding 'what is the unit of versioning', seems to be 'WorkspaceDb'.  But a WorkspaceDb "can contain any number of resources of any kind"... which I guess means that the implementation doesn't care which resources are versioned together.  That's nice but it doesn't answer the original question.  Who will decide what goes into which workspaceDb?  Will it be platform/bis, or app developers, or cad admins?

Likewise, I asked about how new versions of resources will be consumed by iModels.  I get that there will be a setting which will make the new version available.  I guess that setting will be modified by the admin using the as-yet-defined Mango app?  That doesn't explain how text style changes will get propagated to text elements.

All my questions were "answered / not answered" in this same style.  Sure, some of the low lying mechanics are more clear now, but the workflows are not.  Who is responsible for defining the workflows?

Josh-Schifter:
> > a small block diagram showing how Resources are stored inside containers and referred to by Settings
> 
> @Josh-Schifter I am bad at diagramming. Does this come anywhere close to capturing what you want to see?
> 
> image

Since the diagram is meant to go with the first paragraph, I'd like it to include all the main terms from there, and only the ones from there.  As of now, the main terms from the first paragraph are Workspace, Settings, Containers, and Resources.  But the diagram is missing 'Resources' and includes the additional terms 'Settings Dictionary' and 'WorkspaceDb'.

Josh-Schifter:
> I don't want the diagram to lie by omitting dictionaries and workspace dbs. Perhaps the intro paragraph should briefly mention them.

Fine with me.

pmconne:
> All my questions were "answered / not answered" in this same style. Sure, some of the low lying mechanics are more clear now, but the workflows are not. Who is responsible for defining the workflows?

I see the disconnect - thanks. This PR, including the documentation, is about an API that workflows can be built on top of. The API is meant to enable various user workflows, not to impose a specific one. Applications are going to want to configure their default workspaces. Administrators will want to customize those and define their standards. Both will organize their resources in the ways that makes sense for them. Platform has a limited role: providing the APIs and a couple of broadly useful, publicly accessible workspace containers for things like GCSes and fonts.

Certainly a bunch of programmers are not going to define the end user workflows.

pmconne:
@Josh-Schifter does this work better?

image

Josh-Schifter:
> @Josh-Schifter does this work better? Updated intro paragraph as discussed.

I like it.  Thanks.

docs/learning/backend/Workspace.md
```diff
@@ -1,475 +1,226 @@
...
 
-`Settings` are *named parameters* that an application defines and whose values are supplied at runtime. `WorkspaceDb`s hold *named resources* (i.e. data) that the application uses. `Settings` and `WorkspaceDb`s are often related in application logic, e.g.:
+A [Setting]($backend) is simply a name-value pair. The value can be of one of the following types:
+- A `string`, `number`, or `boolean`;
+- An `object` containing properties of any of these types; or
+- An `array` containing elements of one of these types.
 
-- a Setting may contain the "formula" to find a resource
-- a `WorkspaceDb`s may hold a resource that defines a group of `Settings`
+A [SettingName]($backend) must be unique, 1 to 1024 characters long with no leading nor trailing whitespace, and should begin with the schema prefix of the [schema](#settings-schemas) that defines the setting. (More on schemas shortly). For example, LandscapePro might define the following settings:
```
grigasp, line 18: Not sure the part "more on schemas shortly" is even needed, since "schema" is a link. But if you want to keep that, I suggest changing it to this:

```suggestion
A [SettingName]($backend) must be unique, 1 to 1024 characters long with no leading nor trailing whitespace, and should begin with the schema prefix of the [schema](#settings-schemas) that defines the setting - more on schemas shortly. For example, LandscapePro might define the following settings:
```

pmconne: removed.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
pmconne: Done.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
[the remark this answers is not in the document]
pmconne: Resolve if satisfied.

pmconne (approved):

core/backend/src/workspace/Workspace.ts
williamkbentley: Similar question - I think it might be helpful to know what priority a resource has. 

So in the LandscapePro example, if there was a default trees db, and an admin configured DogwoodDb, it might be helpful to show the user a list of all trees grouped by source. I know the list of resources comes sorted by priority, but I dont see how the application can tell which priority the resource has.

pmconne: Resources don't directly have priority, the dictionaries that point to the WorkspaceDbs containing them do. You can use `WorkspaceDb.getWorkspaceDbs` to resolve a setting name to a list of dbs. You can pass a DbListFilter to it, which will supply to you for each db the cloud props and the dictionary from which it was resolved. The dictionary has a priority.

pmconne:
> How do we know this system is suitable for purpose when we don't even know what the purpose is?
>
> Repeating what I wrote before: Who is responsible for defining the workflows?

Come on now. We know what the purpose is. The details of how the functionality is exposed to end users is a question for PM and UX folks from application teams. Could we have started with the workflows and designed the API based on them? Maybe, but we didn't. We didn't define the workflows of how an end user places text into a drawing before we implemented the TextAnnotation APIs either.

Discussion with PM and UX regarding end user workflows isn't going to occur here on github.

PR #6647, opened 2024-04-18
Don't save to lock-database when inserting elements if their Model is locked

aruniverse:
Changes as requested by @abeesh after seeing a 20x slow down with local synchronizations compared to batch connectors.

### DIFF
--- a/common/changes/@itwin/core-backend/lock-perf_2024-04-18-17-47.json
+++ b/common/changes/@itwin/core-backend/lock-perf_2024-04-18-17-47.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Don't lock child if parent is already locked",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/changes/@itwin/core-backend/lock-perf_2024-04-18-17-47.json
+++ b/common/changes/@itwin/core-backend/lock-perf_2024-04-18-17-47.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/core-backend",
-      "comment": "Don't lock child if parent is already locked",
+      "comment": "Don't save to lock-database when inserting elements if their Model is locked",
       "type": "none"
     }
   ],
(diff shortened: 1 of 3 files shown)


kabentley:
What problem does this solve? Storing the fact that we hold the lock on a newly created element seems less expensive (and is innocuous) than the query to find its parents.

aruniverse:
> What problem does this solve? Storing the fact that we hold the lock on a newly created element seems less expensive (and is innocuous) than the query to find its parents.
> 
> Edit: I now see we cache the lock on models, so it doesn't look like we'll need to look too far up the tree to find that an element's model is locked. I'm not sure which way will be more efficient. At any rate, this code doesn't "lock an element", it saves the fact that an element is already locked. What prompted you to change this?

Apologies, I just updated the description. Abeesh and team noticed a massive slow down with local synchronizations compared to batch connectors due to the existing behavior. Taking over 5 hrs for local sync vs 20 mins for a batch connector run.

With the included changes, they saw times reduce from 60 mins to under 3 mins.

kabentley:
It seems to me that this will be a better change (assuming you're trying to optimize insert performance);

change [this](https://github.com/iTwin/itwinjs-core/blob/b1776738667635329d78afe3d986069a904af2ee/core/backend/src/Element.ts#L168-L170) to:
```ts
  protected static onInserted(arg: OnElementIdArg): void {
    const locks = arg.iModel.locks;
    if (locks && !locks.holdsExclusiveLock(arg.model))
        locks.elementWasCreated(arg.id);
  }
   ```

abeesh:
Keith is the expert here and the above-mentioned change should work for local sync.

kabentley:
> With the included changes, they saw times reduce from 60 mins to under 3 mins.

I now see that we call `SaveChanges` on every call to `elementWasCreated`. I'm sure that's what is killing the performance. Probably should add some way to batch this up if we're doing a lot of inserting without the model lock held. We'd need some kind of begin/end batch technique. That can wait for some other time ; for LocalSync, I presume this is the case where they create a new Model so it will be locked. My suggestion above will eliminate almost all the lock overhead. There still is a query to the _lock_ database to determine whether the model is locked, but that's way faster than querying the iModel for an element's model.

aruniverse:
Thanks for the recommendation Keith, I've made the suggested changes. I'll add to the backlog an item to investigate and implement batching.

pmconne (approved):

PR #6653, opened 2024-04-19
Created a GitHub action to add geometry reviewers

saeeedtorabi:
I have created a yaml file such that if a PR with label "geometry" is created, Dave and I are automatically added to the PR reveiewers. 

Dan Rodrigouez test it via https://github.com/iTwin/itwinjs-core/pull/6654


dassaf4 (approved):
Nice! Please wait for a second reviewer before merging, as I'm new to this.

DanRod1999 (approved):

DanRod1999:
based on the test this seems to work, but it might be worth seeing if someone puts up a pr and then adds the label after its created. Don't know how github actions behaves on that, however, it appears to still work as intended otherwise

calebmshafer:
@DanRod1999 should be based off the event of adding a label so it will be added.

@saeeedtorabi you both should be added when you're a code owner. Is this to catch when it's potentially not in the geom package but pertains to geom?

dassaf4:
Right, this is generally not a problem with itwinjs, as we already get notified. The faulty change in question was to geomlibs in imodel02 a couple years ago, and for whatever reason, geometry team was unaware of it. So we're exploring ways to prevent this in ADO and github.

DanRod1999:
> @calebmshafer Yes, recently we found an old issue with geometry tag. However, none of the geometry team members were not added to that review at the time and their fix still had a bug which we later fixed. My change is to prevent those kind of situations.

you'll probably want to set the trigger for your workflow on "issues" as well if that is an option.

saeeedtorabi:
> Right, this is generally not a problem with itwinjs, as we already get notified. The faulty change in question was to geomlibs in imodel02 a couple years ago, and for whatever reason, geometry team was unaware of it. So we're exploring ways to prevent this in ADO and github.

Yes, once this PR is merged, will create similar one for our other GitHub repo(s).

.github/workflows/add-geometry-reviewers.yaml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  build:
+    if: ${{ github.event.label.name == 'geometry' }}
+    runs-on: ubuntu-latest
+    steps:
+    - name: Add Reviewers
+      uses: madrapps/add-reviewers@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        reviewers: dassaf4,saeeedtorabi
```
DanRod1999, line 20: you might want to check the api, don't know if there's a different step for adding assignee vs adding a review, but you could use a condition like if `github.event_name == 'pull_request'` add reviewer, `if github.event_name == 'issue' add assignee.

aruniverse (approved):

PR #6655, opened 2024-04-22
Allow pushing changes from a BriefcaseDb opened readonly

kabentley:
Use `BriefcaseDb.executeWriteable` and make that function open the `ServerBasedLocks` lock control.

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -514,6 +514,8 @@ export class BriefcaseDb extends IModelDb {
     static findByKey(key: string): BriefcaseDb;
     get isBriefcase(): boolean;
     get iTwinId(): GuidString;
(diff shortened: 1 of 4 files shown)


ImVeryLost (approved):
Looks perfect for our needs, thank you

pmconne (approved):

kabentley:
@ImVeryLost Are there any other internal apis in use by iTS that you know of? I don't think that should be necessary/allowed.

ImVeryLost:
> @ImVeryLost Are there any other internal apis in use by iTS that you know of? I don't think that should be necessary/allowed.

@kabentley there is an aggregate issue for such usages - https://github.com/iTwin/itwinjs-backlog/issues/1062
Most violations come from nested dependencies

PR #6656, opened 2024-04-22
Property editing refactor

christophermlawson:
Add schemaitemtype checking to property editing, and moved navigationproperties to the appropriate ECClasses child classes.


core/ecschema-editing/src/Editing/Entities.ts
```diff
@@ -13,16 +13,22 @@ import {
...
 import { ECClasses } from "./ECClasses";
 import { MutableEntityClass } from "./Mutable/MutableEntityClass";
+import { NavigationProperties } from "./Properties";
 
 /**
  * @alpha
  * A class extending ECClasses allowing you to create schema items of type EntityClass.
  */
 export class Entities extends ECClasses {
   public constructor(_schemaEditor: SchemaContextEditor) {
```
StefanApfel-Bentley, line 23: This comment applies to all constructors. The schemaEditor argument starts with an underscore, which is actually the way to mark unused methods arguments, as we use it though (passing to the parent constructor), we should not have a leading underscore here.

core/ecschema-editing/src/Editing/ECClasses.ts
```diff
@@ -18,40 +18,38 @@ import { assert } from "@itwin/core-bentley";
 import { PropertyEditResults, SchemaContextEditor, SchemaItemEditResults } from "./Editor";
 import { MutableClass } from "./Mutable/MutableClass";
 import * as Rules from "../Validation/ECRules";
-import { ArrayProperties, EnumerationProperties, NavigationProperties, PrimitiveProperties, Properties, StructProperties } from "./Properties";
+import { ArrayProperties, EnumerationProperties, PrimitiveProperties, Properties, StructProperties } from "./Properties";
+
+export type ECClassSchemaItems = Extract<SchemaItemType, SchemaItemType.EntityClass | SchemaItemType.StructClass | SchemaItemType.RelationshipClass | SchemaItemType.Mixin | SchemaItemType.CustomAttributeClass>;
```
StefanApfel-Bentley, line 23: Since we do not require dynamic extracting of types here, I'd propose:
`export type ECClassSchemaItems = SchemaItemType, SchemaItemType.EntityClass | SchemaItemType.StructClass | SchemaItemType.RelationshipClass | SchemaItemType.Mixin | SchemaItemType.CustomAttributeClass;`

core/ecschema-editing/src/Editing/Properties.ts
```diff
@@ -17,7 +17,7 @@ type MutablePropertyType = MutableProperty | MutableArrayProperty | MutablePrimi
  * A class allowing editing of attributes of the base Property class.
  */
 export class Properties {
-  public constructor(protected _schemaEditor: SchemaContextEditor) { }
+  public constructor(protected ecClassType: SchemaItemType, protected _schemaEditor: SchemaContextEditor) { }
```
StefanApfel-Bentley, line 20: Shouldn't we restrict to `ECClassSchemaItems` instead of SchemaItemType type here?

christophermlawson: Yeah, I created the type and forgot to use it!  It wasn't being used anywhere. Done.

core/ecschema-editing/src/Editing/Entities.ts
christophermlawson: Done

StefanApfel-Bentley (approved):

PR #6676, opened 2024-04-26
Include BIS classes when forwarding txn events to frontend

### LINKED ISSUE #6607: BriefcaseTxns should provide more context with onChanged events.
When reacting to the `onElementsChanged` and `onModelsChanged` events from [BriefcaseTxns](https://www.itwinjs.org/reference/core-frontend/imodelconnection/briefcasetxns/), we are finding that each listener needs to query for the classes of the elements/models that triggered the event as a filter, or is just dumbly reacting to it anyways.  This creates unnecessary traffic for the backend.

The events should also include information about what class of elements/models was changed.  Preferably, in a way that relates the changed element/model id with the class. 

Nothing worth mentioning.

**Additional context**
We see changes come it 2 forms, small focused changes from editing workflows, and large sets of changes from a pull of the iModel.

pmconne:
WIP.
Fixes #6607.

The backend events provide the class Ids, which listeners can use to, e.g., determine if any geometric elements are included in the changes. This information is dropped when forwarding the events to the frontend, which receives only the entity Ids.

Class Ids would be pretty useless on the frontend, unless we give the frontend a way to interrogate class hierarchy/metadata based on class Id. FWIW, they're pretty awkward to use on the backend too, because MetadataRegistry, EntityMetadata, et al are all based on class name, not Id.

Generally I'd prefer not to duplicate information between backend and frontend. The backend is readily available via the IPC pipe, permitting apps to either query based on class Id for the information they need, or implement the logic that needs to respond to events in the backend instead of the frontend. However, this may not be ideal for txn events, for which there may be many listeners, the listeners care about the ordering of the events (so no arbitrary `await` delays inside event callbacks), and all the rest of the logic is implemented in the front-end.

In its current form, the PR provides a frontend API that can asynchronously load metadata for a set of classes given their Ids, consisting of class full name and base classes. You can query loaded metadata by name or Id. Txn events include the `EntityClassMetadata` representing the ECClass of each changed entity, which provides an `is` method to interrogate its inheritance tree.


pmconne:
Simplified in #6690.

PR #6684, opened 2024-05-01
Improved MapFeatureInfo decorator to display properly highlight geometries when terrain is enabled

mdastous-bentley:
The actual MapFeatureInfo tool has been improved to highlight map-layer geometries when the terrain is turned ON.  

To achieve this, the appropriate resolution terrain tiles needs to be collected from the tile tree, then the `@itwin/core-geometry` functionalities are used to drape the map-layer geometries (which include point, lines and polygons) to be draped on the terrain mesh.  The resulting geometries are displayed as decorations.

In the following screenshot, we can see sample highlighted : 
image


pmconne:
> The resulting geometries are displayed as decorations

How many geometries are we talking about?

> In the following screenshot, we can see sample highlighted 

This refers specifically to the single brigher blue polygon? Are the other polygons also drawn as decorations, or something else?

mdastous-bentley:
> > The resulting geometries are displayed as decorations
> 
> How many geometries are we talking about?
> 
> > In the following screenshot, we can see sample highlighted
The tool only support clicked point selection, so the general workflow is a single geometry at a time.  You can always zoom out, but then the terrain tile resolution and also the map-layers geometries will be low resolution.


> This refers specifically to the single brigher blue polygon? Are the other polygons also drawn as decorations, or something else?
Yes the brighter blue polygon is a decoration.  Decorations are only used to simulate "feature classification".  Other blue polygons are map-layers textures.

mdastous-bentley:
@aruniverse I understand it might be too late, but would be nice to get this included in 4.6.

pmconne (approved):

extensions/map-layers-formats/src/Tools/GeometryTerrainDraper.ts
```diff
@@ -0,0 +1,189 @@
...
+      }
+    }
+
+    return IndexedPolyfaceSubsetVisitor.createSubsetVisitor(mesh, topFacets, 0);
+  }
+
+  public drapeLineString(outStrings: CurvePrimitive[], inPoints: GrowableXYZArray, tolerance: number, range: Range3d): "loading" | "complete" {
+    const tree = this.treeRef.treeOwner.load();
+    if (!tree)
+      return "loading";
```
aruniverse, line 109: nit: i hate the use of literal strings being used over and over in this file; would prefer pulling these into constants.

aruniverse (approved):

aruniverse:
@mergifyio backport release/4.6.x

extensions/map-layers-formats/src/Tools/GeometryTerrainDraper.ts
mdastous-bentley: ok will take note of that in future updates

mdastous-bentley:
@eringram please review

eringram (approved):

extensions/map-layers-formats/src/Tools/GeometryTerrainDraper.ts
mdastous-bentley: @aruniverse can you please resolve

PR #6687, opened 2024-05-01
Add sessionId tracking to mesh export service API call

andremig-bentley:


danieliborra:
cc. @arvindsvenkat 

@andremig-bentley can you test/debug it in DEV and send the sessionId to Arvind, to see if the dashboard shows it?

@arvindsvenkat is the name of the header 'SessionId' correct?

andremig-bentley:
> cc. @arvindsvenkat
> 
> @andremig-bentley can you test/debug it in DEV and send the sessionId to Arvind, to see if the dashboard shows it?

Debugging the backend through Display Test App confirmed the correct session id is being used in the header. The first picture below is the session id being passed through the header, the second picture is the session id which is being used by the backend.
sessionId
sessionId2

markschlosseratbentley (approved):

danieliborra (approved):

PR #6692, opened 2024-05-02
Improvement to SchemaSync for iModel Hub forking/init case

khanaffan:
Changes
### IModel Hub initializing & forking usecase
1. `SyncId` now mean `ContainerId` which of type string (instead of BeGuid) and must be provided by caller when initializing schema sync container. Previously it was random Guid that was problem when forking iModel from specific changeset.
2. For a given iModel a schema sync container can be initialized again with new container. In this case SchemaSync will assign new `containerId` to `SyncId` and also copy `DataVer` from iModel to container.

### Profile upgrades
When iModel goes through profile upgrade its is also pushed to schema sync container. (previously it was not pushed)
1. Under new rule profile version of schema sync container and iModel must follow following rules
   * **On pull**: iModel ECDb profile version <= schema sync container ECDb profile version.
   * **On push**: iModel ECDb profile version >= schema sync container ECDb profile version. 
2. iTwin `IModelDb.upgradeSchemas()`  will ensure profile upgrade is pushed to user.

### Misc
Add ability to pull new schema changes from schema sync on iTwin.js side.

imodel-native: https://github.com/iTwin/imodel-native/pull/744

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -4688,10 +4688,12 @@ export namespace SchemaSync {
         openMode?: OpenMode;
         user?: string;
     }, operation: (access: CloudAccess) => Promise<void>) => Promise<void>;
+    const pull: (iModel: TestCacheIModel) => Promise<void>;
     const // (undocumented)
     initializeForIModel: (arg: {
         iModel: IModelDb;
         containerProps: CloudSqlite.ContainerProps;
+        overrideContainer?: boolean;
     }) => Promise<void>;
     export class SchemaSyncDb extends VersionedSqliteDb {
         // (undocumented)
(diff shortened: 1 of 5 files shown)


pmconne (approved):

core/common/src/ChangesetProps.ts
```diff
@@ -53,6 +53,8 @@ export enum ChangesetType {
   Regular = 0,
   /** changeset *does* contain schema changes. */
   Schema = 1,
+  /** Schema changeset pushed by iModel with SchemaSync enabled */
+  SchemaSync = Schema | 64,
```
khanaffan, line 57: On old code we have condition of equality and not ORing the flag. This mean 65 changeset is detected as Data/Regular changeset if condition say `type===ChangesetType.Schema`. The older code can be on imodel02 and we do not have to worry about 3.x

rschili (approved):

ColinKerr (approved):

Animagne (approved):

aruniverse (approved):

PR #6693, opened 2024-05-02
Fix text rotation around anchor point 

### LINKED ISSUE #6688: Text Annotation orientation should rotate about the anchor point
I expect the property TextAnnotation.orientation to rotate the text geometry about the anchor point.  Currently, it is rotating about the lower left.

**To Reproduce**
Please use the following branch which includes some changes to DisplayTestApp to make it easier to see the bug:
https://github.com/iTwin/itwinjs-core/tree/text-rotation-bug

The keyin sequence below assumes you are using [house.bim](https://bentley-my.sharepoint.com/:u:/p/josh_schifter/EVCf6j_XEQVNhPZ95d_P2KgBdQ44sscW1ESNM5efAxvEng?e=EskLJC) for categoryId and font name.  But any bim file with a 2d view can be used.

1. Start Display Test App, Open an .bim file, choose a 2d view.
2. Enter the following sequence of key-ins:
        dta text init 0x200000000c8
        dta text font arial
        dta text text abcdefghi
        dta text center
        dta text anchor vertical bottom
        dta text rotation 30
        dta text rotation 45
        dta text rotation 90
3. Observe that the text is rotating about the lower left instead of the bottom, middle as expected.

**Expected behavior**
Text should rotate about the anchor point

**Screenshots**
image

- OS: Windows
- iTwin.js Version 4.6.0-dev.19

pmconne:
Fixes #6688.

Added key-ins to display-test-app:
- `dta text rotation <degrees>`
- `dta text anchor <left|center|right|top|middle|bottom>`

Removed `TextAnnotation.origin`. It was not adding anything useful.
Clarified documentation of `TextAnnotationAnchor`.
Fixed inverted addition and subtraction in `TextAnnotation.computeDocumentTransform`, copied from rarely-if-ever-used C++ version.
Made `computeDocumentTransform` rotate about the fixed anchor point.

The results are *almost* correct.
- [x] The text doesn't rotate precisely about the anchor point. Matrix multiplication ordering mistake - fixed.
- [x]  The extremes of the rotated text get clipped out, seemingly by the element's bounding box. Likely a problem with how the bounding box is computed on the backend. This doesn't occur if the rotation is exactly +/-90 degrees. Fixed by https://github.com/iTwin/imodel-native/pull/746

### DIFF
--- a/common/api/core-common.api.md
+++ b/common/api/core-common.api.md
@@ -9458,13 +9458,14 @@ export class TestRpcManager {
 // @beta
 export class TextAnnotation {
     anchor: TextAnnotationAnchor;
+    // @internal (undocumented)
+    computeAnchorPoint(layoutRange: Range2d): Point3d;
     // @internal
     computeDocumentTransform(layoutRange: Range2d): Transform;
     static create(args?: TextAnnotationCreateArgs): TextAnnotation;
     equals(other: TextAnnotation): boolean;
     static fromJSON(props: TextAnnotationProps | undefined): TextAnnotation;
     orientation: YawPitchRollAngles;
-    origin: Point3d;
     textBlock: TextBlock;
     toJSON(): TextAnnotationProps;
 }
@@ -9497,7 +9498,6 @@ export interface TextAnnotationAnchor {
 export interface TextAnnotationCreateArgs {
     anchor?: TextAnnotationAnchor;
     orientation?: YawPitchRollAngles;
-    origin?: Point3d;
     textBlock?: TextBlock;
 }
 
@@ -9505,7 +9505,6 @@ export interface TextAnnotationCreateArgs {
 export interface TextAnnotationProps {
     anchor?: TextAnnotationAnchor;
     orientation?: YawPitchRollProps;
-    origin?: XYZProps;
     textBlock?: TextBlockProps;
 }
 
--- a/common/changes/@itwin/core-backend/pmc-fix-text-anchor_2024-05-03-11-44.json
+++ b/common/changes/@itwin/core-backend/pmc-fix-text-anchor_2024-05-03-11-44.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Fix text annotation rotation around anchor point.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/changes/@itwin/core-common/pmc-fix-text-anchor_2024-05-03-11-44.json
+++ b/common/changes/@itwin/core-common/pmc-fix-text-anchor_2024-05-03-11-44.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
+      "comment": "Remove unused TextAnnotation.origin.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-common"
+}
--- a/core/backend/src/TextAnnotationGeometry.ts
+++ b/core/backend/src/TextAnnotationGeometry.ts
@@ -6,7 +6,7 @@
  * @module ElementGeometry
  */
 
-import { TextAnnotation, TextBlockGeometryProps, TextBlockGeometryPropsEntry, TextString, TextStyleColor } from "@itwin/core-common";
+import { ColorDef, TextAnnotation, TextBlockGeometryProps, TextBlockGeometryPropsEntry, TextString, TextStyleColor } from "@itwin/core-common";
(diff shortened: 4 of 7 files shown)


bbastings (approved):

markschlosseratbentley (approved):

Josh-Schifter:
I don't understand why you removed origin from TextAnnotation.  Is the caller expected to translate the geometry after calling produceTextAnnotationGeometry?   If so, why keep the orientation property?

pmconne:
> I don't understand why you removed origin from TextAnnotation.

TextAnnotation represents a single TextBlock associated with a single element. The element already has an origin. Putting an origin in the annotation doesn't add anything useful.

A different element type (like a dimension) containing multiple TextBlocks would indeed want to provide a way to position the text blocks (ideally, relative to one another and to the host element's origin, IMO).

> Is the caller expected to translate the geometry after calling produceTextAnnotationGeometry?

The caller is expected to tell us where the element is located in world coordinates, by specifying the placement origin. This enables you to change the placement origin, whether persistently or in dynamics, without updating the annotation.

> why keep the orientation property?

Because the extents of the text and therefore the pivot point for rotation are determined by the TextBlock's contents and the annotation's anchor property.

Josh-Schifter:
> TextAnnotation represents a single TextBlock associated with a single element.

I think this is the sticking point.  As far as I know, the only way to get geometry from a TextBlock is to create a TextAnnotation and then call produceTextAnnotationGeometry.  Is that right?

There will be several annotation elements that require multiple TextBlocks each independently positioned relative to that element's single placement.  If 'TextAnnotation' is meant for the single TextBlock case, how should those elements position each TextBlock and produce the geometry?

The way it was before, where each TextAnnotation had its own origin and orientation seemed quite convenient for the general case of an annotation element with multiple TextBlocks.

pmconne:
> If 'TextAnnotation' is meant for the single TextBlock case, how should those elements position each TextBlock and produce the geometry?

Sounds like you will need an exported version of [produceTextBlockGeometry](https://github.com/iTwin/itwinjs-core/blob/8e8f530d79c3769d242acd7d873152e82698efa8/core/backend/src/TextAnnotationGeometry.ts#L128).

bbastings:
> how should those elements position each TextBlock and produce the geometry?
One approach could be to change the localToWorld transform on ElementGeometry.Builder to collect the output relative to the placement transform, could also be done with a separate builder if you don't want to change localToWorld with a single builder...

> Instead, it gives you a way to define text that exists very far away from the element itself, which is not desirable.
I agree that this helps enforce proper use of the element placement. @Josh-Schifter were you using the placement before...or were you creating your geometry at it's world location?

Josh-Schifter:
> Sounds like you will need an exported version of [produceTextBlockGeometry](https://github.com/iTwin/itwinjs-core/blob/8e8f530d79c3769d242acd7d873152e82698efa8/core/backend/src/TextAnnotationGeometry.ts#L128).

Perhaps, although then I will also need computeDocumentTransform  (or I can duplicate it since it is simple).

>I agree that this helps enforce proper use of the element placement. @Josh-Schifter were you using the placement before...or were you creating your geometry at its world location?

I'm still producing all the geometry at the world location.  As you know, whatever I pick for placement origin, the text will always be somewhere else.  So either way I need a way to produce geometry that's transformed relative to the placement.

bbastings:
> So either way I need a way to produce geometry that's transformed relative to the placement.

That's what ElementGeometry.Builder does for you...and if you change localToWorld before calling appendTextBlock you can have it's geometry collected at an offset.

pmconne:
> EDIT: Might be better to re-add origin though, sounds like appendTextBlock doesn't work like everything else now...you can add everything else in world and have the builder automatically transform it relative to the element's placement...

Yes, potentially.
I want to first better understand how Josh will compute the origins of two text blocks within a single dimension though.
I suspect he might have to lay out each of the text blocks to determine their ranges, then compute the origins based on that.

Josh-Schifter:
> I suspect he might have to lay out each of the text blocks to determine their ranges, then compute the origins based on that.

Yes, I will need to know the range in order to determine the origin.  As an example, if the text is too wide, I will need to shift it.

As an example, see the image below.  The '50mm' string fits nicely between the vertical extension lines so it is centered.  By contrast, the '25mm' string is too wide to fit so it and the horizontal dimension line are shifted to the left, also the terminators arrows are reversed.

image

pmconne:
Ok, I can 1) add back the `origin` property and document what it's for and 2) expose a layout function to compute the annotation's range.

bbastings:
Yeah, think we need both...

Now when appendTextBlock calls other append methods they apply worldToLocal...adding back in origin would be necessary for appendTextBlock to account for a placement established by calling setLocalToWorld on the builder (the same as all the other append methods)...

'''
    const builder = new ElementGeometry.Builder();
    const primitive = LineString3d.create(pts);

    builder.setLocalToWorld3d(testOrigin, testAngles); // Establish world to local transform for append...
    const status = builder.appendGeometryQuery(primitive);
'''

PR #6710, opened 2024-05-10
Get rid of imodel02 branch

naveedkhan8067:
The imodel02 branch is not used now.

### DIFF
--- a/common/config/azure-pipelines/templates/core-build.yaml
+++ b/common/config/azure-pipelines/templates/core-build.yaml
@@ -74,7 +74,7 @@ steps:
   - script: node common/scripts/install-run-rush.js audit
     displayName: rush audit
     workingDirectory: ${{ parameters.workingDir }}
-    condition: and(succeeded(), ${{ parameters.runRushAudit }}, ne('${{ parameters.targetBranch }}', 'imodel02'))
+    condition: and(succeeded(), ${{ parameters.runRushAudit }}, ne('${{ parameters.targetBranch }}', 'master'))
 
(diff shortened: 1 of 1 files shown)


DanRod1999 (approved):

common/config/azure-pipelines/templates/core-build.yaml
```diff
@@ -74,7 +74,7 @@ steps:
   - script: node common/scripts/install-run-rush.js audit
     displayName: rush audit
     workingDirectory: ${{ parameters.workingDir }}
-    condition: and(succeeded(), ${{ parameters.runRushAudit }}, ne('${{ parameters.targetBranch }}', 'imodel02'))
+    condition: and(succeeded(), ${{ parameters.runRushAudit }}, ne('${{ parameters.targetBranch }}', 'master'))
```
DanRod1999, line 77: I think `ne('${{ parameters.targetBranch }}', 'master')` can be removed

naveedkhan8067: > I think `ne('${{ parameters.targetBranch }}', 'master')` can be removed

I agree. otherwise audit would not run on master which is a required thing.

DanRod1999 (approved):

aruniverse:
Do we need this pr?

DanRod1999:
> Do we need this pr?

the behavior of the pipeline doesn't change, but it doesn't make sense to include that in the yml file anymore.

aruniverse (approved):

PR #6720, opened 2024-05-13
IndexedDB Tile Cache

andremig-bentley:
Create a local cache with IndexedDB which will be used to store tiles from the mesh export service. The IndexedDBCache class is a standalone class which interfaces with IndexedDB and can be used widely. The class can store, delete, and retrieve content from an IndexedDB database, and has the option to have the content in the database expire. 

If the useIndexedDBCache conditional is set true, tiles being fetched through the itwinjs-batched-models channel will first attempt to fetch their content from an IndexedDB database. If that content cannot be found or is expired, the channel will fetch the content normally and store the content for future requests.


core/frontend/src/tile/IDBTileStorage.ts
```diff
@@ -0,0 +1,182 @@
...
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Tiles
+ */
+
+import { TileRequestChannel } from "./internal";
+
+export class IDBStorage {
```
andremig-bentley, line 11: IDBTileStorage has been changed to IDBStorage, a more generic class that has no mention of tiles. All of its functionality deals solely with the IndexedDB database itself.

core/frontend/src/tile/IDBTileStorage.ts
```diff
@@ -0,0 +1,182 @@
...
+    super(channelName, cacheConcurrency);
+
+    this._tileStorage = new IDBStorage(dbName);
+  }
+
+  public get tileStorage(): IDBStorage {
+    return this._tileStorage;
+  }
+
+  public async fetch(url: string, callback: (url: string) => Promise<Response>): Promise<ArrayBuffer> {
```
andremig-bentley, line 168: This fetch function now takes a callback function so that it can always return content even if the db request is unsuccessful. For now, I've kept this function as a part of the tile request channel, but it can be moved.

extensions/frontend-tiles/src/BatchedTile.ts
```diff
@@ -137,8 +140,9 @@ export class BatchedTile extends Tile {
   public override async requestContent(_isCanceled: () => boolean): Promise<TileRequest.Response> {
     const url = new URL(this.contentId, this.batchedTree.reader.baseUrl);
     url.search = this.batchedTree.reader.baseUrl.search;
-    const response = await fetch(url.toString());
-    return response.arrayBuffer();
+    const response = await this.channel.fetch(url.toString(), fetch);
```
andremig-bentley, line 143: Here is the usage of the new fetch function. For now, I kept this as a part of the tile request channel. No matter what, to access the db like this, we need these tiles to have a member with access to the db. Since the channel already existed and has access to the db, I kept this functionality as a part of the channel. It might be better to revert the channel being used by the batched tiles to a standard TileRequestChannel, and instead give the tiles a member that is an instance of IDBStorage, or some other connector class that communicates between the two. This shouldn't be a difficult change, but needs further discussion.

andremig-bentley:
@danieliborra 

Updates based on your feedback. I believe the IDBStorage class should now be testable on its own, so I'm working on getting that going.

core/frontend/src/tile/IDBTileStorage.ts
danieliborra: This couples things more. Try removing completely IDBTileRequestChannel

core/frontend/src/tile/IDBTileStorage.ts
```diff
@@ -0,0 +1,182 @@
...
+    deleteTransaction.onsuccess = () => {
+      console.log("DELETE TRANSACTION SUCCESS");
+    };
+
+    deleteTransaction.oncomplete = async () => {
+      console.log("DELETE TRANSACTION COMPLETED");
+    };
+  }
+
+  public async addContent(uniqueId: string, content: ArrayBuffer) {
```
danieliborra, line 123: make it private

core/frontend/src/tile/IDBTileStorage.ts
```diff
@@ -0,0 +1,182 @@
...
+      } else {
+        console.log("No matching results found in db");
+      }
+      await this.close();
+      return undefined;
+    };
+    return undefined;
+  }
+
+  public async deleteContent(uniqueId: string) {
```
danieliborra, line 106: make it private

core/frontend/src/tile/IDBTileStorage.ts
```diff
@@ -0,0 +1,182 @@
...
+      initialObjectStore.createIndex("content", "content", {unique: false});
+      initialObjectStore.createIndex("timeOfStorage", "timeOfStorage", {unique: false});
+    };
+  }
+
+  public async close() {
+    await this._db.close();
+  }
+
+  public async requestContent(uniqueId: string): Promise<ArrayBuffer | undefined> {
```
danieliborra, line 63: call it "fetch"

extensions/frontend-tiles/src/BatchedTile.ts
danieliborra: As we discussed, I'd keep this simple, no need to add additional channels here, there is already a channel.
I'd remove the new channel. 
Branch this branch and let's try it out without channels, you will see the code will be more simple and elegant IMO.

danieliborra, line 143:
```suggestion
    const response = await this.cache.fetch(url.toString(), fetch);
```

core/frontend/src/tile/IDBTileStorage.ts
```diff
@@ -0,0 +1,182 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Tiles
+ */
+
```
danieliborra, line 8: Can you create an interface e.g. `IStorageCache` with a virtual function `fetch` and make IDBStorage inherit from it?

Then, as step too, can you create another class `IDBPassThrowCache`, which just recalls fetch? We will need both classes in the future.

danieliborra (changes requested):
Good progress, but we're still not there. As I commented below, I'd suggest branching this branch and see how it looks when removing the new channel.
It won't take you long, and then we can discuss. I think it is worth a try.

andremig-bentley:
IDBTileRequestChannel has now been fully removed, IDBStorage has been renamed to IndexedDBCacheManager (copilot's suggestion). The interface IndexedDBCache has been created with a fetch function, and IndexedDBCacheManager and IndexedDBCachePassThrough implement the interface. Within IndexedDBCacheManager, all functions except for the fetch function have been made private. Within BatchedTiles.ts, the IDBTileRequestChannel has been replaced with _localCache, an instance of IndexedDBCacheManager.

core/frontend/src/tile/IDBTileStorage.ts
andremig-bentley: I decided to keep this fetch function as simple as possible to read since it is the one public function, meaning that instead of having the actual cache searching in this function, I kept the cache searching separate. I think that if the fetch and searchCache functions were combined it would be a much more complicated function to read, but if you think they should be combined, I can certainly do that.

danieliborra:
> IDBTileRequestChannel has now been fully removed, IDBStorage has been renamed to IndexedDBCacheManager (copilot's suggestion). The interface IndexedDBCache has been created with a fetch function, and IndexedDBCacheManager and IndexedDBCachePassThrough implement the interface. Within IndexedDBCacheManager, all functions except for the fetch function have been made private. Within BatchedTiles.ts, the IDBTileRequestChannel has been replaced with _localCache, an instance of IndexedDBCacheManager.

It looks way better! All changes are good.
In this case, co-pilot didn't do great job 😄 
What about:
- IndexedDBCache => ILocalCache or ICloudContentCache or IFetchContentCache (generic name, has nothing to do with IndexDB)
- IndexedDBCacheManager => IndexedDBCache
- IndexedDBCachePassThrough => PassThroughCache

extensions/frontend-tiles/src/BatchedTile.ts
```diff
@@ -29,6 +29,7 @@ export class BatchedTile extends Tile {
   private readonly _unskippable: boolean;
   /** Transform from the tile's local coordinate system to that of the tileset. */
   public readonly transformToRoot?: Transform;
+  private _localCache: any;
```
danieliborra, line 32:
```suggestion
  private _localCache: ILocalCache;
```

andremig-bentley:
Updates from most recent commit:

- Fixed all async issues/issues with database functions happening out of order, etc. Confirmed now that all database functionality is working as intended (searching, retrieving, adding, deleting, all working how/when they're supposed to).
- General cleanup, changed IDBTileStorage.ts to IndexedDBCache.ts
- Moved all changes dealing with IModelTileRequestChannels to a new branch and removed them from this PR

markschlosseratbentley (changes requested):
extract-api

andremig-bentley:
> extract-api

Taken care of.

core/frontend/src/tile/TileAdmin.ts
```diff
@@ -161,6 +161,8 @@ export class TileAdmin {
   public readonly contextPreloadParentSkip: number;
   /** @beta */
   public readonly cesiumIonKey?: string;
+  /** @beta */
```
andremig-bentley, line 164: Potential location for new conditional to use/not use cache, can definitely be moved if somewhere else is better.

danieliborra (approved):

markschlosseratbentley:
is extract-api failing again?

andremig-bentley:
> is extract-api failing again?

Yeah it is, not sure why. I re-ran it and re-committed everything applicable with my latest commit, but it doesn’t seem to be working. Looking into that this morning

markschlosseratbentley:
> > is extract-api failing again?
> 
> Yeah it is, not sure why. I re-ran it and re-committed everything applicable with my latest commit, but it doesn’t seem to be working. Looking into that this morning

All good.  Approving

markschlosseratbentley (approved):

common/api/core-frontend.api.md
```diff
@@ -5989,6 +5989,12 @@ export class IdleTool extends InteractiveTool {
     static toolId: string;
 }
 
+// @beta (undocumented)
+export interface ILocalCache {
```
pmconne, line 5993: > Do not use I as a prefix for interface names.

Please make sure you have reviewed our [coding guidelines](https://www.itwinjs.org/learning/guidelines/typescript-coding-guidelines/#names)

common/api/core-frontend.api.md
```diff
@@ -7365,6 +7371,23 @@ export class IndexBuffer {
     toVertexIndices(): VertexIndices;
 }
 
+// @beta (undocumented)
```
pmconne, line 7374: beta APIs must be documented.

core/frontend/src/test/tile/IndexedDBCache.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  public async convertStringToBuffer(str: string): Promise<ArrayBuffer> {
+    return new TextEncoder().encode(str).buffer;
+  }
+
+  public async convertBufferToString(buffer: ArrayBuffer): Promise<string> {
+    return new TextDecoder().decode(new Uint8Array(buffer)).toString();
+  }
+}
+
+describe.only("IndexedDBCache", () => {
```
pmconne, line 42: rm .only

pmconne (changes requested):
If this is no longer "WIP" then please update your PR description to summarize what you are doing and why.

core/frontend/src/tile/IndexedDBCache.ts
```diff
@@ -0,0 +1,185 @@
...
+  private _expirationTime?: number;
+
+  public constructor(dbName: string, expirationTime?: number) {
+    this._dbName = dbName;
+    if (expirationTime) {
+      this._expirationTime = expirationTime;
+    }
+  }
+
+  protected async open(dbCache: IndexedDBCache){
```
pmconne, line 35: This is an instance method, why does it take an IndexedDBCache argument instead of operating on `this`?

core/frontend/src/tile/IndexedDBCache.ts
```diff
@@ -0,0 +1,185 @@
+
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+/** @beta */
```
pmconne, line 7: Why do we want these APIs to be part of the package's public API? They seem like implementation details.

core/frontend/src/tile/IndexedDBCache.ts
```diff
@@ -0,0 +1,185 @@
...
+          await this.close();
+          return resolve(content);
+        }
+
+        await this.close();
+        return resolve(undefined);
+      };
+
+      storedResponse.onerror = async () => {
+        await this.close();
```
pmconne, line 108: Log errors.

core/frontend/src/test/tile/IndexedDBCache.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  it("should fetch content from the network if expired in the cache", async () => {
+    const pause = async (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+    const expiringCache = new TestCache("ExpiringTestDB", 1);
+    const uniqueId = "test-expiring";
+    const content = await cache.convertStringToBuffer(uniqueId);
+
+    await expiringCache.testAddContent(uniqueId, content);
+    await pause(1000);
```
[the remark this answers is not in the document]
andremig-bentley: Do you mean because of the pause?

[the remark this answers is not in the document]
andremig-bentley: I changed the test to now have an expiration time of -1 so that it expires immediately with no reliance on a pause. I changed the test where the cache does not expire to have a much larger expiration time to avoid flakiness in that test as well.

core/frontend/src/tile/TileAdmin.ts
```diff
@@ -1274,6 +1277,13 @@ export namespace TileAdmin { // eslint-disable-line no-redeclare
      * @alpha This was primarily introduced because the electron version of certa does not serve local assets, so the tests can't locate the worker script.
      */
     decodeImdlInWorker?: boolean;
+
+    /** If true, an indexedDB database will be added to the itwinjs-batched-models tile request channel for use as a local cache.
+     * Tiles requested through this channel will first be searched for in the database, and if not found, will be requested from the backend, then stored for later requests.
+     * Default value: false
+     * @alpha
+     */
+    useIndexedDBCache?: boolean;
```
pmconne, line 1286: Can you move this to `FrontendTilesOptions`?

andremig-bentley: Moved it to `FrontendTilesOptions` and added it wherever necessary within the rest of the options functionality. Let me know if I missed something or incorrectly configured some of it.

extensions/frontend-tiles/src/FrontendTiles.ts
```diff
@@ -25,6 +25,9 @@ function createMeshExportServiceQueryUrl(args: { iModelId: string, urlPrefix?: s
   if (args.enableCDN)
     url = `${url}&cdn=1`;
 
+  if (args.useIndexedDBCache)
+    url = `${url}&idb=1`;
```
pmconne, line 29: Why do you need to alter the URL based on whether or not the cache is used?

pmconne: Now you can remove `useIndexedDBCache` from the two `Args` types.

andremig-bentley: Yeah I saw that too, just pushed that change a minute ago

extensions/frontend-tiles/src/BatchedTile.ts
```diff
@@ -29,6 +30,7 @@ export class BatchedTile extends Tile {
   private readonly _unskippable: boolean;
   /** Transform from the tile's local coordinate system to that of the tileset. */
   public readonly transformToRoot?: Transform;
+  private _localCache: LocalCache;
```
pmconne, line 33:
```suggestion
  private readonly _localCache: LocalCache;
```

pmconne (approved):

pmconne:
Force-merging because unrelated failing reality data integration test.

PR #6724, opened 2024-05-13
Add tests and API for getListOfGCS

EricPaquetB:
A new function (getListOfGCS) was created in imodel-native. 
This PR contains tests for this function, and an API to make it available publicly in itwinjs-core.

imodel-native: https://github.com/iTwin/imodel-native/pull/756


core/backend/src/GeoCoordServices.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @public
+ */
+export abstract class GeoCoordServices {
+  /** Get a list of Geographic Coordinate Systems.
+   * @param ignoreLegacy If true, only return GCS's that are not considered legacy.
+   * @param extent If provided, only return GCS's that contain the given extent. Minimum longitude and latitude correspond to extent.low.x and extent.low.y, respectively.
+   * Maximum longitude and latitude correspond to extent.high.x and extent.high.y, respectively.
+   * @returns The list of Geographic Coordinate Systems, according to the supplied parameters.
+   */
+  public static getListOfGCS(ignoreLegacy: boolean, extent?: Range2dProps): Array<{ name: string, description: string }> {
```
pmconne, line 18:
```suggestion
/** APIs for working with geographic coordinates.
 * @public
 */
export namespace GeoCoordServices {
  /** Get a list of Geographic Coordinate Systems.
   * @param ignoreLegacy If true, only return GCS's that are not considered legacy.
   * @param extent If provided, only return GCS's that contain the given extent. Minimum longitude and latitude correspond to extent.low.x and extent.low.y, respectively.
   * Maximum longitude and latitude correspond to extent.high.x and extent.high.y, respectively.
   * @returns The list of Geographic Coordinate Systems, according to the supplied parameters.
   */
  export getListOfGCS(ignoreLegacy: boolean, extent?: Range2dProps): Array<{ name: string, description: string }> {
```

pmconne, line 18: Define what "legacy" means. Change function to take a single "options" object containing `ignoreLegacy` and `extent` properties.

core/backend/src/GeoCoordServices.ts
```diff
@@ -0,0 +1,21 @@
...
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import { IModelJsNative } from "@bentley/imodeljs-native";
+import { Range2dProps } from "@itwin/core-geometry";
+
+/** Class containing services for working with geographic coordinates.
+ * @public
+ */
+export abstract class GeoCoordServices {
+  /** Get a list of Geographic Coordinate Systems.
```
pmconne, line 12: We use both GCS (geographic coordinate system) and CRS (coordinate reference system) in our docs and APIs...CRS is the more common term, isn't it?

EricPaquetB: There is already a "GeoCoordConfig.ts" file / class in the same directory. I would hesitate to use another term at this level. I think this would bring confusion.

core/backend/src/GeoCoordServices.ts
```diff
@@ -0,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import { IModelJsNative } from "@bentley/imodeljs-native";
+import { Range2dProps } from "@itwin/core-geometry";
+
+/** Class containing services for working with geographic coordinates.
+ * @public
```
aruniverse, line 9: id rather we start off with `@beta` and work up to public

core/backend/src/GeoCoordServices.ts
aruniverse, line 18: to make this more robust, maybe accept an obj instead with named params, instead of a list of args

core/backend/src/GeoCoordServices.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export abstract class GeoCoordServices {
+  /** Get a list of Geographic Coordinate Systems.
+   * @param ignoreLegacy If true, only return GCS's that are not considered legacy.
```
aruniverse, line 13: do we have any docs on what is defined as "legacy" gcs?

AlainRobertAtBentley: In the iTwinjs-core we instead use the term 'deprecated' and is represented as a bool in the HorizontalCRS props at: https://github.com/iTwin/itwinjs-core/blob/9ec9aab98c6893f49065e6a7e14862115eb8f650/core/common/src/geometry/CoordinateReferenceSystem.ts#L97
Internally in the CSMAP compliant database we mark deprecated CRS by indicating they belong to the LEGACY group. We could normalize to deprecated and rename legacy everywhere.
The meaning of 'deprecated' is the same as in the official EPSG database. The entry is still valid and can be used when older data is being referenced, but shall not be used for newer dataset.

pmconne: So probably the option should be `includeDeprecated` (default false) instead of `ignoreLegacy`.

core/backend/src/GeoCoordServices.ts
EricPaquetB: Do you mean to change GCS of CRS in every file (and file name) that I touched ? Because it creates a bit of confusion is some cases since "GCS" is already used for some base types:

bvector<CRSListResponseProps> GeoServicesInterop::GetListOfCRS(bool ignoreLegacy, DRange2dCP extent)
    {
    static GeoCoordinates::BaseGCSPtr ll84CRSPtr = GeoCoordinates::BaseGCS::CreateGCS("LL84");
    bvector<CRSListResponseProps> listOfCRS;

    char csKeyName[128];
    CRSListResponseProps props;
    for (int index = 0; (0 < GeoCoordinates::CSMap::CS_csEnum(index, csKeyName, sizeof(csKeyName))); index++)
        {
        GeoCoordinates::BaseGCSPtr crs = GeoCoordinates::BaseGCS::CreateGCS(csKeyName);
...

pmconne: We don't care about filenames nor about the C++ APIs, only the core-backend APIs.

core/backend/src/GeoCoordServices.ts
EricPaquetB: I don't use "ignoreLegacy" as a parameter anymore. But the "deprecated" term is defined where it appears in the API.

core/backend/src/GeoCoordServices.ts
EricPaquetB: I'm now using "CRS" in the core-backend APIs.

core/backend/src/GeoCoordServices.ts
EricPaquetB: Now using @beta

core/backend/src/GeoCoordServices.ts
EricPaquetB: There is now only 1 parameter, but I'm using an object instead of a single argument anyway, in case of future expansion.

core/backend/src/GeoCoordServices.ts
EricPaquetB: The "includeDeprecated" parameter is not necessary anymore. We decided to return a "deprecated" field with the results.

core/backend/src/GeographicCRSServices.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import { Range2dProps } from "@itwin/core-geometry";
+import { IModelHost } from "./IModelHost";
+
+/** Describes a geographic coordinate reference system produced by [[getAvailableCoordinateReferenceSystems]].
+ * @beta
+ */
+export interface AvailableCoordinateReferenceSystemProps {
+  /** TODO: Document the fields. */
```
pmconne, line 16: Document each field of the available CRS. e.g., what can I do with the name - what APIs can I pass it to? What does "deprecated" mean and under what circumstances should I use or avoid using a deprecated CRS?

You will need to run `build` and then `extract-api` after you add documentation.

EricPaquetB: Done. Running extract-api only changed slightly common/api/core-backend.api.md. I don't know if this is what was expected.

core/backend/src/test/misc/GeoServices.test.ts
```diff
@@ -804,6 +805,45 @@ describe("GeoServices", () => {
...
+          const crsExtentRange: Range2d = Range2d.fromJSON(crs.crsExtent);
+          const intersects = extentRange.intersectsRange(crsExtentRange);
+          assert.isTrue(intersects);
+        }
+      }
+
+      assert.isTrue(listOfCRS.length > expectedCount - allowedRange && listOfCRS.length < expectedCount + allowedRange);
+    };
+
+    it("should get all CRS", async () => {
```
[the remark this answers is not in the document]
EricPaquetB: Done

core/backend/src/GeographicCRSServices.ts
pmconne: Yep, as expected.

pmconne (approved):

pmconne:
This integration test has failed twice. Seems unrelated to your changes at a glance.

```
1) RealityDataAccess (#integration)
     should get RealityDataSource for reality data attachment in iModel:
   AssertionError: expected undefined not to be undefined
    at Context.<anonymous> (home/vsts/work/1/s/full-stack-tests/core/src/frontend/standalone/RealityDataAccess.test.ts:252:27)
```

Issue #6583, opened 2024-04-04
Crash saving geomstream with auxChannel data


Rock2000:
We are creating geometric elements that represent a 3d surface. We tried to add auxChannel data in order to use the Scientific Visualization feature, but after saving the element with the auxChannel data the platform crashes. 

**To Reproduce**
Attached is the mesh with aux data that was used. The aux data in this test is Scalar data that is simply zeros. Save this polyface to a geomstream. The element update call will succeed, but afterwards the platform will disappear. I presume when the tiles are being regenerated.

[aux.txt](https://github.com/iTwin/itwinjs-core/files/14872137/aux.txt)

**Desktop (please complete the applicable information):**

- OS: Win11
- Browser Electron
- iTwin.js Version 4.4.0-dev.10

pmconne:
Show your code please.

Rock2000:
The polyface creation is in the attached file. Then just create the geomstream

    const builder = new CoreElementGeometry.Builder();

    const meshParams = new GeometryParams(categoryId);
    meshParams.fillDisplay = FillDisplay.ByView;
    meshParams.lineColor = sourceParams.color;
    meshParams.fillColor = meshParams.lineColor;
    meshParams.weight = 1;
    builder.appendGeometryParamsChange(meshParams));
    builder.appendGeometryQuery(polyface);
	
    const builderParams = { viewIndependent: false, entryArray: builder.entries };

and assign this to a geometric element's elementGeometryBuilderParams property. It works fine without the auxChannel data.

Rock2000:
I might be reading it wrong, but if the test just stores the geomstream into the database and no other services are running than I expect it won't show the problem. The actual update of the element is not what crashes. While I can't be sure, I believe it is either the tile generation based on this new geomstream, or somehow rendering the tile. 
It may actually be easier to reproduce this way. Here is an imodel that has the stored geomstream (not the exact same as the sample code, but conceptually the same, with zero scalar values). If you open the imodel the platform crashes before the view is shown. You need all the files supplied to see the issue (the bim file alone will open successfully)

[booster.zip](https://github.com/iTwin/itwinjs-core/files/14876315/booster.zip)

pmconne:
Gonna need **actual repro steps**.

pmconne:
Alternatively, consider using the debugging, logging, and crash reporting capabilities built into Studio to capture a native stack trace.

Rock2000:
[...]

PR #6726, opened 2024-05-14
Fix PolyfaceAuxData index blocking flatbuffer format / Reorganize test data

dassaf4:
Addresses https://github.com/iTwin/itwinjs-core/issues/6583.

TypeScript FlatBuffer serialization of `PolyfaceAuxData` indices lacked the 0-blocking of the other `Polyface` index arrays, which led to crashes during mesh processing. Now TypeScript FlatBuffer serialization outputs the correct index blocking, and both TypeScript and native FlatBuffer APIs read both old and new PolyfaceAuxData index formats. The old format is detected using heuristics.

Although there is a similar imodel-native PR, the fixes in each repo are independent of the other.

Also:
* Typescript FlatBuffer and JSON deserialization APIs were incorrectly reading fixed-sized `Polyface` indices, which have `numPerFace > 2` and are only written by the native FlatBuffer API. Move index array reblocking logic to `SerializationHelpers` for common usage.
* New crossPlatform.test.ts verifies that Typescript API can read native API geometry serializations. (The same test exists in imodel-native to verify the native API can read TypeScript API geometry serializations.) Each test case consists of at least four files: FlatBuffer and JSON serializations of a single geometry, written by both APIs. The test verifies that a given API deserializes each file in a test case into the same geometry. Initial test cases include `Polyface` variations in `numPerFace` value, `PolyfaceAuxData` format, and all combinations thereof.
* `IndexedMeshProps` was incomplete and used the wrong label for `TaggedNumericDataProps`.
* Add "Props" interfaces for `PolyfaceAuxData` and use them in the JSON deserializer.
* Use correct `assert(!"foo")` idiom
* Split out index validation from `IndexedPolyface.terminateFacet` so that it can be called once after all facets are added, instead of after each facet. This new index validation caught at least one bug in a mesh created for a `PolyfaceQuery` test.
* Prefer class members of form `foo?: bar` over `foo: bar | undefined`
* `PolyfaceData.compress` now compresses `PolyfaceAuxData` if exactly one `AuxChannelData` is present
* `Sphere` ctor now validates latitude sweep to avoid common user error. `Sphere.createEllipsoid` was erroneously capturing inputs.
* Add type predicates to `Checker.testDefined` and `Checker.testUndefined` to simplify many unit tests. The linter pointed out the simplifications to the tests.
* `GeometryCoreTestIO`: add helper functions for tests that perform FlatBuffer/JSON <-> file/`GeometryQuery` conversions
* Move all unit test input data into one "data" directory, similar to native geomlibs. Hardcoded data paths in tests are updated.
* Improve documentation for `PolyfaceAuxData`. In particular emphasize that the indices must have the same blocking as the other `Polyface` index arrays.


saeeedtorabi (approved):
I only reviewed the last commit where you extended `getPointCurveClosestApproachXYNewton` to support LineString. Looks good!

dassaf4:
> Approving but consider putting the massive code reorganization changes into a separate PR or at the very least changing the PR title to describe what's actually being changed.

Thanks. Yeah, this one got out of control. Will definitely do next time.

PR #6728, opened 2024-05-15
Fix finalize release job

hl662:
Finalize release recently [failed](https://github.com/iTwin/itwinjs-core/actions/runs/9096507413) because of a permission error when trying to install npm packages globally. 

If certain commands need to be run with elevated privileges within Github's hosted runners, their [docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#administrative-privileges) state we can use `sudo`.

Added `sudo` to the 2 `npm install -g` commands within finalize release yaml.

Ran finalize release workflow [manually](https://github.com/iTwin/itwinjs-core/actions/runs/9097784491) on this branch, confirmed the packages were installed, and `zx` commands were able to run inside of `update-changelogs.mjs`


.github/workflows/finalize-release.yaml
```diff
@@ -29,8 +29,8 @@ jobs:
           git config --local user.name imodeljs-admin
       - name: Run update-changelogs.mjs
         run: |
-          npm i -g zx
-          npm install -g @microsoft/rush
+          sudo npm i -g zx
+          sudo npm install -g @microsoft/rush
           zx --install ./.github/workflows/automation-scripts/update-changelogs.mjs
```
aruniverse, line 34: could we just do something like 

```
npx zx@latest --install ./.github/workflows/automation-scripts/update-changelogs.mjs
```

.github/workflows/finalize-release.yaml
```diff
@@ -29,8 +29,8 @@ jobs:
           git config --local user.name imodeljs-admin
       - name: Run update-changelogs.mjs
         run: |
-          npm i -g zx
-          npm install -g @microsoft/rush
+          sudo npm i -g zx
```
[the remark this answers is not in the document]
hl662: Nope, github's linux and mac runners use passwordless sudo

.github/workflows/finalize-release.yaml
hl662: yep, that works, I made the change

hl662:
@anmolshres98 I'll leave the merge to you

anmolshres98:
@Mergifyio backport release/4.6.x

PR #6733, opened 2024-05-17
initial editing refactor

christophermlawson:
I've included all the Error design into this PR now.  It was already big, and since the error design changes pretty much affected the same code, I decided to include it all together.


core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -111,69 +86,65 @@ export class SchemaContextEditor {
...
+  public async addCustomAttribute(schemaKey: SchemaKey, customAttribute: CustomAttribute): Promise<void> {
     const schema = (await this.schemaContext.getCachedSchema<MutableSchema>(schemaKey, SchemaMatchType.Latest));
     if (schema === undefined)
-      return { errorMessage: `Schema Key ${schemaKey.toString(true)} not found in context` };
+      throw new ECEditingError(ECEditingStatus.SchemaNotFound, `Schema Key ${schemaKey.toString(true)} not found in context`);
 
     schema.addCustomAttribute(customAttribute);
 
-    const diagnostics = Rules.validateCustomAttributeInstance(schema, customAttribute);
+    const diagnosticIterable = Rules.validateCustomAttributeInstance(schema, customAttribute);
```
StefanApfel-Bentley, line 137: Since the iterable is only been used in the for loop in line 140, we should probably just use it there:
`for await (const diagnostic of Rules.validateCustomAttributeInstance(schema, customAttribute)) {`

core/ecschema-editing/src/Editing/Constants.ts
```diff
@@ -43,27 +52,34 @@ export class Constants {
...
+    let newConstant: MutableConstant;
+    try {
+      newConstant = (await schema.createConstant(constantProps.name)) as MutableConstant;
+    } catch (e) {
+      if (e instanceof ECObjectsError && e.errorNumber === ECObjectsStatus.DuplicateItem) {
+        throw new ECEditingError(ECEditingStatus.SchemaItemNameAlreadyExists, `Constant ${constantProps.name} already exists in the schema ${schema.fullName}.`);
+      } else {
+        throw new ECEditingError(ECEditingStatus.Unknown, `Failed to create Constant ${constantProps.name} in schema ${schema.fullName}.`);
+      }
+    }
```
StefanApfel-Bentley, line 80: Since we have that block repeating very often, I wonder if we can capsule it away. I could imagine something like this:
```
const newConstant = await schema.createConstant(constantProps.name)
    .catch(handleItemCreationError("Constant", constantProps.name), schema.fullName));
```

the handleItemCreationError has the three variable parts of the repeating code block and would then look like this:
```
function handleError(type: string, name: string, schema: string) {
  return (e: unknown) => {
     if (e instanceof ECObjectsError && e.errorNumber === ECObjectsStatus.DuplicateItem) {
        throw new ECEditingError(ECEditingStatus.SchemaItemNameAlreadyExists, `${type} ${name} already exists in the schema ${schema}.`);
      } else {
        throw new ECEditingError(ECEditingStatus.Unknown, `Failed to create ${type} ${name} in schema ${schema}.`);
      }
  }
}
```

So if createConstant fails, the promise calls the catch callback, there we can check the type and rethrow an updated error up again to the promise which will eventually fail and return the error back to the caller.

core/ecschema-editing/src/Editing/Formats.ts
```diff
@@ -6,65 +6,68 @@
...
+      const schema = await this.getSchema(schemaKey);
+      const boundCreate = schema.createFormat.bind(schema);
+      newFormat = await this.createSchemaItemFromProps<Format>(schemaKey, this.schemaItemType, boundCreate, formatProps) as MutableFormat;
+    } catch (e: any) {
+      throw new SchemaEditingError(ECEditingStatus.CreateSchemaItemFromPropsFailed, schemaItemIdentifierFromName(schemaKey, this.schemaItemType, formatProps.name!), e);
+    }
 
     await newFormat.fromJSON(formatProps);
-    return { itemKey: newFormat.key };
+    return newFormat.key;
```
StefanApfel-Bentley, line 71: maybe we should capsule the whole json reading in the try block, otherwise if something fails on reading the props (line 70) it would not be wrapped with the SchemaEditingError.

Aside from that, we we actually need line 70? as far I understand the code, this.createSchemaItemsFromProps internally calls newItem.fromJSON already itself.

core/ecschema-editing/src/Editing/Constants.ts
```diff
@@ -17,20 +18,28 @@ import { MutableConstant } from "./Mutable/MutableConstant";
...
-    if (newConstant === undefined) {
-      return { errorMessage: `Failed to create class ${name} in schema ${schemaKey.toString(true)}.` };
+    let newConstant: MutableConstant;
+    try {
+      newConstant = (await schema.createConstant(name)) as MutableConstant;
+    } catch (e) {
+      if (e instanceof ECObjectsError && e.errorNumber === ECObjectsStatus.DuplicateItem) {
+        throw new ECEditingError(ECEditingStatus.SchemaItemNameAlreadyExists, `Constant ${name} already exists in the schema ${schema.fullName}.`);
+      } else {
+        throw new ECEditingError(ECEditingStatus.Unknown, `Failed to create Constant ${name} in schema ${schema.fullName}.`);
```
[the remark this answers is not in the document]
christophermlawson: The innerError property is of type AnyEditingError which has the standard Error as a supported type.

core/ecschema-editing/src/Editing/Constants.ts
christophermlawson: My refactor should address all this redundancy.

StefanApfel-Bentley (approved):

ColinKerr (approved):

example-code/snippets/src/pcf/DynamicSchema.ts
```diff
@@ -129,25 +129,19 @@ async function createDynamicSchema(
...
-          const propResult = await editor.entities.createPrimitiveProperty(entityResult.itemKey, prop.name, prop.type as any);
-          if (!propResult.itemKey)
-            throw new Error(`Failed to create EC Property - ${propResult.errorMessage}`);
+          await editor.entities.createPrimitiveProperty(entityKey, prop.name, prop.type as any);
         }
       }
     }
   };
 
   const createRelClasses = async (schema: MetaSchema) => {
```
aruniverse, line 142: is this fn even needed anymore?

calebmshafer (approved):

example-code/snippets/src/pcf/DynamicSchema.ts
christophermlawson: Do you mean the example-code?  If so, not sure.  the createPrimitiveProperty method still exists, but we now throw errors instead of 'result' objects.

christophermlawson: Sorry mis-read your comment. Yes, the function is still needed.

PR #6734, opened 2024-05-20
Update getSchema in SchemaJsonLocater to be truly async

RohitPtnkr1996:
SchemaJsonLocater loads requested schema synchronously, which causes the main thread to block when loaded schema is large or has many referenced schemas.

SchemaJsonLocater.getSchema should have an asynchronous implementation instead of relying on a synchronous one.


core/ecschema-metadata/src/SchemaJsonLocater.ts
```diff
@@ -27,12 +27,17 @@ export class SchemaJsonLocater implements ISchemaLocater {
...
    * @param schemaKey The [SchemaKey] that identifies the schema.
-   * @param matchType The [SchemaMatchType] to used for comparing schema versions.
+   * @param _matchType The [SchemaMatchType] to used for comparing schema versions.
    * @param context The [SchemaContext] used to facilitate schema location.
    * @throws [ECObjectsError]($ecschema-metadata) if the schema exists, but cannot be loaded.
    */
-  public async getSchema<T extends Schema>(schemaKey: Readonly<SchemaKey>, matchType: SchemaMatchType, context: SchemaContext): Promise<T | undefined> {
-    return this.getSchemaSync(schemaKey, matchType, context) as T;
+  public async getSchema<T extends Schema>(schemaKey: Readonly<SchemaKey>, _matchType: SchemaMatchType, context: SchemaContext): Promise<T | undefined> {
+    const schemaProps = this._getSchema(schemaKey.name);
```
ColinKerr, line 35: so long as the SchemaPropsGetter is sync we will still be blocking.  In @grigasp case the call to get schemas is a network request so we'd block until the request is completed.

grigasp: Not really... 

Our library may be used both on the backend and the frontend, and my specific situation is about the backend. There, the `IModelDb.getSchemaProps` is sync and it seems there's not much you can do about that. The problem for us is that the whole `SchemaJsonLocater.getSchema` is synchronous and doesn't give a chance for other tasks to be executed until it's complete. Parsing the schema asynchronously should at least partially fix the problem (the first part - `IModelDb.getSchemaProps` may still take a long time, but at least the second part would be async). However, I'd like to point out that `Schema.fromJSON` lies about being async - it's actually calling `fromJSONSync` internally. This is why I had to add `await BeDuration.wait(0)` between getting schema props and parsing the schema.

If schema props are retrieved using a network request, the function that does that MUST be async, which means it's incompatible with `SchemaJsonLocater` as it is defined now. On the frontend we use our library with `ECSchemaRpcLocater`.

ColinKerr: Good point.

We should have all the code necessary to write a property async schema from json ... The SchemaReadHelper has an async readSchema method that should behave the same a the readSchemaSync method.  We should refactor that as part of this change.

wrt getting the json from the backend ... we should look into being able to do that off the main thread ... but that should be a separate issue.

ColinKerr: I'm sorry I think I misread this.  Schema.fromJSON doesn't do any async operations so there is really no difference between a sync and an async version.  I think the problem with this locater is that it returns the whole schema when you call 'getSchemaInfo'.  The SchemaInfo is the schema key plus the referenced schemas keys.  If we have this we can then make the request to load each reference asynchronously.  

The ECSchemaRpcLocater does this correctly, I think we need to update this locater to match that behavior.

I don't think there is actually any need to change the fromJSON to have its own implementation at that point.

ColinKerr (approved):

core/ecschema-metadata/src/SchemaJsonLocater.ts
ColinKerr: looks good, @grigasp what do you think?

core/ecschema-metadata/src/SchemaJsonLocater.ts
```diff
@@ -41,7 +42,15 @@ export class SchemaJsonLocater implements ISchemaLocater {
    * @param matchType The match type to use when locating the schema
    */
   public async getSchemaInfo(schemaKey: Readonly<SchemaKey>, matchType: SchemaMatchType, context: SchemaContext): Promise<SchemaInfo | undefined> {
-    return this.getSchema(schemaKey, matchType, context);
+    const schemaProps = this._getSchema(schemaKey.name);
+    if (!schemaProps)
+      return undefined;
+
+    const schemaInfo = await Schema.startLoadingFromJson(schemaProps, context || new SchemaContext());
```
grigasp, line 49: `|| new SchemaContext()` seems unnecessary

grigasp (approved):

core/ecschema-metadata/src/SchemaJsonLocater.ts
grigasp: LGTM

core/ecschema-metadata/src/SchemaJsonLocater.ts
ColinKerr: I think this was copied from the ECSchemaRpcLocater ... context used to be `SchemaContext | undefined`.  Since it should always be defined now I think it's fine to remove it

RohitPtnkr1996: I kept that in since getSchemaSync from SchemaJsonLocater also had this check.
I've removed it now.

PR #6753, opened 2024-05-22
Updated relationship constraint class names for schema merging

AureMonke:


common/changes/@itwin/ecschema-editing/schema-editing-relationship-class-merger-fix_2024-05-22-13-57.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-editing",
+      "comment": "Updated relationship constraint class names for merging",
```
StefanApfel-Bentley, line 5: I had a discussion with Grigas a while back, he recommended to only use comments if we change something that could be interesting to the users. As those PR only fixes our relationship merging, I'd propose to leave that commentline empty, that it won't be included to the itwinjs changelog

StefanApfel-Bentley (approved):

PR #6755, opened 2024-05-22
Fix map masking when latest Mesh Export tiles are used

DStradley:
Fix a problem where maps masked by models are not consistently paying attention to models being turned off, when used with the latest Mesh Export tiles.

https://github.com/iTwin/itwin-graphics-backlog/issues/215

During extensive testing we uncovered a few other issues that will be addressed in later PRs
 - New tiles produce a lower resolution mask on average than old tiles.
 - Some views in both old and new tiles produce low resolution masks
 - New tiles don't update mask immediately for some view changes (element overrides and category changes)
 - Need to add DTA UI for optionally applying feature overrides on models and subcategories (not just elements), so that we can create some additional test cases.
 - Need to create some additional test cases


DStradley:
@pmconne I see two ways to fix the second part of this problem:

1. In `FeatureOverrides.buildLookupTable`, add a conditional to select where to get the `BranchStack` from, based on planar classification being active (this was tried in the Fix commit).
2. In `PlanarClassifier`'s constructor, change `this._branchStack` to use the one from `Target` rather than creating a new one (this is in the Fix Alt commit).

Both of these fix the reported problem and both pass ImageTests, but I am unsure of possible side effects of #2 and wondering which you thought was the better approach?

Also, if you prefer #2, should we make the same change to `BackgroundMapDrape`, which also uses its own copy of a `BranchStack`?

pmconne:
> The second part was due to planar classification (which does the map masking) using a different BranchStack than the one on Target.

Please explain why this is a problem.

DStradley:
@pmconne To answer a previous question, I thought that was a problem because the getFeatureAppearance is looking at the Targets branch while it is drawing the mask, and it did not have the AppearanceOverrideProvider there.  By sharing that, it fixed and was correct for the problem in the issue case, which turned out to be a Priority mode case.

However, as you pointed out, we would not want that for the Model mode case.
During looking at the Model mode case (and the others), we discovered that those appeared to be working because in fact it was using old tiles to get the answer, which of course we don't want.

So we have been trying to come up with a set of FeatureOverrides that will work with Batched models for this, but we have had to jump through several hoops to get it to use the combinations of models, categories and element feature overrides in a manner consistent with the way old tiles currently work (as well as figure out what that is in more detail), but basically it is:
 - Priority mode - everything visible in the view creates the mask, so turning off models, categories, or elements (via fully transparent overrides) will make things not be in the mask (or view).
- Model mode - defines a set of models that the mask is based upon, regardless of what model is on or off in the view.  However, in this mode, things that are off from category settings or element feature overrides in the view still also make things not be in the mask for these models.
- The other 3 modes (element, exclude elements and subcategory) are fairly straightforward and operate based on specified sets of models/subcategories/elementIds, and don't have anything to do with the view settings.
And we also had to do something to come up with the list of models (and apply them) for the new tiles.

I also made a set of 11 new tests so far and now have those working, so I think this is close, but could use more testing.  Some of the tests I want to make though will require extending DTA features a little.  I also just launched a pair of ImageTests runs (on old and new tiles).

So we would appreciate you evaluating these code changes again.

pmconne:
> I also made a set of 11 new tests

Please supply them.

pmconne:
Document on `PlanarClipMaskMode` how each mode behaves with respect to the view's settings (as described in your bullet points above).

core/common/src/PlanarClipMask.ts
```diff
@@ -150,8 +150,8 @@ export interface PriorityPlanarClipMaskArgs extends BasicPlanarClipMaskArgs {
 export class PlanarClipMaskSettings {
   /** Specifies how the mask geometry is produced. */
   public readonly mode: PlanarClipMaskMode;
-  /** For any mode other than [[PlanarClipMaskMode.Priority]], the Ids of the [GeometricModel]($backend)s containing the geometry used to produce the mask.
-   * If `undefined`, the set of all models in the view's [ModelSelector]($backend) is used.
+  /** For any mode other than [[PlanarClipMaskMode.Priority]], the Ids of the [GeometricModel]($backend)s containing the geometry used to produce the mask,
+   * and if `undefined`, no mask geometry will be created.
```
pmconne, line 154: Did you change it to produce no mask if modelIds is undefined, or is that the way it already worked in V1?
If the latter, how did you decide that we want the current behavior vs the advertised behavior?

core/frontend/src/render/RenderTarget.ts
```diff
@@ -125,6 +125,7 @@ export abstract class RenderTarget implements IDisposable, RenderMemory.Consumer
   public abstract changeRenderPlan(plan: RenderPlan): void;
   public abstract drawFrame(sceneMilSecElapsed?: number): void;
   public overrideFeatureSymbology(_ovr: FeatureSymbology.Overrides): void { }
+  public getCurrentFeatureSymbologyOverrides(): FeatureSymbology.Overrides | undefined { return undefined; }
```
pmconne, line 128: You added this, and implemented it in subclasses, but I don't see you calling it anywhere.

core/common/src/PlanarClipMask.ts
DStradley: We modified the comment, and tried to clarify it, to match what the actual behavior already was.  The comment was unclear on what cases the undefined applied to - It is undefined in Priority Mode by definition (and does use all the models in the view in that case), but for the rest if undefined it produces no mask.  Since the code has been this way apparently forever, we thought it best to clarify/change the comment as someone may already depend on actual behavior.

DStradley:
> > I also made a set of 11 new tests
> 
> Please supply them.

I was referring to ImageTests tests, and these are already in the system and activated.

pmconne:
> I was referring to ImageTests tests, and these are already in the system and activated.

Can you provide .bim and ESV files, or reveal their names so I can download and test them?

DStradley:
> > I was referring to ImageTests tests, and these are already in the system and activated.
> 
> Can you provide .bim and ESV files, or reveal their names so I can download and test them?

The files you want are 04_Plant_Geo.bim and 04_Plant_Geo_ESV.json and are in Datasets/GeoLocated

core/frontend/src/render/RenderTarget.ts
DStradley: Yes, that was left over from an earlier attempt, now removed.

DStradley:
> Document on `PlanarClipMaskMode` how each mode behaves with respect to the view's settings (as described in your bullet points above).

Done

core/frontend/src/SpatialViewState.ts
```diff
@@ -256,6 +256,24 @@ export class SpatialViewState extends ViewState3d {
     this._treeRefs.setDeactivated(modelIds, deactivated, which);
   }
 
+  /** For getting the [TileTreeReference]s that are in the modelIds, for planar classification.
+   * @param modelIds modelIds for which to get the TileTreeReferences
+   * @param maskTreeRefs where to store the TileTreeReferences
+   * @internal
+   */
+  public setMaskRefs(modelIds: OrderedId64Iterable, maskTreeRefs: TileTreeReference[]): void {
```
pmconne, line 264: Is there are reason why this isn't a `getMaskRefs` returning an array instead of `setMaskRefs` that takes an array to populate? It's confusing, it's not actually "setting" anything.

core/frontend/src/PlanarClipMaskState.ts
```diff
@@ -53,14 +54,8 @@ export class PlanarClipMaskState {
...
       this._tileTreeRefs = new Array<TileTreeReference>();
-      if (this.settings.modelIds) {
-        for (const modelId of this.settings.modelIds) {
-          const model = view.iModel.models.getLoaded(modelId);
-          assert(model !== undefined);   // Models should be loaded by RealityModelTileTree
-          if (model?.asGeometricModel)
-            this._tileTreeRefs.push(createMaskTreeReference(view, model.asGeometricModel));
-        }
-      }
+      if (this.settings.modelIds && view.isSpatialView())
```
pmconne, line 57: Here and in `getPlanarClipMaskSymbologyOverrides`, change `view` to a `SpatialViewState`.

In PlanarClassifier.ts, Ray has a bug where he is obtaining the alleged `ViewState3d` to pass to this the `PlanarClipMaskState`: `collectGraphics` does this:
```
    const viewState = context.viewingSpace.view as ViewState3d;
    if (undefined === viewState)
      return;
```

It's never going to `return` here because `as ViewState3d` just tells the compiler it definitely **is** a `ViewState3d`. Change it to check `view.isSpatialView()`.

core/frontend/src/PlanarClipMaskState.ts
```diff
@@ -69,29 +64,51 @@ export class PlanarClipMaskState {
...
+
+    const overrides = new FeatureSymbology.Overrides();
+
+    if (overrideModels) {
+      // overrideModels is used for batched models.  For those, we need to create model overrides for visibility (using transparency).
+      const appOn = FeatureAppearance.fromTransparency(0.0);
+      const appOff = FeatureAppearance.fromTransparency(1.0);
+      // For Priority or Models mode, we need to start with the current overrides and modify them
+      if (PlanarClipMaskMode.Priority === this.settings.mode || PlanarClipMaskMode.Models === this.settings.mode) {
+        const curOverrides = new FeatureSymbology.Overrides(context.viewport);
```
pmconne, line 83: Why did you decide to get rid of the `RenderTarget.getCurrentSymbologyOverrides` method you previously had, and instead recreate the overrides from scratch here?

pmconne (approved):

core/frontend/src/SpatialViewState.ts
DStradley: Could change the name to `collectMaskRefs`?
Did this because the array is created separately, and its existence is a flag, and this func may not get called.

core/frontend/src/PlanarClipMaskState.ts
DStradley: Because we needed to modify them, and don't want to modify the target one.

core/frontend/src/SpatialViewState.ts
pmconne: Sure. It's a nit, `setMaskRefs` just sounds like it's changing something about the ViewState.

core/frontend/src/PlanarClipMaskState.ts
DStradley: Done

core/frontend/src/SpatialViewState.ts
DStradley: Renamed

extensions/frontend-tiles/src/BatchedSpatialTileTreeRefs.ts
```diff
@@ -139,6 +139,41 @@ class BatchedSpatialTileTreeReferences implements SpatialTileTreeReferences {
...
+            maskTreeRefs.push(ref);
+            break;
+          }
+        }
+      }
+    }
+  }
+
+  public getMaskModels(models: OrderedId64Iterable | undefined, useVisible: boolean): Map<Id64String, boolean> | undefined {
+    const maskModels = new Map<Id64String, boolean>();
```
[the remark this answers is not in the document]
DStradley: I was unaware of this.
Added.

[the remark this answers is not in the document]
pmconne: Do you have a test case?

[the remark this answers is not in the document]
pmconne: Did you consider just calling `this._excludedRefs.collectMaskRefs(modelIds, maskTreeRefs)`?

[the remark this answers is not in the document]
DStradley: No and no, but I will.

danieliborra:
Changes make sense to me (with my current limited knowledge of the iTwin viewer).
I think this could be a good opportunity to include some more documentation/explanations to this part of the code.
It would also be very useful that the comments/doubts that have appeared in this PR are included in the code as comments where appropriate.
@DStradley does this makes sense?

DStradley:
Discovered that the element, exclude elements and subcategory modes behave differently when subCategoryOrElementIds is undefined.  In that case, these modes behave like Models mode in that things that are off from category settings or element feature overrides in the view will not be in the mask for these models (rather than ignoring these view settings when there is a subCategoryOrElementIds).
Documented this while correcting new tiles to work the same.

DStradley:
> I think this could be a good opportunity to include some more documentation/explanations to this part of the code...

@danieliborra I attempted to do at least some of this as I went along.

danieliborra:
> > I think this could be a good opportunity to include some more documentation/explanations to this part of the code...
> 
> @danieliborra I attempted to do at least some of this as I went along.

Yes, I saw your comments and they are great, but if you could comment the rest of the functions you changed/created and the reasoning behind the changes (just a line or two), it would help us a lot if we have to re-visit this in the future. Thanks!

MarcNeely:
Added more documentation.

danieliborra (approved):
Looks very good to me, good work!
The documentation is very helpful in this case, thanks for adding it.

pmconne (approved):
You did a thorough job. Thanks.

DStradley:
@aruniverse This needs to be backported for Design Review, what version would that be?

aruniverse:

aruniverse:
> @aruniverse This needs to be backported for Design Review, what version would that be?

Since we just released 4.7, and theyre on 4.6.x, they will need to pick it up when they move up to 4.7.x

markschlosseratbentley:
> > @aruniverse This needs to be backported for Design Review, what version would that be?
> 
> Since we just released 4.7, and theyre on 4.6.x, they will need to pick it up when they move up to 4.7.x

@danieliborra fyi

danieliborra:
Thank you all!
@aruniverse I should ask Justin when they plan to adopt 4.7.x, right?

aruniverse:
> Thank you all! @aruniverse I should ask Justin when they plan to adopt 4.7.x, right?

@jjbeckman13

PR #6764, opened 2024-05-27
Increase restrictions on native app storage name

GytisCepk:
This should prevent storage files from being created outside storage cache directory.

Alternative solution would be to not allow any sort of path elements in the storage name. However, this would break apps which use storage names likes `my-app/common-settings` (if such apps exist).

### DIFF
--- a/core/backend/src/NativeAppStorage.ts
+++ b/core/backend/src/NativeAppStorage.ts
@@ -6,7 +6,7 @@
  * @module NativeApp
  */
 
-import { join } from "path";
+import { join, relative } from "node:path";
 import { DbResult, IModelStatus } from "@itwin/core-bentley";
 import { IModelError, StorageValue } from "@itwin/core-common";
 import { ECDb, ECDbOpenMode } from "./ECDb";
@@ -191,11 +191,16 @@ export class NativeAppStorage {
       IModelHost.onBeforeShutdown.addOnce(() => this.closeAll());
       this._init = true;
     }
+    if (name.includes("\x00"))
+      throw new Error("Storage name contains illegal characters");
     const fileName = name + this._ext;
     if (!IModelJsFs.existsSync(NativeHost.appSettingsCacheDir))
       IModelJsFs.recursiveMkDirSync(NativeHost.appSettingsCacheDir);
 
     const storageFile = join(NativeHost.appSettingsCacheDir, fileName);
+    if (relative(NativeHost.appSettingsCacheDir, storageFile).startsWith(".."))
+      throw new Error("Storage name should not be a path");
+
     try {
       return this.find(fileName); // see if it's already open
     } catch (err) {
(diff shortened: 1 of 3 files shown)


core/backend/src/NativeAppStorage.ts
```diff
@@ -191,11 +191,16 @@ export class NativeAppStorage {
       IModelHost.onBeforeShutdown.addOnce(() => this.closeAll());
       this._init = true;
     }
+    if (name.includes("\x00"))
```
aruniverse, line 194: any other special chars we need to handle?

pmconne: It's a valid Unicode character. `console.log("ab\x00\x00cd")` prints "abcd".

kabentley: that's crazy. If it's valid, why is this test necessary?

GytisCepk: > If it's valid, why is this test necessary?

We use `name` to create path which will be passed to the native side to create a storage file. If we allow `name` to contain `null`, path will be cut short because of how strings are terminated in C++.

e.g.: if `name` has a value `"ab\x00cd"`, storage file will be created as `ab` instead of `abcd.settings-db`.

GytisCepk: > any other special chars we need to handle?

None for us to handle. There are a few which are not permitted in file names, but most of them are OS specific and are handled by the OS itself (names are rejected or special character are omitted).

kabentley: This is silly. We don't need to protect programmers from themselves. Why would anyone add a 0 in a string.

GytisCepk: This was raised as a security concern.

`name` comes directly from the frontend, so the idea is to prevent any malicious party, who may manage to execute their script in the frontend, from creating files with any file extension they want on user's machine. [Workitem](https://dev.azure.com/bentleycs/ProductAdvancement/_workitems/edit/1434865) explains how this can be potentially exploited.

kabentley: It can only come from the frontend of a _trusted_ app. We have 100's of places we accept strings from front ends that are passed to native code. We can't add this test everywhere, which is why I raise this concern.

cjj884: If there's a belief that we 100% trust frontend code then why bother disabling node integration at all in the frontend and just develop everything there? IMO it's not a question of if there will be a XSS in a frontend, but when. The only thing that stands between XSS being RCE on a user's desktop is the process isolation afforded by separating frontend from backend and the restrictions on IPC that go with it. If this is considered silly then it's effectively saying defense in depth is a silly concept.

aruniverse (approved):

core/backend/src/NativeAppStorage.ts
thomas-chauchefoin-bentley-systems: I agree with @cjj884 on this. I still want to add that I'm not fully satisfied with the NULL-byte check happening in itwinjs-core, in my opinion it should rather happen in imodel-native, at the interface between the NodeJS N-API and system functions. Should we create another security concern?

PR #6769, opened 2024-05-29
Fix text annotation wrapping+justification issues

### LINKED ISSUE #6768: TextAnnotation Width + Anchor point bugs
**Describe the bug**

I've noticed the following issues/possible issues with the [`TextAnnotation`](https://github.com/iTwin/itwinjs-core/blob/1c2c71ee5c0ddd43c5e686b7cebf098830c1d2be/core/common/src/annotation/TextAnnotation.ts) class:

1. **Document width + anchor point**: I noticed that changing the width via `dta text width` does cap the width and word wraps as expected. However, this doesn't seem to be registered by the anchor point or the computed text range. I would expect the anchor point to adjust to the text block width. 
  
  image

2. **Document width + justification**: for this one, if I have multiple lines and try to justify them to the right or to the center, they align themselves based off the previous line. While this makes my test strings super poetic, this is a bug, right? Also the anchor point is thrown way off to the side, which is intriguing to me.

  image

3. **Anchor point + placement origin**: when changing the anchor point, I expect the anchor point to remain fixed at the element/graphics's origin, and for the text to move relative to the anchor point. Let me know if I'm misunderstanding the use of the anchor point. 

  image

4. **The debug tools**: Also, this one's not a super big deal, but I noticed when rotating the text element, the debug graphics don't reflect that rotation. That may be intentional, but just thought I would point that out in case it wasn't. 

  drawing

**To Reproduce**

I'm running display test app. Last pull, Friday 5/24/2024.

For all of these, I started with these base key-ins:
```txt
dta text init "0x20000000087"
dta text font arial
dta text center
dta text width 20
dta text text "It was the best of times. It was the worst of times."
```

0. Set `debugAnchorPointAndRange` to `true`: https://github.com/iTwin/itwinjs-core/blob/8434bf71c4c37ba43a3ed1ef9c2c65308e4c92fa/core/backend/src/TextAnnotationGeometry.ts#L201

1. **Document width + anchor point**: 
    Additional Key-ins: _no extra, I just used the above key-ins_

2. **Document width + justification**: 
    Additional Key-ins:
    ```txt
    dta text justify center
    ```

3. **Anchor point + placement origin**: 
    Additional Key-ins:
    ```txt
    dta text anchor right
    dta text anchor left
    ```

4. **The debug tools**: 
    Additional Key-ins:
    ```txt
    dta text rotation 45
    ```

**Expected behavior**

Summed up in the description.

**Screenshots**

Summed up in the description.

- OS: Windows
- Browser: Electron
- iTwin.js Version: ^4.6.0 ish, I'm using display test app, last pull, Friday 5/24/2024.

**Additional context**

### LINKED ISSUE #6777: TextBlock fraction alignment should default to 'middle'
Currently, TextBlock does not support the concept of fraction alignment from MicroStation.  Presumably it was removed for simplicity which is fine with me.  The issue is that presently we only support bottom alignment which I'm pretty sure is much less desirable compared to middle alignment.

Support middle fraction alignment only.  If there is some use case for bottom, then at least provide a way to use middle.

**Additional context**
This is needed for dimensioning where bottom will look terrible.

image

I used the following keyins to create the first text in the image.

dta text init 0x200000000c8
dta text font arial
dta text text "first line"
dta text paragraph
dta text text "abc "
dta text fraction 1 2
dta text text " def"
dta text paragraph
dta text text "third line"

pmconne:
Fixes #6777: Stacked fractions are taller than ordinary runs. Compute the height of the line, then center each run inside the line.

Fixes #6768: Simplified line-wrapping logic. Codified justification behavior. Basically, compute a bounding box that is at least as wide as `TextBlock.width`. Justify relative to that box. Expand that box if an un-wrappable line exceeds the specified width. When right-justified, it expands to the left, and vice-versa; center-justified expands in both directions. This permits overflow while keeping the anchor point stable.

`ProduceTextAnnotationGeometryArgs` now has a `debugAnchorPointAndRange` option so you can debug without having to edit + reocmpile core-backend. `dta text debug` toggles this option.


bbastings (approved):

core/backend/src/TextAnnotationLayout.ts
```diff
@@ -41,18 +41,12 @@ export type FindFontId = (name: string) => FontId;
 /** @internal */
 export type FindTextStyle = (name: string) => TextStyleSettings;
 
-/** Arguments supplied to [[computeTextBlockExtents]].
- * @beta
- */
-export interface ComputeTextBlockExtentsArgs {
+/** @internal */
```
aruniverse, line 44: do we need to call this out in nextver.md?

aruniverse (approved):

ben-polinsky (approved):

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -146,11 +146,11 @@ export class TextAnnotation {
    * The anchor point is computed as specified by this annotation's [[anchor]] setting. For example, if the text block is anchored
    * at the bottom left, then the transform will be relative to the bottom-left corner of `textBlockExtents`.
    * The text block will be rotated around the fixed anchor point according to [[orientation]], then the anchor point will be translated by [[offset]].
-   * @param textBlockDimensions The width and height of the bounding box containing the text block. You can compute this using [computeTextBlockExtents]($backend).
+   * @param boundingBox A box fully containing the [[textBlock]]. You can compute this using [computeTextBlockBoundingBox]($backend).
```
ben-polinsky, line 149: Does `computeTextBlockBoundingBox` exist? Here and below in `computeTransform`, the jsdoc comment links are not being resolved.

core/backend/src/TextAnnotationLayout.ts
pmconne: It's a beta API that only existed for a few weeks and nobody ever used.

core/common/src/annotation/TextAnnotation.ts
pmconne: No, it doesn't. I thought I was going to have to replace `computeTextBlockExtents` with it but nobody ever used that function.

pmconne: Removed. We can add a link to the new `computeTextBlockLayoutResult` function being added in #6820 after that merges. The existing `layoutTextBlock` function is internal.

PR #6775, opened 2024-05-30
Refactor Mesh Export Service Access in Frontend-Tiles

andremig-bentley:
Refactor FrontendTiles.ts to decouple the Mesh Export Service access points in frontend-tiles from batched tiles. The goal is to be able to use the Mesh Export Service to obtain tiles other than only batched tiles in the future. To do this, `GraphicsService` and `GraphicsServiceProvider` have been created. The `GraphicsService` will work as a connector - outside sources such as frontend tiles will request tileset urls from the `GraphicsService`. The `GraphicsService` will then obtain the url from the `GraphicsServiceProvider` and then send it back to the outside source.

Fixes https://github.com/iTwin/itwin-graphics-backlog/issues/249


extensions/frontend-tiles/src/MeshExportServiceProvider.ts
```diff
@@ -0,0 +1,190 @@
+/*---------------------------------------------------------------------------------------------
```
andremig-bentley, line 1: All functionality for the mesh export service itself (not dealing with batched tiles) has been moved to a separate file, and the class `MeshExportServiceProvider` has been created to house those functions. `obtainMeshExportTilesetUrl` is the main access point to the MES, so it has been left public, while the other functions have been made private.

extensions/frontend-tiles/src/FrontendTiles.ts
```diff
@@ -235,6 +56,7 @@ export const frontendTilesOptions = {
  * @beta
  */
 export function initializeFrontendTiles(options: FrontendTilesOptions): void {
+  const provider = new MeshExportServiceProvider();
```
andremig-bentley, line 59: I'm not convinced that creating an instance of a provider here is the best way to accomplish this, but as of now, I believe it's necessary so that we can call `obtainMeshExportTilesetUrl`. Open to any suggestions.

extensions/frontend-tiles/src/test/FrontendTiles.test.ts
```diff
@@ -2,222 +2,20 @@
...
-    await expectUrl("http://tiles.com/a/tileset.json", { id: "imdl", changesetId: "bbbbbb" });
-    await expectUrl("http://tiles.com/a/tileset.json", { id: "imdl", changesetId: "bbbbbb", exact: false });
-  });
-
-  it("returns undefined if no export matches the changeset Id and caller requires an exact changeset match", async () => {
-    await expectUrl(undefined, { id: "imdl", changesetId: "bbbbbb", exact: true });
+import { expect } from "chai";
+import { frontendTilesOptions, FrontendTilesOptions, initializeFrontendTiles } from "../FrontendTiles";
+
+describe("FrontendTiles", () => {
```
andremig-bentley, line 8: This is not at all an exhaustive test, but this file has now been changed to house test for `FrontendTilesOptions`. Better tests are WIP

extensions/frontend-tiles/src/MeshExportService.ts
```diff
@@ -0,0 +1,186 @@
...
+import { IModelApp, IModelConnection } from "@itwin/core-frontend";
+import { loggerCategory } from "./LoggerCategory";
+
+/** A function that can provide the base URL where a tileset representing all of the spatial models in a given iModel are stored.
+ * The tileset is expected to reside at "baseUrl/tileset.json" and to have been produced by the [mesh export service](https://developer.bentley.com/apis/mesh-export/).
+ * If no such tileset exists for the given iModel, return `undefined`.
+ * @see [[FrontendTilesOptions.computeSpatialTilesetBaseUrl]].
+ * @beta
+ */
+export type ComputeSpatialTilesetBaseUrl = (iModel: IModelConnection) => Promise<URL | undefined>;
```
andremig-bentley, line 16: After looking in to how the pineapple viewer uses the mesh export service, I decided to remove the `MeshExportServiceProvider` class which was wrapping all of this functionality. The viewer accesses `obtainMeshExportTilesetUrl` directly, so wrapping it in a class would be a breaking change for them.

 Now, the functionality for the mesh export service itself has been moved here, into `MeshExportService.ts` while tests for the service have been moved to `MeshExportService.test.ts`. FrontendTiles functionality has been left in `FrontendTiles.ts` and tests of `initializeFrontendTiles` have been added to `FrontendTiles.test.ts`.

pmconne:
Did anything functionally change, or did you simply move existing code to different files?

pmconne:
I question the further proliferation of the term "mesh export" given we want to rename the service and we wish to emphasize streaming of data, not export of it.

common/changes/@itwin/frontend-tiles/andremig-mxprovider_2024-05-30-14-29.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/frontend-tiles",
+      "comment": "Refactor FrontendTiles",
```
pmconne, line 5:
```suggestion
      "comment": "",
```

Changelogs record changes of interest to users. No user cares that we moved some code around internally. (Also, committing this change will trigger the various build pipelines that you would otherwise need to manually, individually trigger)).

andremig-bentley:
> I question the further proliferation of the term "mesh export" given we want to rename the service and we wish to emphasize streaming of data, not export of it.

I'm not at all tied to the name, I didn't know we were looking to change it. Would something like "Mesh Stream" or "Tile Stream" be any better? @danieliborra any thoughts?

danieliborra:
What about 'GraphicsServiceProvider'?
I would prefer 'iTwinGraphicsServiceProvider' but looks like the "'iTwin" may be redundant here.

andremig-bentley:
Major changes with latest commit:

- Create `GraphicsService` and `GraphicsServiceProvider`. The `GraphicsServiceProvider` contains the Mesh Export Service functionality which previously was housed in `FrontendTiles.ts`, meaning it will query the service and obtain tileset urls. The `GraphicsService` acts as a connector between the `GraphicsServiceProvider` and outside callers. Outside callers will request tileset urls from the `GraphicsService` which will then obtain the urls from the `GraphicsServiceProvider`, and then return the associated graphics data to the outside caller.
- Removed all iModel and Mesh Export related names/data from the `GraphicsServiceProvider` functionality (previously the Mesh Export Service) which was possible to remove. The `GraphicsServiceProvider` is meant to deal with type-agnostic graphics data, however, to avoid breaking existing API's some iModel/Mesh Export specifics are still present (e.g. the `ServiceJsonResponse` interface still includes `iModelId` and `changesetId` within the request).
- Added Tests for the `GraphicsServiceProvider` and the `GraphicsService`

common/api/frontend-tiles.api.md
```diff
@@ -28,6 +28,24 @@ export const frontendTilesOptions: {
     useIndexedDBCache: boolean;
 };
 
+// @beta
+export interface GraphicsDataSource {
+    // (undocumented)
```
pmconne, line 33: APIs must be documented.

andremig-bentley:
Closing due to PR scope changes

PR #6776, opened 2024-05-30
close the briefcase db before throwing the error 

jiaruiz717:
Close the briefcase db before throwing the error, if we fail to open the codeService 

fix iTwin/itwinjs-backlog#1122


kabentley:
I'm not sure this is right. I think what should happen if there is a CodeServiceDb but it is not available, is that we create a CodeService that always throws for verifyCode if the code is non-null. It should be legal to open a briefcase and make changes that don't affect codes.

jiaruiz717:
> I'm not sure this is right. I think what should happen if there is a CodeServiceDb but it is not available, is that we create a CodeService that always throws for verifyCode if the code is non-null. It should be legal to open a briefcase and make changes that don't affect codes.

I made some changes correspondingly. I'll also add a test, but before that I have a question, what is an example of a change to a briefcase that doesn't affect codes?

pmconne:
> what is an example of a change to a briefcase that doesn't affect codes?

Deleting an element. Inserting or updating an element that has a null code. Saving a change made via `IModelDb.save/deleteFileProperty`. Updating or deleting a model.

i.e., any operation that either doesn't need to invoke `verifyCode` or for which `verifyCode` is a no-op.

nick4598:
Is there any concern that a user believes they've successfully instantiated a code service as opening the briefcase went smoothly, only for them to later find out we've setup a fake one for them? Meaning they wasted sometime creating an element(?) and is soon about to fail?

Maybe logging a warning that this has occurred is enough..?

kabentley:
> Maybe logging a warning that this has occurred ( a fake codeservice ) is enough..?

Users won't even know what a CodeService is, but logging a warning wouldn't hurt. They will be stopped from performing any operation that requires code validation. I can imagine many scenarios where an entire session will work fine - it depends on what the operations they do. The message about why they can't add/change/delete something should make it clear what's wrong and how to resolve the problem.

core/backend/src/IModelDb.ts
```diff
@@ -2768,8 +2768,22 @@ export class BriefcaseDb extends IModelDb {
         briefcaseDb._codeService = await CodeService.createForIModel(briefcaseDb);
         this.onCodeServiceCreated.raiseEvent(briefcaseDb);
       } catch (e: any) {
-        if ((e as CodeService.Error).errorId !== "NoCodeIndex") // no code index means iModel isn't enforcing codes.
-          throw e;
+        if ((e as CodeService.Error).errorId !== "NoCodeIndex") { // no code index means iModel isn't enforcing codes.
```
kabentley, line 2771: I don't know what the exception string should be. It should probably explain that the codeservice is not available and append the string from `e`. But I'm wondering where this comes up. Didn't we make the CodeService implementation not throw for authorization errors? @johnnyd710?

nick4598: @swwilson-bsi was the one who originally reported this so maybe he would have some ideas about how this occurred.

johnnyd710: I made it throw the "user is not authorized to access CodeService" error (thrown when a user does not have all necessary RBAC permissions) when the code service is actually used, not on initialization. See https://github.com/iTwin/code-service/pull/9

jiaruiz717:
> Deleting an element. Inserting or updating an element that has a null code. Saving a change made via `IModelDb.save/deleteFileProperty`. Updating or deleting a model.

I tried updateFileProperty, it indeed does not affect code, does not invoke verifyCode and worked as expected. Just a question about inserting/updating element for my understanding, onUpdate/onInsert call verifyCode no matter what the code is, based on the function implementation:
```javascript
protected static onUpdate(arg: OnElementPropsArg): void {
    const { iModel, props } = arg;
    iModel.channels.verifyChannel(props.model);
    iModel.locks.checkExclusiveLock(props.id!, "element", "update"); // eslint-disable-line @typescript-eslint/no-non-null-assertion
    iModel.codeService?.verifyCode(arg);
  }
```
Do you mean verifyCode is a no-op if the code is null?

pmconne:
> Do you mean verifyCode is a no-op if the code is null?

Yes.

jiaruiz717:
> > Do you mean verifyCode is a no-op if the code is null?
> 
> Yes.

I want to make sure that verifyCode in our new 'fake' codeService will not throw the error if the code is null. How do we define a null code? Is it Code.createEmpty() or undefined or null or something else?

pmconne:
> How do we define a null code? Is it Code.createEmpty()

Yes.

core/backend/src/IModelDb.ts
```diff
@@ -2772,7 +2772,7 @@ export class BriefcaseDb extends IModelDb {
           Logger.logWarning(loggerCategory, `The CodeService is not available for this briefcase: errorId: ${(e as CodeService.Error).errorId}, errorMessage; ${e.message}. Proceeding with BriefcaseDb.open(), but all operations involving codes will fail.`);
           briefcaseDb._codeService = {
             verifyCode: (props: CodeService.ElementCodeProps) => {
-              if (props.props.code !== null) {
+              if (props.props.code !== Code.createEmpty()) {
```
pmconne, line 2775:
```suggestion
              if (!Code.isEmpty(props.props.code))) {
```

jiaruiz717: Resolved in [commit ff9178df8b77156c34ed76e2b113a7d39811ef28](https://github.com/iTwin/itwinjs-core/pull/6776/commits/ff9178df8b77156c34ed76e2b113a7d39811ef28)

core/backend/src/IModelDb.ts
```diff
@@ -2768,8 +2768,23 @@ export class BriefcaseDb extends IModelDb {
...
         this.onCodeServiceCreated.raiseEvent(briefcaseDb);
       } catch (e: any) {
-        if ((e as CodeService.Error).errorId !== "NoCodeIndex") // no code index means iModel isn't enforcing codes.
-          throw e;
+        if ((e as CodeService.Error).errorId !== "NoCodeIndex") { // no code index means iModel isn't enforcing codes.
+          Logger.logWarning(loggerCategory, `The CodeService is not available for this briefcase: errorId: ${(e as CodeService.Error).errorId}, errorMessage; ${e.message}. Proceeding with BriefcaseDb.open(), but all operations involving codes will fail.`);
+          briefcaseDb._codeService = {
+            verifyCode: (props: CodeService.ElementCodeProps) => {
+              if (!Code.isEmpty(props.props.code)) {
+                e.message = `The CodeService is not available for this briefcase: errorId: ${(e as CodeService.Error).errorId}, errorMessage; ${e.message}. Proceeding with BriefcaseDb.open(), but all operations involving codes will fail.`
```
nick4598, line 2776:
```suggestion
                e.message = `The CodeService is not available for this briefcase: errorId: ${(e as CodeService.Error).errorId}, errorMessage; ${e.message}.`
```

jiaruiz717: Resolved in [commit 0b6ec57e366e631524867756bea8b4c45525fe97](https://github.com/iTwin/itwinjs-core/pull/6776/commits/0b6ec57e366e631524867756bea8b4c45525fe97)

core/backend/src/test/standalone/IModelWrite.test.ts
```diff
@@ -329,6 +331,39 @@ describe("IModelWriteTest", () => {
     rwIModel.close();
   });
 
+  it.only("should set a fake verifyCode for codeService that throws error for operations that affect code, if failed to open codeService ", async () => {
```
nick4598, line 334:
```suggestion
  it("should set a fake verifyCode for codeService that throws error for operations that affect code, if failed to open codeService ", async () => {
```

core/backend/src/test/standalone/IModelWrite.test.ts
```diff
@@ -329,6 +331,39 @@ describe("IModelWriteTest", () => {
...
+      if (stmt.step() === DbResult.BE_SQLITE_ROW) {
+        firstElement = stmt.getRow();
+      }
+    });
+    // make change to the briefcaseDb that does not affect code, e.g., save file property
+    // expect no error from verifyCode
+    expect(() => briefcaseDb.saveFileProperty({ name: "codeServiceProp", namespace: "codeService", id: 1, subId: 1 }, "codeService test")).to.not.throw();
+    // make change to the briefcaseDb that affects code that will invoke verifyCode, e.g., update an element with a non-null code
+    // expect error from verifyCode
+    const newProps = { id: firstElement.id, code: {...Code.createEmpty(), value:firstElement.codeValue}, classFullName: undefined, model: undefined };
```
nick4598, line 359: can you also add a case that we don't throw an error if the Code is an emptyCode somewhere in this test?

jiaruiz717: Added in [commit eec2512af53b9c5b79a8f7fcb99749b1fc7e541a](https://github.com/iTwin/itwinjs-core/pull/6776/commits/eec2512af53b9c5b79a8f7fcb99749b1fc7e541a)

core/backend/src/test/standalone/IModelWrite.test.ts
```diff
@@ -329,6 +331,43 @@ describe("IModelWriteTest", () => {
...
 
+  it("should set a fake verifyCode for codeService that throws error for operations that affect code, if failed to open codeService ", async () => {
+    const iModelProps = {
+      iModelName: "codeServiceTest",
+      iTwinId,
+    };
+    const iModelId = await HubMock.createNewIModel(iModelProps);
+    const briefcaseProps = await BriefcaseManager.downloadBriefcase({ accessToken: "codeServiceTest", iTwinId, iModelId });
+    const originalCreateForIModel = CodeService.createForIModel;
+    // can be any errors except 'NoCodeIndex'
```
nick4598, line 342: Can you assert that we propagate the error if the error is 'NoCodeIndex' and don't set up a fake codeService?

nick4598 (approved):

core/backend/src/test/standalone/IModelWrite.test.ts
jiaruiz717: Added in [commit f818d0130ca3f0cbfa0125ba43c4d290563ebca9](https://github.com/iTwin/itwinjs-core/pull/6776/commits/f818d0130ca3f0cbfa0125ba43c4d290563ebca9)

core/backend/src/test/standalone/IModelWrite.test.ts
```diff
@@ -329,6 +330,55 @@ describe("IModelWriteTest", () => {
...
+    // expect error from verifyCode
+    let newProps = { id: firstNonRootElement.id, code: {...Code.createEmpty(), value:firstNonRootElement.codeValue}, classFullName: undefined, model: undefined };
+    await briefcaseDb.locks.acquireLocks({exclusive: firstNonRootElement.id});
+    expect(() => briefcaseDb.elements.updateElement(newProps)).to.throw(CodeService.Error);
+    // make change to the briefcaseDb that will invoke verifyCode with a null(empty) code, e.g., update an element with a null(empty) code
+    // expect no error from verifyCode
+    newProps = { id: firstNonRootElement.id, code: Code.createEmpty(), classFullName: undefined, model: undefined };
+    expect(() => briefcaseDb.elements.updateElement(newProps)).to.not.throw();
+    briefcaseDb.close();
+    // throw "NoCodeIndex"
```
nick4598, line 366:
```suggestion
    // throw "NoCodeIndex", this error should get ignored because it means the iModel isn't enforcing codes. 
```

core/backend/src/test/standalone/IModelWrite.test.ts
```diff
@@ -329,6 +330,55 @@ describe("IModelWriteTest", () => {
...
+    CodeService.createForIModel = async () => {
+      throw new CodeService.Error("NoCodeIndex", 0x10000 + 1, " ");
+    };
+    const briefcaseDb2 = await BriefcaseDb.open({ fileName: briefcaseProps.fileName});
+    briefcaseDb2.channels.addAllowedChannel(ChannelControl.sharedChannelName);
+    await briefcaseDb2.locks.acquireLocks({exclusive: firstNonRootElement.id});
+    newProps = { id: firstNonRootElement.id, code: {...Code.createEmpty(), value:firstNonRootElement.codeValue}, classFullName: undefined, model: undefined };
+    // make change to the briefcaseDb that affects code that will invoke verifyCode, e.g., update an element with a non-null code
+    // expect no error from verifyCode
+    expect(() => briefcaseDb2.elements.updateElement(newProps)).to.not.throw();
```
nick4598, line 376: Can you test that a empty code works here too

nick4598: To me it looks like you've just called updateelemnt with an emtpy code twice

jiaruiz717:
I encountered a test case that is possible in real use cases:

1. update the code of the root subject to be an empty code using `updateElement(newProps)`
2. close the briefcase db
3. open that breifcase db again

There was an error in step3 from native code. After investigation, it is because GetIModelProps is called when opening the db. `rootSubject->GetCode().GetValueUtf8CP()` in line 1152 returns a `nullptr`, hence that error was thrown in line 1042.
image
image

I bypassed this issue by updating a non-root subject instead of the root subject. But there is a possibility in real use cases to meet the same situation. 

**Question:**
Do we need to modify `updateElement()` to throw an error when users try to assign an empty code to the root subject?

nick4598:
To me, it seems better to error out at the time of updating the element as opposed to when opening the database after this update has occurred. Feels weird to modify updateElement in TS since all it does is call the native implementation of updateElement. So maybe its better to fix it on the native side? 

Or maybe we should detect that we have an empty code and set the subject[json_name] to empty string in that case? I don't know if its necessarily wrong for the rootsubject's code to be empty or not. I think that the rootsubject's codeValue is typically the name of the iModel itself, but not sure if thats just a convention we've setup that we want to enforce by throwing an error or we can allow empty code and just set the name to an empty string.

core/backend/src/IModelDb.ts
jiaruiz717: @kabentley Hi, are we okay to resolve this conversation?

kabentley: Yes, you can resolve this

PR #6780, opened 2024-05-31
Prohibit `Object` methods being called through IPC

GytisCepk:
## Motivation

IPC expects only methods defined in a custom IPC interface to be called, but it does not account for all other methods defined in an IPC handler. For this reason, it is possible to execute methods that exist in the IPC handler but are not intended to be called from the frontend. This includes all methods inherited from the `Object` and any methods defined as `private` in TypeScript code.

This could be exploited by malicious party to weaken security ([work item for more details](https://dev.azure.com/bentleycs/ProductAdvancement/_workitems/edit/1434860)).

## Changes

Changes in this PR prevent methods inherited from `Object` class (e.g.: `toString`, `__defineSetter__`) from being called through IPC. 

It would be great to also prohibit `private` methods in custom IPC handlers from being callable, but it's not possible, since, if they are defined `private` in TypeScript, this information is not available during runtime. Added a recommendation to the documentation to either remove private methods from IPC handlers or define them as [JavaScript private properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties).


GytisCepk:
> Your PR description should specify why you want to do this.

Updated description.

aruniverse (approved):

pmconne (changes requested):
I want to discuss with @wgoehrig first.

pmconne (approved):

Issue #6742, opened 2024-05-22
Make it harder to accidentally use internal APIs


calebmshafer:
With our current release tag setup we use the `@internal` tag to denote something that we do not want consumers of the library to use. However, those APIs are still part of the barrel files and exported from our packages, either as full types or a method/property on a public type. This leads to consumers accidentally using these APIs because they do not know any difference.

We need to come up with a strategy that reduces, or optimally eliminates, the ability for consumers to run into this scenario.

calebmshafer:
As a result of the aforementioned "pow-wow", we are going to take this in steps to validate what a good strategy will be. Below is the summary of the action items of the call.

- [ ] Move as many `internal` APIs our of the main barrel exports
  - The plan to move them out will be to create a separate folder in each package as a peer of `src`, named `src-internal` to make it explicit
  - We need to understand the scope of this change to determine whether we can do it all, or at least, a subset of it within 4.x or it needs to wait until 5.0.
  - [ ] @pmconne is going to get the list of what APIs can be moved out of core-common as a template. Then we break out from there so that more people can tackle each and gather the data
  - [ ] Scope of all packages (after @pmconne's work)
- [ ] Investigate strip internals (@wgoehrig)
  - This will require 5.0 release
- Do not try and tackle `alpha`/`beta` for now, focus on `internal`

calebmshafer:
@GerardasB @grigasp @saskliutas FYI - I know the AppUI and Presentation side is looking at some similar so wanted to share what was discussed.

Presentation pkg related issue - https://github.com/iTwin/presentation/issues/545

pmconne:
@calebmshafer @aruniverse @wgoehrig @ben-polinsky One perennial pain point with our API support policies is that we can never add new required properties to a `@public` interface, just in case somebody outside itwinjs-core is creating their own implementation of it. On the other hand, exposing interfaces instead of implementations (classes) gives us more flexibility to make improvements to the implementation without breaking the API. We really have two kinds of interfaces:
- Those that define the shapes of objects provided by itwinjs-core APIs. Consumers of the library are expected to obtain implementations of these objects from itwinjs-core, not define their own implementations. Maybe call these "output interfaces".
- Those that define the shapes of objects that are passed in to itwinjs-core APIs. Consumers are expected to define their own implementations.

[...]

PR #6783, opened 2024-06-01
Enforce package-internal API visibility

pmconne:
See internal/Internal.ts for the proposed policy. The policy is applied to `IModelDb._nativeDb` as an example.

Part of #6742


pmconne:
Superseded by #6828.

PR #6784, opened 2024-06-03
Approximate elliptical arc with circular arc chain

dassaf4:
New functionality for iTwin consumers who don't support ellipse operations, but who can handle circular arcs. Of course, there is no exact approximation, but with enough arcs in the chain, you can get sufficiently close for many purposes...

Implements https://github.com/iTwin/itwinjs-backlog/issues/1107


core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,131 @@
...
+   * * The instance sweep is ignored.
+   * * Arc is assumed to be non-circular and have perpendicular axes of positive length.
+   * * This is a scaled inverse of [[Arc3d.fractionToCurvature]] restricted to fractions in [0,1/4].
+   * @return radian angle in [0,pi/2] or undefined if the ellipse is invalid, or does not attain the given curvature.
+  */
+  private curvatureToAngle(curvature: number): number | undefined {
+    if (!this.isValidArc)
+      return undefined;
+    if (!this.curvatureRange.containsX(curvature))
+      return undefined; // ellipse does not attain this curvature
```
saeeedtorabi, line 92: I think we should document the math details of lines 93-97.

From Eberly PDF we have `x = a*sqrt( |lambda - a^2| / |b2 - a^2| )` where `lambda = (a*b / K )^(3/2)` so `cos(theta) = x/a = sqrt( |lambda - a^2| / |b2 - a^2| )` which is calculated here. We can add this as a comment to the function body but still derivation of the original formula we used for `x` is not clear. Therefore, I think we should add math details to an internal md file. I can do that.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,131 @@
...
+    const cosTheta = Math.sqrt(Math.abs(numerator / denominator));
+    return Math.acos(cosTheta);
+  }
+
+  /**
+   * Sample the instance arc by interpolating curvature, and return the fractions of the sample points.
+   * @param numPointsInQuadrant number of samples in each ellipse quadrant.
+   * @param structuredOutput false to return a flat array of fractions (default). If true, structured objects are returned:
+   * * Fractions are sorted and assembled by ellipse quadrant into a [[QuadrantFractions]] object.
+   * * If the arc sweep spans adjacent quadrants, `QuadrantFractions`representing those quadrants share the fraction at the quadrant boundary.
```
saeeedtorabi, line 105: This line was a bit vague for me. I think this makes it more clear:
```suggestion
   * * If the arc sweep spans adjacent quadrants, `QuadrantFractions` that represent those quadrants would share the fraction at the quadrant boundary.
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,131 @@
...
+   * * `QuadrantFractions` objects are sorted in increasing fraction order.
+   * * Note that if the arc starts and ends in the same ellipse quadrant, two `QuadrantFractions` objects can be returned.
```
saeeedtorabi, line 107:
```suggestion
   * * Note that if the arc starts and ends in the same ellipse quadrant, two `QuadrantFractions` objects can be returned. Therefore, maximum of 5 `QuadrantFractions` objects can be returned for an arc.
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,131 @@
...
+    const numerator = Math.cbrt(scaledNormalLengthSquared) - this._axx;
+    const denominator = this._ayy - this._axx;
+    const cosTheta = Math.sqrt(Math.abs(numerator / denominator));
+    return Math.acos(cosTheta);
+  }
+
+  /**
+   * Sample the instance arc by interpolating curvature, and return the fractions of the sample points.
+   * @param numPointsInQuadrant number of samples in each ellipse quadrant.
+   * @param structuredOutput false to return a flat array of fractions (default). If true, structured objects are returned:
```
saeeedtorabi, line 103: what does `a flat array of fractions` mean?!  Do you mean naive distribution of fractions?

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,131 @@
...
+      return undefined; // ellipse does not attain this curvature
+    const scaledNormalLengthSquared = this._axx * this._ayy / curvature * curvature;
+    const numerator = Math.cbrt(scaledNormalLengthSquared) - this._axx;
+    const denominator = this._ayy - this._axx;
+    const cosTheta = Math.sqrt(Math.abs(numerator / denominator));
+    return Math.acos(cosTheta);
+  }
+
+  /**
+   * Sample the instance arc by interpolating curvature, and return the fractions of the sample points.
```
saeeedtorabi, line 101:
```suggestion
   * Sample the instance arc by interpolating curvature, and return the fractions of the sample points. Default is 4.
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,131 @@
...
+    const cMin = this.curvatureRange.low;
+    const cMax = this.curvatureRange.high;
+    const tDelta = 1.0 / (numPointsInQuadrant - 1);
+    for (let i = 1; i < numPointsInQuadrant; ++i) {
+      const j = f(i * tDelta);
+      const curvature = (1 - j) * cMin + j * cMax;
+      for (const fraction of curvatureToFractions(ellipticalArc, curvature))
+        fractions.add(fraction);
+    }
+    for (const fraction of [...fractions].sort())
```
saeeedtorabi, line 127: As discussed, we should also drop fractions that are too close together because they are unnecessary and reduce the performance. I was able to do that by doing below rather than a simple `[...fractions].sort()`

``` 
    let tolerance = 0.01;
    const filteredFractions = [...fractions].sort((a, b) => a - b).filter((val, i, arr) => {
      return i === 0 || Math.abs(val - arr[i - 1]) >= tolerance;
    });
```

You may want to chaneg the tolerance but I think 0.01 (which is 1%) is good.

core/geometry/src/test/curve/Arc3d.test.ts
```diff
@@ -552,4 +552,166 @@ describe("Arc3d", () => {
     expect(ck.getNumErrors()).equals(0);
   });
 
+  it.only("ParameterizationSpeed", () => {
```
saeeedtorabi, line 555: The test does not seem to be ready for review so I will review it later.

saeeedtorabi (changes requested):

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,440 @@
...
+  public create(method: EllipticalArcSampleMethod = EllipticalArcSampleMethod.UniformParameter, structuredOutput: boolean = false, numSamplesInQuadrant: number = 4, maxError: number = 0.01, remapFunction?: FractionMapper) {
+    if (numSamplesInQuadrant < 2)
+      numSamplesInQuadrant = 2;
+    return new EllipticalArcApproximationOptions(method, structuredOutput, numSamplesInQuadrant, maxError, remapFunction);
+  }
+  /** Method used to sample the elliptical arc. */
+  public get sampleMethod(): EllipticalArcSampleMethod {
+    return this._sampleMethod;
+  }
+  /** Whether to return sample fractions grouped by quadrant.
```
saeeedtorabi, line 86:
```suggestion
  /** 
   * Whether to return sample fractions grouped by quadrant.
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,440 @@
...
+    this._numSamplesInQuadrant = numPointsInQuadrant;
+    this._maxError = maxError;
+    this._remapFunction = remapFunction;
+  }
+  /**
+   * Construct options with optional defaults.
+   * @param method sample method, default [[EllipticalArcSampleMethod.UniformParameter]].
+   * @param structuredOutput output format, default false.
+   * @param numSamplesInQuadrant samples in each full quadrant for interpolation methods, default 4.
+   * @param maxError max distance to ellipse for subdivision methods, default 1cm.
```
saeeedtorabi, line 74: 1 cm? Really? Do we have units? I thought these numbers are unitless! I mean when we say a length is 1, then it can be 1 cm or 1 inch or whatever!

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: Near duplicates should be disallowed by the comparison function for the relevant `OrderedSet` objects.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: In iTwin and imodel-native, coordinates are assumed to be in meters.

saeeedtorabi: oh nice!

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -14,6 +14,9 @@ import { CurveChain } from "../CurveCollection";
 import { CurveCurve } from "../CurveCurve";
 import { LineSegment3d } from "../LineSegment3d";
 import { LineString3d } from "../LineString3d";
+import { Path } from "../Path";
+import { Point3d } from "../../geometry3d/Point3dVector3d";
+import { NumberArray } from "../../core-geometry";
```
saeeedtorabi, line 19: I will remove this.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -451,75 +448,100 @@ export class EllipticalArcApproximationContext {
...
-   * @return max approximation error, or `Geometry.largeCoordinateResult` if the error could not be computed.
+   * Compute the maximum xy-distance between an elliptical arc, and its segment or circular arc approximation.
+   * * Inputs should be in horizontal plane(s), as z-coordinates are ignored.
+   * @param arc elliptical arc being approximated. Sweep outside the extents of the approximation is ignored.
+   * @param approximation circular arc, or segment, assumed to start and end on the elliptical arc.
+   * @return details of the perpendicular measuring the max approximation error, or undefined if no such perpendicular.
+   * For each of `detailA` and `detailB`:
+   * * `point` is the end of the perpendicular on each curve
+   * * `fraction` is the curve parameter of the point
+   * * `a` is the distance between the points.
```
saeeedtorabi, line 467: which points? The point on the curve and on the approximation when max error happen?

saeeedtorabi (changes requested):
A general question about error. Do you try different approxmiation methods and find the one with min error? If so, do you have a test for it?

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: Yes. We are measuring error in the max norm, which is the farthest distance between nearby portions of the two curves, which means length of the longest segment perpendicular to both an arc in the approximation, and to the portion of the ellipse it approximates. As is typical for detail pairs, we store the segment start/end data in `detailA` and `detailB`, and each detail stores the length of the segment in `a`.

dassaf4:
> A general question about error.

The current test attempts to measure the error for the naive approximation method, the uniform curvature interpolation method, and various non-uniform curvature interpolation methods. I'm seeing Newton fail to converge, or fail on singular matrices in several cases when we try to compute the error. I'm looking into this. It may be related to https://github.com/iTwin/itwinjs-backlog/issues/1138.

The goal is to find a good method that works over a broad range of eccentricities for low sample counts, and then just use that one method. Right now, there doesn't appear to be a clear winner. This is why I'm also working on a new method.

saeeedtorabi (approved):

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+    this._axisAtEnd = onAxis;
+  }
+  /**
+   * Whether the average fraction was added to produce a fractions array of length 3.
+   * * Some sample methods and elliptical arcs may not yield enough samples in a quadrant.
+   * * This flag is set if an extra (average) fraction was added to a 2-fraction array to satisfy
+   * the minimum 3-sample requirement that allows both axis/endpoints in the quadrant to be interpolated
+   * by [[EllipticalArcApproximationContext.constructCircularArcChainApproximation]].
+  */
+  public get averageAdded(): boolean {
```
saeeedtorabi, line 229: ok so we need at least 3 fractions on each quadrant to be able to pass a circle from those 3 points. If we have 2 fractions `f1` and `f2` we add the third fraction of `(f1+f2)/2` and set this flag.

That first line of the comment is confusing. I had to read next lines multiple times to understand it. Also the name `average` is confusing. I would go with `extraFraction` instead.

I also update the comment to something like this:

Whether the extra fraction was added.
To generate a circle, we need 3 points but some sample methods and elliptical arcs may not yield enough samples in a quadrant. 
This flag is set if an extra (average) fraction was added to a 2-fraction array to satisfy the minimum 3-sample requirement that allows both axis/endpoints in the quadrant to be interpolated by [[EllipticalArcApproximationContext.constructCircularArcChainApproximation]].

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+  }
+  /**
+   * Compute quadrant data for the given angles.
+   * @param radians0 first radian angle
+   * @param radians1 second radian angle
+   * @return quadrant number and start/end radian angles for the quadrant that contains both input angles, or
+   * undefined if no such quadrant.
+   * * The returned sweep is always counterclockwise: angle0 < angle1.
+   */
+  public static getQuadrantRadians(
```
saeeedtorabi, line 243: So one can pass `radian0` and `radian1` in different quadrants and then this function returns `undefined`, right? I so, please resolve this comment.

dassaf4: Correct!

saeeedtorabi: Initially I though it's better to make this function `private` but then it cannot be called in this file (because caller is in another class)

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+  protected _options: EllipticalArcApproximationOptions;
+  private _xMag2: number;
+  private _yMag2: number;
+  private _curvatureRange: Range1d;
+  protected constructor(c: EllipticalArcApproximationContext, o: EllipticalArcApproximationOptions) {
+    this._context = c;
+    this._options = o;
+    this._xMag2 = c.arc.matrixRef.columnXMagnitudeSquared();
+    this._yMag2 = c.arc.matrixRef.columnYMagnitudeSquared();
+    // extreme curvatures occur at the axis points
```
saeeedtorabi, line 320: Extreme curvatures occur at the axis points only if major/minor axes are parallel to xy axes, right? Since, that's not always the case, I assume before getting to this line, we already moved the ellipse into xy-plane and rotated it to make it parallel with xy axes, right?

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+   * @return radian angle in [0, pi/2] or undefined if the ellipse is invalid, or does not attain the given curvature.
+   */
+  private curvatureToRadians(curvature: number): number | undefined {
+    /*
+    Let the elliptical arc be parameterized with axes u,v of different length and u.v = 0:
+      f(t) = c + u cos(t) + v sin(t),
+      f'(t) = -u sin(t) + v cos(t),
+      f"(t) = -u cos(t) - v sin(t)
+    We seek a formula for t(K), the inverse of the standard curvature formula
+      K(t) := ||f'(t) x f"(t)||/||f'(t)||^3
```
saeeedtorabi, line 341:
```suggestion
      K(t) := ||f'(t) x f"(t)||/||f'(t)||^3
```
```suggestion
      K(t) := ||f'(t) x f"(t)|| / ||f'(t)||^3
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+    for a parametric function f(t):R->R^3. We'll restrict K to Q1 (i.e., t in [0, pi/2]), where K is monotonic.
+    By linearity of the cross product and the above formulas, the numerator of K(t) reduces to ||u x v||, and so:
+      cbrt(||u x v||/K) = ||f'(t)|| = sqrt(f'(t).f'(t))
```
saeeedtorabi, line 344:
```suggestion
      cbrt(||u x v||/K) = ||f'(t)|| = sqrt(f'(t).f'(t))
```
```suggestion
      cbrt(||u x v|| / K) = ||f'(t)|| = sqrt(f'(t).f'(t))
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+    Leveraging u,v perpendicularity we can define:
+      lambda(K) := (||u x v||/K)^(2/3) = (||u|| ||v|| / K)^(2/3) = cbrt(u.u v.v / K^2)
```
saeeedtorabi, line 346:
```suggestion
      lambda(K) := (||u x v||/K)^(2/3) = (||u|| ||v|| / K)^(2/3) = cbrt(u.u v.v / K^2)
```
```suggestion
      lambda(K) := (||u x v|| / K)^(2/3) = (||u|| ||v|| / K)^(2/3) = cbrt(u.u v.v / K^2)
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
saeeedtorabi: Initially I asked you to do the same for all division but I think next ones are actually better without space. so I deleted those comments.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+        if (undefined !== angle)
+          result.push(angle);
+      }
+    }
+    return result;
+  }
+};
+
+/**
+ * Implementation for method `EllipticalArcSampleMethod.UniformCurvature`
```
saeeedtorabi, line 379: Maybe in the doc say something about this class basically uses non unifrom curvature with identity map which makes it unifrom!

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: Yes, "axis points" is vague, and is true if ellipse axes have been squared. (The ellipse doesn't have to be aligned to global xy axes, however.)  The context ctor stores the squared axis formulation of the input arc, so all logic that uses the context's arc can assume squared axes.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+ */
+class SubdivideForArcsSampler extends SubdivisionSampler {
+  private constructor(c: EllipticalArcApproximationContext, o: EllipticalArcApproximationOptions) {
+    super(c, o);
+  }
+  public static create(context: EllipticalArcApproximationContext, options: EllipticalArcApproximationOptions): SubdivideForArcsSampler {
+    return new SubdivideForArcsSampler(context, options);
+  }
+  public override createErrorProcessor(): SubdivisionErrorProcessor {
+    return new ArcChainSubdivisionErrorProcessor(this.fullArcXY, this._options.maxError);
```
saeeedtorabi, line 456: similarly, please move ArcChainSubdivisionErrorProcessor before this class

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+  */
+  public announceQuadrantBegin(_q: QuadrantFractions): boolean { return true; }
+  /**
+   * Optionally announce a chord of the elliptical arc between the given fractions.
+   * @param pt0 start point of the chord segment
+   * @param pt1 end point of the chord segment
+   * @param f0 elliptical arc parameter in [0,1] of pt0
+   * @param f1 elliptical arc parameter in [0,1] of pt1
+   */
+  public announceChord?(pt0: Point3d, pt1: Point3d, f0: number, f1: number): void;
```
saeeedtorabi, line 477: Maybe it's too much pain to change the word `chord` but I think we use the term `segment` eveywhere else!

dassaf4: Chord has additional meaning: its endpoints are on another curve (in this case, the elliptical arc). That's why I prefer it to segment.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+  public override announceChord(pt0: Point3d, pt1: Point3d, f0: number, f1: number): void {
+    this.updateMaxPerpendicular(LineSegment3d.create(pt0, pt1), f0, f1);
+  }
+}
+
+/**
+ * Intermediate processor class for computing samples for a subdivision-based approximation.
+ * @internal
+ */
+class SubdivisionErrorProcessor extends ErrorProcessor {
```
saeeedtorabi, line 605: It's not clear what this class is doing. What refinement set is, what is the count. Please add some comments to make it clear.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+      this._refinement.add(Geometry.interpolate(f0, 0.5, f1));
+  }
+  public override announceQuadrantEnd(q: QuadrantFractions, _wasReversed: boolean): void {
+    q.fractions = [...this._refinement];
+  }
+  /** Whether the processor refined the sample set to decrease the approximation error. */
+  public get isRefined(): boolean {
+    return this._originalRefinementCount < this._refinement.size;
+  }
+  public getRefinedAngles(result?: number[]): number[] {
```
saeeedtorabi, line 636: similarly what the `result` array?

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+/**
+ * Intermediate class for constructing a sample-based approximation.
+ * @internal
+ */
+class ConstructionProcessor extends QuadrantFractionsProcessor {
+  protected _chain: CurveChain;
+  protected _quadrantChain?: CurveChain;
+  public constructor(ellipticalArc: Arc3d) {
+    super();
+    this._chain = ellipticalArc.sweep.isFullCircle ? Loop.create() : Path.create();
```
saeeedtorabi, line 684: so `isFullCircle` means arc sweep is from 0 to 360 and does not mean the arc is actually a circle, right?

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1092 @@
...
+  }
+  protected set maxPerpendicular(newMaxPerp: CurveLocationDetailPair) {
+    this._maxPerpendicular = newMaxPerp;
+  }
+  /**
+   * Update the chain approximation error for a given chain child that approximates the instance arc between the given
+   * fractions.
+   * * Fractional sweep of the arc is the smaller of the cyclic sweeps.
+   */
+  protected updateMaxPerpendicular(childApproximation: Arc3d | LineSegment3d, f0: number, f1: number): void {
```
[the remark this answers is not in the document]
dassaf4: Right, I'm using the max norm, aka `l_{\infty}`, which is the simplest to compute.  Basically this defines the approximation error as the largest gap between the input elliptical arc, and any approximating circular arc or chord.  There's no summation involved. Other norms involve more work for different measures of error, which tend to reduce the effect of outliers: for `l_1` norm we add up the absolute error for each arc/chord (Manhattan distance); for `l_2` norm we use Euclidean distance = the square root of the sum of the squared errors. Typically, max norm suffices, and is what is historically expected from strokers that take an absolute error input.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
saeeedtorabi: well, both ends are on the arc, are not they?! I mean for example, here we have 4 lines segments and all of start/end points are on the same arc, right?

image

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: I'll document. But its usage further below clarifies.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: Yes, exactly, that's why I call them chords.  More descriptive than segments.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: I'll add more comments. Perhaps it will be clearer when you see how it's used below. The set is initially `q.fractions`. Then it is refined by adding a midpoint in each interval where the error of the announced approximant is too large.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: Correct.  `AngleSweep` is a general class for dealing with angles; it knows nothing about ellipses. `Circle` is descriptive, if perhaps a bit confusing here. Here, `isFullPeriod` would be better, however, `AngleSweep.isFullPeriod` would be incorrect in a context where the period is _not_ 2pi. So we have `AngleSweep.isFullCircle`.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+   * * The processor is allowed to capture `arc`, e.g., to add it to a growing chain approximation.
+   * @param f0 elliptical arc parameter in [0,1], unordered with respect to f1
+   * @param f1 elliptical arc parameter in [0,1], unordered with respect to f0
+   */
+  public announceArc?(arc: Arc3d, f0: number, f1: number): void;
+  /**
+   * Announce the end of processing of the samples in the given quadrant.
+   * @param _wasReversed whether the given quadrant's samples were processed in reverse
+  */
+  public announceQuadrantEnd(_q: QuadrantFractions, _wasReversed: boolean): void { }
```
saeeedtorabi, line 292: I move this a few lines above right after `announceQuadrantBegin`

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+ * Processor for constructing a sample-based arc chain approximation.
+ * @internal
+ */
+class ArcChainConstructionProcessor extends ConstructionProcessor {
+  public constructor(ellipticalArc: Arc3d) {
+    super(ellipticalArc);
+  }
+  public override announceArc(arc: Arc3d, _f0: number, _f1: number): void {
+    if (!this._quadrantChain)
+      this._quadrantChain = Path.create();
```
saeeedtorabi, line 727: `_chain` in `ConstructionProcessor` class can be loop or path but here `_quadrantChain` is path. How so?

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+    this._quadrantChain.tryAddChild(arc); // captured!
+  }
+};
+
+/**
+ * Processor for constructing a sample-based linestring approximation.
+ * @internal
+ */
+class LineStringConstructionProcessor extends ConstructionProcessor {
```
saeeedtorabi, line 736: `ChordConstructionProcessor` ?

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+    super(ellipticalArc);
+  }
+  public override announceArc(arc: Arc3d, _f0: number, _f1: number): void {
+    if (!this._quadrantChain)
+      this._quadrantChain = Path.create();
+    this._quadrantChain.tryAddChild(arc); // captured!
+  }
+};
+
+/**
```
saeeedtorabi, line 732: What is the diff between this and stroking the arc? I assume when we stroke we dont have much control on the stroke points but here we do

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+export class EllipticalArcApproximationContext {
+  private _arc: Arc3d;
+  private _localToWorld: Transform;
+  private _isValidArc: boolean;
+  private static workPt0 = Point3d.createZero();
+  private static workPt1 = Point3d.createZero();
+  private static workPt2 = Point3d.createZero();
+  private static workRay = Ray3d.createZero();
+
+  /** captures input */
```
saeeedtorabi, line 771:
```suggestion
  /** Constructor, captures input. */
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+      // assume non-colinear inputs
+      const myArc = Arc3d.createCircularStartTangentEnd(arcStart.origin, arcStart.direction, arcEnd) as Arc3d;
+      if (reverse)
+        myArc.reverseInPlace();
+      return myArc;
+    };
+    const createFirstArc = (f0: number, f1: number, reverse: boolean) => {
+      ellipticalArc.fractionToPointAndDerivative(f0, ray);
+      pt0.setFrom(ray.origin);
+      pt2.setFrom(ellipticalArc.fractionToPoint(f1, pt1)); // set pt2 in case no inner arcs
```
saeeedtorabi, line 834: is `pt1 = pt2`? If so, why do you even need `pt2`?

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: It's there because the intervening methods occur between announceBegin/End.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: Because in a single quadrant, the approximant can only be a path.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: It returns a LineString3d, so LineString is more descriptive I think.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: That's the difference. Mainly, this is here to facilitate comparison against the arc chain constructor. I fully expect no one to use the linestring stuff when it's much easier to use `emitStrokes`.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+      if (processor.announceArc) {
+        if (reverse)
+          ray.direction.scaleInPlace(-1);
+        processor.announceArc(arcBetween2Samples(ray, pt1, false), f0, f1);
+      }
+      if (processor.announceChord)
+        processor.announceChord(pt0, pt1, f0, f1);
+      return { p0: pt0, p1: pt1, p2: pt2 };
+    };
+    const arcBetweenLast2Of3Samples = (firstPt: Point3d, arcStart: Point3d, arcEnd: Point3d): Arc3d => {
```
saeeedtorabi, line 844: I found this name confusing. I know we use 3 points to create the partial circle and then only take part of it from middle to end. However, since we use 2 points for first and last partial circle, this name confuses me.

Also in general the word `arc` is used for both original arc that we are approximating and also for the partial circles and that's confusing too!

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+      // assume non-colinear inputs
+      const arc = Arc3d.createCircularStartMiddleEnd(firstPt, arcStart, arcEnd) as Arc3d;
+      const startAngle = arc.vector0.signedAngleTo(Vector3d.createStartEnd(arc.center, arcStart), arc.matrixRef.columnZ());
+      arc.sweep.setStartEndRadians(startAngle.radians, arc.sweep.endRadians);
```
saeeedtorabi, line 848: I would add a comment to explain this line saying we take from mid to end. Maybe that also makes the name `arcBetweenLast2Of3Samples` less confusing.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
dassaf4: I'm currently taking suggestions for better distinction of "arcBetween2Samples" and "arcBetweenLast2Of3Samples"!

Yeah, if you find specific doc that needs "arc" clarification, I can fix.  At some point I got tired of writing "elliptical arc" and "circular arc" and assumed reader would know which is inferred from the context.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
saeeedtorabi: Then I would remove 
``` 
pt2.setFrom(ellipticalArc.fractionToPoint(f1, pt1)); // set pt2 in case no inner arcs
```
and change last line to
```
return { p0: pt0, p1: pt1, p2: pt1 }; // set p2 in case no inner arcs
```

dassaf4: Cannot remove that line.  It sets pt1.  And that line contains the comment.

saeeedtorabi: ok

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+      if (reverse)
+        q.fractions.reverse();  // restore original order just to be safe
+      processor.announceQuadrantEnd(q, reverse);
+    }
+  }
+  /**
+   * Compute the maximum error of the given approximation.
+   * * This is measured by the longest perpendicular between the elliptical arc and its approximation.
+   * @param samples structured sample data for the instance's elliptical arc.
+   * @param arcChain whether to compute approximation error of a circular arc chain (true) or linestring (false)
```
saeeedtorabi, line 903:
```suggestion
   * @param arcChain whether to compute approximation error of a circular arc chain (true) or linestring (false).
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+    const arcXY = this.cloneLocalArc();
+    const processor = arcChain ? new ArcChainErrorProcessor(arcXY) : new LineStringErrorProcessor(arcXY);
+    EllipticalArcApproximationContext.processQuadrantFractions(arcXY, samples, processor);
+    const maxError = processor.maxPerpendicular;
+    return (maxError && maxError.tryTransformInPlace(this.localToWorld)) ? maxError : undefined;
+  }
+  /**
+   * Compute samples for the elliptical arc as fraction parameters.
+   * @param options approximation options
+   * @param structuredOutput
```
saeeedtorabi, line 922: Please add what `structuredOutput` is

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+      return undefined;
+    const arcXY = this.cloneLocalArc();
+    const processor = arcChain ? new ArcChainErrorProcessor(arcXY) : new LineStringErrorProcessor(arcXY);
+    EllipticalArcApproximationContext.processQuadrantFractions(arcXY, samples, processor);
+    const maxError = processor.maxPerpendicular;
+    return (maxError && maxError.tryTransformInPlace(this.localToWorld)) ? maxError : undefined;
+  }
+  /**
+   * Compute samples for the elliptical arc as fraction parameters.
+   * @param options approximation options
```
saeeedtorabi, line 921:
```suggestion
   * @param options approximation options.
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
saeeedtorabi: oh you have its details on the next line. Still if you add couple words line `a flag for whatever` makes the doc more readable.

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+      for (const a1 of qEndAngles) {
+        const quadrant = convertQ1RadiansInSweepToQuadrantFractions(anglesInQ1, a0, a1, arcSweep);
+        if (quadrant)
+          quadrants.add(quadrant);
+        a0 = a1;
+      }
+      return [...quadrants];
+    };
+    const computeFlatOutput = (anglesInQ1: number[], arcSweep: AngleSweep): number[] => {
+      // Flat array output: first add quadrant fractions so the set prefers them over nearby interior fractions
```
saeeedtorabi, line 1041: It's not super important but I try to be consistent with comments and start with lower case as you did for all above comments.

```suggestion
      // flat array output: first add quadrant fractions so the set prefers them over nearby interior fractions
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+      const fractions = new OrderedSet<number>(compareFractions);
+      fractions.add(0);
+      fractions.add(1);
+      for (const angle of [0, Angle.piOver2Radians, Angle.piRadians, Angle.pi3Over2Radians])
+        convertAndAddRadiansToFractionInRange(fractions, angle, arcSweep);
+      // Add interior Q1 fractions, reflect to the other quadrants, filter by sweep and extant entry
```
saeeedtorabi, line 1047:
```suggestion
      // add interior Q1 fractions, reflect to the other quadrants, filter by sweep and extant entry
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
```diff
@@ -0,0 +1,1106 @@
...
+      for (const angle0 of anglesInQ1) {
+        for (const angle of [angle0, Angle.piRadians - angle0, Angle.piRadians + angle0, Angle.pi2Radians - angle0])
+          convertAndAddRadiansToFractionInRange(fractions, angle, arcSweep);
+      }
+      return [...fractions];
+    };
+    // Sample the (full) ellipse as angles in strict interior of Quadrant 1
```
saeeedtorabi, line 1054:
```suggestion
    // sample the (full) ellipse as angles in strict interior of Quadrant 1
```

core/geometry/src/curve/internalContexts/EllipticalArcApproximationContext.ts
saeeedtorabi, line 1: This is very nice file and doing lots of things and it makes it harder to follow even for me who knows what is the class doing. Adding 5-10 lines of comments at the start of the file and summerizing what we are doing (approximation an arc using circular arcs or chords, different points distribuition, how we approximate, etc) makes the class much easier to follow and understand.

dassaf4: Typically we put comments close to the code so that it tends not to diverge as much, rather than at top of file. (Almost everywhere you see doc at the top of a file in geomlibs, it is out of date.) The key classes and methods have doc blocks. I'll make a pass to add some additional doc where it is thin.

core/geometry/src/test/curve/Arc3d.test.ts
```diff
@@ -650,6 +650,29 @@ describe("Arc3d", () => {
     expect(ck.getNumErrors()).equals(0);
   });
```
saeeedtorabi, line 652: I moved this nice function outside `EllipseSampler` test so I can use it in my test.

pmconne (approved):
Reviewed core-common. Lots of noise from unrelated changes in this PR.

dassaf4:
Will wait for Saeed's review of latest changes.

PR #6792, opened 2024-06-05
Add GL headless support on linux chrome

mathieu-lemuzic:
Also add a new flag to enable EGL for creating headless GL context on chrome / linux.

This is required to run image tests on headless linux virtual machines, e.g., that do not have a display.

### DIFF
--- a/test-apps/display-performance-test-app/src/backend/WebMain.ts
+++ b/test-apps/display-performance-test-app/src/backend/WebMain.ts
@@ -58,6 +58,10 @@ function startWebServer() {
       browser = arg;
     else if (arg === "headless")
       chromeFlags.push("--headless");
+    else if (arg === "egl"){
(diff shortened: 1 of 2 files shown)


test-apps/display-performance-test-app/src/backend/WebMain.ts
```diff
@@ -58,6 +58,10 @@ function startWebServer() {
       browser = arg;
     else if (arg === "headless")
       chromeFlags.push("--headless");
+    else if (arg === "egl"){
```
mathieu-lemuzic, line 61: @pmconne Is there a way to get chrome to pick up the GPU without these changes ?

markschlosseratbentley: Looks good to me.

markschlosseratbentley (approved):

PR #6806, opened 2024-06-06
Identifiertypes refactor

christophermlawson:
Added SchemaTypeIdentifiers enum and interfaces for discriminating identifier types. 
Carole needed Label editing support (schema items). I added description as well.
Removed BentleyError now deriving from Error
Changed enumeration identifier
Added missing documentation
Exported Exception.ts from module
Make sure exports and tags (alpha) are correct.
Added generic error message for 'task' exceptions (outer errors).


core/ecschema-editing/src/Editing/SchemaItems.ts
```diff
@@ -27,6 +28,32 @@ export abstract class SchemaItems {
...
+   * @param label The new label to set.
+   */
+  public async setDisplayLabel(schemaItemKey: SchemaItemKey, label: string) {
+    const item = await this.getSchemaItem<MutableSchemaItem>(schemaItemKey)
+      .catch((e: any) => {
+        throw new SchemaEditingError(ECEditingStatus.SetLabel, new SchemaItemId(this.schemaItemType, schemaItemKey), e);
+      });
+    item.setDisplayLabel(label);
+  }
+
```
StefanApfel-Bentley, line 56: If I understand that right, we provide the editing methods here in an "unspecified" item-space, but don't call this ourself anywhere yet. I guess the Merger will call it though once available which can save us to cast to the mutables at some places.

However, if we don't call it here (in this PR scope), we might want to have an own PR for that in which we can also change the merger code.

core/ecschema-editing/src/test/Editing/Exception.test.ts
```diff
@@ -34,12 +34,46 @@ describe("SchemaEditingError tests", () => {
...
+  it("Should create correct error message for Schema task error", async () => {
+    const identifier = new SchemaId(testKey);
+    const error = new SchemaEditingError(ECEditingStatus.CreateElement, identifier, new Error("inner error"));
+    expect(error.message).to.equal(`While performing task '${ECEditingStatus[error.errorNumber]}' an error occurred editing Schema ${identifier.name}.`);
+  });
+
+  it("Should create correct error message for SchemaItem task error", async () => {
+    const identifier = new SchemaItemId(SchemaItemType.Unit, "TestUnit", testKey);
+    const error = new SchemaEditingError(ECEditingStatus.CreateSchemaItemFailed, identifier, new Error("inner error"));
+    expect(error.message).to.equal(`While performing task '${ECEditingStatus[error.errorNumber]}' an error occurred editing SchemaItem ${identifier.name}.`);
```
StefanApfel-Bentley, line 55: I know we've been doing it that way for a while now, but I think, we should get in the habit of having static error messages. In this case
```ts
"While performing task 'CreateSchemaItemFailed' an error occurred editing SchemaItem TestUnit."
```
That's a little more effort if we change something, but you see at once what the expected result is, and things don't slip through so easily if we change something on the backend side than if we'd compose the expected results dynamically.

core/ecschema-editing/src/Editing/SchemaItems.ts
christophermlawson: Carole requested these methods (she actually requested the label, but I added description too).  Up to this point, unless we have to due to failing tests in the merger, I've kept editing API code changes separate from any merger changes.

core/ecschema-editing/src/test/Editing/Exception.test.ts
christophermlawson: I think we should discuss this.  I want to make sure I understand what you are saying.

StefanApfel-Bentley (approved):

PR #6816, opened 2024-06-10
Ensure that compared schema item type are compatible to each other

AureMonke:


StefanApfel-Bentley:
At the moment, we only check on properties and custom attributes whether the looked up item matches the type of the comparison. Since our comparer methods all expect certain types being passed in, I think we should better check all lookup calls that the returned item is a supported type of the comparer.

By doing that you can also safe the generic when lookup the schema item, as we'd have the typeguard method afterwards that type out item to be valid for the comparer method.

```ts
public async visitUnit(unitA: Unit): Promise<void> {
  const lookupItem = await this._schemaB.lookupItem(unitA.name);
  if(Unit.isUnit(lookupItem))
    this._schemaComparer.compareUnits(unitA, lookupItem);
}
```

StefanApfel-Bentley (approved):

PR #6818, opened 2024-06-10
Refactor the Mesh Export Service to Create the Graphics Service and the Graphics Service Provider

andremig-bentley:
WIP

Refactor the frontend tiles package to decouple the Mesh Export Service access points from batched tiles. The goal is to be able to use the Mesh Export Service to obtain tiles other than only batched tiles in the future. To do this, the GraphicsService and the GraphicsServiceProvider have been created. The GraphicsService will work as a connector - outside sources such as frontend tiles will request tileset urls from the GraphicsService. The GraphicsService will then obtain the url from the GraphicsServiceProvider and then send it back to the outside source.

Resolves https://github.com/iTwin/itwin-graphics-backlog/issues/249


common/api/frontend-tiles.api.md
```diff
@@ -28,6 +36,36 @@ export const frontendTilesOptions: {
     useIndexedDBCache: boolean;
 };
 
+// @beta
+export interface GraphicRepresentation {
+    dataSource: DataSource;
+    displayName: string;
+    format: GraphicRepresentationFormat;
```
pmconne, line 43: tsc's exhaustiveness checks mean that you will not be able to add new values to the enum without potentially breaking somebody's code. I expect we will add new formats in the future.

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+ * @beta
+ */
+export interface DataSource {
+  /** The iTwinId associated with th data source */
+  iTwinId: string;
+  /** The id of the data source */
+  id: string;
+  /** The change id of the data source */
+  changeId?: string;
+  /** The type of the data source */
```
pmconne, line 41: Describe possible values.

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+ * A data source is usually higher-fidelity and contains more information, but may not be as well suited for visualization
+ * as a graphic representation.
+ * @beta
+ */
+export interface DataSource {
+  /** The iTwinId associated with th data source */
+  iTwinId: string;
+  /** The id of the data source */
+  id: string;
+  /** The change id of the data source */
```
pmconne, line 39: This documentation comment adds no information that is not already conveyed by the property name. Describe what "change id" means.

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * Represents a data source for a graphic representation.
+ * A data source is usually higher-fidelity and contains more information, but may not be as well suited for visualization
+ * as a graphic representation.
+ * @beta
+ */
+export interface DataSource {
+  /** The iTwinId associated with th data source */
+  iTwinId: string;
+  /** The id of the data source */
```
pmconne, line 37: Explain what "id" means.

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+  id: string;
+  /** The change id of the data source */
+  changeId?: string;
+  /** The type of the data source */
+  type: string;
+}
+
+/** represents a visual representation of data from a data source.
+ * This could be a 3D model, a map, or any other kind of graphical data.
+ * @see [[queryGraphicsDataSources]].
```
pmconne, line 47: An `@see` comment should always explain **why** you should consult the linked documentation (e.g., in this case, because you want to find data available data sources).

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+  /** The Id of the Graphic Representation */
+  representationId: string;
+  /** The status of the query from the Graphics Provider for the Graphic Representation */
+  status: GraphicRepresentationStatus;
+  /** The format of the Graphic Representation*/
+  format: GraphicRepresentationFormat;
+  /** The url of the Graphic Representation */
+  url?: string;
+  /** The data source that the representation originates from.
+  * For example, a GraphicRepresentation in the 3D Tiles format might have a dataSource that is a specific iModel changeset.
```
pmconne, line 62: This is an example of a documentation comment that provides useful information. Revisit your other doc comments to make them do the same.

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+  url = `${url}&tileVersion=1&exportType=IMODEL`;
+
+  return url;
+}
+
+/** Arguments supplied to [[queryGraphicsDataSources]].
+ * @beta
+ */
+export interface QueryGraphicRepresentationsArgs {
+  /** The token used to access the mesh export service. */
```
danieliborra, line 86:
```suggestion
  /** The token used to access the data source provider. */
```

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+  accessToken: AccessToken;
+  /** The Session Id */
+  sessionId: string;
+  /** The Data Source for which to query */
```
danieliborra, line 90:
```suggestion
  /** The Data Source for which to query the graphic representations*/
```

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+  /** Chiefly used in testing environments. */
+  urlPrefix?: string;
+  /** If true, exports whose status is not "Complete" (indicating the export successfully finished) will be included in the results */
+  includeIncomplete?: boolean;
+  /** If true, enables a CDN (content delivery network) to access tiles faster. */
+  enableCDN?: boolean;
+}
+
+/** Query Graphic Representations for representations matching the specified criteria.
+ * The representations are sorted from most-recently- to least-recently-produced.
```
danieliborra, line 103:
```suggestion
/** Query graphic representations matching the specified criteria, sorted from most-recently- to least-recently-produced.
```

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+
+/** Arguments supplied  to [[obtainGraphicRepresentationUrl]].
+ * @beta
+ */
+export interface ObtainGraphicRepresentationUrlArgs {
+  /** The token used to access the mesh export service. */
+  accessToken: AccessToken;
+  /** The Session Id */
+  sessionId: string;
+  /** The data source for which to query */
```
danieliborra, line 192:
```suggestion
  /** The data source for which to query the graphic representations */
```

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+ * @beta
+ */
+export interface QueryGraphicRepresentationsArgs {
+  /** The token used to access the mesh export service. */
+  accessToken: AccessToken;
+  /** The Session Id */
+  sessionId: string;
+  /** The Data Source for which to query */
+  dataSource: DataSource;
+  /** The format of the Graphics Data Source */
```
danieliborra, line 92:
```suggestion
  /** The expected format of the graphic representations */
```

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+ * @beta
+ */
+export interface ObtainGraphicRepresentationUrlArgs {
+  /** The token used to access the mesh export service. */
+  accessToken: AccessToken;
+  /** The Session Id */
+  sessionId: string;
+  /** The data source for which to query */
+  dataSource: DataSource;
+  /** The format of the Graphics Data Source */
```
danieliborra, line 194:
```suggestion
  /** The expected format of the graphic representations */
```

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  const queryArgs: QueryGraphicRepresentationsArgs = {
+    accessToken: args.accessToken,
+    sessionId: args.sessionId,
+    dataSource: {
+      iTwinId: args.dataSource.iTwinId,
+      id: args.dataSource.id,
+      changeId: args.dataSource.changeId,
+      type: args.dataSource.type,
+    },
```
danieliborra, line 226:
```suggestion
    dataSource: args.dataSource,
```

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+    format: args.format,
+    urlPrefix: args.urlPrefix,
+    enableCDN: args.enableCDN,
+  };
+
+  let selectedData;
```
danieliborra, line 232: I'm curious if co-pilot could simplify this using "find". Do not change it this PR, it is just an idea to consider in the future.

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,255 @@
...
+  /** The id of the data source */
+  id: string;
+  /** The change id of the data source */
+  changeId?: string;
+  /** The type of the data source */
+  type: string;
+}
+
+/** represents a visual representation of data from a data source.
+ * This could be a 3D model, a map, or any other kind of graphical data.
```
danieliborra, line 46:
```suggestion
/** Represents a visual representation of a data source, for example a 3d tileset.
 * A data source can be an iModel, reality data, or other kind of graphical data.
```

common/api/frontend-tiles.api.md
danieliborra: We had a long discussion about it. While we agree that we will add new formats in the future, in the short-mid term we are the ones that will add new formats/providers, so an enum is good for now. We thought about implementing a union, which could be extended, but we thought about keeping this simple for now.

If we want to add new formats, we should use a "general" provider that provides functionality to register data sources, but if we go in that direction, we should rethink also the other providers (e.g. reality data). So, I'd personally keep this simple for now, and in parallel have a serious discussion about unifying the data sources in the iTwin Viewer in global.

danieliborra (commented):
Looks good to me. Many good discussions/comments about documentation, and the only "serious" discussion is about the graphics representation format enum.
Let's try to wrap this PR up today.

common/api/frontend-tiles.api.md
pmconne: > We had a long discussion about it.

Who is "we"?

> While we agree that we will add new formats in the future, in the short-mid term we are the ones that will add new formats/providers, so an enum is good for now.

When you introduce those new formats, you run the risk of introducing run-time exceptions and/or compile-time errors in other people's code.

danieliborra: > Who is "we"?
Andre, Erin and me.

> When you introduce those new formats, you run the risk of introducing run-time exceptions and/or compile-time errors in other people's code.

Can you give me an example? Adding a new value to the enum shouldn't break anything, right? I guess I'm missing something here or maybe thinking "C++ minded".
The goal is to make sure that users are restricted to only use the supported formats.
The restriction can easily be removed, but isn't it better to restrict it to produce compile-time errors if the format is not well defined?

pmconne:
```
enum Enum { A, B };
function getAThing(enum: Enum): Thing {
  switch (enum) {
    case Enum.A: return new ThingA();
    case Enum.B: return new ThingB();
    // no default required - every possible case covered
  }
  // no default return required - unreachable code
}
```

Now add `Enum.C` and suddenly `getAThing` can return `undefined` at run-time. If the author of the code above recompiles, compilation will fail.

danieliborra: Ahhhh, thanks for sharing!
Unfortunately, that makes all the sense 😅 
Ok, then enum out.

pmconne: For documentation purposes you could use a type like `"IMDL" | "3DTILES" | string`.

andremig-bentley: the enum has been removed and replaced with a type of `"IMDL" | "3DTILES" | string`

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
andremig-bentley: Docs have been rewritten to be more useful.

danieliborra (commented):
Looks good to me, I'll let Paul approve the PR

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,263 @@
...
+   * For example, if a specific version of an iModel is desired, the iModel's changesetId would be attributed to this value.
+  */
+  changeId?: string;
+  /** The type of the data source. For example, a DataSource can be of type "IMODEL" or "RealityData" */
+  type: string;
+}
+
+/** Represents a visual representation of a data source, for example a 3d tileset.
+ * A data source can be an iModel, reality data, or other kind of graphical data.
+ * @see [[queryGraphicsDataSources]] for its construction as a representation of the data produced by a query of data sources.
```
eringram, line 49: Just noticed some of the documentation and test descriptions still have the old naming which I would suggest fixing. There are 2 instances in this file, and all the others are all in GraphicRepresentationProvider.test.ts. I just searched for "GraphicsDataSource" and "graphics data source" to find them

andremig-bentley: Good catch, they should all be updated now.

extensions/frontend-tiles/src/FrontendTiles.ts
```diff
@@ -74,6 +61,8 @@ export interface MeshExports {
 export interface QueryMeshExportsArgs {
   /** The token used to access the mesh export service. */
   accessToken: AccessToken;
+  /** If defined, the iTwinId associated with the Mesh Export */
```
pmconne, line 64: It says "if defined" but it's a required property.

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -0,0 +1,263 @@
...
+  /** The status of the generation of the Graphic Representation from its Data Source.
+   * @see [[GraphicRepresentationStatus]] for possible values.
+   */
+  status: GraphicRepresentationStatus;
+  /** The expected format of the Graphic Representation
+   * @see [[GraphicRepresentationFormat]] for possible values.
+   */
+  format:  GraphicRepresentationFormat;
+  /** The url of the Graphic Representation */
+  url?: string;
```
pmconne, line 66: Under what circumstances would `url` be undefined?

pmconne (approved):

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
eringram: If the status of the graphic representation is not "complete", it won't be guaranteed to have a url for the data. For example if you query a representation that was recently created it might still be "in progress" or if there was an error during generation it could be "failed" and a url may not have been added yet. Do you think it makes sense for it to be optional if this is explained in the docs?

pmconne: Will `url` be valid if and only if `status` === "Complete"?

andremig-bentley: `url` is not valid if and only if `status` === "Complete", but that is the only time we can guarantee that it is valid. Currently, the url will show up at some point while the graphic representation is being generated, meaning at some point while the `status` === "In Progress". Until that point the url is undefined. But this means if you query a representation and get one that is "In Progress", there's a chance there is a url, but we can't guarantee it, and it might be undefined. @eringram correct me if I'm wrong on any of this or if you want to add anything.

extensions/frontend-tiles/src/FrontendTiles.ts
andremig-bentley: The docs were wrong here, at one point we were considering having `iTwinId` be optional, but i didn't update the doc comment when we changed that. That fix is pushed now.

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
andremig-bentley: This change has been pushed, GraphicRepresentation is now a type with the above structure.

danieliborra: I had no idea that was possible in typescript, amazing, thanks @pmconne !!

PR #6825, opened 2024-06-11
Allow App to specify GCS asset Dir

DanRod1999:
imodel-native: https://github.com/iTwin/imodel-native/pull/748
resolves: https://github.com/iTwin/itwinjs-backlog/issues/1134

Be able to specify the location for GCS data in the native code when gcs workspaces are disabled. Visualization/GPB want to use it when loading up GPB in the kubernetes clusters. This is related to one of the experiments during the pineapple viewer hackathon.


kabentley:
I don't understand the reason for this PR (the link above doesn't work for me). What is a "geocoord asset dir" and why does anyone need it?

DanRod1999:
> I don't understand the reason for this PR (the link above doesn't work for me). What is a "geocoord asset dir" and why does anyone need it?

Geocoord asset Dir, is what I'm calling the local directory storing GCS data.
We found a perf hit during imodel opens, and it was due to demands on GCS data. To avoid this the idea was to disable workspaces in GBP in the Kubernetes clusters, and package the GCS data on the nodes and mount the path of the GCS data to all the nodes on the cluster.
Typically when workspaces are disabled it defaults to the asset directory setup in IKnownLocationsAdmin, and when stored in the addon the path is something like `.../@bentley+imodeljs-native@4.5.38` which due to the change versions would require editing the mounted path every time. So to prevent that we wanted the ability to specify the path we store the gcs data

pmconne:
> We found a perf hit during imodel opens

Can you quantify this perf hit? Host many seconds will you save by making the proposed change?

kabentley:
Of course the only performance improvement is startup time, and I'm skeptical this will help much, but I would like to understand where the time actually goes. 

But, I spoke to @wgoehrig about this and explained there is no need to re-open this old wound - just use local WorkspaceDb file(s) and the existing api works just fine. Reading loose CSMap files should be permanently removed from the code.

PR #6828, opened 2024-06-11
Define policy for internal APIs

pmconne:
Part of #6742.

See release-tags-guidelines.md.

I wanted to start at the bottom and work my way up.
I addressed all of the `@internal` APIs in core-bentley. That package doesn't include any nested APIs that we want to keep internal, so doesn't need to apply the `Symbol` approach.
I made representative changes to core-backend to see if I hit any unexpected snafus. I may make a few more.

imodel-native: https://github.com/iTwin/imodel-native/pull/797


docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,188 @@
...
+}
+```
+
+Update all callers of `close` in itwinjs-core to use `[_close]` instead.
+
+The symbols in `Symbols.ts` are not exported from `core-backend.ts`, so they are inaccessible to packages outside of itwinjs-core. We preserve the original `@internal` API to avoid introducing a breaking API change for external callers (who should not be using it in the first place). The deprecation tag will warn them to fix their code to remove their dependency on the internal API. If we were reasonably certain no one was depending on the `close` method, we could simply delete it.
+
+### Transition plan
+
+In iTwin.js 5.0, we will be able to begin enforcing the internal API policy without regard for breaking changes.
```
pmconne, line 174: @aruniverse @calebmshafer @wgoehrig can we commit to the listed changes for iTwin.js 5.0?

aruniverse: I'm ok with this as long as we give fair/advanced warning to App teams and they know what they're getting themselves into.

Changes below, like standardizing on ESM is a dream that I'm not 100% sure we'll be able to realize in time.

pmconne: > Changes below, like standardizing on ESM 

@wgoehrig suggested standardizing on ESM was in scope. Is that the only one you potentially object to? (You said "like" that one). If we can't commit to any of them, we're not going to solve the internal API (and hence the API stability) problem.

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,205 @@
...
 
-## Supported Release Tags
+## Supported tags
 
-The four primary release tags and the "deprecated" tag are described in the [API support policies](../api-support-policies.md). For [Extensions](../frontend/Extensions.md), two additional tags are supported:
+Each API in iTwin.js belongs to one of four API categories as defined by its associated release tag:
 
-- The "extensions" tag indicates that an API is to be included in the `@itwin/core-extension` API. Currently this tag is only useful for APIs in `@itwin/core-frontend` and `@itwin/core-common`; all other packages' APIs can be used directly by extensions. It may become relevant for more packages in the future. A lint rule enforces that the "extensions" tag may only be applied to APIs tagged as "public".
+- `@public` indicates a stable, production-ready API subject to our documented [stability and deprecation policies](../api-support-policies.md).
+- `@beta` indicates an API undergoing stabilization. Users are encouraged to experiment with it in non-production environments.
```
calebmshafer, line 10: I'd like to avoid saying "non-production environments" and rather highlight that they may change in a breaking way in any version subsequent version.

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,205 @@
...
 
-## Supported Release Tags
+## Supported tags
 
-The four primary release tags and the "deprecated" tag are described in the [API support policies](../api-support-policies.md). For [Extensions](../frontend/Extensions.md), two additional tags are supported:
+Each API in iTwin.js belongs to one of four API categories as defined by its associated release tag:
 
-- The "extensions" tag indicates that an API is to be included in the `@itwin/core-extension` API. Currently this tag is only useful for APIs in `@itwin/core-frontend` and `@itwin/core-common`; all other packages' APIs can be used directly by extensions. It may become relevant for more packages in the future. A lint rule enforces that the "extensions" tag may only be applied to APIs tagged as "public".
+- `@public` indicates a stable, production-ready API subject to our documented [stability and deprecation policies](../api-support-policies.md).
```
calebmshafer, line 9: again, I'd like to remove "production-ready" and rather state, "indicates a stable API subject to our documented ...."

pmconne: Please specify *why* you would like to make these changes.

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,205 @@
...
+- `@beta` indicates an API undergoing stabilization. Users are encouraged to experiment with it in non-production environments.
+- `@alpha` indicates a highly experimental API that users should not yet experiment with.
```
calebmshafer, line 11: Thoughts on this?

"indicates a highly experimental API that is subject to substantial change or removal. Use with caution."

It does not explcitly say do not use or try it out but it does warn of the potential consequences.

docs/learning/guidelines/release-tags-guidelines.md
calebmshafer: Also, lets add something to the beta about us looking for feedback.

"We encourage user testing and feedback so that we can know that this feature is ready to be marked as stable." (Took this from the Node website on their stability guidelines)

pmconne: I'm fine with rewording stuff but bear in mind as noted at the top, this article is for contributors. api-support-policies.md is for consumers, who would be offering feedback.

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,205 @@
...
+
+- In 4.x (using the `core-backend` example above), we will export the contents of `cross-package.ts` from `core-backend.ts`.
+- In 5.0, we will standardize on ESModules, allowing, e.g., core-backend to import `InternalApi` from core-common using `import { InternalApi } from "@itwin/core-common/lib/InternalApi";`. We can then delete `cross-package.ts`.
+
+Currently, it's possible for an app to (mis)configure their dependencies such that they end up with multiple versions of single core package - and hence, multiple independent copies of each `Symbol` defined in `Symbols.ts`.
+- In 4.x, we use `Symbol.for` to define those symbols so that they are looked up by name in a global registry, preventing duplication.
+- In 5.0, we will prohibit taking a dependency on multiple versions of the same core package (which can lead to all sorts of other problems), and switch to using the `Symbol` constructor, which does not register the symbol in the global registry. This will make it impossible for anyone outside the package to look up the symbol by its name.
```
calebmshafer, line 186: Are we planning to do a runtime validation or what's the proposal here? maybe I missed that in the GH issue.

pmconne: I don't believe anyone gave specifics during our meeting about how they would enforce this, only that it should be enforced. I think there's a run-time enforcement for core-frontend right now that checks for the presence of a property on `window`. @aruniverse @wgoehrig?

docs/learning/guidelines/release-tags-guidelines.md
calebmshafer: As I stated in the other comment, I think it's more important to give the consumer all of the information on what we are expecting out of them and guaranteeing rather than telling them what they should and should not do with it. That is up for them to decide how much risk they'd like to take on overall.

docs/learning/guidelines/release-tags-guidelines.md
calebmshafer: > Changes below, like standardizing on ESM is a dream that I'm not 100% sure we'll be able to realize in time.

Let's add it to the 5.0 list and see where we land with the impact and effort needed.

@pmconne I'm good with all of the changes suggested here other than a few smaller comments I made.

I'd like @grigasp and @GerardasB to comment as well.

docs/learning/guidelines/release-tags-guidelines.md
aruniverse: Correct, we would look to copy the approach from [frontend](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/extension/ExtensionRuntime.ts#L15-L17) into other packages

calebmshafer: @aruniverse sounds good. Let's add that as its own GH issue for 5.0 release.

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,205 @@
...
 
-## API Items
+By default, any API is also usable by extensions, with the exception of APIs exported by `@itwin/core-frontend` and `@itwin/core-common`; these must be explicitly tagged as `@extensions` for inclusion in the `@itwin/core-extension` package's API. The `@extensions` tag may only be applied to `@public` APIs, as verified by a lint rule.
 
-Every API exported by a package must have a release tag.
-Members of classes, interface, namespace, and enums inherit the release tag applied to their containing type unless explicitly overridden.
-The release tag applied to such a member can only be more restrictive than that applied to its container. For example, a "beta" class can contain "alpha" or "internal" properties, but not "public" ones.
+The `@deprecated` tag can be applied to any API to notify users that it may be removed or changed in a breaking way in the future. For `@public` APIs, deprecations must follow our [deprecation policy](../api-support-policies.md#api-deprecation-policy); marking `@beta`, `@alpha`, or `@internal` APIs as deprecated is optional, but recommended for giving users a heads-up.
```
grigasp, line 16: Should this mention the requirement to specify an alternative API to use instead of the deprecated one?

P.S. I find the idea to deprecate with intention to change in a breaking way, rather than remove the API, quite interesting. Technically, that's just removing + adding a different API with the same name, but from consumer's perspective we're suggesting them to switch from using `X` to `Y` and then start using `X` again.

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,205 @@
...
+
+- Defining APIs in terms of `interface`s (behavior and/or data) rather than `class`es (implementation details).
+- Adhering to [SOLID](https://en.wikipedia.org/wiki/SOLID) principles.
+- Writing functions that accept arguments as a single `object` to which new optional arguments can be added in the future.
+- Avoiding exhaustively enumerated types (e.g., string unions or `enum`s) if new types are likely to be added in the future.
+- Using the `_implementationProhibited` symbol for interfaces whose implementations should only be acquired from the package defining the interface, enabling new required fields to be added in the future (see, e.g., [WorkspaceDb]($backend)).
+
+When introducing a brand-new API intended for use outside of the itwinjs-core repository, choose `@public` if the API is relatively simple and unlikely to require breaking changes. Otherwise, choose `@beta` to enable the API to evolve in response to feedback, but keep in mind the eventual goal of promoting the API to `@public`.
+
+`@alpha` should almost never be used, except perhaps when collaborating with other itwinjs-core developers are on highly experimental API.
```
grigasp, line 75:
```suggestion
`@alpha` should almost never be used, except perhaps when collaborating with other itwinjs-core developers on highly experimental API.
```

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,205 @@
...
+`@internal` APIs require special handling, covered in a dedicated section below.
+
+## Internal APIs
+
+`@internal` APIs are a necessary evil. For one reason or another - as described above - an `@internal` API is **not** to be used by - and, ideally, should be completely inaccessible to - code outside of the itwinjs-core repository. Unfortunately, as of iTwin.js 4.x, all `@internal` APIs are technically available for use to any package. The [@itwin/eslint-plugin](https://www.npmjs.com/package/@itwin/eslint-plugin) provides a `no-internal` lint rule that attempts to flag inappropriate usage of `@internal` APIs, but it has had little impact on non-core packages taking dependencies on `@internal` APIs.
+
+A few categories of `@internal` APIs exist:
+
+- Single-package APIs used only within a single package in itwinjs-core - e.g., a function exported by one source file in `@itwin/core-backend` and imported for use in another source file in that same package.
+- Cross-package APIs used by multiple packages in itwinjs-core - e.g., a function exported by `@itwin/core-common` and imported for use in `@itwin/core-common`.
```
grigasp, line 94:
```suggestion
- Cross-package APIs used by multiple packages in itwinjs-core - e.g., a function exported by `@itwin/core-common` and imported for use in `@itwin/core-backend`.
```

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,205 @@
...
+    // implementation goes here
+  }
+}
+
 ```
+
+Add to `src/internal/Symbols.ts` the following, if a `_close` symbol doesn't already exist (keeping the list of symbols sorted alphabetically):
+
+```ts
+export const _close = Symbol.for("close_INTERNAL_ONLY_DO_NOT_USE");
```
grigasp, line 149: Should the name be more specific about the API, e.g. include class or package name? It seems we can easily get symbol name collisions if we only use method name + constant suffix.

pmconne: What problem do you see arising from, say, core-backend and core-common both having a `_close` Symbol with the same description? They will point to the exact same object.

You would never do something like:
```ts
import { _close as closeBackend } from "@itwin/core-backend";
import { _close as closeCommon } from "@itwin/core-common";

export interface Thing {
  [_closeBackend]: string;
  [_closeCommon]: string;
}
```

That would be a compiler error just as trying to define two properties named "close" would.

pmconne: Ok I [take that back](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBBCeBbGBeGBlJAjEAbAdAGYgBOAFAEQAWAprriBQJQDcAUKJLAogEyoZseIqUq16jVmzYBLMFBolCAQ2A0YAFSpyA5jADebGDADaPALoAuOFBK72xs0l5WbdsDoemL16O89sAL7SnNAwUNoe1lq6AoaOPjAUPBQANEbezq7JzmkZAPT5mYjZACYArgAONHnBHOAQeDT4DDpkEbpOJVKhTS0gbR0eXS5SQA).

docs/learning/guidelines/release-tags-guidelines.md
pmconne: Removed release tags definitions from release-tags-guidelines.md, tweaked them in api-support-policies.md. Resolve comment if satisfied.

docs/learning/guidelines/release-tags-guidelines.md
pmconne: Removed release tags definitions from release-tags-guidelines.md, tweaked them in api-support-policies.md. Resolve comment if satisfied.

docs/learning/guidelines/release-tags-guidelines.md
pmconne: Reworded in api-support-policies.md.

docs/learning/guidelines/release-tags-guidelines.md
pmconne: That - along with the requirement to specify the version in which it was deprecated - is documented above the code example.

docs/learning/guidelines/release-tags-guidelines.md
pmconne: Made this change, resolve if satisfied.

docs/learning/guidelines/release-tags-guidelines.md
calebmshafer: that looks good to me. Thanks!

core/backend/src/ChannelControl.ts
```diff
@@ -73,105 +73,3 @@ export namespace ChannelControl {
   export const sharedChannelName = "shared";
 }
 
-/** @internal */
-export class ChannelAdmin implements ChannelControl {
```
calebmshafer, line 77: Are we sure that someone isn't using this? I know we had a plan to deprecate most of these first to avoid unintended breaking changes.

core/backend/src/core-backend.ts
```diff
@@ -76,6 +74,8 @@ export * from "./workspace/WorkspaceEditor";
 export * from "./SqliteChangesetReader";
 export * from "./ChangesetECAdaptor";
 
+export * from "./internal/cross-package";
```
calebmshafer, line 77: This is what will be removed in 5.0 but is here right now to not break potential usage, correct?

calebmshafer:
Looks good overall @pmconne ! Just a few questions on the implementation in 4.x before we make the larger change in 5.0.

core/backend/src/core-backend.ts
pmconne: Yes. Also because we currently don't yet have a solution for knowing whether to import from lib/cjs or lib/esm. For 5.0 we either need to ditch cjs, or find a way to omit the cross-package.ts export from the published packages.

core/backend/src/ChannelControl.ts
pmconne: ChannelControl is the public interface. I can't imagine why or how anyone would use the concrete implementation directly. If they are, they can fix their code. We held up our end of the API contract.

grigasp (approved):

bbastings (approved):

tcobbs-bentley (approved):

aruniverse (approved):
When i reviewed the first time, didnt realize or see you touched core-backend. Im eager to get this into our desktop apps asap to see how much internal usage they've been getting away with

docs/learning/guidelines/release-tags-guidelines.md
```diff
@@ -1,66 +1,205 @@
...
+
+- Defining APIs in terms of `interface`s (behavior and/or data) rather than `class`es (implementation details).
+- Adhering to [SOLID](https://en.wikipedia.org/wiki/SOLID) principles.
+- Writing functions that accept arguments as a single `object` to which new optional arguments can be added in the future.
+- Avoiding exhaustively enumerated types (e.g., string unions or `enum`s) if new types are likely to be added in the future.
+- Using the `_implementationProhibited` symbol for interfaces whose implementations should only be acquired from the package defining the interface, enabling new required fields to be added in the future (see, e.g., [WorkspaceDb]($backend)).
+
+When introducing a brand-new API intended for use outside of the itwinjs-core repository, choose `@public` if the API is relatively simple and unlikely to require breaking changes. Otherwise, choose `@beta` to enable the API to evolve in response to feedback, but keep in mind the eventual goal of promoting the API to `@public`.
```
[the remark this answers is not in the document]
pmconne: I did qualify with "simple and unlikely to require breaking changes". We have years' worth of beta APIs that nobody ever bothered to promote to public, we need to break that bad habit.

PR #6836, opened 2024-06-12
Improve TextBlock line breaking behavior

### LINKED ISSUE #6834: Improve TextBlock line wrapping
When we layout a TextBlock that has a defined width, we must perform word-wrapping on lines exceeding that width. The current implementation does things like putting a sentence's period on the next line following the sentence's last word, and moving the space after a word onto the line below it.

It should wrap the way people expect from most text editing apps like Word, Notepad, even MicroStation's text editor dialog. It should work for all languages.

The current implementation uses JavaScript's `Intl.Segmenter`, which implements the Unicode word breaking algorithm. We should really be using the [Unicode line breaking algorithm](http://www.unicode.org/reports/tr14/), but JavaScript currently lacks an implementation of that. Several npm packages (all of which seem to be [this](https://www.npmjs.com/package/linebreak) or a fork thereof) purport to provide an implementation, but we have not yet evaluated their quality.

I very much would like to avoid rolling our own implementation.

**Additional context**
MicroStation uses a Bentley-developed library that internally uses Windows-specific APIs.

pmconne:
Fixes #6834.

Switch from `Intl.Segmenter`, which only supports word-breaking, to [linebreak](https://www.npmjs.com/package/linebreak), which implements the Unicode line-breaking algorithm.

In the images below, the width of the TextBlock is set to 0.1 to force each line to be as narrow as possible according to the line-breaking rules. The previous approach would have inserted breaks around each punctuation mark and would have produced lines containing only spaces. The new approach respects the difference between a period and a decimal separator; between a thousands separator and a comma; etc.

image

### DIFF
--- a/common/api/core-common.api.md
+++ b/common/api/core-common.api.md
@@ -3168,6 +3168,8 @@ export enum FeatureIndexType {
 // @public
 export class FeatureOverrides implements FeatureAppearanceSource {
     constructor();
+    // @internal
+    addInvisibleElementOverridesToNeverDrawn(): void;
     get alwaysDrawn(): Id64.Uint32Set;
     // @internal
     protected readonly _alwaysDrawn: Id64.Uint32Set;
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -212,6 +212,7 @@ import { OrbitGtBlobProps } from '@itwin/core-common';
 import { OrbitGtDataManager } from '@itwin/core-orbitgt';
 import { OrderedComparator } from '@itwin/core-bentley';
 import { OrderedId64Array } from '@itwin/core-bentley';
+import { OrderedId64Iterable } from '@itwin/core-bentley';
 import { PackedFeatureTable } from '@itwin/core-common';
 import { ParseResults } from '@itwin/appui-abstract';
 import { ParserSpec } from '@itwin/core-quantity';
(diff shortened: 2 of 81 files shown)


pmconne:
@ben-polinsky It's not clear to me why the docs build is even trying to parse linebreak.d.ts or requiring a `@packageDocumentation` tag for it, but it's now producing this:

```
Invalid Link: "[export=]($core-backend)" in file reference/core-backend/all/index.html"
```

ben-polinsky:
> @ben-polinsky It's not clear to me why the docs build is even trying to parse linebreak.d.ts or requiring a `@packageDocumentation` tag for it, but it's now producing this:
> 
> ```
> Invalid Link: "[export=]($core-backend)" in file reference/core-backend/all/index.html"
> ```
> 
> You can compare to wms-capabilities.d.ts, which the docs build doesn't seem to have a problem with.

I think there are sometimes we might want to document declaration files, so I don't want to entirely exclude them, but needing packageDocumentation in modules without public exports has been annoying. I'll take a look.

aruniverse:
Just want to confirm the `grapheme` and `sentence` [granularity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter/Segmenter#granularity) won't suffice

ben-polinsky:
> The difference between this file and wms-capabilities.d.ts is core-frontend's "docs" script tells you to exclude it. Do you want me to tell you to exclude this one too, or await your investigation?
> 
> ```
> "docs": "betools docs --includes=../../generated-docs/extract
>  --json=../../generated-docs/core/core-frontend/file.json
>  --tsIndexFile=./core-frontend.ts --onlyJson
>  --excludes=webgl/**/*,**/map/*.d.ts,**/tile/*.d.ts,**/*-css.ts",
> ```

Why don't you exclude it for now, and I'll raise an issue to look into it.

pmconne:
> Just want to confirm the `grapheme` and `sentence` [granularity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter/Segmenter#granularity) won't suffice

That's right.

aruniverse:
> > Just want to confirm the `grapheme` and `sentence` [granularity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter/Segmenter#granularity) won't suffice
> 
> That's right. The rules for valid line breaks are different than those for word, grapheme, or sentence breaks. The Intl.Segmenter proposal deemed line-breaking as out of scope.

Just going to link this issue, so we can keep track of it, just in case they eventually add support for it. https://github.com/tc39/proposal-intl-segmenter/issues/139

aruniverse (approved):

pmconne:
@ben-polinsky none of my globs appear to be working.

ben-polinsky:
> @ben-polinsky none of my globs appear to be working.

More layers are revealed. Here's where the cli arguments are parsed in `tools/build/scripts/docs.js`:
image

Perhaps try `excludeGlob`?

PR #6851, opened 2024-06-17
Setcategory and lookupitem changes

christophermlawson:
setCategory validation and small refactor of getSchemaItem and lookupSchemaItem


core/ecschema-editing/src/Editing/Formats.ts
```diff
@@ -28,7 +28,10 @@ export class Formats extends SchemaItems {
 
       if (units !== undefined) {
         for (const unit of units) {
-          const unitItem = await this.getSchemaItem<Unit | InvertedUnit>(unit, null);
+          const unitItem =  await this.schemaEditor.schemaContext.getSchemaItem<Unit | InvertedUnit>(unit);
+          if (!unitItem) {
+            throw new SchemaEditingError(ECEditingStatus.SchemaItemNotFoundInContext, new SchemaItemId(SchemaItemType.Unit, unit));
+          }
```
StefanApfel-Bentley, line 34: Actually, I think we could also use the new method of SchemaContextEditor.getSchemaItem here. It uses the same method to get the item and also throws an error if the returned value is undefined with the same erro rmessage.

christophermlawson: The reason I can't is because it could be either unit or invertedUnit.  If you look at the PR diff, in the old code I had a hacky work around where if you passed in null it would bypass the SchemaItemType check.  I decided this was to ugly, so I just replaced it with the underlying editor call that doesn't have validation.  I also experimented with passing in an array of schemaItemTypes, but thought it was to ugly as well.  I'm only do this in two places, both involving the unit/invertedunit scenario.

core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -201,5 +201,35 @@ export class SchemaContextEditor {
...
     return schema;
   }
+
+  public async lookupSchemaItem<T extends SchemaItem>(schemaOrKey: Schema | SchemaKey, schemaItemKey: SchemaItemKey, schemaItemType: SchemaItemType): Promise<T>{
+    let schema: Schema;
+    if (schemaOrKey instanceof SchemaKey) {
+      schema = await this.getSchema(schemaOrKey);
+    } else {
+      schema = schemaOrKey;
+    }
```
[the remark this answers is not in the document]
christophermlawson: Fixed.

core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -201,5 +201,32 @@ export class SchemaContextEditor {
 
     return schema;
   }
+
+  public async lookupSchemaItem<T extends SchemaItem>(schemaOrKey: Schema | SchemaKey, schemaItemKey: SchemaItemKey, schemaItemType: SchemaItemType): Promise<T>{
```
StefanApfel-Bentley, line 205: I wonder if we should mark it internal to make clear this function is not supposed to be called by anything other than the actual editor method.

core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -71,10 +71,10 @@ export class SchemaContextEditor {
    * Helper method for retrieving a schema, previously added, from the SchemaContext.
    * @param schemaKey The SchemaKey identifying the schema.
   */
-  public async getSchema(schemaKey: SchemaKey): Promise<MutableSchema | undefined> {
+  public async getSchema(schemaKey: SchemaKey): Promise<MutableSchema> {
```
StefanApfel-Bentley, line 74: Maybe we should mark this internal

core/ecschema-editing/src/Editing/Editor.ts
christophermlawson: Done

StefanApfel-Bentley (approved):

common/api/ecschema-editing.api.md
```diff
@@ -1936,12 +1936,17 @@ export class SchemaContextEditor {
     finish(): Promise<SchemaContext>;
     // (undocumented)
     readonly formats: Formats;
-    getSchema(schemaKey: SchemaKey): Promise<MutableSchema | undefined>;
+    // @internal
```
pmconne, line 1939: What's the motivation for switching `getSchema` and `lookupSchemaItem` to `@internal`?

pmconne (approved):

common/api/ecschema-editing.api.md
StefanApfel-Bentley: They are not supposed to be called from outside this package

PR #6852, opened 2024-06-17
Add ConflictingLocksError class

nick4598:
Taken from imodels-clients. So that they can surface this error properly to users and and those users can show some useful UI about who has their locks.


common/api/core-common.api.md
```diff
@@ -5533,6 +5547,13 @@ export interface Localization {
     unregisterNamespace(namespace: string): void;
 }
 
+// @public
+export enum LockLevel {
```
pmconne, line 5551: core-backend has a `LockState` with the same semantics you are capturing here. Move it to core-common?

core/common/src/IModelError.ts
```diff
@@ -32,6 +32,43 @@ export class IModelError extends BentleyError {
...
+export enum LockLevel {
+  /** Objects are not locked. This level is used for releasing already acquired locks. */
+  None = "none",
+  /** Multiple Briefcases can acquire a shared lock on the same object. */
+  Shared = "shared",
+  /** Only one Briefcase can acquire an exclusive lock on a given object at a time. */
+  Exclusive = "exclusive"
+}
+
+/** @public */
```
pmconne, line 62: Public APIs must be documented.

core/common/src/IModelError.ts
```diff
@@ -32,6 +32,43 @@ export class IModelError extends BentleyError {
...
+  /**
+     * The level of conflicting lock. Possible values are {@link LockLevel.Shared}, {@link LockLevel.Exclusive}.
+     * See {@link LockLevel}.
+     */
+  lockLevel: LockLevel;
+  /** An array of Briefcase ids that hold this lock. */
+  briefcaseIds: number[];
+}
+
+/** Supported Lock levels.
```
pmconne, line 50: Link to `LockControl` or ConcurrencyControl.md so people know what "Lock" means.

core/common/src/IModelError.ts
```diff
@@ -32,6 +32,43 @@ export class IModelError extends BentleyError {
   }
 }
 
+/** Detailed information about a particular object Lock that is causing the Lock update conflict.
```
pmconne, line 35: Explain what "the Lock update conflict" means.

core/backend/src/BackendHubAccess.ts
```diff
@@ -232,6 +232,7 @@ export interface BackendHubAccess {
   /**
    * acquire one or more locks. Throws if unsuccessful. If *any* lock cannot be obtained, no locks are acquired
    * @internal
+   * @throws ConflictingLocksError
```
pmconne, line 235:
```suggestion
   * @throws ConflictingLocksError if one or more requested locks are held by other briefcases.
```

I assume that's the condition under which this error is thrown.

common/api/core-common.api.md
austeja-bentley: I am not sure we need `LockLevel` here at all. `LockLevel` is defined in iModels clients as a string enum because iModels API returns lock level as a string. Platform already has `LockState` to convey the same information about a lock and `BackendHubAccess` implementation already has code to [convert the string lock level to `LockState`](https://github.com/iTwin/imodels-clients/blob/3c87539effae754c2da3e959e0494c9b3cb838ec/itwin-platform-access/imodels-access-backend/src/interface-adapters/ClientToPlatformAdapter.ts#L63). We can and should do the same conversion when adapting the [error that iModels API client throws](https://github.com/iTwin/imodels-clients/blob/4a3d178a8acc3228e889171e67f5937f4cf3b7a8/clients/imodels-client-authoring/src/base/types/apiEntities/LockErrorInterfaces.ts#L18).

nick4598: Removed LockLevel and moved LockState to core-common from core-backend

core/common/src/IModelError.ts
nick4598: I added a comment to ConflictingLOcksError

core/common/src/IModelError.ts
nick4598: Linked to ConcurrencyControl.md, specifically the section about how to acquire a lock

core/common/src/IModelError.ts
```diff
@@ -32,34 +32,41 @@ export class IModelError extends BentleyError {
...
   /** Id of the object that is causing conflict. */
   objectId: string;
   /**
-     * The level of conflicting lock. Possible values are {@link LockLevel.Shared}, {@link LockLevel.Exclusive}.
-     * See {@link LockLevel}.
+     * The level of conflicting lock. Possible values are {@link LockState.Shared}, {@link LockState.Exclusive}.
+     * See {@link LockState}.
      */
-  lockLevel: LockLevel;
+  lockLevel: LockState;
```
pmconne, line 60:
```suggestion
  state: LockState;
```

Nitpick but "level" doesn't make sense any more and "lock" is redundant.

core/backend/src/BackendHubAccess.ts
```diff
@@ -10,24 +10,11 @@ import { AccessToken, GuidString, Id64String, IModelHubStatus } from "@itwin/cor
...
   IModelError, IModelVersion, LocalDirName, LocalFileName,
+  LockState,
 } from "@itwin/core-common";
 import { CheckpointProps, DownloadRequest, ProgressFunction } from "./CheckpointManager";
 import type { TokenArg } from "./IModelDb";
 
-/** The state of a lock.
- * @public
- */
-export enum LockState {
```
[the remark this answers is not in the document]
nick4598: `export { LockState } from "@itwin/core-common";` to the core-backend.ts file I assume? Just pushed that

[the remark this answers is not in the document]
pmconne: In BackendHubAccess, something like:
```
import { LockState as CommonLockState } from "@itwin/core-common";

/** @deprecated in 4.7 Use [LockState]($common)
 * @public
 */
export LockState = CommonLockState;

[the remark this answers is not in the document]
pmconne: `export type` won't work for `enum` since it's a type and value.
Could try `export { CommonLockState as LockState };`
The point is to deprecate the one in core-backend so we can delete it at some point.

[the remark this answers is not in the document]
nick4598: the deprecated tag wasn't able to show up with export { commonlockstate as Lockstate} so going with duplicate and deprecate approach.

pmconne (approved):

nick4598:

PR #6860, opened 2024-06-18
Fix case-insensitive  base class name comparison

christophermlawson:
Ensure case-insensitive name comparison.


core/ecschema-editing/src/test/Validation/SchemaComparison.test.ts
```diff
@@ -806,6 +806,38 @@ describe("Schema comparison tests for comparing schemas with different names", (
         expect(findDiagnostic(reporter.changes[0].allDiagnostics, SchemaCompareCodes.BaseClassDelta, "SchemaA.testBaseClass", "SchemaB.testBaseClass")).to.be.undefined;
       });
 
+      it.only("should not report baseClass delta when the base class has the same name with different case and is defined in comparable schemas", async () => {
```
StefanApfel-Bentley, line 809: better remove the only flag 😉

christophermlawson: doh!  done.

StefanApfel-Bentley (approved):

PR #6865, opened 2024-06-18
Take benefit of SQLite optimization when statement is cached vs when its used once or few time and finalized quickly

khanaffan:
Allow passing [SQLITE_PREPARE_PERSISTENT](https://www.sqlite.org/c3ref/c_prepare_normalize.html#sqlitepreparepersistent) to prepared statement when it is prepared with intention to be cached for multiple use.

The option avoid using lookaside memory which is use for shorter lived statement. [lookaside memory](https://www.sqlite.org/malloc.html#lookaside) allow faster prepare for single use query.

Also, note that for single-use statements, there is no guarantee of utilizing the lookaside memory pool if their state does not fit into a slot in the pool or if the pool has no empty slots. In such cases, it's allocated on the heap. By default, the lookaside memory is set to 48KB (40 slots of 1200 bytes), which is generally exhausted by cached statements. Allowing lookaside memory to be used by single-use statements can make single-use statement preparation, stepping, and finalization very fast and may not require any malloc at all.

With this change, we can prevent caching unnecessary SQLite statements and utilize single-use statements where appropriate, as they are faster to prepare and finalize compared to those allocated on the heap. The persistent flag will signal the query planner to allocate the statement on the heap when we intent to cache it.
 
Note: SQLite internally runs SQLite statements using either the execute or statement interface, which are also allocated on the heap. This occurs as the application has already exhausted 40 slots in the lookaside memory.

All of the following now use `SQLITE_PREPARE_PERSISTENT` option when preparing sqlite statement.

* SqliteStatementCache
* ECSqlStatementCache
* ConcurrentQuery
* InstanceQuery
* TypeScript Sqlite/ECSql Statement caches

We should also name the `IModelDb.withPreparedStatement()` as `IModelDb.withCachedStatement()` while keep same name for `IModelDb.withStatement()`. Thought this not part of this PR.

imodel-native: https://github.com/iTwin/imodel-native/pull/791


common/api/core-backend.api.md
```diff
@@ -3175,8 +3175,8 @@ export abstract class IModelDb extends IModel {
     get pathName(): LocalFileName;
     performCheckpoint(): void;
     // @internal
-    prepareSqliteStatement(sql: string, logErrors?: boolean): SqliteStatement;
-    prepareStatement(sql: string, logErrors?: boolean): ECSqlStatement;
+    prepareSqliteStatement(sql: string, logErrors?: boolean, persistent?: boolean): SqliteStatement;
+    prepareStatement(sql: string, logErrors?: boolean, persistent?: boolean): ECSqlStatement;
```
pmconne, line 3179: Please prefer not to add new optional arguments. Consider instead adding an overload that takes a single `args` object.

ColinKerr: This combined with @rschili comment on the C++ PR makes a strong case to refactor this to be an options object that is consistent across C++ and TypeScript code

PR #6872, opened 2024-06-19
Fix serialization of KoQs with undefined or empty string unit labels

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/1141


core/ecschema-metadata/src/Metadata/KindOfQuantity.ts
```diff
@@ -166,7 +166,7 @@ export class KindOfQuantity extends SchemaItem {
           if (undefined === unit)
             throw new ECObjectsError(ECObjectsStatus.InvalidECJson, `Unable to locate SchemaItem ${unitOverride[0]}.`);
 
-          unitAndLabels.push([unit, unitOverride[1]]);
+          unitAndLabels.push([unit, (unitOverride[1] === undefined || unitOverride[1] === "") ? unit.label : unitOverride[1]]);
```
ColinKerr, line 169: we should not serialize the units label if there is no override ... we should serialize nothing.  This will end up setting the override to the label on every format.  We don't want that.  The override should only be set if it is explicitly set.

Think about this from a round trip perspective.  If the format is `f.DefaultRealU[u.IN]` and we round trip it to xml it should remain unchanged.  If it becomes `f.DefaultRealU[u.IN|inch]` we have changed the definition just by round tripping.

ColinKerr (approved):

core/ecschema-metadata/src/Metadata/KindOfQuantity.ts
RohitPtnkr1996: I've reverted the changes to KoQs.
And added undefined checks only to the overriden formats.
Resolving this comment as PR has been approved after the fix.

PR #6876, opened 2024-06-19
Include extents of TiledGraphicsProviders in ViewingSpace extents.

### LINKED ISSUE #6873: TiledGraphicsProvider extents no included when computing near and far planes
_Originally posted by @MYDJH in https://github.com/iTwin/itwinjs-core/discussions/6850#discussioncomment-9816364_

See the discussion above for repro steps. Implement a unit test along the same lines.

`ViewingSpace` calls `ViewState.getViewedExtents` which computes the union of all tile trees known to the view. This does not include tile trees known only to the Viewport via `TiledGraphicProvider`s.  It should.

pmconne:
Fixes #6873.

Awaiting ImageTests results.

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -16419,6 +16419,7 @@ export class ViewingSpace {
     getTerrainHeightRange(): Range1d | undefined;
     // (undocumented)
     getViewCorners(): Range3d;
+    readonly getViewedExtents: () => AxisAlignedBox3d;
     // @internal
     static nearScaleLog24: number;
     // @internal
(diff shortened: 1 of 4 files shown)


pmconne:

bbastings (approved):

MarcNeely (approved):
I see that your ImageTest run from your PR also had some missing BG terrain on a few views.  They must just be having some problems serving their data.  Other than that, there are some diffs with the EWR views that I'm not sure about, so I'll let you be the judge on whether those were expected or not by your change.

pmconne:
> I see that your ImageTest run from your PR also had some missing BG terrain on a few views. They must just be having some problems serving their data. Other than that, there are some diffs with the EWR views that I'm not sure about, so I'll let you be the judge on whether those were expected or not by your change.

Thanks for looking. I queued a run against master to try to determine which if any differences might be related to my changes.

pmconne:
Two different ImageTests runs produced two different sets of image differences, mostly related to map/terrain. Two different runs against master produced similar inconsistencies.

This PR does slightly alter the viewed extents for hypermodeling views, hence the slight differences in EWR. Hypermodeling graphics are provided by a TiledGraphicsProvider.

PR #6878, opened 2024-06-20
Various Schema Editing additions

christophermlawson:
- Set schema item name support
- Set description and label on schema
- Added deleteSchemaItem support, in part because it's needed during modifying the SchemaItem name, but also because we already had deleteClass, so this was another gap in the editing API.


core/ecschema-editing/src/Editing/Mutable/MutableSchemaItem.ts
```diff
@@ -9,6 +9,7 @@ import {SchemaItem} from "@itwin/ecschema-metadata";
  * @internal
  */
 export abstract class MutableSchemaItem extends SchemaItem {
+  public abstract override setName(name: string): void;
```
StefanApfel-Bentley, line 12: We should call this method setSchemaItemName to indicate that you'll set the name of a schema item with this method and not the name of a schema.

PR #6882, opened 2024-06-20
Load all subcategories when creating a default view.

hl662:
- Update querySubCategories to fetch all subcategories when no categoryId iterable is passed in.
- Add a call to ViewCreator3d that loads all subcategories of the iModel into the view state's subcategoriescache when a default view is created.


pmconne:
As previously discussed, paging was introduced because we observed truncated responses when too many subcategories existed.
Can you show how that problem has been addressed?

hl662:
> As previously discussed, paging was introduced because we observed truncated responses when too many subcategories existed. Can you show how that problem has been addressed?

Hey @pmconne, is there a way to reproduce a case in QA, where a response is truncated? It doesn't have to be the query subcategories method. If we're concerned about the response being truncated where either some of the subcategories missing, or the response body itself being incomplete, createDefaultView can fallback to using the original implementation, where querySubCategories is paged.

hl662:
I also wanted to ask this because I couldn't find an iModel in QA with enough subcategories returned that would elicit a truncated response

pmconne:
> I couldn't find an iModel with enough subcategories returned that would elicit a truncated response

I no longer have access to an iModel that produced the problem. Try going back in git history to find when paging was added and see if the PR provides any useful information.

pmconne:
> If the response is bad, or the subcategories retrieved are incomplete, the original behavior will be fell back upon

Got a test to prove it works?

hl662:
> Got a test to prove it works?

I've added a test for the behavior we're looking for.

core/backend/src/IModelDb.ts
```diff
@@ -707,16 +707,25 @@ export abstract class IModelDb extends IModel {
   }
 
   /**
-   * queries the BisCore.SubCategory table for the entries that are children of the passed categoryIds
+   * queries the BisCore.SubCategory table for the entries that are children of the passed categoryIds.
+   * If no iterable is passed, all subcategories that are part of parent categories containing an element are returned. If an empty iterable is passed, no subcategories are returned.
```
pmconne, line 711: > If no iterable is passed, all subcategories that are part of parent categories containing an element are returned.

This is confusing and inaccurate. You are only querying spatial categories and 3d elements.

You have changed the default behavior of this general-purpose function to serve one very specific purpose (ViewCreator3d).
Did you consider creating a new `queryAllUsedSpatialSubCategories` function instead?

core/common/src/rpc/IModelReadRpcInterface.ts
```diff
@@ -94,7 +94,7 @@ export abstract class IModelReadRpcInterface extends RpcInterface { // eslint-di
   public async getConnectionProps(_iModelToken: IModelRpcOpenProps): Promise<IModelConnectionProps> { return this.forward(arguments); }
   public async queryRows(_iModelToken: IModelRpcProps, _request: DbQueryRequest): Promise<DbQueryResponse> { return this.forward(arguments); }
   @RpcOperation.allowResponseCaching(RpcResponseCacheControl.Immutable) // eslint-disable-line deprecation/deprecation
-  public async querySubCategories(_iModelToken: IModelRpcProps, _categoryIds: CompressedId64Set): Promise<SubCategoryResultRow[]> { return this.forward(arguments); }
+  public async querySubCategories(_iModelToken: IModelRpcProps, _categoryIds?: CompressedId64Set): Promise<SubCategoryResultRow[]> { return this.forward(arguments); }
```
pmconne, line 97: You cannot change a required argument of an RPC interface to become optional, this will break if a newer frontend calls an older backend. First bullet point in last section of [guidelines](https://github.com/iTwin/itwinjs-core/blob/master/core/common/src/rpc/README-RpcInterfaceVersioning.md).

pmconne (changes requested):
Add a new RPC interface function `queryAllUsedSpatialSubCategories`. Do not alter the existing APIs.

core/backend/src/IModelDb.ts
hl662: Thanks Paul, I wanted to initially explore an approach that would avoid creating a new rpc method, but I didn't pay attention to the guidelines for RPC versioning then. I've added a new `queryAllUsedSpatialSubCategories` instead.

I also originally thought to all parent categories were synonymous to all spatial categories and 3d elements because of an existing [method](https://github.com/iTwin/itwinjs-core/blob/4fc1a660752e5062c483b7280f440d5c380c418c/core/backend/src/CustomViewState3dCreator.ts#L42) we had in core backend. This is the same method used to get the parent categories for the subcategories of a default view. Maybe that method could be renamed to more accurately reflect its behavior

core/common/src/rpc/IModelReadRpcInterface.ts
hl662: I've reverted the change to this existing querySubCategories, and instead added a new rpc method instead.

core/frontend/src/SubCategoriesCache.ts
```diff
@@ -58,14 +58,25 @@ export class SubCategoriesCache {
...
-
     return {
       missingCategoryIds: missing,
       promise,
       cancel: () => request.cancel(),
     };
   }
 
+  /** Load all subcategories that come from used spatial categories of the iModel into the cache. */
+  public async loadAllSubCategories(): Promise<void> {
```
pmconne, line 69:
```suggestion
  public async loadAllUsedSpatialSubCategories(): Promise<void> {
```
Make function name reflect what it actually does.

core/frontend/src/SubCategoriesCache.ts
```diff
@@ -111,13 +123,14 @@ export class SubCategoriesCache {
   /** Exposed strictly for tests.
    * @internal
    */
-  public add(categoryId: string, subCategoryId: string, appearance: SubCategoryAppearance) {
+  public add(categoryId: string, subCategoryId: string, appearance: SubCategoryAppearance, override: boolean) {
```
pmconne, line 126: What motivates the addition of the `override` parameter?

pmconne (approved):

core/frontend/src/SubCategoriesCache.ts
hl662: There's a case shown in [ViewCreatorTest](https://github.com/iTwin/itwinjs-core/blob/d8f249a31d6d2325aa99e8c135d34aec6ac7c43f/full-stack-tests/core/src/frontend/standalone/ViewCreator3d.test.ts#L58), where the subcategories cache could be populated prior to loading the used spatial subcategories (aka prior to creating a default view). I wanted to keep the behavior of the existing test cases the same, and in case that behavior might be in real life

aruniverse (approved):

PR #6888, opened 2024-06-21
Changed Newton seed to find all intersections and perp lines of geometries.

saeeedtorabi:


core/geometry/src/numerics/Polynomials.ts
```diff
@@ -1667,13 +1667,15 @@ export class SmallSystem {
    * @param cy right hand side row 1 coefficient
    * @param cz right hand side row 2 coefficient
    * @param result optional result.
+   * @returns solution vector (u,v,w) or `undefined` is system is singular.
```
dassaf4:
```suggestion
   * @returns solution vector (u,v,w) or `undefined` if system is singular.
```

core/geometry/src/test/curve/CurveCurveCloseApproachXY.test.ts
```diff
@@ -4,6 +4,7 @@
 *--------------------------------------------------------------------------------------------*/
 import { expect } from "chai";
 import { BSplineCurve3d } from "../../bspline/BSplineCurve";
+import { Angle } from "../../core-geometry";
```
dassaf4, line 7:
```suggestion
import { Angle } from "../../geometry3d/Angle";
```

core/geometry/src/test/curve/CurveCurveCloseApproachXY.test.ts
```diff
@@ -885,6 +888,115 @@ describe("CurveCurveCloseApproachXY", () => {
...
+  ) {
+    const existsInIntersectionSetWithTolerance = (set: Set<Point3d>, point: Point3d, tolerance: number = 1e-6): boolean => {
+      for (const pt of set)
+        if (pt.isAlmostEqual(point, tolerance))
+          return true;
+      return false;
+    };
+
+    const existsInApproachSetWithTolerance = (
+      set: Set<[Point3d, Point3d]>, point0: Point3d, point1: Point3d, tolerance: number = 1e-6,
```
dassaf4, line 906: For future reference: you can avoid this lambda and simplify your code by instead using an `OrderedSet` constructed with a toleranced comparison function.  `OrderedSet` will do all this work for you in its `add` method: an element won't get added to the set unless it's different according to your compare function.

core/geometry/src/test/curve/CurveCurveCloseApproachXY.test.ts
```diff
@@ -885,6 +888,115 @@ describe("CurveCurveCloseApproachXY", () => {
...
+              numUniqueIntersections++;
+            }
+          } else { // closest approach between geometries
+            const approachSegment = LineSegment3d.create(start, end);
+            const lenSqr = start.distanceSquaredXY(end);
+            ck.testLE(
+              Math.sqrt(lenSqr), maxDistance, undefined, undefined, "approach length must be smaller than maxDistance",
+            );
+            const vec1 = Vector3d.createStartEnd(start, end);
+            const vec2 = geometryA.fractionToPointAnd2Derivatives(ap.detailA.fraction).vectorU;
```
dassaf4, line 950: For future reference: you could get by with `fractionToPointAndDerivative` or `fractionToPointAndUnitTangent` here.

dassaf4 (approved):
In Newton we trust, but warily. Nice tests!

core/geometry/src/numerics/Polynomials.ts
saeeedtorabi: already fixed

core/geometry/src/test/curve/CurveCurveCloseApproachXY.test.ts
saeeedtorabi: will fix

core/geometry/src/test/curve/CurveCurveCloseApproachXY.test.ts
saeeedtorabi: good to know

core/geometry/src/test/curve/CurveCurveCloseApproachXY.test.ts
saeeedtorabi: I took `fractionToPointAnd2Derivatives` from one of my sandboxes but `fractionToPointAndDerivative` is more readable so I will use it.

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -638,7 +638,8 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
   private solveArcArcNewton(
     curveP: Arc3d, curveQ: Arc3d, reversed: boolean, newtonEvaluator: NewtonEvaluatorRRtoRRD,
   ): void {
-    const seedDelta = 1 / 11; // HEURISTIC: 2 ellipses have up to 8 perpendiculars and up to 4 intersections
+    // HEURISTIC: 2 ellipses have up to 8 perpendiculars and up to 4 intersections
+    const seedDelta = 1 / 10;
```
dassaf4, line 642:
```suggestion
    const seedDelta = 1 / 10; // denominator 9 fails the unit test
```

dassaf4 (approved):

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
saeeedtorabi: Done

PR #6892, opened 2024-06-24
Reenable iOS backend unit tests

tcobbs-bentley:
These were failing during CI but not failing for me locally, so there is a decent chance they will still fail now that they are reenabled, so I added a small amount of extra output printing during the run.


aruniverse (approved):

tcobbs-bentley:
@aruniverse The tests pass now, but I see that my job ran on a Mac that @chuckkir just updated the OS on (which is expected, actually). I don't have any particular reason to believe that the OS update would be fixing the problem, but since I have no idea why it broke for no reason in the first place, I'm concerned that this will fail when run on a different CI machine.

aruniverse:
> I'm concerned that this will fail when run on a different CI machine

You can re-run the tests and demand a specific agent to use. Just modify the pipeline to test

tcobbs-bentley:
Chuck disabled the update Mac, so I'm rerunning on a different one.

PR #6900, opened 2024-06-26
esm support in @itwin/ecschema-metadata

williamkbentley:
esm build

ecschema-metadata esm changes prevents `Uncaught SyntaxError: The requested module '/@fs/.../itwinjs-core/core/ecschema-metadata/lib/cjs/ecschema-metadata.js' does not provide an export named 'InvertedUnit' (at KoqPropertyValueFormatter.ts:12:3)`. When using core in a CoSpace with other esm modules... esm support for Node is now readily available.


aruniverse (approved):
needed as frontend pkgs are now importing this pkg

ColinKerr (approved):

PR #6901, opened 2024-06-26
Within changeSummaryManager, try to rely on having an authClient defined where possible.

nick4598:
This is so potentially long running operations like pullChanges will refresh their token when downloadingChangesets because hubAccess methods call IModelHost.authorizationClient.getAccessToken if we don't pass a token.


aruniverse (approved):

nick4598:

PR #6902, opened 2024-06-27
Fix Incorrect Commit Message when Manually run Finalize Release Workflow

MikeNBentley:
This involves [Issue #1124](https://github.com/iTwin/itwinjs-backlog/issues/1124)

- Problem: Currently, when Finalize Release Workflow is triggered manually, we get incorrect commit message as seen [here](https://github.com/iTwin/itwinjs-core/commit/dd1cbc1ea98f34df1be0788ca4452d1cfab0090a)

- Approach: Instead of getting version from latest commit, extract version from latest git tag is more consistent

- Affected File(s): .github/workflows/automation-scripts/update-changelogs.mjs

### DIFF
--- a/.github/workflows/automation-scripts/update-changelogs.mjs
+++ b/.github/workflows/automation-scripts/update-changelogs.mjs
@@ -94,64 +94,64 @@ if (targetBranch === `origin/${currentBranch}`) {
   console.log(`The current branch is ${currentBranch}, so the target will be ${targetBranch} branch`)
 }
 
-// // copy all changelogs from the current branch to ./temp-incoming-changelogs, the files will be named: package_name_CHANGELOG.json
-// await $`find ./ -type f -name "CHANGELOG.json" -not -path "*/node_modules/*" -exec sh -c 'cp "{}" "./temp-incoming-changelogs/$(echo "{}" | sed "s/^.\\///; s/\\//_/g")"' \\;`;
+// copy all changelogs from the current branch to ./temp-incoming-changelogs, the files will be named: package_name_CHANGELOG.json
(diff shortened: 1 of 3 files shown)


.github/workflows/automation-scripts/update-changelogs.mjs
```diff
@@ -67,7 +67,12 @@ await $`mkdir ${incomingPath}`
 // find the latest release branch, and make that the target for the changelogs
 let targetBranch = await $`git branch -a --list "origin/release/[0-9]*.[0-9]*.x" | tail -n1 | sed 's/  remotes\\///'`;
 let currentBranch = await $`git branch --show-current`;
-let commitMessage = await $`git log --format=%B -n 1`;
+
+// the version in the commit message can be extracted from the latest tag
+const latestTag = await $`git describe --tags $(git rev-list --tags --max-count=1)`;
+const tag = latestTag.stdout.trim();
+const versionMatch = tag.match(/release\/(\d+\.\d+\.\d+)/);
+let commitMessage = versionMatch ? versionMatch[1] : null;
```
DanRod1999, line 75: So if this is the version of the branch based on the git tag, if we first run this workflow on release/4.7.x then it will get the tags of the current version and commit to master the new logs. However would this work if we were to do a back port on an older minor? So for example if this were on 3.x it would get the tag from there, then push changelogs to release/4.7.x, and then it would run again on release/4.7.x and I think get the tags for 4.x. So maybe using tags won't work here, or perhaps we can use a combination of tags and something else to get the message? I only thought of this based on our discussion on triggers yesterday

PR #6904, opened 2024-06-27
EVO Integration

andremig-bentley:
WIP

Added InitializeGeoscienceTiles function allowing you to create a reality model from Seequent EVO data. Added obtainGeoscienceTilesetUrl function to GraphicsProvider which provides the tileset's url from the Seequent EVO service. Added "fdt seequent demo" key in to Display Test App which calls this init function for testing/demo purposes.


pmconne:
These are all temporary APIs for testing/demo purposes, right? If so please mark them `@alpha` to exclude them from the docs.

PR #6906, opened 2024-06-28
Presentation: Add ability to customize subject ID for ruleset embedder

### LINKED ISSUE #6824: Presentation: Add a way to specify parent `Subject` when embedding a ruleset into an iModel
The `RulesetEmbedder` API should be enhanced to allow specifying parent `Subject` for the inserted Ruleset's model.

Yato333:
Closes #6824

### DIFF
--- a/common/changes/@itwin/presentation-backend/presentation-customize-subject-for-ruleset-embedder_2024-06-28-14-20.json
+++ b/common/changes/@itwin/presentation-backend/presentation-customize-subject-for-ruleset-embedder_2024-06-28-14-20.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/presentation-backend",
-      "comment": "Add ability to specify parent subject id for the rulesets.",
+      "comment": "`RulesetEmbedder`: Add ability to specify parent `bis.Subject` for the presentation rulesets' `Subject`.",
       "type": "none"
     }
   ],
--- a/common/changes/@itwin/presentation-backend/presentation-customize-subject-for-ruleset-embedder_2024-06-28-14-20.json
+++ b/common/changes/@itwin/presentation-backend/presentation-customize-subject-for-ruleset-embedder_2024-06-28-14-20.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/presentation-backend",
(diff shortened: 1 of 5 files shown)


presentation/backend/src/presentation-backend/RulesetEmbedder.ts
```diff
@@ -246,36 +251,39 @@ export class RulesetEmbedder {
     return rulesetList;
   }
 
-  private async getOrCreateRulesetModel(callbacks?: InsertCallbacks): Promise<DefinitionModel> {
-    const rulesetModel = this.queryRulesetModel();
+  private async getOrCreateRulesetModel(callbacks?: InsertCallbacks, subjectId?: Id64String): Promise<DefinitionModel> {
+    const subject = subjectId ? this._imodel.elements.getElement<Subject>(subjectId) : await this.getOrCreateSubject(callbacks);
+
+    const rulesetModel = this.queryRulesetModel(subject);
```
grigasp, line 257: The issue description isn't quite clear on whose parent the given `Subject` should be. It should be a parent for the `Subject` that we insert in `insertSubject`. When not specified, the parent should stay as is - the root `Subject`.

Also, I think we should handle the case when an ID of non-existing element is given - we should throw if we can't find a `Subject` wit that ID.

Yato333: Made described changes. 
> we should throw if we can't find a Subject wit that ID

I expect an error to be thrown at `imodel.getElement(parentSubjectId)`

common/changes/@itwin/presentation-backend/presentation-customize-subject-for-ruleset-embedder_2024-06-28-14-20.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-backend",
+      "comment": "Add ability to specify parent subject id for the rulesets.",
```
grigasp, line 5:
```suggestion
      "comment": "`RulesetEmbedder`: Add ability to specify parent `bis.Subject` for the presentation rulesets' `Subject`.",
```

grigasp (approved):

PR #6909, opened 2024-06-28
Add ITwinError, and if an ITwinError is thrown on the backend during an IPC call rethrow it on the frontend

nick4598:
This PR introduces the interface and namespace `ITwinError`.
```
export interface ITwinError {
  /** namespace for the error */
  namespace: string;
  /** unique key for error, within namespace */
  errorKey: string;
  /** explanation of what went wrong. */
  message: string;
  /** stack trace of the error. */
  stack?: string;
  /** metadata about the exception. */
  metadata?: LoggingMetaData;
}
```
Typically the way the interface `ITwinError` should be used is by extending it to add properties that may be important for a specific type of error. Take the `InUseLocksError` as an example below:

```
export interface InUseLocksError extends ITwinError {
  namespace: "IModelAccess";
  errorKey: "InUseLocks";
  inUseLocks: InUseLock[];
}
```
It defines a specific `namespace` and `errorKey` to always be attached to the `InUseLocksError` interface as well as an additional property `InUseLocks` which adds more specific information about the error that occurred. Note that we could also define a specific `message` property if we felt that the InUseLocksError should always have the same exact message.

With the introduction of `InUseLocksError` or any error which extends `ITwinError` also comes 2 new functions added to the `ITwinError` namespace.
1) A type-guard function based off of the namespace and errorKey
```
  export function isInUseLocksError(error: unknown): error is InUseLocksError {
    return isITwinError(error) && error.namespace === "IModelAccess" && error.errorKey === "InUseLocks";
  }
```
2) A function which throws an InUseLocksError
```
export function throwInUseLocksError(inUseLocks: InUseLock[], message?: string, metadata?: LoggingMetaData): never {
    const errorObject = new Error();
    errorObject.name = "InUseLocksError";
    Error.captureStackTrace(errorObject, throwInUseLocksError);
    const lockError: InUseLocksError = {
      namespace: "IModelAccess",
      errorKey: "InUseLocks",
      message: message ?? "One or more objects are already locked by another briefcase.",
      metadata,
      inUseLocks,
    };
    Object.assign(errorObject, lockError);
    throw errorObject;
  }
```
Which results in (The first test in the screenshot is console.logging the stack and the second is throwing the error without catching it):
Pasted image 20241023153041

The way we throw the error could be up for debate and I have tried different ways to do this which we could discuss on this PR if reviewer is unhappy with the current approach. 

This PR also adds support for propagating ITwinErrors to the frontend when they are thrown on a backend in an IPC setup.


core/backend/src/IpcHost.ts
```diff
@@ -166,16 +166,28 @@ export abstract class IpcHandler {
...
+        // Returning the ConflictingLocksError class results in information loss due to the way IPC handles error objects, so we get around this by reconstructing on the frontend using the provided
+        // constructorName and args.
+        if (err instanceof ConflictingLocksError) {
+          const ret: IpcInvokeReturn = {
+            errorConstructorName: err.constructor.name,
+            argsForErrorConstructor: [err.message, err.getMetaData(), err.conflictingLocks],
+          };
+          if (!IpcHost.noStack)
+            ret.stack = BentleyError.getErrorStack(err);
+          return ret;
```
tcobbs-bentley, line 178: Can't these three lines (and the identical ones in the else clause) be pulled out of the if/then/else?

tcobbs-bentley: Never mind; I misread.

kabentley:
I think this approach is heading in the wrong direction. We should not create *any* subclasses of Error, and certainly no new ones. Testing for `typeof` is problematic and unnecessary. Errors should be an interface, as I proposed earlier. 

BTW, I realize terminology is difficult, but "lock conflict" is confusing. Conflicts are what you get if you _don't_ use locks. The error here must be "lock is in use", right?

nick4598:
> I think this approach is heading in the wrong direction. We should not create _any_ subclasses of Error, and certainly no new ones. Testing for `typeof` is problematic and unnecessary. Errors should be an interface, as I proposed earlier.
> 
> BTW, I realize terminology is difficult, but "lock conflict" is confusing. Conflicts are what you get if you _don't_ use locks. The error here must be "lock is in use", right?

Since the conflictinglockserror passes IModelHubStatus.LockOwnedByAnotherBriefcase as its errorNumber, the message ultimately contains "Lock is owned by another briefcase".

Im guessing you meant instanceof? The imodels-clients repo will start throwing this error class during acquireLocks, so I assume that the instanceof check would pass in that case? 

I agree we can't just keep creating subclasses of error, but I assumed switching over to something more uniform would have to be a breaking change for 5.0. I've been thinking about this problem in the context of a singular ITwinError interface as you suggested as well, in the workitem that I linked in the description.

tcobbs-bentley:
Maybe "AlreadyLocked" would be better than "ConflictingLock"?

kabentley:
> Maybe "AlreadyLocked" would be better than "ConflictingLock"?

Yes, that error makes more sense to me. Locks are either available or "in use" (the distinction is that shared locks block exclusive locks). I'd suggest "LockInUse" or "LockNotAvailable". The problem is that Conflicts come during Merge operations, that's why I was confused when I saw this PR.

kabentley:
> I agree we can't just keep creating subclasses of error, but I assumed switching over to something more uniform would have to be a breaking change for 5.0. 

It would be breaking to attempt to change existing subclasses of Error, but not if you're making up a new class. I suggest this be the first use of ITwinError and we **never** support Error subclasses "spanning" processes.

> I've been thinking about this problem in the context of a singular ITwinError interface as you suggested as well, in the workitem that I linked in the description.

kabentley:
> Since the conflictinglockserror passes IModelHubStatus.LockOwnedByAnotherBriefcase as its errorNumber, the message ultimately contains "Lock is owned by another briefcase".

This will only be useful if we give the "Moniker" of the lock holder (so the localized message will be "Element is in use by @joeSmith"), right? Hopefully the IMS team and IModelHub are working on that?

kabentley:
One last point, Locks lock Elements. When we refer to "locks" in error messages, that will confuse users - they only know about Elements. From their perspective the Element is locked.

[N.B. there is some mystery about locking Models - we really obtain the lock on its modeled element. Also the "schema lock" is really a lock on the root subject Element.]

pmconne:
I agree with most of @kabentley's sentiments. I think this discussion belongs on [the issue](https://github.com/iTwin/itwinjs-backlog/issues/1072), but he lacks access to it.

The proposed error interface looks like this:

```ts
export interface ITwinError {
    namespace: string;   // Localization namespace
    errorKey: string; // unique key for error, within namespace
    message: string; // explanation of what went wrong. (Should provide some format for localization, like ITwinLocalization.getLocalizedKeys but the opposite way)
    metadata?: LoggingMetaData;
}
```

This is great for being able to present meaningful information to the user. Software can recognize the unique combination of namespace+errorKey and react to specific errors if it needs to. It will not, however, have access to additional details like the list of unavailable locks.

> This will only be useful if we give the "Moniker" of the lock holder (so the localized message will be "Element is in use by @joesmith"), right?

If the only error details available are encoded into the localizable string then all of the actionable details must be encoded into it. An operation can require multiple locks held by multiple different people, for example. That could make for a long, ugly user-facing error message.

Presumably ITwinError must include an object containing variables to replace placeholders in the localized message. e.g., "Element is in use by ${lockOwner}". Only the frontend can do the substitution because only it knows the user's locale.

Relying on namespace+errorKey as the "error code" may prove fragile. For example, we may need two different errorKeys for the "lock unavailable" error, depending on whether one or multiple locks are not held. Or, we may decide in the future to rephrase an error such that it requires different formatting (e.g., takes a different number of variables), necessitating a new errorKey to avoid breaking frontends that are using an older version of the localizable error message. Suddenly my `catch` statement block stops recognizing the error.

If we want software to be able to do anything with these errors, then we would effectively be making the localization keys an untracked part of our public API.

karolis-zukauskas:
I'll share my thoughts from an email thread that requested `ConflictingLocksError`.

I feel like I want to localize errors on the Frontend. It is because **Frontend is aware of the context** - error messaging depends on the workflow the user was attempting. I mean, the same `ConflictingLocksError` could be translated differently (just random examples):
- "Cannot edit iModel – user X is currently editing”
- "Cannot synchronize changes – user X is currently pushing changes”

Translating errors on the Backend (returning a key) assumes that an error is always directly translated to the same user facing text - I don't think that is true in general. That will end up feeling "dull" and lack flexibility (limiting UX).

nick4598:
>This will only be useful if we give the "Moniker" of the lock holder (so the localized message will be "Element is in use by @joesmith"), right? Hopefully the IMS team and IModelHub are working on that?

From what I understood, since the ConflictingLocks interface is a part of the error message, it would be used to see the userid of who owns the briefcaseId and and then query the hub's 'users' endpoint to get information about the user.

My suggestion on [the issue](https://github.com/iTwin/itwinjs-backlog/issues/1072) is to add another field 'extraData' whose type is driven by the namespace + errorKey (Note I only know how to do this if we combined namespace and errorkey). In the ConflictingLocks case, extraData would be of type ConflictingLocks[]. [See playground.](https://www.typescriptlang.org/play/?#code/FASwdgLgpgTgZgQwMZQAQGED2Y4BsRITgDmAMpkgNaoDewqqmARgFZSECSAJgFyoDOEGCQDc9ARATQ+g4WGJiGTYVDhIE-KN358wAVwC2TWAG0AumIC+wYAHpbqACoBPAA5oDCV65Kpw0eGQoUEhYRBQnNygAWS9acSwcfEISciodDGw8AiJ5NMpzRVQAeQgAC1gAURgYTBg+Gi4ofiRhVyJsGSESSyL7VABBLi5GcthUWFqYAGkoZ1RPbxJ+VAqYYOs7B2jMLhA4ECgRjkcAd3BqqdQITD8wJFw9JtRiKDBYAmuo1Dg6iZq6rNnCEAuE0CdzmBLnUADzTCYAD2gYC4K0oc0wcEi7lirlQAF5UOjnJjsTEvAA+eIMMAIAzNVxBLpyBSofppBD4ABeUhA2FQtPp-EZKHEk0Bcz40xEbIcejAIAAjno0MSfn9xTAADSoc7lcACukMoLiIX8BCvZmiWWI1y4BC0jpgRhY05lKS6t4QXW1eQAOlQAAoAMplTB6XAjVy1ABuIGe-Ew9PVME83t+MFQuAonJAPKdOvw6NQEPAHO5vOwfteEHLeaOQJWTD03rGjG8mH4IGguoQzgAlKaoJIuFIEFb5EUoEiYAgACJjvguHFeEzTCzATb9SoIul2tB6c2vYBIbCCf71EtnC4AmAwgBEiWyKTyFEo-HvVMJdBpRuFTNQe9HGaCAADk-xFKB7y1MVbyBPhHyyZJcjIN8PxghgzQtKAEMcYo52KPgACVKjjLtWzKEB0KHEdF0Au0gjDSNYHpWjJGgsUZ3nOiTBoRhWHYCBuAQgBGaCJCkHDAOnB5DxAGMoJ1ZRDjUDQtFRPgTBEnUACYzFQSwzBg3obH6RxKJWOB5RSflTnDSNUGjZgECYXB5mMVAmgOd4RgNB0-AgSFoUzQVjQiPVKOdU91j9YArPuJ0-H4YLijgZcoDhRFkVRIkMSxdLcQpQNNT4UsoVvHVNXg1Bpn7PhNSSq8gtvOEqR-VB1ggPQYGdTU-SquYCXxQkBucKxTIccyqJ+azErsiMo1qJgXLc1YEAU65bg8rzwCOVAmHmfyvDxOLZr5Z13RRXAoEuTLp2ytE8rJQritvUrr3KqZWv7akGD8LFAyolK0qiV6ph1RCkhyVI0Pvfsfvav6-v2IM+vu2cF0kP1rvkcpUCpAAGBHxCRpHTzARNruxzBiDBup+q4zGEBMAmzD9ZTVHUTRtEHUm-usJHLAmXBNH+oMgdvVL0rp7VANKNZgrh4m+YYcnKaganabRxmxz9JoWjaJ1+yKBhrGsS6uGuy4wf7IA) This would give consumers compile time errors if we decided to change the type ConflictingLocks for some reason.

This extraData would allow consumers to specialize the message they show to users if they so desire. If they saw no need to, they'd be able to use our fallback message.

I don't think this solves the problem Paul brings up of our messages potentially needing to change formats, but maybe we'll see a need for changing message formats less if consumers can make their own message using the data provided in the 'extraData' field?

karolis-zukauskas:
> This extraData would allow consumers to specialize the message they show to users if they so desire.

This will encourage bad patterns of not caring and doing proper localization, because the responsibility won't be on the app developers, e.g. "it will be handled by the platform".

It seems awkward to me that the Backend needs to deal with localization. It doesn't even deliver the localization files, "how will it know" what keys exist? The Backend has no idea of what is happening on the Frontend - how can it provide a UX friendly error message?

Also, what about application defined errors? Can we get a mechanism to pass those to the Frontend?

nick4598:
>This will encourage bad patterns of not caring and doing proper localization, because the responsibility won't be on the app developers, e.g. "it will be handled by the platform".

Isn't providing this extraData putting it on the app developers to create a message as they see fit? Or do you mean having a message to fallback to will encourage bad patterns? I figured it might be wasteful to force every single app to implement their own error message if we can provide a default that works just fine. I imagined app developers noticing a message they don't like and iteratively making custom ones.

kabentley:
I think there is some confusion about my comments. There are 3 separate topics here:

1. Error subclasses. We should not create any (new) subclasses of Error and certainly we should not attempt to propagate subclasses between processes as is introduced in this PR. Instead, add/test fields on an instance of Error.
2. Lock Terminology. The term "conflict" should not be used to refer to lock contention. Also, locks are for Elements, so terminology about locks should refer to the element they gate. Lastly, the "lock in use" exception should contain the identity (in the form of a username/moniker, not email) of the lock holder. Without that, no message to the user will be helpful.
3. The `ITwinError` interface. It should use two strings to differentiate the cause of the error (namespace and errorName) that replaces the problematic `errorNumber` in BentleyError. The combination of those strings may be presumed unique - and that's all they convey. 

I can't see the rest of the proposal for ITwinError, but the comment above about `message` containing localization strings isn't right. The "message" field is for programmers and logs, not for users. However, the exception should contain enough information so that programmers CAN create a localized message in their application (e.g. the "inUseBy" field for the lock contention error). One way to do that would be to use the namespace/errorName fields as localization keys. That would require some extra consideration when the message has to include other data (e.g. the moniker) so localization will have to be case by case. Note that all Localization headaches are the same as if we continued the subclassing-Error approach.

austeja-bentley:
>  Instead, add/test fields on an instance of Error.

In such case, how would we communicate to application developers what error fields to expect and of what type? If we rely on applications discovering Error fields themselves, it sounds like we could not really change them once we add them, since it is possible they added code that relies on those fields and they will not get any compiler errors like they would if we changed names/property types on a class.

nick4598:
>I can't see the rest of the proposal for ITwinError, but the comment above about message containing localization strings isn't right. The "message" field is for programmers and logs, not for users.

Could you give an example of what the message field would look like? When you initially proposed it [here](https://github.com/iTwin/itwinjs-core/pull/6557#issuecomment-2020573699), you mentioned that the message should be an 

>"explanation of what went wrong. (Should provide some format for localization, like ITwinLocalization.getLocalizedKeys but the opposite way)"

I'm having trouble picturing what this means. GetLocalizedKeys does the below:
`getLocalizedKeys("string with %{MyKeys.Key1} followed by %{MyKeys.Key2}!"") // returns "string with First Value followed by Second Value!"`.

kabentley:
> In such case, how would we communicate to application developers what error fields to expect and of what type? If we rely on applications discovering Error fields themselves, it sounds like we could not really change them once we add them, since it is possible they added code that relies on those fields and they will not get any compiler errors like they would if we changed names/property types on a class.

Use an interface for your Error.

core/common/src/ITwinError.ts
```diff
@@ -0,0 +1,92 @@
...
+   * The level of conflicting lock. Possible values are {@link LockState.Shared}, {@link LockState.Exclusive}.
+   * See {@link LockState}.
+   */
+  state: LockState;
+  /** An array of Briefcase ids that hold this lock. */
+  briefcaseIds: number[];
+}
+
+/**
+ * An interface used to describe an error. This error interface should be extended when needing to throw errors with extra properties defined on them. See [[InUseLocksError]] for an example.
```
pmconne, line 22: Please make it much clearer that it describes an error **for a developer/application**. The message is not intended to be displayed to an end user. Be more explicit about how namespace is a uniqueness qualifier for key, which must itself be unique within its namespace. Ideally, provide an example code snippet where some function receives an `ITwinError`, determines that it is an `InUseLocksError`, and accesses the `inUseLocks` property of that derived type.

core/common/src/ITwinError.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Typically this error would be thrown by [LockControl.acquireLocks]($backend) when you are requesting a lock on an element that is already held by another briefcase.
+ * @beta
+*/
+export interface InUseLocksError extends ITwinError {
+  namespace: "IModelAccess";
+  errorKey: "InUseLocks";
+  inUseLocks: InUseLock[];
+}
+
+interface OtherError extends ITwinError {
```
pmconne, line 50: What is the purpose of this?

core/common/src/ITwinError.ts
```diff
@@ -0,0 +1,92 @@
...
+  metadata?: LoggingMetaData;
+}
+
+/**
+ * An error raised when there is a lock conflict detected.
+ * Typically this error would be thrown by [LockControl.acquireLocks]($backend) when you are requesting a lock on an element that is already held by another briefcase.
+ * @beta
+*/
+export interface InUseLocksError extends ITwinError {
+  namespace: "IModelAccess";
```
pmconne, line 45: What level of granularity on namespace is really wanted/useful? Seems like you could use a single namespace for all of itwinjs-core.

pmconne (commented):
Please consider adding a note to `IModelError` discouraging people from creating subclasses of it, instead using `ITwinError`.

full-stack-tests/core/src/frontend/standalone/Error.test.ts
```diff
@@ -0,0 +1,50 @@
...
+          expect(err.stack?.includes("core\\backend") || err.stack?.includes("core/backend"), `Expected ${err.stack} to have mention of 'core\\backend' or 'core/backend'`).to.be.true;
+          expect(err.message).to.equal(message);
+          expect(err.inUseLocks).to.deep.equal(inUseLocks);
+          expect(ITwinError.getMetaData(err)).to.be.undefined; // Currently not propagating metadata.
+        }
+      }
+      expect(caughtError).to.be.true;
+    });
+
+    // it("should throw error on frontend", async () => {
```
pmconne, line 42: Is this commented out because it doesn't pass? Should it?

full-stack-tests/core/src/frontend/standalone/Error.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    beforeEach(async () => {
+      await TestUtility.startFrontend();
+      await TestUtility.initialize(TestUsers.regular);
+    });
+
+    afterEach(async () => {
+      await TestUtility.shutdownFrontend();
+    });
+
+    it("should receive ConflictingLocksError on the frontend", async () => {
```
pmconne, line 21: AFAICT `ConflictingLocksError` is not involved at all.

nick4598: You're right, changed to InUseLocksError

full-stack-tests/core/src/frontend/standalone/Error.test.ts
nick4598: I removed the test. I was using it locally to make sure the error looked the way I wanted it to when it was thrown.

core/common/src/ITwinError.ts
nick4598: Yeah thats a good point. I switched to "itwinjs-core". Now it feels a little weird that the case of the errorKey and the case of the namespace differ, but "ITwinJsCore" feels wrong to me. Not sure if you have an opinion on that.

core/common/src/ITwinError.ts
nick4598: I removed it, wanted to show how adding new Errors would look. I added to the comment above itwinerror that people should add a typeguard and a function to throw the error to the ITwinError namespace when extending ITwinError.

core/common/src/ITwinError.ts
nick4598: I added the example code snippet, and added extra text to the message, namespace and errorkey properties capturing that information.

core/common/src/ITwinError.ts
pmconne: Adding new functions to `ITwinError` namespace isn't going to scale well and won't work for errors that use types unavailable in core-common.
Did you consider either making your `InUseLocksError`-related functions free functions or putting them inside the `InUseLocksError` namespace?

core/common/src/ITwinError.ts
pmconne: I don't feel strongly about it, but if you want to establish consistency consider doing `"in-use-locks"`. People shouldn't be comparing namespace or key themselves anyway, they should be calling a type assertion function that does the comparison for them.

core/common/src/ITwinError.ts
pmconne: Not to nitpick but example code snippets should go in example-code/snippets and be referenced by docs, to ensure they compile now and in the future.

core/common/src/ITwinError.ts
nick4598: I thought adding a new namespace for each new error might be a lot but I guess it doesn't seem that bad. Part of me wanted to fill the ITwinError namespace up when possible (because otherwise it feels quite empty), but then it seems confusing for others trying to extend ITwinError when they dont fit the mould of adding to the ITwinError namespace. Switching to something that anyone could use anywhere does seem better. I switched to use namespace and suggested others do that or create standalone functions for those.

core/common/src/ITwinError.ts
nick4598: I gave it a try. Let me know if it looks off.

core/common/src/ITwinError.ts
nick4598: Switched to "in-use-locks", thanks

pmconne (approved):

core/common/src/ITwinError.ts
```diff
@@ -0,0 +1,96 @@
...
+ * @beta
+*/
+export interface InUseLocksError extends ITwinError {
+  namespace: "itwinjs-core";
+  errorKey: "in-use-locks";
+  inUseLocks: InUseLock[];
+}
+
+/**
+ * Functions for working with InUseLocksError, specifically a type guard and a function to throw the error.
```
pmconne, line 57: FYI, declaration merging will ignore your doc comments on the namespace.

pmconne:
```ts
/** Put all top-level `Thing` documentation here */
interface Thing { }

/** Top-level documentation on namespace gets ignored, don't put anything here */
namespace Thing {
  /** Documentation on members of namespace will be included */
  export function stuff() { }
}
```

nick4598: Thanks, moved comments to the functions themselves

core/common/src/ITwinError.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /** throws an error which passes the [[isInUseLocksError]] type guard function */
+  export function throwInUseLocksError(inUseLocks: InUseLock[], message?: string, metadata?: LoggingMetaData): never {
+    const errorObject = new Error();
+    errorObject.name = "InUseLocksError"; // optional but makes it so that when the error is thrown and not caught we see InUseLocksError: 'message' instead of Error: 'message'
+    Error.captureStackTrace(errorObject, throwInUseLocksError); // optional: whether we want to hide throwInUseLocksError or not from the stack. not super important
+    const lockError: InUseLocksError = {
+      namespace: "itwinjs-core",
+      errorKey: "in-use-locks",
+      message: message ?? "One or more objects are already locked by another briefcase.",
```
pmconne, line 75: You were previously including metadata here, now it goes unused?

nick4598: Ugh, sorry I meant to remove the TODO and removed that as well.. added it back in

PR #6911, opened 2024-06-28
Fix Incorrect Commit Message when Manually run Finalize Release Workflow

MikeNBentley:
This involves [Issue #1124](https://github.com/iTwin/itwinjs-backlog/issues/1124)

- Problem: Currently, when Finalize Release Workflow is triggered manually, we get incorrect commit message as seen [here](https://github.com/iTwin/itwinjs-core/commit/dd1cbc1ea98f34df1be0788ca4452d1cfab0090a)

- Approach: Instead of getting version from latest commit, extract version from latest commit that starts with major.minor.patch (for example "4.7.2" or "4.7.2 Changelogs") is more consistent. This also works when we do a back port on an older minor for example 3.x

- Affected File(s): .github/workflows/automation-scripts/update-changelogs.mjs


DanRod1999 (approved):

DanRod1999:
just to clarify, in a fake example if you had commits like this (newest on top)
new feature
4.2.1
4.2.2

the expect behavior is to take 4.2.1 because it is the most recent right?
This example is deliberately wrong cause this should not be possible, just making sure I understand the regex

MikeNBentley:
> just to clarify, in a fake example if you had commits like this (newest on top) new feature 4.2.1 4.2.2
> 
> the expect behavior is to take 4.2.1 because it is the most recent right? This example is deliberately wrong cause this should not be possible, just making sure I understand the regex

It should work like that. Thanks!

aruniverse (approved):

hl662 (approved):

MichaelSwigerAtBentley (approved):

MikeNBentley:
[@Mergifyio](https://github.com/Mergifyio) backport release/4.7.x

MikeNBentley:
[@Mergifyio](https://github.com/Mergifyio) backport release/4.7.x

DanRod1999:

PR #6913, opened 2024-06-28
Refactor GraphicsProvider and add APIs for creating Seequent EVO tilesets

eringram:
This PR adds a new concept of TilesetCreators and UrlProviders:
- The `UrlProviders/` directory contains files with functions that call external services such as the mesh export service and Seequent EVO visualization service to retrieve urls to 3D tilesets
- The `TilesetCreators/` directory contains files with functions that create/initialize tilesets using these urls

A pair of new files in these directories are GeoscienceUrlProvider.ts and GeoscienceTileset.ts. GeoscienceUrlProvider.ts contains `getGeoscienceTilesetUrl` which retrieves a url to a 3D Tiles tileset.json from Seequent, as described above. GeoscienceTileset.ts contains `createGeoscienceTileset` which uses that url to create and attach a reality model using `Viewport.displayStyle.attachRealityModel`. These APIs are marked as `@alpha` because they are experimental in collaboration with Seequent.

Used to be #6904. This is a copy of that branch with the display test app changes removed.


pmconne:
> singleton class

Why?

danieliborra:
> > singleton class
> 
> Why?

Good point, we can make the functions static so we won't need a singleton.

The idea of the class is to centralise the access and generation of 3d tilesets.

andremig-bentley:
> > Why?
> 
> Good point, we can make the functions static so we won't need a singleton.

The class has been changed to have only static methods, and the singleton has been removed.

extensions/frontend-tiles/src/graphics-provider/GraphicsProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+import { getIModelTilesetUrlFromConnection } from "./url-providers/IModelUrlProvider";
+import { createRealityModelTilesetFromUrl } from "./tileset-creators/RealityModelTileset";
+
+/**
+ * The GraphicsProvider class is responsible for providing graphics-related functionality.
+ * @beta
+ */
+export class GraphicsProvider {
+
+  private constructor() { /* Private constructor to prevent instantiation */ }
```
pmconne, line 17: Classes that can't be instantiated aren't classes, they are namespaces.

Why do you need to put these functions inside of any container?

danieliborra, line 17:
```suggestion
```

danieliborra: In this case, grouping functions improves organization and readability by providing a single point of access, simplifies imports, and enhances maintainability by avoiding global scope pollution and reducing naming conflicts. It also provides a consistent interface for interacting with the functions, making it easier for developers to understand and use the API.

Using a module, namespace or a static class, are valid approaches in typescript for grouping functions.
The advantage of using a class in this case, is that it can be extended in the future with more complex behaviour or state (e.g. custom providers). There is not a single definitive reason for grouping the functions, but in general I think there are many advantages, especially if we compare the approach with having the functions spread on different files.

danieliborra (changes requested):

extensions/frontend-tiles/src/graphics-provider/GraphicsProvider.ts
pmconne: To group or not to group, mostly a philosophical/stylistic argument that I don't want to have (I generally don't prefer it). However, by grouping them you reduce your code's amenity to tree-shaking.

```ts
import { SomeClass } from "whatever";

SomeClass.justTheOneFunctionINeed();
```

vs
```ts
import { justTheOneFunctionINeed } from "whatever";

justTheOneFunctionINeed();
```

The former pulls in the entire object and all the functions of `SomeClass` that you don't call (and their dependencies, and their dependencies, and so on) may not be eliminated when bundling your app. It's also a great way to end up with circular dependencies (leading to run-time errors) because you have to import **everything** instead of just the thing(s) you actually need.

We already have mechanisms for grouping related sets of APIs together in our docs.

pmconne: > start a parallel conversation

Yes.

andremig-bentley: Based on this conversation, `GraphicsProvider.ts` has been removed.

pmconne (approved):
This is fine for demo purposes. Please mark all new APIs as `@alpha`. We will revisit with an eye to more comprehensive integration of the graphics service.

PR #6914, opened 2024-07-01
Reinstate ecschema-metadata clean script

pmconne:
Removed in #6900. @williamkbentley I presume this was unintentional.
Causes `rush clean` to fail.


williamkbentley:
yeah, didn't mean to remove that. whoops

PR #6915, opened 2024-07-01
Add `ListenerType` helper

GerardasB:
This PR adds `ListenerType` helper to retrieve a `BeEvent` listener type. This is useful when an implicit type of `BeEvent` listener can not be used i.e.:

```ts
const ev = new BeEvent<(args: {x: number, y: string }) => void>();

// Implicit type
ev.addListener((args) => {
  args.x === 0;
  args.y === "a";
});

// Type redefinition is lengthy and error-prone, new property additions need to be maintained as well.
const listener1 = (args: {x: number, y: string }) => {
  args.x === 0;
  args.y === "a";
};
ev.addListener(listener1);

// `ListenerType` helper
const listener2: ListenerType<typeof ev> = (args) => {
  args.x === 0;
  args.y === "a";
};
ev.addListener(listener2);
```

This addition facilitates using inline types for event arguments as well, which would allow package maintainers to add new required properties w/o breaking changes if the event type is narrower i.e. https://github.com/iTwin/presentation/blob/b9598f6702bea0f466abae02378a0c958f8e5fe9/packages/shared/src/shared/Event.ts#L10


common/api/core-bentley.api.md
```diff
@@ -1178,6 +1178,13 @@ export namespace JsonUtils {
 // @public
 export type Listener = (...arg: any[]) => void;
 
+// @beta
+export type ListenerType<TEvent extends {
```
pmconne, line 1182: Why make generic instead of using `BeEvent` explicitly? (`BeUiEvent` derives from `BeEvent`.)

common/api/core-bentley.api.md
```diff
@@ -1178,6 +1178,13 @@ export namespace JsonUtils {
 // @public
 export type Listener = (...arg: any[]) => void;
 
+// @beta
```
pmconne, line 1181: Why beta? What might go wrong? Under what criteria will the type become suitable for promoting to public?

common/api/core-bentley.api.md
GerardasB: It's to support narrower event types, like https://github.com/iTwin/presentation/blob/b9598f6702bea0f466abae02378a0c958f8e5fe9/packages/shared/src/shared/Event.ts#L10
Take a look at test named `Narrowed event type`

common/api/core-bentley.api.md
GerardasB: I've marked it as `@beta` with the idea to leave some room in case some unexpected use case comes up. I.e. initially I had a naive implementation, where `BeEvent` was used directly, but it wouldn't work with narrowed event types.
I've bumped the type to `@public` since it is unlikely we'll need to change the structure of the type.

core/bentley/src/BeEvent.ts
```diff
@@ -171,3 +171,24 @@ export class BeEventList<T extends Listener> {
     this._events[name] = undefined;
   }
 }
+
+/**
+ * Retrieves the listener type of a {@link BeEvent}.
+ * @note This can be used with a narrowed event type.
+ *
+ * @example
```
pmconne, line 179:
```suggestion
 * Retrieves the listener type of an event type like [[BeEvent]].
 * For example:
```

Fix doc tags.

core/bentley/src/BeEvent.ts
```diff
@@ -171,3 +171,24 @@ export class BeEventList<T extends Listener> {
...
+ * ```ts
+ * const ev = new BeEvent<(x: number, y: string) => void>();
+ *
+ * // type EvListener = (x: number, y: string) => void
+ * type EvListener = ListenerType<typeof ev>;
```
pmconne, line 184:
```suggestion
 * const event = new BeEvent<(x: number, y: string) => void>();
 * const callback: ListenerType<typeof event> = (x: number, y: string) => {
 *   console.log(`${x}, ${y}`);
 * };
```
Make example clearer.

pmconne (approved):

core/bentley/src/BeEvent.ts
GerardasB: Applied suggestion without explicit arg types since that's the expected usage.

PR #6921, opened 2024-07-02
AccuDrawShortcuts.processShortcutKey is only for testing - clean up callers

bbastings:


pmconne (commented):
You don't appear to have actually removed it.

bbastings:
> You don't appear to have actually removed it.

I removed it from these tools, we still use it for testing in dta with a ToolAdmin override...

pmconne (approved):

PR #6929, opened 2024-07-03
Add public to constant.ts

nick4598:
Without this, I get 

Analysis will use the bundled TypeScript version 5.4.2
Error: src/Constant.ts:13:1 - (ae-missing-release-tag) "Constant" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)

API Extractor completed with errors

locally. 

As well as in my pipeline https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2997999&view=logs&jobId=98bcb9af-6780-51c4-01f3-7d41d7c82e74&j=98bcb9af-6780-51c4-01f3-7d41d7c82e74&t=93a643e6-ba20-5515-aad8-b904d25149fa

### DIFF
--- a/common/changes/@itwin/core-geometry/nick-add-public_2024-07-03-13-55.json
+++ b/common/changes/@itwin/core-geometry/nick-add-public_2024-07-03-13-55.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-geometry",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-geometry"
+}
--- a/core/geometry/src/Constant.ts
+++ b/core/geometry/src/Constant.ts
@@ -9,6 +9,7 @@
 
 /**
  * Commonly used constant values.
+ * @public
  */
 export class Constant {
   /** symbolic name for 1 millimeter:  0.001 meter */


pmconne:
I'd like to understand how that didn't fail extract-api check on the original PR.

dassaf4 (approved):
D'oh! Missed that. Thanks Nick.

nick4598:
> I'd like to understand how that didn't fail extract-api check on the original PR #6922. The API summary files have it marked as `public`.

I was wondering the same thing but I don't know the answer unfortunately.

I know in the past that we've seen some slight extract-api differences across Windows and Linux. I was able to reproduce the error on windows locally, but maybe @saeeedtorabi is not on a windows machine? I think extract-api only runs on Ubuntu 22.04 in the itwinjs-core pipeline. 

That theory falls apart slightly when you consider my native pipeline failed on extract api in all 3 cases in the same way, Build windows itwinjs-core, build linux itwinjs-core, and build darwin-arm64 itwinjs-core.

The build linux itwinjs-core stage appears to run on a machine from iModeljs linux pool, so maybe not Ubuntu 22.04.

 job: build_linux
    displayName: Build linux-x64 iTwinjs-core
    pool:
      name: 'iModeljs Linux'
      demands:
        - 'Agent.OS -equals Linux'

PR #6931, opened 2024-07-03
Use export maps from electron-auth, update build config

hl662:
- Use @itwin/electron-authorization `0.19.5` in our test apps and core full stack tests.
- Use the export maps exposed in @itwin/electron-authorization `package.json`.
- Modify the tsconfig.json of the 3 packages above with the following justifications:
  -  `resolvePackageJsonExports` to use libraries' `exports` field when importing those libraries, else `tsc` will complain it couldn't find the declaration files or the library itself.
  - To use a library's `exports` field, we also need to use a `moduleResolution` of `NodeNext` or `Bundler`, but the latter will enforce the files you use to strictly ESM. The former supports CJS files, which is what the 3 packages above use.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -1708,7 +1708,7 @@ importers:
       '@itwin/editor-backend': workspace:*
       '@itwin/editor-common': workspace:*
       '@itwin/editor-frontend': workspace:*
-      '@itwin/electron-authorization': ^0.14.1
+      '@itwin/electron-authorization': ^0.19.5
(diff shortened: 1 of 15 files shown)


hl662:
@DStradley can you check if the display test apps still working like normal before approval?

DStradley:
> @DStradley can you check if the display test apps display tests are still working like normal before approval?

DTA seems to work.  I have launched a test with DPTA and will report back.

DStradley:
Okay, DPTA ran successfully, I'll approve.

DStradley (approved):

PR #6932, opened 2024-07-03
Add optional nop fallback to frontend tiles and DPTA

mathieu-lemuzic:
Add optional Nop fallback to frontend tiles and DPTA so that we can catch tileset.json parsing errors earlier

We will only fallback to an empty tile tree only is the tileset.json is missing or is an invalid json files, other fallback logic that may happen afterwards are expected to work as usual.

We also exposed Nop fallback options to display test app test configuration options.


pmconne:
Did you run ImageTests? Some 3d models exist that TilesetPublisher intentionally never includes in its tileset. Primarily, 3d models used in drawings and sheets, but technically any 3d model that is marked "private" and/or is not spatially located. BatchedSpatialTileTreeReferences falls back to using V1 tiles for those models. At a glance, it looks like your "nopFallback" will cause nothing to render for those models.

mathieu-lemuzic:
> Did you run ImageTests with this option enabled? Some 3d models exist that TilesetPublisher intentionally never includes in its tileset. Primarily, 3d models used in drawings and sheets, but technically any 3d model that is marked "private" and/or is not spatially located. BatchedSpatialTileTreeReferences falls back to using V1 tiles for those models. At a glance, it looks like your "nopFallback" will cause nothing to render for those models. Is the assumption that none of the test cases you want to run using the no-op fallback will include such models?

I will try it, but this nop option can be set for the whole test set for each single test, so in the worst case we can chose which tests shoud fallback to nop and which one to V1

mathieu-lemuzic:
> Some 3d models exist that TilesetPublisher intentionally never includes in its tileset. Primarily, 3d models used in drawings and sheets, but technically any 3d model that is marked "private" and/or is not spatially located. BatchedSpatialTileTreeReferences falls back to using V1 tiles for those models.

Does that means that there is another fallback to V1 somewhere ? at the moment it seems that I'm only falling back to V1 if the tileset json is missing or if there is deserialization issue when reading the file...

pmconne:
> @pmconne Does that mean that there is another fallback to V1 somewhere else?

Yes, [here](https://github.com/iTwin/itwinjs-core/blob/18ddecdfdc6275bbf01585b49a4f066e9da0ac92/extensions/frontend-tiles/src/BatchedSpatialTileTreeRefs.ts#L32) for specific models not included in the batched tiles.

mathieu-lemuzic:
> > @pmconne Does that mean that there is another fallback to V1 somewhere else?
> 
> Yes, [here](https://github.com/iTwin/itwinjs-core/blob/18ddecdfdc6275bbf01585b49a4f066e9da0ac92/extensions/frontend-tiles/src/BatchedSpatialTileTreeRefs.ts#L32) for specific models not included in the batched tiles.

I don't think this will be covered by the Nop fallback, here we only want to make sure we don't want to publish an invalid tileset json (it happended already)

extensions/frontend-tiles/src/BatchedSpatialTileTreeRefs.ts
```diff
@@ -311,8 +318,26 @@ async function fetchTilesetSpec(iModel: IModelConnection, computeBaseUrl: Comput
   }
 }
 
+class EmptySpatialTileTreeReferences implements SpatialTileTreeReferences {
+  public update(): void {}
+
+  public setDeactivated(_modelIds: string | string[] | undefined, _deactivated: boolean | undefined, _refs: "all" | "animated" | "primary" | "section" | number[]): void {}
```
pmconne, line 324: Just FYI, no action needed - you can omit the args if you don't use them, e.g.,
`public setDeactivated(): void { }`

pmconne (approved):

PR #6933, opened 2024-07-03
Check for the possibility that a TextureId is a number while cloning it.

nick4598:
imodel-native: https://github.com/iTwin/imodel-native/pull/820


nick4598:
Should team responsible for Revit Connector be contacted and ask that they bring TextureIds over as hexadecimal strings?

pmconne:
> Should team responsible for Revit Connector be contacted and ask that they bring TextureIds over as hexadecimal strings?

100%. It is never correct to store element IDs in JSON as numbers, JavaScript represents them as `double`s which lack sufficient precision. @timlawrence-bentley @jffmarker

nick4598:
> Add more commits by pushing to the `[nick/rendermaterial-enhancement](/iTwin/itwinjs-core/tree/nick/rendermaterial-enhancement)` branch on [iTwin/itwinjs-core](/iTwin/itwinjs-core).

I will reach out to them with more details, i.e. iModel this was observed in

core/backend/src/Material.ts
```diff
@@ -198,6 +198,10 @@ export class RenderMaterialElement extends DefinitionElement {
       }
       if (!Id64.isValid(sourceMap.TextureId) || sourceMap.TextureId === undefined)
         continue;
+      // textureId has a chance to be a base 10 number instead of a hexadecimal string. This was observed in an iModel created with the Revit Connector.
```
pmconne, line 201: This is going to cause problems in places other than `onCloned`.

nick4598:
It was recently discovered that another iModel has textureIds as numbers instead of hexadecimal strings, and there was no revit connector involved. The two mentioned to me were the Mstn and Navisworks connector. So one of the two, or both may also be bringing textureIds over as numbers. 

This has caused crashes in transformation, so I've made a change to map these ids to invalid instead of erroring out on them. The reason for the crash is that findtargetElementId expects a string and not a number.

@jffmarker @timlawrence-bentley Could you have someone look at both connectors to see which is bringing textureIds over as numbers?

jffmarker:
> It was recently discovered that another iModel has textureIds as numbers instead of hexadecimal strings, and there was no revit connector involved. The two mentioned to me were the Mstn and Navisworks connector. So one of the two, or both may also be bringing textureIds over as numbers.
> 
> This has caused crashes in transformation, so I've made a change to map these ids to invalid instead of erroring out on them. The reason for the crash is that findtargetElementId expects a string and not a number.
> 
> @jffmarker @timlawrence-bentley Could you have someone look at both connectors to see which is bringing textureIds over as numbers? I don't have access to the iModel this was found in, but we could likely get that access if necessary. Let me know.

From skimming source, I think NWD is the culprit. I'll file a bug on the internal backlog.

core/backend/src/Material.ts
nick4598: Any issues with mapping textures to invalid in the target instead of trying to recover with a potentially inaccurate base 10 number?

jffmarker: I think JS can represent ints from 0..2**53 − 1, so it still covers a lot of ID space accurately. So technically it can be inaccurate, but probably works out in practice "most" of the time. Depends how pedantic you want to be. Since we cannot easily clean up old iModels, that would also compromise visual fidelity for a notable subset.

pmconne: I am not convinced that always returning an invalid Id when we encounter a number is a good idea, vs returning a valid Id. My comment from 3 weeks ago was pointing out that this specific line of code in this specific function is not the only place that "whoops it's a number not a string" will break stuff.

core/backend/src/test/standalone/RenderMaterialElement.test.ts
```diff
@@ -91,6 +91,52 @@ describe("RenderMaterialElement", () => {
       test({}, {});
     });
 
+    it("should convert TextureIds to hexadecimal string when loading element", () => {
```
pmconne, line 94: Please test the following cases:
- TextureId has a non-zero briefcase Id (e.g., 0x0000bc0000001234)
- TextureId has non-zero digits in the upper 32 bits (e.g., 0x000000123456789a)
- Combination of the two above
- TextureId is a numeric value outside the range of a `double` (e.g., 0xfeffff123456789a)
- The map has other properties besides `TextureId`, and those are preserved after native code converts `TextureId` from integer to Id string.

nick4598: Thanks, I believe I've added all of those test cases.

nick4598: Turns out my test is failing on [darwin arm64](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3061563&view=logs&j=71a49582-1cc1-5c39-deb8-41612c740dcb&t=1d054115-f899-5a59-f14a-abc21087f657):
```
  1) RenderMaterialElement
       insert
         should be able to convert TextureId to hexadecimal string outside javascript's double precision range of 2^53:

      AssertionError: expected '0x7fffffffffffffff' to equal '0x8000000000000000'
  ...1 line omitted...

      -0x7fffffffffffffff
      +0x8000000000000000
      
      at Context.<anonymous> (src/test/standalone/RenderMaterialElement.test.ts:108:107)
      at processImmediate (node:internal/timers:476:21)
```

After looking into it with @tcobbs-bentley's help understanding the differences between intel/arm and float-to-int conversion...

#### Note for my test case:
I modified the jsonProperties of a database to have 2^63 represented as decimal for the textureId using a SQLite db browser for this test.

On the C++ side, I am using BeJsValue's GetId64 which use [GetUint64](https://github.com/iTwin/imodel-native/blob/nick/rendermaterial/iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h#L274-L280).

This casts the number to a double then to an int64 then to a uint64. The initial cast to double and calling GetDouble already seems to be a problem if we are trying to represent numbers closer to that 2^63 limit, as that turns them into floating point numbers which as far as I understand can't represent those large numbers with very much precision.

**Question** for @jffmarker**, do we know for sure that connectors can put numbers in the 2^60 range with full precision into the jsonProperties? 

If the answer to that is yes, then we probably need to approach converting those numbers to hexadecimal strings slightly differently than what GetId64 does if possible, I.e. we can't convert to a double. Not sure if we can just call GetUInt64 on this and have things work? I see within BeRapidJsonUtilities we do have [Uint64FromValue](https://github.com/iTwin/imodel-native/blob/nick/rendermaterial/iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeRapidJson.h#L35), but I don't know the full context behind some of the complexity in BeJsValue's GetId64.

#### Slightly less important but more context

Test passed on intel machines but failed on arm machines, due to the way each handles [float-to-int conversions](https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code#Audit-Code-that-Contains-Float-to-Int-Conversions).
**For intel:**
- Converting 2^63 to a signed integer gives us the largest possible negative int which is -(2^63). 
- Converting back to unsigned gives us 2^63 
- Test passes because I expected to see 0x80000... ( which is 2^63).

**For arm:**
- Converting 2^63 to a signed integer gives us the largest possible positive int which is 2^63 - 1. 
- Converting back to unsigned gives us 2^63 -1.
- Test fails because we saw 0x7ffff.... (which is 2^63 -1)

tcobbs-bentley: Just as a slight clarification to the architecture-specific float-to-int behavior.

When converting a double precision (64-bit) floating point value to a signed 64-bit integer on Intel Architecture, it behaves as expected for values that fit in the possible range of a 64-bit signed integer (-2^63 ... +2^63 - 1). However, if the floating point value is _larger_ than 2^63 - 1, it get converted to the minimum possible signed 64-bit integer (-2^63). That's right, all too-large _positive_ values end up being converted to the lowest possible (_negative_) value. I have no idea why this is the case. The Arm architecture instead converts the floating point value to the maximum possible _positive_ signed 64-bit integer (2^63-1). This is much more reasonable behavior, but since it is different than Intel, apparently it ends up causing bugs like this one.

pmconne: Yes, our C++ APIs (JsonCpp and RapidJson) can insert 64-bit signed or unsigned values with full precision. By default BeJsValue uses RapidJson, but often in the context of the addon it will use Napi::Value instead. DgnElement::m_jsonProps uses RapidJson. BeJsValue::operator=(uint64_t and int64_t) call pure virtual methods, but GetUint64 does not - it assumes we're dealing with JavaScript limits on precision (you can also find a comment on operator=(int64_t) claiming that the input must be < Number.MAX_SAFE_INTEGER, but nothing enforces this).

An option would be to add virtual functions to JsValueRef to retrieve 64-bit integer values. JsonCpp and RapidJson already support this. (JsonCpp asserts if the underlying value is a double, not an integer, and the double is larger than MAX_SAFE_INTEGER). NapiValueRef would have to implement these to convert from double to integer. It would return the default value if the double value is outside of the safe range for the requested type (signed/unsigned 64).

Additionally, we ought to consider deleting the bits of the API that permit storing 64-bit integers in JSON in both imodel-native and imodel02. Whether people are accidentally doing that for storing Ids, or just for storing plain integers, it always carries the risk of overflowing the constraints imposed by JavaScript.

core/backend/src/test/standalone/RenderMaterialElement.test.ts
```diff
@@ -86,15 +86,11 @@ describe("RenderMaterialElement", () => {
...
-      test({}, {});
-    });
-
+  describe("read", () => {
     it("should be able to convert TextureId to hexadecimal string outside javascript's double precision range of 2^53", async () => {
       // LargeNumericRenderMaterialTextureId is a database with a single RenderMaterial whose jsonProperties look something like:
-      // {"materialAssets":{"renderMaterial":{"Map":{"Diffuse":{"TextureId":9223372036854775808}}}}}
-      // 9223372036854775808 is equivalent to 2^63 which is equivalent to the hexadecimal string 0x8000000000000000.
+      // {"materialAssets":{"renderMaterial":{"Map":{"Diffuse":{"TextureId":9223372036854775807}}}}}
+      // 9223372036854775807 is equivalent to 2^63 - 1 which is equivalent to the hexadecimal string 0x7fffffffffffffff.
```
tcobbs-bentley, line 93: It seems to me that since this would pass with the broken native code, making this change is actually detrimental to the purpose of this test, which is to verify that large values work.

nick4598: 9223372036854775807 actually doesn't pass on intel without the changes that paul suggested on the C++ side. It would give me back 0x800000 on intel due to the floating point conversion stuff.

tcobbs-bentley: I didn't realize that this now fails on Intel without any native fixes. I'm guessing that means that it fails on both Intel and Arm, and if that is the case, then I agree that this change is good, and you can resolve this conversation.

nick4598: I think it might've passed on ARM, but not 100% sure. I added a couple more large numbers to my test database in hopes thats a bit more thorough.

core/backend/src/test/standalone/RenderMaterialElement.test.ts
nick4598: I believe I have implemented your suggestions on my [native PR](https://github.com/iTwin/imodel-native/pull/820/files).

I also created a backlog item reflecting your last suggestion to make it impossible to store 64-bit integers in JSON: https://github.com/iTwin/itwinjs-backlog/issues/1194

pmconne (approved):

core/backend/src/test/standalone/RenderMaterialElement.test.ts
pmconne: All tests pass (on native PR, which also runs the TS tests in this PR). @tcobbs-bentley please resolve your comment if satisfied.

nick4598:
> > It was recently discovered that another iModel has textureIds as numbers instead of hexadecimal strings, and there was no revit connector involved. The two mentioned to me were the Mstn and Navisworks connector. So one of the two, or both may also be bringing textureIds over as numbers.
> > This has caused crashes in transformation, so I've made a change to map these ids to invalid instead of erroring out on them. The reason for the crash is that findtargetElementId expects a string and not a number.
> > @jffmarker @timlawrence-bentley Could you have someone look at both connectors to see which is bringing textureIds over as numbers? I don't have access to the iModel this was found in, but we could likely get that access if necessary. Let me know.
> 
> From skimming source, I think NWD is the culprit. I'll file a bug on the internal backlog.

Could you send me a link to that bug so I can track it?

aruniverse (approved):

nick4598:

PR #6934, opened 2024-07-04
Allow apps to filter tooltip display.

bbastings:
Also localize strings used for default map layer tooltip.

[Release Tags](https://github.com/iTwin/itwinjs-backlog/issues/1103#issuecomment-2204522281)


pmconne:
Maybe nitpick but `showTooltip` sounds like it actually shows a tooltip. Your new method seems like it should be called `wantShowTooltip`.

pmconne (approved):

PR #6939, opened 2024-07-05
Update Vite to 5.x, accommodate ESM-only exports in Vite

hl662:
Update display-test-app and display-performance-test-app to Vite 5.x.
[Migration guide](https://vitejs.dev/guide/migration) from Vite.
Notable, impacted changes from the guide above involves Vite dropping CJS support of their modules, and only [exporting ESM](https://github.com/vitejs/vite/discussions/13928). 

Solutions are documented in Vite's official migration page, one involved changing package.json's `type`: `module`, one involved changing vite.config.ts to vite.config.mts.

I chose the latter, and here are the following changes:
- esbuild-plugin-external-global is CJS only, so I've removed the use of that, and switched to `rollup-plugin-external-globals`
- `require.resolve` doesn't work by default, the solution I chose is to use `createRequire` from the `module` module. There's also an [experimental](https://nodejs.org/api/esm.html#importmetaresolvespecifier) `import.meta.resolve()`, but it's not a 1-1 implementation of require.resolve(), and the former is a more stable solution. We can look into using import.meta.resolve() in the future when it's not flagged as experimental.


aruniverse (approved):

DStradley:
> Fixed the issue with the app's build - @DStradley can you run the image tests to verify no regressions?

Started ImageTest run with your branch.

DStradley:
ImageTests looked good.

DStradley (approved):

PR #6944, opened 2024-07-08
Add alpha APIs for creating Seequent EVO tilesets

eringram:
This PR adds a new concept of TilesetCreators and UrlProviders:
- The `UrlProviders/` directory contains files with functions that call external services such as the mesh export service and Seequent EVO visualization service to retrieve urls to 3D tilesets
- The `TilesetCreators/` directory contains files with functions that create/initialize tilesets using these urls

A pair of new files in these directories are GeoscienceUrlProvider.ts and GeoscienceTileset.ts. GeoscienceUrlProvider.ts contains `getGeoscienceTilesetUrl` which retrieves a url to a 3D Tiles tileset.json from Seequent, as described above. GeoscienceTileset.ts contains `createGeoscienceTileset` which uses that url to create and attach a reality model using `Viewport.displayStyle.attachRealityModel`.

All of these APIs are marked as `@alpha` because they are experimental in collaboration with Seequent. Only APIs within GeoscienceTileset.ts and GeoscienceUrlProvider.ts are exported from frontend-tiles.ts because they are necessary for the demo Seequent is working on.


eringram:
@pmconne This is our fix for PR #6913 based on your conversation with @danieliborra, everything new is now alpha and the rest of frontend-tiles is not changed

extensions/frontend-tiles/src/graphics-provider/tileset-creators/IModelTileset.ts
```diff
@@ -0,0 +1,104 @@
...
+  maxLevelsToSkip: 4,
+  enableEdges: false,
+  useIndexedDBCache: false,
+};
+
+/**
+ * Initializes the IModelTiles with the provided options.
+ * @alpha
+ */
+export function createIModelTileset(options: CreateIModelTilesetArgs): void {
```
pmconne, line 73: This does not create an iModel tileset. Who is expected to call this function?

extensions/frontend-tiles/src/graphics-provider/tileset-creators/IModelTileset.ts
```diff
@@ -0,0 +1,104 @@
...
+  * Requested tiles will then first be search for in the database, and if not found, fetched as normal.
+  * @internal
+  */
+  useIndexedDBCache?: boolean;
+}
+
+/** Global configuration initialized by [[initializeFrontendTiles]].
+ * @internal
+ */
+export const frontendTilesOptions = {
```
pmconne, line 63: This was copy-pasted from FrontendTiles.ts. I don't know what happens when you export two variables with the same name from a single package, but I suspect it's not what you intended.

extensions/frontend-tiles/src/graphics-provider/tileset-creators/IModelTileset.ts
eringram: It would be called for example in `initializeFrontendTiles` or wherever that function is called. It does the same thing as
`initializeFrontendTiles`, essentially replacing/moving it. I agree that the name is misleading however. Our thought was to introduce this and similar functions for demo purposes or as a "first draft".

extensions/frontend-tiles/src/graphics-provider/tileset-creators/IModelTileset.ts
eringram: This was initially removed from FrontendTiles.ts, but we are now trying to introduce these new providers as alpha without changing how frontend-tiles currently works. So the result is some duplicate code because they have been moved to new files but not removed from the old location.

eringram:
I feel like refactoring this graphics provider/graphics service code and introducing alpha APIs for seequent have become very coupled when they don't need to be. Andre and I have discussed it, and we think it might be cleaner to remove all the new "__UrlProvider" and "__Tileset" files except the geoscience ones, and save those changes for a later conversation. @danieliborra let me know what you think of this option.

extensions/frontend-tiles/src/graphics-provider/tileset-creators/IModelTileset.ts
pmconne: I thought this PR was about enabling EVO. They don't need this, do they? And everyone else will continue to use the existing APIs like `initializeFrontendTiles`, right?

andremig-bentley: That's correct, this was initially part of a proposed refactor from a previous PR which attempted to set up these sorts of functions for future use. There's no immediate use for it, but I believe it is indicative of the way @danieliborra wants this area of the code to look in the future. This is an example of @eringram's comment about the coupling of this overall refactoring and the seequent demo code. If all this PR is attempting to accomplish is to set up the seequent demo, then this is not necessary.

danieliborra:
> I feel like refactoring this graphics provider/graphics service code and introducing alpha APIs for seequent have become very coupled when they don't need to be. Andre and I have discussed it, and we think it might be cleaner to remove all the new "__UrlProvider" and "__Tileset" files except the geoscience ones, and save those changes for a later conversation. @danieliborra let me know what you think of this option.

andremig-bentley:
After the discussions on this PR, we have decided to pare this PR down to only include changes necessary to set up access to the seequent service marked as alpha. The files including this functionality have been moved to within `GraphicsProvider` for clarity, while they are still under `url-providers` and `tileset-creators` as this is the direction we most likely want to go in the future when further refactoring happens. There are no changes to any beta or public API's or their dependencies, only new alpha API's for the seequent connection.

pmconne (approved):

PR #6947, opened 2024-07-09
add `setAlias` api

christophermlawson:


core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -254,5 +254,28 @@ export class SchemaContextEditor {
       });
     schema.setDisplayLabel(label);
   }
+
+  /**
+   * Sets the Schemas alias.
+   * @param schemaKey The SchemaKey identifying the schema.
+   * @param alias The new alias to set.
+   */
+  public async setAlias(schemaKey: SchemaKey, alias: string) {
```
ColinKerr, line 263: Should the editor check other schemas to see if this alias conflicts with one that is already in the set?

ColinKerr (approved):

PR #6948, opened 2024-07-10
Depict null values in array type ECProperty output

6ar8nas:
imodel-native: https://github.com/iTwin/imodel-native/pull/806

Covering the expected behavior with tests, repurposed a test fixture so that it can make room for test cases not limited to struct arrays.
Appending the null behavior documentation with the agreed-on correct behaviors.

Closes iTwin/itwinjs-backlog#1162


docs/learning/ECSQLNullBehaviors.md
```diff
@@ -27,7 +28,11 @@ Attempting to update any of the properties (`intProp`, `arrBoolProp`, `structPro
...
+- A `struct` type property can have all of their children properties set to null. The child property would end up being cleared from the parent object with the remaining properties remaining intact. If all of the struct's children become null, the struct itself is considered to be null. This means that if `structProp` would include `{ doubleProp: 41.0, stringProp: null }`, the string property would be cleared with the doubleProp successfully updated to the expected value.
+- An `array` type (primitive or struct) property can store null values as any other value of the respective type. It would happen to be represented alongside the ordinary properties as null. However, similarly to structs, if all of the array's entries are nulls, the array itself is considered to be null. E.g. updating `arrBoolProp` to `[false, true, null, false]` is a perfectly valid data entry for a boolean type array property.
```
6ar8nas, line 35: We should agree for whether an array [null, null, null] is considered null itself as well or not.

docs/learning/ECSQLNullBehaviors.md
```diff
@@ -27,7 +28,11 @@ Attempting to update any of the properties (`intProp`, `arrBoolProp`, `structPro
 
 ## Setting some complex property children to null
 
-To be added
+This section would cover partially setting a complex property (point2d, struct, array) to null.
+
+- A `point2d` type property must always have all its coordinates present, hence updating a point2d to be partially null is not a valid input and would result in the property not being updated. This is why updating `p2dProp` to `{ x: null, y: 2.5 }` would be invalid and take no effect on the element's property.
+- A `struct` type property can have all of their children properties set to null. The child property would end up being cleared from the parent object with the remaining properties remaining intact. This means that if `structProp` would include `{ doubleProp: 41.0, stringProp: null }`, the string property would be cleared with the doubleProp successfully updated to the expected value.
+- An `array` type (primitive or struct) property can store null values as any other value of the respective type. It would happen to be represented alongside the ordinary properties as null. E.g. updating `arrBoolProp` to `[false, true, null, false]` is a perfectly valid data entry for a boolean type array property.
```
ColinKerr, line 35: This is the behavior that has been controversial.

There are three possible behaviors:
1 - accept the array as input and persist including the nulls.
2 - consider the array invalid if it has any null inputs.
3 - silently prune the null entries in the array, e.g. persist `[false, true, false]`

I do not think it's acceptable to prune the nulls from the array so only options 1 and 2 are OK.

I think #1 is the best option but I could be convinced to accept #2 if we're sure no one will ever want to be able to have a sparse array or have a fixed length array where the use null to indicate that an entry is not yet set.

pmconne: If we permit the possibility of sparse arrays then everywhere we define a TypeScript interface representing any array, we must define it as `Array<T | undefined>~ instead of `T[]`. We certainly don't do that today, and I don't think that makes anyone's life better.

6ar8nas: In theory, these changes only solve the fact that *if* nulls are already present in the DB, we display the data accurately as we have it. The sparse arrays have already been possible, it's just that we have been using the 3rd behavior that Colin has described above for representation, while storing the data into the respective database columns as has been input for us including all of the nulls and that seems like a bug.

> we must define it as `Array<T | undefined>` instead of `T[]`

Is it a must to match all of the TypeScript interfaces to what exactly can be stored on the native side immediately? Given the business justification some time in the future, the only needed change would be switching up the interfaces on the TS side. If the interfaces are kept as is and everyone were strict about using them, this behavior should not ever be hit (though iTwin/itwinjs-backlog#1063 suggests it's not as uncommon) and the nulls wouldn't ever be persisted in the database, making this whole problem redundant.

Also, we can define some generic `type SparseArray<T> = (T | undefined)[];`, which to me wouldn't appear to be that excessive, though I might be not sufficiently aware of what would seem to complicate someone's life with this approach.

6ar8nas:
Closing as there was no development in the last 9 months

PR #6952, opened 2024-07-10
SchemaSync.pull() should not obtain write-lock on container

khanaffan:
Fixes
* `SchemaSync.pull()` should not obtain write-lock on container. Its a read-only container operation.
*  Set timeout to `1 sec` & `30 retries`. The default was `100 ms` & `20 retries` which is too small for `SchemaSync` operation that may take at least 1 sec or more.

The overall algorithm for retry need to be improved. This issue was reported on internal backlog by local sync.

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -4697,6 +4697,10 @@ export namespace SchemaSync {
         openMode?: OpenMode;
         user?: string;
     }, operation: (access: CloudAccess) => Promise<void>) => Promise<void>;
+    const // (undocumented)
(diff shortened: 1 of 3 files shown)


pmconne:
> This issue was reported on internal backlog by local sync.

Please link to the issue.

khanaffan:
> > This issue was reported on internal backlog by local sync.
> 
> Please link to the issue.

https://dev.azure.com/bentleycs/iModelTechnologies/_workitems/edit/1470768

pmconne (approved):

PR #6954, opened 2024-07-11
Fix property type and primitive type conflicts

AureMonke:
…hema merge


StefanApfel-Bentley:
I'll approve that, although some conflict metadata isn't that helpful.
Fixing that should be covered in this backlog item #6967

StefanApfel-Bentley (approved):

PR #6956, opened 2024-07-11
Imdl tiles meshopt compression

mathieu-lemuzic:


core/frontend/src/common/imdl/ImdlSchema.ts
```diff
@@ -327,6 +331,10 @@ export interface ImdlSurface {
   readonly type: SurfaceType;
   /** The 24-bit indices into the [[ImdlVertexTable]] of each triangle's vertex. */
   readonly indices: string;
+  /** Is the index data compressed. */
+  readonly isCompressed: boolean;
+  /** The number of indices, only set if data is compressed. */
+  readonly indexCount: number;
```
pmconne, line 337:
```suggestion
  /** The number of indices, only set if data is compressed. */
  readonly compressedIndexCount?: number;
```

mathieu-lemuzic: done

mathieu-lemuzic:
@pmconne about consuming meshoptimizer module in parse IMDL workers... MeshoptCompression.decodeMeshoptBuffer is asynchronous, but there is currently no async logic in place in the parser, I could make all parent functions async until I reach the code, but I am not quite sure if that's the right call. I also considered not loading the module dynamically (to avoid await), but I guess that would cancel your effort to avoid loading it only if not needed... any suggestions?

pmconne:
> @pmconne about consuming meshoptimizer module in parse IMDL workers... MeshoptCompression.decodeMeshoptBuffer is asynchronous, but there is currently no async logic in place in the parser, I could make all parent functions async until I reach the code, but I am not quite sure if that's the right call. I also considered not loading the module dynamically (to avoid await), but I guess that would cancel your effort to avoid loading it only if not needed... any suggestions?

It's dynamically loaded because we currently only use it for glTF that contains meshopt-encoded data (very rare). If we're going to start meshopt-encoding every iMdl tile, there's no reason to dynamically load it because we'll always need it.

It currently does the decoding on a worker, which is of course async. That is useful for glTF / reality models, whose decoders run on the main thread. You could make the iMdl parser use the synchronous meshopt decoder instead.

mathieu-lemuzic:
> > @pmconne about consuming meshoptimizer module in parse IMDL workers... MeshoptCompression.decodeMeshoptBuffer is asynchronous, but there is currently no async logic in place in the parser, I could make all parent functions async until I reach the code, but I am not quite sure if that's the right call. I also considered not loading the module dynamically (to avoid await), but I guess that would cancel your effort to avoid loading it only if not needed... any suggestions?
> 
> It's dynamically loaded because we currently only use it for glTF that contains meshopt-encoded data (very rare). If we're going to start meshopt-encoding every iMdl tile, there's no reason to dynamically load it because we'll always need it.
> 
> It currently does the decoding on a worker, which is of course async. That is useful for glTF / reality models, whose decoders run on the main thread. You could make the iMdl parser use the synchronous meshopt decoder instead.

If we want to execute the meshopt decoder in IMDL parse workers, we will have to inject the WASM decode code into the worker as follows: [link](https://github.com/zeux/meshoptimizer/blob/897976bb56ba6851d1cdb98cfb2d7851dd428cb6/js/meshopt_decoder.js#L102)
, which is tricky since the WASM code is hidden in a module function scope, so I can't access it even by casting the decoder to any. The other alternative is to spawn a meshopt decoder worker from the parser worker and call it asynchronously, which requires changing all parsing functions to async ... what do you think ?

pmconne:
> If we want to execute the meshopt decoder in IMDL parser workers, we will have to inject the WASM decode code into the worker

You can't call the synchronous functions like `decodeGltfBuffer` from a worker? Why not?

mathieu-lemuzic:
> > If we want to execute the meshopt decoder in IMDL parser workers, we will have to inject the WASM decode code into the worker
> 
> You can't call the synchronous functions like `decodeGltfBuffer` from inside a worker? Why not?

Nevermind, I was creating the decoder outside the worker and then passing it as argument, which explains why the WASM was not initialized in the worker scope, it works fine now :)

pmconne:
> [WIP do not review]

The native PR has merged, is this now awaiting review?

mathieu-lemuzic:
> > [WIP do not review]
> 
> The native PR has merged, is this now awaiting review?

Sorry wrong link, I updated the description

core/frontend/src/common/imdl/ImdlSchema.ts
```diff
@@ -135,6 +135,10 @@ export interface ImdlVertexTable {
   readonly width: number;
   /** The height of the lookup texture. */
   readonly height: number;
+  /** True if vertex data is compressed. */
+  readonly isCompressed: boolean;
+  /** The size of the compressed data */
+  readonly compressedSize: number;
```
pmconne, line 141:
```suggestion
  /** The size of the compressed data, or undefined if the data is not compressed. */
  readonly compressedSize?: number;
```

You don't appear to ever check for `isCompressed`. You check if `compressedSize` is truthy and greater than zero, but your type definition asserts that it is never undefined.

pmconne (approved):

core/frontend/src/common/imdl/ImdlSchema.ts
mathieu-lemuzic: > Your native code also never sets `isCompressed` and leaves `compressedSize` undefined if not compressed. Make the interface match.

fixed

pmconne:
@itwin/core-backend tests fail with the following:
```
⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯

Vitest caught 40 unhandled errors during the test run.
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.

⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯
RangeError: WebAssembly.instantiate(): Out of memory: Cannot allocate Wasm memory for new instance
[...this repeats many times...]
```

core/frontend/src/workers/ImdlParser/Worker.ts
```diff
@@ -25,12 +25,12 @@ export interface ParseImdlWorker {
   /** Parse the binary content into a document.
    * @note The [[Uint8Array]] containing the binary data is transferred from the caller to the worker - it will become unusable for the caller.
    */
-  parse(options: ImdlParserOptions): ImdlModel.Document | ImdlParseError;
+  parse(options: ImdlParserOptions): Promise<any>;
```
pmconne, line 28: Why did you change the return type?

pmconne: Pushed fix.

mathieu-lemuzic:
@pmconne since vitest was handling poorly the loading wasm modules when running multiple tests (could be caused by a chromium bug) it made me think realize that I should perhaps consolidate error handing in case some browsers are not able to load these modules. I ended up reusing the existing meshopt decoder loader and execute it only when compressed geometries are detected, I also added a custom error message in case a browser attempts to read compressed tile if it does not support it. Finally, I added a test to make sure compressed data is loaded properly and another one to make sure the correct error message is returned in case the browser failed to load the decoder, and since only two tests will load the wasm modules, vitest does not complain anymore (previously all tests using the parser would end up loading the meshopt compression module, even if a tile did not contain compressed data). Please let me know if it's okay to add new error codes and messages, thanks.

core/frontend/src/common/imdl/ParseImdlDocument.ts
```diff
@@ -299,11 +304,29 @@ class Parser {
...
+  public async parse(): Promise<Imdl.Document | ImdlParseError> {
     const featureTable = this.parseFeatureTable();
     if (!featureTable)
       return TileReadStatus.InvalidFeatureTable;
 
+    if (this.hasMeshoptCompression()) {
+      if (this._options.meshoptCompressionNotSupported || !MeshoptDecoder.supported)
+        return TileReadStatus.UnsupportedMeshoptCompression;
+
+      if (meshoptDecoderStatus === "uninitialized") {
```
mathieu-lemuzic, line 316: @pmconne I ended up doing things a bit like in MeshoptCompression.ts except that the module is loaded statically instead, do you reckon that we should perhaps make MeshCompression.ts use static import and use it here instead ?

core/common/src/tile/TileIO.ts
```diff
@@ -62,6 +62,7 @@ export enum TileReadStatus {
   InvalidFeatureTable,
   NewerMajorVersion,
   Canceled,
+  UnsupportedMeshoptCompression,
```
mathieu-lemuzic, line 65: @pmconne I added a dedicated error in case meshopt failed to load AND compressed tiles are provided, yea of nay ?

mathieu-lemuzic:
@pmconne I changed this PR a bit since you last approved it, so I reset your approve so you get a chance to have a second look before merging.

core/frontend/vitest.config.mts
```diff
@@ -3,6 +3,7 @@ import { viteStaticCopy } from 'vite-plugin-static-copy';
 
 export default defineConfig({
   test: {
+    dangerouslyIgnoreUnhandledErrors: true,
```
pmconne, line 6: add comment explaining why.

pmconne:
> what about copying the meshopt module code in itwinjs

Don't love copying code from dependencies but no reason not to perform the experiment and see if it would resolve the error.

core/frontend/src/tile/MeshoptDecoder.ts
mathieu-lemuzic, line 1: @pmconne here is the code I copy pasted from [meshoptimizer](https://github.com/zeux/meshoptimizer/blob/master/js/meshopt_decoder.js.) with a few tweaks to make it pass the ts compilation / lint, do you think that we should move this to the existing loader so that gltf also use the same logic ?

pmconne: If we're going to copy-paste it then yes, we don't want multiple copies of it.

mathieu-lemuzic: I did not bother pasting the code for running the decoder in workers, but I know that gltf does the decoding in a single worker and awaits the results, is that okay to run the gltf decoder on the main thread since the results are always awaited?

pmconne: sounds fine.

mathieu-lemuzic: @pmconne test are now passing without errors in the imodel-native-internal, up to you to approve again, I have a few questions that still needs to be resolved above

core/common/src/tile/TileIO.ts
pmconne: Nay, nobody checks the detailed statuses and if we find ourselves in this situation then we messed up.

mathieu-lemuzic: the error code was removed

mathieu-lemuzic:
@pmconne I removed the error code waiting for your approval.

core/frontend/src/test/tile/ImdlParser.test.ts
```diff
@@ -278,4 +278,175 @@ describe("ImdlParser", () => {
...
+
+    const parser = acquireImdlParser({});
+    const document = await parser.parse({
+      data: tileData,
+      batchModelId: "0x123",
+      is3d: true,
+      maxVertexTableSize: 4096,
+    });
+
+    expect(typeof document).toEqual("object");
```
pmconne, line 448: You don't want to verify you correctly decoded the buffer(s)?

pmconne (approved):

aruniverse (approved):
Approving addon changes

core/frontend/src/test/tile/ImdlParser.test.ts
mathieu-lemuzic: I could but I'm using this code in image tests already with compressed tiles, so I did not bother

pmconne: Is it difficult to verify here? If somebody breaks it and you're relying solely on image tests to tell you, you won't find out until PR merges, nightly run fails, somebody investigates and tracks it down to a meshopt problem and finds the offending PR.

mathieu-lemuzic: will add a test in a following PR

PR #6957, opened 2024-07-11
Fix issue where ECChangesetAdaptor fails when ExclusiveRootClassId is NULL for overflow table

khanaffan:
The issue can be reproduced as follows: It is unclear why the user dataset contains `NULL` for `ExclusiveRootClassId`.

1. Import a schema with a class that spans an overflow table.
2. Insert an element for the class.
3. Push the changes to the hub.
4. Update the element.
5. Push the changes to the hub again.
6. Delete the element.
7. Set ExclusiveRootClassId to NULL for the overflow table to simulate the issue.
8. The ECChangesetAdaptor should be able to read changeset 2, in which the element is updated, against the latest iModel where the element is deleted.

Fixes: https://github.com/iTwin/itwinjs-backlog/issues/1168

### DIFF
--- a/common/changes/@itwin/core-backend/affanK-changeset-reader-fix_2024-07-11-13-09.json
+++ b/common/changes/@itwin/core-backend/affanK-changeset-reader-fix_2024-07-11-13-09.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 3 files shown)


pmconne (approved):

nick4598:

PR #6960, opened 2024-07-11
Fix missing login dialog for some ArcGIS services

mdastous-bentley:
Some ArcGIS service will throw an error 403 (You do not have permissions to access this resource or perform this operation),
when an OAuth token is required ( instead of the typical 499 error - TokenRequired).

So we now check for error 403 before triggering the Oauth2 login process.

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -1191,6 +1191,8 @@ export enum ArcGisErrorCode {
     // (undocumented)
     InvalidToken = 498,
     // (undocumented)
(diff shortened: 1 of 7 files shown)


aruniverse (approved):
i assume linter will fail, but lgtm
can prob be cleaned up further so these checks dont get too much more unruly

jjbeckman13:
@aruniverse, can you backport this to 4.7.x? Needed to address service case.

aruniverse:

aruniverse:

PR #1169, opened 2021-04-09
Presentation: Add support for custom property category renderers

roluk:
- Extends `PropertyCategorySpecification` to include a `renderer` field.
- Adds `PropertyCategoryRendererManager` which acts similarly to `PropertyValueRendererManager`, but is specific to property grid categories.
- Integrates custom category rendering logic into `VirtualizedPropertyGrid`.
- Some previously internal property grid state and types are now exposed to custom category renderers.
- Documents custom property category renderer usage and adds entry to `NextVersion.md`

[Parallel PR for the native addon.](https://dev.azure.com/bentleycs/iModelTechnologies/_git/imodel02/pullrequest/155584)

PR #6973, opened 2024-07-15
should LogLevel.Trace set logMask to ALL

jiaruiz717:
fix iTwin/itwinjs-core#1169

set logMask to ALL if CloudSqlite logger category is Trace. Added a test as well.


full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -188,6 +188,47 @@ describe("CloudSqlite", () => {
     container.disconnect({detach: true});
   });
 
+  it("should LogLevel.Trace set LogMask to ALL", async () => {
+    Logger.initializeToConsole();
```
nick4598, line 192: Rather than spying on console.log, You could pass functions like sinon.spy() to Logger.initialize

`const logTrace = sinon.spy()`
`const loginfo = sinon.spy()`
`Logger.initialize(undefined,. undefined, logInfo, logTrace)`

And then I think you could assert those were called with what you expect.

At the end of the test you should probably also call Logger.initialize() (in addition to any cleaning of the spies you might have to do) to clear up any spies setup in the logger.

jiaruiz717: resolved in [commit d3af9c8c5a565d029d9d040de8e17c3384727671](https://github.com/iTwin/itwinjs-core/pull/6973/commits/d3af9c8c5a565d029d9d040de8e17c3384727671)

nick4598 (approved):

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
nick4598: I meant you could just do sinon.spy() with nothing inside, did you find that didn't work? either way what you switched to is fine. Resolving

jiaruiz717: sinon.spy() didn't work. We need a `LogFunction` in `Logger.initialize()`

nick4598: Hmm maybe could cast it `as LogFunction`? good to know though.

jiaruiz717: Yes. But we also need to assert that certain messages get logged, for instance, "dirty block".  With just `sinon.spy()`, we cannot do that.

nick4598: I assumed that you could still do `spy.getCalls().some((call) =>call.args[1].includes("is now dirty block"));` but I could be wrong

PR #6975, opened 2024-07-15
Support defining graphics in workers

pmconne:
## Don't be intimidated by the number of files "changed" by this PR!

A large number of files were simply moved to a different subdirectory, and other files needed their import statements adjusted accordingly.

## Changes

See NextVersion.md, WorkerProxy.md, and Worker.ts for documentation and example code. You can now create graphics on a Worker. This was required for folks trying to produce graphics from large WFS/GeoJSON data sets.

### Promote `WorkerProxy` APIs to `@beta`.

They greatly simplify the task of writing and communicating with Worker scripts.

### Add `GraphicDescriptionBuilder`

It's identical to `GraphicBuilder` except it can execute on a Worker and its `finish` method produces a `GraphicDescription` instead of a `RenderGraphic`. The rest of the `GraphicBuilder` interface was factored out into a `GraphicAssembler` base class from which both `GraphicBuilder` and `GraphicDescriptionBuilder` derive.

### Make most of the geometry-to-graphic APIs eligible for use on a Worker

This required removing dependencies on IModelConnection, IModelApp, RenderSystem, etc. All of the APIs that are safe for use in both Workers and the main thread are in /core/frontend/src/common/. See core/frontend/README.md where I documented the source code layout for this package.

## Follow-up

- Support creating materials and textures on a Worker.


pmconne:
Running into issues with my webpacked worker script. Path extends CurveChain, but is defined before CurveChain. Circular dependencies in core-geometry no doubt.

dassaf4 (approved):
core-geometry looks good. Hard-coding module load order via barrel file to resolve circular dependencies is a neat trick! Lint will presumably catch the handful of files with multiple curves import statements. It is curious that GeometryQuery.ts lives under /curve/. I'll make a note to try moving it where it belongs.

saeeedtorabi (approved):

core/frontend/src/common/render/GraphicAssembler.ts
```diff
@@ -0,0 +1,462 @@
...
+    let curve;
+    if (isEllipse || filled) {
+      curve = Loop.create(ellipse);
+    } else {
+      curve = Path.create(ellipse);
+    }
+
+    if (filled && !isEllipse && !ellipse.sweep.isFullCircle) {
+      const gapSegment: CurvePrimitive = LineSegment3d.create(ellipse.startPoint(), ellipse.endPoint());
+      (gapSegment as any).markerBits = 0x00010000; // Set the CURVE_PRIMITIVE_BIT_GapCurve marker bit
```
aruniverse, line 229: is there any way around the cast to as any?

core/frontend/src/common/render/GraphicAssembler.ts
```diff
@@ -0,0 +1,462 @@
...
+
+  /** Add Frustum sides. Useful for debugging. */
+  public addFrustumSides(frustum: Frustum) {
+    this.addRangeBoxSidesFromCorners(frustum.points);
+  }
+
+  /** Add range edges from corner points */
+  public addRangeBoxFromCorners(p: Point3d[]) {
+    this.addLineString([
+      p[Npc.LeftBottomFront],
```
aruniverse, line 372: should the rest of these be cloned?

aruniverse (approved):

aruniverse (approved):
lgtm; excited to see more usage of workers.

would be good if @itwin/itwinjs-core-display reviewed as well

markschlosseratbentley (approved):
Nice documentation.

core/frontend/src/common/render/GraphicAssembler.ts
pmconne: It's been there since at least 2018, from the folks who initially ported this code from C++. It is the only occurrence of `markerBits` in the entire repository.

pmconne: Removed.

dassaf4:
Any lessons for core-geometry with the webpack circular dependency adventure?

PR #6978, opened 2024-07-16
Support EXT_instance_features 

mathieu-lemuzic:
Add support for EXT_instance_features that allows assigning identifiers to individual instances of a mesh, which can be used to look up per-instance data in the structural metadata property tables

The structural metadata tables and instance properties are parsed in the gltf reader and then stored in the tile tree BatchTableProperties.

Each instance is given an identifier that maps to the instance properties, so two instances with identical properties will be assigned the same identifier.

Other changes:

- Added a DTA tool to load a reality data model from seequent.
- The reality data tile loading logic was also patched to support absolute URL for tile references in the tileset.json
- Return undefined in getGeoscienceTilesetUrl() in case the response from the seequent server was invalid


core/frontend/src/common/gltf/GltfSchema.ts
```diff
@@ -618,7 +631,7 @@ export namespace GltfStructuralMetadata {
     name?: string;
     description?: string;
     version?: string;
-    classes?: Class[];
+    classes?: any;
```
pmconne, line 634: Why did you change this? It was intended to reflect the [spec](https://github.com/CesiumGS/glTF/blob/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata/schema/class.schema.json).

pmconne: Ah, `Schema.classes` is a dictionary not an array per [spec](https://github.com/CesiumGS/glTF/blob/00578e29c554b9081373d0a6f916b48c4a8ce700/extensions/2.0/Vendor/EXT_structural_metadata/schema/schema.schema.json#L35). But the dictionary values are defined as `Class`es.

mathieu-lemuzic: the structural metadata schema does not hold an array of classes:

image

pmconne: [Example](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata#class).

```

  "extensions": {
    "EXT_structural_metadata": {
      "schema": {
        "classes": {
          "tree": {
            "name": "Tree",
            "description": "Woody, perennial plant.",
            "properties": {
              "species": { ... },
              "age": { ... },
              "height": { ... },
              "diameter": { ... }
```

mathieu-lemuzic: Here is what I changed:

```export interface Schema extends GltfProperty {
    id: string;
    name?: string;
    description?: string;
    version?: string;
    classes?: {
      [classId: string]: Class | undefined;
    };
    enums?: Enum[];
  }```

File: common/config/rush/browser-approved-packages.json [hunk 2]
```diff
@@ -272,7 +272,7 @@
     },
     {
       "name": "@itwin/frontend-tiles",
-      "allowedCategories": [ "internal" ]
+      "allowedCategories": [ "frontend", "internal" ]
```
mathieu-lemuzic, line 275: @pmconne I have no idea what this is...

mathieu-lemuzic:
@pmconne I forgot to add the changes that erin and andre did in their branch, which is:

- Support absolute tile URL in the tileset.json
- Add DTA tool to load a reality model from Seequent

Would you mind having a look at these changes ?

core/frontend-devtools/package.json
```diff
@@ -39,6 +39,7 @@
     "@itwin/core-common": "workspace:*",
     "@itwin/core-frontend": "workspace:*",
     "@itwin/core-geometry": "workspace:*",
+    "@itwin/frontend-tiles": "workspace:*",
```
pmconne, line 42: Prefer to put the new tool in display-test-app and not introduce a dependency here on frontend-tiles.

core/frontend/src/tile/BatchedTileIdMap.ts
```diff
@@ -24,20 +24,25 @@ export class BatchedTileIdMap implements BatchTableProperties {
     this._iModel = iModel;
   }
 
-  /** Obtains or allocates the Id64String corresponding to the supplied set of JSON properties. */
-  public getBatchId(properties: any): Id64String {
+  /** Obtains or allocates the Id64String corresponding to the supplied set of JSON properties.
+   * If allowDuplicates is true, entries with the same properties will be allowed to have different Id64Strings, default is false.
+  */
+  public getBatchId(properties: any, allowDuplicates?: boolean): Id64String {
```
pmconne, line 30: My reading of the spec suggests there is no concept of the same "feature" existing within multiple tiles (e.g., crossing tile boundaries or LODs) - you will just have multiple features in each tile that happen to have all the same property values (and may or may not have one or more property values serving as a unique identifier across tiles). So, when the user clicks on a single feature in a tile, that might hilite only a portion of the "real" feature, and the object will no longer be hilited if the user zooms in/out causing that specific tile to fail the LOD test.

Did I overlook something? If not, we're eventually going to have to solve that problem (ideally by amending the spec and/or adding an extension).

pmconne (commented):
No tests?

core/frontend/src/tile/BatchedTileIdMap.ts
mathieu-lemuzic: Correct, selecting a single element by clicking on it will no longer work as you zoom in or out, this information is just not present in the tileset, and even it is present we would need a non-gltf-standard way to assign it as element id, unless we try to match elements with identical properties across tiles, with using the default allowDuplicates=false. Maybe it's not a huge problem to "merge" two instances after all as they are very likely located next to each other anyway, and selection on the UI side will always be driven by their properties.

mathieu-lemuzic:
> No tests?

Added gltf reader tests

core/frontend/src/tile/BatchedTileIdMap.ts
mathieu-lemuzic: Added element id matching for instances with similar properties

core/frontend/src/test/tile/GltfReader.test.ts
```diff
@@ -437,4 +439,635 @@ describe("GltfReader", () => {
...
+              expect(compareArrays(array, expectedValuesMat4[instanceFeatures[entryCount]] )).to.be.true;
+              propertyCount1++;
+            }
+          }
+          expect(propertyCount1 === 6).to.be.true;
+        }
+
+        entryCount ++;
+      }
+      expect(entryCount === 4).to.be.true;
```
pmconne, line 1070: Don't make any changes but in future prefer `expect(entryCount).to.equal(4)` so people don't have to start a debugger or edit the test to find out what `entryCount` actually is when the test fails.

pmconne (approved):

core/frontend/src/test/tile/GltfReader.test.ts
mathieu-lemuzic: Thanks, I knew that there was a reason why this stuff existed, but I could not figure why, now I know 👍

extensions/frontend-tiles/src/GraphicsProvider/url-providers/GeoscienceUrlProvider.ts
```diff
@@ -61,11 +61,17 @@ export async function getGeoscienceTilesetUrl(args: GetGeoscienceTilesetArgs): P
   const baseUrl = "https://351mt.api.integration.seequent.com";
```
aruniverse, line 61: should we be exposing this endpoint?
why not add a temporary arg to GetGeoscienceTilesetArgs to allow you to pass this in?

aruniverse (approved):

extensions/frontend-tiles/src/GraphicsProvider/url-providers/GeoscienceUrlProvider.ts
mathieu-lemuzic: good idea, I did not code that part, but I noticed that the GetGeoscienceTilesetArgs already had an argument to specify a prefix url, so I changed the logic to use it and default to `https://351mt.api.integration.seequent.com` otherwise...

aruniverse: i dont want to expose a default endpoint at all

mathieu-lemuzic: done

PR #6990, opened 2024-07-18
Fix the infinite loop for polyface triangulation

saeeedtorabi:
Several edge cases in the point triangulation algorithm were missing. These cases happen when points are within xy-tolerance of each other, or just beyond, such as when the input contains nearly-vertically separated skirt points in the DTM mesh obtained from a Cesium tile.

Implements https://github.com/iTwin/itwinjs-backlog/issues/1120


core/geometry/src/topology/HalfEdgePointInGraphSearch.ts
```diff
@@ -95,88 +96,109 @@ export class PointSearchContext {
...
-          // See if the other side of the sector is turned even beyond that ...
+          // search direction is exactly opposite this edge
+          // see if the other side of the sector is turned even beyond that
           if (v1 > this._tol) {
             result = searchBase.resetAsFace(outboundEdge, outboundEdge);
             return result;
           }
         }
-      } else if (v0 < -this._tol) {
+      } else if (v0 < -this._tol) { // if sweep from ray to outboundEdge.faceSuccessor is CCW
```
saeeedtorabi, line 156: Comment should be CW because v0 is negative
```suggestion
      } else if (v0 < -this._tol) { // if sweep from ray to outboundEdge.faceSuccessor is CW
```

core/geometry/src/topology/HalfEdgePointInGraphSearch.ts
```diff
@@ -95,88 +96,109 @@ export class PointSearchContext {
...
         if (v1 > this._tol) {
-          // The usual simple entry into an angle < 180
+          // the usual simple entry into an angle < 180
           result = searchBase.resetAsFace(outboundEdge, outboundEdge);
           return result;
         }
+      } else { // if (v0 > this._tol) or if sweep from ray to outboundEdge.faceSuccessor is CW
```
saeeedtorabi, line 162: Comment should be CCW  because v0 is positive
```suggestion
      } else { // if (v0 > this._tol) or if sweep from ray to outboundEdge.faceSuccessor is CCW
```

core/geometry/src/topology/HalfEdgePointInGraphSearch.ts
saeeedtorabi: Addressed

core/geometry/src/topology/HalfEdgePointInGraphSearch.ts
saeeedtorabi: Addressed

dassaf4 (approved):
Self-approving. Will wait for Saeed's review.

core/geometry/src/topology/Triangulation.ts
```diff
@@ -177,25 +177,9 @@ export class Triangulator {
...
-        numInsert++;
-        if (numInsert > 16) {
-          /*
-          context.reset();
-          Triangulator.flipTriangles(context.graph);
-          // console.log (" intermediate flips " + numFlip);
-          */
-          numInsert = 0;
-        }
+        numInsert++; // eslint-disable-line @typescript-eslint/no-unused-vars
```
saeeedtorabi: wht did you keep `numInsert`?

dassaf4: Indispensable for debugging!  Harmless to keep.

bbastings (approved):

Issue #6966, opened 2024-07-12
Documentation Issue: Missing code examples


Wilou1428:
It would be great if there were code examples for different code specs and their different implementations.

Found on page https://www.itwinjs.org/bis/guide/fundamentals/codes - "Element Codes"

ben-polinsky:
@diegoalexdiaz can I assign to you?

diegoalexdiaz:
Sure - I get @Wilou1428's request as we need to introduce examples of CodeSpecs in light of Codes in general. However, I'm not sure what @Wilou1428 means regarding "different implementations".

PR #6991, opened 2024-07-22
Expanding example of CodeSpecs in guide to address doc issue #6966

diegoalexdiaz:
https://github.com/iTwin/itwinjs-core/issues/6966

@Wilou1428, does the additional example proposed in this PR address the doc issue?


docs/bis/guide/fundamentals/codes.md
```diff
@@ -45,13 +45,15 @@ When the identified entity is pure information in nature, it will often have its
 
 A `CodeSpec` (aka **Code Specification**) names and specifies a new *classification* for Codes.
 A `CodeSpec` also captures the rules for encoding and decoding significant business information into and from a Code.
-For example, the Codes for ViewDefinitions and the Codes for Equipment have different *encoding* rules and uniqueness constraints, so would each have a separate `CodeSpec`.
+For example, the Codes for software artifacts, like ViewDefinitions, and the Codes for Equipment, such as Piping Components, have different *encoding* rules and uniqueness constraints, so would each have a separate `CodeSpec`.
```
hl662, line 48:
```suggestion
For example, the Codes for software artifacts, like ViewDefinitions, and the Codes for Equipment, such as Piping Components, have different *encoding* rules and uniqueness constraints, so each would have a separate `CodeSpec`.
```

docs/bis/guide/fundamentals/codes.md
```diff
@@ -45,13 +45,15 @@ When the identified entity is pure information in nature, it will often have its
...
 
-Typically, a `CodeSpec` has a strong correlation with a branch of the Element class hierarchy and is often named after an abstract base class that defines the starting point of that branch. It is common for all subclasses (direct or indirect) descending from that base class to share the same `CodeSpec`.
+For instance, an Industrial Process Plant-focused implementation may require Codes for Piping Components to follow a Specification such as: [plant area code]-[equipment-type code]-[equipment number] (e.g. OIL-AAV-001, which *encodes* a Valve of a particular type "AAV" in an area known as "OIL" at a specific Industrial Plant).
+
+In many cases, a `CodeSpec` has a strong correlation with a branch of the Element class hierarchy and is often named after an abstract base class that defines the starting point of that branch. It is common for all subclasses (direct or indirect) descending from that base class to share the same `CodeSpec`.
 
 For example, the standard `CodeSpec` called "bis:ViewDefinition" helps ensure unique names for all subclasses of the `BisCore:ViewDefinition` Element class.
```
hl662, line 54:
```suggestion
For example, the standard `CodeSpec` for "bis:ViewDefinition" helps ensure unique names for all subclasses of the `BisCore:ViewDefinition` Element class.
```

diegoalexdiaz: Thanks for your suggestions @hl662. In this specific case, the original text is correct. What may be confusing at first is that the names of the standard CodeSpecs match the name of classes in BisCore (e.g. the CodeSpec name for subclasses of bis:ViewDefinition (which could also be referred to as BisCore:ViewDefinition) is "bis:ViewDefinition").

docs/bis/guide/fundamentals/codes.md
```diff
@@ -45,13 +45,15 @@ When the identified entity is pure information in nature, it will often have its
...
 
-Configuration can define the association between Element class and CodeSpec so that a shared service (e.g. 'Identification Code Service') can be used to generate and validate Codes.
+Additional Configuration in an application can define the association between Element class and CodeSpec so that a shared service (e.g. 'Identification Code Service') can be used to generate and validate Codes.
```
hl662, line 56:
```suggestion
Additional configuration in an application can define the association between an Element class and its CodeSpec, allowing a shared service (e.g., 'Identification Code Service') to generate and validate Codes.
```

hl662:
@ColinKerr please review

ColinKerr (approved):

PR #6992, opened 2024-07-22
TS Implementation for Sheet Indexes

Ellord207:
## Context:

My team is working on creating features to facilitate the production of 2d drawings and sheets for users using iTwin.js.  To do this, we need to be able to insert [Sheet Indexes](https://www.itwinjs.org/bis/domains/drawings-sheets/#sheet-index) and their relational elements into the iModel.  Since there are no Sheet Index TS implementations, we took the initiative and started creating them ourselves.
We have created a draft PR for these TS implementations and have a few questions about it.  I feel like larger discussion needs happen.  I will include my questions here and hope the right people will see it.

## Questions:

1.	Code
a.	How to add a BisCodeSpec?
b.	Is it needed for use case?
c.	What Scope Type is need?
2.	Should events such as [Element.onChildUpdate](https://www.itwinjs.org/reference/core-backend/elements/element/onchildupdatestatic/) manage the entryPriority of Sheet Index elements to ensure they are sequential?
3.	How do we handle user defined page labels?  There are scenarios where a user may want page number “%pagenumber%-a” or something like that.  Would DisplayLabel be appropriate?
4.	Do we need to write tests for these types where should they go? (Already started working on them porting them over from our development repo)
a.	Proposed location: core\backend\src\test\sheetIndex\SheetIndex.test.ts
b.	How to run tests for specific files, because running all backend tests takes long.


core/backend/src/test/sheetindex/SheetIndex.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  beforeEach(async () => {
+    const iModelFile: string = IModelTestUtils.prepareOutputFile("IModel", "TestSheetIndex.bim");
+
+    const iModelDb = SnapshotDb.createEmpty(iModelFile, { rootSubject: { name: "SheetIndex" } });
+    await ExtensiveTestScenario.prepareDb(iModelDb);
+    ExtensiveTestScenario.populateDb(iModelDb);
+    iModel = iModelDb;
+
+    await insertCodeSpec(iModel);
```
Ellord207, line 76: Is this necessary?  This steps feel like it should already have happened.  Is this what was meant by, "Ensure the new codes are included in pre-existing iModels"?

core/backend/src/test/sheetindex/SheetIndex.test.ts
```diff
@@ -0,0 +1,447 @@
...
+      folder!.parent = new SheetIndexOwnsEntries(sheetIndex2Id);
+      folder!.update();
+
+      // Need to register class scheme?
+      // const relationship = iModel.relationships.tryGetInstance(SheetIndexOwnsEntries.classFullName, { sourceId: sheetIndexId, targetId: folderId });
+      const parentRel12 = iModel.relationships.tryGetInstanceProps(ElementOwnsChildElements.classFullName, { sourceId: sheetIndex1Id, targetId: folderId });
+      expect(parentRel12).to.be.undefined;
+
+      // Need to register class scheme?
+      // const relationship = iModel.relationships.tryGetInstance(SheetIndexOwnsEntries.classFullName, { sourceId: sheetIndexId, targetId: folderId });
```
Ellord207, line 167: This isn't necessary, but is there a way to get an instance of a RelatedElement directly?  When I try to run this method I get the following error:

```
Bad Arg: class BisCore,ElementOwnsChildElements has no superclass
      at Function.generateClassForEntity (src\ClassRegistry.ts:103:13)
```
I assume this is because [NavigationRelationships are not Entities](https://github.com/iTwin/itwinjs-core/blob/c9d607822932cdc444205221c48f58fab3bab7d0/core/backend/src/NavigationRelationship.ts#L9).

I also found the [Relationship](https://www.itwinjs.org/reference/core-backend/relationships/relationship/) class.  Should `SheetIndexOwnsEntries` and `SheetIndexReferenceRefersToSheetIndex` be extending this class instead?

Ellord207: More specifically, [ElementRefersToElement](https://www.itwinjs.org/reference/core-backend/relationships/elementreferstoelements/) seems like an appropriate base class for `SheetIndexReferenceRefersToSheetIndex`.

diegoalexdiaz: Your assumption is correct: A Navigation-relationship is just backing-up a "foreign-key" column that is stored along an Element, so there's no separate "row" in the database for such relationship. 

On the other hand, a link-table relationship (like ElementRefersToElement subclasses) lead to separate rows stored in their own table, different from its source or target Element. 

Picking between the two during schema design primarily lays on questions about multiplicity (1-n vs. n-n) as well as data-ownership. 

It seems the tryGetInstance API only works with the latter kind because it returns an instance of "Relationship" - which expects it to be link-table. Maybe such API could be refactored to work more seamlessly with both kinds of relationships. ECSQL itself is cleaner on that regard... you could write a query based on any kind of relationship the same way (e.g. SELECT SourceECInstanceId, TargetECInstanceId FROM <any relationship - nav-property or link-table>).

So, finally answering your question, I don't think there's an API way to get to an instance of a RelatedElement directly, associated via a nav-property.

core/backend/src/test/sheetindex/SheetIndex.test.ts
diegoalexdiaz: I'll defer to @aruniverse to investigate and confirm this question.

core/backend/src/test/sheetindex/SheetIndex.test.ts
Ellord207: Thanks for the informative answer.  I will leave the implementation as is, and will clean up these comments in the test file.

core/backend/src/SheetIndex.ts
```diff
@@ -0,0 +1,326 @@
+import { BisCodeSpec, Code, CodeScopeProps, CodeSpec, ElementProps, EntityReferenceSet, IModelError, RelatedElementProps, SheetIndexEntryProps, SheetIndexFolderProps, SheetIndexReferenceProps, SheetReferenceProps } from "@itwin/core-common";
+import { InformationReferenceElement, Sheet } from "./Element";
+import { IModelDb } from "./IModelDb";
+import { Id64String, IModelStatus } from "@itwin/core-bentley";
+import { SheetIndexFolderOwnsEntries, SheetIndexOwnsEntries, SheetIndexReferenceRefersToSheetIndex, SheetReferenceRefersToSheet } from "./NavigationRelationship";
+
+/**
+ * A bis:InformationReferenceElement used to organize bis:Sheet instances into a hierarchy with the assistance
```
pmconne, line 8: Please link to the Sheet Index section of docs/bis/domains/drawings-sheets.md where people can learn what a sheet index is.

core/backend/src/SheetIndex.ts
```diff
@@ -0,0 +1,326 @@
...
+   * @param iModelDb The iModel
+   * @param sheetIndexModel The [[SheetIndexModel]]
+   * @param parentId The [[SheetIndex]] or [[SheetIndexFolder]] that is parent to this Entry
+   * @param name The name of the SheetIndexReference
+   * @param priority The priority of the SheetIndexReference
+   * @param sheetIndexId The Sheet Index referenced by the SheetIndexReference
+   * @returns The newly constructed SheetIndexReference element.
+   * @throws [[IModelError]] if unable to create the element.
+   */
+  public static create(
```
pmconne, line 197: Too many args here and in `insert`, same for `SheetIndexFolder` methods. Consolidate into an "arguments" object so people can't screw up the order of the 4 string arguments.

core/backend/src/test/sheetindex/SheetIndex.test.ts
pmconne: Built-in code specs are registered in native code when creating a new DgnDb by [DgnDb::CreateCodeSpecs](https://github.com/iTwin/imodel-native/blob/f6244d40dd38146680bbdd0f7fc711b6a3f7f02c/iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp#L287).
That function notably does not register sheet index-related code specs.

core/backend/src/Element.ts
```diff
@@ -1339,6 +1339,15 @@ export class SpatialLocationPartition extends InformationPartitionElement {
   public static override get className(): string { return "SpatialLocationPartition"; }
 }
 
+/** A SheetIndexPartition element establishes a *Sheet Index* modeling perspective for its parent Subject
+ * A SheetIndexPartition is always sub-modeled by a SheetIndexModel.
+ * @see [[SheetIndexModel]]
```
pmconne, line 1344:
```suggestion
/** A SheetIndexPartition element establishes a [[SheetIndex]] modeling perspective for its parent [[Subject]].
 * A SheetIndexPartition is always sub-modeled by a [[SheetIndexModel]].
```

core/backend/src/Model.ts
```diff
@@ -437,6 +437,35 @@ export abstract class GroupInformationModel extends InformationModel {
   public static override get className(): string { return "GroupInformationModel"; }
 }
 
+/** A container for persisting SheetIndexEntry and SheetIndex elements.
+ * @see [[SheetIndexPartition]]
```
pmconne, line 441:
```suggestion
/** A sub-model of a [[SheetIndexPartition]] serving as a container for persisting [[SheetIndexEntry]] and [[SheetIndex]] elements.
```

core/backend/src/NavigationRelationship.ts
```diff
@@ -104,6 +104,26 @@ export class FolderContainsRepositories extends ElementOwnsChildElements {
   }
 }
 
+/** Relates a [[SheetIndexFolder]] and [[Entries]] that it owns.
```
pmconne, line 107:
```suggestion
/** Relates a [[SheetIndexFolder]] and [[SheetIndexEntry]]'s that it owns.
```

core/backend/src/NavigationRelationship.ts
```diff
@@ -104,6 +104,26 @@ export class FolderContainsRepositories extends ElementOwnsChildElements {
...
+ * @beta
+ */
+export class SheetIndexFolderOwnsEntries extends ElementOwnsChildElements {
+  public static override classFullName = "BisCore:SheetIndexFolderOwnsEntries";
+  public constructor(parentId: Id64String, relClassName: string = SheetIndexFolderOwnsEntries.classFullName) {
+    super(parentId, relClassName );
+  }
+}
+
+/** Relates a [[SheetIndex]] and [[Entries]] that it owns.
```
pmconne, line 117:
```suggestion
/** Relates a [[SheetIndex]] and [SheetIndexEntry]]'s that it owns.
```

core/backend/src/test/sheetindex/SheetIndex.test.ts
pmconne: CodeSpecs use server-issued Ids. There is [some attempt](https://github.com/iTwin/itwinjs-core/blob/10d8e52c046250c8650688a84bf9a196c978b722/core/backend/src/CodeSpecs.ts#L102) to prevent conflicts between briefcases. I think you will need to make your `createCode` functions ensure that they register the relevant CodeSpec if not already present in the iModel. You may wish to add documentation to your new BisCodeSpec members indicating that they are only created in an iModel on first use.

@khanaffan should we consider a DgnDb schema upgrade that adds these code specs if not present?

core/backend/src/SheetIndex.ts
Ellord207: Added a URL to the doc site.  I wasn't sure if there's a better way to do that or not.  If there is, I couldn't find an example of it in the repo.

ben-polinsky: You can use `$docs` to reference the base url without hard coding a specific environment. Something like: `[SheetIndex]($docs/bis/domains/drawings-sheets#sheet-index)`

Ellord207: Thanks for the suggestion.  Updated

core/backend/src/SheetIndex.ts
Ellord207: Refactored

core/backend/src/test/sheetindex/SheetIndex.test.ts
Ellord207: Updating the iModel indirectly seems like a side effect we should avoid, but I'm not sure if there's an instance where a user would want to create a code an not immediately update the iModel with the next element.  Plus, I can't think of a more elegant way.

Implemented as suggested

khanaffan: Yes we can.

khanaffan: Though feel like we should have different approach where its like define in custom attribute in BisCore. So we do not have to add them via profile version change instead biscore can be modified to add builtin code that will exist in seed file or upgraded file. So we can tie it to biscore rather then runtime.

pmconne: @khanaffan can you ping me sometime today to clarify your suggestion please?

Ellord207: After discussing with @khanaffan, it was decided that the suggest implementation should be reverted.

Currently, these added classes are all in beta and will main be used by my team.  Once they have matured a bit and are ready to be released, the new Code Specs will be added to the profile and method of updating existing iModels will the added then.  In the mean time, the exact way to add them will be discussed more.

ben-polinsky (approved):

PR #6997, opened 2024-07-23
implementing schema changeinfo, changeoptions, etc...

christophermlawson:


core/ecschema-editing/src/Editing/ChangeInfo/ChangeInfo.ts
```diff
@@ -0,0 +1,273 @@
...
+  readonly schemaItemType?: string;
+}
+
+export interface ISchemaEditChangeInfo {
+  readonly contextEditor: SchemaContextEditor;
+  readonly changeOptions: ChangeOptions;
+  readonly editType: SchemaEditType;
+  readonly schemaItemType: SchemaItemType;
+  set sequence(value: number)
+  get sequence(): number;
```
StefanApfel-Bentley, line 23: if we both have a public getter and setter, we could just defined it as property.
`sequence: number`

core/ecschema-editing/src/Editing/ChangeInfo/ChangeInfo.ts
```diff
@@ -0,0 +1,273 @@
...
+//   readonly removeCustomAttribute?: boolean;
+// }
+
+// export type RemoveReferenceCallback = (changeInfo: RemoveReferenceChangeInfo) => IRemoveReferenceChangeOptions;
+
+// export interface ISetBaseClassChangeOptions {
+//   readonly cancel?: boolean;
+// }
+
+// export type SetBaseClassCallback = (changeInfo: SetBaseClassChangeInfo) => ISetBaseClassChangeOptions;
```
StefanApfel-Bentley, line 273: I assume, that will reactivated for other cases than SetName?

core/ecschema-editing/src/Editing/SchmaEditType.ts
```diff
@@ -0,0 +1,60 @@
+export enum SchemaEditType {
+  CreateSchemaItemFailed = "CreateSchemaItemFailed",
```
StefanApfel-Bentley, line 2: I'm not sure about this one. Actually, everything in this enum is actually representing an editing operation except of this, which is an editing result (something failed) and therefore a but misleading in this enum I think.

Maybe it make sense to have another enum SchemaEditResultType which derives from SchemaEditType and extend it by the failure type. For the editing calls, we use SchemaEditTypes, in the Errors, the ResultType.
```
export enum SchemaEditResultType = SchemaEditType | {
  CreateSchemaItemFailed = "CreateSchemaItemFailed",
}
```

core/ecschema-editing/src/test/Editing/Entities.test.ts
```diff
@@ -261,74 +262,138 @@ describe("Entities tests", () => {
...
+    await testEditor.entities.createPrimitiveProperty(entityRes, "TestProp", PrimitiveType.DateTime);
+
+    await expect(testEditor.entities.setBaseClass(entityRes, baseClassRes)).to.be.eventually.rejected.then(function (error) {
+      expect(error).to.have.property("schemaEditType", SchemaEditType.SetBaseClass);
+      expect(error).to.have.nested.property("innerError.message", `Found property 'testEntity.TestProp' of type 'dateTime' which is not compatible with the base property 'testBaseClass.TestProp' of type 'boolean'.`);
+      expect(error).to.have.nested.property("innerError.errorStatus", ECEditingStatus.InvalidPropertyType);
+    });
+  });
+
+  it("try setting base class that contains, through inheritance, an incompatible property type with the entity class (grandchild), returns error", async () => {
```
christophermlawson, line 344: @ColinKerr Is this a valid unit test? It passes because of some new validation code I wrote that does not exist in C#.  Seems like a good check to me...

StefanApfel-Bentley:
It seems the only time we need that SchemaContextEditor in the SchemaEditChangeBase implementations is to [add the edit info](https://github.com/iTwin/itwinjs-core/pull/6997/files#diff-ee30444bedc1125a5eff46d8f11706a927c16810a45c587abdd6f9796d021244R77) (and to call [changeCancelled](https://github.com/iTwin/itwinjs-core/pull/6997/files#diff-ee30444bedc1125a5eff46d8f11706a927c16810a45c587abdd6f9796d021244R147).

I think it would be cleaner if the adding to a context is done in the context since it also manages the changes. That would prevent a reference to the Editor in the ChangeInfo objects (which would be purely informative) and move the majority of change-tracking and reverting logic to the context only.

core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -61,9 +66,41 @@ export class SchemaContextEditor {
   public get schemaContext(): SchemaContext { return this._schemaContext; }
 
   public async finish(): Promise<SchemaContext> {
```
StefanApfel-Bentley, line 68: Returning the SchemaContext here, doesn't give us a lot of benefits. Since we know at that point what changes has been made to the schema, it would be probably more interesting to return the applied changes to the context in the current session. Especially as we can get the SchemaContext associated with the editor through the[ Editor's schemaContext property](https://github.com/iTwin/itwinjs-core/pull/6997/files#diff-fee38c49a6a2c6681281283ef37b7cccbc3a54af372ada73375407be15586b32R66).

core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -61,9 +66,41 @@ export class SchemaContextEditor {
...
+    await change.revertChange();
+    this.removeChange(change);
+  }
+
+  public changeCancelled(change: ISchemaEditChangeInfo) {
+    this.removeChange(change);
+  }
+
+  public removeChange(change: ISchemaEditChangeInfo) {
+    this._currentChanges = this._currentChanges.splice(change.sequence, 1);
```
StefanApfel-Bentley, line 96: I think that won't work. If I add 3 changes, the list would look like this (mocking up)...
```ts
0: { sequence: 0, editType: "SetLabel", ... }
1: { sequence: 1, editType: "SetPriority", ... }
2: { sequence: 2, editType: "SetDescription", ... }
```
If I now remove, the 2n change, and since all methods are public, we might want to have that possibility, it would get sliced out from the list
```ts
0: { sequence: 0, editType: "SetLabel", ... }
1: { sequence: 2, editType: "SetDescription", ... }
```
As the sequence of all other edits remain, if you want to call removeChange with the SetDescription edit, you'd call splice to remove the entry on index 2 (as that is the sequence value), but actually want to remove index 1.

So I guess we either want to prevent that changes can be removed anytime, or we need a lookup before calling splice like:
```ts
const changeIndex = this._currentChanges.find((entry) => entry.sequence === change.sequence)); // alternativly this._currentChanges.indexOf(change)
this._currentChanges = this._currentChanges.splice(changeIndex, 1);
```

core/ecschema-editing/src/Editing/ChangeInfo/ChangeInfo.ts
```diff
@@ -0,0 +1,151 @@
...
+  readonly changeOptions?: ChangeOptionProps;
+  readonly editType?: string;
+  readonly schemaItemType?: string;
+}
+
+/**
+ * Defines the properties required for a specific schema edit operation.
+ * @alpha
+ */
+export interface ISchemaEditChangeInfo {
```
StefanApfel-Bentley, line 31: Do we really need that interface? Since we have an abstract base class of all edits, we could also use that as carry type in the lists and references. As far as I see it, only toJson and beginChange is hidden in that interface, so we don't gain much by introducing another interface.

By the way, you can import classes acting as "interfaces" by using the type indicator:
```ts 
import type { SchemaEditChangeBase } from "./ChangeInfo"
```

StefanApfel-Bentley:
After changes became edits, we need to update a few more places (file names, properties and methods especially in ChangeInfo/Options) :)

core/ecschema-editing/src/Editing/ChangeInfo/ChangeInfo.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (!this.revertCallback)
+      return;
+
+    await this.revertCallback(this);
+  }
+
+  /**
+   * Serializes the object to json format.
+   */
+  public toJson() {
```
StefanApfel-Bentley, line 131: This doesn't do much at the moment.

christophermlawson: Removed

core/ecschema-editing/src/Editing/Editor.ts
christophermlawson: I removed the sequence property entirely.

core/ecschema-editing/src/Editing/Editor.ts
christophermlawson: Done.

christophermlawson:
This PR should remain open for now as it contains unfinished Schema Editing changes that we will most likely need in the future. This code should be reviewed for necessary changes, current conflicts resolved, and get buy-in from PR reviewers.

PR #6998, opened 2024-07-23
delete layered definitions

swwilson-bsi:
To handle layered dependencies, deleteElementTree must make multiple passes. In each pass, it creates and uses a new ElementTreeDeleter, which detects and classifies the remaining elements and calls IModelDb.deleteDefinitionElements on the remaining definition elements that it finds. 

Note that deleteDefinitionElements will delete only definition elements that are not "used" (as defined by NativeDb usage calculator).

Scenario: A LineStyle refers to a GeometryPart (which captures a LS symbol). The GeometryStreams and SubCategories that refer to the LineStyle must be deleted first. Then the LS symbol GeometryPart can be deleted. 

That's just one example. There can be further layers of dependencies.

### DIFF
--- a/core/backend/src/ElementTreeWalker.ts
+++ b/core/backend/src/ElementTreeWalker.ts
@@ -450,12 +450,16 @@ export class ElementSubTreeDeleter extends ElementTreeTopDown {
 /** Deletes an element tree starting with the specified top element. The top element is also deleted. Uses ElementTreeDeleter.
  * @param iModel The iModel
  * @param topElement The parent of the sub-tree
+ * @param maxPasses Maximum number of passes to make when deleting definitions
  * @beta
  */
-export function deleteElementTree(iModel: IModelDb, topElement: Id64String): void {
-  const del = new ElementTreeDeleter(iModel);
-  del.deleteNormalElements(topElement);
-  del.deleteSpecialElements();
+export function deleteElementTree(iModel: IModelDb, topElement: Id64String, maxPasses: number = 5): void {
+  let pass = 0;
+  do {
+    const del = new ElementTreeDeleter(iModel);
+    del.deleteNormalElements(topElement);
+    del.deleteSpecialElements();
+  } while ((iModel.elements.tryGetElement(topElement) !== undefined) && (++pass < maxPasses));
 }
 
 /** Deletes all element sub-trees that are selected by the supplied filter. Uses ElementSubTreeDeleter.
@@ -464,10 +468,14 @@ export function deleteElementTree(iModel: IModelDb, topElement: Id64String): voi
  * @param iModel The iModel
  * @param topElement Where to start the search.
  * @param filter Callback that selects sub-trees that should be deleted.
+ * @param maxPasses Maximum number of passes to make when deleting definitions
  * @beta
  */
-export function deleteElementSubTrees(iModel: IModelDb, topElement: Id64String, filter: ElementSubTreeDeleteFilter): void {
-  const del = new ElementSubTreeDeleter(iModel, filter);
-  del.deleteNormalElementSubTrees(topElement);
-  del.deleteSpecialElementSubTrees();
+export function deleteElementSubTrees(iModel: IModelDb, topElement: Id64String, filter: ElementSubTreeDeleteFilter, maxPasses: number = 5): void {
+  let pass = 0;
+  do {
+    const del = new ElementSubTreeDeleter(iModel, filter);
+    del.deleteNormalElementSubTrees(topElement);
+    del.deleteSpecialElementSubTrees();
+  } while ((iModel.elements.tryGetElement(topElement) !== undefined) && (++pass < maxPasses));
 }
--- a/core/backend/src/ElementTreeWalker.ts
+++ b/core/backend/src/ElementTreeWalker.ts
@@ -465,13 +465,14 @@ export function deleteElementTree(iModel: IModelDb, topElement: Id64String, maxP
 /** Deletes all element sub-trees that are selected by the supplied filter. Uses ElementSubTreeDeleter.
  * If the filter selects the top element itself, then the entire tree (including the top element) is deleted.
  * That has the same effect as calling [[deleteElementTree]] on the top element.
+ * @note The caller may have to call this function multiple times if there are multiple layers of dependencies among definition elements.
  * @param iModel The iModel
  * @param topElement Where to start the search.
  * @param filter Callback that selects sub-trees that should be deleted.
  * @param maxPasses Maximum number of passes to make when deleting definitions
  * @beta
  */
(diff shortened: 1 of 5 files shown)


pmconne:
> Note that deleteDefinitionElements will delete only definition elements that are not "used" (as defined by NativeDb usage calculator).

This means that your attempt to delete the subtree may fail no matter how many passes you make, right?
You currently just loop `maxPasses` times (or until the top element is successfully deleted).
Do you have a way to detect that deletion will never succeed because of used definition elements, and bail before reaching `maxPasses`?

core/backend/src/ElementTreeWalker.ts
```diff
@@ -450,17 +450,22 @@ export class ElementSubTreeDeleter extends ElementTreeTopDown {
...
  */
-export function deleteElementTree(iModel: IModelDb, topElement: Id64String): void {
-  const del = new ElementTreeDeleter(iModel);
-  del.deleteNormalElements(topElement);
-  del.deleteSpecialElements();
+export function deleteElementTree(iModel: IModelDb, topElement: Id64String, maxPasses: number = 5): void {
+  let pass = 0;
+  do {
+    const del = new ElementTreeDeleter(iModel);
+    del.deleteNormalElements(topElement);
```
nick4598, line 460: Can you explain this to me? The comment above deleteElementTree mentions the top element is also deleted, yet you now have to try and delete the element many times, because it can't be deleted until all of its definitionElements have been deleted? And deleting some definitionElements may make other definitionElements unused, allowing you to delete more?

core/backend/src/ElementTreeWalker.ts
```diff
@@ -450,17 +450,22 @@ export class ElementSubTreeDeleter extends ElementTreeTopDown {
...
  * @param iModel The iModel
  * @param topElement The parent of the sub-tree
+ * @param maxPasses Maximum number of passes to make when deleting definitions
  * @beta
  */
-export function deleteElementTree(iModel: IModelDb, topElement: Id64String): void {
-  const del = new ElementTreeDeleter(iModel);
-  del.deleteNormalElements(topElement);
-  del.deleteSpecialElements();
+export function deleteElementTree(iModel: IModelDb, topElement: Id64String, maxPasses: number = 5): void {
```
pmconne, line 456: Avoid piling on optional arguments. Something like this instead, keeping the original function with two arguments to avoid breaking existing callers, and adding an overload taking an `Args` object where new stuff can be added moving forward.
```ts
/** Arguments supplied to [[deleteElementTree]].
 * @beta
 */
export interface DeleteElementTreeArgs {
  /** The iModel containing the elements to delete. */
  iModel: IModelDb;
  /** The Id of the root element of the tree to delete. */
  topElement: Id64String;
  /** The maximum number of passes to make when deleting definition elements.
   * Default: 5
   */
  maxPasses?: number;
}
/** @beta */
export function deleteElementTree(iModel: IModelDb, topElement: Id64String): void;
/** @beta */
export function deleteElementTree(args: DeleteElementTreeArgs): void;
/** @internal */
export function deleteElementTree(arg0: DeleteElementTreeArgs | IModelDb, arg1?: IModelDb): void {
  let maxPasses = 5;
  let iModel: IModelDb;
  let topElement: Id64String;
  if (arg0 instanceof IModelDb) {
    assert(typeof arg1 === "string";
    iModel = arg0;
    topElement = arg1;
  } else {
    iModel = arg0.iModel
    topElement = arg0.topElement;
    maxPasses = arg0.maxPasses;
  }

  /* ... */
}
```

core/backend/src/ElementTreeWalker.ts
swwilson-bsi: Yes, that's right. There may be definitions in DefinitionPartitions/Containers that are children of topElement. If those definitions cannot be deleted because they are still in use, then the parent, topElement, can't be deleted. So, checking to see if topElement was not deleted is the indirect way of telling that some of its child definitions could not be deleted.

core/backend/src/ElementTreeWalker.ts
swwilson-bsi: > Do you have a way to detect that deletion will never succeed because of used definition elements, and bail before reaching maxPasses?

No way that I know of. That would come up if some of the definitions were used by elements outside of the tree parented by topElement. NativeDb.queryDefinitionElementUsagedoes not return information about where definitions are used. My code would have to essentially repeat the select statements that it executes and further analyze the results. Does not seem worth it to me.

swwilson-bsi: I added the overloads

core/backend/src/ElementTreeWalker.ts
```diff
@@ -447,20 +447,60 @@ export class ElementSubTreeDeleter extends ElementTreeTopDown {
...
+  /** The maximum number of passes to make when deleting definition elements.
+   * Default: 5
+   */
+  maxPasses?: number;
+}
+
 /** Deletes an element tree starting with the specified top element. The top element is also deleted. Uses ElementTreeDeleter.
  * @param iModel The iModel
  * @param topElement The parent of the sub-tree
+ * @param maxPasses Maximum number of passes to make when deleting definitions
```
pmconne, line 467:
```suggestion
```

core/backend/src/ElementTreeWalker.ts
```diff
@@ -447,20 +447,59 @@ export class ElementSubTreeDeleter extends ElementTreeTopDown {
...
-  del.deleteNormalElements(topElement);
-  del.deleteSpecialElements();
+export function deleteElementTree(iModel: IModelDb, topElement: Id64String): void;
+/** Deletes an element tree starting with the specified top element. The top element is also deleted. Uses ElementTreeDeleter.
+ * @param args Specifies the iModel and top element.
+ * @beta
+ */
+export function deleteElementTree(args: DeleteElementTreeArgs): void;
+/** @internal */
+export function deleteElementTree(arg0: DeleteElementTreeArgs | IModelDb, arg1?: Id64String): void {
```
nick4598, line 476: Can we keep arg1 as `topElement`?

nick4598 (approved):

core/backend/src/ElementTreeWalker.ts
swwilson-bsi: @pmconne ?

pmconne: Sounds a little nitpicky. @nick4598, why not use the "Add a suggestion" to propose your changes?

nick4598: I agree its nitpicky, but I didn't like that I had to read the code to figure out that topElement was still arg1. Maybe that is obvious to people more comfortable with overloads. 

I also didn't use add a suggestion because I figured it would cascade through the function, so I started looking at it locally and it doesn't feel like there is a good way to have arg1 be topElement and still use topElement within the function itself (would have to use a different variable I think within the function). I could be missing something, but it seems bad to change `arg1` for that reason. I will resolve.

PR #7000, opened 2024-07-23
check if current user still holds the write lock when upload changes

jiaruiz717:
iTwin/itwinjs-backlog#1170
 
There is a possible scenario that two users try to upload local changes to the same cloud container:
 
1. User1 grabs the write lock and sit there for a very long time. User1 does not release the write lock even after it has expired
2. User2 grabs the write lock after User1's write lock expires, and upload something to the cloud container
3. User1 then tries to upload changes,  we should disable User1 from uploading his changes because the write lock has expired
 
However, if there are no other users grab the write lock after User1, we allow User1 to upload changes(and refresh the write lock) even if the write lock has expired. 

Native change:

### DIFF
--- a/full-stack-tests/backend/src/integration/AzuriteTest.ts
+++ b/full-stack-tests/backend/src/integration/AzuriteTest.ts
@@ -4,13 +4,16 @@
 *--------------------------------------------------------------------------------------------*/
 
 import { expect } from "chai";
+import * as crypto from "crypto";
+import * as fs from "fs";
 import { emptyDirSync, mkdirsSync } from "fs-extra";
 import { join } from "path";
+import { open } from "sqlite";
+import * as sqlite3 from "sqlite3";
 import * as azureBlob from "@azure/storage-blob";
-import { BlobContainer, CloudSqlite, IModelHost, SettingsContainer } from "@itwin/core-backend";
+import { BlobContainer, CloudSqlite, IModelHost, KnownLocations, SettingsContainer } from "@itwin/core-backend";
 import { AccessToken, Guid } from "@itwin/core-bentley";
 import { LocalDirName, LocalFileName } from "@itwin/core-common";
-import * as crypto from "crypto";
 
 // spell:ignore imodelid itwinid mkdirs devstoreaccount racwdl
 
@@ -51,6 +54,29 @@ export namespace AzuriteTest {
       const azClient = createAzClient(container.containerId);
       const blobClient = azClient.getBlockBlobClient(blockName);
       return blobClient.exists();
+    };
+
+    export const overwriteWriteLockValue = async (container: CloudSqlite.CloudContainer, blockName: string) => {
+      const azClient = createAzClient(container.containerId);
+      const blobClient = azClient.getBlockBlobClient(blockName);
+      const tempFilePath = join(KnownLocations.tmpdir, blockName);
+      await blobClient.downloadToFile(tempFilePath);
+
+      const bcv_kv = await open({
+        filename: tempFilePath,
+        driver: sqlite3.Database,
+      });
+      const row = await bcv_kv.get<{ v: string }>("SELECT v FROM kv WHERE k = 'writeLock'");
+      if (row) {
+        const writeLockData = JSON.parse(row.v);
+        const expiresDate = new Date(writeLockData.expires);
+        expiresDate.setMinutes(expiresDate.getMinutes() - 5);
+        writeLockData.expires = expiresDate.toISOString();
+        await bcv_kv.run("UPDATE kv SET v = ? WHERE k = 'writeLock'", JSON.stringify(writeLockData));
+      }
+      await bcv_kv.close();
+      await blobClient.uploadFile(tempFilePath);
+      fs.unlinkSync(tempFilePath);
     };
 
     export const setSasToken = async (container: CloudSqlite.CloudContainer, accessLevel: BlobContainer.RequestAccessLevel) => {
--- a/full-stack-tests/backend/src/integration/AzuriteTest.ts
+++ b/full-stack-tests/backend/src/integration/AzuriteTest.ts
@@ -94,7 +94,7 @@ export namespace AzuriteTest {
       emptyDirSync(name);
     };
 
-    export interface TestContainerProps { containerId: string, logId?: string, isPublic?: boolean, writeable?: boolean }
+    export interface TestContainerProps { containerId: string, logId?: string, isPublic?: boolean, writeable?: boolean, lockExpireSeconds?: number}
 
     export const makeContainer = async (arg: TestContainerProps): Promise<TestContainer> => {
       const containerProps = { ...arg, writeable: true, baseUri, storageType };
--- a/full-stack-tests/backend/src/integration/CloudSqlite.test.ts
+++ b/full-stack-tests/backend/src/integration/CloudSqlite.test.ts
@@ -227,7 +227,7 @@ describe("CloudSqlite", () => {
     testContainer0.disconnect();
   });
 
-  it.only("Should user1 fail to upload changes if user2 holds write lock after user1's expiration time", async () => {
+  it("Should user1 fail to upload changes if user2 holds write lock after user1's expiration time", async () => {
     // simulate two users in two processes
     // test container 1 and test container 2 are actually the same cloud container
     const testContainer1 = testContainers[3];
@@ -249,16 +249,7 @@ describe("CloudSqlite", () => {
     testContainer2.connect(testCache2);
     testContainer2.acquireWriteLock(user2);
     // user1 tries to upload changes, it should fail because user1's write lock has expired and user2 is using it
-    // await expect(testContainer1.uploadChanges()).to.eventually.be.rejectedWith("container is currently locked by another user");
     expect( async () => testContainer1.uploadChanges()).to.throw(`Container [${testContainer1.containerId}] is currently locked by another user`);
-
-    // expect(testContainer1.uploadChanges()).throws("container is currently locked by another user");
-    // try {
-    //   await testContainer1.uploadChanges();
-    // } catch (error: any) {
-    //   console.log(error.message);
-    // }
-    // await BeDuration.wait(6000);
     testContainer1.disconnect();
     testContainer2.disconnect();
   });
--- a/full-stack-tests/backend/src/integration/CloudSqlite.test.ts
+++ b/full-stack-tests/backend/src/integration/CloudSqlite.test.ts
@@ -195,36 +195,18 @@ describe("CloudSqlite", () => {
   });
 
   it("Should a single user upload changes successfully even if write lock expires", async () => {
-    const testContainer0 = testContainers[0];
-    // const testContainerArg ={
-    //   containerId: "test0",
-    //   logId: "logId-0",
-    //   writeable: true,
-    //   baseUri: "http://127.0.0.1:10000/devstoreaccount1",
-    //   storageType: "azure" as const,
-    // };
-    // await azSqlite.createAzContainer(testContainerArg);
-    // const accessToken = await CloudSqlite.requestToken(testContainerArg);
-    // const lockExpireSeconds = 5;
-    // const testContainer0 = CloudSqlite.createCloudContainer({ ...testContainerArg, accessToken, lockExpireSeconds});
+    const testContainer1 = testContainers[0];
     const testCache = azSqlite.makeCaches(["testCache"])[0];
-    testContainer0.connect(testCache);
-    testContainer0.acquireWriteLock(user1);
-    await testContainer0.copyDatabase("testBim", "testBimCopy");
-    const db = await BriefcaseDb.open({ fileName: "testBimCopy", container: testContainer0 });
+    testContainer1.connect(testCache);
+    testContainer1.acquireWriteLock(user1);
+    await testContainer1.copyDatabase("testBim", "testBimCopy");
+    const db = await BriefcaseDb.open({ fileName: "testBimCopy", container: testContainer1 });
     db.saveFileProperty({ name: "logMask", namespace: "logMaskTest", id: 1, subId: 1 }, "this is a test");
     db.close();
-    await BeDuration.wait(10000);
-    // await testContainer0.uploadChanges();
-    await expect(testContainer0.uploadChanges()).to.eventually.be.fulfilled;
-    // testContainer0.releaseWriteLock();
-    // await CloudSqlite.withWriteLock({ user: user1, container: testContainer0 }, async () => {
-    //   await testContainer0.copyDatabase("testBim", "testBimCopy");
-    //   const db = await BriefcaseDb.open({ fileName: "testBimCopy", container: testContainer0 });
-    //   db.saveFileProperty({ name: "logMask", namespace: "logMaskTest", id: 1, subId: 1 }, "this is a test");
-    //   db.close();
-    // });
-    testContainer0.disconnect();
+    // set the expires time to five mins later, which avoids waiting for 5 mins here.
+    await azSqlite.overwriteWriteLockValue(testContainer1, "bcv_kv.bcv");
+    await expect(testContainer1.uploadChanges()).to.eventually.be.fulfilled;
+    testContainer1.disconnect();
   });
 
   it("Should user1 fail to upload changes if user2 holds write lock after user1's expiration time", async () => {
@@ -244,12 +226,13 @@ describe("CloudSqlite", () => {
     const db1 = await BriefcaseDb.open({ fileName: "testBimCopy1", container: testContainer1 });
     db1.saveFileProperty({ name: "logMask", namespace: "logMaskTest", id: 1, subId: 1 }, "this is a test");
     db1.close();
-    await BeDuration.wait(6000);
+    // set the expires time to five mins later, which avoids waiting for 5 mins here.
+    await azSqlite.overwriteWriteLockValue(testContainer1, "bcv_kv.bcv");
     // user2 grabs the write lock
     testContainer2.connect(testCache2);
     testContainer2.acquireWriteLock(user2);
     // user1 tries to upload changes, it should fail because user1's write lock has expired and user2 is using it
-    expect( async () => testContainer1.uploadChanges()).to.throw(`Container [${testContainer1.containerId}] is currently locked by another user`);
+    expect(() => testContainer1.uploadChanges()).to.throw(`Container [${testContainer1.containerId}] is currently locked by another user`);
     testContainer1.disconnect();
     testContainer2.disconnect();
   });
--- a/full-stack-tests/backend/src/integration/CloudSqlite.test.ts
+++ b/full-stack-tests/backend/src/integration/CloudSqlite.test.ts
@@ -10,7 +10,7 @@ import { join } from "path";
 import * as sinon from "sinon";
 import { _nativeDb, BlobContainer, BriefcaseDb, CloudSqlite, IModelHost, IModelJsFs, KnownLocations, PropertyStore, SnapshotDb, SQLiteDb } from "@itwin/core-backend";
 import { KnownTestLocations } from "@itwin/core-backend/lib/cjs/test";
-import { assert, BeDuration, DbResult, Guid, GuidString, Logger, LogLevel, OpenMode } from "@itwin/core-bentley";
+import { assert, BeDuration, DbResult, Guid, GuidString, OpenMode } from "@itwin/core-bentley";
 import { AzuriteTest } from "./AzuriteTest";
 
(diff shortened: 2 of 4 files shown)


full-stack-tests/backend/src/integration/AzuriteTest.ts
```diff
@@ -53,6 +56,29 @@ export namespace AzuriteTest {
       return blobClient.exists();
     };
 
+    export const overwriteWriteLockValue = async (container: CloudSqlite.CloudContainer, blockName: string) => {
+      const azClient = createAzClient(container.containerId);
+      const blobClient = azClient.getBlockBlobClient(blockName);
+      const tempFilePath = join(KnownLocations.tmpdir, blockName);
+      await blobClient.downloadToFile(tempFilePath);
+
+      const bcv_kv = await open({
```
nick4598, line 65: Instead of introducing the dependency on sqlite into full-stack-tests/backend, you should try doing

`const db = new SQLiteDb()`
`db.openDb("Path-to-db", ...)`

nick4598: I would also maybe rename the function to be setWriteLockToExpired? Just want to capture that the function is setting it to a time in the past.

jiaruiz717: Resolved in [commit c79429aefcf07d9be974cdf095add0ce657c9c55](https://github.com/iTwin/itwinjs-core/pull/7000/commits/c79429aefcf07d9be974cdf095add0ce657c9c55)

full-stack-tests/backend/src/integration/AzuriteTest.ts
```diff
@@ -56,26 +56,31 @@ export namespace AzuriteTest {
...
-      const bcv_kv = await open({
-        filename: tempFilePath,
-        driver: sqlite3.Database,
+      const bcv_kv = new SQLiteDb();
+      bcv_kv.openDb(tempFilePath, {openMode: OpenMode.ReadWrite, rawSQLite: true});
+      // read writeLock.expires and modify it
+      bcv_kv.withSqliteStatement("SELECT v FROM kv WHERE k = 'writeLock'", (stmt) => {
+        if (stmt.nextRow()) {
+          const writeLockData = JSON.parse(stmt.getValueString(0));
+          const expiresDate = new Date(writeLockData.expires);
```
nick4598, line 71: Rereading this more closely, this is actually subtracting 5 minutes from the expiry date of the write lock right? I thought at first it was doing current time - 5 mins to guarantee expired. So my naming suggestion might be wrong in that case. Maybe it could be changed to do current time - 5 minutes? Or maybe a parameter that says numMinsToSubtract?

jiaruiz717: m_lockExpireSeconds = std::min((int) (12*SECONDS_PER_HOUR), std::max(5*(int)SECONDS_PER_MINUTE, m_lockExpireSeconds));

this is the formula we use to calculate m_lockExpireSeconds . In our test case, it is 5mins.
we also have GetServerTime() on native to read/set the expires time in bcv_kv.bcv. 
According to the comment of GetServerTime:

   * Get the date(s) returned by the server in the http header when the bcv_kv blob was fetched.
   * This avoids any dependence on the local computer's time for detecting expired write locks.

If we do current time -5 mins, then there's a possibility that (current time -5 mins) > GetServerTime(), it will prevent user2 from grabbing the write lock.

nick4598: Since we use azurite, a local storage emulator I would expect that the server's time is also the local computer's time. I only want the function name to accurately represent what its doing. The name 'makewritelockexpired' is only true in the specific case that the write lock is already grabbed for 5 minutes.

jiaruiz717: Resolved in [commit 1ea569504b20baaa08796b742a7521f363f75423](https://github.com/iTwin/itwinjs-core/pull/7000/commits/1ea569504b20baaa08796b742a7521f363f75423)

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -235,6 +235,22 @@ describe("CloudSqlite", () => {
...
+    const testContainer1 = testContainers[3];
+    const testCache1 = azSqlite.makeCaches(["testCache1"])[0];
+    testContainer1.connect(testCache1);
+    testContainer1.acquireWriteLock(user1);
+    await testContainer1.copyDatabase("testBim", "testBimCopy1");
+    const db1 = await BriefcaseDb.open({ fileName: "testBimCopy1", container: testContainer1 });
+    db1.saveFileProperty({ name: "refresh", namespace: "refreshTest", id: 1, subId: 1 }, "this is a test");
+    db1.close();
+    await azSqlite.setWriteLockToExpired(testContainer1, "bcv_kv.bcv");
+    // there are no other users, uploadChanges should refresh the write lock for the current user.
```
nick4598, line 248: What about when writeLock isn't expired?

jiaruiz717: added in [commit 0bd6507cf820f33d8598b933ffe4f75734b80465](https://github.com/iTwin/itwinjs-core/pull/7000/commits/0bd6507cf820f33d8598b933ffe4f75734b80465)

full-stack-tests/backend/src/integration/AzuriteTest.ts
```diff
@@ -94,9 +94,10 @@ export namespace AzuriteTest {
         bcvDb.withSqliteStatement("SELECT v FROM kv WHERE k = 'writeLock'", (stmt) => {
           if (stmt.nextRow()) {
             const writeLockData = JSON.parse(stmt.getValueString(0));
-            const expiresDate = new Date(writeLockData.expires);
-            const currentTime = new Date();
-            resolve(expiresDate > currentTime);
+            const expiresTime = new Date(writeLockData.expires);
+            // should be 5 min but let's use 4.5min to avoid time conflict
+            const writeLockTimeRemains = 4.5*60*1000;
```
nick4598, line 99: How about changing 'isWriteLockRefreshed' to 'isWriteLockValidForAtleast' And taking writeLockTimeRemains as a parameter?

full-stack-tests/backend/src/integration/AzuriteTest.ts
```diff
@@ -54,7 +54,7 @@ export namespace AzuriteTest {
       return blobClient.exists();
     };
 
-    export const setWriteLockToExpired = async (container: CloudSqlite.CloudContainer, blockName: string) => {
+    export const resetWriteLockExpireTime = async (container: CloudSqlite.CloudContainer, blockName: string, numOfMin: number) => {
```
nick4598, line 57: maybe this should be called 'subtractFromCurrentWriteLockExpiryTime' ?

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -239,15 +239,28 @@ describe("CloudSqlite", () => {
...
+    // there are no other users, uploadChanges should refresh the write lock for the current user
     await testContainer1.uploadChanges();
-    expect (await azSqlite.isWriteLockRefreshed(testContainer1, "bcv_kv.bcv")).to.be.true;
+    expect (await azSqlite.isWriteLockRefreshed(testContainer1, "bcv_kv.bcv", currentTime)).to.be.true;
+    // case 2: current write lock has not expired
+    currentTime = new Date();
+    // the write lock is refreshed to 5 minutes later
+    // mock 4 min operation time by user1
+    await azSqlite.resetWriteLockExpireTime(testContainer1, "bcv_kv.bcv", 4);
+    // now user1 only has 1 min to perform further actions
```
nick4598, line 260: You could also use the isValidForAtleast function here as well to expect that isValidForAtleast 1.5 minutes ends up being false or something along those lines.

jiaruiz717: Resolved in [commit 8a0fdb5818d3144a873d3ae42a56c9a99db05473](https://github.com/iTwin/itwinjs-core/pull/7000/commits/8a0fdb5818d3144a873d3ae42a56c9a99db05473)

full-stack-tests/backend/src/integration/AzuriteTest.ts
jiaruiz717: Resolved in [commit 8a0fdb5818d3144a873d3ae42a56c9a99db05473](https://github.com/iTwin/itwinjs-core/pull/7000/commits/8a0fdb5818d3144a873d3ae42a56c9a99db05473)

full-stack-tests/backend/src/integration/AzuriteTest.ts
jiaruiz717: Resolved in [commit 8a0fdb5818d3144a873d3ae42a56c9a99db05473](https://github.com/iTwin/itwinjs-core/pull/7000/commits/8a0fdb5818d3144a873d3ae42a56c9a99db05473)

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -188,6 +191,82 @@ describe("CloudSqlite", () => {
...
+    db1.saveFileProperty({ name: "upload", namespace: "uploadTest", id: 1, subId: 1 }, "this is a test");
+    db1.close();
+    // set the expires time to five mins earlier, which avoids waiting for 5 mins here.
+    await azSqlite.subtractFromCurrentWriteLockExpiryTime(testContainer1, "bcv_kv.bcv", 5);
+    // user2 grabs the write lock
+    testContainer2.connect(testCache2);
+    testContainer2.acquireWriteLock(user2);
+    // user1 tries to upload changes, it should fail because user1's write lock has expired and user2 is using it
+    // eslint-disable-next-line @typescript-eslint/promise-function-async
+    expect(() => testContainer1.uploadChanges()).to.throw(`Container [${testContainer1.containerId}] is currently locked by another user`);
```
nick4598, line 233: Did you test the case where user2 has the writelock, makes changes and user1 should still fail?

full-stack-tests/backend/src/integration/AzuriteTest.ts
```diff
@@ -4,13 +4,14 @@
 *--------------------------------------------------------------------------------------------*/
 
 import { expect } from "chai";
+import * as crypto from "crypto";
```
MichaelSwigerAtBentley, line 7: For imports from node, you can use `from "node:crypto"` or `from "node:fs"` to add clarity as to where these utils are coming from. It is functionally the same, but makes it easier to read.

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -188,6 +191,90 @@ describe("CloudSqlite", () => {
     container.disconnect({detach: true});
   });
 
+  it("Should a single user upload changes successfully even if write lock expires", async () => {
```
MichaelSwigerAtBentley, line 194: Maybe this should be "(it) should upload a single user's changes successfully when a write lock is expired"

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -188,6 +191,90 @@ describe("CloudSqlite", () => {
...
+    const testCache = azSqlite.makeCaches(["testCache"])[0];
+    testContainer1.connect(testCache);
+    testContainer1.acquireWriteLock(user1);
+    await testContainer1.copyDatabase("testBim", "testBimCopy");
+    const db = await BriefcaseDb.open({ fileName: "testBimCopy", container: testContainer1 });
+    db.saveFileProperty({ name: "logMask", namespace: "logMaskTest", id: 1, subId: 1 }, "this is a test");
+    db.close();
+    // set the expires time to five mins earlier, which avoids waiting for 5 mins here.
+    await azSqlite.subtractFromCurrentWriteLockExpiryTime(testContainer1, "bcv_kv.bcv", 5);
+    await expect(testContainer1.uploadChanges()).to.eventually.be.fulfilled;
```
MichaelSwigerAtBentley, line 205: Is `eventually.be.fulfilled` enough to verify a successfully upload?

jiaruiz717: The point of `await expect(testContainer1.uploadChanges()).to.eventually.be.fulfilled;` is to check that no error is thrown from `RequireWriteLock()` on the native side. We don't really care if the upload itself is successful or not imo.

MichaelSwigerAtBentley: Makes sense, just wanted to double check

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
jiaruiz717: resolved in [commit 7ad75934bda38b7d23a0730443088800a94e3716](https://github.com/iTwin/itwinjs-core/pull/7000/commits/7ad75934bda38b7d23a0730443088800a94e3716)

full-stack-tests/backend/src/integration/AzuriteTest.ts
jiaruiz717: resolved in [commit 7ad75934bda38b7d23a0730443088800a94e3716](https://github.com/iTwin/itwinjs-core/pull/7000/commits/7ad75934bda38b7d23a0730443088800a94e3716)

full-stack-tests/backend/src/integration/AzuriteTest.ts
```diff
@@ -53,6 +54,58 @@ export namespace AzuriteTest {
       return blobClient.exists();
     };
 
+    export const subtractFromCurrentWriteLockExpiryTime = async (container: CloudSqlite.CloudContainer, blockName: string, numOfMin: number) => {
```
nick4598, line 57: can blockName just be hardcoded in this function? I think the fxn only makes sense if we're working with bcv_kv.bcv

full-stack-tests/backend/src/integration/AzuriteTest.ts
```diff
@@ -53,6 +54,58 @@ export namespace AzuriteTest {
...
+          bcvDb.saveChanges();
+        }
+      });
+      await blobClient.uploadFile(tempFilePath);
+      // clean up
+      bcvDb.closeDb();
+      fs.unlinkSync(tempFilePath);
+    };
+
+    export const isWriteLockValidForAtLeast = async (container: CloudSqlite.CloudContainer, blockName: string, currentTime: Date, writeLockTimeRemainsMs: number): Promise<boolean> => {
```
nick4598, line 85: same wtih this blockName, bcv_kv.bcv is implied to be the blockName since we're dealing with the writelock

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -188,6 +191,90 @@ describe("CloudSqlite", () => {
...
+    const testContainer1 = testContainers[0];
+    const testCache = azSqlite.makeCaches(["testCache"])[0];
+    testContainer1.connect(testCache);
+    testContainer1.acquireWriteLock(user1);
+    await testContainer1.copyDatabase("testBim", "testBimCopy");
+    const db = await BriefcaseDb.open({ fileName: "testBimCopy", container: testContainer1 });
+    db.saveFileProperty({ name: "logMask", namespace: "logMaskTest", id: 1, subId: 1 }, "this is a test");
+    db.close();
+    // set the expires time to five mins earlier, which avoids waiting for 5 mins here.
+    await azSqlite.subtractFromCurrentWriteLockExpiryTime(testContainer1, "bcv_kv.bcv", 5);
```
nick4598, line 204: Is the write lock actually expired here? isn't testContainers[0] not passing a lockExpireSeconds which means we default to 1hr for the write lock?

nick4598: We should assert that our assumption the write lock is expired is true using that new function you added

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -188,6 +191,90 @@ describe("CloudSqlite", () => {
...
+    // user1 grabs the write lock, make some changes but don't release the write lock
+    // after a specific amount of time, the write lock expires
+    testContainer1.connect(testCache1);
+    testContainer1.acquireWriteLock(user1);
+    await testContainer1.copyDatabase("testBim", "testBimCopy1");
+    const db1 = await BriefcaseDb.open({ fileName: "testBimCopy1", container: testContainer1 });
+    db1.saveFileProperty({ name: "upload", namespace: "uploadTest", id: 1, subId: 1 }, "this is a test");
+    db1.close();
+    // set the expires time to five mins earlier, which avoids waiting for 5 mins here.
+    await azSqlite.subtractFromCurrentWriteLockExpiryTime(testContainer1, "bcv_kv.bcv", 5);
```
nick4598, line 227: Similar to above, youre subtracting 5 mins here, but isn't the writeLockExpireSeconds on the testContainers[3] and testContainers[4] in seconds? You should convert them to minutes up above by multiplying by 60.

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
```diff
@@ -188,6 +191,90 @@ describe("CloudSqlite", () => {
...
+    expect (await azSqlite.isWriteLockValidForAtLeast(testContainer1, "bcv_kv.bcv", currentTime, 4.5*60*1000)).to.be.true;
+    // case 2: current write lock has not expired
+    currentTime = new Date();
+    // the write lock is refreshed to 5 minutes later
+    // mock 4 min operation time by user1
+    await azSqlite.subtractFromCurrentWriteLockExpiryTime(testContainer1, "bcv_kv.bcv", 4);
+    // now user1 only has 1 min to perform further actions
+    // uploadChanges does nothing this time but to refresh write lock
+    await testContainer1.uploadChanges();
+    // should be 5 min but let's use 4.5 min to avoid time conflict
```
nick4598, line 272: Since we had 1 min left we're actually adding 5 mins to get up to 6 mins right?

full-stack-tests/backend/src/integration/AzuriteTest.ts
MichaelSwigerAtBentley: cloud just make it optional with default, get the best of both worlds

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
jiaruiz717: `m_lockExpireSeconds = std::min((int) (12*SECONDS_PER_HOUR), std::max(5*((int)SECONDS_PER_MINUTE), m_lockExpireSeconds));`

So for testContainers[3] and testContainers[4], the m_lockExpireSeconds is 5 mins.

full-stack-tests/backend/src/integration/AzuriteTest.ts
jiaruiz717: Resolved in [commit b6de82106a3f57bb9db95b4dbd3dcfd47076e2c8](https://github.com/iTwin/itwinjs-core/pull/7000/commits/b6de82106a3f57bb9db95b4dbd3dcfd47076e2c8)

full-stack-tests/backend/src/integration/AzuriteTest.ts
jiaruiz717: Resolved in [commit b6de82106a3f57bb9db95b4dbd3dcfd47076e2c8](https://github.com/iTwin/itwinjs-core/pull/7000/commits/b6de82106a3f57bb9db95b4dbd3dcfd47076e2c8)

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
jiaruiz717: Resolved in [commit b6de82106a3f57bb9db95b4dbd3dcfd47076e2c8](https://github.com/iTwin/itwinjs-core/pull/7000/commits/b6de82106a3f57bb9db95b4dbd3dcfd47076e2c8)

full-stack-tests/backend/src/integration/CloudSqlite.test.ts
jiaruiz717: No, as long as user `uploadChanges`, `AcquireWriteLock()` refreshes the write lock expiry time to (SERVER_TIME+ 5 mins). In other words, the new expiry time is based on server time, not the previous expiry time.

So after the refresh, users should be able to own write lock for another 5 mins instead of 6 mins

PR #7010, opened 2024-07-26
Mark Bing Maps as deprecated, re-enable support for Azure Maps

### LINKED ISSUE #7009: Bing Maps Deprecation
On June 30, Microsoft [disabled the ability](https://www.microsoft.com/en-us/maps/bing-maps/discontinued-services) for users to generate new Bing Map keys. Existing hyperlinks that pointed our users to create a bing Map key would be redirected to the Url above, recommending migrating to Azure Maps. Existing Bing Map keys will continue to be supported until June 20, 2028, after which the service will be shut down (as explained in Microsoft's deprecation notice).

I'll highlight two components in core that make use of Bing Maps, and the effect of the deprecation:

- [BingLocationProvider](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/BingLocation.ts): Azure Maps has the [same feature available](https://learn.microsoft.com/en-us/azure/azure-maps/migrate-find-location-query), with some slight differences in the response body of the new endpoint. We can create a new AzureLocationProvider to support the same functionality.
- [BingElevationProvider](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/tile/map/BingElevation.ts): Azure Maps [deprecated](https://azure.microsoft.com/en-us/updates/azure-maps-elevation-apis-and-render-v2-dem-tiles-will-be-retired-on-5-may-2023/) their elevation API as of May last year, and there doesn't seem to be an easy way to replace that. This might warrant more discussion on alternatives.

- In 4.8 or 4.9, depending on how quick the PR moves, we will flag relevant APIs and components that make use of BIngMaps as deprecated, without modifying the existing default behavior.

- We need to reenable the AzureMapsLayerFormat, revisit core's ability to support and use AzureMaps, without modifying the existing behavior of fetching Bing Maps for now. In 5.x, we can look to migrate to another map tile imagery provider, we do have 4 years before our hand is forced...

hl662:
closes #7009 
- Marks BingLocationProvider and BingElevationProvider as deprecated.
- Mark `BingMap` in `MapLayerOptions` interface as deprecated.
- Extend BackgroundMapProviderName to allow `AzureProvider`
- Creates a new AzureLocationProvider component.
- Uncomment `AzureMapsMapLayerFormat` in MapLayerImageryFormats.ts
- Adds option to provide a `IMJS_AZURE_MAPS_KEY` in display test app.
- Add a new Azure Maps object in `MapLayerSources.json`

Tested Azure Maps support on display test app, using an Azure Maps Key generated on my sandbox resource.
Enable support for AzureMaps in `MapLayerSettings`. Azure Maps provides different [tilesets](https://learn.microsoft.com/en-us/rest/api/maps/render/get-map-tile?view=rest-maps-2024-04-01&tabs=HTTP#tilesetid) for their imagery.
BackgroundMapType.Street uses microsoft's base.hybrid.road, which doesn't show water or green spots denoting parks or fields:

Contrast that to base.road, which BackgroundMapType.Hybrid will use:

I have to note that there's no replacement for [AerialWithLabels](https://learn.microsoft.com/en-us/azure/azure-maps/migrate-get-imagery-metadata), which is what we use often in our Bing Maps implementation. The closest is microsoft.imagery, which has satellite imagery, but no road labels. I set BackgroundMapType.Aerial to use this:


core/frontend/src/tools/ViewTool.ts
```diff
@@ -3316,6 +3316,8 @@ export class ViewGlobeLocationTool extends ViewTool {
 
     if (this._globalLocation === undefined) {
       const locationString = args.join(" ");
+      // We'll look to move to AzureLocationProvider as the default in 5.0.
```
aruniverse, line 3319: add this to the 5.0 backlog

pmconne:
@mdastous-bentley is it necessary and desirable to introduce another hard-coded map imagery provider as "special" the way we did MapBox and Bing? Especially when the new provider can't support our API contract like "aerial+labels"? BackgroundMapProps.providerName and BackgroundMapProps.providerData.mapType are [long-deprecated](https://github.com/iTwin/itwinjs-core/blob/34d29425f9a39c202c16f4a96c229a1990b28db8/core/common/src/BackgroundMapSettings.ts#L67).

mdastous-bentley:
> @mdastous-bentley is it necessary and desirable to introduce another hard-coded map imagery provider as "special" the way we did MapBox and Bing? 

@pmconne They are not **necessary**, but without those special types you have to provide a complete 'ImageMapLayerSettings' definition to `DisplayStyleState.MapImagerySettings.backgroundBase`, and this include the service endpoint URL.  So it depends if we want to keep this 'level' of functionality in itwin.js or not.

pmconne:
> They are not **necessary**, but without those special types you have to provide a complete 'ImageMapLayerSettings' definition to `DisplayStyleState.MapImagerySettings.backgroundBase`, and this include the service endpoint URL. So it depends if we want to keep this 'level' of functionality in itwin.js or not.

It would be nice if the BackgroundMapProvider could disclose what different types of imagery it provides, and allow user to select among them; rather than the 3 hard-coded possibilities enumerated by BackgroundMapType. And if applications could register whatever BackgroundMapProviders they want, rather than being constrained to either using our small hard-coded possibilities or providing a complete ImageMapLayerSettings definition.

We're trying to fit a square peg into a round hole here with Azure maps.

mdastous-bentley:
> > They are not **necessary**, but without those special types you have to provide a complete 'ImageMapLayerSettings' definition to `DisplayStyleState.MapImagerySettings.backgroundBase`, and this include the service endpoint URL. So it depends if we want to keep this 'level' of functionality in itwin.js or not.
> 
> It would be nice if the BackgroundMapProvider could disclose what different types of imagery it provides, and allow user to select among them; rather than pretending that every provider supports all of and only the 3 hard-coded possibilities enumerated by BackgroundMapType. And if applications could register whatever BackgroundMapProviders they want, rather than being constrained to either using our small hard-coded possibilities or providing a complete ImageMapLayerSettings definition.
> 
> We're trying to fit a square peg into a round hole here with Azure maps.

Actually this should be `MapLayerFormatRegistry`'s job... but you still have have to provide a complete 'ImageMapLayerSettings'  definition.

core/frontend/src/AzureLocation.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Cartographic } from "@itwin/core-common";
+import { request } from "./request/Request";
+import { IModelApp } from "./IModelApp";
+import { GlobalLocation } from "./ViewGlobalLocation";
+
+/** Provides an interface to the [Azure Maps Geocoding services](https://learn.microsoft.com/en-us/rest/api/maps/search/get-geocoding).
+ * @public
+ * @extensions
+ */
+export class AzureLocationProvider {
```
hl662, line 18: Personal reminder for me to update NextVer.md with this new provider

pmconne:
> @pmconne I noticed we this PR is not using `map-layers-formats` package.. I guess it would be hard to move because we want to to keep AzureMaps a "well-known" format right?

I am not sure exactly what you are asking, but the point of my questioning thus far has been: do we really want to introduce a new "well-known format", when I think we all agreed quite long ago that denoting Bing and MapBox as well-known formats was probably a mistake.

mdastous-bentley:
> > @pmconne I noticed we this PR is not using `map-layers-formats` package.. I guess it would be hard to move because we want to to keep AzureMaps a "well-known" format right?
> 
> I am not sure exactly what you are asking, but the point of my questioning thus far has been: do we really want to introduce a new "well-known format", when I think we all agreed quite long ago that denoting Bing and MapBox as well-known formats was probably a mistake. Did you all discuss that privately and decide to move forward with the approach in this PR?

I don't know what was the initial thought process of the actual design, but if the goal is to assist the API consumer in creating a `MapLayerSettings` object, I'm sure we could create helper functions that doesn't hardcoding anything in `core-frontend`.

The other thing that feels wrong to me, is that we store the URL of the azure service endpoint in the view definition... as soon as Azure changes its API, any saved view get invalidated.  Technically if we know the format id (AzureMaps), and style id (currently not stored), the provider should have enough information to generate the URL.

hl662:
> if applications could register whatever BackgroundMapProviders they want, rather than being constrained to either using our small hard-coded possibilities or providing a complete ImageMapLayerSettings definition.

I would prefer this, I wasn't familiar with how background map tiles are fetched coming into working on this PR.  On the side, AzureMapsLayerImageryProvider already existed in core-frontend prior to this PR, but it was disabled. Based on the discussion, do you want to immediately move that out of core-frontend, maybe to map-layer-formats in the extensions folder? 


> I don't know what was the initial thought process of the actual design, but if the goal is to assist the API consumer in creating a MapLayerSettings object, I'm sure we could create helper functions that doesn't hardcoding anything in core-frontend.

We do have an existing helper function to [create an ImageryProvider](https://github.com/iTwin/itwinjs-core/blob/67cbeb1b01c9f1f7ed9e29bb4a22615f35fbbbd4/core/frontend/src/tile/map/MapLayerFormatRegistry.ts#L182C10-L182C31) in MapLayerFormatRegistry, given a MapLayerSettings input, although that function is marked as internal. Not to mention MapLayerFormats containing a function of the same [name](https://github.com/iTwin/itwinjs-core/blob/67cbeb1b01c9f1f7ed9e29bb4a22615f35fbbbd4/core/frontend/src/tile/map/MapLayerFormatRegistry.ts#L54C17-L54C38). 
Perhaps moving the former away from @internal, adding some examples of how to create a ImageMapLayerSettings (using the [fromJSON](https://github.com/iTwin/itwinjs-core/blob/master/core/common/src/MapLayerSettings.ts) method for example) and documenting how to register your own ImageryProvider to be used for background mapping via iModelApp can help here. If the guide already exists then my bad.

PR #7011, opened 2024-07-29
build-tools: Extract API script improvements

### LINKED ISSUE #6980: Enhance `extract-api` script in `itwinjs-core` to allow creating API summary of APIs that aren't exported through barrel
Should work similar to what's here: https://github.com/iTwin/presentation/tree/master/apps/build-tools. But by default should work as it does today to avoid breaking existing users.

Yato333:
Closes #6980.

extract-api.js:
Added an option to extract unexported APIs.

This can be helpful to detect breaking changes in types that are not exposed by the authors directly but still can be indirectly 
 accessible by the consumers. For example, exporting interface of a React component properties can be undesirable because even adding an additional optional property can be considered a breaking change by the consumers that extend from this interface. 

This workflow is already utilized in Presentation libraries and a custom scripts had to be written to enforce this: https://github.com/iTwin/presentation/tree/master/apps/build-tools/bin

extract-api-summary.js
Fixed API regex, which was ignoring types with multi-line types such as
```ts
type X = Iterable<{
  a: number;
  b: number;
};
```
Added an option for improved output which will now contain type of the API (function, type, interface), and discard extra text right of the name of the APIs. This will help to distinguish duplicate lines when one of them signifies a type and another -- a namespace for that type. See changes to `betools.js` where an example is provided in the help text.

### DIFF
--- a/common/api/summary/appui-abstract.exports.csv
+++ b/common/api/summary/appui-abstract.exports.csv
@@ -185,6 +185,8 @@ deprecated;UiEvent
 public;UiEventDispatcher
 deprecated;UiEventDispatcher
 public;UiFlags
+public;UiItemProviderOverrides = MarkRequired
+deprecated;UiItemProviderOverrides = MarkRequired
 public;UiItemProviderRegisteredEventArgs
 deprecated;UiItemProviderRegisteredEventArgs
 public;UiItemsApplicationAction
--- a/common/api/summary/core-backend.exports.csv
+++ b/common/api/summary/core-backend.exports.csv
@@ -320,6 +320,7 @@ public;PullChangesArgs = ToChangesetArgs &
 public;PushChangesArgs 
 beta;QueryLocalChangesArgs
 beta;QueryWorkspaceResourcesArgs
+beta;QueryWorkspaceResourcesCallback = (resources: Iterable
 beta;class RecipeDefinitionElement 
 public;Relationship 
 public;Relationships
--- a/common/api/summary/core-common.exports.csv
+++ b/common/api/summary/core-common.exports.csv
@@ -132,6 +132,7 @@ public;CommonMapLayerProps
 internal;compareIModelTileTreeIds(lhs: IModelTileTreeId, rhs: IModelTileTreeId): number
 internal;CompositeTileHeader 
 internal;computeChildTileProps(parent: TileMetadata, idProvider: ContentIdProvider, root: TileTreeMetadata):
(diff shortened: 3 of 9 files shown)


pmconne:
The current output makes it very easy to spot potentially-breaking changes to union types.
```
public;JSONSchemaType = string | number | boolean | object | JSONSchemaType[]
public;JSONSchemaTypeName = "string" | "number" | "integer" | "boolean" | "null" | "array" | "object"
```

The new output would not even show a diff in the API summary if a breaking change was made:
```
public;type;JSONSchemaType
public;type;JSONSchemaTypeName
```

pmconne:
PR description points out several "flaws" but does not explicitly state the motivation for these changes. Are you writing code that needs to parse the output, and if so for what purpose?

Yato333:
@pmconne @grigasp
We wanted to add an option to the `extract-api` script, such that it would extract non-exported APIs to the API summary. We have already implemented this in `presentation` repository and wanted to add this as an option to `build-scripts` so that it could be reused in `presentation-components` (see https://github.com/iTwin/itwinjs-core/issues/6980).

While implementing these changes we've notices that the output  of `extract-api-summary` could use some improvements. 
Since changes to .csv files are usually reviewed in PRs, I'd argue this would help reviewers easier see and understand changes to what is exported. Let's walk through the flaws that I've mentioned.

> Some exported API items are missing

This is simply a bug that this PR would fix.

> The last column of the output is inconsistent

The last column of the .csv file is named "API Item" so it should either represent only the name of the API item or the entire definition of an API item. The current implementation doesn't consistently provide neither first, or the second. Since the purpose of the .csv file is to simply list exported APIs to help notice when something is added or removed from exports, we figured that a name of the API would suffice, and changes to the definition of the API item can be seen in the api.md file.

> When having types and namespaces with the same name, the current implementation simply outputs duplicate lines

I think this change could help reviewers easier see what exactly is added or removed from exports. For example, if a new type and a namespace is added, the reviewer may easier notice if only the namespace was exported and not the interface, or vice versa.

grigasp:
@pmconne, did the last comment answer your questions?

pmconne:
@grigasp yes, thanks.

pmconne (approved):

grigasp (approved):

wgoehrig (approved):

diegoalexdiaz (approved):

dassaf4 (approved):
Nice!

aruniverse (approved):

aruniverse:

PR #7012, opened 2024-07-29
Support showing AccuDraw input fields in a popup at the cursor.

bbastings:
Added method that is called when AccuDraw compass display changes.

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -398,8 +398,8 @@ export class AccuDraw {
     protected _compassSizeInches: number;
     // (undocumented)
     contextSensitive: boolean;
-    // (undocumented)
-    currentState: CurrentState;
+    get currentState(): CurrentState;
+    set currentState(state: CurrentState);
     // @internal (undocumented)
     currentView?: ScreenViewport;
     // @internal (undocumented)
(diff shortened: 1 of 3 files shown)


mathieu-fournier (approved):

pmconne (changes requested):
It would be bad if somebody directly set `currentState` instead of calling `setCurrentState`, right? If so, can you do something to prevent or discourage that?

bbastings:
@pmconne Like what, it's public? Same issue with compassMode and rotationMode. Pretty much relying on the class documentation saying not to call this stuff directly here...

mathieu-fournier:
Why not add public get private set ?

bbastings:
> Why not add public get private set ?

That's a breaking change...

pmconne:
> > Why not add public get private set ?
> 
> That's a breaking change...

You could change`currentState` to a getter+setter, move your `setCurrentState` logic to the setter, delete `setCurrentState`.

bbastings:
> You could changecurrentState to a getter+setter, move your setCurrentState logic to the setter, delete setCurrentState.
@pmconne Done. Didn't think this was a good idea, but since we do it in other setters...

mathieu-fournier (approved):

pmconne (approved):

PR #7023, opened 2024-07-31
Fix error reading ArcGIS Map service capabilities when Kerberos authentication is used

mdastous-bentley:
When Kerberos authentication is enabled on a ArcGIS service, two requests are made: the first one is expected to return 401, and then if all conditions are met, further requests will include windows authentication information.  

It was found the second request made to read the service capabilities would be missing the initial query parameters.


pmconne (approved):

mdastous-bentley:

PR #7026, opened 2024-08-01
exclude internal folders from typedoc processing

ben-polinsky:
Typedoc [entryPoints/strategies](https://typedoc.org/options/input/#entrypoints) don't provide a way to use a barrel file as an entryPoint without significant post-processing to retain our current documentation struture. 

As long as all of our internal implementations are in an internal folder, as specified in #6828 (#6742), adding them to typedoc's [exclude flag](https://typedoc.org/options/input/#exclude) should suffice.


ben-polinsky:
@pmconne

aruniverse (approved):

pmconne:
I need to move PolylineArgs and MeshArgs, apparently they were part of the public API.

pmconne (approved):

PR #7034, opened 2024-08-02
Clean up and promote EntityMetaData

pmconne:
Part of https://github.com/iTwin/itwinjs-backlog/issues/1103.

This was one of the very first APIs ever created for iTwin.js and boy does it show.
Despite still being tagged as `@beta` after all these years, it has always been de facto part of the `@public` API due to its use in `@public` methods like `Entity.forEachProperty` and `Element.getClassMetaData`.

I tried to do minimal cleanup without introducing breaking changes and add/clarify documentation. Mostly I preferred to link to the learning articles in docs/bis/ec; those need some separate cleanup, though.


core/common/src/EntityProps.ts
```diff
@@ -83,69 +84,113 @@ export enum PrimitiveTypeCode {
...
   minimumLength?: number;
+  /** See [[PropertyMetaData.maximumLength]]. */
   maximumLength?: number;
+  /** See [[PropertyMetaData.readOnly]]. */
   readOnly?: boolean;
+  /** See [[PropertyMetaData.kindOfQuantity]]. */
   kindOfQuantity?: string;
+  /** See [[PropertyMetaData.isCustomHandled]]. */
   isCustomHandled?: boolean;
+  /** See [[PropertyMetaData.isCustomHandledOrphan]]. */
```
pmconne, line 136: This is supposed to be determined by `AutoHandledPropertiesCollection::DetectOrphanCustomHandlerProperty`. Nobody ever calls that function. @khanaffan did I overlook something or is this property indeed meaningless and never true?

ColinKerr:
Can we take a look at how this stuff is actually used before we promote to public?

pmconne:
> Can we take a look at how this stuff is actually used before we promote to public?

We can, but I want to reiterate: it is already public and has been for years, beta tag notwithstanding.
[Booster's usage](https://dev.azure.com/bentleycs/Civil-iTwin/_git/Booster?path=/apps/booster/src/backend/imodel/ElementOperations.ts&version=GBmain&line=301&lineEnd=302&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents).

core/common/src/EntityProps.ts
khanaffan: No, you are right. From what i see `s_orphanCustomHandledProperties` will always be empty as `DetectOrphanCustomHandledProperty` is never called from any where.

In addition even if its called the result will be exactly same is `isCustomHandled`. I am not sure why its called "Orphan" when it also check for exactly same customAttribute  presence on the property as the isCustomHandled does.

As of now if you remove this it be fine as it will always be `false`.

khanaffan (approved):

core/backend/src/ClassRegistry.ts
```diff
@@ -95,12 +95,12 @@ export class ClassRegistry {
...
-    const name = entityMetaData.ecclass.split(":");
-    const domainName = name[0];
-    const className = name[1];
+    const nameParts = entityMetaData.ecclass.split(":");
+    const domainName = nameParts[0];
+    const className = nameParts[1];
 
     if (0 === entityMetaData.baseClasses.length) // metadata must contain a superclass
-      throw new IModelError(IModelStatus.BadArg, `class ${name} has no superclass`);
+      throw new IModelError(IModelStatus.BadArg, `class ${nameParts} has no superclass`);
```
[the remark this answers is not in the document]
pmconne: Addressed (same problem a few lines down from this one too).

diegoalexdiaz (approved):

PR #7035, opened 2024-08-02
Add subcategory to appearance on default view creation if not exists

hl662:
A bug was found when testing the pre-release pkgs of core using a viewer and a GPB serving the new 3.7.0 imodelreadrpcinterface. When creating a default view, the categoryIds map would be populated, but the appearances map would not.

Resolved by ensuring subcategories that don't already exist in subCategoriesCache `appearances` would be added 

This won't affect existing behavior where users would add subcategories after view creation (because the override flag would always be set to true outside of default view creation).


hl662:
Thank you @MikeNBentley for detecting the bug and validating the fix through his e2e viewer tests.

pmconne (approved):

full-stack-tests/core/src/frontend/standalone/ViewCreator3d.test.ts
aruniverse, line 1: I know you added this test to catch this next time, but wonder if we should update dta/dpta to open up default view.
That can be a separate pr, just mentioning it here so this doesnt get lost.

common/changes/@itwin/core-frontend/nam-fix-appearance_2024-08-02-13-56.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Add subcategory to appearance on default view creation if id doesn't exist in appearance map",
```
aruniverse, line 5: Is this needed if this is due to a change in a previous pre-release and never seen by users?

PR #7040, opened 2024-08-05
Basic grapheme offset function for visual caret positioning

Alfonso-Martello:
Related to https://github.com/iTwin/itwinjs-backlog/issues/1152.

Introduces a new computeGraphemeOffsets function to allow computing the range of a grapheme. Really, it is just computing the trailing edge (since each range is from the start of the RunLayoutResult to the trailing edge of the grapheme).


pmconne:
I have only skimmed the changes. A couple of things to consider:

1. A given TextBlock will contain many graphemes. The JSON representation of a grapheme is very verbose. Can you reduce it?
1. Not every caller of layoutTextBlock cares about individual graphemes. They shouldn't have to pay the price (in terms of processing and extra data). Can you make it opt-in?

core/backend/src/TextAnnotationLayout.ts
```diff
@@ -244,6 +244,17 @@ function split(source: string): Segment[] {
   return segments;
 }
 
+/**
+ * TypeScript only provides type declarations for Intl.Segmenter if targeting ES2022+.
+ * Use this type as a stand-in until ES2022+ is supported.
+ * @internal
```
ben-polinsky, line 250: Since this interface isn't exported, `@internal` isn't necessary?

core/backend/src/TextAnnotationLayout.ts
```diff
@@ -349,7 +360,43 @@ export class RunLayout {
...
+    const segmenter = new (global as any).Intl.Segmenter(undefined, {granularity: "grapheme"});
+    const graphemes: IntlSegmenterStandInType[] = Array.from(segmenter.segment(content));
+    const graphemeOffsets: GraphemeOffset[] = [];
+    let processedText = "";
+    let prevGraphemeOffset = 0;
+
+    graphemes.forEach((grapheme, index) => {
+      const nextGraphemeCharIndex = graphemes[index+1]?.index ?? content.length;
+      processedText += content.substring(grapheme.index, nextGraphemeCharIndex);
+      const rangeForText = layoutContext.computeRangeForText(processedText, style, baselineShift);
```
pmconne, line 387: Ask @jffmarker if you actually need to include all the preceding characters to get an accurate range for the current grapheme. It's a lot of redundant processing (e.g., for a 6-character run we must cumulatively process 21 characters).

Alfonso-Martello:
> 1. A given TextBlock will contain many graphemes. The JSON representation of a grapheme is very verbose. Can you reduce it?

By reduce, I assume you mean reduce the number of properties? All of the properties are being used. I may be able to remove the `charCount` property. I could combine the `leadingGraphemeOffset` and `trailingGraphemeOffset` properties into just a `range`.
> 2. Not every caller of layoutTextBlock cares about individual graphemes. They shouldn't have to pay the price (in terms of processing and extra data). Can you make it opt-in - or, make it a separate function?

The call to compute grapheme offsets should only occur when the user calls the separate [computeLayoutTextBlockResult](https://github.com/iTwin/itwinjs-core/blob/a30dd73d6734cda36408b0009ba0c116b6717229/core/backend/src/TextAnnotationLayout.ts#L87) function, so only when they are asking for the layout information, not when a caller is just calling `layoutTextBlock`. I can make it even more opt-in by adding a boolean arg required to be true to opt-in. Or are you suggesting a separate function from the "toResult" workflow that a user can call if they explicitly want grapheme information (such that they pass a `RunLayoutResult` and get grapheme info for that `RunLayoutResult`)?

Based on the above questions, I'm starting to like the idea of having a separate function that I can call to get the grapheme information for a RunLayoutResult. This would allow me to do the logic for determining what constitutes a grapheme on my end, and only ask core for the range of the characters. This would allow me to remove the GraphemeOffset interface and decouple grapheme offsets form the layout result.

If you agree I can modify this PR to introduce that function instead.

pmconne:
> By reduce, I assume you mean reduce the number of properties? All of the properties are being used. I may be able to remove the `charCount` property. I could combine the `leadingGraphemeOffset` and `trailingGraphemeOffset` properties into just a `range`.

I mean that this:
```json
graphemeOffsets: [{
  charCount: 1,
  charOffset: 2,
  leadingGraphemeOffset: 3,
  trailingGraphemeOffset: 4,
}, {
  // ...repeat for each grapheme...
```

is very verbose. There are 4 characters that contain actual data, and a ton of characters that are just repeated property names. The same information could be transmitted much more compactly.

pmconne:
> I'm starting to like the idea of having a separate function that I can call to get the grapheme information for a RunLayoutResult. 

Me too.

core/backend/src/TextAnnotationLayout.ts
jffmarker: I would be concerned about glyph shaping and TrueType kerning features. For example:

image
(Word with Aptos (Body) font)

The leading edge of the second character is dependent upon the previous character (notice the 'e' is slightly tucked under the 'T', but not so for the B). I don't have good examples off-hand, but other script languages such as Devanagari and Arabic need to shape glyphs based on the preceding, so I would also expect it's important there.

image
(https://github.com/n8willis/opentype-shaping-documents/blob/master/opentype-shaping-devanagari.md)

Then again, that also influences the trailing position of the previous glyph, so you may have to consider backfilling as well. Seems a shame the browser can't give a list of grapheme positions, and you need to rely on incrementally computing it like this.

core/backend/src/TextAnnotationLayout.ts
Alfonso-Martello: Thanks Ben, this interface has been removed

core/backend/src/TextAnnotationLayout.ts
Alfonso-Martello: I tried it both ways (considering all prior text vs. considering only each grapheme). When considering only each grapheme, the computed offset would become incorrect if glyph shaping/kerning caused a dependency on the previous grapheme (for example, if "Te" was encountered, the cursor would appear in the middle of following graphemes, not on their edges).

Alfonso-Martello:
This PR is now introducing a new `computeGraphemeOffsets` function to allow computing the range of a grapheme. Really, it is just computing the trailing edge (since each range is from the start of the RunLayoutResult to the trailing edge of the grapheme).

core/backend/src/TextAnnotationLayout.ts
ben-polinsky: Thanks!

core/backend/src/test/annotations/TextAnnotation.test.ts
```diff
@@ -551,6 +551,242 @@ describe("layoutTextBlock", () => {
...
+          expect(resultText).to.equal(originalText);
+        }
+      }
+    }
+  });
+
+  describe("grapheme offsets", () => {
+    it("should return an empty array if source type is not text", function () {
+      const textBlock = TextBlock.create({ styleName: "" });
+      const fractionRun = FractionRun.create({ numerator: "1", denominator: "2", styleName: "fraction" });
```
pmconne, line 673: What does/will your text editing tool do when user tries to move caret over a fraction? Is it possible to edit or delete a fraction?

Alfonso-Martello: Right now, the caret API treats a fraction as an atomic unit. It uses the range of the FractionRun to determine the cursor's offset. 

Fractions are not yet implemented with the text editing tool. With what I have right now, it would not be possible to edit a fraction. It would be possible to delete a fraction (this would be the same as removing the entire FractionRun). 

I think it would be very nice to be able to edit fractions in-place, but I have not started thinking about how to implement that.

pmconne: I mainly ask because it may have implications for your new function and interfaces. They are beta, so they can evolve as you iterate, but it may be worth taking some time to consider in what direction they might need to evolve to support fractions, so you don't have to radically revise them later.

ben-polinsky (approved):

core/backend/src/test/annotations/TextAnnotation.test.ts
```diff
@@ -685,122 +921,6 @@ describe("layoutTextBlock", () => {
       expect(r1.justification.xLength()).to.equal(r2.justification.xLength());
     });
   });
```
pmconne, line 923: PR merged while I was reviewing...why was this test deleted?

pmconne: nvm it was moved for some reason.

core/backend/src/test/annotations/TextAnnotation.test.ts
Alfonso-Martello: Spending only the last 30 minutes thinking about it, I put together some pseudo code that should allow the function and interface to remain mostly the same for fractions. I would need to add one more property to the interface to know if the function should get grapheme range's for the numerator or denominator (meaning for each fraction the function would need to be called twice, once for the numerator and once for the denominator). I'm not sure I love that, but it should be possible.

However, there are a lot of big unknowns with text (mainly around international language support), and I am viewing this function and interface as the first steps towards validating those unknowns. I think it is too early to conclude if this approach is “good enough” or not, so it will probably change regardless as we iterate

PR #7042, opened 2024-08-06
Bearing and Azimuth direction support for quantity framework

rschili:
Support for Bearing and Azimuth formatting in the quantity package.
2 Bugfixes for minWidth and spacer fields on format.


core/quantity/src/Formatter/Formatter.ts
```diff
@@ -179,15 +181,15 @@ export class Formatter {
...
+      }
     } else {
       componentText = Formatter.formatMagnitude(compositeValue, spec);
     }
 
     if (spec.format.hasFormatTraitSet(FormatTraits.ShowUnitLabel)) {
       componentText = componentText + spec.format.uomSeparator + label;
-    } else {
-      if (!isLastPart)
-        componentText = `${componentText}:`;
```
rschili, line 190: @anmolshres98 @hl662 @ColinKerr @pmconne 
This is one of two breaking fixes that I had to make to the formatting code in order to support this. There are 2 fields which didn't work as in native, but their description on https://www.itwinjs.org/bis/ec/ec-format/ states that the C++ implementation is correct. Still, fixing that here is a breaking change. Let's discuss how to deal with this.
These are the 2 fields:

- "minwidth". The documentation says it will pad left with zeros, but the code actually took the entire formatted string and padded with spaces. In C++ it applies to each component in a composite while the Typescript code just used it on the overall result. In addition, the implementation in ts was buggy, it did effectively nothing, ever. But fixing it will suddenly make it work.
- "Spacer" which is the thing that separates the pieces in a composite format. Somehow, instead of using the spacer, the code always put a ":" between the pieces plus the spacer. That's not what c++ does and the hard coded : (which is the piece of code highlighted by this comment) is also undocumented and I believe unwanted.

For now I just went ahead and fixed the logic, not caring about the break.....

core/quantity/src/Formatter/Format.ts
```diff
@@ -176,6 +194,18 @@ export class BaseFormat {
         throw new QuantityError(QuantityStatus.InvalidJson, `The Format ${this.name} has an invalid 'stationSeparator' attribute. It should be an empty or one character string.`);
       this._stationSeparator = formatProps.stationSeparator;
     }
+
+    if (undefined !== formatProps.azimuthBase) { // optional; default is 0.0
```
hl662, line 198: if the default is 0.0, shouldn't that be reflected in line 41 above, where the azimuthBasee property is defined? Same with azimuthCounterClockwise

core/quantity/src/Formatter/Format.ts
```diff
@@ -316,6 +350,37 @@ export class Format extends BaseFormat {
       if (undefined === this.units || this.units.length === 0)
         throw new QuantityError(QuantityStatus.InvalidJson, `The Format ${this.name} has a Composite with no valid 'units'`);
     }
+
+    if(this.type === FormatType.Azimuth || this.type === FormatType.Bearing) {
+      // these units cannot be loaded from loadFormatProperties() because they require an async call, and the method signature is already public
+
+      if (undefined !== jsonObj.azimuthBaseUnit) {
+        if (typeof (jsonObj.azimuthBaseUnit) !== "string")
```
hl662, line 358: type validation is already ddone in loadFormatProperties() above - do we need to do this again?

hl662: My bad, I didn't see this is for azimuthBaseUnit, not azimuthBase @@

core/quantity/src/Formatter/Format.ts
```diff
@@ -316,6 +350,37 @@ export class Format extends BaseFormat {
...
+          throw new QuantityError(QuantityStatus.InvalidJson, `The Format ${this.name} has an invalid 'azimuthBaseUnit' attribute. It should be of type 'number'.`);
```
hl662, line 359:
```suggestion
          throw new QuantityError(QuantityStatus.InvalidJson, `The Format ${this.name} has an invalid 'azimuthBaseUnit' attribute. It should be of type 'string'.`);
```

core/quantity/src/test/DirectionFormatting.test.ts
```diff
@@ -0,0 +1,282 @@
...
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import { assert, expect } from "chai";
+import { Format } from "../Formatter/Format";
+import { FormatterSpec } from "../Formatter/FormatterSpec";
+import { Formatter } from "../Formatter/Formatter";
+import { TestUnitsProvider } from "./TestUtils/TestHelper";
+import { FormatProps, UnitProps } from "../core-quantity";
+
+describe.only("Bearing direction tests:", () => {
```
hl662, line 12: remove this `only`

core/quantity/src/Formatter/Formatter.ts
```diff
@@ -355,9 +357,24 @@ export class Formatter {
...
+
+    if(spec.format.minWidth) {
+      formattedValue = this.countAndPad(formattedValue, spec.format.minWidth);
+    }
+
     return formattedValue;
   }
 
+  private static countAndPad(value: string, minWidth: number): string {
+    const regex = /[\d,.]/g;
```
hl662, line 369: what if a user specified a different thousand and decimal separator? would the regex catch this?

hl662: maybe having a separate unit test in NumericFormat.test.ts where a min width is specified and used, alongside a custom decimal separator would be good

core/quantity/src/Formatter/Format.ts
rschili: My thinking was, if I leave it "undefined" above and use a default in code, I can then tell if the value was explicitly set or not. So the serializer would know if it should serialize the value or not.
As alternatives, there could be a separate "isXXXExplicitlySet"-flag which we do often but I don't like that very much. Or I could not serialize if the value matches the default, but in that case the serialized value JSON could be different from the loaded value JSON and code might think it has changed even though it represents the same format....
So out of those options I chose this one..

core/quantity/src/Formatter/Format.ts
rschili: Thanks!

core/quantity/src/test/DirectionFormatting.test.ts
rschili: I'll do a sweep of those once I'm finished with the changes.

core/quantity/src/Formatter/Formatter.ts
rschili: Good point. I copied this from the C++ implementation. I think you're right. I don't expect anybody is really using a different separator than comma and dot so this had gone unnoticed. Could you guys revisit this later at some point? It would have to be adjusted in native as well.

docs/bis/ec/ec-format.md
```diff
@@ -72,9 +75,9 @@ The supported options are:
 
 **thousandSeparator** the character used to separate the integer part of a value in groups of 3 digits. If not defined a localized separator will be used.
 
-**uomSeparator** the character used to separate the magnitude and the Unit label.
+**uomSeparator** the character used to separate the magnitude and the Unit label. For composite formats applies to each component.
```
hl662, line 78:
```suggestion
**uomSeparator** the character used to separate the magnitude and the Unit label. For composite formats, this applies to each component.
```

core/quantity/src/Formatter/Formatter.ts
```diff
@@ -232,7 +234,7 @@ export class Formatter {
       }
     }
 
-    return compositeStrings.join(spec.format.spacer ? spec.format.spacer : "");
+    return compositeStrings.join((spec.format.spacer !== undefined) ? spec.format.spacer : " ");
```
hl662, line 237: Hmm, maybe if a spacer is undefined, it would mean that it shouldn't be spaced? So, I see joining on an empty string like previously would make more sense

hl662: formats have the spacer attribute set to `" "` anyways, so the only way its undefined would be if someone deliberately passed spacer as null to the format props, or if they explicitly want an empty string as their spacer

rschili: good observation. Still, since spacer is allowed to be undefined, I guess there is no way to avoid this check. Or did you have a better idea? Previously, we always added a ":"  between components in addition to the spacer, which I removed. So before it was never really possible to have "no spacing". But now it can be done. I think if somebody really wants it, setting the spacer to "" is a good precondition.

hl662: yes... now its possible to have no spacing we should properly support it. We can keep the check make the spacer set to "" when the check fails, and just gotta properly document that this fix enables no spacing if a user wants it (Then again, if no one ever raised the `:` issue in the past 3 years maybe no one needed it hah)

core/quantity/src/Formatter/Formatter.ts
hl662: I'm okay with both of these changes, seeing as its inherently a fix. The bad behavior was undetectable, or not big enough to warrant for product teams to raise an issue about it. Since it could cause users to assume this bug was intended for a long time, it won't hurt to document these bug fixes in `NextVer.md` though. Anyone else have any thoughts?

core/quantity/src/Formatter/Format.ts
hl662: following up on this, now that we have serialization - is there an intention behind determining if it should serialize the value or not? Based on internal conversations we've also established what the default values should be right? The parser and formatter won't use these attributes anyways unless they are of type `Azimuth`?

rschili: Coming from XML serialization, we do not guarantee, but try to preserve the contents of the file as much as we can. So if the spacer for example is set to " " in the file, even though that is the default, we try to remember that and serialize it back.
So far this has been a good practice, so if the input JSON does not contain an azimuthBase, and we serialize it back to Json, there should not be one. The default is 90, and if the input matches 90 we still want to serialize that back out again.
I like this approach. The code only uses the fields when the type is azimuth, but I'd still try and preserve them if they are not used.
Or would you like to change that? I do not have a strong opinion on this, I can be persuaded :-)

core/quantity/src/test/Formatting.test.ts
```diff
@@ -815,4 +815,122 @@ describe("Formatting tests:", () => {
...
+      },
+    };
+
+    const format = await Format.createFromJSON("test", unitsProvider, formatProps);
+
+    const jsonObj: FormatProps = format.toJSON();
+
+    // Assertions
+    // eslint-disable-next-line guard-for-in
+    for (const key in formatProps) {
```
rschili, line 836: I believe you may be able to use chai's assert.deepEqual here

core/quantity/src/Formatter/Format.ts
```diff
@@ -326,6 +359,37 @@ export class Format extends BaseFormat {
       if (undefined === this.units || this.units.length === 0)
         throw new QuantityError(QuantityStatus.InvalidJson, `The Format ${this.name} has a Composite with no valid 'units'`);
     }
+
+    if(this.type === FormatType.Azimuth || this.type === FormatType.Bearing) {
+      // these units cannot be loaded from loadFormatProperties() because they require an async call, and the method signature is already public
```
ColinKerr, line 364: loadFormatProperties is public but it is in a beta class

core/quantity/src/Formatter/Format.ts
hl662: I like that approach, it's just we already have existing fields that are created by default (like uom, deicmal separator, etc), so if the input json doesn't contain those fields, the output json would. If we've established default values for these 2 new properties, I think concretely putting it in the code will reflect that's the default we've established, rather than leaving the onus to the user to interpret and enter manually. I also don't have a strong opinion on this either so I'd rather have another person chime in 😄

core/quantity/src/test/Formatting.test.ts
hl662: I originally used that as well, but it can't be done because of `formatTraits`. deepEqual expects the order of two compared arrays to match too, which is something we reasonably don't care about. That's why I wrote down logic within for comparing members of the 2 arrays, without caring about the order

core/quantity/src/Formatter/Formatter.ts
```diff
@@ -414,10 +437,104 @@ export class Formatter {
...
+        prefix = "S";
+
+      if (quadrant === 0 || quadrant === 3)
+        suffix = "E";
+
+      if (quadrant === 1 || quadrant === 2)
+        suffix = "W";
+
+      // special case, if in quadrant 2 and value is very small, turn suffix to E because S00E is preferred over S00W
+      if (quadrant === 2 && magnitude < 0.00000000001) // 1e-11 is a small value that is close to 0
```
ColinKerr, line 481: I'm kind of surprised that this is the first place where we've had to use some small value to represent 'almost zero'.

In this case I'm not sure what you've done is enough.  If you never want to show S00W then shouldn't you be comparing the formatted value rather than the raw magnitude?  The precision of the formatted output might treat much larger values as zero

core/quantity/src/Formatter/Formatter.ts
```diff
@@ -414,10 +437,104 @@ export class Formatter {
...
+        suffix = "E";
+
+      return {magnitude, prefix, suffix: suffix!};
+    }
+
+    if (type === FormatType.Azimuth) {
+      let azimuthBase = quarterRevolution; // default base is North
```
ColinKerr, line 488: Is this assuming the persisted values base is east?

core/quantity/src/Formatter/FormatterSpec.ts
```diff
@@ -100,7 +104,23 @@ export class FormatterSpec {
    */
   public static async create(name: string, format: Format, unitsProvider: UnitsProvider, inputUnit?: UnitProps): Promise<FormatterSpec> {
     const conversions: UnitConversionSpec[] = await FormatterSpec.getUnitConversions(format, unitsProvider, inputUnit);
-    return new FormatterSpec(name, format, conversions, inputUnit);
+    const spec = new FormatterSpec(name, format, conversions, inputUnit);
```
ColinKerr, line 107: it seems wrong that you have to do extra to the FormatterSpec after creating it even though the information is being passed into the constructor.

docs/bis/ec/ec-format.md
```diff
@@ -14,6 +14,8 @@
 - `fractional`
 - `scientific`
 - `station`
+- `bearing`
```
ColinKerr, line 17: we have to be careful here ... this is not supported in EC at this point.  When it is we will need to identify that it's only available in a new EC version ... I guess 3.3

core/quantity/src/Formatter/Format.ts
rschili: That is true, I did consider making it async, but there is one call from ECFormat.fromJSONSync()
... And now I realize, that fromJSONSync in metadata has a copy of what this method here does, so we'll have to put a duplicate of this code there.... meh. I'll bring this up with Anmol and Nam.

core/quantity/src/Formatter/Formatter.ts
rschili: I'll have to think about that for a bit. Imagine the unit is whole degrees, and the value is 0.4. Which formats to "0" can/should we consider that zero? I feel like I'm not qualified to make that decision. We do have an AlmostEqual() method in native quantity.

core/quantity/src/Formatter/Formatter.ts
rschili: yes. North is quarter revolution, which in degrees would be 90.

core/quantity/src/Formatter/Format.ts
hl662: I've added dedupe of method calls to the backlog as part of quantity formatting.

docs/bis/ec/ec-format.md
```diff
@@ -89,6 +92,14 @@ Supported scientific type:
 
 **stationSeparator** the character used to separate the station and off set portions of a `station` formatted value.
 
+**azimuthBase** A numeric value indicating the base when type is set to azimuth. It fet, the `azimuthBaseUnit` has to also be set to indicate which unit this value is in. Defaults to north (quarter rotation). The value is indicated from east counter-clockwise.
```
hl662, line 95:
```suggestion
**azimuthBase** A numeric value indicating the base when type is set to azimuth. If set, the `azimuthBaseUnit` has to also be set to indicate which unit this value is in. Defaults to north (quarter rotation). The value is indicated from east counter-clockwise.
```

core/quantity/src/Formatter/FormatterSpec.ts
rschili: Fixed

core/quantity/src/Formatter/Formatter.ts
ColinKerr: Well you don't make that decision the format's definition makes the decision how many decimal digits to show.  That's why I was suggesting that you have to look at it after formatting the double value instead of looking at the double value.

docs/changehistory/NextVersion.md
```diff
@@ -6,6 +6,14 @@ publish: false
...
 Table of contents:
 
+- [Quantity](#quantity)
+- [Electron 32 support](#electron-32-support)
+
+## Quantity
+
+- The `minWidth` property on FormatProps now works as documented.
+- The `spacer` property on FormatProps now indicates the space used between composite components, it defaults to a single space, and there is no longer a ':' prepended. If a ':' spacer is desired, `spacer` has to be set accordingly. This is to streamline the behavior with the documentation and native APIs.
+- Added support for bearing and azimuth format types (e.g. bearing `N45°30'10"E`). A new phenomenon "Direction" for these will be added to our units library soon, but they work just as well with the angle phenomenon for now. Persistence values for both bearing and azimuth are to be provided counter-clockwise from an east-base (inspired by PowerPlatform).
```
aruniverse, line 16: should we be mentioning "PowerPlatform"? how much of itwinjs consumers know about that? isnt it an internal concept?

core/quantity/src/Formatter/Interfaces.ts
```diff
@@ -25,9 +25,13 @@ export interface FormatProps {
   readonly scientificType?: string; // conditionally required
   readonly stationOffsetSize?: number; // conditionally required
   readonly stationSeparator?: string;
+  readonly azimuthBase?: number; // value is specified from east counter-clockwise
```
aruniverse, line 28: should prob add comments above using `/** ... */` syntax so they show up in vscode, this can be done in a separate pr.

docs/changehistory/NextVersion.md
rschili: It is the reason why we chose this default instead of North-Clockwise. Historically, MicroStation has done that, and much of our data is probably stored in that way. So.... not sure. By not mentioning PowerPlatform we'd be making this choice seem arbitrary. What do you think?

aruniverse: I searched for [Power Platform](https://bentleysystems.service-now.com/community?id=kb_article&sysparm_article=KB0099402) and I guess this is good enough of an explanation for it, so im ok leaving this in.

core/quantity/src/Formatter/Format.ts
rschili: I changed some of the fields to have a getXXOrDefault getter alongside the normal getter, which will fill in the default if the value is undefined. For azimuthBase I couldn't do it because the default depends on the unit. But I think that's a good compromise.

core/quantity/src/Formatter/Formatter.ts
rschili: closing the conversation. Nobody seems to object.

docs/bis/ec/ec-format.md
rschili: I added a note saying this is only supported in the quantity package at the moment

core/quantity/src/Formatter/Interfaces.ts
rschili: changed it.

core/quantity/src/Formatter/Formatter.ts
rschili: For bearing, we usually use more than one presentation unit in a composite.
That formatting happens after this code.
So what I had to do to achieve the logic you describe was I had to convert the unit and compare it using the format's precision. The check is merely a preference, both results would be fine, but Kevin said "E" is preferred.
The adjusted code is a bit gross, but the alternative would be breaking encapsulation of bearing inside this method, and putting some bearing logic outside after the formatting code, which looked even more gross.
Please have a look at the updated code.

ColinKerr (approved):

PR #7055, opened 2024-08-08
Create Mesh Export Service Integration Test

andremig-bentley:
WIP

Currently facing webpack issues, creating this PR to get help. Not ready for review.

### DIFF
--- a/extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
+++ b/extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
@@ -1,0 +1,126 @@
+// /*---------------------------------------------------------------------------------------------
+//  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+//  * See LICENSE.md in the project root for license terms and full copyright
+//  *notice.
+//  *--------------------------------------------------------------------------------------------*/
+
+const path = require('path');
+const {globSync} = require('glob');
+const frontendLib = path.resolve(__dirname, '../../../lib/cjs');
+const Dotenv = require('dotenv-webpack');
+const webpack = require('webpack');
+
+function createConfig(shouldInstrument) {
+  const config = {
+    mode: 'development',
+    entry:
+        globSync(path.resolve(frontendLib, 'integration-test/**/*.test.js'), {
+          windowsPathsNoEscape: true,
+        }),
+    output: {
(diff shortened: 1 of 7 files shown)


extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
```diff
@@ -0,0 +1,126 @@
+// /*---------------------------------------------------------------------------------------------
```
andremig-bentley, line 1: webpack-config is here.

aruniverse: i think you want to use the [CopyExternalsPlugin](https://github.com/iTwin/itwinjs-core/blob/master/tools/webpack-core/src/plugins/CopyExternalsPlugin.ts#L94)

andremig-bentley: I've added the plugin, I'm not sure if it's done correctly, I couldn't find examples of it being used in a webpack-config. If it is set up correctly, it doesn't seem to have done much. By removing the call to externalsPreset{node: true}, webpack still doesn't successfully compile/bundle, citing the same errors as before. The NodePolyfillPlugin above it does help to reduce the number of errors signigicantly (250 -> 110), but there are still many present. When the externalsPreset is present, webpack still compiles correctly, but then I cant run the tests and I get the same "require is not defined" error.

extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
```diff
@@ -0,0 +1,131 @@
...
+//  * See LICENSE.md in the project root for license terms and full copyright
+//  *notice.
+//  *--------------------------------------------------------------------------------------------*/
+
+const path = require('path');
+const {globSync} = require('glob');
+const frontendLib = path.resolve(__dirname, '../../../lib/cjs');
+const Dotenv = require('dotenv-webpack');
+const webpack = require('webpack');
+const copyExternalsPlugin =
```
aruniverse, line 12: you need to add a dev dep on this and import it from the pkg not this super deep import

andremig-bentley: I believe that should be fixed now, but same behavior as my previous comment.

andremig-bentley:
Closing this PR as it's no longer necessary. Thank you for the help!

PR #7058, opened 2024-08-08
attempt to improve ao's default value

danielzhong-bentley:
This pull request addresses the issue where the current Ambient Occlusion (AO) implementation has a noticeable "glow" effect. To resolve this, the following changes have been made:

1. **Changed Default Value of Z Length Cap**: Lowered the default value to reduce the "glow" effect.
https://github.com/iTwin/itwinjs-core/blob/e49833447653a3e98c332120d7070bfb823d0726/core/common/src/AmbientOcclusion.ts#L38

2. **Updated Slider Range and Step Value**: Adjusted the slider to accommodate the new Z length cap value more effectively.
https://github.com/iTwin/itwinjs-core/blob/e49833447653a3e98c332120d7070bfb823d0726/test-apps/display-test-app/src/frontend/AmbientOcclusion.ts#L72

3. **Modified Default Bias Value**: Tweaked the default bias value to enhance the AO effect without introducing artifacts.
https://github.com/iTwin/itwinjs-core/blob/e49833447653a3e98c332120d7070bfb823d0726/core/common/src/AmbientOcclusion.ts#L37

#### Before:
image

#### After:
Snipaste_2024-08-08_14-29-18

### DIFF
--- a/core/common/src/AmbientOcclusion.ts
+++ b/core/common/src/AmbientOcclusion.ts
@@ -34,8 +34,8 @@ export namespace AmbientOcclusion {
 
   /** Describes the properties with which ambient occlusion should be drawn. These properties correspond to a horizon-based ambient occlusion approach. */
   export class Settings implements Props {
-    private static _defaultBias: number = 0.25;
-    private static _defaultZLengthCap: number = 0.0025;
+    private static _defaultBias: number = 0.5;
+    private static _defaultZLengthCap: number = 0.00001385;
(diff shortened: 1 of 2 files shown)


DStradley:
@danielzhong-bentley I got an ImageTests run for AO and found that this change basically disabled the AO effect for all the images.  See https://github.com/iTwin/itwin-graphics-backlog/issues/360 for an example.

PR #7066, opened 2024-08-09
display-test-app key-in to attach a secondary iModel

pmconne:
`toggleExternalIModel` was written as a prototype long ago for use cases like version comparison, for adding a second IModelConnection to be visualized in a viewport. It was implemented using `TiledGraphicsProvider`. Other folks have copied this approach.

The existing implementation only supported V1 tiles (1 tileset per model, with tiles generated on demand), not V4 tiles (1 tileset for the whole iModel, with tiles published in advance). I simplified the implementation to use `ViewCreator3d` to create a view that displays all the spatial models and categories of the secondary iModel. So, now it works with both V1 and V4 tiles.

This feature is now invokable using the new `dta imodel attach` keyin.


markschlosseratbentley (approved):

pmconne:
@ben-polinsky @grigasp nothing to do with my changes
```
Invalid Link: $docs link found in markdown file './presentation/advanced/localization.md'. Use relative links for md -> md linking.
One or more invalid links found.
/Volumes/Builds/vsts_work_a/76/itwinjs-core/node_modules/@bentley/bemetalsmith-link-helpers/dist/index.js:83
            throw new Error("One or more invalid links found.");
```

ben-polinsky:
> /Volumes/Builds/vsts_work_a/76/itwinjs-core/node_modules/@bentley/bemetalsmith-link-helpers/dist/index.js:83

@aruniverse It's the localization change we merged. The link works, I see no reason for enforcing this rule. Maybe we should revert https://github.com/iTwin/itwinjs-core/pull/6999 until I make the change in BeMetalsmith.

aruniverse:
> > /Volumes/Builds/vsts_work_a/76/itwinjs-core/node_modules/@bentley/bemetalsmith-link-helpers/dist/index.js:83
> 
> @aruniverse It's the localization change we merged. The link works, I see no reason for enforcing this rule. Maybe we should revert #6999 until I make the change in BeMetalsmith.

if that pr is gonna issues for others, lets revert it in this pr. If not, lets force merge this one if docs fail, and just pr a fix for BeMetalsmith (if this is something we can do quickly)

PR #7077, opened 2024-08-12
Create Mesh Export Service Integration Test

andremig-bentley:
Create an integration test which calls the MES using a test user and determines if the tile version of the requested model is supported by the frontend under which the test is running. The test will:

- Start an export from the MES
- Once that export is complete, obtain the tileset from that export
- Check the tile version of the provided tileset
- Check if the frontend under which the test is running will support rendering the tileset

While this test in and of itself is not the most useful or robust test, it serves two major purposes. Once the integration testing framework is set up, it will be significantly easier to expand upon the test that is here and create more robust tests. Additionally, the test can serve as an example to other teams of a way tile version compatibility can be checked in their own use cases.


hl662:
Where will this test be run? As part of the iTwin.js Integration pipeline, or an external repo?

markschlosseratbentley:
> Where will this test be run? As part of the iTwin.js Integration pipeline, or an external repo?

Currently undecided.  Most likely not iTwin.js integration pipeline.

andremig-bentley:
> Does it really make sense to have this in this repo then?

That's part of the discussion we're still having. Based on conversations with @arvindsvenkat, we will most likely want it to be an automated test, but we aren't set on where exactly, yet. Open to any suggestions.

arvindsvenkat:
> > > Where will this test be run? As part of the iTwin.js Integration pipeline, or an external repo?
> > 
> > 
> > Currently undecided. Most likely not iTwin.js integration pipeline.
> 
> Does it really make sense to have this in this repo then?

@aruniverse, This test is intended to prevent the error that was encountered when the export tile version changed and wasn't supported by the iTwin viewer. I would assume it's directly correlated to upgrades in the iTwin.js version and the Mesh Export Service exports.

In my opinion, it should be run in the iTwin.js Integration test pipeline and also in the Mesh Export Service release pipeline (possibly scheduled in QA).

cc: @andremig-bentley  @markschlosseratbentley , @danieliborra

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      format: "IMDL",
+      urlPrefix: "qa-",
+    };
+
+    await IModelApp.startup();
+  });
+
+  after(async () => IModelApp.shutdown());
+
+  it ("should obtain a tileset with a tile version supported by the frontend", async () => {
```
markschlosseratbentley, line 35: As discussed in the meeting with @arvindsvenkat - this test should probably request a new export.

markschlosseratbentley (changes requested):

pmconne:
When exactly will this test actually run?
`rush test:integration` expects a script named `test:integration`, which your package.json lacks.
The iTwin.js Integration - GitHub job runs a hard-coded set of integration tests, which doesn't include this one.

andremig-bentley:
> When exactly will this test actually run? `rush test:integration` expects a script named `test:integration`, which your package.json lacks. The iTwin.js Integration - GitHub job runs a hard-coded set of integration tests, which doesn't include this one.

After further discussions, we have decided to add this test to the iTwin.js integration pipeline. The command to run the test has been changed to `test:integration:tileversion` to follow the general integration test naming conventions in core, but we don't want it to run automatically with `rush test:integration`, so that command does not exist in this package. As of now, this PR will only include the addition of the test, and the pipeline integration will be handled in a later PR.

Additionally, based on discussions with @arvindsvenkat and @markschlosseratbentley, the test has been changed to begin by kicking off an export, then polling for the status of that export, and once it is complete, going forward with the tile version testing on that new export, ensuring we get the latest tile version information from the MES.

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: These changes have been made and are included in the most recent commit.

markschlosseratbentley (approved):

aruniverse (changes requested):
where is this actually going to run?
i dont see any pipeline changes. you can grep for this within the repo and look at the other tests as an example

pmconne:
> where is this actually going to run? i dont see any pipeline changes. you can grep for this within the repo and look at the other tests as an example

I had the same question. @andremig-bentley later stated above:
> After further discussions, we have decided to add this test to the iTwin.js integration pipeline. [...snip...]
> The command to run the test has been changed to test:integration:tileversion to follow the general integration test naming conventions in core, but we don't want it to run automatically with rush test:integration, so that command does not exist in this package. As of now, this PR will only include the addition of the test, and the pipeline integration will be handled in a later PR.

I'm not sure where the further discussions occurred or whether you were involved in them. I was not.

andremig-bentley:
> where is this actually going to run? i dont see any pipeline changes. you can grep for this within the repo and look at the other tests as an example

We had decided to have this PR include only the addition of the test, with pipeline changes coming later as time allowed. If you think we should roll it all into this PR, that can be done.

andremig-bentley:
@pmconne @aruniverse 
To clear up any confusion, since the goal of this test is to validate the interoperability between the frontend here and the MES backend, we have decided that this test would be useful to run in two locations: As part of the Mesh Export Service Release Pipeline, and pertinent to this PR, in itwinjs-core. We think it is useful to run it here to validate that any frontend changes do not unintentionally break this connection. 

Initially, we had chosen to only include it in the iTwinjs Integration - Github job as a matter of simplicity. However after discussing more with @markschlosseratbentley, we have decided that we will add it as well to `rush test:integration` as it would be just as useful there.

Additionally, we had also considered having this PR only include the actual addition of the test and not the changes to the pipeline etc. needed to add the test to the iTwinjs Integration - Github job and `rush test:integration` due to prioritization of other work. Our thought being we could finish this PR, have the test in place, and work on actually connecting it to the correct testing infrastructure as time allowed. However, it seems that it would make more sense to have this PR include all necessary changes for the test to run, so those changes are now WIP and will be included in this PR.

Due to this change in plan and the PR now being considered WIP, I am converting it back to draft and will re-request reviews when it is ready. Let me know if there are any other questions I can answer in the mean time.

extensions/frontend-tiles/package.json
```diff
@@ -63,6 +67,7 @@
     "sinon": "^17.0.2",
     "source-map-loader": "^4.0.0",
     "typescript": "~5.3.3",
-    "webpack": "^5.76.0"
+    "webpack": "^5.76.0",
+    "dotenv-webpack": "^8.1.0"
```
aruniverse, line 71: lets alphabetize our deps

extensions/frontend-tiles/src/integration-test/TestContext.ts
```diff
@@ -0,0 +1,57 @@
...
+  private async initialize() {
+    const oidcConfig = {
+      clientId: process.env.MES_OIDC_CLIENT_ID || "",
+      redirectUri: process.env.OIDC_REDIRECT || "",
+      scope: process.env.MES_OIDC_SCOPE || "",
+      authority: process.env.MES_OIDC_AUTH_URL || "",
+    };
+
+    const user = {
+      email: process.env.mes_integration_testuser || "",
```
aruniverse, line 32: why are these env vars all lowercase vs the others above which are all capitalized?

extensions/frontend-tiles/src/integration-test/TestContext.ts
```diff
@@ -0,0 +1,57 @@
...
+import { getAccessTokenFromBackend } from "@itwin/oidc-signin-tool/lib/cjs/frontend";
+
+export class TestContext {
+  private _accessToken: AccessToken | undefined;
+  private static _instance?: TestContext = undefined;
+
+  private constructor() { }
+
+  public static async instance(): Promise<TestContext> {
+
```
aruniverse, line 11: nuke empty line, fix whitespacing on line 40 too

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,108 @@
+import { IModelApp } from "@itwin/core-frontend";
+import { expect } from "chai";
+import { TestContext } from "./TestContext";
+import { GraphicRepresentationStatus } from "../GraphicsProvider/GraphicRepresentationProvider";
+
+async function getExport(exportId: string, accessToken: string, urlPrefix?: string): Promise<any> {
```
aruniverse, line 6: any? we dont have this type defined anywhere?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  try {
+    const response = await fetch(url, { headers });
+    const result = await response.json();
+    return result;
+  } catch (err) {
+    return undefined;
+  }
+}
+
+describe("MeshExportServiceIntegrationTest", () => {
```
aruniverse, line 25: dont use the global, import this

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  let accessToken: any;
+  let urlPrefix: any;
+  let exportId: any;
+
+  before(async () => {
```
aruniverse, line 30: dont use the global, import this

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    // obtain access token
+    const testContext = await TestContext.instance();
+    accessToken = testContext.getAccessToken();
+
+    const iModelId = process.env.MES_IMODEL_ID || "";
+    urlPrefix = process.env.imjs_url_prefix || "";
```
aruniverse, line 37: inconsistent casing

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      },
+      body: JSON.stringify({
+        iModelId,
+        changesetId: "",
+        exportType:"IMODEL",
+      }),
+    };
+
+    // initiate mesh export
+    const response = await fetch(`https://qa-api.bentley.com/mesh-export/`, requestOptions);
```
aruniverse, line 58: why is the env hardcoded here?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    const result = JSON.parse(JSON.stringify(await response.json()));
+    exportId = result.export.id;
+  });
+
+  after(async () => IModelApp.shutdown());
```
aruniverse, line 63: dont use the global, import this

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  it ("should obtain a tileset with a tile version supported by the frontend", async () => {
+    let exportComplete = false;
+    let tilesetUrl: any;
+
+    // poll for export status
+    const start = Date.now();
+    while (Date.now() - start < 90000) {
```
aruniverse, line 71: what is this magical/mysterious number?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (!exportComplete) {
+      throw new Error("Export did not complete in time");
+    }
+
+    if (!tilesetUrl) {
+      throw new Error("Url is undefined");
+    }
+
+    // obtain tileset from url
+    const tileset = JSON.parse(JSON.stringify(await (await fetch(tilesetUrl)).json()));
```
aruniverse, line 93: wut

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    // obtain major tile version from the returned tileset (removing minor version from the string)
+    const tilesetMajorTileVersion = tileset.asset.version.slice(0,4);
+
+    // obtain frontend major tile version
+    const frontendMajorTileVersion = IModelApp.tileAdmin.maximumMajorTileFormatVersion;
```
aruniverse, line 99: do we really need IModelApp? why cant we just use TileAdmin directly?

extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
```diff
@@ -0,0 +1,73 @@
...
+//  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+//  * See LICENSE.md in the project root for license terms and full copyright
+//  *notice.
+//  *--------------------------------------------------------------------------------------------*/
+
+const path = require('path');
+const {globSync} = require('glob');
+const frontendLib = path.resolve(__dirname, '../../../lib/cjs');
+const Dotenv = require('dotenv-webpack');
+const copyExternalsPlugin =
```
aruniverse, line 11: what do you need this plugin for?

extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
```diff
@@ -0,0 +1,73 @@
...
+      },
+      enforce: 'post',
+    });
+  }
+
+  return config;
+}
+
+// Runs webpack once for each config in the export array
+module.exports = [createConfig(false), createConfig(true)]
```
aruniverse, line 73: lets not bother doing any of the instrumentation stuff
just remove that and only export 1 config

common/config/azure-pipelines/templates/setup-integration-users.yaml
```diff
@@ -18,6 +18,8 @@ steps:
     echo "##vso[task.setvariable variable=test_bing_maps_key;]$bing_key"
     echo "##vso[task.setvariable variable=test_mapbox_key;]$mapbox_key"
     echo "##vso[task.setvariable variable=imjs_cesium_ion_key;]$cesium_ion_key"
+    echo "##vso[task.setvariable variable=mes_testuser;]$mes_user"
```
aruniverse, line 21: do we need a separate user account & pw?
couldnt use the exisisting test users?

common/config/azure-pipelines/templates/integration-test-steps.yaml
```diff
@@ -173,3 +180,10 @@ steps:
       testResultsFiles: core/electron/lib/test/*_junit_results.xml
       testRunTitle: "Core - Electron Tests - ${{ parameters.nodeVersion }}"
     condition: succeededOrFailed()
+
+      - task: PublishTestResults@2
+    displayName: "Publish Frontend Tiles Test Results"
+    inputs:
+      testResultsFiles: extensions/frontend-tiles/lib/test/*_junit_results.xml
```
aruniverse, line 187: what with the `*`? do we not know the file path?

ben-polinsky: timestamps?

common/config/azure-pipelines/templates/setup-integration-users.yaml
andremig-bentley: I'm currently using this separate user since it was one I know we had control over and could therefore give access to the right imodels to use in the test, etc.

extensions/frontend-tiles/src/integration-test/TestContext.ts
andremig-bentley: fixed

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: mistake, fixed.

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: This was originally what we were using as a timeout for the test, it's now been changed to five minutes on the suggestion of @arvindsvenkat, and it has been moved to the certa.json file. This while loop now doesn't include any magical numbers and only runs based on the status of the export.

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: accidental hold out from earlier iterations, it's been pared down to only what's necessary now.

extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
andremig-bentley: Wasn't needed, accidental hold out from earlier iterations, it's been removed.

extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
andremig-bentley: fixed

common/config/azure-pipelines/templates/integration-test-steps.yaml
andremig-bentley: fixed to now point to the specific file.

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: I'm not sure what you mean by the global here, what should I be importing?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: same question, what should I be importing here?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: Should I be importing `IModelApp.shutdown`?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
aruniverse: im talking about `describe`, `before`, `after`, `it`, `test`, etc

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
aruniverse: extract the type out, mark it as internal, and re-use it here

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
aruniverse: imo, you went the wrong way.
all the env vars should be capitalized. this is the standard across the entire monorepo, why start a new trend?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
aruniverse: why cant you create your own instance of TileAdmin instead?

common/config/azure-pipelines/templates/setup-integration-users.yaml
aruniverse: I think we should work on getting the test accounts access to these models, instead of creating another set of credentials

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: I see these being imported for tests running using vitest instead of certa, which currently this one is set up to run with. Should this be changed to run with vitest? Unless I'm missing it, I don't see other tests run with certa which import describe, etc.

common/config/azure-pipelines/templates/setup-integration-users.yaml
andremig-bentley: The existing IMJS_REGULAR_TEST_USER now has access to the model we will use to test this.

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: This was fixed in the latest commit

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: Due to current difficulty using the oidc-signin-tool with vitest, we've elected to keep the test running with certa. Because of that, I don't believe we need to import these things here, let me know if I'm mistaken on that.

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: I was actually incorrect. We don't have this type defined anywhere in the frontend-tiles package. We have a very similar type defined in `GraphicRepresentationProvider.ts`, but that one is the response back from using the get-exports API, whereas here I am using the get-export (no s) API. That structure is slightly different, and as far as I can see, is not defined anywhere here.

common/config/azure-pipelines/templates/setup-integration-users.yaml
aruniverse, line 1: lets revert the whitespace changes in this file to clean up the diff

extensions/frontend-tiles/src/integration-test/TestContext.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  public getAccessToken(): AccessToken | undefined {
+    return this._accessToken;
+  }
+
+  private async initialize() {
+    const urlPrefix = process.env.IMJS_URL_PREFIX || "";
+    const oidcConfig = {
+      clientId: process.env.MES_OIDC_CLIENT_ID || "",
```
aruniverse, line 25: lets make sure this env var follows the same prefix like all the other env vars

extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
```diff
@@ -0,0 +1,48 @@
...
+//  * See LICENSE.md in the project root for license terms and full copyright
+//  *notice.
+//  *--------------------------------------------------------------------------------------------*/
+
+const path = require('path');
+const {globSync} = require('glob');
+const frontendLib = path.resolve(__dirname, '../../../lib/cjs');
+const Dotenv = require('dotenv-webpack');
+
+module.exports = [{
```
aruniverse, line 12: since we dont have multiple configs, lets just return a single config object, not an array

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
aruniverse: are we going to type this out?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
aruniverse: are they not exported from chai or mocha?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  let urlPrefix: any;
+  let exportId: any;
+
+  before(async () => {
+
+    // obtain access token
+    const testContext = await TestContext.instance();
+    accessToken = testContext.getAccessToken();
+
+    const iModelId = process.env.MES_IMODEL_ID || "";
```
aruniverse, line 35: lets try to stay consistent with env var prefixes

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      body: JSON.stringify({
+        iModelId,
+        changesetId: "",
+        exportType:"IMODEL",
+      }),
+    };
+
+    // initiate mesh export
+    const response = await fetch(`https://${urlPrefix}api.bentley.com/mesh-export/`, requestOptions);
+    const result = await response.json();
```
aruniverse, line 57: shouldnt you check the response was ok first?

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    exportId = result.export.id;
+  });
+
+  it ("should obtain a tileset with a tile version supported by the frontend", async () => {
+    let exportComplete = false;
+    let tilesetUrl: any;
+
+    // poll for export status
+    while (!exportComplete) {
```
aruniverse, line 66: shoudlnt you wait a few seconds to getExport, because otherwise couldnt this be a bit too spammy?

aruniverse (commented):
looking much better and close to merging

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
aruniverse: https://github.com/iTwin/itwinjs-core/pull/7077#discussion_r1788255530

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
aruniverse: https://github.com/iTwin/itwinjs-core/pull/7077#discussion_r1788255530

extensions/frontend-tiles/package.json
```diff
@@ -46,6 +48,8 @@
     "@itwin/core-common": "workspace:*",
     "@itwin/core-frontend": "workspace:*",
     "@itwin/core-geometry": "workspace:*",
+    "@itwin/core-webpack-tools": "workspace:*",
```
aruniverse, line 51: we dont seem to be using this dep here anymore?

common/config/azure-pipelines/templates/setup-integration-users.yaml
andremig-bentley: removed in latest commit

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: I don't believe it's necessary to type it out, as it's only used here for this test. If that's a sticking point, I can type it out maybe in `GraphicRepresentationProvider.ts` and export it here.

extensions/frontend-tiles/src/integration-test/TestContext.ts
andremig-bentley: these env variables have now been changed to start with IMJS_MES_INTEGRATION so that they still begin with the IMJS prefix, but they are still also specified as part of the mes integration test so that there is no accidental value mix up when running these in a pipeline.

extensions/frontend-tiles/src/integration-test/utils/webpack-config.js
andremig-bentley: fixed in latest commit.

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: check has been added in latest commit

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: a three second delay has been added in the latest commit

extensions/frontend-tiles/package.json
andremig-bentley: removed in latest commit

extensions/frontend-tiles/src/integration-test/IntegrationTest.test.ts
andremig-bentley: They are exported from mocha, but I don't think it's necessary to import them in this case. I don't see any other tests which still use certa importing them - example [here](https://github.com/iTwin/itwinjs-core/blob/18403314367ce0879c609d308931823050f77c30/extensions/frontend-tiles/src/test/ModelGroup.test.ts#L82). Additionally, when I did import them from mocha, I was hit with webpack errors when trying to wrap them up.

extensions/frontend-tiles/src/integration-test/TestContext.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  public getAccessToken(): AccessToken | undefined {
+    return this._accessToken;
+  }
+
+  private async initialize() {
+    const urlPrefix = process.env.IMJS_URL_PREFIX || "";
+    const oidcConfig = {
+      clientId: process.env.IMJS_MES_INTEGRATION_OIDC_CLIENT_ID || "",
```
eringram, line 25: Just an optimization: might be worth it to check at the beginning of `initialize()` whether the env variables IMJS_MES_INTEGRATION_OIDC_CLIENT_ID, IMJS_TEST_REGULAR_USER_NAME, and IMJS_TEST_REGULAR_PASSWORD exist. Because if they don't exist, the test will definitely fail so there's no need for retries (doesn't apply to the url prefix, which can correctly default to "")

eringram (approved):

andremig-bentley:
This PR will be closed with this comment. This is due to significant scope-creep, a very large amount of conversation as we narrowed in on what we wanted this PR to accomplish that will be difficult to decipher by someone in the future, and the fact that it has existed for an incredibly long time untouched, has gotten stale, and accumulated some significant conflicts. A new PR will be created soon that will be much cleaner from the start and will better serve us in the future.

PR #7079, opened 2024-08-13
Add test to assert that we don't fail on textures without TextureIds

nick4598:
imodel-native: https://github.com/iTwin/imodel-native/pull/832


nick4598:
Tried adding the problematic json to my test db, but it doesn't fail to open. The code I added OnLoadedJsonProperties is not problematic in this case, because I check if TextureId is a numericMember before doing anything.


The line of code `auto havePatternMap = patternMap.IsValid() && patternMap.IsPatternEnabled() && patternMap.GetTextureId().IsValid();` in Render.cpp

calls `GetTextureId()` which calls `return DgnTextureId(m_value[RENDER_MATERIAL_TextureId].asUInt64());`

`m_value[RENDER_MATERIAL_TextureId]` results in the IsNull() BeJsConst.

pmconne:
> The line of code `auto havePatternMap = patternMap.IsValid() && patternMap.IsPatternEnabled() && patternMap.GetTextureId().IsValid();` in Render.cpp
> 
> calls `GetTextureId()` which calls `return DgnTextureId(m_value[RENDER_MATERIAL_TextureId].asUInt64());`
> 
> `m_value[RENDER_MATERIAL_TextureId]` results in the IsNull() BeJsConst.

Can you write a test that opens the db and makes the calls you described to trigger the bug?

nick4598:
> > The line of code `auto havePatternMap = patternMap.IsValid() && patternMap.IsPatternEnabled() && patternMap.GetTextureId().IsValid();` in Render.cpp
> > calls `GetTextureId()` which calls `return DgnTextureId(m_value[RENDER_MATERIAL_TextureId].asUInt64());`
> > `m_value[RENDER_MATERIAL_TextureId]` results in the IsNull() BeJsConst.
> 
> Can you write a test that opens the db and makes the calls you described to trigger the bug?

Thanks, it was much easier to write a test on the C++ side. I pushed a change there and will close this PR, feel free to comment there if you have any concerns.

PR #7080, opened 2024-08-13
Add openGroup() & writeToFile()  to SqliteChangesetReader

khanaffan:
This change supports the use case for version comparison, where it is necessary to group or combine changesets before enumerating the changes.

* `SqliteChangesetReader.openGroup()` combines a set of changesets and enumerates over the resulting change group.
* `SqliteChangesetReader.writeToFile()` allows writing a change group to a file. This can be beneficial for large change groups, where an application can write them to disk first before processing to conserve memory by streaming off disk.

imodel-native: https://github.com/iTwin/imodel-native/pull/834


mattbjordan:
Looks pretty good at a glance. One test case that might be worth adding is that Update+Update=Update

core/backend/src/SqliteChangesetReader.ts
```diff
@@ -106,7 +104,31 @@ export class SqliteChangesetReader implements IDisposable {
...
+    reader._nativeReader.openGroup(args.changesetFiles, args.invert ?? false);
+    return reader;
+  }
+  /**
+   * Writes the changeset to a file.
+   * @note can be use with openGroup() or openLocalChanges() to persist changeset.
+   * @param args - The arguments for writing to the file.
+   * @param args.fileName - The name of the file to write to.
+   * @param args.containsSchemaChanges - Indicates whether the changeset contains schema changes.
+   * @param args.overrideFile - Indicates whether to override the file if it already exists. Default is false.
```
pmconne, line 127: nit but you probably meant "overwrite".

pmconne (approved):

core/backend/src/SqliteChangesetReader.ts
khanaffan: fixed

DanRod1999:
it looks like you had a typo in you PR validation for native @khanaffan 
image
(missing a space)

This means your native code was tested against itwinjs-core master, and not your PR. The addon you generated from this [PR](https://github.com/iTwin/imodel-native/pull/834) should already be in addon version >=4.9.5. So you may need to address your errors in these builds because they were not properly tested during native pr val

Issue #6275, opened 2023-12-07
View clipping error when zooming after moving graphic 


matthaleky:
**Describe the bug**
In the following video, notice that I can zoom in very close. After moving the graphic it then clips too soon as I zoom in.
![2023-12-07_09-30-07](https://github.com/iTwin/itwinjs-core/assets/50327481/00f54cf0-e1f9-4c5f-8880-52d3d7e2158a)
Here is a simpler model that shows the problem with a cylinder and using only iTwinjs Edit Tools.
![2023-12-07_13-14-27](https://github.com/iTwin/itwinjs-core/assets/50327481/a3cbbd20-bbaa-4b20-861a-eab1a3e92b75)
If I close that model and open another model or restart the app, zooming is correct again. Turning EditScope off and on again does not help.

**To Reproduce**
[cylinder.zip](https://github.com/iTwin/itwinjs-core/files/13603548/honkingaround.zip) has the bim file in the second example. I placed the cylinder using iTwin Edit Tools. I used the keyin "editor move elements" to move the cylinder.

1. Open the example bim with an App that has the [Edit Tools](https://github.com/iTwin/itwinjs-core/blob/master/editor/frontend/README.md) enabled. I tried in Display-Test-App but it errors with "Key-in Failed to run"
2. Use the keyin "editor move elements" to move the cylinder to other locations
3. There is a certain distance from the original location when zooming in close clips too soon.

Videos are using latest Mango/Studio with Substation
Version: 0.1.0 (b62a306) 
Build: iTwinStudio 2023.11.27-1 
iTwin.js: 4.3.0-dev.25 
iTwinUI: 2.11.10 
AppUI: 4.6.1 
React: 17.0.2 
Node.js: 18.16.1 
Electron: 26.0.0 
Build Tools Commit: 842412e

pmconne:
Are you moving the element outside of the initial project extents?

matthaleky:
It is within the project extents.

matthaleky:
Notice project extents are displayed. I moved the right cylinder until zoom started clipping too soon. I then placed the right cylinder and recorded.
![2023-12-12_08-24-27](https://github.com/iTwin/itwinjs-core/assets/50327481/9e68960e-b9df-4f67-b9e5-d97cd2c2b8e3)

markschlosseratbentley:
How small is your geometry? (meters)

matthaleky:
[...]

PR #7087, opened 2024-08-14
Fix ranges on tiles during edit mode.

MarcNeely:
When elements are edited and their range changes, the range on their temporary tiles used during edit were being changed, but their boundingSpheres were not updated to match the range changes.  This caused them to disappear sometimes when the un-updated bounding sphere was out of range of the view.

I believe this should fix issue [#6275](https://github.com/iTwin/itwinjs-core/issues/6275).


pmconne:
[These tests](https://github.com/iTwin/itwinjs-core/blob/30584ce0c380812067a3f6bfdc0028d876d0d024/full-stack-tests/core/src/frontend/standalone/GraphicalEditingScope.test.ts#L298) already verify the tile's range is updated when elements are moved. Can you adjust it to also test that the bounding sphere is updated?

MarcNeely:
>These tests already verify the tile's range is updated when elements are moved. Can you adjust it to also verify that the bounding sphere is updated?

OK, I'll take a stab at that in the morning.

MarcNeely:
@pmconne Test added.

markschlosseratbentley (approved):

pmconne (approved):

PR #7088, opened 2024-08-14
Allow async methods on worker proxy

matmarchand:
Adding async support to enable worker proxy so we can run async operation like fetch().  In my case, it is a large geojson stream that would like to avoid moving from the the main thread to the worker thread.

### DIFF
--- a/core/frontend/src/workers/RegisterWorker.ts
+++ b/core/frontend/src/workers/RegisterWorker.ts
@@ -30,7 +30,10 @@ export function registerWorker<T>(impl: WorkerImplementation<T>): void {
       assert(typeof req === "object" && "operation" in req && "payload" in req && "msgId" in req);
       const func = (impl as any)[req.operation];
       assert(typeof func === "function");
-      const ret = func.constructor.name === "AsyncFunction" ? await func(req.payload) : func(req.payload);
+      let ret = func(req.payload);
+      if (ret instanceof Promise) {
+        ret = await ret;
+      }
       if (typeof ret === "object" && "transfer" in ret)
         postMessage({ result: ret.result, msgId }, { transfer: ret.transfer });
       else
--- a/core/frontend/src/workers/RegisterWorker.ts
+++ b/core/frontend/src/workers/RegisterWorker.ts
@@ -23,14 +23,14 @@ interface WorkerRequest {
  * @beta
  */
 export function registerWorker<T>(impl: WorkerImplementation<T>): void {
-  onmessage = (e: MessageEvent) => {
+  onmessage = async (e: MessageEvent) => {
     const req = e.data as WorkerRequest;
     const msgId = req.msgId;
     try {
       assert(typeof req === "object" && "operation" in req && "payload" in req && "msgId" in req);
       const func = (impl as any)[req.operation];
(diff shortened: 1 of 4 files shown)


core/frontend/src/workers/RegisterWorker.ts
```diff
@@ -23,14 +23,14 @@ interface WorkerRequest {
...
-  onmessage = (e: MessageEvent) => {
+  onmessage = async (e: MessageEvent) => {
     const req = e.data as WorkerRequest;
     const msgId = req.msgId;
     try {
       assert(typeof req === "object" && "operation" in req && "payload" in req && "msgId" in req);
       const func = (impl as any)[req.operation];
       assert(typeof func === "function");
-      const ret = func(req.payload);
+      const ret = func.constructor.name === "AsyncFunction" ? await func(req.payload) : func(req.payload);
```
pmconne, line 33:
```suggestion
      let ret = func(req.payload);
      if (ret instanceof Promise) {
        ret = await ret;
      }
```

pmconne:
Please explain use case.
Please write a test that confirms that when you do:
```ts
await workerProxy.someLongRunningAsyncOperation();
await workerProxy.someFastSynchronousOperation();
```
the first operation always completes before the second.

matmarchand:
> Please explain use case. Please write a test that confirms that when you do:
> 
> ```ts
> await workerProxy.someLongRunningAsyncOperation();
> await workerProxy.someFastSynchronousOperation();
> ```
> 
> the first operation always completes before the second.

Could you explain what you have in mind with this test? javascript 'await' contract is _"to wait for a promise and get its fulfillment value"_ so the second line will never execute before the first one has ended.

pmconne:
> javascript 'await' contract is "to wait for a promise and get its fulfillment value" so the second line will never execute before the first one has ended.

No, it doesn't block the JS event loop, it yields to the JS event loop.
This means that, for example, `onmessage` can be invoked again while you're awaiting your promise.
The message queue built into WorkerProxy *should* prevent `onmessage` from being reinvoked until the first operation completes, if the caller `await`ed it. The test is supposed to prove that it actually *does*.

matmarchand:
> > javascript 'await' contract is "to wait for a promise and get its fulfillment value" so the second line will never execute before the first one has ended.
> 
> No, it doesn't block the JS event loop, it yields to the JS event loop. This means that, for example, `onmessage` can be invoked again while you're awaiting your promise. The message queue built into WorkerProxy _should_ prevent `onmessage` from being reinvoked until the first operation completes. The test is supposed to prove that it actually _does_.
> 
> I asked about use cases because in some cases it may be important that operations execute in sequence - e.g., "first pass some data to be processed, then invoke an operation to process that data". In others, it may not matter (e.g., "perform this atomic operation and return the result"). Yours sounds like the latter?

From now, it's the latter, I'm expecting to make atomic requests so sequence it no important. I updated the test and the message queue does not seem to produce a FIFO result.

pmconne:
> I updated the test and the message queue does not seem to produce a FIFO result.

You're testing the opposite of what I suggested, which is fine, but different.
I'll update the tests.

pmconne (approved):

PR #7090, opened 2024-08-15
add `setKindOfQuantity` api

christophermlawson:


core/ecschema-editing/src/Editing/Properties.ts
```diff
@@ -143,6 +143,26 @@ export class Properties {
     property.setCategory(new DelayedPromiseWithProps<SchemaItemKey, PropertyCategory>(categoryKey, async () => category));
   }
 
+  /**
```
ColinKerr, line 146: How do we want to handle the validation that we have not changed the persistence unit with this edit?  If it's a breaking schema change then it's OK to change the persistence unit ... if not then it's not OK.

Also this should be consistent up and down the stack ... changing the persistence unit for one property should change it for all the overrides.

christophermlawson: Let me look into it and I'll let you know when I have a solution commited.

christophermlawson: I've created the backlog item for this: https://github.com/orgs/iTwin/projects/29/views/1?pane=issue&itemId=75487150

christophermlawson:
@Mergifyio backport release/4.8x

christophermlawson:
https://github.com/Mergifyio backport release/4.8.x

PR #7100, opened 2024-08-21
Frontend-tiles bug fix and argument clean-up

eringram:
- Fixed an incorrect property name in `queryGraphicRepresentations`
- Changed `obtainMeshExportTilesetUrl` function arguments to directly take the  iTwin, iModel, and changeset IDs instead of an `IModelConnection` object
    - The reason for this is that sometimes it's desirable to call this function before an `IModelConnection` is created, and the function just uses the connection to get the IDs anyway
- Moved logger initialization earlier during the Display Test App startup. I noticed I was missing useful info logged before `Logger.initializeToConsole` was called.


extensions/frontend-tiles/src/GraphicsProvider/GraphicsProvider.ts
```diff
@@ -8,12 +8,12 @@ import { IModelApp, IModelConnection} from "@itwin/core-frontend";
...
 import { loggerCategory} from "../LoggerCategory";
 
-/** Arguments supplied  to [[obtainMeshExportTilesetUrl]].
+/** Arguments supplied  to [[obtainIModelTilesetUrl]].
  * @beta
  */
 export interface ObtainIModelTilesetUrlArgs {
   /** The iModel for which to obtain a tileset URl. */
-  iModel: IModelConnection;
+  iModel: IModelConnection | { iTwinId: string, iModelId: string, changesetId: string };
```
jason-crow, line 16: `changesetId?: string` 
^^ optional right?

eringram: Yeah it should be, fixed

jason-crow:
https://github.com/Mergifyio backport release/4.8.x

markschlosseratbentley (approved):

jason-crow (approved):

PR #7101, opened 2024-08-22
Schema Editing API refinements

StefanApfel-Bentley:
- Utils get exported as _DifferenceUtils_ namespace to reduce the exported api 
- Differences use actual contextual property instead of generic path (like propertyName, constraint, enumerator)
- Schema Conflict Code now has prefix _SchemaConflict_ instead of _C_


christophermlawson (approved):
Looks pretty straightforward.  I don't see anything to comment one.

StefanApfel-Bentley:
This PR contains a couple of changes that seem to be useful (like the contextural conflict interfaces) others might break existing consumers (eg when changing the conflict codes) so, before this will continue, we should contact existing consumers of that api (like transformation).

StefanApfel-Bentley:
I converted the PR as draft, since we won't work on that in the short future.

PR #7105, opened 2024-08-23
Fix `error` metadata being overwritten in logs

GytisCepk:
At some point, the value of the `error` metadata property is being overwritten to `true`, and as a result, the original value is not available in telemetry.


paulius-valiunas (approved):

aruniverse (approved):

paulius-valiunas:

PR #7109, opened 2024-08-26
Add disableClipStyle to BranchState

andremig-bentley:
This PR is to address this issue: https://github.com/iTwin/itwin-graphics-backlog/issues/343

An optional `disableClipStyle` boolean has been added to `GraphicBranchOptions`, `BranchStateOptions`, `BranchState`, and `RealityMeshParams`. If set `true`, any clip in the branch will override the view's `DisplayStyleSettings.clipStyle`, and instead, no `ClipStyle.insideColor` or `ClipStyle.outsideColor` will be applied to the clip. This is accomplished within `BranchUniforms` every time a new branch is pushed to the stack by setting the alpha values of `ClipStack.insideColor`, `ClipStack.outsideColor`, and `ClipStack.intersectionStyle` to 0 if `disableClipStyle` is true. Otherwise, it is left to the viewport's `DisplayStyleSettings.clipStyle` to determine these values.

This is being applied to fix the above bug by overriding the clip style in branches created for rendering map tiles in `MapTile.ts`, and then passing that information along through `RealityMeshParams`.


core/frontend/src/render/webgl/BranchUniforms.ts
```diff
@@ -103,6 +105,12 @@ export class BranchUniforms {
   public pushBranch(branch: Branch): void {
     desync(this);
     this._stack.pushBranch(branch);
+    if (branch.overrideClipStyle) {
+      this.overrideClipStyle = true;
+    } else {
+      this.overrideClipStyle = false;
+    }
```
ben-polinsky, line 112:
```suggestion
    this.overrideClipStyle = !!branch.overrideClipStyle
```

DStradley (approved):

andremig-bentley:
Since this PR is not critical, we will wait until @pmconne returns to office next week and has a chance to look at it before merging it in.

core/frontend/src/render/webgl/glsl/Clipping.ts
```diff
@@ -99,13 +99,13 @@ const applyClipPlanesPostlude = `
 
   numSetsClippedBy += int(clippedByCurrentPlaneSet);
   if (numSetsClippedBy == numPlaneSets) {
-    if (u_outsideRgba.a > 0.0) {
+    if (!u_overrideClipStyle && (u_outsideRgba.a > 0.0)) {
```
pmconne, line 102: What about ClipIntersectionStyle?

core/frontend/src/render/webgl/BranchUniforms.ts
```diff
@@ -70,6 +70,8 @@ export class BranchUniforms {
   private readonly _scratchVIModelMatrix = Transform.createIdentity();
   private readonly _zeroPoint = new Point3d(0, 0, 0);
 
+  public overrideClipStyle = false;
```
pmconne, line 73: This is really `disableClipStyle`, isn't it? What led you to decide to use a boolean checked in the shader, vs moving ClipStyle into BranchState?

andremig-bentley: Since the bug this was trying to solve was from a mis-application of the clip style, passing a single boolean along seemed to be the most efficient way to solve the bug. That being said, it does limit the application of this to only solving the related bug and other very similar scenarios where we wanted specifically no ClipStyle. If we wanted a solution that could be applied to more in the future, especially if we wanted to allow multiple ClipStyles per viewport, then adding ClipStyle to BranchState would be a better solution.

core/frontend/src/render/webgl/glsl/Clipping.ts
andremig-bentley: My initial testing didn't show that ClipIntersectionStyle was part of the related bug, but I believe that was because I limited the width so it wasn't noticable. After a little more testing with a wider width just now, I see the ClipIntersectionStyle is being applied to the map tiles, so I'll work on fixing this.

core/frontend/src/render/webgl/RealityMesh.ts
```diff
@@ -376,7 +376,7 @@ export class RealityMeshGeometry extends IndexedGeometry implements IDisposable,
       branch.add(system.createBatch(primitive!, featureTable, mesh.getRange(), { tileId }));
     }
 
-    return system.createBranch(branch, realityMesh._transform ? realityMesh._transform : Transform.createIdentity());
+    return system.createBranch(branch, realityMesh._transform ? realityMesh._transform : Transform.createIdentity(), {overrideClipStyle: true});
```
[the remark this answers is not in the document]
andremig-bentley: To fix this, a `disableClipStyle` boolean has been added to `RealityMeshParams`, and it is now being set in `MapTile.ts`, so `RealityMesh.ts` now only passes that information along instead of dictating it.

core/frontend/src/render/webgl/BranchUniforms.ts
andremig-bentley: As of the latest commit, `overrideClipStyle` has been renamed to `disableClipStyle`. The logic for using this boolean has been moved to within `PushBranch` in `BranchUniforms` and has been removed from the shader. Now, in `PushBranch`, the alpha values of `ClipStack.insideColor`, `ClipStack.outsideColor`, and `ClipStack.intersectionStyle` are set to 0 if `disableClipStyle` is true, thereby disabling them. Otherwise, it is left to the viewport's clip style to decide the alpha values.

andremig-bentley:
> Missing tests.

A unit test has been added to `BranchUniforms.test.ts` which will test whether or not the clip style is successfully disabled when a branch with `disableClipStyle` set to true is pushed to the stack.

core/frontend/src/render/webgl/BranchUniforms.ts
```diff
@@ -103,6 +104,19 @@ export class BranchUniforms {
...
     desync(this);
     this._stack.pushBranch(branch);
+
+    const vp = IModelApp.viewManager.selectedView;
+    if (vp) {
+
+      /** If the branch has disableClipStyle set, we set these alphas to 0, therefore disabling the clip style.
+      * Otherwise, we leave it to the viewport's clip style.
+      */
+      const style = vp.view.displayStyle.settings.clipStyle;
```
pmconne, line 114: It looks to me like if I have a graphic with the following structure:
```
Branch: disableClipStyle = true
  Branch: disableClipStyle = false (or undefined)
    Some visible graphics
```

Then when you push the second branch you will turn the clip style back on, where we want the disableClipStyle flag to be inherited from the top of the stack when pushing, and restored when popping.

Can you write a test that proves me wrong?

andremig-bentley: You were right, some logic has been added to `BranchUniforms.pushBranch` to ensure `disableClipStyle` is inherited from the top of the stack, and a test for this has been added to `BranchUniforms.test.ts`.

pmconne: Your test doesn't appear to validate the "resotre when popping" behavior and  your implementation here doesn't appear to handle it.

If I have this scene graph, where only branch A1 disables the clip style:
```
Branch A
  Branch A1: disableClipStyle = true
    Branch A1a
  Branch A2
Branch B
  Branch B1
```

and the view has a clip style, then the clip style should be applied to branches A2, B, and B1; and not applied to branches A1 and A1a.

AFAICT, the way you've implemented it, as soon as any branch disables the clip style, it remains disabled for everything subsequently drawn, even after that branch is popped. Did I overlook something?

andremig-bentley: You're right, some changes are in progress. A quick question, assume we had this graph:
```
Branch A
  Branch A1: disableClipStyle = true
    Branch A1a
      Branch1aa
    Branch1b
  Branch A2
Branch B
  Branch B1
```

In this scenario, am I correct in saying we want both Branch1aa and Branch1b to both have a disabled clip style?

pmconne: Yep. It's a stack. State changes on push, resets on pop.

core/frontend/src/render/GraphicBranch.ts
```diff
@@ -160,6 +160,10 @@ export interface GraphicBranchOptions {
    * @internal
    */
   viewAttachmentId?: Id64String;
+  /** If true, the view's [[DisplayStyleSettings.clipStyle]] will be disabled for this branch.
+   * No [[ClipStyle.insideColor]], [[ClipStyle.outsideColor]], or [[ClipStyle.intersectionStyle]] will be applied.
+   */
+  disableClipStyle?: boolean;
```
pmconne, line 166: Consider changing the type to `true | undefined`. Otherwise, you will have to handle the case where you push a branch that disables the clip style, containing another branch that re-enables it, which doesn't seem useful.

andremig-bentley: done in latest commit

core/frontend/src/render/webgl/BranchUniforms.ts
andremig-bentley: In `BranchUniforms.ts`, logic has been added to `pushBranch` to ensure `disableClipstyle` is inherited even if there are multiple branches in a row with it undefined like the scenario above,  logic has been added to `pop` to restore the clipstyle when a branch is popped, and two new functions `setClipStyle` and `resetClipStyle` were created for clarity within `pushBranch`. `testBranches` in `BranchUniforms.test.ts` now also optionally tests for the clipstyle being restored after a branch is popped.

core/frontend/src/render/webgl/BranchUniforms.ts
```diff
@@ -103,19 +103,12 @@ export class BranchUniforms {
...
-      /** If the branch has disableClipStyle set, we set these alphas to 0, therefore disabling the clip style.
-      * Otherwise, we leave it to the viewport's clip style.
-      */
-      const style = vp.view.displayStyle.settings.clipStyle;
-      this.clipStack.insideColor.alpha = disableClipStyle ? 0 : (style.insideColor ? 1 : 0);
-      this.clipStack.outsideColor.alpha = disableClipStyle ? 0 : (style.outsideColor ? 1 : 0);
-      this.clipStack.intersectionStyle.alpha = disableClipStyle ? 0 : (style.intersectionStyle ? style.intersectionStyle.width : 0);
-    }
+    if (this.top.disableClipStyle)
+      branch.disableClipStyle = true;
```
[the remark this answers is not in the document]
andremig-bentley: that's been removed, and the clip style is now being set with `setClipStyle(this.top.disableClipStyle)` after the branch is pushed, meaning the newly created `BranchState` is being used to determine the clip style and therefore, `disableClipStyle` will be fully inherited without modifying the input `Branch`.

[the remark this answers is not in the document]
pmconne: Ex:
```
A: disableClipStyle = true
  B
  C
```

Both B and C should draw with clip style disabled.
But in your implementation, C will draw with clip style enabled, because you re-enable it when you pop B.

core/frontend/src/render/webgl/BranchState.ts
```diff
@@ -46,6 +46,10 @@ export interface BranchStateOptions {
   forceViewCoords?: boolean;
   readonly viewAttachmentId?: Id64String;
   groupNodeId?: number;
+  /** If true, the view's [[DisplayStyleSettings.clipStyle]] will be disabled for this branch.
+   * No [[ClipStyle.insideColor]], [[ClipStyle.outsideColor]], or [[ClipStyle.intersectionStyle]] will be applied.
```
pmconne, line 50: These links are invalid - e.g., should be `[ClipStyle.insideColor]($common)` not `[[ClipStyle]]`.
@ben-polinsky docs build wouldn't catch this?

pmconne (approved):

core/frontend/src/render/webgl/BranchState.ts
andremig-bentley: Links have been updated to the above form.

andremig-bentley:
Final step before merging is to figure out why `rush cover` is consistently failing on the `RPCInterface - should set cache-control headers when applicable` test and solve that issue if necessary.

core/frontend/src/test/render/webgl/BranchUniforms.test.ts
```diff
@@ -86,17 +98,23 @@ function testBranches(viewClip: ClipInfo, branches: ClipInfo[], expectViewClip:
...
+
   expect(uniforms.clipStack.hasViewClip).to.equal(hadViewClip);
   expect(uniforms.clipStack.hasClip).to.equal(hadClip);
   expectClipStack(target, prevClips);
 
   dispose(target);
 }
 
-describe("BranchUniforms", async () => {
+describe.only("BranchUniforms", async () => {
```
pmconne, line 117:
```suggestion
describe("BranchUniforms", async () => {
```

core/frontend/src/render/webgl/BranchState.ts
ben-polinsky: @pmconne It should but apparently there is an issue. I'll add it to the backlog but not sure when I'll have time to take a look.

toddsouthenbentley:
@Mergifyio backport release/4.9.x

PR #7111, opened 2024-08-26
Ratio support for quantity framework

naronchen:
Support for ratio formatting in the quantity package.


core/quantity/src/Formatter/Formatter.ts
```diff
@@ -231,6 +232,15 @@ export class Formatter {
         const componentText = Formatter.formatCompositePart(unitValue, true, currentLabel, spec);
         compositeStrings.push(componentText);
       }
+
+      if (spec.format.type === FormatType.Ratio){
+        if ( spec.persistenceUnit.phenomenon !== "Units.SLOPE")
```
rschili, line 237: We should not hard code slope here, also we should not ignore the additional composite parts if ratio is selected. Instead, either throw an error if there is more than one part in the composite (saying we require a single part) or format each part as a ratio. Not sure if the second makes sense, we could clarify with Kevin

core/quantity/src/Formatter/Formatter.ts
```diff
@@ -546,4 +556,45 @@ export class Formatter {
 
     return converted.magnitude;
   }
+
+  private static formatRatio(magnitude: number, spec: FormatterSpec): string {
+    if (null === spec.format.ratioType)
+      throw new QuantityError(QuantityStatus.InvalidCompositeFormat, `The Format ${spec.format.name} must have a ratio type specified.`);
+
+    if (spec.format.hasUnits && spec.format.units && spec.format.units[0][0].phenomenon !== "Units.SLOPE") // TODO - <Naron>: this should be handled at convert?
```
rschili, line 564: Same here, do not check for Slope because the phenomenon could have a different name based on its provider/schema

core/quantity/src/Parser.ts
```diff
@@ -796,6 +800,94 @@ export class Parser {
...
+    const parts = inString.split(":");
+    if (parts.length !== 2)
+      throw new QuantityError(QuantityStatus.UnableToConvertParseTokensToQuantity, `Invalid ratio: ${inString}, needs to be in form x:y`);
+
+    const numerator = parseFloat(parts[0]);
+    const denominator = parseFloat(parts[1]);
+
+    if (isNaN(numerator) || isNaN(denominator))
+      return { ok: false, error: ParseError.NoValueOrUnitFoundInString };
+    // TODO - Naron: there is also a QuantityStatus.NoValueOrUnitFoundInString
```
rschili, line 816: ParseError is better, keep it and remove comment

core/quantity/src/Parser.ts
```diff
@@ -796,6 +800,94 @@ export class Parser {
     return { ok: true, value: magnitude };
   }
 
+  private static parseRatioFormat(inString: string, spec: ParserSpec): QuantityParseResult {
+    if (!inString || !inString.includes(":"))
+      return { ok: false, error: ParseError.NoValueOrUnitFoundInString };
+
+    const parts = inString.split(":");
+    if (parts.length !== 2)
+      throw new QuantityError(QuantityStatus.UnableToConvertParseTokensToQuantity, `Invalid ratio: ${inString}, needs to be in form x:y`);
```
rschili, line 809: I say an invalid input string is not an exception (unexpected). It's user input, so it can be malformed. Also, if the input string is just a single number, like "30", we may interpret that as "30:1". I reckon that's what a user would expect for that input

core/quantity/src/Parser.ts
```diff
@@ -796,6 +800,94 @@ export class Parser {
...
+        return { ok: true, value: 0.0 };
+      // if (spec.format.units[0][0].name is invert of spec.outUnit && numerator === 1 && denominator === 0)
+      //   return { ok: true, value: 0.0 };
+    } else {
+      throw new QuantityError(QuantityStatus.MissingRequiredProperty, "Missing presentation unit or persistence unit for ratio format.");
+    }
+
+    let magnitude = 0.0;
+    switch (spec.format.ratioType) {
+      case RatioType.OneToN:
```
rschili, line 830: This can be more simple. I say we do not validate the RatioType here. We are parsing a ratio, so we should try and accept "2:1" as well as "4:2" or "8:4". It translates into the same persisted value. No need to enforce the RatioType here. Only use it in the Formatter.

core/quantity/src/Parser.ts
```diff
@@ -796,6 +800,94 @@ export class Parser {
...
+        if (!gcdResult.ok)
+          return gcdResult;
+        magnitude = numerator / denominator;
+        break;
+
+      default:
+        throw new QuantityError(QuantityStatus.InvalidJson, `Unsupported ratio type: ${spec.format.ratioType}`);
+    }
+
+    // TODO: convert the magnitude to OutUnit, the invert unit conversion is not handled yet in unit conversion
```
rschili, line 869: Like the comment says, this needs to be implemented before we merge.

core/quantity/src/Formatter/Formatter.ts
```diff
@@ -214,7 +214,26 @@ export class Formatter {
...
+      // let unitValue = applyConversion(posMagnitude, unitConversion) + this.FPV_MINTHRESHOLD;
+      // put this in try catch, if catched QuantityError(QuantityStatus.InvertingZero), if the spec.format.type is Ratio, we will return 1:0
+
+      let unitValue = 0.0;
+      try {
+        unitValue = applyConversion(posMagnitude, unitConversion) + this.FPV_MINTHRESHOLD;
+      } catch (e) {
+        // error code 35049 is thrown when inverting zero
+        // This is special case for ratio, just return 1:0
+        if (spec.format.type === FormatType.Ratio && e instanceof QuantityError && e.errorNumber === QuantityStatus.InvertingZero) {
```
rschili, line 226: Oh, I didn't think of this when I created the applyConversion function. I don't like that we are using an exception here for something that's expected. All the other spots that dealt with conversion before which I adjusted to use this method didn't support that conversion could fail. So I thought raising an exception is the best choice, so we can be sure it's not swallowed. But looking at this here.... In native we return a UnitsProblemCode which contains these errors:
```cpp
enum class UnitsProblemCode
    {
    NoProblem = 0,
    UncomparableUnits = 31,     //!< Units provided on the argument list are not comparable
    InvalidUnitName = 32,       //!< Not-recognizd unit name or unit is not associated with a Phenomenon
    InvertingZero = 33,         //!< Attempting to invert a zero value
    UnspecifiedProblem = 34,    //!< A problem occurred but the cause is not specified
    NaN = 35,                   //!< A value is NaN
    };
```

Let me think about this for a bit. I'm leaning towards a tryConvert() function which will return errors instead of throwing exceptions

core/quantity/src/Formatter/Formatter.ts
```diff
@@ -214,7 +214,26 @@ export class Formatter {
...
+            return "1:0";
+        }
+      }
+
+      if (spec.format.type === FormatType.Ratio){
+        if (1 !== spec.format.units!.length)
+          throw new QuantityError(QuantityStatus.InvalidCompositeFormat, `The Format ${spec.format.name} has an invalid unit specification, we require single presentation unit`);
```
rschili, line 233:
```suggestion
          throw new QuantityError(QuantityStatus.InvalidCompositeFormat, `The Format ${spec.format.name} has an invalid unit specification, we require single presentation unit when using format type 'ratio'`);
```

core/quantity/src/Parser.ts
```diff
@@ -796,6 +800,62 @@ export class Parser {
...
+      quantity = new Quantity(spec.format.units[0][0], numerator / denominator);
+    } else {
+      throw new QuantityError(QuantityStatus.MissingRequiredProperty, "Missing presentation unit or persistence unit for ratio format.");
+    }
+
+    let converted: Quantity | undefined;
+    try {
+      converted = quantity.convertTo(spec.outUnit, unitConversion);
+    } catch (err){
+      if (err instanceof QuantityError && err.errorNumber === QuantityStatus.InvertingZero){
```
rschili, line 847: I think you caught this further up. I don't see any test hitting this code, and I think it's probably wrong. The result of a zero inversion should not be zero.

core/quantity/src/Formatter/Formatter.ts
rschili: Alright, I thought about it and decided to keep it as is. Most of those problemCodes would not apply here, and creating a separate API just for this corner case seems overkill.

rschili (approved):

mathieu-fournier:
Looking at the tests, we dont test math operation like plus or minus. 
This deviates from the standard of other units that all support these math operators.
This might be for another PR, but I think that all units should act the same way with maths.

mathieu-fournier:
The parser has 2 APIs to parse strings. 
One asynchonous [`Parser.ParseIntoQuantity`](https://github.com/iTwin/itwinjs-core/blob/de8f59995fb70cf963568cfb15465a3fbe125574/core/quantity/src/Parser.ts#L470C23-L470C40). 
One synchronous [`Parser.ParseQuantityString`](https://github.com/iTwin/itwinjs-core/blob/de8f59995fb70cf963568cfb15465a3fbe125574/core/quantity/src/Parser.ts#L645C17-L645C36). 
Both has their own return type and error handling mecanism. 
I think we should test both to be sure that they support the new format correctly.

rschili:
> Looking at the tests, we dont test math operation like plus or minus. This deviates from the standard of other units that all support these math operators. This might be for another PR, but I think that all units should act the same way with maths.

Good point. I was wondering about that but dismissed the idea because I could not imagine how that would look like. For example, what would "2:1+3" mean? Should it follow common maths rules? But I'd argue that string is not a ratio.

docs/bis/ec/ec-format.md
```diff
@@ -86,6 +87,23 @@ Supported scientific type:
 - `normalized`
 - `zeroNormalized`
 
+**RatioType** determines the type of ratio format to use. Only valid and its required when the type is ratio.
```
hl662, line 90:
```suggestion
**RatioType** determines the type of ratio format to use. Only valid and is required when the type is ratio.
```

hl662: It seems like other attributes of Format uses camelCase, not PascalCase - should ratioType follow the same manner?

docs/changehistory/NextVersion.md
```diff
@@ -3,3 +3,11 @@ publish: false
 ---
 # NextVersion
 
+Table of contents:
+
+- [Quantity](#quantity)
+
+## Quantity
+
+- Add support for 'Ratio' format type (e.g. "1:2")
```
hl662, line 12: Can you add an example of formatting a ratio here? You can denote outside the code snippet that you're assuming a UnitsProvider (like the SchemaUnitsProvider) has been registered or etc. Might be good to then put the example in either ec-format or QuantityFomratting.md

core/quantity/src/test/Ratio.test.ts
```diff
@@ -0,0 +1,449 @@
...
+    const defaultPrecision = 3;
+
+    const ratioJson: FormatProps = {
+      type: "Ratio",
+      ratioType,
+      precision: defaultPrecision,
+      composite: {
+        includeZero: true,
+        units: [
+          { name: presentationUnitStr}, // presentation unit
```
hl662, line 23:
```suggestion
          { name: presentationUnitStr }, // presentation unit
```

core/quantity/src/test/TestUtils/TestHelper.ts
```diff
@@ -153,6 +158,12 @@ export class TestUnitsProvider implements UnitsProvider, AlternateUnitLabelsProv
       const conversionData = new ConversionData();
       conversionData.factor = deltaNumerator / deltaDenominator;
       conversionData.offset = deltaOffset;
+      if (fromUnitData.isInverted && !toUnitData.isInverted) {
```
hl662, line 161: @rschili do we want to update core-frontend's basicUnitsProvider to reflect this too...?

rschili:
> The parser has 2 APIs to parse strings. One asynchonous [`Parser.ParseIntoQuantity`](https://github.com/iTwin/itwinjs-core/blob/de8f59995fb70cf963568cfb15465a3fbe125574/core/quantity/src/Parser.ts#L470C23-L470C40). One synchronous [`Parser.ParseQuantityString`](https://github.com/iTwin/itwinjs-core/blob/de8f59995fb70cf963568cfb15465a3fbe125574/core/quantity/src/Parser.ts#L645C17-L645C36). <- only this one is tested now. Both has their own return type and error handling mecanism. I think we should test both to be sure that they support the new format correctly.

I agree we should test both, though, the async really only exists to dynamically look-up units by label. Ratio never uses labels, so this makes no sense. The actual parsing does not have any other async logic. The API is not obvious about this, but we should prefer calling the sync method if we know the units we are dealing with.

core/quantity/src/test/TestUtils/TestHelper.ts
rschili: I thought about this, too. For now, BasicUnitsProvider has no inverted units, it also does not have the SLOPE Phenomenon. And we want to eventually deprecate it for the SchemaUnitsProvider. So I think this is fine.

core/quantity/src/Formatter/Formatter.ts
```diff
@@ -211,10 +211,30 @@ export class Formatter {
 
       if (i > 0 && unitConversion.factor < 1.0)
         throw new QuantityError(QuantityStatus.InvalidCompositeFormat, `The Format ${spec.format.name} has a invalid unit specification..`);
-      if (i > 0 && unitConversion.offset !== 0)
+      if (i > 0 && unitConversion.offset !== 0) // offset should only ever be defined for major unit
         throw new QuantityError(QuantityStatus.InvalidCompositeFormat, `The Format ${spec.format.name} has a invalid unit specification..`);
 
-      let unitValue = (posMagnitude * unitConversion.factor) + unitConversion.offset + Formatter.FPV_MINTHRESHOLD; // offset should only ever be defined for major unit
+      // let unitValue = applyConversion(posMagnitude, unitConversion) + this.FPV_MINTHRESHOLD;
```
hl662, line 217: remove this commented out code?

core/quantity/src/Formatter/FormatEnums.ts
```diff
@@ -112,6 +112,8 @@ export enum FormatType {
   Bearing,
   /** Azimuth angle e.g. 45°30'00". Requires provided quantities to be of the angle phenomenon */
   Azimuth,
+  /** Eg. 1:2 or 0.3:1. Requires provided quantities to be of the slope phenomenon */
```
ColinKerr, line 115: is slope required?  Is something invertible required if we don't do inversion?  Would miles per gallon work so long as we didn't ask for gallons per mile?

core/quantity/src/Formatter/Formatter.ts
ColinKerr: also the comment below this line should be removed.

core/quantity/src/test/Ratio.test.ts
```diff
@@ -0,0 +1,456 @@
...
+        { input: 0.6667, ratio: "667:1000" },
+      ];
+      await testRatioType("UseGreatestCommonDivisor", testData);
+    });
+  });
+
+  describe("RatioType Tests with different precision", () => {
+    it("ratioType precision test | One To N", async () => {
+      const testData: { input: number, ratio: string, precision: number}[] = [
+        // { input: 3, ratio: "1:0", precision: 0 }, commented out since its expected to fail on parsing for now
```
ColinKerr, line 122: why is it expected to fail?  That is important context for someone in the future who is going to look at this

core/quantity/src/test/Ratio.test.ts
```diff
@@ -0,0 +1,456 @@
...
+        { input: 2.0, ratio: "2:1" },
+        { input: 0.5, ratio: "1:2" },
+        { input: 0.333, ratio: "333:1000" },
+        { input: 0.3333, ratio: "333:1000" },
+        { input: 0.2857, ratio: "143:500" },
+        { input: 0.25, ratio: "1:4" },
+        { input: 0.6667, ratio: "667:1000" },
+      ];
+      await testRatioType("UseGreatestCommonDivisor", testData);
+    });
```
ColinKerr, line 116: it seems like there would be some interesting cases here with varying precision that aren't covered by the tests below

core/quantity/src/test/Ratio.test.ts
```diff
@@ -0,0 +1,456 @@
...
+    it("zero value", async () => {
+      const testData: { input: number, ratio: string }[] = [
+        { input: 0.0, ratio: "0:1" },
+      ];
+      await testRatioType("NToOne", testData);
+    });
+
+    it("large/small value", async () => {
+      const testData: { input: number, ratio: string }[] = [
+        { input: 0.00000001, ratio: "0:1" },
```
ColinKerr, line 211: might be nice to see the threshold where precision and number meet so we see with one less digit of precision it is 0:1 and with more more it is 0.00001:1

core/quantity/src/test/Ratio.test.ts
```diff
@@ -0,0 +1,456 @@
...
+        { input: 1.0 / 7, ratio: "0.143:1" },
+        { input: 2.0 / 7, ratio: "0.286:1" },
+      ];
+      await testRatioType("NToOne", testData);
+    });
+
+    it("irrational number", async () => {
+      const testData: { input: number, ratio: string }[] = [
+        { input: 1.0 / 7, ratio: "143:1000" },
+        { input: 2.0 / 7, ratio: "143:500" }, // loses precision from 0.28571428571 to 0.286
```
ColinKerr, line 237: once again might be nice to test how precision impacts the result

docs/bis/ec/ec-format.md
```diff
@@ -86,6 +87,23 @@ Supported scientific type:
 - `normalized`
 - `zeroNormalized`
 
+**ratioType** determines the type of ratio format to use. Only valid and is required when the type is ratio.
```
ColinKerr, line 90: we need to add an example of a ratio format ... we should have added examples of bearing and azimuth formats as well ... @rschili can we add some examples?

core/quantity/src/Formatter/FormatEnums.ts
rschili: yeah it's not required that was a doc oversight. Naron has deleted the requirement.

core/quantity/src/Formatter/FormatEnums.ts
```diff
@@ -123,6 +125,19 @@ export enum ScientificType {
   ZeroNormalized,
 }
 
+/** required if type is ratio
+ * @beta */
+export enum RatioType {
```
aruniverse, line 130: why dont we use string enums?
with that could we get rid of `parseRatioType` and `ratioTypeToString`?

docs/changehistory/NextVersion.md
```diff
@@ -27,3 +28,27 @@ Some detail and requirements are as following.
...
+            { name: "Units.HORIZONTAL_PER_VERTICAL" },
+        ],
+    },
+};
+
+const ratioFormat = new Format("Ratio");
+ratioFormat.fromJSON(unitsProvider, ratioFormatProps).catch(() => {});
+```
+
+- Add support for unit inversion during unit conversion
```
hl662, line 53: do we have any further documentation explaining how to enable unit inversion? I remember going through the code that there are different enums for UnitConversionInversion. Perhaps writing down the enums available and explaining what they do will be helpful.

rschili: I added more description to the enum "UnitConversionInvert" values explaining their purpose.

core/quantity/src/Formatter/FormatEnums.ts
rschili: Fair point, I'm changing it. It's not consistent with all the other enums in this same file though. We should probably go and adjust them all in a separate PR @hl662

docs/bis/ec/ec-format.md
rschili: done

rbreukelaar (approved):
Code looks good and readable... especially like the option UseGreatestCommonDivisor. Looks good.

core/quantity/src/Formatter/FormatEnums.ts
hl662: agreed, I'll have @franklbh tackle this, will create a separate issue

aruniverse:
> [CLA assistant check](https://cla-assistant.io/iTwin/itwinjs-core?pullRequest=7111) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/iTwin/itwinjs-core?pullRequest=7111) before we can accept your contribution.**3** out of **4** committers have signed the CLA.✅ rschili✅ hl662✅ anmolshres98❌ naronchen
> 
> **naronchen** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/iTwin/itwinjs-core?pullRequest=7111) it.

@naronchen  can you sign the CLA?

core/quantity/src/Formatter/FormatEnums.ts
```diff
@@ -154,6 +169,25 @@ export function scientificTypeToString(scientificType: ScientificType): string {
   return (scientificType === ScientificType.Normalized) ? "Normalized" : "ZeroNormalized";
 }
 
+/**  @beta   */
+export function parseRatioType(ratioType: string, formatName: string): RatioType {
+  const normalizedValue = ratioType.toLowerCase();
+  for (const key in RatioType) {
```
aruniverse, line 175: do we really need the for loop and the first if statement?
wouldnt this suffice?
```
const enumValue = RatioType[key as keyof typeof RatioType];
if (enumValue.toLowerCase() === normalizedValue) {
  return enumValue as RatioType;
}
```

core/quantity/src/Formatter/FormatEnums.ts
```diff
@@ -154,6 +169,25 @@ export function scientificTypeToString(scientificType: ScientificType): string {
...
+      if (enumValue.toLowerCase() === normalizedValue) {
+        return enumValue as RatioType;
+      }
+    }
+  }
+  throw new QuantityError(QuantityStatus.InvalidJson, `The Format ${formatName} has an invalid 'ratioType' attribute.`);
+}
+
+/**  @beta   */
+export function ratioTypeToString(ratioType: RatioType): string {
```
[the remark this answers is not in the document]
rschili: removed

core/quantity/src/Formatter/FormatEnums.ts
rschili: If we do that, eslint complains because that could include other unwanted properties from the prototype.
This rule here: https://eslint.org/docs/latest/rules/guard-for-in

rschili: I checked other methods where we do the same thing. Turns out we do not have a lot of string enums which we parse from strings. One other example is SchemaItemType, here: https://github.com/iTwin/itwinjs-core/blob/master/core/ecschema-metadata/src/ECObjects.ts#L162
That uses a switch case method which I'd argue is worse.

rschili: I'm marking the conversation resolved as it's the only thing preventing the merge. We can revisit this later if you want to follow up @aruniverse

PR #7112, opened 2024-08-26
Allow revert changes from imodel timeline

khanaffan:
### Revert timeline changes

At present, the sole method to reverse a defective changeset is to remove it from the iModel hub, which can lead to numerous side effects. A preferable approach would be to reverse the changeset in the timeline and introduce it as a new changeset. Although this method remains intrusive and necessitates a schema lock, it is safer because it allows for the reversal to restore previous changes, ensuring that nothing is permanently lost from the timeline.

`IModelDb.revertAndPushChanges()` Allow to push a single changeset that undo all changeset from tip to specified changeset in history.

Some detail and requirements are as following.

- When invoking the iModel, it must not have any local modifications.
- The operation is atomic; if it fails, the database will revert to its previous state.
- The revert operation necessitates a schema lock (an exclusive lock on the iModel) because it does not lock each individual element affected by the revert.
- If no description is provided after a revert, a default description for the changeset will be created and pushed, which releases the schema lock.
- Schema changes are not reverted during SchemaSync, or they can be optionally skipped when SchemaSync is not utilized.

imodel-native: https://github.com/iTwin/imodel-native/pull/844


core/backend/src/BriefcaseManager.ts
```diff
@@ -432,6 +444,59 @@ export class BriefcaseManager {
...
+    }
+    if (!db.holdsSchemaLock) {
+      throw new IModelError(ChangeSetStatus.ApplyError, "Cannot revert timeline changesets without holding a schema lock");
+    }
+
+    // Download change sets
+    const changesets = await IModelHost.hubAccess.downloadChangesets({
+      accessToken: arg.accessToken,
+      iModelId: db.iModelId,
+      range: { first: arg.toIndex, end: currentIndex },
```
JulijaRamoskiene, line 470: Why not pass `end: undefined` to make sure that all change sets are retrieved, other option could be to lock schema before pulling latest changes:
image

core/backend/src/IModelDb.ts
```diff
@@ -2992,6 +2992,59 @@ export class BriefcaseDb extends IModelDb {
...
+        accessToken: arg.accessToken,
+        mergeRetryCount: arg.mergeRetryCount,
+        mergeRetryDelay: arg.mergeRetryDelay,
+        pushRetryCount: arg.pushRetryCount,
+        pushRetryDelay: arg.pushRetryDelay,
+      });
+      this.clearCaches();
+    } finally {
+      this.abandonChanges();
+      await this.locks.releaseAllLocks();
```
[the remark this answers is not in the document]
khanaffan: Make sense. Let me fix this.

core/backend/src/IModelDb.ts
```diff
@@ -2992,6 +2992,59 @@ export class BriefcaseDb extends IModelDb {
...
+      throw new IModelError(ChangeSetStatus.HasLocalChanges, "Cannot revert with pending txns");
+    }
+
+    this.clearCaches();
+    while (true) {
+      await this.pullChanges({ ...arg, toIndex: undefined });
+      if (nativeDb.hasPendingTxns()) {
+        // schema sync create txn on pull. We need to push it.
+        const description = nativeDb.getTxnDescription(nativeDb.getCurrentTxnId());
+        await this.pushChanges({ description, accessToken: arg.accessToken });
```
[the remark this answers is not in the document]
nick4598: I thought skipSchemaChanges was set to true explicitly in case of schema sync?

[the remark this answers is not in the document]
khanaffan: Ah typo,  i meant true :)  thanks for noticing. I have test for this so we do not revert schema changes in case of schema sync.

core/backend/src/BriefcaseManager.ts
khanaffan: I had to push because in schema sync we can get a schema changeset. I have remove this code and now we never do schemaSync.pull and so there will be no local changeset after pull.

core/backend/src/BriefcaseManager.ts
```diff
@@ -432,6 +444,59 @@ export class BriefcaseManager {
...
+
+    let currentIndex = db.changeset.index;
+    if (currentIndex === undefined)
+      currentIndex = (await IModelHost.hubAccess.queryChangeset({ accessToken: arg.accessToken, iModelId: db.iModelId, changeset: { id: db.changeset.id } })).index;
+
+    if (!arg.toIndex) {
+      throw new IModelError(ChangeSetStatus.ApplyError, "toIndex must be specified to revert changesets");
+    }
+    if (arg.toIndex > currentIndex) {
+      throw new IModelError(ChangeSetStatus.ApplyError, "toIndex must be less than the current index to revert changesets");
```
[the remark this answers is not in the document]
khanaffan: fixed

core/backend/src/BriefcaseManager.ts
khanaffan: This case is now also covered by the test.

core/backend/src/IModelDb.ts
```diff
@@ -2984,14 +2990,71 @@ export class BriefcaseDb extends IModelDb {
...
+        mergeRetryCount: arg.mergeRetryCount,
+        mergeRetryDelay: arg.mergeRetryDelay,
+        pushRetryCount: arg.pushRetryCount,
+        pushRetryDelay: arg.pushRetryDelay,
+        retainLocks: arg.retainLocks,
+      };
+      await skipSchemaSyncPull(async () => this.pushChanges(pushArgs));
+      this.clearCaches();
+    } finally {
+      this.abandonChanges();
```
JulijaRamoskiene, line 3054: schema lock is not always released e.g. if revertTimelineChanges() throws. suggest adding catch{} that would release schema lock if `retainLocks is false`

core/backend/src/BriefcaseManager.ts
```diff
@@ -432,6 +444,59 @@ export class BriefcaseManager {
     IModelJsFs.removeSync(changesetFile.pathname);
   }
 
+  /** @internal */
+  public static async revertTimelineChanges(db: IModelDb, arg: RevertChangesArgs): Promise<void> {
+    if (!db.isOpen || db[_nativeDb].isReadonly()) // don't use db.isReadonly - we reopen the file writable just for this operation but db.isReadonly is still true
```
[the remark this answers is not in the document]
JulijaRamoskiene: My point was that this comment is out of sync with code, as it is not true that 'we reopen the file writable just for `revertTimelineChanges` operation'. It seems like a copy - paste from pullAndApplyChangesets, where that comment is true as `pullChanges() in fact open file as writable just for pullAndApplyChangesets() operaion`. Suggest to change comment to something like : `// don't use db.isReadonly - it might be still true in cases when file is reopened writable just for this operation`

full-stack-tests/backend/src/integration/SchemaSync.test.ts
```diff
@@ -1035,4 +1036,216 @@ describe("Schema synchronization", function (this: Suite) {
...
+    assert.isDefined(findEl(el3));
+    assert.isDefined(findEl(el4));
+    assert.deepEqual(Object.getOwnPropertyNames(b1.getMetaData("TestDomain:Test2dElement").properties), ["p1", "p2"]);
+    // 6. Revert to timeline 2
+    await b2.revertAndPushChanges({ toIndex: 3, description: "revert to timeline 2" });
+
+    assert.equal((await getChanges()).at(-1)!.description, "revert to timeline 2");
+    await b1.pullChanges();
+    assert.isUndefined(findEl(el1));
+    assert.isUndefined(findEl(el2));
```
JulijaRamoskiene, line 1163: [Question] Why el1 and el2 are expected to be undefined in timeline2. Aren't they created in 2nd changeset:

image

Is `timeline 2` same as `2nd changeset`?

core/backend/src/BriefcaseManager.ts
```diff
@@ -76,6 +76,18 @@ export type PullChangesArgs = ToChangesetArgs & {
...
 
+/** Arguments for [[BriefcaseManager.revertTimelineChanges]]
+ * @public
+ */
+export type RevertChangesArgs = Optional<PushChangesArgs, "description"> & {
+  /** If present, a function called periodically during the download to indicate progress.
+   * @note return non-zero from this function to abort the download.
+   */
+  onProgress?: ProgressFunction;
+  toIndex: ChangesetIndex;
```
pmconne, line 87: Please document all properties of this type.

full-stack-tests/backend/src/integration/SchemaSync.test.ts
khanaffan: they are created in changeset 3. 

* The initialize schema sync push changeset 1
* Setup category push changeset 2
* Insert el1 & el2 make changeset 3

core/backend/src/IModelDb.ts
khanaffan: done

core/backend/src/BriefcaseManager.ts
khanaffan: done

pmconne (approved):

core/backend/src/BriefcaseManager.ts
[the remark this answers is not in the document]
khanaffan: I think I have fixed all you comments. See if you can close the comments.

PR #7118, opened 2024-08-29
Allow AccuDraw UI to specify quantity parsers.

bbastings:


ben-polinsky (approved):

core/frontend/src/AccuDraw.ts
```diff
@@ -1002,23 +1002,35 @@ export class AccuDraw {
...
   private stringToDistance(str: string): QuantityParseResult {
-    const parserSpec = IModelApp.quantityFormatter.findParserSpecByQuantityType(QuantityType.Length);
+    const parserSpec = this.getLengthParser();
     if (parserSpec)
       return parserSpec.parseToQuantityValue(str);
     return { ok: false, error: ParseError.InvalidParserSpec };
   }
 
+  /** Allow the AccuDraw user interface to specify bearing */
+  public get isBearingMode(): boolean { return false; }
```
mathieu-fournier, line 1018: Instead of a boolean flag, we could have a getter that detect if the `getAngleParser` returns a bearing or not. Like this we will never have a mismatch between the boolean and the actual parser used.

bbastings: I need the flag in processInputField...

PR #7119, opened 2024-08-30
Difference Validator Refactoring

StefanApfel-Bentley:
- Extracted the difference validation from the diagnostic visitor to a separate class.
- Added SchemaDifferenceVisitor and SchemaDifferenceValidator to the internal differencing api.
- Added conflict if KoQ on a property differs.
- Added conflict if schema reference alias already exists in the target schema context
- Added conflict if class modifiers don't match or violate


core/ecschema-editing/src/Differencing/SchemaDifferenceValidator.ts
```diff
@@ -0,0 +1,502 @@
...
+        this.addConflict({
+          code: ConflictCode.ConflictingPropertyKindOfQuantity,
+          schemaType: targetClass.schemaItemType,
+          itemName: targetClass.name,
+          path: targetProperty.name,
+          source: entry.difference.kindOfQuantity,
+          target: resolveLazyItemName(targetProperty.kindOfQuantity),
+          description: "The property has different kind of quantities defined.",
+        });
+      }
```
ColinKerr, line 404: There is nothing wrong with changing the koq so long as both KOQs have the same persistence unit.  Changing the persistence Unit when you change KOQs is not allowed ... everything else about the KOQs can be different

core/ecschema-editing/src/Differencing/SchemaDifferenceValidator.ts
```diff
@@ -0,0 +1,502 @@
...
+ * a class with the given baseClassName.
+ */
+async function derivedFrom(ecClass: ECClass | undefined, baseClassName: string): Promise<boolean> {
+  if (ecClass === undefined) {
+    return false;
+  }
+  if (ecClass && ecClass.name === baseClassName) {
+    return true;
+  }
+  return derivedFrom(await ecClass.baseClass, baseClassName);
```
ColinKerr, line 501: This concerns me because classes can have the same name but come from different classes ... I know there is some wiggly bits handling schemas with different names here but we have to ensure that we will not match if base classes are from a different schema but the input class name comes from this schema.


There should be at least one test for this explicit case

core/ecschema-editing/src/test/Differencing/Conflicts.test.ts
```diff
@@ -858,21 +776,183 @@ describe("Difference Conflict Reporting", () => {
...
+        expect(conflict).to.have.a.property("schemaType", "EntityClass");
+        expect(conflict).to.have.a.property("itemName", "TestItem");
+        expect(conflict).to.have.a.property("path", "TestProperty");
+        expect(conflict).to.have.a.property("source", "ConflictSchema.TestRelationship");
+        expect(conflict).to.have.a.property("target", "ConflictSchema.TestStruct");
+        expect(conflict).to.have.a.property("description", "Target class already contains a property with a different type.");
+      });
+    });
+
+    it("should find a conflict between on different kind of quantities on primitive properties", async () => {
```
ColinKerr, line 790: as noted the issue is when the persistence unit differs ... also this test name doesn't read very well.

core/ecschema-editing/src/Differencing/SchemaDifferenceValidator.ts
```diff
@@ -231,25 +212,20 @@ class SchemaDifferenceValidationVisitor implements ISchemaDifferenceVisitor {
...
-      const sourceSchemaItem = await this._sourceSchema.getItem<EntityClass>(entry.itemName);
-      if (sourceMixin === undefined) {
-        return;
-      }
-
-      const appliesTo = sourceMixin.appliesTo as Readonly<SchemaItemKey>;
-      if (appliesTo && !await derivedFrom(sourceSchemaItem, appliesTo.name)) {
+      const sourceMixin = await this._sourceSchema.lookupItem(addedMixin) as Mixin;
+      const sourceSchemaItem = await this._sourceSchema.getItem(entry.itemName) as EntityClass;
+      if (sourceMixin.appliesTo && !await this.derivedFrom(sourceSchemaItem, sourceMixin.appliesTo)) {
```
ColinKerr, line 222: I don't know why this needs any special code.  Both classes are from the source schema so you should be able to use ECClass.is

core/ecschema-editing/src/Differencing/SchemaDifferenceValidator.ts
```diff
@@ -462,6 +438,28 @@ class SchemaDifferenceValidationVisitor implements ISchemaDifferenceVisitor {
...
+    }
+
+    // First check for name which must be same in any case...
+    if (ecClass.name === baseClassKey.name) {
+      // ... then check if the class is in the same schema as the expected base class...
+      if(ecClass.schema.name === baseClassKey.schemaName)
+        return true;
+      // ... if not, whether it's in the other schema, which could be the case if the base class
+      // gets added to the target schema...
+      if(ecClass.schema.name === this._sourceSchema.name || ecClass.schema.name === this._targetSchema.name)
```
ColinKerr, line 458: I don't understand how the ecClass in this case could have the target schema name because you load the ecClass from the source.

I think the right check is that if the ecClass.schema.name === source AND baseClassKey.schemaName === target then we can consider them equal.


You should be able to make a test case where the current code fails when both classes have the same name but the target class is defined in a referenced schema and the source class is defined in the source schema.

Issue #6782, opened 2024-05-31
`CurveOps.collectChains` behavior doesn't match its documentation


dassaf4:
Documentation states that this method creates a `Path` when `planeTol === undefined`, but it is creating a `Loop` instead. User also states this API is "pretty goofy".

Same for `CurveOps.collectChainsAsLineString3d`, for which it also makes no sense to discuss `Loop` and `Path` in the doc as this method only announces linestrings.

The problem occurs because the underlying internal context `MultiChainCollector` gives `planeTol` a default tolerance when it comes in undefined. Hence, the context logic never branches on undefined planeTol.

dassaf4:
I think if we just remove the default tolerance from `planeTol` in callers and internal context, we restore the original behavior.

PR #7125, opened 2024-09-03
Prevent flipped facets and extraneous edges in some geometry constructions

dassaf4:
Addresses iTwin/itwinjs-backlog#1237:

Some geometry constructions were creating closed `Polyface`s with flipped (inward-facing) normals. There were several causes:
* `PolygonOps.orientLoopsCCWForOutwardNormalInPlace` was always reversing a facet orientation perpendicular to the input `outwardNormal`.
* `PolyfaceBuilder.addLinearSweep` created a capped solid with inward-facing cap facet normals when the input geometry's contour's normal did not align with its sweep direction.
* `PolyfaceClip.clipPolyfaceInsideOutside` generated flipped facets in each clip plane.
* `SweepContour.createForPolygon` was always flipping the frame returned by `FrameBuilder` when the polygon normal wasn't perpendicular to the input `defaultNormal`.

`PolyfaceQuery.cloneWithMaximalPlanarFacets` was also flipping facets, and also failing to remove some extraneous edges:
* Non-convex facets could be flipped by this method's reliance on `FrameBuilder` to reconstitute the normal of each maximized facet for the purposes of rotating to xy for topological computations (see next bullet). `FrameBuilder`'s greedy algorithm can produce a frame with reversed z-column if it starts accumulating cross products within a concave portion of the input polygon. Now we keep track of each maximally planar partition's average normal to avoid this quirk of `FrameBuilder`.
* A topological "regularization" operation is employed by this method as part of computing the boundary of a maximally planar partition of the mesh, each of which becomes a facet in the resulting `Polyface`. Regularization is essential for adding "bridge edges" from the partition outer boundary to holes in its interior, but this operation also prepares a polygon for future triangulation by adding additional interior edges. These redundant edges were surviving in the output. Now, if a partition has no holes, regularization is skipped since we expect neither to have bridge edges nor to triangulate. If a planar partition _does_ have holes, regularization is still necessary to create bridge edges, but this means these partitions may still be split with additional non-bridge edges added by regularization. See iTwin/itwinjs-backlog#1088.

Also addresses #6782:

The APIs for `CurveOps.collectChains` and `CurveOps.collectChainsAsLineString3d` were confusing, and did not agree with output.
* Removing the default value to `planeTolerance` fixes this discrepancy.
* Callers that were passing in `undefined` were sometimes erroneously getting `Loop`s returned; these callers will now only get `Path`s, as per documentation.

Additional changes:
* Tightened some unnecessarily broad input/return type unions.
* `FrameBuilder.createRightHandedFrame` did not return a frame aligned to the input `defaultUpVector` when the input data is linear (as it does in the nonlinear case).
* `PolyfaceQuery.announceBoundaryEdges` is more efficient.

### DIFF
--- a/core/geometry/src/curve/CurveOps.ts
+++ b/core/geometry/src/curve/CurveOps.ts
@@ -130,7 +130,7 @@ export class CurveOps {
    * @param planeTolerance tolerance for considering a closed chain to be planar. If undefined, only create Path. If defined, create Loops for closed chains within tolerance of a plane.
    * @returns chains, possibly wrapped in a [[BagOfCurves]].
    */
-  public static collectChains(fragments: AnyCurve[], gapTolerance: number = Geometry.smallMetricDistance, planeTolerance: number | undefined = Geometry.smallMetricDistance): AnyChain | undefined {
+  public static collectChains(fragments: AnyCurve[], gapTolerance: number = Geometry.smallMetricDistance, planeTolerance: number = Geometry.smallMetricDistance): AnyChain | undefined {
     const collector = new MultiChainCollector(gapTolerance, planeTolerance);
     for (const s of fragments) {
       collector.captureCurve(s);
--- a/core/geometry/src/curve/CurveOps.ts
+++ b/core/geometry/src/curve/CurveOps.ts
@@ -90,7 +90,7 @@ export class CurveOps {
    * @returns object with named chains, insideOffsets, outsideOffsets
    */
   public static collectInsideAndOutsideXYOffsets(fragments: AnyCurve[], offsetDistance: number, gapTolerance: number): { insideOffsets: AnyCurve[], outsideOffsets: AnyCurve[], chains?: AnyChain } {
-    const collector = new MultiChainCollector(gapTolerance);
+    const collector = new MultiChainCollector(gapTolerance, gapTolerance);
     for (const s of fragments) {
       collector.captureCurve(s);
     }
--- a/core/geometry/src/curve/CurveOps.ts
+++ b/core/geometry/src/curve/CurveOps.ts
@@ -130,7 +130,7 @@ export class CurveOps {
    * @param planeTolerance tolerance for considering a closed chain to be planar. If undefined, only create Path. If defined, create Loops for closed chains within tolerance of a plane.
    * @returns chains, possibly wrapped in a [[BagOfCurves]].
    */
-  public static collectChains(fragments: AnyCurve[], gapTolerance: number = Geometry.smallMetricDistance, planeTolerance: number = Geometry.smallMetricDistance): AnyChain | undefined {
+  public static collectChains(fragments: AnyCurve[], gapTolerance: number = Geometry.smallMetricDistance, planeTolerance?: number): AnyChain | undefined {
     const collector = new MultiChainCollector(gapTolerance, planeTolerance);
     for (const s of fragments) {
       collector.captureCurve(s);
@@ -146,7 +146,7 @@ export class CurveOps {
    * @param gapTolerance distance to be treated as "effectively zero" when assembling fragments head-to-tail. Also used for removing duplicate points in the stroked chains.
    * @param planeTolerance tolerance for considering a closed chain to be planar. If undefined, only create Path. If defined, create Loops for closed chains within tolerance of a plane.
    */
-  public static collectChainsAsLineString3d(fragments: AnyCurve[], announceChain: (chainPoints: LineString3d) => void, strokeOptions?: StrokeOptions, gapTolerance: number = Geometry.smallMetricDistance, planeTolerance: number | undefined = Geometry.smallMetricDistance) {
+  public static collectChainsAsLineString3d(fragments: AnyCurve[], announceChain: (chainPoints: LineString3d) => void, strokeOptions?: StrokeOptions, gapTolerance: number = Geometry.smallMetricDistance, planeTolerance?: number) {
     const collector = new MultiChainCollector(gapTolerance, planeTolerance);
     for (const s of fragments) {
       collector.captureCurve(s);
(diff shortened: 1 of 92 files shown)


DStradley:
@dassaf4 I ran ImageTests and it did not show any differences.

core/geometry/src/curve/CurveOps.ts
```diff
@@ -130,24 +130,24 @@ export class CurveOps {
    * @param planeTolerance tolerance for considering a closed chain to be planar. If undefined, only create Path. If defined, create Loops for closed chains within tolerance of a plane.
    * @returns chains, possibly wrapped in a [[BagOfCurves]].
    */
-  public static collectChains(fragments: AnyCurve[], gapTolerance: number = Geometry.smallMetricDistance, planeTolerance: number | undefined = Geometry.smallMetricDistance): AnyChain | undefined {
+  public static collectChains(fragments: AnyCurve[], gapTolerance: number = Geometry.smallMetricDistance, planeTolerance?: number): AnyChain | undefined {
```
saeeedtorabi, line 133: why did you remove the default value for `planeTolerance`?

core/geometry/src/curve/CurveOps.ts
```diff
@@ -130,24 +130,24 @@ export class CurveOps {
...
   /**
-   * Restructure curve fragments as Paths and Loops, to be stroked and passed into the callback.
+   * Restructure curve fragments, to be stroked and passed into the callback.
    * * Chain formation is dependent upon input fragment order, as a greedy algorithm is employed.
    * @param fragments fragments to be chained and stroked
    * @param announceChain callback to process each stroked Path and Loop
    * @param strokeOptions options for stroking the chains
    * @param gapTolerance distance to be treated as "effectively zero" when assembling fragments head-to-tail. Also used for removing duplicate points in the stroked chains.
-   * @param planeTolerance tolerance for considering a closed chain to be planar. If undefined, only create Path. If defined, create Loops for closed chains within tolerance of a plane.
+   * @param _planeTolerance unused, pass undefined
```
saeeedtorabi, line 147: ok I assume it's a public API and we cannot simply remove `_planeTolerance` so we are stuck with it!

core/geometry/src/curve/internalContexts/MultiChainCollector.ts
```diff
@@ -44,7 +44,7 @@ export class MultiChainCollector {
   private _gapTolerance: number;
   /** end point snap tolerance (assumed to be as tight or tighter than gapTolerance) */
   private _snapTolerance: number;
-  /** tolerance for choosing Path or Loop. If undefined, always Path. */
+  /** planarity tolerance, used to determine whether to return a Path or Loop in `grabResult(true)`. If undefined, always Path. */
```
saeeedtorabi, line 47: Not a big deal but I prefer to start docs with upper case: `Planarity tolerance, ...`
Same for lines 41, 43, and 45

core/geometry/src/curve/internalContexts/MultiChainCollector.ts
```diff
@@ -56,7 +56,7 @@ export class MultiChainCollector {
    * @param gapTolerance tolerance for calling endpoints identical
    * @param planeTolerance tolerance for considering a closed chain to be planar. If undefined, only create Path. If defined, create Loops for closed chains within tolerance of a plane.
    */
-  public constructor(gapTolerance = Geometry.smallMetricDistance, planeTolerance: number | undefined = Geometry.smallMetricDistance) {
+  public constructor(gapTolerance = Geometry.smallMetricDistance, planeTolerance?: number) {
```
saeeedtorabi, line 59: same question about default value removal

core/geometry/src/polyface/PolyfaceClip.ts
```diff
@@ -47,15 +47,7 @@ class ClipCandidate {
 }
 
 /**
- * A pair of PolyfaceBuilder objects, for use by clippers that emit inside and outside parts.
- * * There are nominally 4 builders:
```
saeeedtorabi, line 51: Any specific reason for removing all these doc details?!

core/geometry/src/polyface/PolyfaceQuery.ts
```diff
@@ -1024,43 +1018,59 @@ export class PolyfaceQuery {
...
+        for (const facetIndex of partition) {
+          if (visitor.moveToReadIndex(facetIndex))
+            if (PolygonOps.areaNormalGo(visitor.point, normal))
+              averageNormal.addInPlace(normal);
+        }
+        partitionNormals.push(Ray3d.createCapture(point0, averageNormal));
         planarPartitions.push(partition);
       }
     }
     const fragmentPolyfaces = PolyfaceQuery.clonePartitions(mesh, planarPartitions);
```
saeeedtorabi, line 1041: No need for `else`?

core/geometry/src/curve/CurveOps.ts
dassaf4: Because "undefined" has a specific behavior (return only Paths) that a default value was not allowing. (This bug is explained further in the PR description and issue #6782.)

core/geometry/src/curve/CurveOps.ts
dassaf4: Yes, unfortunately. I didn't think it worth deprecating and coming up with a new method name.

core/geometry/src/curve/internalContexts/MultiChainCollector.ts
dassaf4: Same answer!

core/geometry/src/polyface/PolyfaceClip.ts
dassaf4: They were incorrect.  Probably copied from native, where the implementation is different.

core/geometry/src/curve/internalContexts/MultiChainCollector.ts
dassaf4: Ah, missed that, thanks.

saeeedtorabi (approved):

Issue #7126, opened 2024-09-04
@itwin/ecschema-metadata - almostEqual is not a function


karolis-zukauskas:
**Describe the bug**
Application crashes when attempting to edit properties in PropertyGrid with unit conversions enabled (Frontend). Code from @itwin/ecschema-metadata crashes with "almostEqual is not a function". Something is wrong with the generated code for ESM (Frontend). The problematic file is `lib/esm/UnitConversion/UnitConversion.js`. If you apply the following patch:
-import * as almostEqual from "almost-equal";
+import almostEqual from "almost-equal";

everything works fine. If not - object almostEqual ends up being a module object, not the actual function. The actual function in Frontend in this scenario is `almostEqual.default` (but code doesn't call that - it tries to call `almostEqual` directly)

**To Reproduce**
I don't know .. I experience this while attempting to edit properties in PropertyGrid. The same behaviour can be reproduced in Presentation testApp. Reach out for more details.

**Desktop (please complete the applicable information):**

- OS: Windows
- iTwin.js Version: 4.8.5

PR #7129, opened 2024-09-05
Remove almost-equal package dependency

rschili:
fixes: https://github.com/iTwin/itwinjs-core/issues/7126

We include a package named "almost-equal" using this syntax:
`import * as almostEqual from "almost-equal";`
 
This results in an error "almost-equal is not a function" because the module exports its function by using the CommonJS syntax
`module.exports = almostEqual`
which results in us importing the entire module as an object rather than the function itself.
 
possible fixes are either doing
`import almostEqual = require("almost-equal");`
Which doesn't work as we have our module resolution strategy set to target ECMAScript modules which prevents this syntax.
The other alternative is
`import almostEqual from "almost-equal";`
Which I also am not allowed to do because that would require esModuleInterop and allowSyntheticDefaultImports to be enabled in tsconfig.jsonwhich we explicitly disallow.

After some discussion in the itwin team we decided to drop the dependency and just put a method directly into the code as it is very trivial.

The implementation is more complex than I imagined.
 - Our native almost_equal method uses a tolerance value which defaults to 1000
 - The almost_equal package uses an absolute and relativeError which both default to 2.2204460492503131e-16.
 - I literally tried Arun's generated method, but turns out with that many tests fail because they seem to have relied on the relativeError mechanism in almost-equal.
 - There is a chai-almost package which defaults to a tolerance of 1e-6 , but that just calculates the absolute difference and compares it.
 
After some thinking I decided to go with a single tolerance parameter which will be multiplied by 1 or one of the provided values whichever is bigger.


rschili:
Never mind, I am not allowed to use that syntax because our module resolution strategy is set to target ECMAScript modules. I cannot use the other syntax either because our tsconfig.json sets esModuleInterop to false, which would be needed for this.
So I guess there is no way we are allowed to use this module...
Looking for an alternative solution

pmconne (approved):

common/changes/@itwin/ecschema-metadata/master_2024-09-05-13-14.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-metadata",
+      "comment": "almost-equal dependency",
```
aruniverse, line 5:
```suggestion
      "comment": "remove almost-equal dependency",
```

core/ecschema-metadata/src/test/UnitConversion/Convert.test.ts
```diff
@@ -17,9 +16,19 @@ interface TestData {
...
 }
 
+/**
+ * Checks if two numbers are approximately equal within a given tolerance.
+ * @param a - The first number to compare.
+ * @param b - The second number to compare.
+ * @param epsilon - The tolerance within which the numbers are considered equal.
+ * @returns True if the numbers are approximately equal, false otherwise.
+ */
+function almostEqual(a: number, b: number, epsilon: number = 1.19209290e-7): boolean {
```
aruniverse, line 26: do we need to copy this into the test code? why cant we export from UnitConversion and import from there for the tests?

aruniverse: I notice the different epsilon values, but still think we can import the utility fn from the UnitConversion module and actually pass in the tolerance

aruniverse: no i was against the idea of adding a static method, we need to be more vigilant with not immediately going to statics in JS world. its fine to export it from the UnitConversion.ts module itself, it wont be apart of our public api unless you explicitly export it in the barrel file & the fn is marked as public. if api-extractor complains about it not having a release tag, just mark it as internal.

rschili: We are going to need a similar method in the quantity package for inverting units, which this package depends on. So I'd rather put it there and include it here. It's also similar to what native did. https://github.com/iTwin/imodel-native/blob/main/iModelCore/Units/src/Quantity.cpp#L112

christophermlawson (approved):

rschili:
This is more complex than I imagined. The native almost_equal method uses a tolerance value which defaults to 1000, while the almost_equal package uses an absolute and relativeError which both default to 2.2204460492503131e-16.
I just literally took Arun's generated method, but turns out with that many tests fail because they seem to have relied on the relativeError mechanism.
There is a chai-almost package which defaults to a tolerance of 1e-6 , but that also just calcilates the absolute difference and compares it.
 
It seems like there is no standard of what "almost" means. I'll think about it a bit and put something out for review tomorrow.

core/ecschema-metadata/src/test/UnitConversion/Convert.test.ts
rschili: Hey @aruniverse I've updated the code according to the latest discussions, can you have a quick look? Thanks

common/changes/@itwin/core-quantity/rob-almost-equal-fix_2024-09-06-07-05.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-quantity",
+      "comment": "New almostEqual method for comparing numbers",
```
aruniverse, line 5: this is an internal api, not exposed to the user, therefore no need to mention it in the changelog

```suggestion
      "comment": "",
```

aruniverse (approved):

ColinKerr (approved):

PR #7134, opened 2024-09-06
new circular arc from start + end + radius

dassaf4:
New constructor `Arc3d.createCircularStartEndRadius` takes start point, end point, radius, and either a third point in the arc plane or a normal vector. Invalid inputs result in a line segment, like other start + end arc constructors.

Resolves iTwin/itwinjs-backlog#1239 for Booster editing tools.


core/geometry/src/curve/Arc3d.ts
```diff
@@ -445,6 +446,36 @@ export class Arc3d extends CurvePrimitive implements BeJSONFunctions {
     }
     return LineSegment3d.create(start, end);
   }
+  /**
+   * Create a circular arc defined by start and end points and radius.
```
saeeedtorabi, line 450: Because an arc cannot be defined uniquely bt start, end, and radius I will mention the helper point too.

core/geometry/src/curve/Arc3d.ts
```diff
@@ -445,6 +446,36 @@ export class Arc3d extends CurvePrimitive implements BeJSONFunctions {
...
+   * @param start start point of the arc
+   * @param end end point of the arc
+   * @param helper a third point near the arc in its plane, or a vector in the direction of the arc normal
+   * @returns the constructed arc, or line segment if desired arc cannot be constructed
+   */
+  public static createCircularStartEndRadius(start: Point3d, end: Point3d, radius: number, helper: Point3d | Vector3d): Arc3d | LineSegment3d {
+    const semiChordLen2 = 0.25 * start.distanceSquared(end);
```
saeeedtorabi, line 457: When I implemented `createCircularStartTangentEnd` I added couple lines of doc (inside the function body) explaining the algorithm I used. It makes it much easier to read and understand the function body. Please do the same here.

saeeedtorabi (changes requested):

core/geometry/src/test/curve/Arc3d.test.ts
```diff
@@ -717,6 +719,60 @@ describe("Arc3d", () => {
     GeometryCoreTestIO.saveGeometry(allGeometry, "Arc3d", "createCircularStartTangentEnd");
     expect(ck.getNumErrors()).equals(0);
   });
+  it.only("createCircularStartEndRadius", () => {
```
saeeedtorabi, line 722: `only` and `true,true` should be removed.

bbastings (approved):
Works as advertised, I hooked this up to the place arc tool (local copy for now) using a help point (not the vector).

core/geometry/src/curve/Arc3d.ts
dassaf4: Eh, it's a small detail. We want that first sentence to be short.

core/geometry/src/test/curve/Arc3d.test.ts
dassaf4: Ack! Good catch.

core/geometry/src/curve/Arc3d.ts
dassaf4: Done!

saeeedtorabi (approved):

core/geometry/src/curve/Arc3d.ts
```diff
@@ -445,6 +446,36 @@ export class Arc3d extends CurvePrimitive implements BeJSONFunctions {
...
+      return LineSegment3d.create(start, end);
+    const height = Math.sqrt(radius2 - semiChordLen2);
+    const normal = Vector3d.createZero(this._workVectorU);
+    const vecToCenter = Vector3d.createZero(this._workVectorV);
+    if (helper instanceof Point3d)
+      start.crossProductToPoints(helper, end, normal);
+    else
+      normal.setFrom(helper);
+    if (!normal.normalizeInPlace() || !normal.crossProductStartEnd(start, end, vecToCenter).scaleToLength(height, vecToCenter))
+      return LineSegment3d.create(start, end);
```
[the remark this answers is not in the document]
bbastings: Ugh, forgot to "submit" my review again. :)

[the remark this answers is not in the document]
dassaf4: Yeah, good point. I was thinking of maintaining consistency, but I never liked returning `LineSegment3d` anyway. Done.

PR #7137, opened 2024-09-06
Change MapLayerImageryProvider.supportsMapFeatureInfo from alpha to public

wilmaier:
This getter is needed in @itwin/map-layers package of viewer-components-react to show the map feature info tool only when a provider is registered to support it.

### DIFF
--- a/core/frontend/src/tile/map/MapLayerImageryProvider.ts
+++ b/core/frontend/src/tile/map/MapLayerImageryProvider.ts
@@ -67,7 +67,7 @@ export abstract class MapLayerImageryProvider {
   /** @internal */
   public get status() { return this._status; }
 
-  /** @alpha */
+  /** @beta */
   public get supportsMapFeatureInfo() { return false; }
 
   public resetStatus() { this.setStatus(MapLayerImageryProviderStatus.Valid); }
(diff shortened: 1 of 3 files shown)


core/frontend/src/tile/map/MapLayerImageryProvider.ts
```diff
@@ -67,7 +67,7 @@ export abstract class MapLayerImageryProvider {
   /** @internal */
   public get status() { return this._status; }
 
-  /** @alpha */
+  /** @beta */
   public get supportsMapFeatureInfo() { return false; }
```
pmconne, line 71:
```suggestion
  public get supportsMapFeatureInfo(): boolean { return false; }
```

core/frontend/src/tile/map/MapLayerImageryProvider.ts
```diff
@@ -67,7 +67,7 @@ export abstract class MapLayerImageryProvider {
   /** @internal */
   public get status() { return this._status; }
 
-  /** @alpha */
+  /** @beta */
```
pmconne, line 70: Document this. What's it for, why do I need it?

core/frontend/src/tile/map/MapLayerImageryProvider.ts
wilmaier: Was asked to make this public instead of beta now

pmconne (approved):

wilmaier:

PR #7138, opened 2024-09-09
Fixed the issue with invalid XYZ intersections.

saeeedtorabi:
The XYZ intersections of middle children of Path, Loop, and CurveChainWithDistanceIndex were found which is wrong. Only first and last child intersections should be found. Also fixed issues with intersections when arcs were extended.


dassaf4:
Pushed some changes:
* Store the correct period-shifted local exterior arc fraction in the `childDetail` of the chain details returned by `CurveCurve` intersection methods on an extended `CurveChainWithDistanceIndex`. Previously, the detail for an extended intersection of a chain with first/last arc might have reported the arc's exterior fraction with the wrong sign.
  * `CurveChainWithDistanceIndex.clonePartialCurve` can now take such chain details as input and will correctly extend the chain using the child detail. Without such local detail, this method resorts to approximating the local exterior fraction using the distance index, and this can result in inaccurate extensions of chains starting/ending with an elliptical arc.
* `AngleSweep`: refactor angle -> fraction logic so we can compute period-shifted exterior arc fractions off the specified end of [0,1].
* Add `StrokeOptions` to `CurveChainWithDistanceIndex` clone methods to customize the distance index.
* Clarifications to doc in response to user queries.

dassaf4 (approved):
Looks good. Excellent test coverage for intersecting coplanar curves! I see a new XYZ test verifying that curves in parallel planes *don't* intersect, but I do not see any new XYZ tests verifying actual intersections between a) non-planar and b) non-coplanar curves. Can you add some? Also, please review my additions. Thanks!

Issue #7139, opened 2024-09-09
add baseUri to BlobContainer.TokenProps interface


YashodipD:
**Is your feature request related to a problem? Please describe.**
ContainerService.requestToken() does not return base uri in it's response. base uri is required to perform any operation on workspace container / db.  There is no other method which expose base uri for subsequent usage. 

**Describe the solution you'd like**
ContainerService.requestToken() should return base uri. for that we need to modify BlobContainer.TokenProps interface.

PR #7140, opened 2024-09-09
Update BlobContainer.TokenProps interface

YashodipD:
ContainerService.requestToken() returns TokenProps, inside the TokenProps, base uri seems to be missing. 

base uri points to the location where the actual workspace container lives, it is required for every operation. Internal api (storage-container) also return base uri, we just need to expose it through interface and add it to response.

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -446,6 +446,7 @@ export namespace BlobContainer {
         ownerGuid?: GuidString;
     }
     export interface TokenProps {
+        baseUri?: string;
         expiration: Date;
         metadata: Metadata;
         provider: Provider;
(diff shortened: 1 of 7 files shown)


YashodipD:
fixex https://github.com/iTwin/itwinjs-core/issues/7139

YashodipD:
I made it optional so no breaking change will be introduced.

kabentley:
I'm not sure where/why you need this (the caller to `requestToken` must supply it), but making it optional makes it useless. The whole interfaces is @beta, I'd just add it.

YashodipD:
I am creating a `client` using `BlobContainerService` which calls `requestToken`. BlobContainerService accepts baseUrl, It is optional. I think url accepted by BlobContainerService is used to fetch information about `storage-container` api and the base uri which I am looking for is the url of actual blob container (https://xyz.blob.core.windows.net).

pmconne (approved):

PR #7144, opened 2024-09-10
Improved DisplayStyle cloning performance

ViliusRuskys:
`DisplayStyle.excludedElements` array may potentially have hundreds of thousands of element ids. Splicing the array to remove each unmapped element has a major performance impact.
Tested this with an iModel that had multiple DisplayStyles with 250k + excludedElements. Time to process went from ~1minute per DisplayStyle to <1s per DisplayStyle.


pmconne:
Good fix but

> may potentially have hundreds of thousands of element ids

Do we understand (or can we find out) why so many?

pmconne (approved):

pmconne:
> It should still be possible with recent connector versions as well. I guess you could just select a bunch of elements, hide them and create a view definition. If the selection is arbitrary enough, then the only solution is to exclude elements by id from the saved view.
The iModel I encountered this issue with had changes done by Microstation and IFC connectors and is quite recent.

@jffmarker @abeesh under what circumstances would MicroStation or IFC connector stick 250,000 elements into a display style's "excluded elements" list? (Or for that matter, any elements at all - we invented that property for the Revit connector's "pseudo-2d" models).

PR #7145, opened 2024-09-10
Pass includeIncomplete flag to queryGraphicRepresentations

eringram:
- Fixed a bug in FrontendTiles.ts that prevented exports with a status other than "Complete" from being returned by `queryMeshExports`, even when they should be included via the `includeIncomplete` flag
- Added unit tests for this
- Also fixed a related issue where because we were assuming all exports were complete, we were also assuming they all had a `_links` property. Fixed by checking if `_links` is undefined.
- Added a note that the `GraphicRepresentationStatus` enum is outdated to what statuses the mesh export service can return. This does not throw an error anywhere, but it should be addressed in a future PR.


markschlosseratbentley (approved):
LGTM

eringram:
Currently waiting on PR #7146 that updates packages to fix the failing rush audit

andremig-bentley (approved):
Mock tests LGTM

eringram:

PR #7146, opened 2024-09-11
mitigate GHSA-qwcr-r2fm-qrc7

ben-polinsky:
Handle audit failures:
`body-parser`: https://github.com/advisories/GHSA-qwcr-r2fm-qrc7
~`path-to-regexp`: https://github.com/advisories/GHSA-9wv6-86v2-598j~

Packages updated:
~`sinon` -> `^18.0.1`~
`body-parser` -> `^1.20.3`
`express` -> `^4.20.0`
`fetch-mock` -> `~11.1.3`

Working with @tcobbs-bentley to resolve the failing ios tests after upgrade. Seems to be a unicode issue related to [this change in path-to-regexp](https://github.com/pillarjs/path-to-regexp/blob/v8.0.0/src/index.ts#L3). Working theory (from Travis): 

> it looks like both ID_Start and ID_Continue are special regex tokens. I found the following, and I'm not sure if it's related or not: https://stackoverflow.com/questions/71155109/invalid-regular-expression-invalid-property-name-in-character-class. Digging further, it looks like this regex will only work with full ICU (International Components for Unicode). The mobile build of Node does not have this (and probably cannot have it any time soon).
 
Also upgrade actions/upload-artifact@v2 -> actions/upload-artifact@v4 since v2 is no longer supported. This task is used in our extract-api action. 

We'll get this through to unblock builds and take another look tomorrow.


aruniverse (approved):

eringram (approved):

pmconne (approved):

hl662:

PR #7150, opened 2024-09-12
Added test cases to test support for boolean exps in select statements

soham-bentley:
This is a following PR for the following PR
https://github.com/iTwin/imodel-native/pull/857


core/backend/src/test/ecdb/ECSqlAst.test.ts
```diff
@@ -475,6 +475,26 @@ describe("ECSql Abstract Syntax Tree", () => {
       assert.equal(test.expectedECSql, await toNormalizeECSql(test.expectedECSql));
     }
   });
+  it.only("parse Boolean Exp in Select statements", async () => {
```
pmconne, line 478:
```suggestion
  it("parse Boolean Exp in Select statements", async () => {
```

pmconne (approved):

core/backend/src/test/ecdb/ECSqlAst.test.ts
soham-bentley: done

PR #7152, opened 2024-09-12
Update schema deserialization and serialization to handle schema evolutions

RohitPtnkr1996:
The PR contains changes to the ecschema-metadata to make it more flexible when handling schemas that have unknown elements used with newer versions of ECDb.

A schema with a greater ECXml version than the one currently supported by ECDb can be deserialized, but it cannot be imported into the Db, nor can it be serialized to Xml or JSON.


pmconne (approved):

common/api/ecschema-metadata.api.md
```diff
@@ -425,6 +425,8 @@ export enum ECObjectsStatus {
     // (undocumented)
     MissingSchemaUrl = 35070,
     // (undocumented)
+    NewerSchemaVersion = 35081,
```
ColinKerr, line 428: This could be confused with the version of the schema rather than the version of the EC Spec.

common/api/ecschema-metadata.api.md
```diff
@@ -488,6 +490,14 @@ export class ECVersion {
...
 }
 
+// @internal
+export interface ECXmlVersion {
+    // (undocumented)
+    readVersion: number;
+    // (undocumented)
+    writeVersion: number;
+}
+
```
ColinKerr, line 500: I want to move away from calling this the ECXml version because it applies to the JSON as well ... we even do it in this code where we set the xml version from the json version.  I propose ECSpec instead of ECXml.

common/api/ecschema-metadata.api.md
RohitPtnkr1996: Yeah, true !
I'll update it to NewerECSpecVersion

common/api/ecschema-metadata.api.md
RohitPtnkr1996: Cool, I'll create a backlog item on native to track it.

ColinKerr (approved):

Issue #7142, opened 2024-09-09
LogException not checking for null error


Jake-Screen:
**Describe the bug**
When encountering an error that is null when trying to be logged, an exception will occur. The way we found this issue was through this itwinui issue we also ran into. The rapid resizing would trigger a resize observer loop error, which would trigger an ErrorEvent. 

![image](https://github.com/user-attachments/assets/f73a8823-8183-4c26-8a45-2b5376a39946)

this error which is null was then passed down to logException method [here](https://github.com/iTwin/itwinjs-core/blob/b9e7a4cf8d15e8e7d44b8056dbecd3a00ed5fef7/core/bentley/src/Logger.ts#L258-L262). Once there, the method would try to access the constructor property of the null error object, causing our app to fail.

![image](https://github.com/user-attachments/assets/6423bf4f-0e8a-426e-beaf-68f687f9db6f)

**To Reproduce**
The root cause that threw the initial error was on the itwinui side, but has been fixed. The underlying error handing does not check if the error object is null before trying to operate with it.

- OS - windows
- Browser - chrome
Version: 1000.0.79 (104b0d4) 
iTwin.js: 4.9.0-dev.22 
iTwinUI: 3.13.4 
AppUI: 4.16.4 
React: 17.0.2 
Node.js: 20.9.0 
Electron: 29.1.0

PR #7153, opened 2024-09-12
Fix Logger.logException when err is undefined or null

tcobbs-bentley:
Fix for https://github.com/iTwin/itwinjs-core/issues/7142.

### DIFF
--- a/core/bentley/src/Logger.ts
+++ b/core/bentley/src/Logger.ts
@@ -246,6 +246,12 @@ export class Logger {
   }
 
   private static getExceptionMessage(err: unknown): string {
+    if (err === undefined) {
+      return "Error: err is undefined.";
+    }
+    if (err === null) {
+      return "Error: err is null.";
+    }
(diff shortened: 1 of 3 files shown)


core/bentley/src/Logger.ts
```diff
@@ -246,6 +246,12 @@ export class Logger {
   }
 
   private static getExceptionMessage(err: unknown): string {
+    if (err === undefined) {
```
ben-polinsky, line 249: If we think it's valuable to convey the error was specifically null or undefined this is cool. But if not, it's an opportunity for the only good use of JS loose equality:

`err == null` will cover both undefined and null.

ben-polinsky (approved):

core/bentley/src/Logger.ts
tcobbs-bentley: Note that the exact string used here also shows up in the updated unit test.

ben-polinsky: Yeah I can see that. Sounds good,

swwilson-bsi (approved):

PR #6890, opened 2024-06-21
[feasibility] switch from `mocha` to `vitest` for esm & native ts based testing

aruniverse:

PR #7165, opened 2024-09-16
Introduce Vitest

hl662:
Continuing off of the work from #6890 

For `core-bentley`,`core-geometry`, `core-common`:
- Refactors all tests
- Uses vitest for coverage as well, needs explicit `@vitest/coverage-istanbul` (I'm open to switching to node's native code coverage `v8` too, no opinions on either)
- Removed explicit `chai`, `chai-as-promised`, `mocha`, `nyc` dependencies

OneAtATime tests in core-bentley are skipped for now, haven't finished refactoring them.

Switching the integration tests to vitest is out of the scope of this PR, so just looking at the unit tests within each sub repo instead. We'll investigate this more after.

A couple more issues to circle back to later on as a result of this: #7200 #7201


core/geometry/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "iTwin.js Core Geometry library",
   "main": "lib/cjs/core-geometry.js",
   "module": "lib/esm/core-geometry.js",
-  "typings": "lib/cjs/core-geometry",
+  "typings": "lib/esm/core-geometry",
```
aruniverse, line 7: dont forget to revert

core/geometry/src/test/setupTests.ts
```diff
@@ -0,0 +1 @@
+import "../core-geometry"; // Needed to avoid circular import runtime errors with vitest
```
aruniverse, line 1: please add a comment with a link to the vitest docs you found this in

full-stack-tests/core/webpack.config.js
```diff
@@ -47,6 +47,10 @@ function createConfig(shouldInstrument) {
         path: require.resolve("path-browserify"),
         stream: require.resolve("stream-browserify"),
         zlib: require.resolve("browserify-zlib"),
+        tty: require.resolve("tty-browserify"),
```
aruniverse, line 50: where are these additions coming from?

.vscode/settings.json
```diff
@@ -7,7 +7,7 @@
     "source.fixAll.markdownlint": "explicit"
   },
   "editor.insertSpaces": true,
-  "editor.formatOnSave": true,
+  // "editor.formatOnSave": true,
```
aruniverse, line 10: dont forget to revert this file

full-stack-tests/core/webpack.config.js
hl662: [error.txt](https://github.com/user-attachments/files/17138427/error.txt)
See the above for the failures in particular, I saw these errors pop up when I picked up your old PR, I assumed it was related to changing the typings, but after reverting the typings changes it still stuck around...

.vscode/settings.json
hl662: done

core/geometry/src/test/setupTests.ts
hl662: added

core/geometry/package.json
hl662: done

core/common/vitest.config.mts
```diff
@@ -0,0 +1,12 @@
+import { defineConfig } from 'vitest/config';
+export default defineConfig({
+  test: {
+    dir: "src",
+    coverage: {
+      provider: "istanbul", // akin to nyc,
```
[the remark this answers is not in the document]
hl662: agh, forgot to add that, just added

core/bentley/package.json
```diff
@@ -4,22 +4,22 @@
   "description": "Bentley JavaScript core components",
   "main": "lib/cjs/core-bentley.js",
   "module": "lib/esm/core-bentley.js",
-  "typings": "lib/cjs/core-bentley",
+  "typings": "lib/esm/core-bentley",
```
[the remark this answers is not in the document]
hl662: done

full-stack-tests/core/webpack.config.js
aruniverse: well looking at the errors, prob coming from a dependency update, which is fine...

core/common/src/test/FeatureSymbology.test.ts
```diff
@@ -84,42 +84,39 @@ describe("FeatureAppearance", () => {
...
-        const app = FeatureAppearance.fromJSON(appProps);
-        expect(app.viewDependentTransparency).to.equal(expected);
-        expect(app.toJSON().viewDependentTransparency).to.equal(expected);
-      }
-
-      test({ }, false);
-      test({ transparency: undefined }, false);
-      test({ transparency: 1 }, false);
-      test({ transparency: 0 }, false );
+  it("view-dependent transparency serialization", () => {
```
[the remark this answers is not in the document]
hl662: nested describes can work, but it doesn't seem necessary - it doesn't make use of any test hooks, there's only one test case nested inside the outer one, so I figured let's just move them out and coalesce into just one test case

core/geometry/package.json
```diff
@@ -39,28 +39,16 @@
...
     "rimraf": "^3.0.2",
-    "typescript": "~5.3.3"
+    "typescript": "~5.3.3",
+    "vitest": "^2.1.0"
   },
   "dependencies": {
     "@itwin/core-bentley": "workspace:*",
     "flatbuffers": "~1.12.0"
-  },
-  "nyc": {
```
hl662, line 58: @aruniverse @saeeedtorabi did this work in the past? I copied the functionality of the coverage below into geometry's vitest.config.mts and it's now failing coverage, saying `Coverage for functions (53.06%) does not meet global threshold (85%)`

.vscode/settings.json
```diff
@@ -104,7 +100,5 @@
     "NODE_ENV": "development"
   },
   "typescript.tsdk": "common/temp/node_modules/typescript/lib",
-  "cSpell.words": [
-    "queryables"
-  ],
-}
\ No newline at end of file
+  "cSpell.words": ["queryables"]
```
aruniverse, line 103: move this to the cspell file, and revert changes to this file since they all come from prettier formatter

core/bentley/package.json
```diff
@@ -12,14 +12,14 @@
     "directory": "core/bentley"
   },
   "scripts": {
-    "build": "npm run -s build:cjs && npm run -s build:esm",
+    "build": "npm run -s build:esm && npm run -s build:cjs",
```
aruniverse, line 15: lets revert this change

core/geometry/package.json
saeeedtorabi: Currently on the master branch our coverage for `core-geometry` is 
```
Statements   : 93.53% ( 38252/40895 )
Branches     : 82.55% ( 15344/18587 )
Functions    : 92.47% ( 6318/6832 )
Lines        : 93.53% ( 36824/39368 )
```

@dassaf4 Any idea why @hl662  is getting 53% coverage?!

core/bentley/src/test/OneAtATime.test.ts
```diff
@@ -3,20 +3,21 @@
...
-const expect = chai.expect;
+// const assert = chai.assert;
+// const expect = chai.expect;
 
-chai.use(chaiAsPromised);
+// chai.use(chaiAsPromised);
 describe("OneAtATime test", () => {
 
-  it("OneAtATime", async () => {
+  it.skip("OneAtATime", async () => {
```
aruniverse, line 20: idk if i ever finished refactoring this test

core/geometry/package.json
hl662: I'll check for regressions - maybe the one in master excludes certain file paths

core/bentley/src/test/OneAtATime.test.ts
hl662: This is a tricky one I skipped for now as well, I need to put this in the PR description

core/geometry/package.json
hl662: vitest's `expect` has different syntaxes than chai's `expect`, perhaps when we use the old syntax, that isn't caught by the coverage - I'll refactor all of the `expect` methods and see if it helps

aruniverse: take a look at `tools/build/.nycrc`

hl662: none on the test files - I'm looking at the local coverage report right now, there's no `test.ts`

dassaf4: I see only one core-geometry test file with the `expect` calls changed to `expectTo`. I assume this is an experiment and the goal is to match the current coverage percentages with the new test framework. Otherwise, core-geometry changes look good so far. Let us know if you need help @hl662!

core/bentley/src/test/OneAtATime.test.ts
aruniverse: make a todo in the pr description, and create issues so these things dont get lost

core/geometry/package.json
hl662: image
Update: I updated the vitest config coverage's `include` and `exclude` and got the numbers above, we might be missing some covered lines and statements by a small fraction, lmk what you think @saeeedtorabi @dassaf4

core/bentley/src/test/OneAtATime.test.ts
hl662: Issue opened: https://github.com/iTwin/itwinjs-core/issues/7200

core/bentley/package.json
hl662: done

.vscode/settings.json
hl662: moved, and reverted the other formatting

aruniverse (approved):
approving on behalf of core bentley & common

core/geometry/src/test/clipping/ClipPrimitives.test.ts
```diff
@@ -320,7 +320,7 @@ describe("ClipPrimitive", () => {
   let polygonB: Point3d[];
   let polygonC: Point3d[];
 
-  before(() => {
+  beforeEach(() => {
```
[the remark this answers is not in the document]
hl662: yes, I'll fix that, thanks!

eringram (approved):

core/geometry/src/test/clipping/ClipPrimitives.test.ts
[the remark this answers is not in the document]
eringram: Np. Possibly the same change applies to ClipVector.test.ts

core/geometry/package.json
dassaf4: Sweet! I'm fine with those numbers. These are estimates of estimates, so the decimal differences are unimportant. Besides, according to L. Kronecker, "God created the integers, all else is the work of man." ;)

core/geometry/src/test/clipping/ClipPrimitives.test.ts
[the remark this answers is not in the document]
dassaf4: Good catch @eringram. I think there were other `beforeEach` changes in core-geom...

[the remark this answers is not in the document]
hl662: I checked there were only those 2 - vitest **needs** things to switch from `before` -> `beforeAll` and `after` -> `afterAll`. `beforeEach` and `afterEach` stays the same between vitest and mocha. If there were any other beforeEach imports that were missing, we would have caught it at build time.

dassaf4 (approved):

core/geometry/src/test/clipping/ClipPrimitives.test.ts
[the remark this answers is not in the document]
hl662: okay, fixed both

core/geometry/src/test/geometry3d/Angle.test.ts
```diff
@@ -129,7 +130,7 @@ describe("Angle.AlmostEqual", () => {
     ck.testFalse(Angle.createRadians(1.0e-20).isExactZero);
 
     ck.checkpoint("End Angle.AlmostEqual");
-    expect(ck.getNumErrors()).equals(0);
+    expect(ck.getNumErrors()).toEqual(0);
```
dassaf4, line 133: Are these equals -> toEqual changes still needed?

hl662: not necessarily, functionally both behave the same, but I was going to leave it in there as an example of switching over from chai to vitest's testing syntax. chai, out of the box, doesn't support asynchronous assertions like how vitest does. The author of vitest [recommends](https://github.com/vitest-dev/vitest/issues/2843#issuecomment-1424352509) to not mix and match, but if we want to continue using the assertions API, we can continue importing that from within vitest. I can remove it this, to avoid inconsistency

dassaf4: I guess we could use some best practices for vitest going forward. What is the difference between `expect(1).equals(0)` and `expect(1).toEqual(0)`?

hl662: For primitive types, they are the exact same, but objects are more complex. In vitest, when comparing the expected value to the actual value, you can choose between toBe() or toEqual(). The former is strict equality (think ===). The latter **deeply**, recursively checks the equality of all property fields, not the object's identity. [Reference](https://stackoverflow.com/questions/45195025/what-is-the-difference-between-tobe-and-toequal-in-jest). Chai's .[equals](https://www.chaijs.com/api/bdd/#method_equal)() (or `eq`) would check an object is strictly equal, while for deeply equals you would use `eql`. Unfamiliar people would not be able to tell as easily in chai, that `equal` and `eql` behave differently, whereas vitest tries to be more explicit in their method naming.

dassaf4: Ah, I see. Nice explanation! All of the occurrences of ".equals" in core-geometry operate on numbers or booleans. So to stick with vitest syntax, it would be more natural to switch them to `.toBe`. If you want to do this, feel free. Otherwise, we'll do it.

ben-polinsky (approved):

core/geometry/src/test/geometry3d/Angle.test.ts
hl662: Yes, toBe would be appropriate in those cases. I'll defer to you to update the syntax, I'll be giving a demo internally next week on some things to catch when refactoring

aruniverse: David can we do that in a separate follow up pr then?

dassaf4: Looking forward to your demo!

aruniverse:
> BTW, just pulled and tried out vitest, and the multithreading is _so nice_. All core-geometry tests done in 50s. Wow.
- @dassaf4 
- 
just in case anyone comes back to this pr, dont want you to miss this comment :)

PR #7170, opened 2024-09-17
Allow context reality model visibility to be toggled on/off

mdastous-bentley:
Similarly to map-layers, we want to be a able to turn off the display of a context reality model, without having to go trough the detach / attach process over and over..

Also, it should be possible to persist this visibility flag part of the view serialization process so that a reality can be loaded again, but remains invisible.


mdastous-bentley:
@pmconne does it look good to you so far?

pmconne:
> @pmconne does it look good to you so far?

Please explain the motivation in the PR description.

I have concerns that just making `draw` a no-op will introduce surprising behavior. There is plenty of code that iterates reality models and/or tile trees without drawing them - what are they supposed to do if it's flagged as invisible? (Currently, they will have no idea, and behave as if it is visible).

mdastous-bentley:
@pmconne A more brutal approach would to skip the model in `DisplayStyleState.forEachRealityModel`, but I feel like it could cause confusion too.

pmconne:
> when want to turn off the display of a reality model, without having to go trough the detach / attach process over and over..
> Also, it should be possible to persist this visibility flag part of the view serialization process so that a reality can be loaded again, but remains invisible.

Who specifically wants this - a user, or a developer? In what app - iTwin Experience?
Are we talking about managing the visibility of scene objects within an "iTwin scene", which is the province of the app, not the core?

MarcBedard8:
This request come from the "Unification widget" project where we will merge reality-data, map-layer, annotations and feature into the same widgets.

MarcBedard8 (approved):

pmconne:
> iModel support this concept with model selector and we have a reality data model in the iModel which expose that behavior.

The context reality models have to follow same behavior as those do. e.g., they don't contribute to the computed scene extents, they aren't included when we iterate tile trees in the view, etc.

core/frontend/src/tile/RealityModelTileTree.ts
```diff
@@ -577,6 +579,7 @@ export namespace RealityModelTileTree {
     protected _classifier?: SpatialClassifierTileTreeReference;
     protected _mapDrapeTree?: TileTreeReference;
     protected _getDisplaySettings: () => RealityModelDisplaySettings;
+    protected _contextRealityModel?: ContextRealityModel;
```
mdastous-bentley, line 582: @pmconne are you ok having optional `ContextRealityModel` object on  `RealityModelTileTree.ReferenceBaseProps` ?

pmconne: Strictly for accessing the `invisible` flag? Would prefer looser coupling like `protected _isInvisible?: () => boolean`.

mdastous-bentley: @pmconne updated

core/common/src/ContextRealityModel.ts
```diff
@@ -234,6 +241,10 @@ export class ContextRealityModel {
   public readonly description: string;
   /** An optional identifier that, if present, can be used to elide a request to the reality data service. */
   public readonly realityDataId?: string;
+  /** If true, reality model is not drawn.
+   * @beta
+  */
+  public readonly invisible?: boolean;
```
pmconne, line 247: You probably want to be able to change this, right (not `readonly`)? If so we will want the setter to raise an event so we can, e.g., invalidate a viewport's scene when the visibility changes.

mdastous-bentley: I had the same thought process as you but then realized this can be achieved by calling `ContextRealityModels.update`, or maybe I'm missing something and that's a terrible idea?

pmconne: Not terrible but more awkward. ContextRealityModel already has 3 mutable properties with corresponding events anyway.

mdastous-bentley: alright, implemented a new event.

pmconne (changes requested):
Make `DisplayStyleState.forEachRealityTileTreeRef` skip over invisible models. You can then delete your `invisible` check in `addToScene`. It looks like that's the only place you needed `_isInvisible` so that can be deleted too.

pmconne (changes requested):
No tests?

mdastous-bentley:
> `forEachRealityTileTreeRef`

Doing so will make the reality model behaves like it has not been attached.  You think that's a better over a non-op draw?

pmconne:
> > `forEachRealityTileTreeRef`
> 
> Doing so will make the reality model behaves like it has not been attached. You think that's a better over a non-op draw?

Isn't that the objective? If you don't skip its tile tree when it's invisible then, for example, fitting the view will cause fit to a volume that includes this invisible reality model. I'd like to see a unit test that proves that `ViewState.computeFitRange` does not include the bounding boxes of invisible reality models.

What are the circumstances in which you would want to process the tile tree of an invisible context reality model?

mdastous-bentley:
@pmconne Im testing the suggested skip 'forEachRealityTileTreeRef', and now having a bug when toggling on/off:
If the view is changed (zoomed in/out) while the reality model is invisible (and no other content is part of the view), toggling the reality model to invisible=false again wont make it appear UNTIL the view is changed (i.e. zoomed).

For some reasons, when the reality model is back visible, no tiles get selected... 
`RealityTile.computeVisibilityFactor` returns -1 because `isFrustumCulled` is true.

core/frontend/src/Viewport.ts
```diff
@@ -1231,6 +1231,7 @@ export abstract class Viewport implements IDisposable, TileUser {
     removals.push(settings.contextRealityModels.onPlanarClipMaskChanged.addListener(displayStyleChanged));
     removals.push(settings.contextRealityModels.onAppearanceOverridesChanged.addListener(displayStyleChanged));
     removals.push(settings.contextRealityModels.onDisplaySettingsChanged.addListener(displayStyleChanged));
+    removals.push(settings.contextRealityModels.onInvisibleChanged.addListener(displayStyleChanged));
```
pmconne, line 1234: You need to `invalidateController` (which will invalidate the render plan), because toggling visibility can affect the viewed volume. (re: your comment about frustum culling).

core/common/src/ContextRealityModel.ts
```diff
@@ -320,6 +331,18 @@ export class ContextRealityModel {
...
   }
 
+  /** If true, reality model is not drawn.
+   * @beta
+  */
+  public get invisible(): boolean {
+    return this._invisible;
+  }
+  public set invisible(invisible: boolean) {
+    this.onInvisibleChanged.raiseEvent(invisible, this);
```
pmconne, line 341: Only raise the event if `invisible !== this._invisible`.

core/frontend/src/Viewport.ts
mdastous-bentley: ok, do I also need a changeFlag ?

pmconne: If you mean `this._changeFlags.setDisplayStyle()`, yes.

aruniverse:
If an attached reality model is set to be invisible, and I for example load from a Saved View, does it still make GET reqs to fetch the tiles from the reality data servers?

pmconne:
> If an attached reality model is set to be invisible, and I for example load from a Saved View, does it still make GET reqs to fetch the tiles from the reality data servers?

If we skip processing its tile tree as suggested, no. A unit test can prove this.

core/frontend/src/Viewport.ts
```diff
@@ -1215,6 +1215,11 @@ export abstract class Viewport implements IDisposable, TileUser {
       this._changeFlags.setDisplayStyle();
     };
 
+    const invalidateControllerAndDisplayStyleChanged = () => {
+      this.invalidateController();
+      this._changeFlags.setViewedCategoriesPerModel();
```
pmconne, line 1220:
```suggestion
      this._changeFlags.setDisplayStyle();
```

pmconne (changes requested):
Good PR - thanks.
Please add a note about this change to the Display section of NextVersion.md.

ben-polinsky (approved):

mdastous-bentley:
@pmconne should be good now

mdastous-bentley:
@Mergifyio backport release/4.9.x

pmconne (approved):

PR #7171, opened 2024-09-18
Make `SnapshotConnection.openFile` IPC only

GytisCepk:
## Public API changes

- ~Deprecate `SnapshotConnection.openRemote` API.~
- Make `SnapshotConnection.openFile` only available in IPC apps.

## Motivation

`SnapshotConnection` relies on `SnapshotIModelRpcInterface` for it's communication with the backend. With deprecation and eventual replacement of RPC, all RPC interfaces owned by core need to reviewed and rewritten as needed. While some of the other RPC interfaces (e.g. `IModelReadRpcInterface`) will need to be converted to the new architecture, I believe, `SnapshotIModelRpcInterface` can be moved to IPC as it doesn't make sense to have Web implementation for it.

## Details

`SnapshotConnection` / `SnapshotIModelRpcInterface` has two main methods:
- `openFile(filePath: string, ...)` - opening iModel using file path only makes sense if backend is on the same machine. This is the case in Electron and mobile in which IPC can be used. For that reason new `IpcApp.appFunctionIpc.openSnapshot` was added.
- `openRemote(key: string, ...)` - as far as I can tell, this was intended to be used in Web Viewer. Web Viewer and other iTwin.js web applications should be using `CheckpointConnection.openRemote(...)`. Since GPB doesn't have implementation registered for `SnapshotIModelRpcInterface`, this is not even usable in the web without custom backend. 
  - Deprecation already merged: https://github.com/iTwin/itwinjs-core/pull/7215

## TODO

- [x] Confirm if this is a breaking change or not.
  - ~While changing `SnapshotConnection.openFile` to work with IPC and not RPC could be considered a breaking change as it's excludes web applications from calling it, it's highly unlikely we're going to break anyone since someone needs to have a custom backend set up to use `openFile` in the web application.~ 
  - Some web apps use `SnapshotConnection.openFile` to open iModels for tests. This change will break their tests.
- [x] Find a way to replace `SnapshotConnection.openFile` in _full-stack-tests/core_ as we will want to keep running those tests in Chrome.
  - Added new `TestSnapshotConnection` class which opens `iModelDb` through custom .
- [x] Find a way to replace `SnapshotConnection.openFile` in _full-stack-tests/presentation_ or support IPC there.
  - Added new `TestIModelConnection` class which opens `iModelDb` directly.
- [x] Don't break display-test-app ([usage](https://github.com/iTwin/itwinjs-core/blob/5117e4bab54e9ec9be0a7d46aff873bbff2235ca/test-apps/display-test-app/src/frontend/openIModel.ts#L77)).
  - display-test-app has IPC available in all 3 environments (web, desktop and mobile).
  - display-performance-test-app does not have IPC available in web, so added `TestSnapshotConnection` similar to the one in core full-stack-tests.
- [x] Update _NextVersion.md_.
- [x] Review and update documentation.


pmconne:
Please add to your TODO list: Don't break display-test-app. See usage [here](https://github.com/iTwin/itwinjs-core/blob/5117e4bab54e9ec9be0a7d46aff873bbff2235ca/test-apps/display-test-app/src/frontend/openIModel.ts#L77).

core/common/src/rpc/SnapshotIModelRpcInterface.ts
```diff
@@ -37,11 +40,20 @@ export abstract class SnapshotIModelRpcInterface extends RpcInterface { // eslin
...
   ===========================================================================================*/
 
+  /**
+   * @deprecated in 4.10. Use [[IpcAppFunctions.openSnapshot]] in IPC applications, no replacement for Web applications.
+   */
   @RpcOperation.setRoutingProps(unknownIModelId)
   public async openFile(_filePath: string, _opts?: SnapshotOpenOptions): Promise<IModelConnectionProps> { return this.forward(arguments); }
 
+  /**
+   * @deprecated in 4.10. Use [[ConnectionConnection.openRemote]].
```
hl662, line 50: Do you mean CheckpointConnection?

core/backend/src/IModelHost.ts
```diff
@@ -487,7 +487,7 @@ export class IModelHost {
     [
       IModelReadRpcImpl,
       IModelTileRpcImpl,
-      SnapshotIModelRpcImpl,
+      SnapshotIModelRpcImpl, // eslint-disable-line deprecation/deprecation
       WipRpcImpl,
```
aruniverse, line 491: Didnt realize we actually used this WipRpc...
not in this PR's context but we should review its usage and look at killing it off

core/backend/src/IpcHost.ts
```diff
@@ -224,6 +224,15 @@ class IpcAppHandler extends IpcHandler implements IpcAppFunctions {
   public async openStandalone(filePath: string, openMode: OpenMode, opts?: StandaloneOpenOptions): Promise<IModelConnectionProps> {
     return StandaloneDb.openFile(filePath, openMode, opts).getConnectionProps();
   }
+  public async openSnapshot(filePath: string, opts?: SnapshotOpenOptions | undefined): Promise<IModelConnectionProps> {
+    let resolvedFileName: string | undefined = filePath;
+    if (IModelHost.snapshotFileNameResolver) {
```
aruniverse, line 229: i dont think anyone is actually using this fileNameResolver, id like to try and not carry fwd junk with us

core/backend/src/IModelHost.ts
GytisCepk: `WipRpcInterface` has no usage in core and shouldn't be used outside core as it's internal.

core/backend/src/IpcHost.ts
GytisCepk: `IModelHost.snapshotFileNameResolver` is public and, if we start ignoring it, it would be a breaking change.

We could deprecate it now and stop using it in 5.0.

docs/changehistory/NextVersion.md
```diff
@@ -71,3 +74,9 @@ A new optional [`extendedData`]($docs/presentation/content/CalculatedPropertiesS
 - `LayoutFragmentProps`, `ContentLayoutProps`, `LayoutSplitPropsBase`, `LayoutHorizontalSplitProps`, `LayoutVerticalSplitProps`, and `StandardContentLayouts` have been deprecated. Use the same APIs from `@itwin/appui-react` instead.
 
 - `BackendItemsManager` is internal and should never have been consumed. It has been deprecated and will be removed in 5.0.0. Use `UiFramework.backstage` from `@itwin/appui-react` instead.
+
+## Deprecations
+
+### @itwin/core-frontend
+
+- [SnapshotConnection.openRemote]($frontend) is deprecated. To open iModel for web workflows use [CheckpointConnection.openRemote]($frontend) instead.
```
grigasp, line 82: This should be merged with "API deprecations section". Table of contents needs to be fixed as well.

grigasp (approved):
approving presentation changes

core/backend/src/IpcHost.ts
```diff
@@ -224,6 +224,15 @@ class IpcAppHandler extends IpcHandler implements IpcAppFunctions {
   public async openStandalone(filePath: string, openMode: OpenMode, opts?: StandaloneOpenOptions): Promise<IModelConnectionProps> {
     return StandaloneDb.openFile(filePath, openMode, opts).getConnectionProps();
   }
+  public async openSnapshot(filePath: string, opts?: SnapshotOpenOptions | undefined): Promise<IModelConnectionProps> {
```
ben-polinsky, line 227: Why is `SnapshotOpenOptions` optional and also a union with `undefined`? Looks like `SnapshotDb.openFile` is just optional.

GytisCepk: No reason, I'll remove `undefined`.

In the end, `opts` type will still resolve to same type: `SnapshotOpenOptions | undefined`

full-stack-tests/core/src/backend/RpcImpl.ts
```diff
@@ -23,11 +23,25 @@ async function initializeContainer(containerId: string) {
   await ViewStore.CloudAccess.initializeDb({ baseUri: AzuriteTest.baseUri, storageType, containerId, accessToken });
 }
 
-export class TestRpcImpl extends RpcInterface implements TestRpcInterface { // eslint-disable-line deprecation/deprecation
+export class TestRpcImpl extends RpcInterface implements TestRpcInterface {
   public static register() {
     RpcManager.registerImpl(TestRpcInterface, TestRpcImpl);
   }
 
+  public async openSnapshot(filePath: string): Promise<IModelConnectionProps> {
```
tcobbs-bentley, line 31: This function appears to be a copy of `IpcHost.openSnapshot`, but without the optional open options. Since it doesn't appear that this function or `IpcHost.openSnapshot` makes use of their `this` pointers, it seems like having a public static function in `IpcHost` that both non-static openSnapshot functions call would be better, to avoid code copying. (Maybe I'm misreading these two functions.)

GytisCepk: I would like to avoid increasing `IpcHost` API size just to avoid some code duplication in tests.

Additionally, `IModelHost.snapshotFileNameResolver` may be deprecated soon. If it is deprecated, this function would be the first to be refactored while `IpcHost.openSnapshot` would need to keep supporting it until deprecated property is removed.

aruniverse (approved):
lgtm;

@markschlosseratbentley please review changes to DTA

markschlosseratbentley:
> lgtm;
> 
> @markschlosseratbentley please review changes to DTA

@MarcNeely FYI

pmconne:
> I did some very basic testing (successfully opened an iModel in a DTA built off this branch) and it works.

Did you confirm you can [edit an iModel](https://github.com/iTwin/itwinjs-core/blob/master/test-apps/display-test-app/README.md#editing)?

markschlosseratbentley:
> > I did some very basic testing (successfully opened an iModel in a DTA built off this branch) and it works.
> 
> Did you confirm you can [edit an iModel](https://github.com/iTwin/itwinjs-core/blob/master/test-apps/display-test-app/README.md#editing)?

I did not confirm this.

GytisCepk:
> Make `SnapshotConnection.openFile` only available in IPC apps.

This will break integration tests of at least 1 application. I don't want to break anything on minor release, I'm moving this PR back to draft state. I'll reopen it once we're working on 5.0. 

Separate PR will be opened soon to deprecate `SnapshotConnection.openRemote`.

pmconne (approved):

PR #7187, opened 2024-09-23
pass pullMerge to getIModelProps which causes extents to be loaded from the database instead of the cached value

nick4598:
imodel-native: https://github.com/iTwin/imodel-native/pull/866

Fixes https://github.com/iTwin/itwinjs-backlog/issues/1240


core/backend/src/IModelDb.ts
```diff
@@ -305,6 +305,13 @@ export abstract class IModelDb extends IModel {
     GeoCoordConfig.loadForImodel(this.workspace.settings); // load gcs data specified by iModel's settings dictionaries, must be done before calling initializeIModelDb
 
     this.initializeIModelDb();
+    this.onChangesetApplied.addListener(() => {
+      const extents = this.queryFilePropertyString({ name: "Extents", namespace: "dgn_Db" });
+      // TODO: What if extents is undefined? And was previously defined? Is this a possibility?
```
nick4598, line 310: Any thoughts on this? My guess is that it is unlikely for an imodel to go from having project extents to no project extents so it is not worth handling, but not sure.

aruniverse: I dont think its possible to from defined extents to undefined, the other way around yes

core/backend/src/IModelDb.ts
```diff
@@ -305,6 +305,13 @@ export abstract class IModelDb extends IModel {
     GeoCoordConfig.loadForImodel(this.workspace.settings); // load gcs data specified by iModel's settings dictionaries, must be done before calling initializeIModelDb
 
     this.initializeIModelDb();
```
khanaffan, line 307: `initializeIModelDb()` is called after pullMerge() after applying all change set. It internally calls following. If I understand correctly, you want to notify native side of any change to project extent or ecf location. If this is the case, you should not listen to apply change set event, rather just do that in initializeIModelDb(). You may want to add parameter to `initializeIModelDb(when: "pullMerge" | undefined)`

```ts
  protected initialize(name: string, props: IModelProps) {
    this.name = name;
    this.rootSubject = props.rootSubject;
    this.projectExtents = Range3d.fromJSON(props.projectExtents);
    this.globalOrigin = Point3d.fromJSON(props.globalOrigin);

    const ecefLocation = props.ecefLocation ? new EcefLocation(props.ecefLocation) : undefined;
    this.ecefLocation = ecefLocation?.isValid ? ecefLocation : undefined;

    this.geographicCoordinateSystem = props.geographicCoordinateSystem ? new GeographicCRS(props.geographicCoordinateSystem) : undefined;
  }
```

nick4598: Switched to passing pullMerge into intializeIModelDb

common/api/core-backend.api.md
```diff
@@ -3116,7 +3116,7 @@ export abstract class IModelDb extends IModel {
     // @alpha
     importSchemaStrings(serializedXmlSchemas: string[]): Promise<void>;
     // @internal (undocumented)
-    protected initializeIModelDb(): void;
+    protected initializeIModelDb(when?: "pullMerge"): void;
```
aruniverse, line 3119: are we really just going to use a string literal?
what other possible option is there for this in the future?

nick4598: I also personally wanted to do a boolean "afterPullMerge" or something along those lines that defaults to false because it seems unlikely we'd need to special case anything else. But, affan mentioned that he was working on a Pr to support table conflict handlers which are called before applying changests, after applying changesets, and eventually after calling pullMerge. That was where his suggestion came from. Im not tied to either approach.

aruniverse: @khanaffan

toddsouthenbentley:
@nick4598 Any updates on this?

nick4598:
> @nick4598 Any updates on this?

This PR is ready for another round of review

common/api/core-backend.api.md
khanaffan: Look good

khanaffan (approved):

aruniverse (approved):

aruniverse:

PR #7195, opened 2024-09-25
Change bearing/azimuth values from east ccw to north cw

rschili:
During the first implementation of this, it was agreed we would store bearing/azimuth values as angles from east and counter-clockwise.

After some discussions it was now decided to change that behavior to north clockwise.

Changes reflect that decision.


common/changes/@itwin/core-quantity/rob-bearingNorthDefault_2024-09-25-10-42.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-quantity",
+      "comment": "Change default azimuth/bearing base from east to north",
```
aruniverse, line 5: can we have a callout for this in nextver.md?

rschili: done

ColinKerr (approved):

PR #7199, opened 2024-09-26
Fix link in NextVersion.md

saskliutas:
Fixed invalid link in NextVersion.md

@DanRod1999 @aruniverse Do you have any idea why Docs pipeline does not fail on invalid link? You can check latest run on master. It's successful but warning is logged. In Presentation repo validate docs job fails for that invalid link.


aruniverse:
@ben-polinsky any idea?

ben-polinsky:
We do not error for missing links in changehistory files because old changelogs may have now-missing APIs. We should check `NextVersion.md`. As to why presentation is erroring, I'm not sure. Will look into it when I get a chance.

ben-polinsky (approved):

PR #7204, opened 2024-09-27
Require db for OpenGroup and use it to resolve schema changes

mattbjordan:
Opening changesets as a group requires the context of an iModel db to resolve any schema changes that may have happened in those changesets.

Passing in `db` is now required.

imodel-native: https://github.com/iTwin/imodel-native/pull/872


common/api/core-backend.api.md
```diff
@@ -5175,7 +5175,7 @@ export class SqliteChangesetReader implements IDisposable {
     } & SqliteChangesetReaderArgs): SqliteChangesetReader;
     static openGroup(args: {
         readonly changesetFiles: string[];
-    } & SqliteChangesetReaderArgs): SqliteChangesetReader;
+    } & MarkRequired<SqliteChangesetReaderArgs, "db">): SqliteChangesetReader;
```
mattbjordan, line 5178: Do I need to deprecate the old version of this to make this change? @iTwin/itwinjs-core-admins

ben-polinsky: Technically no since the class is `@beta` but for something like this, I'd defer to someone who knows this particular code better. (Was this a mistake? Is there a valid case for not passing a db in the args?)

aruniverse: @nick4598

nick4598: This is a brand new API I believe (around a month old).

From [SQLite Docs on changegroups](https://www.sqlite.org/session/sqlite3changegroup_add.html):

>If the new changeset contains changes to a table that is already present in the changegroup, then the number of columns and the position of the primary key columns for the table must be consistent. If this is not the case, this function fails with SQLITE_SCHEMA. Except, if the changegroup object has been configured with a database schema using the sqlite3changegroup_schema() API, then it is possible to combine changesets with different numbers of columns for a single table, provided that they are otherwise compatible.
 
 It is possible to use the changegroup API with or without a database provided, but I don't know how practical it would be to not provide a database in the context of iTwin.js. We should expect in pretty much every case that we atleast have one to pass... but which version of the database should be passed? 
 
 The way the changegroup API resolves the differences requires that there exists a database table with:
- the table name identified by the changeset, 
- at least as many columns as recorded in the changeset, and
- the primary key columns in the same position as recorded in the changeset.

Since we want the database table to have more columns than our changeset, I assume we'd want to pass the database which has the schema already applied to it, just incase that schema added a new column. (What if a schema removes a column? Does this happen? I don't know)
 
 @khanaffan might be worried about how the changegroup API currently resolves the differences? A changeset table with fewer columns than the database schema would just have the default value used for the missing columns. I don't know if that would be a problem or not.

khanaffan: So  `SqliteChangesetReader` does not require db context.  If found it useful in debugging some changeset where all i wanted to is to e.g. print out table/pk/op.  Grouping require a db context due to schema changes. And ECChangeReader definitely require db context.

We can simplify this by making db a required argument. The debug case is not really important. But there is still no control on what version the db is on. May be added docs there that say it should be a head of change set been open.

Also, openLoalChanges() also take iModel from which it open local changes. Maybe we would not need the iModel arg if we make db arg non-optional

`public static openLocalChanges(args: { iModel: IModelJsNative.DgnDb, includeInMemoryChanges?: true } & SqliteChangesetReaderArgs): SqliteChangesetReader`


The api is beta so we can make this change.

khanaffan: @nick4598 we do not really have default values all new columns are default to null at least in ecdb/imodel

common/api/core-common.api.md
```diff
@@ -11210,7 +11210,7 @@ export class WhiteOnWhiteReversalSettings {
     toJSON(): WhiteOnWhiteReversalProps | undefined;
 }
 
-// @internal @deprecated
+// @internal
 export abstract class WipRpcInterface extends RpcInterface {
```
aruniverse, line 11214: revert change here

common/api/core-backend.api.md
mattbjordan: I've made `db` required in `SqliteChangesetReaderArgs` and never undefined in `SqliteChangesetReader`.

openLocalChanges() only works with an `IModelDb` database, not an `ECDb` database (even looking down into the C++ code, that's what is required). So only for that function, I decided to keep the `args` parameter since all of the args are used, but I forced the type of `SqliteChangesetReaderArgs.db` to be `IModelDb`.

tcobbs-bentley:
The only mobile changes appear to be updating the mobile add-on version for the iOS and Android versions of display-test-app. I'm fine with those changes, but don't feel I have enough knowledge to comment on the main changes in this PR.

core/backend/src/SqliteChangesetReader.ts
```diff
@@ -132,12 +132,13 @@ export class SqliteChangesetReader implements IDisposable {
   /**
    * Open local changes in iModel.
    * @param args iModel and other options.
+   * @param args.db must be of type IModelDb
    * @returns SqliteChangesetReader instance
    */
-  public static openLocalChanges(args: { iModel: IModelJsNative.DgnDb, includeInMemoryChanges?: true } & SqliteChangesetReaderArgs): SqliteChangesetReader {
+  public static openLocalChanges(args: Omit<SqliteChangesetReaderArgs, "db"> & { db: IModelDb, includeInMemoryChanges?: true }): SqliteChangesetReader {
```
aruniverse, line 138: does the omit still make sense now? since you made db required in the SqliteChangesetReaderArgs?

common/changes/@itwin/core-backend/mattbjordan-pass-db-to-opengroup_2024-09-27-16-31.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Require db for openGroup",
```
aruniverse, line 5:
```suggestion
      "comment": "Require db in `SqliteChangesetReaderArgs`",
```

core/backend/src/SqliteChangesetReader.ts
mattbjordan: The Omit is just to change the type of `db` when passing SqliteChangesetReaderArgs into this function. The `db` argument is added back in `db: IModelDb`. It looks messy in the function definition, but it just resolves to `SqliteChangesetReaderArgs.db` being of type `IModelDb` here.

It has to be of type `IModelDb` because only that works here. The Omit allows me to override the default value of `db` being `IModelDb | ECDb`.

aruniverse: All youre doing here is narrowing the type from `AnyDB` (`IModelDb | ECDb`) to `IModelDB`. Where do we use `ECDb` in here now?

mattbjordan: Yes. Much simpler way of putting it.

`openGroup()` allows for either database type. And then anywhere else that wants to use a `SqliteChangesetReader` can allow for either Db type to be passed in. In particular, `ChangesetECAdaptor` uses a `SqliteChangesetReader` for reading changesets.

aruniverse (approved):

pmconne:
@mattbjordan @aruniverse multiple imodel-native PRs are failing.
```
--[ FAILURE: @itwin/core-backend ]---------------------------[ 3.78 seconds ]--

src/SqliteChangesetReader.ts(118,26): error TS2554: Expected 3 arguments, but got 2.
```

[This one](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3262784&view=logs&j=e538c5b1-ca02-55cf-272d-45e0b52dd4fb&t=780476b4-e1c4-5c11-6e4c-3f715c15818c) and [this one](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3262088&view=logs&jobId=217fbd95-5c2e-5a16-52a7-4136f1b85287&j=217fbd95-5c2e-5a16-52a7-4136f1b85287&t=0375ca9c-605b-5f0a-2b86-c33701b6bc11), for example.

aruniverse:
> @mattbjordan @aruniverse multiple imodel-native PRs are failing.
> 
> ```
> --[ FAILURE: @itwin/core-backend ]---------------------------[ 3.78 seconds ]--
> 
> src/SqliteChangesetReader.ts(118,26): error TS2554: Expected 3 arguments, but got 2.
> ```
> 
> [This one](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3262784&view=logs&j=e538c5b1-ca02-55cf-272d-45e0b52dd4fb&t=780476b4-e1c4-5c11-6e4c-3f715c15818c) and [this one](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3262088&view=logs&jobId=217fbd95-5c2e-5a16-52a7-4136f1b85287&j=217fbd95-5c2e-5a16-52a7-4136f1b85287&t=0375ca9c-605b-5f0a-2b86-c33701b6bc11), for example. @khanaffan @mathieu-lemuzic fyi.

I believe the native pipelines were failing because the linked js pr didnt have the recent changes from master. I just updated Affans PR, and I would expect this [native pipeline run](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3263906&view=results) will pass now

mattbjordan:
> > @mattbjordan @aruniverse multiple imodel-native PRs are failing.
> > ```
> > --[ FAILURE: @itwin/core-backend ]---------------------------[ 3.78 seconds ]--
> > 
> > src/SqliteChangesetReader.ts(118,26): error TS2554: Expected 3 arguments, but got 2.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [This one](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3262784&view=logs&j=e538c5b1-ca02-55cf-272d-45e0b52dd4fb&t=780476b4-e1c4-5c11-6e4c-3f715c15818c) and [this one](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3262088&view=logs&jobId=217fbd95-5c2e-5a16-52a7-4136f1b85287&j=217fbd95-5c2e-5a16-52a7-4136f1b85287&t=0375ca9c-605b-5f0a-2b86-c33701b6bc11), for example. @khanaffan @mathieu-lemuzic fyi.
> 
> I believe the native pipelines were failing because the linked js pr didnt have the recent changes from master. I just updated Affans PR, and I would expect this [native pipeline run](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3263906&view=results) will pass now

Seems like all is good so long as the builds have the latest changes from main. I see that [this build](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3263850&view=results) succeeded after merging in main [on this PR](https://github.com/iTwin/imodel-native/pull/874/commits/bc46d55877069bc3f0cc4f4d97e519b369144063).

PR #7205, opened 2024-09-27
Contour line display

markschlosseratbentley:
This PR adds support to iTwin.js for contour display of specified subcategories within the iModel.

TODO (before this PR merges):
- [x] Support "just lines" display (show just lines, not the underlying geometry)

TODO (in near future):
- [x] Bug: At certain LODs, certain elements within certain tiles can display the incorrect (or undefined) contour styling
- [x] Maybe not Requirement: Apply contour lines default (Apply to all other geometry)
- [ ] Snapping
- [x] Ability to enable transparency in the underlying mesh, but not in the contour lines

Example rendering:
contour-display


core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,263 @@
...
+      return props;
+    }
+  }
+
+  /** JSON representation of the description of how contours should appear for a particular set of subcategories. */
+  export interface TerrainProps {
+    /** See [[Terrain.contourDef]]. */
+    contourDef?: ContourProps;
+    /** See [[Terrain.subCategories]]. */
+    subCategories?: Id64String[];
```
pmconne, line 160: Use `CompressedId64Set`.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  /** Contains a description of how contours should appear for a particular set of subcategories. A terrain is an organizational concept which associates a contour appearance with a list of subcategories within an iModel.
+   * @see [[ContourDisplay.Contour]]
+   */
+  export class Terrain {
+    /** A [[ContourDisplay.Contour]] object describing how the contours for this terrain should appear. Defaults to an instantation of [[ContourDisplay.Contour]] using all of its own default properties. */
+    public readonly contourDef: Contour;
+    /** List of subcategory IDs to which this terrain styling will be applied. */
+    public readonly subCategories: Id64String[];
```
pmconne, line 170: Expose underlying `CompressedId64Set` as an `OrderedId64Iterable`.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  /** The rendering styling settings that apply to a specific set of subcategories within a [[ContourDisplay.Terrain]].
+   * This actually describes stylings for two sets of contours: major and minor. These stylings are separate from each other.
+   * The minor contour occurs at a defined interval in meters. These intervals draw at a fixed height; they are not dependent on the range of the geometry to which they are applied.
+   * The major contour is dependent on the minor contour. The interval of its occurence is not measured directly in meters; rather it is a count of minor contour intervals between its occurrences.
+   */
+  export class Contour {
+    /** Color that a major contour line will use. Defaults to [[ColorDef.black]].*/
+    public readonly majorColor: ColorDef;
```
pmconne, line 45: Did you consider having a type `{ color, width, pattern }` and two properties `major` and `minor` each of that type?

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,263 @@
...
+
+import { compareNumbers, Id64String } from "@itwin/core-bentley";
+import { ColorDef, ColorDefProps } from "./ColorDef";
+import { LinePixels } from "./LinePixels";
+
+/** Namespace containing types controlling how contour lines are rendered.
+ * @public
+ * @extensions
+ */
+export namespace ContourDisplay {
```
pmconne, line 17: Nitpick but this namespace doesn't do anything except make people have to read/type "Contour" a bunch of extra times.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,263 @@
...
+
+      if (5 !== this.majorIntervalCount)
+        props.majorIntervalCount = this.majorIntervalCount;
+
+      return props;
+    }
+  }
+
+  /** JSON representation of the description of how contours should appear for a particular set of subcategories. */
+  export interface TerrainProps {
```
pmconne, line 156: Contour lines are applicable to more than just terrain, and confusingly do not currently apply to the actual global terrain, which happens also to be defined by a type called `TerrainProps`. Consider something more generic and descriptive like `ContourGroup`.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,263 @@
...
+  /** JSON representation of the contour display setup of a [[DisplayStyle3d]]. */
+  export interface SettingsProps {
+    /** See [[ContourDisplay.terrains]]. */
+    terrains?: (TerrainProps | undefined)[];
+  }
+
+  /** The contour display setup of a [[DisplayStyle3d]].
+   * Contour display allows a user to apply specific contour line renderings to subcategories within a scene.
+   */
+  export class Settings {
```
pmconne, line 225: Consider adding a `ViewFlags.contourLines` so that 1. people can toggle them on and off without destroying and recreating the contour settings and 2. people can create a `GraphicBranch` to contain graphics that shouldn't have contour lines applied to them. See `ViewFlags.thematicDisplay` for comparison.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,263 @@
...
+    /** A list of the terrains which contain their own specific contour display settings. Defaults to an empty array. */
+    public readonly terrains: (Terrain | undefined)[] = [];
```
pmconne, line 227: Why does it explicitly permit a sparse array?

DStradley: So that the contour definitions can be used independently (and cleared, by having no subcategories) without them moving around.

core/common/src/ContourDisplay.ts
DStradley: One problem with that is that there is not a major/minor interval per se, those are 2 different things (count and interval).

core/frontend/src/render/webgl/ContourUniforms.ts
```diff
@@ -0,0 +1,135 @@
...
+/** @packageDocumentation
+ * @module WebGL
+ */
+
+import { ColorDef, ContourDisplay, LinePixels } from "@itwin/core-common";
+import { UniformHandle } from "./UniformHandle";
+import { desync, sync } from "./Sync";
+import { Target } from "./Target";
+
+function convertLinePixelsToIndexInShader(linePixels: LinePixels): number {
```
pmconne, line 14: The only difference from `lineCodeFromLinePixels` appears to be that it treats an invalid input as "invisible" instead of "solid". What's the reason for that? Under what circumstances do you expect an input that is not a valid `LinePixels` value?

core/frontend/src/render/webgl/Contours.ts
```diff
@@ -0,0 +1,167 @@
...
+import { GL } from "./GL";
+import { UniformHandle } from "./UniformHandle";
+import { TextureUnit } from "./RenderFlags";
+import { System } from "./System";
+import { Target } from "./Target";
+import { Texture2DDataUpdater, Texture2DHandle, TextureHandle } from "./Texture";
+import { BatchOptions } from "../../common/render/BatchOptions";
+import { ContourUniforms } from "./ContourUniforms";
+
+function computeWidthAndHeight(nEntries: number, nRgbaPerEntry: number): { width: number, height: number } {
```
pmconne, line 21: Why isn't this just `computeDimensions(nEntries, nRgbaPerEntry, 0, System.instance.maxTextureSize)`?

docs/changehistory/NextVersion.md
```diff
@@ -47,9 +47,86 @@ For the stop sign example described above, you might have a [glTF model](https:/
...
+  - `majorPixelWidth` is the width in pixels of a major contour line. (Range 1.5 to 9 in 0.5 increments). Defaults to 2.
+  - `minorPixelWidth` is the width in pixels of a minor contour line. (Range 1.5 to 9 in 0.5 increments). Defaults to 1.
+  - `majorPattern` is the pattern for a major contour line. Defaults to [LinePixels.Solid]($common).
+  - `minorPattern` is the pattern for a minor contour line. Defaults to [LinePixels.Solid]($common).
+  - `minorInterval` is the interval for the minor contour in the associated terrain in meters. Defaults to 1.
+  - `majorIntervalCount` is the count of minor contour intervals that define a major interval (integer > 0). Defaults to 5.
+
+Consult the following code for an example of enabling and configuring contour display in iTwin.js:
+
+```ts
```
pmconne, line 76: Before merging PR please move example code to the example-code-snippets package and reference it from here, to ensure it compiles now and in the future.

core/frontend/src/render/webgl/Contours.ts
DStradley: I was following pattern in FeatureOverrides.ts.  Looking now, I see that FeatureOverrides could do what you are suggesting.
However, I remember I had to change the code here to deal with the fractional nRgbaPerEntry.  I think it ended up just being the added ceil on line 23.
Also, neither this nor FeatureOverrides have the extra test and bump height at the end, but I am not sure how that could happen?

DStradley: Oh, I also changed line 25 from < to <=.

DStradley: Will it work to incorporate those two changes into computeDimensions and then replace this function and the one in FeatureOverrides with computeDimensions call?

core/common/src/ContourDisplay.ts
markschlosseratbentley: `ContourStyleProps` and `ContourStyle` groupings for these properties have been written in latest code push.

core/common/src/ContourDisplay.ts
markschlosseratbentley: This has been done in the latest code push.

core/common/src/ContourDisplay.ts
markschlosseratbentley: This has been done in the latest code push.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,309 @@
...
+  public equals(other: ContourGroup | undefined): boolean {
+    if (this === undefined && other === undefined)
+      return true;
+    if (this === undefined || other === undefined)
+      return false;
+    if (!this.contourDef.equals(other.contourDef))
+      return false;
+    let subCount = 0;
+    let otherSubCount = 0;
+    for (const subCategory of this.subCategories) {
```
markschlosseratbentley, line 222: @pmconne This equality comparison nested loop is not ideal compared to the way it used to be before the latest API changes I pushed. Is it safe for us to just compare the underlying `CompressedId64Set` as a string against the other one?

pmconne: Yes, that's what I meant by making comparison trivial. Two CompressedId64Sets contain the same Ids if and only if they are identical strings.

markschlosseratbentley: Ok, thanks for verifying. The enforced ordering makes a lot of sense in that context. I'll change that comparison logic and resolve this comment when I push it.

markschlosseratbentley: I fixed the subcategory comparison.

core/frontend/src/render/webgl/ContourUniforms.ts
markschlosseratbentley: I neglected to find that existing function when I wrote this function. I now changed the code to use the existing function (through `LineCode.valueFromLinePixels`). Treating invalid input as solid rather than invisible is fine, and probably preferable.

core/common/src/ContourDisplay.ts
pmconne: Was that done to make it easier to implement the UI in display-test-app? It seems harder to use than a `ReadonlyArray<Terrain>`.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,296 @@
...
+  pattern?: LinePixels;
+}
+
+/** The style settings used by either a minor or major contour.
+   * @see [[Contour.majorStyle]]
+   * @see [[Contour.minorStyle]]
+   */
+export class ContourStyle {
+  /** Color that a major or minor contour line will use. Defaults to [[ColorDef.black]].*/
+  public readonly color: ColorDef;
```
pmconne, line 26: What if anything is the color's transparency used for?

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,296 @@
...
+  color?: ColorDefProps;
+  pixelWidth?: number;
+  pattern?: LinePixels;
+}
+
+/** The style settings used by either a minor or major contour.
+   * @see [[Contour.majorStyle]]
+   * @see [[Contour.minorStyle]]
+   */
+export class ContourStyle {
```
pmconne, line 24: Add `clone` methods to the various immutable types in this file following the pattern used elsewhere, accepting a `Partial<WhateverProps>`, so people can create modified copies.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,296 @@
...
+    return props;
+  }
+}
+
+/** JSON representation of the description of how contours should appear for a particular set of subcategories. */
+export interface ContourGroupProps {
+  /** See [[ContourGroup.contourDef]]. */
+  contourDef?: ContourProps;
+  /** See [[ContourGroup.subCategories]]. */
+  subCategories?: CompressedId64Set;
```
pmconne, line 196: These Ids will be stored in the JSON properties of DisplayStyle elements. You'll need to update `DisplayStyle.collectReferenceIds` and `DisplayStyle.onCloned` to correctly disclose and remap them.

core/common/src/ContourDisplay.ts
markschlosseratbentley: @DStradley the transparency is ignored -- right?

core/common/src/ContourDisplay.ts
markschlosseratbentley: I added a `displayContours` toggle on the display settings.

pmconne: In that case you will need to raise another event when that flag changes.
Seems simpler to put the flag in the ContourDisplay object.

markschlosseratbentley: That would mean the ContourDisplay object will need to destroy and recreate itself in order to change the flag, which this change was hoping to avoid.  That is, if we make the flag living on ContourDisplay follow the pattern it currently does --- exposing access to everything only through fromJSON() and toJSON().

Are you proposing the flag on the ContourDisplay object be accessible without going through those methods?

pmconne: ContourDisplay and the objects it contains are all immutable, there is no need to destroy any of them. Look at `Environment.withDisplay` for the equivalent.

core/common/src/ContourDisplay.ts
pmconne: Did you consider using `RgbColor` instead to remove ambiguity?

core/common/src/ContourDisplay.ts
DStradley: It was done mainly to make it easier to use (not implement) the UI, and it seemed to me likely the same would be true for Civil.  Currently we don't have a name for a particular contour definition group, just a number, so when you want to change something you select that number and change the definition (color, whatever).  You quickly remember what number goes with what group, so if that was changing whenever you deleted a lower one (or whatever else might rearrange them), that would get very confusing to use.  (To me, identifying a group by remembering a list of subCategoryIds is very unlikely.)

core/common/src/ContourDisplay.ts
DStradley: I'll refer that to @markschlosseratbentley.

core/common/src/ContourDisplay.ts
pmconne: Probably should ask Civil for examples of how their UI for this stuff works in their existing app.

core/common/src/ContourDisplay.ts
markschlosseratbentley: Yes, I agree. We should make it use `RgbColor` to remove ambiguity.

core/frontend/src/render/webgl/Contours.ts
DStradley: Okay, I went ahead and modified computeDimensions to add the ceil, changed < to <=, and removed the 2 redundant lines at the end.  Then I changed both FeatureOverrides.ts and Contours.ts to call computeDimensions instead of having their own functions.  I ran an ImageTests run with this and did not see any issues.

core/common/src/ContourDisplay.ts
markschlosseratbentley: Based on a side conversation, it was proposed by @pmconne to add a "name" property to each ContourGroup which identifies them. That way they can be kept track of without the need for a sparse array.

DStradley: So will we allow creation of a ContourGroup with no / empty subCategories now?

pmconne: I think there are two separate questions.
1. Is empty subcategory list allowed? If not, you'd have to throw an exception or silently ignore when people try to create one. I think it's better to accept it.
2. What does empty subcategory list signify? Does it mean no subcategories will have this contour style applied, or does it mean this is the default style that applies to subcategories not included in other contour groups? If the latter, and we decide later to add support for a default contour group, we'll need a different way to indicate that. But you would be able to punt for now on adding support for default group.

I vote for whichever is more expedient. Is default group difficult or time-consuming?

pmconne: For reference, [ModelClipGroup](https://github.com/iTwin/itwinjs-core/blob/8ab7ad738c5127c3bdc874ba4019b9e3c54d0171/core/common/src/ModelClipGroup.ts#L28) interprets an empty list of models as indicating the clip should apply to all otherwise-ungrouped models.

DStradley: What is the difference in meaning between "supporting a default group" and "supporting using a ContourGroup with an empty subcategory list as a default style for subcategories not included in other contour groups"?

I guess I am unsure of the definition of the first.  (The latter I think would not be difficult, but we probably also would want a switch to enable/disable, and if there is more than one defined this way, we just use one.)

pmconne: > What is the difference in meaning between "supporting a default group" and "supporting using a ContourGroup with an empty subcategory list as a default style for subcategories not included in other contour groups"?

The latter is one way of implementing the former. I'm unclear what you mean by "a switch to enable/disable" - are you talking about the UI in display-test-app?

DStradley: I was thinking of some flag in the API (and also the UI I guess) to disable having the default definition applied, kind of like the switch to display contours or not in general.

pmconne: You can turn off the default group by removing the default group, just like any other group.

DStradley: Okay, I just pushed an implementation that uses a definition with an empty subcategory list as the default for any features that don't match any of the subCategoryIds. (If there is more than one definition it uses the last, if there are none, then default is not to draw contours.)

core/common/src/ContourDisplay.ts
markschlosseratbentley: As part of attempting to resolve [this comment chain](https://github.com/iTwin/itwinjs-core/pull/7205#discussion_r1792386401), I began working on adding this to all the other ContourDisplay structures --- namely ContourGroup, Contour, and ContourStyle.

However, I ran into difficulty with ContourGroup and allowing it to use the Partial paradigm, particularly the way subCategories are exposed as an OrderedId64Iterable but stored internally as a CompressedId64Set.

Do you have any suggestions?

Otherwise, we could just skip this for now because at least ContourDisplay now supports creating itself with partial properties, namely the displayContours flag.

core/common/src/ContourDisplay.ts
markschlosseratbentley: I changed this to use `RgbColor`.

core/common/src/ContourDisplay.ts
markschlosseratbentley: Code has been pushed that makes this array of contour groupings no longer a sparse array.
Test UI functions properly with this new paradigm.

docs/changehistory/NextVersion.md
markschlosseratbentley: Can you please provide an existing example of the iTwin-preferred style for how to reference a snippet from a NextVersion.md / numbered release markdown file?

markschlosseratbentley: Snippet added. Linked within NextVersion.md. Reworked NextVersion.md to fit new API.

core/common/src/ContourDisplay.ts
markschlosseratbentley: I pushed some code for this. I am not too familiar with this, so if you could let me know if it looks good I'd appreciate it.

docs/changehistory/NextVersion.md
```diff
@@ -57,6 +57,36 @@ For the stop sign example described above, you might have a [glTF model](https:/
 
 Context reality models that have been attached using `DisplayStyleState.attachRealityModel`, can now be hidden by turning ON the `ContextRealityModel.invisible` flag.  Previous implementation requiered context reality models to be detached in order to hide it from the scene.
 
+### Contour Display
```
pmconne, line 60: Add to table of contents.

docs/changehistory/NextVersion.md
```diff
@@ -57,6 +57,36 @@ For the stop sign example described above, you might have a [glTF model](https:/
...
+- [ContourGroup]($common) describes an association of subcategories to contour styling. It contains a set of subcategory IDs titled `subCategories`. Those subcategories will have the contour styling within the same group's [Contour]($common) `contourDef` object applied to them.
+- [Contour]($common) describes the rendering settings that apply to a specific set of subcategories within a [ContourGroup]($common). This actually describes stylings for two sets of contours: major and minor. These stylings are separate from each other. The minor contour occurs at a defined interval in meters. These intervals draw at a fixed height; they are not dependent on the range of the geometry to which they are applied. The major contour is dependent on the minor contour. The interval of its occurence is not measured directly in meters; rather it is a count of minor contour intervals between its occurrences. The properties describing how major and minor contours are styled are listed here:
```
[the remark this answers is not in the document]
markschlosseratbentley: Sounds good. I can copy it to somewhere like docs/learning/ContourDisplay.md.

docs/changehistory/NextVersion.md
markschlosseratbentley: I will also add "Context Reality model visibility" which is not in the TOC.

markschlosseratbentley: Updated.

docs/changehistory/NextVersion.md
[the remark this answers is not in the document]
markschlosseratbentley: Added.

core/common/src/ContourDisplay.ts
pmconne: I can't tell if you did this or not, or what specific difficulties you encountered. People (e.g., Booster) who want to adjust the current contour settings will need these methods.

core/common/src/ContourDisplay.ts
pmconne: Prefer to use a unit test to prove you got it right.

markschlosseratbentley: Do you happen to know if we have any existing unit tests that execute this code flow that we can take a look at to base this one?

pmconne: I would have to grep.

core/frontend/src/render/webgl/Contours.ts
DStradley: @pmconne you didn't comment on my changes to computeDimensions, and now I see the AuxChannelTable test is failing (line 147) because it didn't expect it to fit in width 8 with only 1 row.  With my change, it allows it to use the full width (maxSize here) before going to square, so this is what I would expect.  But since the test was coded not to, I feel I should ask.  Is there some reason not to use the full max width, or should I just change the test?

core/common/src/ContourDisplay.ts
pmconne: [Right here](https://github.com/iTwin/itwinjs-core/blob/12362e64b682cbff804cbd7becd75003e522ab0b/core/backend/src/test/standalone/DisplayStyle.test.ts#L83).

docs/changehistory/NextVersion.md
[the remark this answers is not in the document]
pmconne: I linked to your article from the main display system docs index and the ContourDisplay API docs.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,368 @@
...
+  /** See [[Contour.majorIntervalCount]]. */
+  majorIntervalCount?: number;
+  /** See [[Contour.showGeometry]] */
+  showGeometry?: boolean;
+}
+
+/** The rendering styling settings that apply to a specific set of subcategories within a [[ContourGroup]].
+ * This actually describes stylings for two sets of contours: major and minor. These stylings are separate from each other.
+ * The minor contour occurs at a defined interval in meters. These intervals draw at a fixed height; they are not dependent on the range of the geometry to which they are applied.
+ * The major contour is dependent on the minor contour. The interval of its occurence is not measured directly in meters; rather it is a count of minor contour intervals between its occurrences.
```
pmconne, line 111: > it is a count of minor contour intervals between its occurrences.

Is this accurate? i.e.,:
majorInterval = 0 means every contour is major,
1 means every other contour is major, and
2 means every third contour is major?
I thought I recalled from your demos that every nth contour was major, where n = majorInterval.

markschlosseratbentley: Thanks for the catch - you are right; this is unclear and indicating the wrong thing.  Your understanding from the demo is the correct one.  I pushed a different explanation to both the .md and the .ts.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,387 @@
...
+  /** See [[Contour.majorIntervalCount]]. */
+  majorIntervalCount?: number;
+  /** See [[Contour.showGeometry]] */
+  showGeometry?: boolean;
+}
+
+/** Describes how to generate and style contour lines for geometry within a single [[ContourGroup]].
+ * Contours provide a way to visualize elevation within a 3d scene by drawing lines at fixed intervals along the z-axis.
+ * There are actually 2 kinds of contour lines: major and minor. Each kind can be styled independently.
+ * A contour line is generated every [[minorInterval]] meters. Every `nth` line will be a *major* contour, where `n` = [[majorInterval]]; the intervening lines will
```
pmconne, line 118: Document what happens if `majorInterval` is less than 1 or not an integer, and what happens if `minorInterval` is <= 0.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,387 @@
...
+  public readonly name: string;
+
+  /** The set of subcategories belonging to this group, or an empty set if this is a default group.
+   * @see [[isDefaultGroup]] to test if this is a default group.
+   */
+  public get subCategories(): OrderedId64Iterable {
+    return CompressedId64Set.iterable(this._subCategories);
+  }
+
+  /** Returns true if [[subCategories]] is an empty set, indicating that any subcategory not included in any other [[ContourGroup]] is implicitly
```
pmconne, line 237: Document what happens if more than one default group exists.

core/common/src/ContourDisplay.ts
```diff
@@ -0,0 +1,463 @@
...
+/** The style settings used by either a minor or major contour.
+ * @see [[Contour.majorStyle]]
+ * @see [[Contour.minorStyle]]
+ * @public
+ */
+export class ContourStyle {
+  /** The color in which to draw the contour lines. Default: black. */
+  public readonly color: RgbColor;
+  /** The width in screen pixels of the contour lines.
+   * Useful values range between 1 and 8.5, in increments of 0.5. Other values will be rounded to meet these criteria.
```
pmconne, line 40: I assumed but did not verify that rounding is what happens, please adjust if I guessed wrong.

markschlosseratbentley: @DStradley please take a look when you can.

core/common/src/ContourDisplay.ts
markschlosseratbentley: documented.

core/common/src/ContourDisplay.ts
markschlosseratbentley: @DStradley please take a look when you can.

core/common/src/ContourDisplay.ts
markschlosseratbentley: this was pushed this morning, and I added tests for it.

core/frontend/src/render/webgl/Contours.ts
pmconne: No, it's just a result of you swapping `<` and `<=`.

core/common/src/ContourDisplay.ts
markschlosseratbentley: Thanks. I added a unit test for contour subCategory remapping in DisplayStyle.test.ts in the backend dir. It found a problem and I also pushed a fix for that.

core/backend/src/DisplayStyle.ts
```diff
@@ -242,6 +250,23 @@ export class DisplayStyle3d extends DisplayStyle {
         }
         targetElementProps.jsonProperties.styles.planProjections = remappedPlanProjections;
       }
+
+      if (targetElementProps?.jsonProperties?.styles?.contours?.groups) {
+        const groups = targetElementProps.jsonProperties.styles.contours.groups;
```
pmconne, line 255:
```suggestion
      const groups = targetElementProps?.jsonProperties?.styles?.contours?.groups;
      if (groups) {
```

core/common/src/ContourDisplay.ts
DStradley: Values outside of that range are clamped to that range, values inside that range are rounded to 0.5 increments.

pmconne (approved):
Awesome.

core/common/src/ContourDisplay.ts
DStradley: Major interval is rounded to an integer.  If it is 1, no minor intervals will be drawn.  If it is <= 0, no major intervals will be drawn.  (The latter is observed behavior, probably due to nan's in the shader.)
Minor intervals = 0 causes no contour lines to be drawn (likely due to shader nan).
Minor intervals < 0 behave as if they were positive (abs(interval, and not explicitly checked for).
Should we add some additional checks for Major <=0 and minor <=0, and if so what should be the behavior?

pmconne: The docs should say what the allowed values are and what happens to values that don't meet that criterion (e.g., round non-integers, clamp to allowed range, whatever). Ideally don't rely on NaN in shader just happening to produce a result.

DStradley: I pushed code changes to keep in range, and @markschlosseratbentley pushed documentation changes for this, so should be done.

PR #7220, opened 2024-10-01
Run veracode scan on CI runs on master

nick4598:
If reviewer would prefer, I could also do a cron job for this (maybe once a week)?

Successful run: https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3244516&view=logs&j=a99115e4-c23d-52c8-02f5-2471221e66e4

Scan is currently in progress. When I did this on my local machine awhile back, I produced a 19.8 MB zip file. This pipeline produced a 21 MB zip file. 

19.8 MB scan took roughly 1 hour to complete.


common/config/azure-pipelines/jobs/fast-ci.yaml
```diff
@@ -35,8 +35,37 @@ pr:
...
+  - job: VeracodeScan
+    condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
+    displayName: Veracode SAST Scan
+    pool:
+      vmImage: ubuntu-latest
+    steps:
+      - template: templates/appsec/veracode/package-veracode.yaml@build-pipeline-scripts
+        parameters:
+          foldersToInclude: "core,editor,domains,example-code,extensions,presentation,test-apps,tools,ui,utils" #note I skipped common, docs, full-stack-tests
+          foldersToExclude: "test"
```
hl662, line 54: I saw in the CI run that files in subdirs `public` folders are being copied - should we exclude those as well?

nick4598, line 54:
```suggestion
          foldersToExclude: "test,public"
```

nick4598: It doesn't seem there are any code files in those folders so that makes sense, I pushed that change.

hl662 (approved):

nick4598:
closes https://github.com/iTwin/itwinjs-backlog/issues/1257

common/config/azure-pipelines/jobs/fast-ci.yaml
```diff
@@ -35,8 +35,37 @@ pr:
...
 jobs:
+  - job: VeracodeScan
+    condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
+    displayName: Veracode SAST Scan
+    pool:
+      vmImage: ubuntu-latest
+    steps:
+      - template: templates/appsec/veracode/package-veracode.yaml@build-pipeline-scripts
+        parameters:
+          foldersToInclude: "core,editor,domains,example-code,extensions,presentation,test-apps,tools,ui,utils" #note I skipped common, docs, full-stack-tests
```
aruniverse, line 53: im wondering if we should exclude test-apps

nick4598: I'm not opposed to it. I suppose the goal is to minimize vulnerabilities in our code that we ship out, so it seems reasonable to me to exclude test-apps.

PR #7223, opened 2024-10-01
Limit the number of returned exports to 5

danieliborra:
We can reduce the response times of querying exports (mesh export service), by limiting the number of exports we want to receive. That reduces the size of the response and the time to generate the information for each export.


markschlosseratbentley (approved):

jason-crow (approved):

eringram (approved):

eringram:

PR #7227, opened 2024-10-02
Fixed issue with arc-arc intersection returning invalid intersections

saeeedtorabi:
Code was returning imaginary roots for one of the quadratics derived from the quartic by way of the resolvent cubic. Now no roots are returned for this case. This was apparently a fat-finger from the native port.


dassaf4 (approved):
Looks good. I made some changes to the markdown file, and suggested a few comment changes elsewhere.

PR #7228, opened 2024-10-02
Make pnpm a prerequisite for development, update debugger config for vitest enabled packages

hl662:
Developers using `npm` might not exit vitest coverage runs before the timeout for the underlying `close()` kicks in, and the solution would be to use `pnpm` instead, which is more performant. `pnpm` isn't an explicit requirement, but in light of this, it should be the recommended package manager devs use when contributing. Updated our root README to reflect that.

Also, updated our `launch.json` to support debugging vitest. Just enabling it per package that already moved to vitest. Dropped `outFiles`, we don't need to explicitly specify this to enable source maps when debugging. Most important part of the configuration is to pass in `autoAttachChildProcesses`, because vitest uses child processes to run tests in multiple threads.

Also, documenting here that the vitest maintainer says you can't filter a test like how we did it with chai: https://github.com/vitest-dev/vitest/issues/1497
 
It seems that `.only` does not filter the test files, it filters out the tests cases inside of a found test file. 
To quickly run a single test file you can go to vitest.config.mts and add include: "<testfileName>", right below setupFiles - that will tell vitest to run test files that match that pattern. Then if there is a .only inside the test file, vitest will run only the describe.only or it.only found within.


README.md
```diff
@@ -20,6 +20,7 @@ Each package has its own **node_modules** directory that contains symbolic links
 
 - [Git](https://git-scm.com/)
 - [Node](https://nodejs.org/en/): an installation of the latest security patch of Node 20. The Node installation also includes the **npm** package manager.
+- [pnpm](https://pnpm.io/installation#using-npm); `pnpm` is our preferred package manager, it is more performant and monorepo friendly than `npm`. To install, do `npm install -g pnpm`
```
aruniverse, line 23: i dont recommend installing pnpm via npm. use corepack

hl662:
@saeeedtorabi @dassaf4 try setting a breakpoint in one of the geometry test files, and launch the debugger profile for `Geometry Tests`. You can also filter out the one test file you are setting a breakpoint in via vitest.config.mts `include` if you don't want to run all the other tests...

aruniverse:
can you update the readme to reference using pnpm instead of npm too?
i see at the bottom for example that says run `npm run docs`

nick4598:
Is it weird to say pnpm when this is a rush repo? Rush does use pnpm under the hood im pretty sure, but not sure if that would be confusing at all? I could just as easily run `rushx docs` in core/backend, which I think is essentially the same thing.

It also does seem the majority of the messaging in that readme is rush based

hl662:
> Is it weird to say pnpm when this is a rush repo? Rush does use pnpm under the hood im pretty sure, but not sure if that would be confusing at all? I could just as easily run `rushx docs` in core/backend, which I think is essentially the same thing.
> 
> It also does seem the majority of the messaging in that readme is rush based

Yes, it may be weird, but it's not easy for developers to tell that there's a difference. You can easily run `npm docs` or `npm test` in the individual folders as well. 

Maybe we can encourage developers to use `pnpm`, or `rushx` in individual folders, by explicitly stating that in the root README of our repo

README.md
hl662: updated readme - we should be aware the instructions might change in the future depending on what the node maintainers [decide to do](https://medium.com/@pacoarjona/node-js-leaves-npm-55d022e95a4f). A later problem though

dassaf4:
> > I could just as easily run `rushx docs` in core/backend, which I think is essentially the same thing.

I'll add here that on my laptop, pnpm is faster than rushx. Just today we were testing vitest, and found that only `pnpm cover` successfully completes in core-geometry, while `rushx cover` times out just like `npm run cover`.

saeeedtorabi:
I tested debugging in VSCode using this PR. This PR fixed the debugging issue.

However, the old method of debugging was nuch faster (specially when you test a single unit test using `.only`). It used to take a couple seconds to get to the first breakpoint. The new method is slow to get to the first breakpoint no matter you test a single unit test or all unit tests. It took me around 15 sec to get to the first breakpoint in VSCode debugger (first time took 60 sec but I repeated and it settled at 15 sec and did not reduce no matter how many times I ran the tests).

I see this message in VSCode and wondering if that can improve the speed:

Screenshot 2024-10-02 134544

hl662:
> I tested debugging in VSCode using this PR. This PR fixed the debugging issue.
> 
> However, the old method of debugging was nuch faster (specially when you test a single unit test using `.only`). It used to take a couple seconds to get to the first breakpoint. The new method is slow to get to the first breakpoint no matter you test a single unit test or all unit tests. It took me around 15 sec to get to the first breakpoint in VSCode debugger (first time took 60 sec but I repeated and it settled at 15 sec and did not reduce no matter how many times I ran the tests).

I can look to add another command in launch.json that can debug the current file that is open, I just need to state that command is for vitest converted test files only, since we still are using mocha+chai in other places

hl662:
Added another launch configuration: 'Debug Current Vitest File`, and it will debug just the active tab you have opened in vs code

dassaf4:
> I can look to add another command in launch.json that can debug the current file that is open in your vs code editor, I just need to state that command is for vitest converted test files only, since we still are using mocha+chai in other places

Since all core-geometry tests are converted to vitest, can this be made the default for the Geometry Tests section?

hl662:
> > I can look to add another command in launch.json that can debug the current file that is open in your vs code editor, I just need to state that command is for vitest converted test files only, since we still are using mocha+chai in other places
> 
> Since all core-geometry tests are converted to vitest, can this be made the default for the Geometry Tests section?

I've already updated the Geometry Tests section to have the debugger work, it just runs against all of geometry's vitest files by default. You can use this other launch config to debug just the one test file you might have opened. Both launch configs will catch your breakpoints, but the latter will make it quicker to get to that breakpoint since it only runs one test file

saeeedtorabi:
> Added another launch configuration: 'Debug Current Vitest File`, and it will debug just the active tab you have opened in vs code

I tested your last commit. Now it takes 50 sec to get to the first breakpoint! I ran it multiple times but it made it worse.

hl662:
> I tested your last commit. Now it takes 50 sec to get to the first breakpoint! I ran it multiple times but it made it worse.

@saeeedtorabi try the 'debug current file in vitest' again - I updated that config to not spawn multiple child threads when debugging just a single file, since attaching the debugger to each thread would be a reason the startup time took so long. Me and a colleague noticed it was faster now

nick4598:
"To quickly run a single test file you can go to vitest.config.mts and add include: "", right below setupFiles - that will tell vitest to run test files that match that pattern. "

is this information captured anywhere besides the description of the PR?

nick4598:
image
Do we know that this screenshot from Saeed's posts isn't important? I see you mentoined that we don't need to explicitly set this in the description, but the ss is a bit concerning for performance reasons isn't it?

.vscode/launch.json
```diff
@@ -13,6 +13,23 @@
...
+      "type": "node",
+      "autoAttachChildProcesses": true,
+      "skipFiles": ["<node_internals>/**", "**/node_modules/**"],
+      "program": "${workspaceFolder}/common/temp/node_modules/vitest/vitest.mjs",
+      "args": [
+        "run",
+        "${relativeFile}",
+        "--no-file-parallelism"
+      ],
+      "smartStep": true,
```
nick4598, line 30: what does smartStep do?

README.md
tcobbs-bentley: In other words, I feel that "preferred" is incorrect and should be "required". It's true that you can use npm to run scripts from package.json files in the various subdirectories, and I guess pnpm (or better yet rushx?) might be "preferred" for doing that.

.vscode/launch.json
```diff
@@ -58,13 +75,13 @@
       "cwd": "${workspaceFolder}/core/bentley/",
       "type": "node",
       "request": "launch",
-      "runtimeExecutable": "npm",
+      "runtimeExecutable": "pnpm",
       "runtimeArgs": [
         "test"
       ],
-      "outFiles": [
```
MichaelSwigerAtBentley, line 65: I'm guessing outFiles is unneeded? I figure CWD is enough to locate the tests, and constraining the search to the outFiles doesn't save time?

README.md
hl662: That's true, I should change it to `required` - as for preference, it was my intention to update this readme to recommend that people **do** use `pnpm` to run scripts within subdirectories, I think I'll find a spot in the readme to explicitly say that

.vscode/launch.json
hl662: It helps when running through transpiled .js code - some parts of the code aren't covered by source maps, like async/await processes. We would frequently hit these weird parts as we debug, normally you'd have to 'step out' until you go back to a part of the code that is resolved by source maps. `smartStep` skips showing you those parts automatically for you. [Reference](https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_smart-stepping)

README.md
tcobbs-bentley: I definitely don't have a problem with indicating that pnpm is preferred (vs. required) over npm when running specific scripts in subdirectories. This line implies to me that pnpm is only preferred and not required for the repository as a whole, since this is the top-level README.md.

hl662:
> "To quickly run a single test file you can go to vitest.config.mts and add include: "", right below setupFiles - that will tell vitest to run test files that match that pattern. "
> 
> is this information captured anywhere besides the description of the PR?

David provided an example within core geometry's vitest.config.mts in a recent [commit](https://github.com/iTwin/itwinjs-core/blob/master/core/geometry/vitest.config.mts) to master

saeeedtorabi:
@hl662 I have issues with debugging in VS Code using this change. `Step Into` does not work properly.

hl662:
> @hl662 I have issues with debugging in VS Code using this change. `Step Into` does not work properly.

I hopped on a call with Saeed to look at the issue - the debugger is working as expected - when setting a breakpoint on a line that calls a static method, like `Line.doThis`, he expected 'step into' to go into the `doThis` method, but instead it actually goes into 'Line', pointing to the closing brace of the class. That's expected - step into should retrieve the class object containing the static method first, before diving into the static method itself.

hl662:
>  I see you mentoined that we don't need to explicitly set this in the description, but the ss is a bit concerning for performance reasons isn't it?
It's hard to justify including the outFiles - vitest uses vite to build and compile, and by default, the outDir of that is `dist`, not `lib/` like how we specify. By default, vitest includes everything in the workspaceFolder, it doesn't change any of the options specified [here](https://github.com/microsoft/vscode-js-debug/blob/main/OPTIONS.md#outfiles), but even so, maybe I should keep outFiles in, maybe in case of stepping into adjacent subdirectories and their methods...

saeeedtorabi:
> > @hl662 I have issues with debugging in VS Code using this change. `Step Into` does not work properly.
> 
> I hopped on a call with Saeed to look at the issue - the debugger is working as expected - when setting a breakpoint on a line that calls a static method, like `Line.doThis`, he expected 'step into' to go into the `doThis` method, but instead it actually goes into 'Line', pointing to the closing brace of the class. That's expected - step into should retrieve the class object containing the static method first, before diving into the static method itself.

This might be the expected new behaviour but I'm pretty sure it was not the old behaviour. I think running the tests is made faster by new changes but debugging is harder and slower now. I hope it can be improved because in the long run people spend more time on debugging than just running the tests.

.vscode/launch.json
hl662: it... could? Although from my comment below, responding to Nick, I'm a little skeptical it would have an impact. I'll re add the outFiles, but I need to holistically investigate vitest's performance in general, as a separate PR

README.md
hl662: I've updated this section to point out `pnpm` is required, and added a quote block recommending people to use `pnpm` over `npm` when running scripts within subdirectories

.vscode/launch.json
hl662: added back

README.md
tcobbs-bentley: Looks good. I won't resolve this conversation, since I didn't start it.

.vscode/launch.json
MichaelSwigerAtBentley: Yeah, skeptical it would make any difference too, but if there is no cost to adding it in, might as well keep it for now. Resolving

nick4598:
> It's hard to justify including the outFiles - vitest uses vite to build and compile, and by default, the outDir of that is `dist`, not `lib/` like how we specify. By default, vitest includes everything in the workspaceFolder, it doesn't change any of the options specified [here](https://github.com/microsoft/vscode-js-debug/blob/main/OPTIONS.md#outfiles), but even so, maybe I should keep outFiles in, maybe in case of stepping into adjacent subdirectories and their methods...

Your link takes me here: 
image

Is that what you intended?

vitest includes everything in the workspaceFolder by default when searching for outFiles? Right now the launch.json is more specific than workspaceFolder for outFiles, i.e. `"${workspaceFolder}/extensions/frontend-tiles/lib/**/*.js"`. 

if vite uses an outDir of dist by default, why shouldn't we change our `outFiles` to have dist instead of lib as opposed to removing them completely?

README.md
dassaf4: > i dont recommend installing pnpm via npm. use corepack

Before @aruniverse's recommendation, I installed with "npm install -g pnpm". This works fine. Then I tried "corepack enable pnpm" but this gave an error unless I used an administrator shell. Also I'm not sure this did anything since I already have pnpm. So:
* Is "corepack enable pnpm" all one needs to do to get pnpm the recommended way?
* If so, maybe just say that instead of providing the link?
* And maybe add a note about admin privileges.

hl662:
> Is that what you intended?
> 
> vitest includes everything in the workspaceFolder by default when searching for outFiles? Right now the launch.json is more specific than workspaceFolder for outFiles, i.e. `"${workspaceFolder}/extensions/frontend-tiles/lib/**/*.js"`.
> 
> if vite uses an outDir of dist by default, why shouldn't we change our `outFiles` to have dist instead of lib as opposed to removing them completely?

Oof yeah, I should have pointed out the `outFiles` param is `node:launch` dropdown in that link.

So I just tried out outFiles to match any `.js` files in either `dist` or `lib` and it made no difference to the performance. It also caused bad behavior in the debugger - it wasn't tracing the lines properly, and the block scoped variables in the debugging panel didn't match where I was stepping through. I had to remove outFiles to get the debugger to work properly.
image

.vscode/launch.json
```diff
@@ -13,6 +13,23 @@
       "request": "attach",
       "processId": "${command:PickProcess}"
     },
+    {
+      "name": "Debug Current Vitest File",
```
dassaf4, line 17: Code tells me the "Debug Current Vitest File" scope is missing a "request" property.

hl662: thanks, I've added it

README.md
hl662: I've edited the README according to your 2 bullet points dave, the third one I didn't see anything new added to the top package.json... I did run npm uninstall -g pnpm before doing it, but no dice

dassaf4: Did your next "pnpm test" install pnpm? That's when the "packageManager" line got appended to package.json. Or maybe you already had that line. It did inform me that I lacked such a line, and so it was adding it.

hl662: Yes, I ran pnpm test after running corepack. The thing is, we shouldn't need a packageManager property to be appended to package.json, not at the root level for sure, since our rush.json has a [property](https://github.com/iTwin/itwinjs-core/blob/52e966a8b3ddaafa57315a06755d14ffb22b0d65/rush.json#L4) that tells the whole repo to use pnpm as a package manager

dassaf4 (approved):

hl662:
@dassaf4 @saeeedtorabi 
I just pushed a commit that adds support for a test explorer for vitest. It adds a suggestion to install an extension called 'vitest.explorer'. Can you install it andd see if in the `Testing` tab of VSCode, you can see something like this: 
image
 The UI should be pretty straight forward to interpret how to debug a single test case, i tried setting a breakpoint and it correctly steps over as well

dassaf4:
> @dassaf4 @saeeedtorabi I just pushed a commit that adds support for a test explorer for vitest. It adds a suggestion to install an extension called 'vitest.explorer'.

Brilliant! This adds a "Testing" icon that you can use instead of "Run and Debug". No more "it.only". No more editing vitest.config.mts to test a single file. Run a single test, or all the tests at the describe/file/repo level directly within Code. And it journals the results in a new "Test Results" tab. 
image

hl662:
> Brilliant! This extension adds a "Testing" icon that you can use instead of "Run and Debug". 
Great, I'm going to remove the 'debug current vitest file' from launch.json, it has a lot of problems with it and executing that in a monorepo, I'll recommend using the vitest extension instead

README.md
```diff
@@ -69,6 +70,8 @@ For incremental builds, the `rush build` command can be used to only build packa
 11. Add and commit the changelog JSON files and any API signature updates.
 12. Publish changes on the branch and open a pull request.
 
+> If executing scripts from 'package.json` files in any of the subdirectories, we recommend using pnpm over npm, as it is more performant.
```
anmolshres98, line 73: let's recommend using `rushx` instead since the default package manager is already `pnpm`

README.md
```diff
@@ -92,4 +95,4 @@ Use these instructions to update dependencies and devDependencies on external pa
 ## Other NPM Scripts
 
 1. Build TypeDoc documentation for all packages: `rush docs`
-2. Build TypeDoc documentation for a single package: `cd core\backend` and then `npm run docs`
+2. Build TypeDoc documentation for a single package: `cd core\backend` and then `pnpm run docs`
```
anmolshres98, line 98:
```suggestion
2. Build TypeDoc documentation for a single package: `cd core\backend` and then `rushx docs`
```

README.md
hl662: done

anmolshres98:
> This might be the expected new behaviour but I'm pretty sure it was not the old behaviour. 

I agree with what Saeed said here. I don't remember a scenario where while stepping into a call the debugger stepped into the closing brace of the class

README.md
```diff
@@ -69,6 +70,8 @@ For incremental builds, the `rush build` command can be used to only build packa
 11. Add and commit the changelog JSON files and any API signature updates.
 12. Publish changes on the branch and open a pull request.
 
+> If executing scripts from 'package.json` files in any of the subdirectories, we recommend using rushx over npm, as it is more performant.
```
aruniverse, line 73:
```suggestion
> If executing scripts from 'package.json` files in any of the subdirectories, we recommend using [`rushx`](https://rushjs.io/pages/commands/rushx/) over `npm`.

aruniverse (approved):

PR #7230, opened 2024-10-02
skip flaky CloudSqlite test 'should LogLevel.Trace set LogMask to ALL' temporarily

nick4598:


pmconne (approved):

nick4598:
@pmconne I think I see why my previous fix did nothing, my check function did not throw. Could you take a look again and let me know if you're comfortable trying this out?

pmconne:
> let me know if you're comfortable trying this out?

go for it.

PR #7232, opened 2024-10-02
Update to TypeScript 5.6

wgoehrig:
It's TypeScript update time again!

Since our last update, there have been 3 major TypeScript versions released (remember that TypeScript does _not_ follow semver).  As always, the release notes are the best way to catch up on what's changed since 5.3:
- [TS 5.4 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html)
- [TS 5.5 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html)
- [TS 5.6 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/)

Even more than usual, these new versions bring better compiler performance and (especially) better type inferencing.  Other highlights include:

- [Regex Syntax Checking](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#regular-expression-syntax-checking) - TypeScript can now tell you when you have syntax errors in your regular expressions.
- [Isolated Declarations](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#isolated-declarations) - We're certainly not ready to opt-in to this new, stricter enforcement for type annotations, but someday this may allow us to _fully_ parallelize the _entire_ monorepo's build!
- [Disallowed Nullish and Truthy Checks](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#disallowed-nullish-and-truthy-checks) - Some expressions that always evaluate either truthy or nullish will now be flagged as errors.  Very handy to catch these at compile time since these can be easy mistakes to make.
- [Iterator Helper Methods](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#iterator-helper-methods) - This implements support for a very cool new ECMAScript proposal that will make it _much_ easier to work with generators and iterators.

Most of the breaking compiler changes were due to the newly-disallowed nullish and truthy checks affecting some patterns in testing/debugging code.  However, better type inferencing also means some new lint errors, and while I'm going to try to hold off on updating ESLint until #7233 there are of course a few lint failures caused just by updating our TypeScript version.

PR #7233, opened 2024-10-02
Update to ESLint 9.13

wgoehrig:
Note that in the latest version of @itwin/eslint-plugin, we've removed all formatting rules, as these were [deprecated in ESLint 9](https://eslint.org/blog/2023/10/deprecating-formatting-rules/).  This should significantly improve linting speed.

Also note that the formatting of lint output has changed, since the `visualstudio` reporter was [removed in ESLint 9](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/#removed-all-formatters-except-stylish%2C-html%2C-json%2C-and-json-with-meta).  Personally, I prefer the default output anyway.


core/frontend/src/AccuDraw.ts
```diff
@@ -2305,7 +2303,7 @@ export class AccuDraw {
     if (!this._distanceRoundOff.active || !this._distanceRoundOff.units.size)
       return undefined;
 
-    let roundValue = this._distanceRoundOff.units.values().next().value!; // eslint-disable-line @typescript-eslint/no-non-null-assertion
+    let roundValue = this._distanceRoundOff.units.values().next().value!;
```
pmconne, line 2306: Are we no longer flagging non-null assertions as? If so why?

pmconne (approved):
I made it to Viewport.test.ts at which point github wanted me to click "Load diff" for every subsequent file. I did some random sampling of the remaining files, I will assume if it compiles+lints then it's fine.

core/frontend/src/AccuDraw.ts
wgoehrig: The rule is enabled for non-test files in core-backend.  As far as I can tell it's never been enabled in core-frontend.

The comment here was just copypasta I accidentally inserted in #7232

pmconne: That is dismaying. Do you know why it's not enabled in the base config? I have lost count of how many bugs/crashes in production have been traced back to somebody's overly-optimistic exclamation mark.

grigasp (approved):
Approving presentation changes. I like that the new version reports unused "ignore" directives - clearly we had a number of them dangling unnecessarily.

core/frontend/src/AccuDraw.ts
grigasp: If that somebody is willing to add `!` to make life easier, what makes you think she/he won't add `eslint-disable-...` for it too? TS already lets you know that there's a problem and allows you to ignore it by adding the exclamation mark. I don't think adding another warning and another way to ignore it will solve anything - it's up to the developer and reviewer.

pmconne: Exclamation mark is way too easy to overlook during review, vs `assert`.
[Exhibit A](https://github.com/iTwin/itwinjs-backlog/issues/1243#issuecomment-2333862507).
```ts
  private ecefToPixelFraction(point: Point3d, applyTerrain: boolean): Point3d {
    const cartoGraphic = Cartographic.fromEcef(point)!;
    return Point3d.create(this.longitudeToXFraction(cartoGraphic.longitude), this.latitudeToYFraction(cartoGraphic.latitude), applyTerrain ? cartoGraphic.height : 0);
```

But this PR isn't about changing the lint config.

dassaf4 (approved):
core-geometry looks good

aruniverse (approved):

calebmshafer (approved):

core/electron/src/backend/ElectronHost.ts
```diff
@@ -123,8 +123,8 @@ export class ElectronHost {
     // However, if the file doesn't exist, it's fine to return an invalid path here - the request will just fail with net::ERR_FILE_NOT_FOUND
     try {
       assetPath = fs.realpathSync(assetPath);
-    } catch (error) {
-      // eslint-disable-next-line no-console
+    } catch {
+      // // eslint-disable-next-line no-console
```
ColinKerr, line 127: double `//`?

pmconne:
```suggestion
      // eslint-disable-next-line no-console

```

wgoehrig: I did that in a number of places - the actual console.log below is commented out - my preference would be to remove _both_ commented-out lines, but thought that was overstepping in a lint PR.

I saw this a lot in tests and figured if these are there for quick debugging it's not hard to uncomment both lines.  I can always just remove all double-commented directives tho, that probably makes more sense if the console.logs should always be temporary.  Thoughts?

wgoehrig: the reason for the double-comment btw is that these are now flagged as unused directives - so the above suggestion will break

ColinKerr (approved):

ben-polinsky (approved):

PR #7238, opened 2024-10-04
Fix regression caused by ClassId conversion to class name vs hexid

khanaffan:
imodel-native: https://github.com/iTwin/imodel-native/pull/878

### DIFF
--- a/core/backend/src/ECSqlStatement.ts
+++ b/core/backend/src/ECSqlStatement.ts
@@ -343,9 +343,6 @@ export class ECSqlStatement implements IterableIterator<any>, IDisposable {
     if (args.rowFormat === undefined) {
       args.rowFormat = QueryRowFormat.UseJsPropertyNames;
     }
-    if (args.classIdsToClassNames === undefined) {
-      args.classIdsToClassNames = true;
-    }
     const resp = this._stmt.toRow({
       classIdsToClassNames: args.classIdsToClassNames,
       useJsName: args.rowFormat === QueryRowFormat.UseJsPropertyNames,
--- a/core/backend/src/ECSqlStatement.ts
+++ b/core/backend/src/ECSqlStatement.ts
@@ -34,6 +34,7 @@ export class ECSqlInsertResult {
 export interface ECSqlRowArg {
   /** Determine row format. */
   rowFormat?: QueryRowFormat;
+  classIdsToClassNames?: boolean;
 }
 
 /** Executes ECSQL statements.
@@ -332,46 +333,42 @@ export class ECSqlStatement implements IterableIterator<any>, IDisposable {
    * - [Code Samples]($docs/learning/backend/ECSQLCodeExamples#working-with-the-query-result)
    */
   public getRow(args?: ECSqlRowArg): any {
-    args = args ?? {};
+    if (!this._stmt)
+      throw new Error("ECSqlStatement is not prepared");
 
-    // Set default rowFormat if not provided
+    args = args ?? {};
     if (args.rowFormat === undefined) {
       args.rowFormat = QueryRowFormat.UseJsPropertyNames;
     }
-
-    const toRowOptions = {
-      classIdsToClassNames: true,
-      rowFormat: args.rowFormat,
-    };
-    const resp = this._stmt!.toRow(toRowOptions); // eslint-disable-line @typescript-eslint/no-non-null-assertion
+    if (args.classIdsToClassNames === undefined) {
+      args.classIdsToClassNames = true;
+    }
+    const resp = this._stmt.toRow({
+      classIdsToClassNames: args.classIdsToClassNames,
+      useJsName: args.rowFormat === QueryRowFormat.UseJsPropertyNames,
+      abbreviateBlobs: false,
+      doNotConvertClassIdsToClassNamesWhenAliased: true,
+    });
     return this.formatCurrentRow(resp, args.rowFormat);
   }
 
-  /**
-   * @internal
-   */
-  public formatCurrentRow(currentResp: any, rowFormat: QueryRowFormat = QueryRowFormat.UseJsPropertyNames): any[] | object {
+  private formatCurrentRow(currentResp: any, rowFormat: QueryRowFormat = QueryRowFormat.UseJsPropertyNames): any[] | object {
+    if (!this._stmt)
+      throw new Error("ECSqlStatement is not prepared");
+
+    if (rowFormat === QueryRowFormat.UseECSqlPropertyIndexes)
+      return currentResp.data;
+
     if (this._props.length === 0) {
-      const resp = this._stmt!.getMetadata(); // eslint-disable-line @typescript-eslint/no-non-null-assertion
+      const resp = this._stmt.getMetadata();
       this._props = new PropertyMetaDataMap(resp.meta);
     }
     const formattedRow = {};
-    const uniqueNames = new Map<string, number>();
     for (const prop of this._props) {
       const propName = rowFormat === QueryRowFormat.UseJsPropertyNames ? prop.jsonName : prop.name;
       const val = currentResp.data[prop.index];
       if (typeof val !== "undefined" && val !== null) {
-        let uniquePropName = propName;
-        if (uniqueNames.has(propName)) {
-          let currentValue = uniqueNames.get(propName);
-          currentValue = currentValue ? currentValue + 1 : 1;
-          uniqueNames.set(propName, currentValue);
-          uniquePropName = `${propName}_${currentValue}`;
-        } else {
-          uniqueNames.set(propName,0);
-        }
-
-        Object.defineProperty(formattedRow, uniquePropName, {
+        Object.defineProperty(formattedRow, propName, {
           value: val,
           enumerable: true,
           writable: true,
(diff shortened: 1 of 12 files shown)


core/backend/src/ECSqlStatement.ts
```diff
@@ -332,46 +336,43 @@ export class ECSqlStatement implements IterableIterator<any>, IDisposable {
...
       args.rowFormat = QueryRowFormat.UseJsPropertyNames;
     }
-
-    const toRowOptions = {
-      classIdsToClassNames: true,
-      rowFormat: args.rowFormat,
-    };
-    const resp = this._stmt!.toRow(toRowOptions); // eslint-disable-line @typescript-eslint/no-non-null-assertion
+    if (args.classIdsToClassNames === undefined) {
+      args.classIdsToClassNames = true;
```
nick4598, line 347: Can you explain why its default true?

core/backend/src/ECSqlStatement.ts
```diff
@@ -332,46 +336,43 @@ export class ECSqlStatement implements IterableIterator<any>, IDisposable {
...
+    }
+    const resp = this._stmt.toRow({
+      classIdsToClassNames: args.classIdsToClassNames,
+      useJsName: args.rowFormat === QueryRowFormat.UseJsPropertyNames,
+      abbreviateBlobs: false,
+      // on 5.x this should be removed or set to false. It is only here to maintain compatibility with 4.x
+      doNotConvertClassIdsToClassNamesWhenAliased: true,
```
nick4598, line 354: Does 'whenAliased' mean if I were to `select classId as aliasedClassId from table` ?

nick4598: It seems that I'm wrong since you ended up changing some tests from `assert.isTrue(Id64.isValidId64(row.myParentRelClassId));` to `assert.isFalse(Id64.isValidId64(row.myParentRelClassId));`

Wouldnt the above be an aliased class Id and therefore it should stay as a classId?

core/backend/src/ECSqlStatement.ts
```diff
@@ -332,46 +336,43 @@ export class ECSqlStatement implements IterableIterator<any>, IDisposable {
...
+      const resp = this._stmt.getMetadata();
       this._props = new PropertyMetaDataMap(resp.meta);
     }
     const formattedRow = {};
-    const uniqueNames = new Map<string, number>();
     for (const prop of this._props) {
       const propName = rowFormat === QueryRowFormat.UseJsPropertyNames ? prop.jsonName : prop.name;
       const val = currentResp.data[prop.index];
       if (typeof val !== "undefined" && val !== null) {
-        let uniquePropName = propName;
```
nick4598, line 364: Is this snippet no longer necessary because you're checking for uniqueECSqlMembers in EcSqlRowAdaptor::GetMetaData on the native side now?

core/backend/src/ECSqlStatement.ts
khanaffan: I am removing it. It's previously set to true and is noop when useJsNames is true. It's useful when not using jsName.

core/backend/src/ECSqlStatement.ts
khanaffan: You are not wrong. It's a part I left to do for today. For some reason, the name reported when we convert to json is ECRelClassId instead of myParentRelClassId. I was going to debug this one to find why is that case.

core/backend/src/ECSqlStatement.ts
nick4598: Should your response or some variation of it be documented on the classIdsToClassNames property?

core/backend/src/ECSqlStatement.ts
khanaffan: So I look into it and it expected. I was confused at first as well. But the result is from two different API. I first aligned output of both API `ECSqlStatement.toRow()` & `ECSqlReader`.

Then I added `doNotConvertClassIdsToClassNamesWhenAliased` flag for `ECSqlStatement` and set it to true and then updated the test. 

Following is set for ECSqlReader which is standard behavior. ECSqlStatement has different behavior.  On 5.x we will ensure both have same behavior by remove `doNotConvertClassIdsToClassNamesWhenAliased` or setting it to `false`

`assert.isTrue(Id64.isValidId64(row.myParentRelClassId));` to `assert.isFalse(Id64.isValidId64(row.myParentRelClassId));`

nick4598: So if I understand correctly...

ECSqlReader by default converts classIds to classNames (I guess because they are generally more useful to read than an ID)?. ECSQLStatement used to not convert classIds to classNames but it recently started to in some recent PR which caused the regression and broke some queries? Thats the reason for this option, to temporarily make that behavior opt-in in the case of ECSQLStatement. 

What was the motivation behind moving ECSQLStatement to start converting classIds to classNames in that previous PR? I'm guessing to unify the outputs of the Reader and the Statement( seems like a good reason).

khanaffan: > What was the motivation behind moving ECSQLStatement to start converting classIds to classNames in that previous PR? I'm guessing to unify the outputs of the Reader and the Statement( seems like a good reason).
That’s right, we’ve transitioned from using TypeScript base values to object serialization, allowing us to leverage common code on the native side. Additionally, the performance of toRow() has significantly improved, so you’ll notice it’s much faster now.

The aliased property to retrieve the actual ID has been a bug since itwin.js 1.x. Users who need the ID rather than the className find this particularly frustrating, as there’s no way to bypass the default conversion from classID to className other than aliasing it. Some users have resorted to joining the ECClassDef class to get the ID, while others have used casting methods.

This is why I’m completely against this smart conversion. It might benefit a few, but it frustrates the majority. If someone wants the className, they should use SELECT ec_classname(ECClassId). We shouldn’t apply this conversion by default. We should return exactly what the user specifies in their query.

If they want ECInstanceId to be id, they should use `SELECT ECInstanceId AS id`. For camel case, they can simply alias the properties as needed.

Hopefully on 5.x we will need no flags at all.

core/backend/src/ECSqlStatement.ts
```diff
@@ -332,46 +336,40 @@ export class ECSqlStatement implements IterableIterator<any>, IDisposable {
...
-    const toRowOptions = {
-      classIdsToClassNames: true,
-      rowFormat: args.rowFormat,
-    };
-    const resp = this._stmt!.toRow(toRowOptions); // eslint-disable-line @typescript-eslint/no-non-null-assertion
+    const resp = this._stmt.toRow({
+      classIdsToClassNames: args.classIdsToClassNames,
+      useJsName: args.rowFormat === QueryRowFormat.UseJsPropertyNames,
+      abbreviateBlobs: false,
+      // on 5.x this should be removed or set to false. It is only here to maintain compatibility with 4.x
```
nick4598, line 350:
```suggestion
      // on 5.x this should be removed or set to false. It is only here to maintain compatibility with 4.x
      In 4.x, people are currently dependent on the behavior of aliased classIds `select classId as aliasedClassId` not being converted into classNames which is a bug that we must now support. This option preserves this special behavior until it can be removed in a future version.
```

Probably could be better, but I would like to better capture why this option is necessary

core/backend/src/ECSqlStatement.ts
nick4598: Totally agree with not doing smart conversion. While reading the PR I was also confused by the ECIsnstanceId->Id conversions happening seemingly magically. It also seemed to happen for SourceECInstnaceId, and even any properties with the name ECClassId in them as well.

At the very least, any and all smart conversion we do should be documented somewhere for people to read. Is that already the case?
When to use ECSqlStatement vs ECSqlReader also seems worth documenting if that isn't already. 

Doesn't have to happen in this PR, but if we could atleast have workitems that would be nice

aruniverse:

rschili (approved):

core/backend/src/ECSqlStatement.ts
nick4598: Before your PR wasn't `classIdsToClassNames` always true? Why shouldn't it stay that way for 4.x? If someone currently has a query that specifies a rowFormat which is not `useJsPropertyNames`, and they update to a 4.x version containing this PR, wouldn't the output of their query change?  

I'm not talking about the other one `donotconvertclassidstoclassnameswhenaliased` which I assume is the one that was added for backward compatibility, since the comment above it says 'it is only here to maintain compatibility with 4.x'

core/backend/src/ECSqlStatement.ts
khanaffan: Yes, I intend to write a git issue for this explaining the issue and direction we should take on 5.x

core/backend/src/ECSqlStatement.ts
khanaffan: Yes that is correct, though this has been rolled out just recently and that rollout is what broke hatch. It's not advertised (and we have no choice to make it public given its input to a public api otherwise we would have marked it beta). I will not worry about it.

nick4598: Ah okay, I see now thanks. Didn't realize that the `toRowOptions` were part of the PR that broke hatch.

nick4598 (approved):

pmconne (approved):

aruniverse (approved):
Approving addon commit

PR #7249, opened 2024-10-09
Fix async worker return type

### LINKED ISSUE #7248: `registerWorker` typing does now allow async function returning a transferable struct
Trying to call `registerWorker` on a `WorkerImplementation` that contain an async function returning a transferrable (i.e. `{result, transfer}`), will give this typing error:
```
Type '() => Promise<{ result: number; transfer: never[]; }>' is not assignable to type '() => WorkerReturnType<() => Promise<number>>'.
  Type 'Promise<{ result: number; transfer: never[]; }>' is not assignable to type 'WorkerReturnType<() => Promise<number>>'.
    Type 'Promise<{ result: number; transfer: never[]; }>' is not assignable to type 'Promise<number>'.
      Type '{ result: number; transfer: never[]; }' is not assignable to type 'number'.
```

**To Reproduce**
This [PR ](https://github.com/iTwin/itwinjs-core/pull/7247) can be used to reproduce the error with the actual tests.

**Expected behavior**
No typing error.

- OS: Windows
- Browser : chrome
- Version: 129
- iTwin.js Version: 4.9.1

pmconne:
Fixes #7248.

#7088 enabled support for asynchronous implementations of individual worker operations, to prevent long-running operations from blocking subsequent requests. But, the types did not accurately capture these semantics, so you couldn't return `{ result: T, transfer: Transferable[] }` from async implementations without the compiler complaining. It would expect you to return `{ result: Promise<T>, transfer: Transferable[] }` instead.

Adjusted the types to fix this.


aruniverse (approved):

mdastous-bentley:
@aruniverse I would like to get it backported to 4.9.x

aruniverse:

PR #7251, opened 2024-10-09
Support independent color/alpha overrides for lines

### LINKED ISSUE #7244: Support overriding line and fill color independently
Two similar problems:
- @FelixGirard wants the user to be able to change the line+fill colors for subcategories and preview the changes in the viewport before saving them to the iModel.
- @matmarchand wants the user to be able to specify the line+fill colors to be applied to GeoJSON and preview them in the viewport before saving them.

SubCategoryAppearance and GeometryParams permit line color to differ from fill color. The preview could use FeatureAppearance to dynamically override the color, but FeatureAppearance only supports a single color.

Enhance FeatureAppearance to support overriding fill+line color independently.

The only real alternative is to write the changes to the subcategory/GeoJSON and generate new graphics from the persistent representation. This would not meet the requirements for an efficient, interactive preview mode to tweak the changes before saving them.

**Additional context**
Here, "line color" refers to the color of:
- curves and line strings,
- stick text, and
- the outlines of planar regions.

It does not apply to the edges of surfaces other than filled planar regions. e.g., a sphere's edges are the same color as its surface.

pmconne:
Fixes #7244.

Added to FeatureAppearance `lineRgb` and `lineTransparency` with the following possible values:
- `undefined`: apply `rgb` and/or `lineTransparency` to linear geometry, as before;
- `false`: do not override color and/or transparency for linear geometry; or
- a color and/or transparency value to apply specifically to linear geometry.

Added two new `OvrFlags` for these new overrides, and another RGBA value to each entry in `FeatureOverrides` texture.

Adjusted the vertex shader to choose which color to apply based on the type of geometry, as specified by `u_renderOrder`.

display-test-app's UI got a little clunkier as a result:
image

TODO:
- [x] NextVersion.md
- [x] Run ImageTests
- [x] Unit tests


common/api/core-common.api.md
```diff
@@ -3169,6 +3173,8 @@ export interface FeatureAppearanceProps {
     emphasized?: true;
     ignoresMaterial?: true;
     linePixels?: LinePixels;
+    lineRgb?: RgbColorProps | false;
```
aruniverse, line 3176: do we really need to type this as false?
wouldnt it being undefined be enough to do whatever you have for the "false" behavior?

pmconne: Did you read the documentation comments?

markschlosseratbentley (approved):

common/api/core-common.api.md
aruniverse: sorry i did not earlier. i now see this is the best way to not break users and be more flexible

hl662 (approved):

ben-polinsky (approved):

aruniverse (approved):

PR #7261, opened 2024-10-14
Merging presentation formats on kind of quantities

AureMonke:


common/api/ecschema-editing.api.md
```diff
@@ -751,7 +751,7 @@ export enum DiagnosticType {
 }
 
 // @beta (undocumented)
-export function diagnosticTypeToString(type: DiagnosticType): "CustomAttributeContainer" | "None" | "Property" | "RelationshipConstraint" | "Schema" | "SchemaItem";
```
StefanApfel-Bentley, line 754: That is often an indication, that your local builds (aside of ecschema-editing) is out of date. Probably do a TMR, but to have extract-api check pass, the strings always need to be in alphabetical order

common/api/ecschema-editing.api.md
```diff
@@ -95,7 +95,7 @@ export type AnySchemaDifference = SchemaDifference | SchemaReferenceDifference |
 export type AnySchemaEdits = SkipEdit | RenameSchemaItemEdit | RenamePropertyEdit;
 
 // @alpha
-export type AnySchemaItemDifference = AnyClassItemDifference | ConstantDifference | EnumerationDifference | EntityClassMixinDifference | FormatDifference | KindOfQuantityDifference | InvertedUnitDifference | PhenomenonDifference | PropertyCategoryDifference | UnitDifference | UnitSystemDifference;
+export type AnySchemaItemDifference = AnyClassItemDifference | ConstantDifference | EnumerationDifference | EntityClassMixinDifference | FormatDifference | KindOfQuantityDifference | KoqPresentationFormatDifference | InvertedUnitDifference | PhenomenonDifference | PropertyCategoryDifference | UnitDifference | UnitSystemDifference;
```
StefanApfel-Bentley, line 98: KoqPresentationFormatDifference shouldn't be an AnySchemaItemDifference, rather put it in the AnySchemaDifference union like EntityClassMixinDifference which moved there with a recent fix too

common/api/ecschema-editing.api.md
```diff
@@ -751,7 +751,7 @@ export enum DiagnosticType {
 }
 
 // @beta (undocumented)
-export function diagnosticTypeToString(type: DiagnosticType): "CustomAttributeContainer" | "None" | "Property" | "RelationshipConstraint" | "Schema" | "SchemaItem";
+export function diagnosticTypeToString(type: DiagnosticType): "Property" | "Schema" | "SchemaItem" | "RelationshipConstraint" | "CustomAttributeContainer" | "None";
```
StefanApfel-Bentley, line 754: That is often an indication, that your local builds (aside of ecschema-editing) is out of date. Probably do a TMR, but to have extract-api check pass, the strings always need to be in alphabetical order.

core/ecschema-metadata/src/Metadata/KindOfQuantity.ts
```diff
@@ -141,7 +141,7 @@ export class KindOfQuantity extends SchemaItem {
     return new OverrideFormat(parent, precision, unitLabelOverrides);
   }
 
-  private async processPresentationUnits(presentationUnitsJson: string | string[]): Promise<void> {
+  protected async processPresentationUnits(presentationUnitsJson: string | string[]): Promise<void> {
```
StefanApfel-Bentley, line 144: I'm not sure about that one. We already have a public method addPresentationFormat and I think we should rather use that one instead of extending our public api. We might need to discuss that further.

core/ecschema-editing/src/Differencing/SchemaDiagnosticVisitor.ts
```diff
@@ -449,6 +452,27 @@ export class SchemaDiagnosticVisitor {
     }
     return;
   }
+
+  private visitMissingPresentationUnit(diagnostic: AnyDiagnostic) {
+    const koq = diagnostic.ecDefinition as KindOfQuantity;
+    const [presentationFormat] = diagnostic.messageArgs as [Format | OverrideFormat];
+
+    let modifyEntry = this.schemaItemDifferences.find((entry): entry is KindOfQuantityPresentationFormatDifference => {
```
StefanApfel-Bentley, line 460: that should actually be an addEntry. I saw it's also wrong for the mixins on classes and relationship constrains. Looks like that slipped through when we changed the change type from modify to add there.

core/ecschema-editing/src/Merging/KindOfQuantityMerger.ts
```diff
@@ -60,6 +60,17 @@ export async function modifyKindOfQuantity(context: SchemaMergeContext, change:
...
 }
+/**
+ * Merges a new presentation format into the target kind of quantity
+ * @internal
+*/
+export async function addPresentationFormat(context: SchemaMergeContext, change: KindOfQuantityPresentationFormatDifference) {
+  for (const presentationFormat of change.difference) {
+    const koqKey = new SchemaItemKey(change.itemName, context.targetSchemaKey);
+    const formatString = await updateOverrideFormat(context, presentationFormat);
+    await context.editor.kindOfQuantities.addPresentationFormatString(koqKey, formatString);
```
StefanApfel-Bentley, line 71: I think, the better approach would be to parse the format string using the KindOfQuantities parseFormatString method (which should be probably moving as a static method to OverrideFormat) and then use the existing editor method addPresentationOverrideFormat instead of adding another addPresentationFormat editor method.

christophermlawson (approved):

PR #7262, opened 2024-10-14
Add Condition to bump job

DanRod1999:
Recent change to this yaml file accidentally changed some of the job conditions. This caused nightly releases to be produced even if no changes were committed to itwinjs-core in that day.

Adding the condition back to the the Bump Job will skip the bump task when no commits into itwinjs-core are detected.

### DIFF
--- a/common/config/azure-pipelines/jobs/version-bump.yaml
+++ b/common/config/azure-pipelines/jobs/version-bump.yaml
@@ -137,8 +137,10 @@ jobs:
 
   - job: Bump
     displayName: Bump Version
-    dependsOn: CreateBranch
-    condition: in(dependencies.CreateBranch.result, 'Succeeded', 'Skipped')
+    dependsOn:
+    - CreateBranch
+    - checkPrevCommit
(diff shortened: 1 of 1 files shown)


hl662 (approved):

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -137,8 +137,10 @@ jobs:
 
   - job: Bump
     displayName: Bump Version
-    dependsOn: CreateBranch
-    condition: in(dependencies.CreateBranch.result, 'Succeeded', 'Skipped')
+    dependsOn:
+    - CreateBranch
+    - checkPrevCommit
```
MichaelSwigerAtBentley, line 142: The Previous Commit job is named "CheckPrevCommit", don't know if dependsOn is case sensitive, but should be consistent regardless. 

https://github.com/iTwin/itwinjs-core/blob/020769b2973cdd7b9d6bfb622ef4b9d34af03e1e/common/config/azure-pipelines/jobs/version-bump.yaml#L43

MichaelSwigerAtBentley (approved):

aruniverse (approved):

PR #7264, opened 2024-10-14
Make classIdsToClassNames explicit parameter independent of UseJsPropertyName

khanaffan:
**Previous behavior**

`UseJsPropertyName` also implicitly mean `ClassIdToClassName` was `true`. But if explicitly specified `ClassIdToClassName: false` we did not return classId(s) instead it will still return class names.

**New behavior**

`UseJsPropertyName` still default to `ClassIdToClassName: true` unless `ClassIdToClassName: false` is specified. If false was specified, then classId(s) are returned.

Instance query when use with option `USE_JS_NAMES` continue to return `UseJsPropertyName: true, ClassIdToClassName: true`.

@soham-bentley added  new test for various permutations of `ECSqlRowFormat`, `classIdToClassNames` & `abbrivateBlob`.

imodel-native: https://github.com/iTwin/imodel-native/pull/883


core/common/src/ECSqlReader.ts
```diff
@@ -342,6 +342,10 @@ export class ECSqlReader implements AsyncIterableIterator<QueryRowProxy> {
       return [];
     }
     const valueFormat = this._options.rowFormat === QueryRowFormat.UseJsPropertyNames ? DbValueFormat.JsNames : DbValueFormat.ECSqlNames;
+    if (this._options.rowFormat === QueryRowFormat.UseJsPropertyNames && typeof this._options.convertClassIdsToClassNames === "undefined") {
```
aruniverse, line 345: can we please deprecate `QueryRowFormat.UseJsPropertyNames` ?

aruniverse: fyi @ColinKerr 
deprecation doesnt have to be done in this specific pr, but this needs to be done asap

rschili (approved):

aruniverse:
does this need to be in a 4.9 patch?

aruniverse:
@ColinKerr can we close this pr in favor of yours, https://github.com/iTwin/itwinjs-core/pull/7315 ?

PR #7268, opened 2024-10-15
Fixed issue with intersection not found at ellipse angle -pi/2

saeeedtorabi:
Addresses https://github.com/iTwin/itwinjs-backlog/issues/1265


dassaf4 (approved):
lgtm, added some doc improvements in light of the math we figured out

PR #7270, opened 2024-10-15
Enable ContentFormatter formatting struct and array type properties with 'undefined' value

JonasDov:


presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
```diff
@@ -73,11 +73,11 @@ export class ContentFormatter {
   }
 
   private async formatPropertyValue(value: Value, field: PropertiesField): Promise<DisplayValue> {
-    if (field.isArrayPropertiesField()) {
+    if (field.isArrayPropertiesField() && value !== undefined) {
```
saskliutas, line 76: I think `value !== undefined` check can be moved to `ContentFormatter.formatValues` function to make sure that we do not try formatting value if it is `undefined`.

saskliutas (changes requested):

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
JonasDov: Moved

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
```diff
@@ -49,6 +49,11 @@ export class ContentFormatter {
         continue;
       }
 
+      // do not add undefined value to display values, it's raw value will be formatted
```
saskliutas, line 52:
```suggestion
      // do not add undefined value to display values
```

saskliutas (approved):

saskliutas:

Issue #6595, opened 2024-04-08
MapLayersWidget - sections of background map are not visible around the imodel


Jake-Screen:
We are utilizing the MapLayersWidget to provide a background map in our studio application. When adding this to application we use a classId of "UnifiedSelectionViewportControl2D". 

Upon loading an iModel from iModel hub, the map appears blank around the area where our model is centered. Upon zooming out, some parts of the map appear, with clipping appearing in different sections of the map. When zoomed far enough out, we can see a circle of area that is not covered in map.

**Expected behavior**
For the map to be visible in the area of our iModel.

**Screenshots**
![map error booster](https://github.com/iTwin/itwinjs-core/assets/108944699/e064abaf-273b-47cc-99a2-9f122810d175)
![map error booster 2](https://github.com/iTwin/itwinjs-core/assets/108944699/404a860b-c455-4931-a2b9-93ab4f5c7cd5)

- OS: Windows
- Browser : Chrome
- iTwin.js Version: 4.4.0-dev.10

Jake-Screen:
@ashar-bentley @FelixGirard

pmconne:
Can you please add to your description steps required for somebody else to reproduce the problem?

ashar-bentley:
The issue happens when the setAllow3dManipulations is set to false

Steps:

1. Open the heatmap sample that has the following view setting set to false:

line 21:  `viewState.setAllow3dManipulations(false);`

https://www.itwinjs.org/sandboxes/iTwinPlatform/Heatmap%20Decorator

2. Zoom out until you start seeing the background map getting clipped:

![image](https://github.com/user-attachments/assets/e7f84305-ecee-4291-be12-7d96fdc94b65)


Now open my fork of the same sample with line 21 modified to set the value to true:

line 21:  `viewState.setAllow3dManipulations(true);`

https://www.itwinjs.org/sandboxes/AsharGhani/HeatMapDecoratorWith3D-On

The issue is no longer reproducible:
![image](https://github.com/user-attachments/assets/a94b52d0-4e1e-4ab4-9784-a6bf092e6eeb)

ashar-bentley:
For some iModels we are seeing a more exacerbated version of this problem, where the problem occurs when fully zoomed in.

pmconne:
@markschlosseratbentley @danieliborra

danieliborra:
We'll take a look at it in our next sprint

MarcNeely:
[...]

PR #7275, opened 2024-10-17
Fix zoom out problem when allow3dManipulations is off and bg map is on

MarcNeely:
This fixes an issue where the background map starts to get culled by the frustum when zooming out if allow3dManipulations is turned off.

It fixes issue [6595](https://github.com/iTwin/itwinjs-core/issues/6595).

ImageTests were run with no visual artifacts.


pmconne (approved):
Good fix, thanks!

PR #7276, opened 2024-10-18
`Matrix3d.factorRigidSkew`

dassaf4:
Add new factorization M = R * G, where R is rigid/orthogonal/rotation and G is everything else (scale/mirror/skew).
* Equivalent to native method `RotMatrix::RotateAndSkewFactors`.

Also:
* Improve doc for several matrix methods
* Remove unnecessary import alias in Matrix3d.test.ts
* Fix flaky test "Point4d.Planes" (HT @GytisCepk)


pmconne (approved):

core/geometry/src/test/geometry3d/Matrix3d.test.ts
```diff
@@ -605,8 +605,51 @@ describe("Matrix3d.Factors", () => {
     expect(ck.getNumErrors()).toBe(0);
   });
 
+  it("FactorRigidSkew", () => {
```
saeeedtorabi, line 608: nice test!

saeeedtorabi (approved):

PR #7280, opened 2024-10-18
GeometryStream builder/iterator wasn't applying correct transform to brep.

bbastings:
Discovered by code to apply a scale factor to geometry stream entries.


pmconne (approved):

dassaf4 (commented):
lgtm

PR #7281, opened 2024-10-19
Introduce vitest to core/quantity

anmolshres98:
Continuation of this effort: #7165


hl662:
can you edit the launch.json debug profile for running core quantity tests to be the same as core bentley and common

core/quantity/src/test/BearingAzimuth.test.ts
```diff
@@ -65,18 +65,18 @@ describe("Bearing format tests:", () => {
 
     const bearingDMS = new Format("BearingDMS");
     await bearingDMS.fromJSON(unitsProvider, bearingDMSJson).catch(() => { });
-    assert.isTrue(bearingDMS.hasUnits);
+    expect(bearingDMS.hasUnits).to.be.true;
```
hl662, line 68: why not toBe(true)? this is nitpicky really,  cuz I switched all my other tests to use jest assertions

rschili (approved):

core/quantity/src/test/BearingAzimuth.test.ts
anmolshres98: no reason really. When I began, I started using `.to.be.true` and just kept using that for everything to keep things consistent

hl662: should we decide on a standard? toBe(true) is a JestAssertion, .to.be.true is chai. Vitest will support both ways, because it has a peerDep on chai and exports chai assertions, but it won't come with async support out of the box. So we would have to handle promises by adding chaiAsPromised explicitly

core/quantity/src/test/Formatting.test.ts
```diff
@@ -276,7 +287,7 @@ describe("Formatting tests:", () => {
     const testFormat = new Format("test");
 
     await testFormat.fromJSON(unitsProvider, json).catch(() => {
-      assert.isTrue(false);
+      expect.fail();
     });
```
hl662, line 291:
```suggestion
    await expect(testFormat.fromJSON(unitsProvider, json)).resolves.not.toThrow();
```

hl662: if you think this looks good, you can replace other similar instances in this file

core/quantity/src/test/BearingAzimuth.test.ts
anmolshres98: we should decide on a standard for consistency's sake. Should we have a separate PR that does this everywhere in the code? I am in favor of whatever will lead to less deps/code. However, how would we enforce this? Maybe a custom lint rule?

core/quantity/src/test/Formatting.test.ts
anmolshres98: I do prefer that. Updated thanks 😄

anmolshres98:
> can you edit the launch.json debug profile for running core quantity tests to be the same as core bentley and common

done

core/quantity/src/test/BearingAzimuth.test.ts
hl662: i straight up did this to core frontend, david did so for core geometry individually, but you're right that enforcing it would be better. I have no idea how to enforce this though :(

hl662: After discussion with Rob, there's no need to enforce this. It's a recommendation to use jest assertions instead of chai, in cases where people want to drop chaiAsPromised usage

hl662 (approved):

eringram:
My only feedback is that in some places you go from [assert.strictEqual](https://nodejs.org/api/assert.html#assertstrictequalactual-expected-message) to vitest's toEqual, when technically they are slightly different including that strictEqual asserts the types are identical and toEqual doesn't ([toStrictEqual ](https://vitest.dev/api/expect.html#tostrictequal)would be the equivalent). If the tests pass I don't think this is worth changing though

eringram (approved):

hl662 (approved):
Picked up this PR again, previous `unhandled rejection` errors are gone, but if it re appears in some way, we can handle it now utilizing `process.on('unhandledRejection', ...)`. Approving this in the meantime

core/quantity/src/test/BearingAzimuth.test.ts
```diff
@@ -2,7 +2,7 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
-import { assert, expect } from "chai";
+import { beforeEach, describe, expect, it } from "vitest";
```
ben-polinsky, line 5: Any reason for the `assert` to `expect` change?

ColinKerr (approved):

core/quantity/src/test/BearingAzimuth.test.ts
hl662: less imports - in vitest, chai and jest both have test assertions that use `expect`, and we can move towards using one syntax. I see a lot more usage of `expect` than `assert` in core-quantity in particular

ben-polinsky: Just saw @eringram's older comment who noticed the same thing. The PR would be cleaner without the change, but I understand coalescing around one or the other.

hl662: unless a test is explicitly testing that the types match, using `toEqual` is fine for many cases. When a user writes a test that tests types, they'll probably learn the hard way that `toEqual` will not do the job and can discover `toStrictEqual`.

aruniverse (approved):

PR #7284, opened 2024-10-21
Empty string property tag fix

christophermlawson:
Any null, undefined, or empty (string) values are considered invalid and exception is thrown. An exception should be made for string properties where the value is empty.

### DIFF
--- a/core/ecschema-metadata/src/test/Deserialization/XmlParser.test.ts
+++ b/core/ecschema-metadata/src/test/Deserialization/XmlParser.test.ts
@@ -1868,6 +1868,61 @@ describe("XmlParser", () => {
     });
 
     describe("Property Parsing Tests", () => {
+      describe("String Parsing Tests", () => {
+        it("With value set to valid string, CustomAttributeProvider should provide valid instance.", async () => {
+          const itemXml = `
+            <ECCustomAttributes>
+              <TestAttribute xmlns="TestSchema.1.0">
+                <TestProperty>TestString</TestProperty>
+              </TestAttribute>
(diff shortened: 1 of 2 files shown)


core/ecschema-metadata/src/test/Deserialization/XmlParser.test.ts
```diff
@@ -1868,6 +1868,61 @@ describe("XmlParser", () => {
     });
 
     describe("Property Parsing Tests", () => {
+      describe("String Parsing Tests", () => {
+        it("With value set to valid string, CustomAttributeProvider should provide valid instance.", async () => {
+          const itemXml = `
+            <ECCustomAttributes>
+              <TestAttribute xmlns="TestSchema.1.0">
```
rschili, line 1875: Is this an older schema? I think we require 3 digits in EC 3.2 schemas, and 2 digits in EC 3.1 and earlier

christophermlawson: All the XmlParser tests have this.  Should I update all of them?

ColinKerr: xmlns should include all three digits of the schema version.  At least update this test so we don't make the world worse even if we're not making it better.

christophermlawson: I updated all of them.

rschili (approved):

StefanApfel-Bentley (approved):

Issue #7245, opened 2024-10-08
AccuSnap Hints on SectionDrawings


Ellord207:
**Describe the bug**

While viewing a section drawing, Tools with AccuSnap does not give hints for geometry coming from the spatial view.

**To Reproduce**

1. Create and load a ViewStete of SectionDrawing
2. Start a tool with AccuSnap (such as the measure tool)
3. Observe AccuSnap hints on geometry not in the spatial view
4. Observe NO AccuSnap hints on geometry from the Spatial View.

You can reference the SectionDrawings and ViewDefintions in this [iModel](https://qa-developer.bentley.com/my-itwins/8d6e2269-0816-4b63-a404-0ddd996497cc/83c0afdf-ccf7-4782-8b0d-0b7680d8ee5b/home/).  If requested, I can provide access or a local snapshot.  I have confirm this bug is Observable in DTA with the ViewDefinitions `section-with-annotations` and `test-section-drawing` in this iModel.

**Expected behavior**
I would expect AccuSnap to provide hints on the geometry from the spatial view.

**Screenshots**
Here is a video it happening.  In the video, the view is a sheet with an attached SectionDrawing.  The result is the same if you just view the section drawing directly.

https://github.com/user-attachments/assets/bcf4e1a5-2114-4081-9f46-b89b48a94fca

**Desktop (please complete the applicable information):**
- iTwin.js Version 4.9.0 (not specific to this version)

**Additional context**
I am able to Select the geometry with the SelectTool, just no AccuSnap Hints.

pmconne:
#5633 added support for snapping to geometry inside ViewAttachments in sheets. It did not do anything to support snapping to geometry inside a section drawing's attached spatial view. The solution would presumably be very similar.

Ellord207:
> [@Ellord207](https://github.com/Ellord207) can you please describe what specifically you are trying to accomplish? Like, are you trying to draw a leader between a dimension and the dimensioned element? Are you trying to measure distances between things in the sheet and in the spatial view or drawing? etc.

We expect the users will need snapping to:
1. Drawing leaders for dimensions, bubble properties, and other such annotations inside the SectionDrawing
2. Snap and offset annotations from existing annotations for a clean layout.
3. Measuring distances of elements inside the SectionDrawing
4. Drawing annotations and markup on a Sheet with one or more attached views of SectionDrawings
[...]

PR #7291, opened 2024-10-22
Snap inside section drawings on sheets

pmconne:
More completely addresses #7245, reverting previous attempt in #7267.

With reluctance, introduce the concept of `HitPath` to `HitDetail`. It is not a generic series of arbitrary reference transforms like in MicroStation; instead it can be one of the following:
- Sheet > ViewAttachment
- SectionDrawing > SpatialView
- Sheet > ViewAttachment (to SectionDrawing) > SpatialView


bbastings (approved):
Looks reasonable. Were you planning to address the TODOs in the computeDisplayTransform implementations, or defer that?

PR #7294, opened 2024-10-24
Add support for inverted units and refactor SchemaUnitsProvider

rschili:
SchemaUnitsProvider was not properly wired up to support Inverted Units which have recently been added to the quantity package.

While at it I made a few more internal changes to that class.
When failing to obtain a Unit, it will now return a BadUnit, as intended by the quantity package instead of throwing.

Left a few TODO comments in the class for known issues (unrelated to this PR)


ColinKerr (approved):

core/ecschema-metadata/src/UnitProvider/SchemaUnitProvider.ts
```diff
@@ -149,55 +168,72 @@ export class SchemaUnitProvider implements UnitsProvider {
...
-    const item = await this._context.getSchemaItem<Unit>(itemKey);
-    if (!item)
-      throw new BentleyError(BentleyStatus.ERROR, "Cannot find schema item/unit", () => {
-        return { item: schemaItemName, schema: schemaName };
-      });
-
-    if (item.schemaItemType === SchemaItemType.Unit)
-      return item;
+    const invertedUnit = await context.getSchemaItem<InvertedUnit>(itemKey);
+    if (invertedUnit && InvertedUnit.isInvertedUnit(invertedUnit) && invertedUnit.invertsUnit) {
```
hl662, line 206: Hmm, could the 2 latter checks here not be reduced into 1?

hl662: Actually, can an InvertedUnit schema item exist, but the unit it is inverting does not exist? I guess that can explain the need for the 3rd check, but if not, we could instead check invertedUnit.invertsUnit in this new `isInvertedUnit()` method no?

hl662: the comment does help yep. This should be fine as is then

hl662 (approved):

PR #7298, opened 2024-10-28
Introduce vitest to core/ecschema-metadata

rschili:
> If unfamiliar with debugging using vitest, take a quick look at our contributing.md section on [debugging](https://github.com/iTwin/itwinjs-core/blob/master/CONTRIBUTING.md#filtering-test-suites)

This pull request migrates the `core/ecschema-metadata` package from Mocha/Chai/Sinon to Vitest for testing, updates test assertions to use Vitest syntax, and removes related legacy dependencies. It also updates test scripts and coverage commands. These changes modernize the test suite, simplify dependency management, and ensure better compatibility with current tooling.

**Test framework migration and dependency cleanup:**
* Replaced Mocha, Chai, and Sinon with Vitest for all unit tests, updating imports and assertion syntax throughout `src/test` files.
* Removed all related legacy dependencies (`chai` related) from `package.json` and added Vitest and coverage plugins.

**Test script and configuration updates:**
* Changed `test` and `cover` scripts in `package.json` to use Vitest
* Updated VSCode launch configuration for ecschema-metadata to use `pnpm`

**Test code modernization:**
* Updated all test assertions from Chai to Vitest syntax, including promise rejection and equality checks, and replaced deprecated hooks with Vitest equivalents. 
* Since Vitest doesn't have a 1-1 rejection assertion similar to chai to asynchronously assert a subclassed Error alongside the error message, added a `expectAsyncToThrow` helper function inside `ecschema-metadata/test/TestUtils/AssertionHelpers.ts` to facilitate, alongside a `expectToThrow` for consistency.
* Added `test/setupTests.ts` that imports the barrel file of ecschema-metadata, to avoid circular import errors.

### DIFF
--- a/core/ecschema-metadata/src/test/Metadata/Phenomenon.test.ts
+++ b/core/ecschema-metadata/src/test/Metadata/Phenomenon.test.ts
@@ -3,7 +3,7 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
-import { assert, beforeEach, describe, expect, it } from "vitest";
+import { beforeEach, describe, expect, it } from "vitest";
 import { SchemaContext } from "../../Context";
 import { Phenomenon } from "../../Metadata/Phenomenon";
 import { Schema } from "../../Metadata/Schema";


core/ecschema-metadata/src/test/Context/SchemaCache.test.ts
```diff
@@ -26,16 +26,20 @@ describe("Schema Cache", () => {
     await cache.addSchema(schema1);
 
     const schema2 = new Schema(context, new SchemaKey("TestSchema"), "ts");
-    await expect(cache.addSchema(schema2)).to.be.rejectedWith(ECObjectsError, "The schema, TestSchema.00.00.00, already exists within this cache.");
+    await expect(cache.addSchema(schema2)).rejects.toThrow(ECObjectsError);
+    await expect(cache.addSchema(schema2)).rejects.toThrow("The schema, TestSchema.00.00.00, already exists within this cache.");
```
rschili, line 30: @hl662 is this the way to do it? This will call addSchema twice just to verify the message of the Error?

rschili:
giving to stefan. I was going to dedicate some time towards this frequently, but it just didn't happen. It probably needs a lot of updating, but is better than starting from scratch.

core/ecschema-metadata/src/test/Context/SchemaCache.test.ts
hl662: I added a helper function called `expectAsyncToThrow` that'll help test both the error obj and the message with one call

StefanApfel-Bentley (approved):

PR #7306, opened 2024-10-30
iOS tests fix

tcobbs-bentley:
iOS tests fix:
* Reenable iOS backend unit tests in CI
* Copy test output XML instead of extracting from console log
* Show error output if present when running display-test-app

Hopefully this will fix the CI test failures for the backend unit tests. Unfortunately, I have no idea what was causing the failures, but I no longer execute the code that appeared to be failing, so hopefully that fixes the problem. (Specifically, instead of printing the XML file contents to the console during execution, this now prints the XML filename, and that file is then copied to the appropriate place by the script running the iOS Simulator.)

### DIFF
--- a/tools/internal/ios/core-test-runner/core-test-runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/tools/internal/ios/core-test-runner/core-test-runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/iTwin/mobile-native-ios",
       "state" : {
-        "revision" : "b365b0ca9fe2b16cddaf26f59e8f38148027608c",
-        "version" : "4.8.11"
+        "revision" : "c011577eed1b76b408a7d997be918b41380fb836",
+        "version" : "4.10.21"
       }
     }
   ],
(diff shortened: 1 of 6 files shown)


tools/internal/ios/core-test-runner/core-test-runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/iTwin/mobile-native-ios",
       "state" : {
-        "revision" : "b365b0ca9fe2b16cddaf26f59e8f38148027608c",
-        "version" : "4.8.11"
+        "revision" : "c011577eed1b76b408a7d997be918b41380fb836",
```
aruniverse, line 8: curious, where does this hash come from? we have scripts to update addon version across the monorepo, should we automate this too, or rather could we?

aruniverse (approved):

tools/internal/ios/core-test-runner/core-test-runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
tcobbs-bentley: The hash is normally populated by Xcode itself when it updates this file (as my Xcode did). I'm not entirely sure where it comes from. Having said that, 4.8.11 was wrong for the master branch, so I included this change in my PR.

aruniverse:

markschlosseratbentley (approved):
LGTM

PR #7310, opened 2024-10-31
Fix reading of KHR compressed meshes

MarcNeely:
Fixes the reading of KHR compressed meshes.  This makes it possible to read in the meshes that we export with the cesium option which were in many cases not scaling properly when attached back into the original iModel.

Partially fixes issue [#420](https://github.com/iTwin/itwin-graphics-backlog/issues/420)

### DIFF
--- a/core/frontend/src/tile/GltfReader.ts
+++ b/core/frontend/src/tile/GltfReader.ts
@@ -1342,9 +1342,9 @@ export abstract class GltfReader {
           // Assume KHR_mesh_quantization...
           min = [0, 0, 0];
           if (GltfDataType.UnsignedShort === posView.type)
-            max = [65535, 65535, 65535];
+            max = [0xFFFF, 0xFFFF, 0xFFFF];
           else
-            max = [255, 255, 255];
+            max = [0xFF, 0xFF, 0xFF];
           }
 
         if (undefined === min || undefined === max) {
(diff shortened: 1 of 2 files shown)


core/frontend/src/tile/GltfReader.ts
```diff
@@ -1340,9 +1340,9 @@ export abstract class GltfReader {
           max = ext.decodedMax;
         } else {
           // Assume KHR_mesh_quantization...
-          min = posView.accessor.min;
-          max = posView.accessor.max;
-        }
+          min = [0, 0, 0];
+          max = [65535, 65535, 65535];
```
aruniverse, line 1344: where doess this value come from? should this be a constant, or should there be a comment for this?

MarcNeely: It's just the maximum value of an unsigned short.

pmconne (approved):

core/frontend/src/tile/GltfReader.ts
MarcNeely: Ahh, I didn't see that that could be an unsigned byte as well.  I'll push a fix.

MarcNeely: Fix pushed.  Or is there some defined constant for these values that I should use?

MarcNeely: OK, I changed them to 0xFFFF and 0xFF which is what I see being used around in other places.

markschlosseratbentley (approved):
Thanks for the fix!

PR #7319, opened 2024-11-04
deprecate PresentationRpcInterface

paulius-valiunas:
We are planning to deprecate RPC altogether in the future, and while we don't have a replacement just yet, we can at least encourage people to use various wrappers instead of calling RPC interfaces directly. This way the transition to new APIs will be less disruptive. Presentation is the only RPC interface where all the methods are public and not internal.


hl662:
Maybe add a section in NextVersion.md saying the interface is deprecated

paulius-valiunas:
> should the PresentationRpcInterface class itself also be marked as deprecated?

people still need a way to register it in their frontend/backend, so I can't just deprecate it... we would need to provide an alternative registration path, but that would be a big refactor and I think it's fine to just keep things this way for now. Once we move all the existing APIs, we can then deprecate `RPCManager` and related classes, as well as the RPC interfaces themselves. But at that point, we will probably be close to 6.0 where we can actually remove all the deprecated methods, so there wouldn't be any actual code in the deprecated interfaces.

docs/changehistory/NextVersion.md
```diff
@@ -175,3 +175,7 @@ In addition to [already supported Electron versions](../learning/SupportedPlatfo
 - Refactored `FormatType`, `ScientificType`, `ShowSignOption` from int enums to string enums and added `RatioType` as a string enum. Relevant toString functions, including [formatTypeToString]($quantity), [scientificTypeToString]($quantity), and [showSignOptionToString]($quantity), have been deprecated because they don't need serialization methods.
 
 - [Parser.parseToQuantityValue]($quantity) have been deprecated. Use the existing method [Parser.parseQuantityString]($quantity) instead.
+
+### @itwin/presentation-common
+
+- All public methods of [PresentationRpcInterface]($presentation-common) have been deprecated. Going forward, RPC interfaces should not be called directly. Public wrappers such as [PresentationManager]$(presentation-backend) should be used instead.
```
grigasp, line 181:
```suggestion
- All public methods of [PresentationRpcInterface]($presentation-common) have been deprecated. Going forward, RPC interfaces should not be called directly. Public wrappers such as [PresentationManager]$(presentation-frontend) should be used instead.
```

common/changes/@itwin/presentation-common/deprecate-presentation-rpc_2024-11-04-15-22.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-common",
+      "comment": "deprecate PresentationRpcInterface",
+      "type": "none"
+    }
```
grigasp, line 7:
```suggestion
    {
      "packageName": "@itwin/presentation-common",
      "comment": "All public methods of `PresentationRpcInterface` have been deprecated. Going forward, RPC interfaces should not be called directly. Public wrappers such as `PresentationManager` should be used instead.",
      "type": "minor"
    }
```

presentation/common/src/presentation-common/PresentationRpcInterface.ts
```diff
@@ -197,38 +197,45 @@ export class PresentationRpcInterface extends RpcInterface {
     NOTE: Please consult the README in core/common/src/rpc for the semantic versioning rules.
   ===========================================================================================*/
 
+  /** @deprecated in 4.10. Use [PresentationManager]($presentation-backend) instead of calling the RPC interface directly. */
```
grigasp, line 200: `master` is now pointing to `5.0` - are you going to rebase this branch to `release/4.10.x`?

presentation/common/src/presentation-common/PresentationRpcInterface.ts
```diff
@@ -197,38 +197,45 @@ export class PresentationRpcInterface extends RpcInterface {
...
   public async getNodesCount(_token: IModelRpcProps, _options: HierarchyRpcRequestOptions): PresentationRpcResponse<number> {
     return this.forward(arguments);
   }
 
+  /** @deprecated in 4.10. Use [PresentationManager]($presentation-backend) instead of calling the RPC interface directly. */
```
grigasp, line 205:
```suggestion
  /** @deprecated in 4.10. Use [PresentationManager]($presentation-frontend) instead of calling the RPC interface directly. */
```

This suggestion applies to all similar deprecation messages in this file.

common/changes/@itwin/presentation-common/deprecate-presentation-rpc_2024-11-04-15-22.json
grigasp: I'm not totally sure about `minor`, but the deprecation should definitely be mentioned in the changelog. Chose `minor` over `patch`, because consumers will start getting deprecation warnings, if they're using these apis.

paulius-valiunas: Does the type even matter here? The version will be bumped in lock step anyway

presentation/common/src/presentation-common/PresentationRpcInterface.ts
paulius-valiunas: fixed

common/changes/@itwin/presentation-common/deprecate-presentation-rpc_2024-11-04-15-22.json
grigasp: Ah, right. As long as it's not "none", it probably doesn't matter.

grigasp (approved):

paulius-valiunas:

paulius-valiunas:

PR #7340, opened 2024-11-08
Use vitest v3, Resolve GHSA-9crc-q9x8-ghqq 

hl662:
- Resolve https://github.com/advisories/GHSA-9crc-q9x8-hgqq 

- Consume vitest 3.0.5 (and adjacent vitest packages like `browser`, `coverage-v8`. 
- All vitest deps follow the same version number, to avoid dependency mismatch warnings when running tests.
- Bump core-geometry's global testTimeout to a minute - there's no slowdown to the CI times or locally, since the tests are running in parallel.
- Drop coverage reporting in core-frontend until the dynamic imports issue can be resolved (Opened [issue](https://github.com/iTwin/itwinjs-core/issues/7711)). 
- Import core-frontend's barrel file before running any core-frontend test suite.
- Enable parallel tests in core-frontend again. (This helps counteract the performance slowdown importing the barrel file above will have)


aruniverse (approved):
Why didn't rush catch this?

hl662:
> Why didn't rush catch this?

The mismatched dependencies were in the pnpm lockfile. Some were '2.1.1', while others were '2.1.0'. I guess rush doesn't catch mismatched deps that are in lockfiles

hl662:
setting back to draft for now, circling back to fix regression for vitest/browser at a later point

hl662:
Marked as active again - I had to update the setup file for core frontend to import the barrel file. Vitest now [inlines](https://github.com/vitest-dev/vitest/pull/6689) setup files into test files , messing up dependency resolution in test files and reintroducing circular imports otherwise

aruniverse (approved):

core/frontend/src/test/setupTests.ts
```diff
@@ -1,7 +1,6 @@
 // By importing a barrel file within a setup file, we would be disabling vitest's ability to mock modules. But it's the easiest way to avoid circular import runtime errors within vitest.
 // Link to a section covering this issue: https://vitest.dev/guide/common-errors.html#cannot-mock-mocked-file-js-because-it-is-already-loaded
 // Not importing all of core frontend, as it's not needed. Only MockRender namespace is causing circular import runtime errors.
-import "../internal/render/MockRender";
-
+import "../core-frontend";
```
ben-polinsky, line 4: The comment above states "not importing all of core frontend" yet...

core/geometry/src/test/SimpleFactory.ts
```diff
@@ -213,7 +213,7 @@ export class SimpleFactory {
         return g.Matrix4d.createIdentity();
       case 2:
         return g.Matrix4d.createTranslationAndScaleXYZ(1, 3, 2, 3, 2, 4);
-      case 2:
+      case 3:
```
hl662, line 216: @saeeedtorabi @dassaf4 heads up, vitest emitted a warning that this case was previously unreachable, so I updated it. No existing test failed

dassaf4: Nice, wtg vitest!

core/frontend/src/test/setupTests.ts
hl662: okay, updated comment, we have to import all of core-frontend

dassaf4 (approved):
core-geometry looks good

hl662:
@Mergifyio backport release/4.10.x release/4.11.x

PR #7342, opened 2024-11-08
Investigate potential schedule script bug

pmconne:
Don't review. I'm trying (and thus far failing) to reproduct the issue described in #7330


pmconne:
Problem was fixed sometime between 3.7.5 and 4.10.0. Advised user to update.

PR #7345, opened 2024-11-11
Presentation: Fix enum values formatting

MartynasStrazdas:
closes https://github.com/iTwin/presentation/issues/763


presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
```diff
@@ -120,21 +120,26 @@ export class ContentPropertyValueFormatter {
         }
       : async (rawValue: number) => formatDouble(rawValue);
 
-    return this.formatValue(field.type, value, { doubleFormatter });
+    return this.formatValue(field.type, value, field, { doubleFormatter });
```
grigasp, line 123: doesn't look like we need to pass `field.type` anymore, since we're now passing `field`

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
```diff
@@ -168,28 +173,31 @@ export class ContentPropertyValueFormatter {
...
-  private async formatStructValue(type: StructTypeDescription, value: Value) {
+  private async formatStructValue(type: StructTypeDescription, value: Value, field: Field) {
     if (!Value.isMap(value)) {
       return {};
     }
 
     const formattedMember: DisplayValuesMap = {};
     for (const member of type.members) {
-      formattedMember[member.name] = await this.formatValue(member.type, value[member.name]);
+      formattedMember[member.name] = await this.formatValue(member.type, value[member.name], field);
```
grigasp, line 190: Here you're formatting a struct member, but passing struct field - that's not going to work. I think you should ensure `field` is `StructPropertiesField` and find appropriate member field to pass to `formatValue`.

Same problem with array items below.  Please add tests for struct members and array items.

grigasp (changes requested):

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
MartynasStrazdas: Removed type where type is generic.

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
MartynasStrazdas: Changed the workflow for struct and array formatting, still need to adjust a few test.

MartynasStrazdas: Fixed

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
```diff
@@ -120,76 +120,89 @@ export class ContentPropertyValueFormatter {
...
       return formatDoubleValue(value);
     }
-    if (type.typeName === "navigation") {
+    if (field.type.typeName === "navigation") {
       assert(Value.isNavigationValue(value));
       return value.label.displayValue;
     }
 
+    if (field.type.typeName === "enum" && field.isPropertiesField()) {
+      return field.properties[0].property.enumerationInfo?.choices.find(({ value: enumValue }) => enumValue === value)?.label;
```
grigasp, line 172: Enumerations may have `isStrict = false`, which should allow them to receive values other than the ones in `choices` list. So if the enum is not strict and we don't find a choice, we should return raw value.

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
```diff
@@ -120,76 +120,89 @@ export class ContentPropertyValueFormatter {
...
+    }
     // eslint-disable-next-line @typescript-eslint/no-base-to-string
     return value.toString();
   }
 
-  private async formatStructValue(type: StructTypeDescription, value: Value) {
-    if (!Value.isMap(value)) {
+  private async formatStructValue(type: StructTypeDescription, value: Value, field: Field) {
```
grigasp, line 178: I think we could take `StructPropertiesField` here

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
```diff
@@ -120,76 +120,89 @@ export class ContentPropertyValueFormatter {
...
     for (const member of type.members) {
-      formattedMember[member.name] = await this.formatValue(member.type, value[member.name]);
+      const fieldMember = field.memberFields.find((memberField) => {
+        return memberField.name === member.name;
+      });
+
+      if (fieldMember !== undefined) {
+        formattedMember[member.name] = await this.formatValue(fieldMember, value[member.name]);
+      }
     }
```
grigasp, line 192: Looks like we could iterate over `field.memberFields` and in that case we wouldn't even need the `type` argument

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
```diff
@@ -120,76 +120,89 @@ export class ContentPropertyValueFormatter {
...
     return formattedMember;
   }
 
-  private async formatArrayValue(type: ArrayTypeDescription, value: Value) {
-    if (!Value.isArray(value)) {
+  private async formatArrayValue(_type: ArrayTypeDescription, value: Value, field: Field) {
```
grigasp, line 196: I think here we could take `ArrayPropertiesField` and remove the `_type` argument.

MartynasStrazdas: Removed

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
MartynasStrazdas: Changed

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
MartynasStrazdas: Changed

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
```diff
@@ -120,76 +119,92 @@ export class ContentPropertyValueFormatter {
...
-  private async formatValue(type: TypeDescription, value: Value, ctx?: { doubleFormatter: (raw: number) => Promise<string> }): Promise<DisplayValue> {
-    switch (type.valueFormat) {
-      case PropertyValueFormat.Primitive:
-        return this.formatPrimitiveValue(type, value, ctx);
-      case PropertyValueFormat.Array:
-        return this.formatArrayValue(type, value);
-      case PropertyValueFormat.Struct:
-        return this.formatStructValue(type, value);
+  private async formatValue(field: Field, value: Value, ctx?: { doubleFormatter: (raw: number) => Promise<string> }): Promise<DisplayValue> {
+    if (!field.isPropertiesField()) {
```
grigasp, line 126: This could break calculated properties - they're not based on specific property, but they are primitive and have type information. I think `formatPrimitiveValue` should take `Field` rather than `PropertiesField`.

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
grigasp: could the type of `field` be changed to `ArrayPropertiesField` to avoid runtime checks?

MartynasStrazdas: Sorry file was not saved before commit, changed.

presentation/common/src/presentation-common/content/PropertyValueFormatter.ts
MartynasStrazdas: In this case `formatValue` should call `formatPrimitiveValue` when field is not propertyField instead of returning undefined right?

grigasp: yes, looks good now

grigasp (approved):

grigasp:

grigasp:

aruniverse:

Issue #6145, opened 2023-10-25
It is difficult to locate the standard schemas delivered as part of the core-backend package


ColinKerr:
The schemas can be found in the native assets directory as specified here: https://www.itwinjs.org/reference/core-backend/imodelhost/knownlocations/  

The user then needs to know that schemas are in ECSchemas/Standard sub directory.

Consider making a locater that can find these schemas and can be added to any context.
Consider also having a locater that can locate the delivered domain schemas (like BisCore).  This may not be directly possible due to the fact that core-backend does not depend on ecschema-metadata and ecschema-locaters should not depend on core-backend.
> BisCore and other domain schemas return their path via a property on the class for that schema (e.g. BisCoreSchema).  So long as it's easy to load them given the full paths that may be enough

Locating these schemas should be opt in or lowest priority so any location specified by the user takes precedence.

PR #7347, opened 2024-11-12
New schema locater to locate the standard schemas delivered as part of the core-backend package

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-core/issues/6145


core/ecschema-locaters/src/SchemaJsonFileLocater.ts
```diff
@@ -116,3 +116,80 @@ export class SchemaJsonFileLocater extends SchemaFileLocater implements ISchemaL
...
   }
 }
+
+/**
+ * A SchemaLocator implementation for locating JSON Schema files
+ * from the file system using configurable search paths.
+ * @beta This is a workaround the current lack of a full xml parser.
+ */
+export class PublishedSchemaJsonFileLocater extends SchemaJsonFileLocater implements ISchemaLocater {
+  public static defaultSchemaSearchPaths = new Set([
```
aruniverse, line 126: this seems to be repeated a few times, wonder if we extract this out to a var that is referenced everywhere
what happens today if you update one of these places but not the others?

having a list of these schemas grouped is helpful, but if its not needed you could use a glob to pull in all schemas

RohitPtnkr1996: Done!

ColinKerr:
I don't think this is the right solution to the problem.

This change adds three locaters.  Only one is necessary.  A schema context takes multiple locaters so if a user wants to add paths they can add additional locaters to the context.  

This change also combines ability to load from packages installed in node_modules with the requested ability to load schemas deployed with the backend.  Loading assuming node_modules is fixed has limited use as things may be packaged in a real deployment.

I suggest implementing one locater that takes the backend assets dir string (must be defined) and internally adds paths for the sub directories needed.  It should not accept any other paths from the user.  It should be explicitly documented so the user knows how to construct it and what schemas it will locate.

Locaters added to a context are checked in the order they are added, so the last locater added is the last one checked.  So it is possible to construct a schema context with this locater as the lowest precedence locater but that is easy to mess up.  One possible strategy to improve this would be to follow the pattern from imodel_native where you can add a last chance locater.

It would be better if it 'just worked' without having to take a path or requiring the user to register this locater correctly but that is not possible because core-backend does not have a dependency on any of the ecschema packages.

RohitPtnkr1996:
Okay sounds good ! I will strip down the locater to focus only on the backend schemas.
I was thinking of this as a possible replacement/extension of the existing locater with the added functionality of being able to locater backend schemas.
I added the node-modules search code to be able to load in a wider set of schemas including their json versions as the backend only puts out xml files.

core/ecschema-locaters/src/SchemaXmlFileLocater.ts
```diff
@@ -120,3 +122,43 @@ export class SchemaXmlFileLocater extends SchemaFileLocater implements ISchemaLo
...
+    this.searchPaths.push(...globSync(path.join(knownBackendAssetsDir, "ECSchemas", "*/"), { windowsPathsNoEscape: true }).filter(fs.existsSync));
+  }
+
+  /**
+   * Overrides the addSchemaSearchPath method to prevent adding new schema search paths.
+   *
+   * @param _schemaPath - The schema path to add (ignored).
+   */
+  public override addSchemaSearchPath(_schemaPath: string): void {
+    return; // Do nothing, this is a read-only locator
```
rschili, line 153: Can you elaborate on this? I'd rather actually add to searchPaths, or throw here.

rschili (approved):

core/ecschema-locaters/src/SchemaXmlFileLocater.ts
rschili: Okay, Rohit explained it to me. It still feels odd, but I understand the reasoning. Closing the conversation.

StefanApfel-Bentley (approved):

PR #7352, opened 2024-11-12
Add Search Paramater Re-Insertion to Child Tile URLs in RealityDataSourceTilesetUrlImpl

andremig-bentley:
This PR adds `_searchParams` to the `RealityDataSourceTilesetUrlImpl` class in `RealityDataSourceTilesetUrlImpl.ts`. This paramater is an empty string until `setBaseUrl` is called, at which point, it will be set to the search parameters of the associated url if they exist. The search parameters will then be reinserted into the tile url in `getTileUrl`. 

This is a necessary step when attaching reality data using a url from blob storage, an example being attaching a Cesium export of an iModel from the Mesh Export Service. Without these changes, using `attachRealityModel()` with the url provided by the MES would not work. Requests for child tiles would be unauthorized because the url would not include the search parameters/sas token from the original tileset url. 

We did not think any sort of toggle or flag was necessary and that this behavior could be applied to all urls because we believe that any instance of a url containing search parameters would be an instance that needed those parameters passed on, and in all other instances, `_searchParams` would remain as an empty string.

@eringram @markschlosseratbentley

### DIFF
--- a/core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
+++ b/core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
@@ -110,9 +110,16 @@ export class RealityDataSourceTilesetUrlImpl implements RealityDataSource {
     return true;
   }
 
-  /** Returns the tile URL. If the tile path is a full URL, it is returned as is. Otherwise, the base URL is prepended to the tile path. */
+  /** Returns the tile URL. If the tile path is a full URL, the URL's search parameters are checked.
+   * If the search parameters are empty, the root url's search parameters will be appended to the tile path.
+   * Otherwise, the base URL is prepended to the tile path, and the root url's search parameters are appended.
+   */
   private getTileUrl(tilePath: string){
-    return this.isValidURL(tilePath) ? tilePath : this._baseUrl + tilePath + this._searchParams;
+    if (this.isValidURL(tilePath)) {
+      const url = new URL(tilePath);
+      return url.search === "" ? tilePath + this._searchParams : tilePath;
+    }
+    return this._baseUrl + tilePath + this._searchParams;
   }
 
   /**
--- a/core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
+++ b/core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
@@ -118,9 +118,9 @@ export class RealityDataSourceTilesetUrlImpl implements RealityDataSource {
   private getTileUrl(tilePath: string){
     if (this.isValidURL(tilePath)) {
       const url = new URL(tilePath);
-      return url.search === "" ? tilePath + this._searchParams : tilePath;
+      return url.search === "" ? `${tilePath}${this._searchParams}` : tilePath;
     }
-    return tilePath.includes("?") ? this._baseUrl + tilePath : this._baseUrl + tilePath + this._searchParams;
+    return tilePath.includes("?") ? `${this._baseUrl}${tilePath}` : `${this._baseUrl}${tilePath}${this._searchParams}`;
   }
 
   /**
--- a/core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
+++ b/core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
@@ -110,10 +110,9 @@ export class RealityDataSourceTilesetUrlImpl implements RealityDataSource {
     return true;
   }
 
-  /** Returns the tile URL. If the tile path is a full URL, the URL's search parameters are checked.
-   * If the search parameters are empty, the root url's search parameters will be appended to the tile path.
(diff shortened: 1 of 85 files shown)


eringram (approved):

eringram:
Looks good to me but I'd recommend waiting to merge for another approval too.

Also a note on what the 3D Tiles spec says about this, I found this discussion in the past where apparently query parameters shouldn't be inherited by child tiles according to the specification: https://github.com/CesiumGS/3d-tiles/issues/746. However, in practice, they are inherited when CesiumJS reads 3D Tiles tilesets, and also Google Photorealistic Tiles uses some query params on the root tile similar to the MES

core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
```diff
@@ -107,7 +112,7 @@ export class RealityDataSourceTilesetUrlImpl implements RealityDataSource {
 
   /** Returns the tile URL. If the tile path is a full URL, it is returned as is. Otherwise, the base URL is prepended to the tile path. */
   private getTileUrl(tilePath: string){
-    return this.isValidURL(tilePath) ? tilePath : this._baseUrl + tilePath;
+    return this.isValidURL(tilePath) ? tilePath : this._baseUrl + tilePath + this._searchParams;
```
danieliborra, line 115: Shouldn't it be
```suggestion
    return (this.isValidURL(tilePath) ? tilePath : this._baseUrl + tilePath) + this._searchParams;
```
?

andremig-bentley: I'm not sure if this is correct. The search parameters only get lost above in `setBaseUrl()` when the original url is split to obtain a base url that would be prepended to a relative tile path. I believe this means the search parameters would then only need to be re-inserted in the case of a relative tile path, so the second option in the conditional. The tile path being a full url should mean that it has everything it needs for a successful request, and therefore doesn't need the search parameters appended to them.

markschlosseratbentley: > The tile path being a full url **_should_** mean that it has everything it needs for a successful request, and therefore doesn't need the search parameters appended to them.

(emphasis above is mine)

Has this theory been tested?

andremig-bentley: This has been done in the latest commits

markschlosseratbentley: Looks good, thanks Andre. @danieliborra When you have a chance, could you look and possibly resolve this comment?

core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
```diff
@@ -105,9 +110,17 @@ export class RealityDataSourceTilesetUrlImpl implements RealityDataSource {
     return true;
   }
 
-  /** Returns the tile URL. If the tile path is a full URL, it is returned as is. Otherwise, the base URL is prepended to the tile path. */
+  /** Returns the tile URL. If the tile path is a full URL, the URL's search parameters are checked.
+   * If the search parameters are empty, the root url's search parameters will be appended to the tile path.
+   * Otherwise, the base URL is prepended to the tile path, and the root url's search parameters are appended,
+   * if no search parameters exist on the existing tile path.
+   */
```
eringram, line 117:
```suggestion
  /** Returns the tile URL.
   * If the tile path is a relative URL, the base URL is prepended to it.
   * For both absolute and relative tile path URLs, the search parameters are checked. If the search params are empty, the base URL's search params are appended to the tile path.
   */
```
Suggestion to simplify the description

core/frontend/src/RealityDataSourceTilesetUrlImpl.ts
```diff
@@ -105,9 +110,17 @@ export class RealityDataSourceTilesetUrlImpl implements RealityDataSource {
...
   private getTileUrl(tilePath: string){
-    return this.isValidURL(tilePath) ? tilePath : this._baseUrl + tilePath;
+    if (this.isValidURL(tilePath)) {
+      const url = new URL(tilePath);
+      return url.search === "" ? tilePath + this._searchParams : tilePath;
+    }
+    return tilePath.includes("?") ? this._baseUrl + tilePath : this._baseUrl + tilePath + this._searchParams;
```
aruniverse, line 123: nit:
maybe use templates strings instead of string concat?
feel free to ignore & resolve on your own

andremig-bentley: Agreed, much cleaner. Done in latest commit.

danieliborra (approved):

andremig-bentley:
@aruniverse @markschlosseratbentley @ben-polinsky Can this PR be backported? It's currently blocking integration of the Repositories API into iTwin viewers, and as of discussions with @danieliborra, it's incredibly important to get it through as quickly as possible.

PR #7353, opened 2024-11-13
Build tools: Bump `typedoc` to a version that supports typescript `5.6`

grigasp:
This should allow consumers avoid the following message when installing packages:

```txt
 WARN  Issues with peer dependencies found
apps/build-tools
└─┬ @itwin/build-tools 4.10.0-dev.33
  └─┬ typedoc 0.25.13
    └── ✕ unmet peer typescript@"4.6.x || 4.7.x || 4.8.x || 4.9.x || 5.0.x || 5.1.x || 5.2.x || 5.3.x || 5.4.x": found 5.6.3 in @itwin/build-tools
```

Will need to backport to 4.10 and 4.11.


grigasp:
CI is failing, because all packages' `docs` scripts are passing `--includes=../../generated-docs/extract` command line option to the `betools docs` script. This results in the option being passed down to `typedoc` and `0.26` [removed it](https://typedoc.org/guides/changelog/#v0.26.0-(2024-06-22)). It seems [there are some replacements](https://typedoc.org/guides/documents/) for the option, but I don't understand why that `--includes` option was used in the first place. AFAIK, all the extractions from `generated-docs/extract` are injected into documentation by the doc site build tool rather than `typedoc`. 

I also dug through the history of the `docs` script to see if any commit messages could explain it, but the `--includes` option is there from the initial commit, so that was no help.

@aruniverse @ben-polinsky I'm thinking about removing the option, but not sure if we have any way to tell if the change breaks anything. Any ideas?

JonasDov (approved):

ben-polinsky:
>I'm thinking about removing the option, but not sure if we have any way to tell if the change breaks anything. Any ideas?
Not sure of the history here, but I can test out a local build and ensure extractions are working without the flag. Then we can test in dev.

aruniverse:

ben-polinsky:
Looks good locally!

aruniverse (approved):

ben-polinsky:
Tested on dev and looks good.

PR #7355, opened 2024-11-13
Support SAS URL in test apps

mathieu-lemuzic:
In display-test-app and display-performance-test-app it is possible to provide a tileset URL (IMJS_FRONTEND_TILES_URL_TEMPLATE) to load static tiles via the frontend-tiles extension, however internal logic in these apps prevents using a URL with search parameters, and thus it fails when trying to load a tileset located in an azure container.


pmconne (approved):

markschlosseratbentley (approved):
LGTM

Issue #6751, opened 2024-05-22
Integrate fonts into workspaces


pmconne:
**Is your feature request related to a problem? Please describe.**
We now have the ability to store fonts in workspace dbs but it is entirely unused and it is not integrated with the Workspace APIs.

**Describe the solution you'd like**
- Add "font" as a new type of WorkspaceResource.
- Populate the native font db list from the workspaces
- Support per-iModel font dbs
- Make the native font db list thread-safe
- Define and document collation rules for font names
- Deprecate FontMap
- Create a public font workspace db containing all fonts delivered with MicroStation (similar to our public GCS workspace db).
- Fix tile generator failing to use last resort default fonts when it fails to find a requested font.

**Describe alternatives you've considered**

**Additional context**

kabentley:
> Define and document collation rules for font names

I sure hope that @khanaffan finishes his PR to make COLLATE NOCASE mean all Latin-1 characters.

ColinKerr:
> > > Define and document collation rules for font names
> > 
> > 
> > I sure hope that @khanaffan finishes his PR to make COLLATE NOCASE mean all Latin-1 characters.
> 
> Its finished & merged but not enabled by default. I am still waiting for iQacs results.

Since @khanaffan's pr allow us to turn the Latin-1 behavior on and off we can consider making this configurable via a setting in be_Props.  We could set it to true for all new iModels and leave old iModels as is.  This avoids us having to waste effort adding support for updating older iModels and would let us turn it on more quickly where it actually matters (new iModels).

markschlosseratbentley:
@rbreukelaar Can you assign this a priority?

PR #7377, opened 2024-11-14
Font map casing

### LINKED ISSUE #7376: `FontMap.getFont` is case sensitive
**Describe the bug**
`FontMap.getFont` is case sensitive. If you have an iModel with the font "Arial", calling `FontMap.getFont(“arial”)` does not find the font.

**To Reproduce**

1. Go to [My iTwins](https://developer.bentley.com/my-itwins/).
2. Create an iTwin and create an iModel from the existing House Sample:
Image
3. This iModel has a font with the name “Arial”. Open the iModel in any app of your choice.
4. Call load the font map via a call to `await IModelConnection.fontMap.loadFontMap()`. Then call `IModelConnection.fontMap.getFont("arial")`. Observe that the font is not found.

OR

1. Run the modified test in the associated PR.

**Expected behavior**
`FontMap.getFont` should find a font by name regardless of casing.

**Additional context**
I have pushed a PR with the fix. I modified the tests to catch this case. You can reproduce this issue by running the modified test.

Alfonso-Martello:
Fixes #7376


common/changes/@itwin/core-backend/fontMapCase_2024-11-14-21-53.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "IModel.test.ts now validates that FontMap.getFont uses case insensitive matching for font names",
```
aruniverse, line 5:
```suggestion
      "comment": "",
```

change logs are for end consumers, and they wont care about tests like this

aruniverse (approved):

common/changes/@itwin/core-backend/fontMapCase_2024-11-14-21-53.json
Alfonso-Martello: Thanks, suggestion committed

pmconne (changes requested):
This is a known issue, see discussion in #6751. Please don't merge yet. JavaScript's case-insensitive comparison uses a different collation sequence than that used by iModels.

In what contexts is this causing a problem? Are people manually typing in font names?

Alfonso-Martello:
It is only causing problems in a niche situation for now. No one is manually typing font names. In drawing production, we are using `TextStyleSettings` stored in workspaces. `TextStyleSettings` reference their font by name.

Somehow we found two iModels that reference the same font but have different spellings (the first iModel has the font named “Arial” embedded while the second iModel has the font named “arial” embedded).

Our UI only shows text styles that have their font available in the iModel, so we were wondering why our “Arial” text style was not being listed for the iModel with “arial”. 

We saw that line of code and thought it was a quick fix so opened this PR/issue. Happy to close as the issue is not pressing.

pmconne:
@khanaffan @ColinKerr what was the outcome of your iQacs collection re https://github.com/iTwin/imodel-native/pull/722? Do we have any plans to change the collation sequence for newly-created iModels?

pmconne:
The [native API](https://github.com/iTwin/imodel-native/blob/eae3dd7c30e8da58892f823f701cc5712ddfd406/iModelCore/iModelPlatform/DgnCore/FontManager.cpp#L152) for embedding fonts does a Unicode case-insensitive comparison to reject the input font if a font already exists with the same name. So you could only end up with an issue if two briefcases imported the same font with different casing (where the difference is in something other than ASCII characters), then one pushes their changes and the other pulls them. But that would only occur if both briefcases neglect to first obtain the lock on the entire iModel before writing to the props table. Since no public TypeScript APIs exist for embedding fonts, we're talking about connectors, which surely must obtain that lock.

So I'm fine with merging this change. @khanaffan do you disagree?

Alfonso-Martello:
So, I have some thoughts/questions.
1. I know there was a font discussion the other day with @Josh-Schifter  and @williamkbentley . This is second hand, but my understanding was that a TypeScript API for embedding fonts would be added (or maybe another way for embedding fonts would be exposed)? If that occurs, then would this be more of an issue (i.e. not just applicable to the connectors)?
2. I do not see a clear way to guarantee consistent behavior between JavaScript's collation and our iModel collation. Which then makes me wonder if it is possible for `FontMap.getFont` to return accurate information in all scenarios (either we have inconsistent collation or we have potentially missed fonts due to casing). Maybe it does not matter if the only way to get into a bad data situation is to have a niche workflow as outlined above.
3. I see in #6751 that the plan is to deprecate `FontMap`. I'm not sure if this is still the plan after the most recent discussion, but if yes, then this PR is unneeded.

khanaffan:
> @khanaffan @ColinKerr what was the outcome of your iQacs collection re [iTwin/imodel-native#722](https://github.com/iTwin/imodel-native/pull/722)? Do we have any plans to change the collation sequence for newly-created iModels?

We found several imodel with Latin1 case sensitive codeValues. But it could be done for new imodels.

pmconne (approved):

khanaffan:
> The [native API](https://github.com/iTwin/imodel-native/blob/eae3dd7c30e8da58892f823f701cc5712ddfd406/iModelCore/iModelPlatform/DgnCore/FontManager.cpp#L152) for embedding fonts does a Unicode case-insensitive comparison to reject the input font if a font already exists with the same name. So you could only end up with an issue if two briefcases imported the same font with different casing (where the difference is in something other than ASCII characters), then one pushes their changes and the other pulls them. But that would only occur if both briefcases neglect to first obtain the lock on the entire iModel before writing to the props table. Since no public TypeScript APIs exist for embedding fonts, we're talking about connectors, which surely must obtain that lock.
> 
> So I'm fine with merging this change. @khanaffan do you disagree?

EmbedFont() seem to also use MAX(id) value for given namespace / name to generate new id that it use for new font been stored. This id is part of primary key. I mean without lock two briefcases can easily cause primary key conflict. The strData (contain json) property change from two briefcase only cause data conflict. We can add rule to handle data conflict by skipping/replace. But for Primary key we can only skip or abort (abort been default).

BeProp is not really good place to store fonts. Workspaces may be better and they are by default only let one writer.

pmconne:
> EmbedFont() seem to also use MAX(id) value for given namespace / name to generate new id that it use for new font been stored. This id is part of primary key. I mean without lock two briefcases can easily cause primary key conflict.

We're supposed to use schema sync db to prevent conflicts between server-issued Ids.

> BeProp is not really good place to store fonts. Workspaces may be better and they are by default only let one writer.

Workspaces involve uploading resources to blob storage and redistributing them from there, something expressly prohibited by virtually every licensed font. For example, [per Microsoft](https://learn.microsoft.com/en-us/typography/fonts/font-faq#redistribution-and-extended-rights) most fonts delivered with Windows permit redistribution strictly through embedding.

PR #7387, opened 2024-11-18
Update compile target to `ES2023` and module to `ES2022`

GytisCepk:
### [Compatibility table](https://compat-table.github.io/compat-table/es2016plus/)


aruniverse (approved):
Lgtm, but plz have someone who owns the ecschema metadata review those changes

tools/build/tsconfig-base.json
```diff
@@ -23,6 +23,7 @@
     "strict": true,
     "strictNullChecks": true,
     "stripInternal": false,
-    "target": "ES2021"
+    "useDefineForClassFields": false,
```
GytisCepk, line 26: Default value of `useDefineForClassFields` switched from `false` to `true` with `ES2022`. 

Enabling `useDefineForClassFields` is recommended but currently breaks code in several packages. For that reason, this will be addressed in separate PR.

[More on `useDefineForClassFields`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier)

aruniverse: Sounds good, thx for mentioning this. Please create an issue for this and if you could list all the pkgs that break due to this field, thatd be helpful

GytisCepk: https://github.com/iTwin/itwinjs-core/issues/7474

saskliutas (approved):

GytisCepk:
@ColinKerr @christophermlawson @naveedkhan8067, please review.

christophermlawson (approved):

PR #7394, opened 2024-11-18
documentation: fix link to isInUseLocksError

ben-polinsky:
Links to namespace children must be prefaced by namespace name, even when inside of the namespace. Docs link validation failed to catch this error due an oversight in easing the upgrade to TypeScript 5.3/TypeDoc 0.24.x.

### DIFF
--- a/common/changes/@itwin/core-common/docs-isInUseLocksError-link-fix_2024-11-18-19-36.json
+++ b/common/changes/@itwin/core-common/docs-isInUseLocksError-link-fix_2024-11-18-19-36.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-common"
+}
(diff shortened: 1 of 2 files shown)


pmconne:
> Docs link validation failed to catch this error due an oversight in easing the upgrade to TypeScript 5.3/TypeDoc 0.24.x.

Has the oversight been fixed? Are we confident no other instances of this same error exist elsewhere in the docs?

ben-polinsky:
> Has the oversight been fixed? 

In progress. Link issues were still being logged, but builds would not fail because of them. I'll push a new version of bemetalsmith without this check. 

> Are we confident no other instances of this same error exist elsewhere in the docs?

Yes. The warnings would still be logged. Currently we have a couple changelog warnings, but those are ok and expected if we remove an API. (They probably shouldn't be logged to avoid confusion.) Otherwise, there is one last warning which I am trying to track down `Add remarks: target file bis/domains/SchemaA.ecschema.md could not be located`.

ben-polinsky:
@mergifyio backport release/4.11.x release/4.10.x

PR #7400, opened 2024-11-19
add hydrateDefaultViewState RPC method

hl662:
Adds experimental rpc method, as part of default view creation refactoring. 
- Also adds a method to iModelDb to query for all spatially located models of an iModel, moving the logic from the custom view state creator file over.

This is one half of the refactoring for default view creation.

### DIFF
--- a/core/backend/src/rpc-impl/IModelReadRpcImpl.ts
+++ b/core/backend/src/rpc-impl/IModelReadRpcImpl.ts
@@ -119,6 +119,17 @@ export class IModelReadRpcImpl extends RpcInterface implements IModelReadRpcInte
     return viewHydrater.getHydrateResponseProps(options);
   }
 
+  public async hydrateDefaultViewState(tokenProps: IModelRpcProps): Promise<HydrateViewStateResponseProps> {
+    const iModelDb = await getIModelForRpc(tokenProps);
+    const modelIds = await iModelDb.queryAllSpatial3dModelIds();
+    const options: HydrateViewStateRequestProps = {
+      notLoadedModelSelectorStateModels: CompressedId64Set.compressArray(modelIds)
+    }
+
+    const viewHydrator = new ViewStateHydrator(iModelDb);
+    return viewHydrator.getHydrateResponseProps(options);
+  }
+
   public async queryAllUsedSpatialSubCategories(tokenProps: IModelRpcProps): Promise<SubCategoryResultRow[]> {
     const iModelDb = await getIModelForRpc(tokenProps);
     return iModelDb.queryAllUsedSpatialSubCategories();
--- a/core/backend/src/test/imodel/IModel.test.ts
+++ b/core/backend/src/test/imodel/IModel.test.ts
@@ -811,6 +811,11 @@ describe("iModel", () => {
     });
   });
 
+  it.only("should query for all spatially located models", async () => {
+    const results = await imodel2.queryAllSpatial3dModelIds();
+    assert.isAtLeast(results.length, 1, "Expected query to find some spatially locaated models");
+  })
+
   it("should be able to query for ViewDefinitionProps", () => {
     const viewDefinitionProps: ViewDefinitionProps[] = imodel2.views.queryViewDefinitionProps(); // query for all ViewDefinitions
     assert.isAtLeast(viewDefinitionProps.length, 3);
--- a/core/backend/src/test/imodel/IModel.test.ts
+++ b/core/backend/src/test/imodel/IModel.test.ts
@@ -811,7 +811,7 @@ describe("iModel", () => {
     });
   });
 
-  it.only("should query for all spatially located models", async () => {
+  it("should query for all spatially located models", async () => {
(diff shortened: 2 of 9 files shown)


aruniverse:
> This is one half of the refactoring for default view creation.

Please elaborate what the other half is

hl662:
> > This is one half of the refactoring for default view creation.
> 
> Please elaborate what the other half is

I'll be updating the createDefaultView() in ViewCreator3d to use the new rpc method, and may add a method to SpatialViewState to support that

pmconne:
"experimental rpc method" suggests you may want to remove it based on the outcome of the experiment, but that's not really an option with rpc interfaces used by apps.

Is there a github issue describing the proposed refactoring/experiment and motivations thereof?

hl662:
> "experimental rpc method" suggests you may want to remove it based on the outcome of the experiment, but that's not really an option with rpc interfaces used by apps.
> 
> Is there a github issue describing the proposed refactoring/experiment and motivations thereof?

Yes... I've just linked an opened issue from our private backlog

core/backend/src/rpc-impl/IModelReadRpcImpl.ts
```diff
@@ -119,6 +119,17 @@ export class IModelReadRpcImpl extends RpcInterface implements IModelReadRpcInte
     return viewHydrater.getHydrateResponseProps(options);
   }
 
+  public async hydrateDefaultViewState(tokenProps: IModelRpcProps): Promise<HydrateViewStateResponseProps> {
```
nick4598, line 122: While we were discussing in person, you mentioned that there was a chance 'getCustomViewState3dData' returned the modelIds in a different order which then lead to the hydrateViewState request from the frontend requesting information on the models in a different order. This reduces the cacheability which is the motivation for this PR I believe (Please correct me if I'm wrong). Could you give more details on that?

The backend calls sortAndCompress on the modelIds before returning them to the frontend, but you're saying calling sortAndCompress on the same modelIds can sometimes yield different results?

hl662: It's not sortAndCompress that's the problem - it's that the marshaling/unmarshaling of the rpc system is (most likely) the problem. For example, I have two apps that will call a backend - one through the rpc system, the other through a standard HTTP request. The impl for marshaling (or serialization) of the rpc request parameters are transparent and viewable within this repo, so I follow that same impl for my HTTP request.  And yet, they produce two **near-identical** serialized parameters, but they are not identical - thus, preventing the caching capabilities. 
image
In the image above, I highlighted the discrepeancy between the two.

core/backend/src/test/imodel/IModel.test.ts
```diff
@@ -811,6 +811,11 @@ describe("iModel", () => {
     });
   });
 
+  it("should query for all spatially located models", async () => {
+    const results = await imodel2.queryAllSpatial3dModelIds();
+    assert.isAtLeast(results.length, 1, "Expected query to find some spatially locaated models");
```
MichaelSwigerAtBentley, line 816:
```suggestion
    assert.isAtLeast(results.length, 1, "Expected query to find some spatially located models");
```

hl662:
Me, @nick4598 and @wgoehrig dove into the root of the problem and figured out the reason the serialization is inconsistent is because of the possibility that URI encoding may not happen. 

From the backend side, we don't know for sure if a fetch() (or other HTTP clients) will URI encode the query parameters of a URL before the request is sent. 

On the frontend side, some web browsers will automatically URI encode requests for you, but that's not guaranteed either. And we don't know how previous browser versions behavior work then.

All of this comes down to inconsistency, when making a request. Our RPC system verifies if a request is [base64 encoded](https://github.com/iTwin/itwinjs-core/blob/99a3c7ae87d690737904d73955dc9db14030ba5b/core/common/src/rpc/web/BentleyCloudRpcProtocol.ts#L59-L61) , but if the request is URI encoded on top of that, it will fail. However, it never will, because of the [URLSearchParams.get() ](https://github.com/iTwin/itwinjs-core/blob/99a3c7ae87d690737904d73955dc9db14030ba5b/core/common/src/rpc/web/BentleyCloudRpcProtocol.ts#L57) in the line above - which will automatically decode values for you. This example below shows the functionality available in web browsers:

```
let url = new URL("https://example.com?foo=1&bar=2");
let params = new URLSearchParams(url.search);

// Add a third parameter.
params.set("baz", "=="); // auto encoded
params.toString(); // "foo=1&bar=2&baz=%3D%3D"
params.get("baz"); // Outputs "==" - auto decoded 
```

So in other words, our RPC system implicitly handles URI encoding/decoding, but it is not easy to determine that - the official [MDN documentation ](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get)doesn't mention anything about it either. Perhaps it would be better for an RPC rewrite to explicitly handle that?

My solution to the root problem is to have my backend service URI encode my parameter to ensure consistency, and I can close this PR, knowing the changes in this PR are not needed to do what I need to.

Issue #6214, opened 2023-11-20
Project map imagery onto reality meshes


pmconne:
**Is your feature request related to a problem? Please describe.**

Some scenes include reality meshes representing portions of the earth's surface. Currently, map layer imagery is only draped onto the global terrain/ellipsoid, not onto reality meshes. Unlike the global terrain, these reality meshes usually have their own baked-in texture imagery too.

**Describe the solution you'd like**

Provide a way for a reality mesh to specify that it should be treated as terrain, with map layer imagery draped onto it. Allow it to optionally specify that the base layer (e.g. Bing map imagery) should not be applied to it.

**Describe alternatives you've considered**

None.

**Additional context**

Global terrain uses the same shaders as reality meshes, just sampling the textures differently.

A "treat as terrain" flag could also be useful when the reality mesh represents a large area of the earth's surface, to help the display system reduce precision issues caused by excessively-expanded frustum planes. On the other hand, we might want to avoid conflating that flag with a narrower "drape map imagery on me" flag.

Consider whether we may one day need to support projecting map imagery onto point clouds and design the API accordingly.

alarkbentley:
@danieliborra @markschlosseratbentley Any status update / estimates on development for this?

danieliborra:
We're currently working on this issue (In progress) and it is high in our priority.
No estimates yet.

alarkbentley:
@markschlosseratbentley 

Here is an example from iTwin Experience QA.
Select "Groups & Layers" -> "Reality Mesh" and it will try to render a reality mesh with a WMS map layer on-top.
See video.

https://qa-itwinexperience.bentley.com/dcd5e824-d9fe-40db-961f-83fd5ae95484/map?featureFlagOnCsv=should-ocp-use3dtiles&sceneid=db54b083-e71e-43f3-8b4c-d10c5df9059c

https://github.com/user-attachments/assets/8bf4050d-cc61-48ec-8dbe-7fcde5f4ae3f

Expected:
![Image](https://github.com/user-attachments/assets/6238e9c8-aebf-465d-9543-40e2e0f68714)

markschlosseratbentley:
@alarkbentley Thank you! I'm searching through Chrome dev-tools' network tab, trying to find the sources of this data, but not having much luck.

Could you tell us where we can locally download the reality model itself?  Or a URL directly to it?

Could you also direct us to a URL of the map layer we would be able to attach in our own app?

[...]

PR #7401, opened 2024-11-19
Add ability to drape the background map onto reality meshes.

MarcNeely:
This ability already somewhat existed.  This PR simply exposes it more so that context reality models which are meshes as well as other reality models which are meshes can have the background map draped onto them as well at will.  This is done by adding a RealityMeshDisplaySettings member to the RealityModelDisplaySettings class.

This is a partial fix for issue [#6214](https://github.com/iTwin/itwinjs-core/issues/6214)


MarcNeely:
@pmconne Paul, the change I'm wondering about is in RealityModelTileTree:addToScene.  Currently the display setting for the draping cannot turn off the draping if it is set in the model tile data with the doDrapeBackgroundMap flag.  I'm actually not sure how we would tell that they wanted it to override it anyhow since the bgMapDrape flag is off by default in the RealityMeshDisplaySettings unless we change this from a boolean to an enum and provide a third option for overriding what the tiles say.

pmconne:
@MarcNeely the flag on the tileset indicates a rare, legacy scalable mesh export whose texture imagery comes from a map service instead of being baked in to the tiles. To me, that seems orthogonal to your new RealityMeshDisplaySettings flag that controls whether or not we replace the tileset's texture imagery with map imagery. Kinda like how overriding an element's color to blue would have no visible effect if the element is already intrinsically blue.

alarkbentley:
@MarcNeely any update on this? will it be completed?

markschlosseratbentley:
> @markschlosseratbentley can we close this in favor of #7637?

#7637 is more about draping model map layers. @MarcNeely can likely comment further on whether this one should be kept open.

pmconne:
> > @markschlosseratbentley can we close this in favor of #7637?
> 
> #7637 is more about draping model map layers. @MarcNeely can likely comment further on whether this one should be kept open.

Draping map imagery onto reality models (this PR) is a requirement for iTwin Experience, which will be using Cesium.js as its renderer. We don't need to implement this feature in the iTwin.js renderer. This PR can be closed. Someone will need to determine whether Cesium.js already provides this capability, or otherwise file an issue in that repo.

#7637 is needed in iTwin.js renderer.

PR #7407, opened 2024-11-20
Add note to checkforChanges call to explain what happens with existing transactions

nick4598:


core/backend/src/CloudSqlite.ts
```diff
@@ -502,6 +502,7 @@ export namespace CloudSqlite {
      * Notes:
      * - no changes made by other processes are visible to this CloudContainer unless/until this method is called.
      * - note this is automatically called whenever the write lock is obtained to ensure all changes are against the latest version.
```
hl662, line 504:
```suggestion
     * - this is automatically called whenever the write lock is obtained to ensure all changes are against the latest version.
```

pmconne (approved):

PR #7409, opened 2024-11-20
add setup node to finalize-release.yaml

DanRod1999:
Add setup node step to pipeline. Otherwise we run the risk of a vm not using a supported version of node.

This caused an error in another pipeline, so we should use node setup or nodeTool whenever possible to avoid this problem

### DIFF
--- a/.github/workflows/finalize-release.yaml
+++ b/.github/workflows/finalize-release.yaml
@@ -29,7 +29,7 @@ jobs:
           git config --local user.name imodeljs-admin
 
       - name: Setup node
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
           node-version: "22"
 
--- a/.github/workflows/finalize-release.yaml
+++ b/.github/workflows/finalize-release.yaml
@@ -27,6 +27,12 @@ jobs:
         run: |
           git config --local user.email 38288322+imodeljs-admin@users.noreply.github.com
           git config --local user.name imodeljs-admin
+
+      - name: Setup node
+        uses: actions/setup-node@v2
+        with:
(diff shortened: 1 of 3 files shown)


aruniverse:
this is the only place where we need to do this?
this is a gh pipeline, i was expecting to see changes for ado pipelines

DanRod1999:
> this is the only place where we need to do this? this is a gh pipeline, i was expecting to see changes for ado pipelines

most of our ado pipelines already have this for core already. Otherwise I'm pretty sure we would have experienced breakages from Gytis' PR.
I was just aware that this pipeline might not have it since its GH, so i double checked it.

.github/workflows/finalize-release.yaml
```diff
@@ -27,6 +27,12 @@ jobs:
         run: |
           git config --local user.email 38288322+imodeljs-admin@users.noreply.github.com
           git config --local user.name imodeljs-admin
+
+      - name: Setup node
+        uses: actions/setup-node@v2
```
aruniverse, line 32: please update to the use the latest version of these gh actions, 
https://github.com/actions/setup-node/releases

aruniverse:
linking https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md as an fyi re what ver of node is being used by the image

.github/workflows/finalize-release.yaml
aruniverse: please update the other actions and any other gh workflow, 
eg actions/checkout@v2

DanRod1999:

aruniverse (approved):

PR #7421, opened 2024-11-22
Update UseNode and Checkout in yml files

DanRod1999:
upgrade to UseNode because node tool is depricated, update checkout to v4

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -28,7 +28,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
 
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v2
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -31,7 +31,7 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
+      uses: github/codeql-action/init@v3
       with:
         languages: 'javascript'
         # If you wish to specify custom queries, you can do so here or in a config file.
@@ -40,4 +40,4 @@ jobs:
         # queries: ./path/to/local/query, your-org/your-repo/queries@main
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v2
+      uses: github/codeql-action/analyze@v3
--- a/.github/workflows/finalize-release.yaml
+++ b/.github/workflows/finalize-release.yaml
@@ -25,7 +25,7 @@ jobs:
           fetch-depth: 0
       - name: Set Git Config
         run: |
-          git config --local user.email 38288322+imodeljs-admin@users.noreply.github.com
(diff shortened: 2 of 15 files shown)


hl662 (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -28,7 +28,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
 
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v2
```
aruniverse, line 34: can we update all actions while were doing these updates to make sure we're not falling behind?

.github/workflows/finalize-release.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
     name: Cherry-pick Changelogs
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           token: ${{ secrets.IMJS_ADMIN_GH_TOKEN }}
           ref: ${{ github.ref }} # checkouts the branch that triggered the workflow
```
aruniverse, line 24: update the user.email on line 28,
we dont need the random numbers and the + in front
keep it consistent with the other pipelines
not often we review all these pipelines, lets try to make sure theyre all up to date and clean

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -146,10 +146,10 @@ jobs:
     steps:
       - checkout: self
 
-      - task: NodeTool@0
+      - task: UseNode@1
         displayName: Use Node 20.x
         inputs:
-          versionSpec: 20.x
+          version: 20.x
```
aruniverse, line 152: why is this on 20 when everything else has moved to 22?

common/config/azure-pipelines/ci-ios.yaml
```diff
@@ -44,10 +44,10 @@ jobs:
       - checkout: self
         clean: true
 
-      - task: NodeTool@0
+      - task: UseNode@1
         displayName: "Use Node 20"
         inputs:
-          versionSpec: 20.x
+          version: 20.x
```
aruniverse, line 50: this one we can leave at 20 since @tcobbs-bentley is still working on node 22 support for mobile

common/config/azure-pipelines/valgrind.yaml
```diff
@@ -26,10 +26,10 @@ jobs:
       - checkout: self
         clean: true
 
-      - task: NodeTool@0
+      - task: UseNode@1
         displayName: "Use Node 20.x"
         inputs:
-          versionSpec: 20.x
+          version: 20.x
```
aruniverse, line 32: why not 22?

aruniverse (approved):

common/config/azure-pipelines/ci-ios.yaml
tcobbs-bentley: It's worth mentioning that the build-time Node version on the build host (Mac) doesn't need to match the runtime Node version on the device (iOS/Android). So using Node 22 for the iOS build is probably fine.

PR #7429, opened 2024-11-26
Derived class caching

christophermlawson:
Introduced derived class caching, which required a breaking change to the ECClass.baseClass property setter to the setBaseClass method.


core/ecschema-editing/src/Editing/RelationshipClasses.ts
```diff
@@ -115,7 +115,7 @@ export class RelationshipClasses extends ECClasses {
     try {
       await this.validate(relClass!);
     } catch(e: any) {
-      relClass!.baseClass = baseClass;
+      await relClass!.setBaseClass(baseClass);
```
StefanApfel-Bentley, line 118: Should we add some check here (or actually in line 111) to ensure that relClass is not undefined? Since it's a lookup by key, I assume it could be happen (eg when loading from a referenced schema) or did I missed some previous validation here?

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -49,6 +48,37 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
+   * Gets the base class if it exists, otherwise returns undefined.
+   */
+  public get baseClass(): LazyLoadedECClass | undefined {
+    return this._baseClass;
+  }
+
+  /**
+   * Sets the base class of the ECClass. Pass undefined to 'remove' the base class.
+   */
+  public async setBaseClass(baseClass: LazyLoadedECClass | undefined) {
```
StefanApfel-Bentley, line 61: Do we have to use a LazyLoaded class here? In all calls I saw, we've always had the baseClass resolved already, so there wouldn't be a need to do the detour via the delayed loaded proxies. Even when loading from JSON we expect the baseClass to be in the schema already (lookupItemSync in line 458).

That makes we wonder if we either missed a case where the baseClass is not yet present in the schema (aka lookup would fail) or if having the base class lazy-loaded is over-complicating things there as we have the items resolved already.

christophermlawson: I think this should be internal and protected. Let me think on this. The editing API should call this. The old property setter  took the LazyLoaded object, and I don't think we should change it.

core/ecschema-editing/src/Editing/RelationshipClasses.ts
```diff
@@ -107,15 +108,18 @@ export class RelationshipClasses extends ECClasses {
...
+      });
     const baseClass = relClass?.baseClass;
 
     await super.setBaseClass(itemKey, baseClassKey);
 
     try {
       await this.validate(relClass!);
     } catch(e: any) {
-      await relClass!.setBaseClass(baseClass);
+      await (relClass! as ECClass as MutableClass).setBaseClass(baseClass);
```
jason-crow, line 122: why is a double cast necessary? especially given you already have  the definitely assigned operator, which together makes this about the least typescript typescript way to accomplish what you seem to want to do here

core/ecschema-editing/src/Editing/RelationshipClasses.ts
christophermlawson: I added a catch block on line 112.

core/ecschema-metadata/src/Metadata/Class.ts
christophermlawson: FYI, I updated the method to protected. I know we discussed this on a call a while ago and decided to keep the LazyLoaded stuff for now.

core/ecschema-editing/src/Editing/RelationshipClasses.ts
christophermlawson: @jason-crow Jason, if you are OK with my response can you resolve this conversation?  Thanks!

StefanApfel-Bentley: As this PR is lingering around for a while, and indeed the double cast is not nice, I'd propose we keep that double casting by now and revisit that during the upcoming itwinjs 5 fitting work.

@christophermlawson , can you create an issue on our backlog that we don't forget that.

StefanApfel-Bentley (approved):

PR #7444, opened 2024-12-03
New workflow for creating release branches

DanRod1999:
New workflow only for creating release branches off of already existing release branches. This is intended only to use when master is on a different minor branch then the one we are currently using. This will be needed for new 4.x minors while 5.0 is currently in development.

Related issue: https://github.com/iTwin/itwinjs-backlog/issues/1316

The purpose of a new workflow is to avoid changing the logic of the currently existing bump version pipeline. I believe re working the current scripts (here and in native) would be more difficult and would risk breaking what already works.

I will admit there are already a few parts of the current scripts I don't like, so maybe it would be worth while to rework in the future (I have an issue already created for it). So if anyone does work on that script, it might be better to address everything in one large change instead of multiple (most likely still large) changes.

### DIFF
--- a/.github/workflows/create-release-branch.yaml
+++ b/.github/workflows/create-release-branch.yaml
@@ -9,13 +9,6 @@ on:
       releaseBranch:
         description: 'The name of the release branch to create'
         required: true
-        default: 'TEST/0.0.x' # FIXME temp default while testing. Otherwise it will create protected branches
-  pull_request:
-    types: [opened, synchronize]
-
-# FIXME FOR TESTING PURPOSES ONLY
-env:
-  RELEASE_BRANCH: 'TEST/0.0.x'
 
 jobs:
   create-release-branch:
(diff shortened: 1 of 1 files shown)


DanRod1999:
This [commit ](https://github.com/iTwin/itwinjs-core/commit/374abb134293eb6135a9b5045495339f818957b5#diff-1b49fbae15655fe232b3b457fdd1a6cb8f253c495b42c3d3d220a6cd4aa89504) on branch TEST/0.0x is an example of what we would see when running this off of a release branch. Branch TEST/0.0.x was auto created by this pipeline, so that value represents what will be a normal release branch ex. release/4.12.x. 
[Here ](https://github.com/iTwin/itwinjs-core/actions/runs/12204722594/job/34050436289?pr=7444) is the workflow run that performed this.

Slight differences from my test and what would actually take place:
- as mentioned this was triggered by my PR. Normally this should be triggered manually, but GH does not seem to allow manual triggers for workflows that do not exist on master.
- I hard coded values for the branch name `TEST/0.0.x` as an env var. This should really be an input var set under workflow_dispatch. This is very similar to parameters that we set on run time for ADO.
- this was ran off of a branch based on master. So this commit might look a little different compared to when running off of a release branch. However, I believe it should look mostly the same, and the logic still all holds true.

PR #7447, opened 2024-12-04
Change expected number of CRS to 11975 in GeoServices.test.ts

nick4598:
image
Arun mentioned that gcs workspace was deployed to prod yesterday.


aruniverse:
@wgoehrig this seems pretty fragile no?

The gcs workspace doesnt get updated often, but it theoretically could.
Should we rather just expect the gcs workspace to not be empty?

wgoehrig (approved):

wgoehrig:
@aruniverse I actually really like this test, I just think we should move it to run as part of the workspace deployment...

aruniverse:

ben-polinsky:

PR #7450, opened 2024-12-05
Obtain Base Url for Texture Requests within Reality Tile Loader

andremig-bentley:
Currently, there is a bug attaching a 3dTiles tileset to the itwin viewer using `attachRealityModel` where if that model has textures, and the tileset has relative URIs for those textures, they will not be requested. This is because the base url for those requests was not being passed along to the gltf reader, and so an attempt to resolve the images for these textures would fail. This PR addresses that issue by creating the new optional `baseUrl` parameter passed down through the creation of the `RealityTileTree`. `baseUrl` is created using the new `getTilesetUrlFromTilesetUrlImpl` in `RealityDataSource`, which obtains the tileset url from a source which is an instance of `RealityDataSourceTilesetUrlImpl`. In `RealityTileLoader.ts`, the url is checked to determine if it's a valid url, and if so, it is passed in to the creation of the `GltfReaderProps`, and therefore, to the gltf reader.


aruniverse:
does this need to be backported to 4.10? or can it go in 4.11?

core/frontend/src/tile/RealityTileLoader.ts
```diff
@@ -143,7 +143,23 @@ export abstract class RealityTileLoader {
...
       case TileFormat.Gltf:
-        const props = GltfReaderProps.create(streamBuffer.nextBytes(streamBuffer.arrayBuffer.byteLength), yAxisUp);
+        /* We can obtain the base url from the tile's tree id.
+        * The id is constructed with parts from the reality data sourcekey like this - provider:format:id:itwinId
+        * The id in this case is the base url of the reality data source. We can split with ":" and parse to get the base url.
+        */
+        const treeId = tile.tree.id;
+        const split = treeId.split(":");
+
+        /* Splitting with ":" will end up splitting up our base url as well as the id.
```
aruniverse, line 153: all of this seems very fragile,
please add unit test for these

core/frontend/src/tile/RealityTileLoader.ts
```diff
@@ -143,7 +143,23 @@ export abstract class RealityTileLoader {
         reader = I3dmReader.create(streamBuffer, iModel, modelId, is3d, tile.contentRange, system, yAxisUp, tile.isLeaf, isCanceled, undefined, this.wantDeduplicatedVertices);
         break;
       case TileFormat.Gltf:
-        const props = GltfReaderProps.create(streamBuffer.nextBytes(streamBuffer.arrayBuffer.byteLength), yAxisUp);
+        /* We can obtain the base url from the tile's tree id.
+        * The id is constructed with parts from the reality data sourcekey like this - provider:format:id:itwinId
```
pmconne, line 147: Are you assuming here that the tileset was produced by mesh export service from an iModel, and will follow this specific URL scheme? Any 3d tileset that uses glTF as its payload (e.g., every 3D Tiles 1.1 tileset) is going to come through this path, with no standard URL scheme.

Are you trying to solve the general problem of needing a base URL, or the specific problem of attaching iModel tiles as a reality model (something nobody in the real world will do)?

andremig-bentley: This is a general problem that happens when a tileset has relative URIs for textures within the .glb/.gltf. It was initially discovered trying to attach an export (the exports have relative texture URIs), but the same behavior would show up with any tileset with relative texture URIs. I'll update the PR description to make that more clear.

The main thinking was that if a tileset came down this path, and had a valid full url in its tree id, then it would need the base url to be passed along to the creation of the gltf reader. I guess this does lead to the possibility of a full valid url being present in the tree id, with full texture urls also being present in the tileset, in which case the base url wouldn't need to be passed into the creation of the reader. A better solution might be to create a separate path for only the case where relative URIs are present and leave the main path as is.

andremig-bentley:
An update on this PR - Due to the fragility of the tree id parsing method, the methods of obtaining the base url for this PR have changed. Now, all tree id parsing has been removed, and instead, the id has been passed down through the creation of the `RealityTileTree` as the new optional member `rdSourceId`. By passing down the id directly, we avoid any possible issues that would come with parsing and avoid errors that might have been caused by unexpected aspects of the tree id. 

Tests have been added to a new file, `RealityTileLoader.test.ts`, which test the creation of `GltfReaderProps` including or not including a base url as expected. This new test file includes a lot of set up code from `GltfReader.test.ts` which was necessary to successfully create `GltfReaderProps`. In my opinion, these tests make more sense being in their own file rather than being in `GltfReader.test.ts`, but they can be moved/changed if necessary. 

Finally, we were able to test the case where a base url is passed in to the gltf reader for a  tileset with full texture urls - the case discussed in my comment above. The texture requests worked as expected with no issues, so that edge case is a non-issue.

core/frontend/src/tile/RealityModelTileTree.ts
```diff
@@ -711,7 +713,7 @@ export namespace RealityModelTileTree {
       const props = await getTileTreeProps(rdSource, tilesetToDb, iModel);
       const loader = new RealityModelTileLoader(props, new BatchedTileIdMap(iModel), opts);
       const gcsConverterAvailable = await getGcsConverterAvailable(iModel);
-      const params = new RealityModelTileTreeParams(tileTreeId, iModel, modelId, loader, gcsConverterAvailable, props.tilesetToEcef);
+      const params = new RealityModelTileTreeParams(tileTreeId, iModel, modelId, loader, gcsConverterAvailable, props.tilesetToEcef, rdSource.key.id);
```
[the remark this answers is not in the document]
andremig-bentley: I was interpreting it as a base url due to its creation in `createKeyFromUrl`, [here](https://github.com/iTwin/itwinjs-core/blob/7bdb8cb11bbbe9a3f7fe0c9eb0d3f7191ba48934/core/frontend/src/RealityDataSource.ts#L131), where the function defaults to setting the reality data source key id as the tileset url. The thinking was that within the other two paths in that function, the id wouldn't be a valid url, so then when we check for a valid url later, the ids wouldn't be added to the reader, so no issue. 

After looking into this more, this does leave the possibility that the source id is created using one of the other paths in `createKeyFromUrl`, meaning the id might not be a valid url, but there still could be relative texture uris within the tileset, meaning we still need to pass along a base url. Changes are underway now to pass along the tileset url used to create the reality data source key instead of just the key id itself, thereby avoiding these issues.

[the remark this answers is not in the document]
pmconne: Did you notice that RealityDataSourceTilesetUrlImpl already has a base URL?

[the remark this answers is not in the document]
andremig-bentley: Thanks for the suggestions on where to look, a few things have now been updated - `getTilesetUrlFromTilesetUrlImpl` has been added to `RealityDataSource` which obtains the tileset url from a source which is an instance of `RealityDataSourceTilesetUrlImpl`. This is being used to set the `baseUrl` property in `RealityModelTileTree`. A note: we need to get the full tileset url, not just the base url, from `RealityDataSourceTilesetUrlImpl`. This is because `ResolveUrl` in `GltfReader`, [here](https://github.com/iTwin/itwinjs-core/blob/4a92d6853949bada726a3d14ec29b0c1b85a7622/core/frontend/src/tile/GltfReader.ts#L1958), expects the base url to have search parameters if they are necessary, and if we were to only use the base url from `RealityDataSourceTilesetUrlImpl` for something like a tileset in blob storage, these search parameters would not be present, and the texture requests would return forbidden.

core/frontend/src/tile/RealityTileLoader.ts
markschlosseratbentley: @aruniverse does the added unit test suffice?

core/frontend/src/RealityDataSource.ts
```diff
@@ -208,6 +208,13 @@ export namespace RealityDataSource {
 
     return provider.createRealityDataSource(key, iTwinId);
   }
+
+  /** Returns the source's tileset url if the source is an instance of RealityDataSourceTilesetUrlImpl.
+   * @internal
+   */
+  export function getTilesetUrlFromTilesetUrlImpl(rdSource: RealityDataSource): string | undefined {
```
pmconne, line 215: Delete this, add an optional method on the `RealityDataSource` interface, and implement it in `RealityDataSourceTilesetUrlImpl`. That will allow any other implementations of `RealityDataSource` to implement it themselves if they so choose.

core/frontend/src/tile/RealityTileLoader.ts
```diff
@@ -242,3 +243,23 @@ export abstract class RealityTileLoader {
...
+}
+
+function isValidURL(url: string){
+  try {
+    new URL(url);
+  } catch {
+    return false;
+  }
+  return true;
+}
```
pmconne, line 265:
```suggestion
export function createReaderPropsWithBaseUrl(streamBuffer: ByteStream, yAxisUp: boolean, baseUrl?: string): GltfReaderProps | undefined {
  let url: URL | undefined;
  try {
    url = new URL(baseUrl);
  } catch (_) { }
  
  return GltfReaderProps.create(streamBuffer.nextBytes(streamBuffer.arrayBuffer.byteLength), yAxisUp, url);
```
You appear to accept `Uint8Array` strictly because your test wants to pass one in - make the test pass in a ByteStream.

core/frontend/src/test/tile/RealityTileLoader.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  if (json)
+    chunks.push({ type: GltfV2ChunkTypes.JSON, data: jsonToBytes(json) });
+
+  if (binary)
+    chunks.push({ type: GltfV2ChunkTypes.Binary, data: binary });
+
+  return glbFromChunks(chunks, header);
+}
+
+describe("createReaderPropsWithBaseUrl", () => {
```
pmconne, line 75: These tests don't prove much. A much more valuable test would confirm that you successfully resolve a resource inside the glTF with a relative URL where prior to your change you would have failed to do so.

core/frontend/src/RealityDataSource.ts
andremig-bentley: Done in the latest commit

core/frontend/src/test/tile/RealityTileLoader.test.ts
andremig-bentley: Tests have been added now to `GltfReader.test.ts`. These tests set up a `GltfReader` as the `RealityTileLoader` now does with the changes in this PR. They then get a relative texture url from a glTF document, then use the created reader to test the resolution of that url. If the url is created correctly, we know it will fetch correctly, so we can show that we will successfully resolve a resource we wouldn't have previously resolved.

common/changes/@itwin/core-frontend/andremig-texture-baseurl_2024-12-05-15-48.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "",
```
pmconne, line 5:
```suggestion
      "comment": "Fix failure to resolve relative texture URIs in 3D Tiles",
```

pmconne (approved):

core/frontend/src/tile/RealityTileLoader.ts
pmconne: @aruniverse

pmconne: nvm resolving as all the parsing logic is gone now.

PR #7458, opened 2024-12-06
Adding Rebase strategy to pull/merge 

khanaffan:
# Pull merge & conflict resolution

> NOTE This document is work in progress and Rebase API are **experimental**.

## What is change merging?

Change merging is when a briefcase pulls new changes from the iModel Hub and applies them locally. If the user has any local changes, the incoming changes need to be merged with the local changes. In this document, we are currently strictly talking about low-level SQLite changesets and high-level concepts defined in ECSchemas.

## What is change conflicts

Git is used for text and text is two-dimensional data. Conflicts in two-dimensional data happen when two users change overlapping ranges of text. In a database, an overlapping change can be a change to the same row identified by a primary key. However, it could be more complex as that row might have database constraints like unique index, check, foreign key, triggers, etc., that make this row part of a larger chunk of information. SQLite tracks changes to individual tables/rows. It is up to the application to make changes consistently so as not to violate database constraints. If the database only contains a single table with no constraints, then it's basically the same as a text file as long as there is a primary key.

SQLite records changes by recording INSERT/UPDATE/DELETE operations with old and new values into a binary changeset. The order of changes in the changeset is arbitrary, but once the whole changeset is applied, it should bring the database to a new state that is consistent with all specified database constraints.

Things get interesting when there are local changes specifically to the same data as the incoming changeset. This operation is called pull-merge. Doing that can result in conflicts which from the SQLite perspective is as follows.

|  | Operation | Conflict description                                                   | Type         |
|--|-----------|------------------------------------------------------------------------|--------------|
|  | `INSERT`  | 1. PRIMARY KEY already exists                                          | `Conflict`   |
|  |           | 2. Database constraint voliation e.g. UNIQUE, CHECK                    | `Constraint` |
|  | `DELETE`  | 1. PRIMARY KEY does not exist                                          | `NotFound`   |
|  |           | 2. PRIMARY KEY does exist but other fields values does not match       | `Data`       |
|  |           | 3. Database constraint voliation e.g. UNIQUE, CHECK caused by update   | `Constraint` |
|  | `UPDATE`  | 1. PRIMARY KEY does not exist                                          | `NotFound`   |
|  |           | 2. PRIMARY KEY does not exist but data fields values does not match    | `Data`       |
|  |           | 3. Database constraint voliation e.g. UNIQUE, CHECK caused by update   | `Constraint` |
|  | -         | Forignkey voliations. Its is not for given row but for whole changeset | `ForeignKey` |

Above conflict can be resolved in on of the followed allowed ways. A `REPLACE` resolution may cause `CONSTRAINT` conflict afterword if db constrain are voilated by `REPLACE` action. If `CONSTRAINT` conflict is skipped then it can cause `ForeignKey` voliation at the end of changeset apply.

|  | Operation | Conflict     | Skip                        | Replace                                   | Abort   | Default   |
|--|-----------|--------------|-----------------------------|-------------------------------------------|---------|-----------|
|  | `INSERT`  | `Conflict`   | Allowed                     | Allowed but may cause constraint conflict | Allowed | `ABORT`   |
|  |           | `Constraint` | Allowed. Change is ignored. | n/a                                       | Allowed | `SKIP`    |
|  | `DELETE`  | `NotFound`   | Allowed                     | n/a                                       | Allowed | `ABORT`   |
|  |           | `Data`       | Allowed                     | Allowed but may cause constraint conflict | Allowed | `REPLACE` |
|  |           | `Constraint` | Allowed. Change is ignored. | n/a                                       | Allowed | `SKIP`    |
|  | `UPDATE`  | `NotFound`   | Allowed                     | n/a                                       | Allowed | `ABORT`   |
|  |           | `Data`       | Allowed                     | Allowed but may cause constraint conflict | Allowed | `REPLACE` |
|  |           | `Constraint` | Allowed. Change is ignored. | n/a                                       | Allowed | `SKIP`    |
|  | -         | `ForeignKey` | n/a                         | n/a                                       | Allowed | `ABORT`   |

> Note even ForeignKey can be skipped but this will cause db integrity check to fail or any change to db in future may fail.

The foreign key one is the only one that is not associated with a specific row/table as it is a count of FK left unresolved and is a fatal error pointing to an error in application logic.

When SQLite applies a changeset and detects a conflict, it will call the conflict handler and request application feedback on how to resolve it. For the above, we can generally choose `SKIP`, `REPLACE`, or `ABORT`. If we choose `REPLACE`, in some cases, our conflict handler may be called again if replacing the local row causes another conflict due to database constraints. We are looking at one row at a time, but these rows do not exist in isolation. They are connected to the rest of the rows in the same table via constraints, to the table definition via CHECK constraints, or to other tables via foreign keys. This makes change merging in the database a bit more complex.

## Types of change merging methods

Now that we have a somewhat understanding of key concepts about change merging and conflicts, we can talk about how to merge conflicting changes.

itwinjs supports two types of change merging:

1. **Merge**: pull new changes & apply these changes on top of local changes (default)
2. **Rebase**: pull new changes, rollback local changes, apply new changes, and then apply local changes. (new as in 5.x but not default, WIP/internal)

Note: Since a briefcase as of now use locks to serialize changes between two or more users. The two methods specified does not really make any difference. The locks ensure only one user can change a given element and thus we do not expect low level sqlite conflict for any directly changed data.

### "Merge" method

Merge is simple and straightforward but not flexible and can easily cause the creation of a changeset that might look fine at the local file level but fail when applied to a new file.

It is done as follows:

1. Briefcase pulls new changes.
2. It applies new changes on top of local changes.
   * It depends on locks for changes to be non-conflicting.
   * This method does not record the effect of applying incoming changes. This locally recorded changeset is an unreliable source of truth without locking.
3. If a conflict happens, there is not much we can do except for either skipping the incoming change or replacing the local change. There is no room for actually merging.
4. Conflict resolution is recorded in a rebase buffer, so when pulling more changes, the same resolution will be applied.
   * This is also dangerous because let's say we choose REPLACE for a DATA conflict. And then change that row later in the session. When we push changes, the local change to the row will be ignored.
5. Finally, when all changes are merged, we can push the changes, but without a lock, this could create a changeset that cannot be applied by others.

As we can see from the above, this merge method works fine if you lock individual rows in a table so only one user can change it. This workflow cannot really scale, and without a lock, it is completely useless.

## "Rebase" method

This is a new method that has been implemented that is way more flexible. It is similar to git rebase. It is simple and easy to understand and will never lead to the push of a changeset that cannot be applied without locks.

1. Briefcase pulls new changes.
2. Briefcase local changes are rolled back, so now it's the same as a point in history on the master timeline.
3. New changes from the master are applied and advance local head to the master. No conflicts are expected as we simply update the local briefcase to the master.
4. Local changes from oldest to newest are now applied on top of the master. *Note that the current disk file represents the master and the changeset been applied is the local change made by the user.*
5. This can cause conflicts as we replay local changes on top of the new master tip.
   * With rebase, we have flexibility that we can change data beside simple choices of `SKIP`, `REPLACE`, & `ABORT`.
   * If the application finds a conflict that requires creating new changes or merging incoming and local changes, then it can do so.
   * It also has a 3-way view of what changed. The local changeset been applied has old/new values while the current value in the db represents the master.
   * The whole merge activity is recorded in the changeset, and the local changeset is updated after merging. In other words, the local change is recomputed against the master.
   * Any tool that made those changes can also react to rebase event and update there change against what is new on master.
6. If at any point the rebase fails, it can be resumed. The db is unable to push unstable changes until it resolves it locally first.
7. Finally user can still undo his local changes. Or there undo stack is preserved. They can also push out there changes if needed.

As one of the most important take away in rebase is application control of merge. It has tools that made those changes and can correct redo them over master if needed. Where as "Merge" method is one way and does not give application any control over merge.

## Rebase API workflow

At the moment only one set of changes does not require locks and we will use that as example. Its local properties of briefcase. Currently two user can change them and it can cause conflicts.

After downloading briefcase you can explicitly change merge method to rebase.

```ts
    b1.txns.changeMergeManager.setMergeMethod("Rebase");
```

you can also set it by default this will cause any new briefcase to automatcally use "Rebase" as default method.

```ts
  await IModelHost.startup({ pullMergeMethod: "Rebase" });
```

We download and open two briefcases for the same iModel, then set a property on b1 and push the changes. Note that the property being set did not exist, so it causes an INSERT operation at the database level.

```ts
  const b1 = ...
  const b2 = ...

  // save a property
  b1.saveFileProperty({ namespace: "test", name: "test" }, "foo");
  b1.saveChanges();
  await b1.pushChanges({ description: "set test to foo" });
```

Now b2 does the same and sets a different value, which also causes an INSERT operation being recorded.

```ts
  b2.saveFileProperty({ namespace: "test", name: "test" }, "goo");
  b2.saveChanges();

  // following will throw exception with error
  // `PRIMARY KEY INSERT CONFLICT - rejecting this changeset`
  await b2.pushChanges({ description: "set test to goo" });
```

Above threw exception as after undoing local changes then applying incoming changes we now have an entry for the key `test` in the db. While locally we recorded an insert operation for the `test` property. This causes a primary key violation. This action leaves the database in rebase mode.

Next, we will install a conflict handler to resolve this issue. This conflict handler concatenates the master value with the local.

```ts
  b2.txns.changeMergeManager.addConflictHandler({
    id: "my", handler: (args: RebaseChangesetConflictArgs) => {
      if (args.cause === DbConflictCause.Conflict) {
        if (args.tableName === "be_Prop") {
          if (args.opcode === DbOpcode.Insert) {
            const localChangedVal = args.getValueText(5, DbChangeStage.New);
            const tipValue = b2.queryFilePropertyString({ namespace: "test", name: "test" });
            b2.saveFileProperty({ namespace: "test", name: "test" }, `${tipValue} + ${localChangedVal}`);
            return DbConflictResolution.Skip; // skip incomming value and continue
          }
        }
      }
      return undefined;
    }
  });
```

We resume rebase and it will resolve the conflict, allowing us to push the changes.

```ts
  // resume rebase see if it resolve the conflict
  b2.txns.changeMergeManager.resume();
  await b2.pushChanges({ description: "set test to goo" });
```

The final value of test pushed will be `foo + goo`.

imodel-native: https://github.com/iTwin/imodel-native/pull/933


docs/learning/backend/PullMerge.md
```diff
@@ -0,0 +1,148 @@
...
+Things get interesting when there are local changes specifically to the same data as the incoming changeset. This operation is called pull-merge. Doing that can result in conflicts which from the SQLite perspective is as follows.
+
+|  | Operation | Conflict description                                                    | Type         |
+|--|-----------|-------------------------------------------------------------------------|--------------|
+|  | `INSERT`  | 1. PRIMARY KEY already exists                                           | `Conflict`   |
+|  | `DELETE`  | 2. PRIMARY KEY does not exist                                           | `NotFound`   |
+|  | `UPDATE`  | 3. PRIMARY KEY does not exist                                           | `NotFound`   |
+|  |           | 1. Database constraint voliation e.g. UNIQUE, CHECK                     | `Constraint` |
+|  |           | 2. PRIMARY KEY does not exist but other fields values does not match    | `Data`       |
+|  |           | 3. PRIMARY KEY does not exist but other fields values does not match.   | `Data`       |
```
pmconne, line 24: These two lines are identical and I think (?) it's supposed to say "PRIMARY KEY matches", not "PRIMARY KEY does not match".

docs/learning/backend/PullMerge.md
```diff
@@ -0,0 +1,148 @@
...
+|  |           | 4. Database constraint voliation e.g. UNIQUE, CHECK                     | `Constraint` |
```
pmconne, line 25: Duplicate from 3 rows above. I think maybe you meant to put different operation names in the first column?

docs/learning/backend/PullMerge.md
```diff
@@ -0,0 +1,148 @@
...
+
+SQLite records changes by recording INSERT/UPDATE/DELETE operations with old and new values into a binary changeset. The order of changes in the changeset is arbitrary, but once the whole changeset is applied, it should bring the database to a new state that is consistent with all specified database constraints.
+
+Things get interesting when there are local changes specifically to the same data as the incoming changeset. This operation is called pull-merge. Doing that can result in conflicts which from the SQLite perspective is as follows.
+
+|  | Operation | Conflict description                                                    | Type         |
+|--|-----------|-------------------------------------------------------------------------|--------------|
+|  | `INSERT`  | 1. PRIMARY KEY already exists                                           | `Conflict`   |
+|  | `DELETE`  | 2. PRIMARY KEY does not exist                                           | `NotFound`   |
+|  | `UPDATE`  | 3. PRIMARY KEY does not exist                                           | `NotFound`   |
```
pmconne, line 21:
```suggestion
|  | `DELETE/UPDATE`  | 2. PRIMARY KEY does not exist                                           | `NotFound`   |
```

common/api/core-bentley.api.md
```diff
@@ -1296,6 +1296,9 @@ export class LRUMap<K, V> extends LRUCache<K, V> {
 // @public
 export type MarkRequired<T, K extends keyof T> = Pick<Required<T>, K> & Omit<T, K>;
 
+// @public
+export type Maybe<T> = T | undefined;
```
pmconne, line 1300: Nit but do we really need/want this? `x?: X` vs `x: Maybe<X>`. One uses terse, standard TypeScript syntax; the other uses a non-standard type to express the same thing more verbosely. The only place this saves you characters (if that's the goal) is when you want to require someone to specify `undefined` explicitly.

core/backend/src/IModelHost.ts
```diff
@@ -195,6 +198,9 @@ export interface IModelHostOptions {
    * Will be changed to default to `false` in 5.0.
    */
   allowSharedChannel?: boolean;
+
+  /** @internal */
```
pmconne, line 202: Are this and `defaultPullMergeMethod` really supposed to remain `internal`? Currently you provide no way for apps to choose between rebase and merge.

core/backend/src/IModelHost.ts
```diff
@@ -453,6 +462,23 @@ export class IModelHost {
...
+    return this.configuration?.pullMergeMethod ?? IModelHostConfiguration.defaultPullMergeMethod;
+  }
+  /** Gets the pull merge method used by the IModelHost.
+   * @internal
+   * @note If a custom pull merge method is not configured, the default pull merge method is used.
+   * @returns The pull merge method.
+   */
+  // public static set pullMergeMethod(method: PullMergeMethod):  {
+  //   return this.configuration?.pullMergeMethod ?? IModelHostConfiguration.defaultPullMergeMethod;
+  // }
```
pmconne, line 480:
```suggestion
```

pmconne (commented):
This is a big exciting new feature, but nobody can currently use it. What needs to happen before they can?

kabentley:
I don't think this should be a public API at all, because it doesn't make any sense to have to choose between them. If there's a conflict, the only way out is to rebase. There doesn't need to be an option - just always do merge-then-rebaseOnFailure. That will be the most efficient anyway. BTW, you mentioned that you think rebasing eliminates the need for a schemaDb, but I don't understand how they're related at all. Maybe we should talk?

core/backend/src/TxnManager.ts
```diff
@@ -463,6 +632,16 @@ export class TxnManager {
    */
   public readonly onReplayedExternalTxns = new BeEvent<() => void>();
 
+  /** @internal */
+  public readonly onRebaseTxnBegin = new BeEvent<(txn: TxnArgs) => void>();
+  /** @internal */
+  public readonly onRebaseLTxnEnd = new BeEvent<(txn: TxnArgs) => void>();
```
aruniverse, line 638:
```suggestion
  public readonly onRebaseTxnEnd = new BeEvent<(txn: TxnArgs) => void>();
```

core/backend/src/TxnManager.ts
```diff
@@ -406,6 +483,98 @@ export class TxnManager {
...
+      args.setLastError(msg);
+      return DbConflictResolution.Abort;
+    }
+
+    if (args.cause === DbConflictCause.NotFound) {
+      return DbConflictResolution.Skip;
+    }
+
+    if (args.cause === DbConflictCause.Constraint) {
+      if (Logger.isEnabled(category, LogLevel.Info)) {
```
aruniverse, line 563: do we really need to wrap log statements to specific levels with checks if that level is enabled?

khanaffan: Its copied to produce same errors/warning as in native. To match logs. But in case of rebase we do not need to be similar as its new code. I will update this method to minimally log and use different category.

core/backend/src/IModelHost.ts
pmconne: Keeping them internal. As @kabentley points out, we can't really allow individual apps to decide. We'll keep the internal APIs temporarily for testing.

docs/learning/backend/PullMerge.md
khanaffan: ah i think some where i click sort on markdown table in vscode which messed it up. I am fixing this.

khanaffan:
> I don't think this should be a public API at all, because it doesn't make any sense to have to choose between them. If there's a conflict, the only way out is to rebase. There doesn't need to be an option - just always do merge-then-rebaseOnFailure. That will be the most efficient anyway. BTW, you mentioned that you think rebasing eliminates the need for a schemaDb, but I don't understand how they're related at all. Maybe we should talk?

So I tested following locally that remove need to choose pullmerge method. 

**FastForward**: In this method we apply changeset on top of local changes but abort on any directly change conflict. Indirect changes conflict does not cause failure they are REPLACED.  Schema changes cannot be applied using FastForward as they bound to cause conflict.

1. If there was local changes
   1. For each changeset attempt to do fast-forward. In fastforward we fail on any direct change conflict.
   2. Any remaining changesets that could not be applied using fastforward will be applied using rebase. 
2. If there was no local changes (do what we do now during merge)
   1. Our imodel history on hub has many direct change conflicts beside indirect. We know this from tons of conflict logging when v2 checkpoints are created. So we are not going to change anything about this workflow. The new method will in future push clean changesets with zero conflict when applied during v2 checkpoint creation.
   
Above simplify things. App does not need to decide the merge method. Its all internal.

docs/learning/backend/PullMerge.md
```diff
@@ -0,0 +1,165 @@
...
+
+|  | Operation | Conflict description                                                   | Type         |
+|--|-----------|------------------------------------------------------------------------|--------------|
+|  | `INSERT`  | 1. PRIMARY KEY already exists                                          | `Conflict`   |
+|  |           | 2. Database constraint voliation e.g. UNIQUE, CHECK                    | `Constraint` |
+|  | `DELETE`  | 1. PRIMARY KEY does not exist                                          | `NotFound`   |
+|  |           | 2. PRIMARY KEY does exist but other fields values does not match       | `Data`       |
+|  |           | 3. Database constraint voliation e.g. UNIQUE, CHECK caused by update   | `Constraint` |
+|  | `UPDATE`  | 1. PRIMARY KEY does not exist                                          | `NotFound`   |
+|  |           | 2. PRIMARY KEY does not exist but data fields values does not match    | `Data`       |
```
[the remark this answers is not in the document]
khanaffan: Changeset is map of changes using its table name and primary key. PRIMARY KEY does not exists mean the change row has primary key which was not found in the database to which the change was applied.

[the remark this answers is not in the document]
khanaffan: corrected.

core/backend/src/TxnManager.ts
khanaffan: fixed

core/backend/src/TxnManager.ts
khanaffan: fixed

common/api/core-backend.api.md
```diff
@@ -5944,6 +5976,16 @@ export class TxnManager {
...
+    // @internal (undocumented)
+    protected _onRebaseLocalTxnConflict(internalArg: DbRebaseChangesetConflictArgs): DbConflictResolution;
+    // @internal (undocumented)
+    readonly onRebaseTxnBegin: BeEvent<(txn: TxnArgs) => void>;
+    // @internal (undocumented)
+    protected _onRebaseTxnBegin(txn: TxnArgs): void;
+    // @internal (undocumented)
+    readonly onRebaseTxnEnd: BeEvent<(txn: TxnArgs) => void>;
+    // @internal (undocumented)
+    protected _onRebaseTxnEnd(txn: TxnArgs): void;
```
pmconne, line 5988: TxnManager isn't intended to be subclassed. If you make these private instead of protected then @internal is not needed.

docs/learning/backend/PullMerge.md
```diff
@@ -0,0 +1,154 @@
+# Pull merge & conflict resolution
```
pmconne, line 1: This is a very informative article. I see no links to it - how are people supposed to find it? Please add some links from appropriate API docs.

pmconne (approved):

docs/learning/backend/PullMerge.md
khanaffan: I think this article need to be referenced from https://www.itwinjs.org/learning/backend/concurrencycontrol/#optimistic-concurrency-control and that doc has to be updated as its content also now out dated.

pmconne: Ok, let's leave it for a follow-up PR.

common/api/core-backend.api.md
khanaffan: I made the callbacks that I added to private.

docs/changehistory/NextVersion.md
```diff
@@ -127,3 +128,30 @@ As of iTwin.js 5.0, the following packages have been removed and are no longer a
 | ------------------------------ | --------------------------------------------------------- |
 | `@itwin/core-webpack-tools`    | We no longer recommend using [webpack](https://webpack.js.org/) and instead recommend using [Vite](https://vite.dev/). |
 | `@itwin/backend-webpack-tools` | We no longer recommend webpack-ing backends, which was previously recommended to shrink the size of backends. |
+
+### Change to pullMerge
```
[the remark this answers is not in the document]
khanaffan: I added the link to nextversion.md

tcobbs-bentley (approved):
The mobile changes simply update the native version, and look fine.

aruniverse (approved):
approving addon ver bump changes

PR #7469, opened 2024-12-09
[breaking] drop custom webpack pkgs

aruniverse:
As of iTwin.js 5.0, the following packages have been removed and are no longer available:

| Removed                        | Replacement                                               |
| ------------------------------ | --------------------------------------------------------- |
| `@itwin/core-webpack-tools`    | We no longer recommend using [webpack](https://webpack.js.org/) and instead recommend using [Vite](https://vite.dev/). |
| `@itwin/backend-webpack-tools` | We no longer recommend webpack-ing backends, which was previously recommended to shrink the size of backends. |


aruniverse:
@tcobbs-bentley can you verify the dta ios app runs fine with the webpack changes in here?

tcobbs-bentley:
> @tcobbs-bentley can you verify the dta ios app runs fine with the webpack changes in here?

I will check.

pmconne (approved):

ben-polinsky (approved):

Issue #7441, opened 2024-12-02
Allow selecting models and sub-categories in `SelectionSet`


grigasp:
**Is your feature request related to a problem? Please describe.**

Some components allow selecting Geometric Model and Category instances. When that happens, we add relevant Model / SubCategory ids to `HiliteSet` to hilite elements in selected model or category. However, `SelectionSet` only allows adding geometric element ids, so we end up with the situation where users think that something is selected (because something is hilited), but the tools don't see that (because `SelectionSet` is empty). For example, the "Clear selection" tool doesn't even show up, because it doesn't get notified that something is selected. Even if it did, the "Clear" action on `SelectionSet` would be a no-op, because it's empty.

**Describe the solution you'd like**

Make `SelectionSet` store selection in a similar structure as `HiliteSet`. This way, when a Model / Category is selected, we can add Model / SubCategories to `SelectionSet`, like we do in `HiliteSet`.

- Adding such an id would trigger the `onChanged` event and affect the `empty`, `size`, `emptyAll` methods.
- The `elements` getter and existing `SelectionSet` modification methods would continue working as before.
- We'd add `models` and `subCategories` getters to modify and access selected Model / SubCategory ids.
- Existing tools would only be aware of elements' selection, but could be enhanced to also consider Model / Category selection.

Things to consider with this solution:

- Changing `ElementSetTool.setPreferredElementSource` to also check if any models or sub-categories are selected when there are no geometric elements selected and outputting an informational message or something to let users know why hilited stuff is rejected.
- Make tools work with models and sub-categories:
  1. Provide a way to change Model / SubCategory selection to Element selection.
  2. Enhance the tools to work with Models / SubCategories as first class citizen. IMO, this should be preferred over the above, because handling a "model" could be made much more efficient than handling potentially millions of models elements directly.

**Describe alternatives you've considered**

- When a Model / Category is selected, add their elements to `SelectionSet`.
  
  Reasons rejected: 
  - This could potentially add very large sets of elements to `SelectionSet`, which could cause performance issues.
  - Loss of context.
  - Difficulties synchronizing `SelectionSet` with unified selection storage.

- Enhance `SelectionSet` to have something like "selection mode", which would default to Element, but could also be Model or Category. The mode would tell what the selected ids mean, changing mode or adding ids with different mode would clear existing ids.

  Reason rejected: 
  - This would only support one type of ids being selected at a time, which is not flexible enough - our UI components support selecting multiple types of ids at once.

[...]

PR #7478, opened 2024-12-11
Support Models' and SubCategories' selection in `SelectionSet`

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/7441.

**From `NextVersion.md`:**

There are two similar selection-related concepts in `@itwin/core-frontend` - `SelectionSet` and `HiliteSet`. The former is generally used by interactive tools (e.g. the "Move element" tool), so it contains what tools think is selected. The latter is used by the graphics system to know what elements to highlight, so it contains what users think is selected. Generally, we want the two sets to be in sync to avoid confusion why tools act on different elements than what users think are selected. Keeping them in sync was not always possible, because `HiliteSet` may store Model and SubCategory ids, but `SelectionSet` could only store Element ids. So we could end up in situations where a Model id is added to `HiliteSet` and `SelectionSet` is empty, making users think that all elements in that model are selected, but tools not knowing anything about it.

To alleviate this problem, the `SelectionSet`-related APIs have been enhanced to support storing Model and SubCategory ids, similar to what `HiliteSet` does. The change has been made in a backwards compatible way, so all existing code using `SelectionSet` should continue to work as before:

- `SelectionSet` modification methods `add`, `addAndRemove`, `remove`, `replace` now, in addition to existing `Id64Arg` argument, accept the `SelectableIds` structure.
- `SelectionSetEvent` attributes `added` and `removed` have been deprecated, but continue to work as before, containing only element ids. In addition, the event object now contains new `additions` and `removals` attributes, which are instances of `SelectableIds` and contain all ids that were added or removed from the selection set, including those of Model and SubCategory.

Because the `SelectionSet` now stores additional types of ids, existing code that listens to `onChange` event may start getting extra invocations that don't affect the element selection (e.g. `SelectAddEvent` with `added: []` and `additions: { models: ["0x1"] }`). Also, the `isActive` getter may return `true` even though `elements` set is empty.


common/api/core-frontend.api.md
```diff
@@ -10654,9 +10658,21 @@ export class SectionDrawingModelState extends DrawingModelState {
     static get className(): string;
 }
 
+// @public
+export interface SelectableIds {
```
pmconne, line 10662: Why make every property required here if you're going to use `Partial<SelectableIds>` everywhere in the API?

grigasp: Agree, makes no sense. Made the members optional.

pmconne (approved):

core/frontend/src/SelectionSet.ts
```diff
@@ -209,233 +176,559 @@ export class HiliteSet {
...
+
+  /** The IDs of the selected subcategories.
+   * @note Do not modify this set directly. Instead, use methods like [[SelectionSet.add]].
+   */
+  public get subcategories(): Set<Id64String> {
+    return this.#selection.subcategories;
+  }
+
+  /** Get the active selection as a collection of geometric element, model and subcategory ids.
+   * @note Do not the sets in returned collection directly. Instead, use methods like [[SelectionSet.add]].
```
bbastings, line 351: nitpick...missing "modify" in note...

bbastings (approved):

saskliutas (approved):

core/frontend/src/SelectionSet.ts
grigasp: fixed

core/frontend/src/test/SelectionSet.test.ts
```diff
@@ -0,0 +1,1191 @@
...
+      expect(onChangedSpy).lastCalledWith({
+        type: SelectionSetEventType.Add,
+        set: ss,
+        added: expect.toBeId64Arg("0x1"),
+        additions: {
+          elements: expect.toBeId64Arg("0x1"),
+          models: expect.toBeId64Arg(["0x2"]),
+          subcategories: expect.toBeId64Arg(["0x3"]),
+        },
+      });
```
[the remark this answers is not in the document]
grigasp: These are two different things and they won't necessarily match. 

The event is raised with changes (e.g. if you call `add` with `["0x1", "0x2"]` and `0x1` is already in the selection set, then the event will only have `0x2` in `additions.elements`.

There are a number of tests where event arguments don't match selection set content. Take `adds element ids to selection set` as an example - we add `0x123` element id, the event is raised with it, but the selection set also contains a model, a subcategory and a different element.

JonasDov (approved):

PR #7488, opened 2024-12-13
Changes to make it easier to implement FrameworkAccuDraw...

bbastings:
...and support bearing direction formatting.

FrameworkAccuDraw should be changed as follows:

```
public override onMotion(_ev: BeButtonEvent): void {
  if (UiFramework.isContextMenuOpen)
    return;
  this.processMotion();
}

FrameworkAccuDraw.getFieldDisplayValue = (index: ItemField): string => {
  return IModelApp.accuDraw.getFormattedValueByIndex(index);
}

AccuDrawInputField.tsx needs to be changed to correctly manage the keyin status :

  React.useEffect(() => {
    return FrameworkAccuDraw.onAccuDrawSetFieldValueToUiEvent.addListener(
      (args) => {
        if (args.field === field && stringValue !== args.formattedValue) {
          setStringValue(args.formattedValue);
+         AccuDrawShortcuts.itemFieldNewInput(field);
          if (isFocusField) setNeedSelection(true);
        }
      }
    );
  }, [field, isFocusField, stringValue]);

```
When the text field accepts the new value, ex. on focus out, it needs to call AccuDrawShortcuts.itemFieldCompletedInput(field). Not sure exactly where to add this...

[Addressing use of internal AccuDraw.vector](https://github.com/iTwin/itwinjs-backlog/issues/1326#issuecomment-2528536738)

To support showing a bearing direction instead of angle, override getAngleFormatter and getLengthFormatter to use the same quantity type currently specified by getAngleParser and getLengthParser.

When initiializing AccuDraw, set IModel.accuDraw.bearingFixedToPlane2d = true to use only XY for bearing direction in a 3d model and ignore Z.

It is no longer necessary to override isBearingMode as default implementation now checks the format type from the angle parser.


bbastings:

pmconne (approved):

PR #7508, opened 2024-12-21
Reinstate Logger.staticMetaData

pmconne:
This `@internal` API was removed in #6828. Reinstate a constrained version of it for people like @c-macdonald who were using the internal version.


pmconne:
@grigasp fyi this test in presentation-full-stack-tests flaked out on Windows:
```
  1) PresentationManager
       Cancel request
         cancels 'getNodes' request:
     AssertionError: expected promise to be rejected with 'PresentationError' but it was fulfilled with [ { key: { version: 2, …(5) }, …(3) } ]
```

Issue #7489, opened 2024-12-14
Quantity parser parses Bearing angle incorrectly in some cases.


mathieu-fournier:
## parsing an incomplete Bearing
Parsing an incomplete Bearing angle result in a wrong value. 
Example : 
When typing a bearing it usually goes like this. Step by step : 
`S`
`S4`
`S45` 
... 
`S45°00'00"E`

When parsing the first two letters `S4`, which is an incomplete Bearing because we dont know the direction yet `East` or `West`, the Parser return a value that's wrong : `North something...`

I think the result should be an error. 

## parsing an bearing with lowercase letters 
When parsing a bearing with lowercase letters, it also returns a wrong value. Ex: `s45°00'00"e` -> result : `North something...`

I think the result should be a valid bearing. 

Thanks

rschili:
I agree.
In addition, I have seen several special formats which are considered valid Bearings, at least some people type it like this:
- N (stands for N00°E
- NE (stands for N45°E)
- NW (N45°W)
I'm a bit conflicted how smart we make this, and if we live-parse the string several times while the user is typing, at certain points in the string, it may represent a valid bearing, while others are invalid.
Are we good if we follow the rule "Once there is a number specified, it only is a valid value, if the suffix is there"?

Naron will be picking this up.

naronchen:
Hi,

the existing code does return ParseError when parsing incomplete bearing string
[itwinjs-core/core/quantity/src/test/BearingAzimuth.test.ts at 46b679a54bd5447b6a2258cf7473eb2b32a843ee · iTwin/itwinjs-core · GitHub](https://github.com/iTwin/itwinjs-core/blob/46b679a54bd5447b6a2258cf7473eb2b32a843ee/core/quantity/src/test/BearingAzimuth.test.ts#L146)

i just tested putting in "S" and "S4" and it does return an ParseError of missing prefix or suffix. 

Which api was it being used that returned `North something...`?

mathieu-fournier:
After some reflection, I'm not sure that it returns North something. The AccuDraw input field displays North something tho.

hl662:
Closing this, parsing bearing with lowercase letters was introduced in https://github.com/iTwin/platform-bentley-community/issues/121

[...]

PR #7513, opened 2024-12-23
handling special bearing format + lower/upper casing + tests on invalid input

naronchen:
#7489 

In addition addressing casing and special strings, I added tests for out-of-bound numbers/possible invalid inputs.

Couple observations:

1. incomplete bearing works if just missing numbers
```
N45E // returns 45 degree
N45:30E // returns 45.5 degree
```
these work as treating the missing sections as 00

2. ignores invalid symbols/numbers
```
N45:00@!$:00 // returns 45 degree
N45:0030:00 // returns 45.5 degree
```

3. allows first number to have negative sign but return ParseError.MathematicOperationFoundButIsNotAllowed for negative sign being on other numbers (this has to do with how Parser.isMathematicOperation is written, it skips the first element)
 ```
N-45:00:00 // returns 315 degree
N45:00:-30 // returns ParseError
```

4, allows numbers bigger than 90 to occur (tests added in this pr)
```
N100:00:00E // returns 100 degree
```

The parseAndProcessTokens implementation is relatively loose, which seems acceptable to me. Let me know if any behavior is unexpected.

### CODE REVIEW

File: core/quantity/src/Parser.ts [hunk 1]
```diff
@@ -712,29 +723,38 @@ export class Parser {
...
       matchedSuffix = DirectionLabel.East;
-    } else if (inString.endsWith(DirectionLabel.West)) {
+    } else if (inString.toUpperCase().endsWith(DirectionLabel.West)) {
       inString = inString.substring(0, inString.length - DirectionLabel.West.length);
       matchedSuffix = DirectionLabel.West;
     }
 
+    // check if the remaining string is a special direction
+    if (inString.trim() === "") {
+      const prefix = matchedPrefix?.toLowerCase() || "";
```
hl662, line 745: seeing how matchedPrefix/matchedSuffix is an enum returning upper case characters, why not avoid having to convert to lowercase here, and change the keys of specialDirections to be all uppercase instead?

core/quantity/src/Parser.ts
```diff
@@ -712,29 +723,38 @@ export class Parser {
...
+        return { ok: true, value: specialDirection };
+    }
+
     if (matchedPrefix === null || matchedSuffix === null) {
       return { ok: false, error: ParseError.BearingPrefixOrSuffixMissing };
     }
 
     const parsedResult = this.parseAndProcessTokens(inString, spec.format, spec.unitConversions);
-    if(this.isParseError(parsedResult) || !parsedResult.ok) {
+    if(this.isParseError(parsedResult)) {
```
hl662, line 757: maybe also remove the RHS of the `||` at the start of `parseAzimuthFormat()` as well, to be consistent

naronchen:
@hl662 @rschili this one is ready to be reviewed

hl662 (approved):

Issue #7487, opened 2024-12-13
Quantity parser returns `NaN` instead of an Error in some edge cases.


mathieu-fournier:
When parsing two dots `..` or two comas `,,`  the length formatter return `NaN` instead of an error. 
I am parsing using the engineering length parser, in imperial. 
I suggest that whenever we parse a `NaN` value, we convert the result to an error instead.

rschili:
By error, do you mean returning a ParseError?
I agree, for parseRatioFormat this is already what we do. We should streamline the behavior across the code.

We also still have the split between throwing a QuantityError and returning (not throwing) a ParseError.
The reasoning there is: QuanityError is for unexpected things, like misconfiguration or broken metadata while ParseError is for expected things (we expect parsing to fail frequently with unvalidated inputs). I think keeping it like that makes sense.

naronchen:
I think async method Parser.parseIntoQuantity right now returns a QuantityProps which dont contain info on ParseError. It does return whether the returned Quantity is Valid

```
export interface QuantityProps {
  readonly magnitude: number;
  readonly unit: UnitProps;
  readonly isValid: boolean;
}
```

the non async method parseQuantityString does return the ParseQuantityError though

For the Parser.parseIntoQuantity method, i tested a couple inputs
Invalid input with only symbols
```
      ".", // return true, NaN
      ",", // return true, 0
      "..", // return true, NaN
      ",,", // return true, 0
      "!$#$%%^", // return true, 0
```
the boolean here is QuantityProps.isValid
So for these cases, probably returning `false` is desired?

Also by the way, I tested the invalid mix of numbers and symbols
```
,123 // return true, magnitude: 123
.1232. // return true, magnitude: 0.1232
.12323.2323 // return true, magnitude: 12323
..123 // return true, magnitude: Nan
.1232!@#$ // return true, magnitude: 1232
```

seemed like its able to extract the number part for most cases and returns true on QuantityProps.isValid.
Let me know if thats desired.

mathieu-fournier:
Note, I'm not 100% certain that it returns NaN. 
The AccuDraw input fields displays NaN. 

When an error is returned, the AccuDraw input field handles it correctly, it ignores the value and uses another valid one. 
When typing two dots `..` something else unexpected occurs because NaN is displayed.

hl662:
[...]

PR #7514, opened 2024-12-24
handling commas and dots input in parseToken of Parser

naronchen:
#7487 

Right now I have any input consists only special characters returning a parseError or a quantity with isValid field being false. 
 
 this should be all the special characters.
 /^[!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]+$/;
 
Let me know if thats expected, or we wanna handle just the dots/comas occurrences


hl662 (approved):
@mathieu-fournier I'll let you chime in on whether we should check only commas and dots, or check all special characters (and if the special characters mentioned are enough or we missing any)

naronchen:
Just pushed a different fix because Mathieu asked to handle multiple dots and comas mixed into numbers
Now I got these tests cases returning parseError.UnableToConvertParseTokensToQuantity
```
      "10..",
      "1.2,,3..4...7",
      ",,3..",
      "12..34",
      "1.2.3",
      "..10",
      "1...2",
      "1..",
      "1..,",
      "10,20,30..40",
      "1..e2",
      "1e..2",
      "1e2..",
      "1...2e3",
      "1e2..,3",
```

the ones with comas is a bit tricky
```
     "1,,2", // returns 12
     "1.,2", // returns 1.2
```
 the parsing skips comas because the parsing method thinks its a thousands separator so it ignores it, will look into it more.

rschili:
> the ones with comas is a bit tricky
> 
> ```
>      "1,,2", // returns 12
>      "1.,2", // returns 1.2
> ```
> 
> the parsing skips comas because the parsing method thinks its a thousands separator so it ignores it, will look into it more.

Oh, but if the algorithm detects this as thousands separator, it should not just strip it but instead consider its meaning.
I cannot find any notion that can be considered a standard, were for example "1,," would be used to represent "1,000,000"
"1,,2" probably does not represent a valid number in any standard. Neither does "1,"
I'd vote to fail here...

naronchen:
@rschili @hl662 just pushed a fix handling multiple comas case. it passes all the tests but I just dont feel very secure with regular expression solutions. Couldnt think of a better fix at the moment. Let me know what you think.

(cpp side behaves similarly with ignoring comas, I can work on that afterwards)

PR #7531, opened 2025-01-08
Remove re-export of core-bentley apis in core-common 

aruniverse:
Remove the re-exports from core-bentley in core-common (/core/common/src/IModelError.ts)
React to changes

### DIFF
--- a/docs/changehistory/NextVersion.md
+++ b/docs/changehistory/NextVersion.md
@@ -23,6 +23,7 @@ Table of contents:
     - [@itwin/core-backend](#itwincore-backend-1)
     - [@itwin/appui-abstract](#itwinappui-abstract)
     - [@itwin/core-electron](#itwincore-electron)
+  - [API removals](#api-removals)
   - [Packages dropped](#packages-dropped)
 
 ## Selection set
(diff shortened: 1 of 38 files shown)


docs/changehistory/NextVersion.md
```diff
@@ -119,11 +121,29 @@ All three `nativeDb` fields and `IModelHost.platform` have always been `@interna
 | `ElectronApp.callDialog`            | [ElectronApp.dialogIpc]($electron)                        |
 | `ElectronHost.getWindowSizeSetting` | [ElectronHost.getWindowSizeAndPositionSetting]($electron) |
 
+### API removals
+
+The following APIs have been removed:
```
tcobbs-bentley, line 126: I feel like this is misleading (in a way that could easily alarm users). These APIs weren't exactly removed; it's just that they are no longer re-exported. Even if someone reads this whole section, I'm not sure it's obvious what was really done.

tcobbs-bentley (approved):

docs/changehistory/NextVersion.md
aruniverse: Yea i wasnt sure the best way to categorize this. I will mention we no longer re-export these apis.

wgoehrig (approved):

grigasp (approved):

pmconne (approved):
These are breaking changes to public, non-deprecated APIs. You can get away with them for 4.x -> 5.0. Such changes would not fly in a world where evergreen Mango apps are officially released.

aruniverse:
> These are breaking changes to public, non-deprecated APIs. You can get away with them for 4.x -> 5.0. Such changes would not fly in a world where evergreen Mango apps are officially released.

Yes this is definitely breaking, and we gave no heads up. I didn't see a clean way to deprecate this previously, or one that fit in our existing support policy. I agree these kind of changes can't fly post 5.0, so we need to act now if there are others

PR #7539, opened 2025-01-10
Fix copyImageToCanvas bug and add ReadImageToCanvasOptions

andremig-bentley:
This PR is to address this issue: https://github.com/iTwin/itwinjs-backlog/issues/1310

This PR has two main goals. First, to address the bug outlined in the above issue. Currently, when there is only a single viewport open and that canvas is copied, such as with `copyImageToCanvas`, canvas decorations do not show up on the resulting image. This is because the single view is rendered onto an on-screen webgl canvas for higher performance, while the canvas decorations were rendered onto a separate 2d canvas then laid on top. When saving, only the webgl canvas was being copied. This bug is fixed by combining the two canvases before saving them in this single viewport case.

Secondly, the issue above lays out a desire for a  way to control whether or not decorations are in the image saved through `Viewport.ReadImageToCanvas`. This is accomplished through the addition of `ReadImageToCanvasOptions`. If undefined, or if `ReadImageToCanvasOptions.includeCanvasDecorations` is `false`, then the original behavior will persist. If `includeCanvasDecorations` is set to true, then the saved image will include the canvas decorations.

Also included in this PR is the `SaveImageWithCanvasDecorations Tool` for Display Test App. This tool gives an example of how to use the new `ReadImageToCanvasOptions` by allowing you to save an image of the canvas with the decorations either on or off.


pmconne:
> This bug is fixed by combining the two canvases before saving them in this single viewport case.

Arguably, you have introduced a different bug by changing the existing behavior. Do we know if anyone was relying on the existing behavior?

markschlosseratbentley:
> > This bug is fixed by combining the two canvases before saving them in this single viewport case.
> 
> Arguably, you have introduced a different bug by changing the existing behavior. Do we know if anyone was relying on the existing behavior?

Yes, Design Review, who was relying on this bug and their own "internal API usage to work aroun

> > This bug is fixed by combining the two canvases before saving them in this single viewport case.
> 
> Arguably, you have introduced a different bug by changing the existing behavior. Do we know if anyone was relying on the existing behavior?

Based on the GitHub issue and discussions: basically Design Review relies on this bug to be able to disable canvas-decorations-only when reading a viewport.  However, this bug-reliance does not work if more than one viewport exists, so it's not a good behavior to rely on anyway.  The bug only happens in the single viewport case.

The idea here was to fix the bug so they no longer have to work around the bug by calling internal API and messing around with our direct-to-screen logic from their app.  Then, we provide them with a way to truly disable canvas decorations as they desired.

pmconne:
>  Then, we provide them with a way to truly disable canvas decorations as they desired.

If they truly only care about canvas decorations (I think there's some confusion about what is a canvas decoration and what is not), a dead-simple solution = 
```ts
interface ReadImageToCanvasOptions {
  includeCanvasDecorations?: boolean;
}

class Viewport {
  public readImageToCanvas(options?: ReadImageToCanvasOptions): HTMLCanvasElement { ... }
}
```

If `includeCanvasDecorations` is undefined then existing behavior is preserved.

I would be reluctant to provide a `Viewport.disableCanvasDecorations` because canvas decorations include logos (such as Google Maps) that we are required contractually to display. We don't want to make it easy for applications to turn those off indefinitely.

markschlosseratbentley:
> If they truly only care about canvas decorations (I think there's some confusion about what is a canvas decoration and what is not)

@ChrisBAtBentley @jason-crow Can you verify our understanding?

andremig-bentley:
The changes necessary for the `canvasDecoration` view flag have been removed. 

@jason-crow can you confirm whether the simple solution outlined in [this](https://github.com/iTwin/itwinjs-core/pull/7539#issuecomment-2583479653) comment above will satisfy what you need? Or, do you think you need something with more control over the decorations akin to the view flags solution (although not the view flags solution as discussed above)? Let me know if you have any other questions.

jason-crow:
@markschlosseratbentley @pmconne apologies for the delay....

The proposed solution for allowing a prop to specify whether to include decorations:
```typescript
interface ReadImageToCanvasOptions {
  includeCanvasDecorations?: boolean;
}

class Viewport {
  public readImageToCanvas(options?: ReadImageToCanvasOptions): HTMLCanvasElement { ... }
}
```
Is a perfect solution for us. We do not actually want to turn off decorations in the viewport, we just want a way to capture a screenshot of the viewport without them using readImageToCanvas. This is preferred for us since a saved view has no mechanism to restore the forms' markers that were possibly active when the screenshot was taken, so ideally the saved view thumbnail would keep them hidden so it doesn't give the user a false expectation about how it will appear when applied.

Since there was a bug that prevented the decorations from showing when a single viewport was active, versus 2, I don't really care if the default is for includeCanvasDecorations to be off or on, since I suppose either way there is a change in behavior; however, the suggested solution of making includeCanvasDecorations off by default probably preserves the most since most consumers likely work with a single viewport. 

Thanks for the work on this!

andremig-bentley:
As of the latest commit, this PR now includes the proposed `ReadImageToCanvasOptions` solution, as well the `SaveImageWithCanvasDecorations Tool` in display test app which provides an example of how to use the new options.

test-apps/display-test-app/public/locales/en/SVTTools.json
```diff
@@ -55,6 +55,9 @@
     "SaveImage": {
       "keyin": "dta save image"
     },
+    "SaveImageWithCanvasDecorations": {
```
eringram, line 58: Did you consider adding the option to turn canvas decorations on/off to the existing SaveImageTool, rather than creating a new one? It slightly increases the complexity of that tool with a new argument, but I feel like it logically makes more sense than a new one

eringram (commented):
core-frontend changes look good to me. Another thing I'd suggest is updating DTA docs for any tool changes

test-apps/display-test-app/public/locales/en/SVTTools.json
andremig-bentley: I considered it, the main reason I rejected it was the added complexity to the original tool and the fact that the url is acquired in a different way in each tool (lines 43 and 118). That being said, it's not at all impossible to combine the tools, so I can definitely make that change if needed.

andremig-bentley: As of latest commit, the `SaveImageWithCanvasDecorations` tool has been removed, and the option to include canvas decorations within the existing `SaveImage` tool has been added.

andremig-bentley:
> core-frontend changes look good to me. Another thing I'd suggest is updating DTA docs for any tool changes

DTA docs have been updated to include the new include decorations functionality on the existing save image tool

test-apps/display-test-app/public/locales/en/SVTTools.json
markschlosseratbentley: @eringram Please resolve comment if this looks good to you.

markschlosseratbentley (changes requested):
I think this probably warrants a NextVersion.md entry describing the changes to behavior / flags.

andremig-bentley:
> I think this probably warrants a NextVersion.md entry describing the changes to API behavior / flags.

A NextVersion.md entry has been added.

markschlosseratbentley (changes requested):
Can you add a simple test or alter an existing test to use this new flag?

core/frontend/src/Viewport.ts
```diff
@@ -262,6 +262,16 @@ export interface ReadPixelsArgs {
   excludedElements?: Iterable<Id64String>;
 }
 
+/** Arguments supplied to [[Viewport.readImageToCanvas]].
+ * @public
+ */
+export interface ReadImageToCanvasOptions {
+  /** If defined, canvas decorations will be included in the saved image.
+   * @note This only affects single viewport applications. For multi-viewport applications, the canvas decorations are always included.
```
pmconne, line 270: No. Make the currently inconsistent behavior consistent.

andremig-bentley: Do you mean when `includeCanvasDecorations` is undefined, remove decorations for both single and multi viewport applications?

pmconne: If I say `{ includeCanvasDecorations: false }`, the canvas decorations should be omitted, regardless of how many viewports I have open. Your `@note` says they will be included even if I tell you to omit them, if more than one viewport is open.

pmconne: Is your note wrong? The comment above it is wrong (it claims any value other than `undefined` will cause the decorations to be included, but the property is a boolean).

andremig-bentley: The comment above it is wrong, should say if true they will be included. The note was correct with the current behavior - basically, if `includeCanvasDecorations !== true` then current behavior will persist, this meaning if one viewport, no decorations, if multiple viewports, decorations included. You're right that this is inconsistent, working on changing it to: 
if `includeCanvasDecorations === true`, then canvas decorations will be included, regardless of number of viewports. Otherwise, no canvas decorations will be included, regardless of number of viewports.

pmconne: My recommendation is to make `includeCanvasDecorations` a non-optional property. The new `options` argument to `readImageToCanvas` is necessarily optional. So if you don't supply any options, you get the current dumb behavior for backwards compatibility. If you supply the options, you must choose whether or not to include canvas decorations. Then you only need to document the dumb "how many viewports" logic in one place (on `readImageToCanvas`).

pmconne: Probably better:
```ts
  /** [document dumb behavior here].
   * @deprecated in 5.0 Use the overload accepting a ReadImageToCanvasOptions.
   */
  public readImageToCanvas(): HTMLCanvasElement;
  /** [document sane behavior here] */
  public readImageToCanvas(options: ReadImageToCanvasOptions): HTMLCanvasElement;
  /** @internal */
  public readImageToCanvas(options: ReadImageToCanvasOptions | undefined): HTMLCanvasElement {
    // implementation goes here.
  }
```

Then you can make `includeCanvasDecorations` optional and give it a sane default behavior (include the decorations by default, in which case you should probably invert its name to `omitCanvasDecorations` since undefined is falsy), and we'll be able to remove the dumb behavior eventually.

core/frontend/src/test/Viewport.test.ts
```diff
@@ -716,4 +717,104 @@ describe("Viewport", () => {
...
+    afterEach(() => {
+      for (const dec of activeDecorators) {
+        IModelApp.viewManager.dropDecorator(dec);
+      }
+
+      activeDecorators.length = 0;
+    });
+
+    it("should not include canvas decorations if omitCanvasDecorations is true", () => {
+      const vp = createViewport();
```
andremig-bentley, line 770: The goal of these tests is to: Open a viewport, place a canvas decoration, use readImageToCanvas to get a copied canvas, check if the decoration is still there. This works and passes when the decoration is supposed to be included in the saved image (when omitCanvasDecorations is false). I am looking for feedback for the case in which the decorations should not be included in the saved image. 

That result is relying on the fact that we render directly to the webgl canvas for a single viewport, and for decorations to be included in the saved image, the overlaying canvas must be combined with the webgl canvas. Despite no other viewports present and no combination taking place (confirmed through debug), the decorations remain present in this case within the test, meaning the test is not using the standard single viewport behavior.

Any suggestions on forcing the test to use the direct render to screen methodology? Other things I've tried are written out in a comment in the code, but I'll add them here as well:

     * Use openBlankViewport to get a blank viewport instead of creating a screen viewport directly
     * Set vp.rendersToScreen to true before calling vp.renderFrame (for both vp types)
     * Use vp.target.SetRenderToScreen to do the same thing
     * Making sure all Viewports are removed before adding a new one at the beginning of the test
     * Fully shutting down IModel App and then restarting it at the beginning of the test

andremig-bentley:
Still To Do:

- [ ] Unit Tests
- [ ] Clean up readImageToCanvas declarations/implementations
- [ ] Clean up docs/readme/nextversion

core/frontend/src/test/Viewport.test.ts
andremig-bentley: All tests have been fixed and are passing in latest commit.

core/frontend/src/Viewport.ts
andremig-bentley: This implementation has been added in the latest commit

andremig-bentley:
> Still To Do:
> 
> * [x]  Unit Tests
> * [x]  Clean up readImageToCanvas declarations/implementations
> * [x]  Clean up docs/readme/nextversion

All pending to-dos have been taken care of.

core/frontend/src/Viewport.ts
```diff
@@ -262,6 +262,14 @@ export interface ReadPixelsArgs {
   excludedElements?: Iterable<Id64String>;
 }
 
+/** Arguments supplied to [[Viewport.readImageToCanvas]].
+ * @public
+ */
+export interface ReadImageToCanvasOptions {
+  /** If true, canvas decorations will be included in the saved image. */
+  omitCanvasDecorations: boolean;
```
pmconne, line 270:
```suggestion
  omitCanvasDecorations?: boolean;
```
Don't require people to specify this. Default false.

core/frontend/src/Viewport.ts
```diff
@@ -262,6 +262,14 @@ export interface ReadPixelsArgs {
   excludedElements?: Iterable<Id64String>;
 }
 
+/** Arguments supplied to [[Viewport.readImageToCanvas]].
+ * @public
+ */
+export interface ReadImageToCanvasOptions {
+  /** If true, canvas decorations will be included in the saved image. */
```
pmconne, line 269:
```suggestion
  /** If true, canvas decorations will not be included in the saved image. */
```

pmconne (changes requested):

docs/changehistory/NextVersion.md
```diff
@@ -293,4 +293,6 @@ For more information read [Pull merge & conflict resolution](../learning/backend
...
+Previously, when using [Viewport.readImageToCanvas]($core-frontend) with a single open viewport, canvas decorations were not included in the saved image. Sometimes this behavior was useful, so an overload to [Viewport.readImageToCanvas]($core-frontend) using the new [ReadImageToCanvasOptions]($core-frontend) interface. This now allows the option to choose whether or not canvas decorations are omitted in the saved image. Now, if [ReadImageToCanvasOptions.omitCanvasDecorations]($core-frontend) is true, canvas decorations will be omitted in the saved image.
```
ben-polinsky, line 332:
```suggestion
Previously, when using [Viewport.readImageToCanvas]($core-frontend) with a single open viewport, canvas decorations were not included in the saved image. Sometimes this behavior was useful, so an overload to [Viewport.readImageToCanvas]($core-frontend) using the new [ReadImageToCanvasOptions]($core-frontend) interface was created. This now allows the option to choose whether or not canvas decorations are omitted in the saved image: if [ReadImageToCanvasOptions.omitCanvasDecorations]($core-frontend) is true, canvas decorations will be omitted.
```

ben-polinsky (approved):
Docs look good.

andremig-bentley:
Latest commit adds two tests to ensure correct behavior when `omitCanvasDecorations` is undefined.

markschlosseratbentley (approved):
Thanks for adding the test and NextVersion entry!

core/frontend/src/Viewport.ts
markschlosseratbentley: Is this able to be resolved?

core/frontend/src/render/webgl/Target.ts
```diff
@@ -1093,12 +1095,24 @@ export abstract class Target extends RenderTarget implements RenderTargetDebugCo
     return image;
   }
 
-  public copyImageToCanvas(): HTMLCanvasElement {
+  public copyImageToCanvas(options?: ReadImageToCanvasOptions): HTMLCanvasElement {
+    const vp = IModelApp.viewManager.selectedView;
```
pmconne, line 1099: No, `Target` should not be doing anything with a Viewport, and there's no reason to believe `this` target is associated with the view manager's currently-selected view. If you need access to the Viewport's canvas, the Viewport should pass it in.

andremig-bentley: As of latest commit, the canvas is no longer obtained by the viewport in `copyImageToCanvas`. Instead it is obtained from the viewport and then passed along within the original call to `readImageToCanvas`.

pmconne: Target needs `overlayCanvas` if and only if canvas decorations are supposed to be included, and the only thing the options specify is whether or not to include canvas decorations - right?
So Target should just accept an optional overlayCanvas, no options; argument will be omitted if decorations are to be omitted.

andremig-bentley: This is much cleaner now, done in latest commit.

docs/changehistory/NextVersion.md
```diff
@@ -392,3 +395,11 @@ class MyElement extends Element {
   ...
 }
 ```
+
+## Graphics
+
+### Read Image To Canvas
+
+Previously, when using [Viewport.readImageToCanvas]($core-frontend) with a single open viewport, canvas decorations were not included in the saved image. Sometimes this behavior was useful, so an overload to [Viewport.readImageToCanvas]($core-frontend) using the new [ReadImageToCanvasOptions]($core-frontend) interface was created. This now allows the option to choose whether or not canvas decorations are omitted in the saved image: if [ReadImageToCanvasOptions.omitCanvasDecorations]($core-frontend) is true, canvas decorations will be omitted.
```
pmconne, line 403:
```suggestion
Previously, when using [Viewport.readImageToCanvas]($core-frontend) with a single open viewport, canvas decorations were not included in the saved image. Sometimes this behavior was useful, so an overload to [Viewport.readImageToCanvas]($core-frontend) using the new [ReadImageToCanvasOptions]($core-frontend) interface was [created](https://github.com/iTwin/itwinjs-core/pull/7539). This now allows the option to choose whether or not canvas decorations are omitted in the saved image: if [ReadImageToCanvasOptions.omitCanvasDecorations]($core-frontend) is true, canvas decorations will be omitted.
```

pmconne (approved):

ChrisBAtBentley:
Any ETA as to when this will be in a release?

markschlosseratbentley:
> Any ETA as to when this will be in a release?

This was slated for iTwin.js 5.0 because that was our understanding from the discussion leading up to this development work.  @aruniverse can comment further on that.

aruniverse:
it will be included in 5.0, expecting a rc in the coming days

PR #7540, opened 2025-01-13
Manual backport 2 fixes from master (Column Infos)

rschili:
imodel-native: https://github.com/iTwin/imodel-native/pull/965

backport from PR https://github.com/iTwin/imodel-native/pull/965

The automatic backport would do too much as this fix was bundled with the new ECSql test suite which we do not want to backport.
So this includes only the test adjustments that are needed with the fix.


rschili:
Needs to get https://github.com/iTwin/imodel-native/pull/965 through

aruniverse (approved):

Issue #7541, opened 2025-01-13
Review .updated.json usage


ben-polinsky:
When building and publishing our documentation, we keep a running map of when pages were last updated. To do this, we keep a map of pages->time last updated, publishing it as a build artifact for consumption by the next documentation CI build. 

This has mostly worked, but a couple issues have surfaced:

1. The last updated times do not always match the file's commit history (example to come)
2. When creating new release branches, we must remember to update the download task branch _after_ an artifact is published from that build. For example, in creating a new branch `release/10.x` we initially point the branch to `release/9.x` as that's the most recently published branch. Then, after one run of `release/10.x` which publishes an updated artifact, we can change the branch to release/10.x. Remembering to do this is error-prone. 

@DanRod1999

PR #7542, opened 2025-01-13
update branch to find docs updated artifact

ben-polinsky:
Our CI docs builds are failing on `release/4.10.x` - we never updated the branch in this download artifact task to point to 4.10.x. This can be a bit tricky since the first CI run on a new release branch will obviously not have published an updated `.updated` file. I bet we can come up with a better way to handle this.

In addition, we should also re-look at this workflow as I've noted the "last updated" time on documentation pages can be inaccurate compared to git commit history - I've added #7541.


aruniverse (approved):

MichaelSwigerAtBentley (approved):

aruniverse:
thx ben, lets make sure this step is part of our release docs if its not already

MichaelSwigerAtBentley:
> thx ben, lets make sure this step is part of our release docs if its not already

Its step 7, but the step is strikethroughed, as this is automated (and has been for maybe 4 minors now?). Might need to look into what happened here. 

@DanRod1999 Is the automation in version bump?

DanRod1999:
I believe this part of the finalize release workflow, which calls a update-changelog js file

wgoehrig (approved):

DanRod1999:
@MichaelSwigerAtBentley you were right that this is partly in the version bump pipelines. This commit is directly from a run of the bump pipeline
https://github.com/iTwin/itwinjs-core/blame/release/4.10.x/common/config/azure-pipelines/templates/gather-docs.yaml#L115

However, we do have a snippet of code that is supposed to correct this branch name in the update-changelogs.mjs file
```
// before checking out to target branch
// if it is a major or minor release, we need to update `gather-docs.yaml`'s branchName value to be the release branch
if (commitMessage.endsWith(".0")) {
  const docsYamlPath = "common/config/azure-pipelines/templates/gather-docs.yaml";
  editFileInPlaceSynchronously(docsYamlPath, /release\/\d+\.\d+\.\w+/g, currentBranch);
  // commit these changes to our release branch
  await $`git add ${docsYamlPath}`;
  await $`git commit -m "Update gather-docs.yaml's branch name to the release branch"`;
  await $`git push origin HEAD:${currentBranch}`;
}
```
I think moving this out of the mjs file and finding a better and more reliable way of making this change is probably best. I think something that triggers directly from the build pipeline if packages are release might be a better approach. We already know that this script has problems with triggering on minor versions (i believe due to GH having trouble with triggers on large commits), so that could be a reason why this current logic may not work as anticipated

PR #7546, opened 2025-01-13
[ci] Use pipeline to update gather-docs.yaml during releases

DanRod1999:
Use a script to find and replace the branchName in gather-docs.ymal file. 

We would want to run this script whenever we publish build artifacts on a release branch, this is because the first time we build on a new release branch we target the previous release branch. Once that build publishes artifacts we want to update the yml file to the current release branch.

I created a new yaml file to trigger this, which runs every time we publish an artifact during the itwin.js builds. Don't know if this is the best way to call the script. Maybe just putting it at the end of the release pipeline which has a similar trigger, however that lives in the imodeljs-build-pipeline-scripts repo.

### DIFF
--- a/common/scripts/check_gather_docs.mjs
+++ b/common/scripts/check_gather_docs.mjs
@@ -1,0 +1,70 @@
+import fs from "fs";
+import path from "path";
+
+function checkFile(filePath, currentBranch) {
+  // const filePath = path.normalize("./common/config/azure-pipelines/templates/gather-docs.yaml")
+  if (!fs.existsSync(filePath)) {
+    console.error("File not found.");
+    process.exit(1);
+  }
+
+  const fileContent = fs.readFileSync(filePath, "utf8");
+
+  const textToFind = /branchName: refs\/heads\/release\/\d+\.\d+\.x/gi;
+  const branchInFile = fileContent.match(textToFind);
+
+  if (!branchInFile) {
+    console.error("No match for branch name found");
+    process.exit(1);
+  }
+
+  if (branchInFile[0] === `branchName: ${currentBranch}`) {
+    console.log("The current branch  matches the branch in gather-docs.yaml. No update Needed.");
+    return false;
+  }
+  else {
+    const newContent = fileContent.replace(textToFind, `branchName: ${currentBranch}`);
+    fs.writeFileSync(filePath, newContent, "utf8");
+    console.log("Updated gather-docs.yaml with the current branch.");
+    return true;
+  }
+}
+
+function main() {
+  const filePath = path.normalize("./common/config/azure-pipelines/templates/gather-docs.yaml")
+  const currentBranch = process.env.CURRENT_BRANCH;
+  checkFile(filePath, currentBranch);
+}
+main();
+
+// Tests
+function createTempFile(name, content) {
+  const tempFilePath = path.join('./', `${name}`);
+
+  fs.writeFileSync(tempFilePath, content, "utf8");
+  console.log(`Temporary file created at: ${tempFilePath}`);
+
+  return tempFilePath;
+}
+
+function deleteTempFile(filePath) {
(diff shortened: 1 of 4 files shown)


DanRod1999:
Maybe calling this script in core-build.yaml instead could be better? Then the condition would just be to run this script as long as the package publish step runs? I think I like that idea better than running the script in the release pipeline, but not entirely sold on either.

common/scripts/check_gather_docs.mjs
```diff
@@ -0,0 +1,70 @@
...
+  } else {
+    console.log(`File not found: ${filePath}`);
+  }
+}
+
+function tests() {
+  const currentBranch = "refs/heads/release/1.2.x";
+  // Test for when file needs to be updated
+  let filePath = createTempFile("should-change.txt", "branchName: refs/heads/release/0.0.x");
+  console.assert(checkFile(filePath, currentBranch) === true, "File should be updated");
```
MichaelSwigerAtBentley, line 63: Maybe should check that the file contains the proper branch 'branchName: refs/heads/release/1.2.x'

common/scripts/check_gather_docs.mjs
```diff
@@ -0,0 +1,70 @@
...
+    const newContent = fileContent.replace(textToFind, `branchName: ${currentBranch}`);
+    fs.writeFileSync(filePath, newContent, "utf8");
+    console.log("Updated gather-docs.yaml with the current branch.");
+    return true;
+  }
+}
+
+function main() {
+  const filePath = path.normalize("./common/config/azure-pipelines/templates/gather-docs.yaml")
+  const currentBranch = process.env.CURRENT_BRANCH;
```
MichaelSwigerAtBentley, line 35: Should we also add a quick sanity check here to make sure the new branchName also follows 'refs\/heads\/release\/\d+\.\d+\.x/gi' before we update it?

DanRod1999: Good idea, this should be an easy way to make sure its not master branch or something

common/scripts/check_gather_docs.mjs
DanRod1999: Don't know if this is needed since I created the temp file above, and the contents are hard coded. Lmk if you think otherwise. Unless the intent was to just check the validity of the helper function. But I think thats covered in the checkFile func

common/scripts/check_gather_docs.mjs
DanRod1999: done

common/scripts/check_gather_docs.mjs
```diff
@@ -0,0 +1,70 @@
+import fs from "fs";
+import path from "path";
+
+function checkFile(filePath, currentBranch) {
+  // const filePath = path.normalize("./common/config/azure-pipelines/templates/gather-docs.yaml")
```
MichaelSwigerAtBentley, line 5: Can we remove this comment since this is called in main()? Maybe add a comment to the function explaining that the filePath Param is expected to be normalized

common/scripts/check_gather_docs.mjs
```diff
@@ -33,6 +33,13 @@ function checkFile(filePath, currentBranch) {
 function main() {
   const filePath = path.normalize("./common/config/azure-pipelines/templates/gather-docs.yaml")
   const currentBranch = process.env.CURRENT_BRANCH;
+
+  const textToFind = /refs\/heads\/release\/\d+\.\d+\.x/gi;
+  if (!currentBranch.match(textToFind)) {
+    console.error("Invalid branch name.");
```
MichaelSwigerAtBentley, line 39: Maybe this message should be more specific: "current branch name does not match the expected format"

DanRod1999:
closing as this is no longer a problem. Was resolved in other work

PR #7550, opened 2025-01-14
Clean up internal rendering-related APIs in core-common

pmconne:


core/common/src/TextureMapping.ts
```diff
@@ -74,17 +78,17 @@ export namespace TextureMapping {
     Parametric = 0,
     ElevationDrape = 1,
     Planar = 2,
-    /** @internal */
+    /** Currently unsupported. */
```
aruniverse, line 81: if theyre not supported any reason we just dont remove them?

aruniverse (approved):

core/common/src/TextureMapping.ts
pmconne: They exist. Connectors preserve them. The renderer just currently does not handle them.

pmconne:
Multiple integration test suites failed with "Error: ITwin iModelJsIntegrationTest returned with no data when queried."

PR #7560, opened 2025-01-16
Presentation: Fix Presentation RPC returning details contained within thrown errors

grigasp:
For security reasons we don't want the backend's error response to contain any details about the error. The RPC implementation was converting all errors to a 200 response containing the error message. Now we only do that for `PresentationError` errors, which contain errors that we want to return (e.g. "invalid element id", or "unknown presentation rule", etc.). The generic `Error` errors are just re-thrown to the RPC system, where they can be handled in a generic way.

### DIFF
--- a/common/changes/@itwin/presentation-backend/presentation-fix-backend-returning-error-details_2025-01-16-09-38.json
+++ b/common/changes/@itwin/presentation-backend/presentation-fix-backend-returning-error-details_2025-01-16-09-38.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-backend",
+      "comment": "Fix Presentation RPC returning details contained within thrown errors. Instead of doing that, we now re-throw the error and let RPC system handle it (by re-throwing a generic error instead).",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/presentation-backend"
+}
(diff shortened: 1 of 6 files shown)


paulius-valiunas:
> Now we only do that for `PresentationError` errors, which contain errors that we want to return (e.g. "invalid element id", or "unknown presentation rule", etc.)

I still don't think those should be 200. For example, invalid element id definitely sounds like 400 (bad request)

grigasp:
> > Now we only do that for `PresentationError` errors, which contain errors that we want to return (e.g. "invalid element id", or "unknown presentation rule", etc.)
> 
> I still don't think those should be 200. For example, invalid element id definitely sounds like 400 (bad request)

We'll have to rework all that with the no-rpc changes, which are scheduled for 5.0 timeframe, won't we? If that's the case, I see no point in fixing this in this PR just to throw it await afterwards.

paulius-valiunas:
> > > Now we only do that for `PresentationError` errors, which contain errors that we want to return (e.g. "invalid element id", or "unknown presentation rule", etc.)
> > 
> > 
> > I still don't think those should be 200. For example, invalid element id definitely sounds like 400 (bad request)
> 
> We'll have to rework all that with the no-rpc changes, which are scheduled for 5.0 timeframe, won't we? If that's the case, I see no point in fixing this in this PR just to throw it await afterwards.

RPC will co-exist with no-rpc for a while. We're not targeting to cover the full API surface with 5.0, at least not with the current manpower.

grigasp:
> > > > Now we only do that for `PresentationError` errors, which contain errors that we want to return (e.g. "invalid element id", or "unknown presentation rule", etc.)
> > > 
> > > 
> > > I still don't think those should be 200. For example, invalid element id definitely sounds like 400 (bad request)
> > 
> > 
> > We'll have to rework all that with the no-rpc changes, which are scheduled for 5.0 timeframe, won't we? If that's the case, I see no point in fixing this in this PR just to throw it await afterwards.
> 
> RPC will co-exist with no-rpc for a while. We're not targeting to cover the full API surface with 5.0, at least not with the current manpower.

Okay, scheduled further improvements for 5.0: https://github.com/iTwin/itwinjs-core/issues/7562

paulius-valiunas (approved):

saskliutas (approved):

Issue #7474, opened 2024-12-10
Update TypeScript config to enable `useDefineForClassFields`


GytisCepk:
`useDefineForClassFields` changes behavior of how class properties are initialized ([TypeScript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier)). TypeScript recommends enabling it as it makes TypeScript code closer to ECMA runtime behavior

This breaks following:
- All classes derived from `Entity`:
[...]

PR #3017, opened 2022-01-11
Samw/element geometry builder params

swwilson-bsi:
Allow an app (including a two-process app) to specify an element's geometry in terms of an ElementGeometryDataEntry array, rather than a JSON geometry stream (or a fully cooked geomBinary). This allows the app to define the geometry that goes into the array using ElementGeometry.Builder (rather than GeometryStreamBuilder).

This PR should not be completed until the [PR of the same name](https://bentleycs.visualstudio.com/iModelTechnologies/_git/imodel02/pullrequest/218619) in ADO imodel02 is completed.

PR #7563, opened 2025-01-16
Enable `useDefineForClassFields` TS flag

GytisCepk:
TypeScript recommends setting `useDefineForClassFields` to `true` (default value starting ES2022). However, enabling this flag caused several issues. More info: https://github.com/iTwin/itwinjs-core/issues/7474

Additionally, noticed that some TypeScript types are incorrectly defined (e.g. `AuxCoordSystem2d.angle` is defined as `number`, but in reality it can be a `number`, `object` or `undefined`). This PR fixes some of these.

### DIFF
--- a/core/backend/src/Element.ts
+++ b/core/backend/src/Element.ts
@@ -477,6 +477,7 @@ export abstract class GeometricElement extends Element {
     super(props, iModel);
     this.category = Id64.fromJSON(props.category);
     this.geom = props.geom;
+    this.elementGeometryBuilderParams = props.elementGeometryBuilderParams;
   }
 
   /** Type guard for instanceof [[GeometricElement3d]] */
@@ -707,7 +708,10 @@ export abstract class InformationReferenceElement extends InformationContentElem
 export class Subject extends InformationReferenceElement {
   public static override get className(): string { return "Subject"; }
   public description?: string;
-  protected constructor(props: SubjectProps, iModel: IModelDb) { super(props, iModel); }
+  protected constructor(props: SubjectProps, iModel: IModelDb) {
+    super(props, iModel);
+    this.description = props.description;
+  }
 
   public override toJSON(): SubjectProps { // This override only specializes the return type
     return super.toJSON() as SubjectProps; // Entity.toJSON takes care of auto-handled properties
@@ -880,7 +884,11 @@ export class SheetBorderTemplate extends Document {
   public static override get className(): string { return "SheetBorderTemplate"; }
   public height?: number;
   public width?: number;
-  protected constructor(props: SheetBorderTemplateProps, iModel: IModelDb) { super(props, iModel); }
+  protected constructor(props: SheetBorderTemplateProps, iModel: IModelDb) {
+    super(props, iModel);
+    this.height = props.height;
+    this.width = props.width;
+  }
 }
 
 /** The template for a [[Sheet]]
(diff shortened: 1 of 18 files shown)


core/backend/src/ViewDefinition.ts
```diff
@@ -614,9 +614,13 @@ export class TemplateViewDefinition3d extends ViewDefinition3d {
...
 export abstract class AuxCoordSystem extends DefinitionElement {
   public static override get className(): string { return "AuxCoordSystem"; }
-  public type!: number;
+  public type?: number;
   public description?: string;
-  public constructor(props: AuxCoordSystemProps, iModel: IModelDb) { super(props, iModel); }
+  public constructor(props: AuxCoordSystemProps, iModel: IModelDb) {
+    super(props, iModel);
+    this.type = props.type;
+    this.description = props.description;
```
wgoehrig, line 622: For fields like these (where the type annotation already matches what the `Entity` constructor was setting), why not just use the `declare` keyword instead?

core/backend/src/Element.ts
```diff
@@ -477,6 +477,7 @@ export abstract class GeometricElement extends Element {
     super(props, iModel);
     this.category = Id64.fromJSON(props.category);
     this.geom = props.geom;
+    this.elementGeometryBuilderParams = props.elementGeometryBuilderParams;
```
wgoehrig, line 480: This class field is interesting... @GytisCepk were there any test or build failures that prompted you to add this?

I did some debugging and `elementGeometryBuilderParams` definitely is **not** set by the `Entity` constructor, and it also doesn't get serialized in [`GeometricElement.toJSON()`](https://github.com/iTwin/itwinjs-core/blob/f0164a3a92d23d77112a8213617804f28dfe67b6/core/backend/src/Element.ts#L490).  However, the corresponding [`GeometricElementProps.elementGeometryBuilderParams`](https://github.com/iTwin/itwinjs-core/blob/f0164a3a92d23d77112a8213617804f28dfe67b6/core/common/src/ElementProps.ts#L113) member is used in a [bunch of places](https://github.com/search?q=repo%3AiTwin%2Fitwinjs-core+elementGeometryBuilderParams+language%3ATypeScript&type=code&l=TypeScript)...

@pmconne @bbastings is it possible this class field is just broken and has never been used?  Should we remove it entirely?  If we want to fix it instead, we should probably also make sure it gets included in `toJSON()`...

core/backend/src/ViewDefinition.ts
GytisCepk: Two things:
- Code readability. In my opinion, this is much easier to understand
- `declare` just like `!` makes code more error prone

core/backend/src/Element.ts
pmconne: > is it possible this class field is just broken and has never been used? Should we remove it entirely? If we want to fix it instead, we should probably also make sure it gets included in toJSON()

The corresponding field in the interface was always a bit of a hack. My guess is nobody has ever tried to set the field on an instance of the class. Sam initially added it in #3017 for usage by local sync connectors. @bbastings uses it for some tools.

bbastings: elementGeometryBuilderParams is used and does work...

It is intended to be used by insert/update to provide the geometry stream in flatbuffer format to the backend as a napi object. It's supposed to be undefined when reading an element.

You can look at GeometricElement::_FromJson(BeJsConst props) in imodel-native to see how it's used.

pmconne: @bbastings the question is specifically about `GeometricElement.elementGeometryBuilderParams` (the class), not `GeometricElementProps.elementGeometryBuilderParams` (the JSON interface).

bbastings: @pmconne Ah, I don't see a need to set this on the class then.

GeometryPart/GeometryPartProps is similar...

pmconne: Thanks. It's `@beta` and never worked, so let's remove it.

GytisCepk: Removed `GeometricElement.elementGeometryBuilderParams` and `GeometryPart.elementGeometryBuilderParams` (identical situation to `GeometricElement`)

wgoehrig (approved):

aruniverse (approved):

christophermlawson (approved):

PR #7566, opened 2025-01-16
Add link to editing API after it is mentioned.

tcobbs-bentley:
The original documentation stated that ECSQL could only use SELECT, and that modification must be done through the API. This provides an example of one API that allows for data modification.


docs/learning/backend/ExecutingECSQL.md
```diff
@@ -10,6 +10,7 @@ Executing an ECSQL statement typically consists of these steps:
 1. Reset the statement and clear its parameter bindings, if the statement should be executed again.
 
 > For iModels only ECSQL SELECT statements can be executed. Data modification must be done through the API.
+> For example, see [Elements.updateElement]($backend).
```
ben-polinsky, line 13:
```suggestion
> For example, see `[IModelDb.Elements.updateElement]($backend)`.
```

aruniverse (approved):

PR #7570, opened 2025-01-16
Fix GraphicRepresentation types according to latest Mesh Export API docs

eringram:
Update `GraphicRepresentationFormat` type and `GraphicRepresentationStatus` enum to reflect the latest [Mesh Export Service docs](https://developer.bentley.com/apis/mesh-export/operations/get-export/#definitions-ref). These are breaking changes to beta APIs intended to go into iTwin.js 5.0.


common/api/frontend-tiles.api.md
```diff
@@ -88,18 +88,18 @@ export type GraphicRepresentation = {
 });
 
 // @beta
-export type GraphicRepresentationFormat = "IMDL" | "3DTILES" | string;
+export type GraphicRepresentationFormat = "IMODEL" | "3DTiles" | "CESIUM";
```
eringram, line 91: @danieliborra @arvindsvenkat To clarify, is "3DFT" as an export type still returned by the service? If so I'll add it here

markschlosseratbentley (commented):
Looks good to me.
I will defer to @danieliborra and @arvindsvenkat for final approval, though.
Do we need to notify anyone of this change?

aruniverse (commented):
will this be needed in 4.10 or 4.11?

eringram:
> @aruniverse will this be needed in 4.10 or 4.11?

No, we were planning for it to go in 5.0.

extensions/frontend-tiles/src/FrontendTiles.ts
```diff
@@ -92,7 +92,7 @@ export async function* queryMeshExports(args: QueryMeshExportsArgs): AsyncIterab
       changeId: args.changesetId,
       type: "IMODEL",
     },
-    format: "IMDL",
+    format: "IMODEL" as GraphicRepresentationFormat,
```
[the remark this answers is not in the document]
eringram: Without the cast the compiler was complaining about the type of `format` being incorrect, inferred to be string instead of GraphicRepresentationFormat. Explicitly stating the type is wordier but is that preferred? E.g.
```
const testArgs: {
  accessToken: string;
  sessionId: string;
  dataSource: {
    iTwinId: string;
    id: string;
    changeId?: string;
    type: string;
  };
  format: GraphicRepresentationFormat;
} = {
  accessToken: "this-is-a-fake-access-token",
  sessionId: "testSession",
  dataSource: {
    iTwinId: "iTwinId",
    id: "srcId",
    changeId: undefined,
    type: "srcType",
  },
  format: "IMODEL",
};
```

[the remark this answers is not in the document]
aruniverse: while wordier, i think its safer to extract the type out for an internal interface to use

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -10,19 +10,17 @@ import { IModelApp, ITWINJS_CORE_VERSION } from "@itwin/core-frontend";
...
-/* eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents */
-export type GraphicRepresentationFormat = "IMDL" | "3DTILES" | string;
+export type GraphicRepresentationFormat = "IMODEL" | "3DTiles" | "CESIUM";
 
 /** Graphic representations are generated from Data Sources.
  * The status of a Graphic Representation indicates the progress of that generation process.
  * @beta
  */
-// ###TODO this needs to be expanded to include more statuses, and/or "failed" needs to be replaced with "invalid".
 export enum GraphicRepresentationStatus {
```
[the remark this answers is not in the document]
eringram: I don't completely remember, but my guess is because GraphicRepresentationFormat used to have enum values that were strings different from the constant names, like this:
```
export enum GraphicRepresentationStatus {
  InProgress = "In progress",
  Complete = "Complete",
  NotStarted = "Not started",
  Failed = "Failed"
}
```
Why those different strings were necessary to begin with, I'm not sure and it could have just been a poor decision. Since all that is needed are the strings "InProgress", "Complete", "NotStarted", and "Invalid", it could become a type like GraphicRepresentationFormat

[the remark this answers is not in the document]
eringram: I realized one reason we did this is the enum is used in a conditional type definition of GraphicRepresentation [here](https://github.com/iTwin/itwinjs-core/pull/7570/files#diff-3cd0b96cd8e6e82bb18bda068b487efdf59dd65df6c0145634d66b04b7fc9621R73-R79):

```
export type GraphicRepresentation = {
  displayName: string;
  representationId: string;
  status: GraphicRepresentationStatus;
  format: GraphicRepresentationFormat;
  dataSource: DataSource;
  /** The url of the graphic representation
   * @note The url can only be guaranteed to be valid if the status is complete.
   * Therefore, the url is optional if the status is not complete, and required if the status is complete.
   */
} & ({
  status: Omit<GraphicRepresentationStatus, GraphicRepresentationStatus.Complete>;
  url?: string;
} | {
  status: GraphicRepresentationStatus.Complete;
  url: string;
});
```

Where the `url` prop is only guaranteed to exist if status = GraphicRepresentationStatus.Complete. This could also be done with a union type though with `Exclude<GraphicRepresentationStatus, "Complete">`, etc. instead of `Omit`.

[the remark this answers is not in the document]
eringram: @aruniverse do you have a preference between enum and string literal type here?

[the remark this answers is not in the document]
aruniverse: I personally would prefer the string literals, not a huge fan of enums. I think the Exclude utility would be perfect for the discriminated union type

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
```diff
@@ -184,21 +175,40 @@ export async function* queryGraphicRepresentations(args: QueryGraphicRepresentat
...
-        url: foundSource._links?.mesh.href,
-        dataSource: {
-          iTwinId: args.dataSource.iTwinId,
-          id: foundSource.request.iModelId,
-          changeId: foundSource.request.changesetId,
-          type: foundSource.request.exportType,
-        },
-      };
+      let graphicRepresentation: GraphicRepresentation;
+      if (foundSource.status === "Complete") {
```
eringram, line 181: Updated the `GraphicRepresentationFormat` type to be a union of strings, which unfortunately requires this long if statement to enforce that `url` has type `string` only if `status === "Complete"`, and type `string | undefined` otherwise. I tried to do this with a ternary operator but that doesn't work so I think this is the best solution to avoid more type assertions.

common/api/frontend-tiles.api.md
```diff
@@ -80,27 +80,18 @@ export type GraphicRepresentation = {
...
     url?: string;
 } | {
-    status: GraphicRepresentationStatus.Complete;
+    status: "Complete";
     url: string;
 });
 
 // @beta
-export type GraphicRepresentationFormat = "IMDL" | "3DTILES" | string;
+export type GraphicRepresentationFormat = "3DFT" | "3DTiles" | "CESIUM" | "IMODEL";
```
pmconne, line 91: Why did you remove `| string`? Are we never going to add support for a new format?

eringram: It seemed pointless to have a type that includes some string literals but also could be any string, so I thought it would be better to enforce that the string must be one of the values currently returned by the mesh export service.

But you're right that new formats could be added at any time, and waiting until a new major version to update this type could be a problem considering the MES is still in development/tech preview.

aruniverse: @eringram take a look at https://stackoverflow.com/questions/61047551/typescript-union-of-string-and-string-literals

Im kinda in favor of removing the `| string` because its a super type and pretty much invalidates any type safety we have

eringram: Maybe since `| string` is pointless for type safety but we want to keep this open to new formats, I can remove GraphicRepresentationFormat as a type completely and just have the `format` property be a string. We can just put the list of current formats like "IMODEL", "3DTiles", etc. in the `format` prop documentation to be clear what values it should have.

I don't think this is a problem because as far as I know GraphicRepresentationFormat is not used outside this file, and if it is this will be in 5.0 with the changelog so people can react to it. 

I feel like ideally the implementation would be similar to the MapLayerFormat class that is extensible, but so far there's no need for a whole class with properties and methods in this case.

pmconne: Right. The string union exists for documentation of currently supported formats, the "| string" is there for forward compatibility.

aruniverse: Well if youre adding support for a new format, wouldn't it be "widening" the type, therefore not breaking? I would think narrowing a scope (eg removing a format) is the only time it would be considered breaking. But now as I write this, I'm guessing when you mean forward compatibility that also includes removing "old" formats

pmconne:
```
function doSomethingCool(graphicFormat: GraphicRepresentationFormat): string {
  switch (graphicFormat) {
    case "3DFT":
    case "3DTiles":
    case "CESIUM":
    case "IMODEL":
      return "cool";
  }

  // unreachable code becomes reachable once GraphicRepresentationFormat is expanded,
  // suddenly my function implicitly returns `undefined` and I crash at run-time.
}

File: extensions/frontend-tiles/src/FrontendTiles.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (eringram) [reply extensions/frontend-tiles/src/FrontendTiles.ts]: Fixed

File: extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts (hunk 3 above)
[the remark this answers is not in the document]
Comment (eringram) [reply extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts]: Also fixed this

File: common/api/frontend-tiles.api.md (hunk 5 above)
Comment (eringram) [reply common/api/frontend-tiles.api.md]: Added `| string` back to the type so it can remain extendable but the existing formats are documented. 
Also I saw this is done in a couple other places ([Tileset3dSchema.ts](https://github.com/iTwin/itwinjs-core/blob/master/core/common/src/tile/Tileset3dSchema.ts#L50), [GltfSchema.ts](https://github.com/iTwin/itwinjs-core/blob/5fc3a18aae5752681a6eef86e1ecea33bdef6ce2/core/frontend/src/common/gltf/GltfSchema.ts#L583)) so it's not completely without precedent

File: common/changes/@itwin/frontend-tiles/eringram-graphicrepresentation-fix_2025-01-21-17-59.json [hunk 6]
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/frontend-tiles",
+      "comment": "Update GraphicRepresentationFormat type and GraphicRepresentationStatus enum to reflect the latest Mesh Export Service docs.",
```
aruniverse, line 5: should probably mention/update that GraphicRepresentationStatus is no longer an enum but a string literal type

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
aruniverse: this doesnt need to be a massive if statement, 
if the only difference is the url, you can create the object and in the inline definition of the url property use a ternary to set the value

common/changes/@itwin/frontend-tiles/eringram-graphicrepresentation-fix_2025-01-21-17-59.json
eringram: Good point, added

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
arvindsvenkat: I agree, would be easier without a huge if block

```
let graphicRepresentation: GraphicRepresentation = {
     displayName: foundSource.displayName,
     representationId: foundSource.id,
     status: foundSource.status,
     format: args.format,
     url: (foundSource.status === "Complete") ? foundSource._links!.mesh.href : foundSource._links?.mesh.href,
     dataSource: {
        iTwinId: args.dataSource.iTwinId,
        id: foundSource.request.iModelId,
        changeId: foundSource.request.changesetId,
        type: foundSource.request.exportType,
     },
};
yield graphicRepresentation;
```

extensions/frontend-tiles/src/test/GraphicsProvider/GraphicRepresentationProvider.test.ts
arvindsvenkat, line 1: A minor nitpick: Could we move "interface ServiceJsonResponse" and "interface ServiceJsonResponses" outside the function ?

common/api/frontend-tiles.api.md
```diff
@@ -80,27 +80,18 @@ export type GraphicRepresentation = {
...
     url?: string;
 } | {
-    status: GraphicRepresentationStatus.Complete;
+    status: "Complete";
     url: string;
 });
 
 // @beta
-export type GraphicRepresentationFormat = "IMDL" | "3DTILES" | string;
+export type GraphicRepresentationFormat = "3DFT" | "3DTiles" | "CESIUM" | "IMODEL" | string;
```
arvindsvenkat, line 91: Do we need type "string" here?

pmconne: Please refer to [previous comments](https://github.com/iTwin/itwinjs-core/pull/7570#discussion_r1927383203).

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
eringram: That was my initial thought too, but if I set url to that ternary the compiler doesn't like it but I don't fully understand the error message. My guess is because url is ultimately still inferred to be type `string | undefined`,  which is incompatible with status "Complete" according to the GraphicRepresentation definition. 

This is the full error message:
```
Type '{ displayName: string; representationId: string; status: GraphicRepresentationStatus; format: string; url: string | undefined; dataSource: { ... }' is not assignable to type 'GraphicRepresentation'.
  Type '{ displayName: string; representationId: string; status: GraphicRepresentationStatus; format: string; url: string | undefined; dataSource: { ... }' is not assignable to type '{ displayName: string; representationId: string; status: GraphicRepresentationStatus; format: string; dataSource: DataSource; } & { status: "Complete"; url: string; }'.
    Type '{ displayName: string; representationId: string; status: GraphicRepresentationStatus; format: string; url: string | undefined; dataSource: { ... }' is not assignable to type '{ status: "Complete"; url: string; }'.
      Types of property 'status' are incompatible.
        Type 'GraphicRepresentationStatus' is not assignable to type '"Complete"'.
          Type '"InProgress"' is not assignable to type '"Complete"'.
```

I could also force it by not assigning the `graphicRepresentation` variable a type, but then doing `yield graphicRepresentation as GraphicRepresentation` at the end (the tests pass this way but `as` seems unsafe).

extensions/frontend-tiles/src/test/GraphicsProvider/GraphicRepresentationProvider.test.ts
eringram: Added

common/api/frontend-tiles.api.md
arvindsvenkat: Makes sense.

extensions/frontend-tiles/src/GraphicsProvider/GraphicRepresentationProvider.ts
eringram: @aruniverse Any suggestions here?

aruniverse: please take a look at the latest commit, which should hopefully clean the api up

eringram: Thanks. I think simplifying the API is the right way to go, it was the other option we considered when first creating GraphicRepresentation. As Paul said in [this comment](https://github.com/iTwin/itwinjs-core/pull/6818#discussion_r1638340975) it just means the caller has to enforce that url is defined

aruniverse (approved):

danieliborra (approved):

markschlosseratbentley (approved):

PR #7572, opened 2025-01-17
Remove deprecated 3.x APIs in core-common

MichaelSwigerAtBentley:
Removes deprecated APIs from core-common package

Note: This PR does not remove RPC methods deprecated in 3.6 or update the RPC interface. Deferring those changes for a larger scale RPC PR. 

Files not edited in favor of a later RPC specific PR:
BackendTypes.ts
RpcConstants.ts
OpenAPI.ts
WebAppRpcProtocol.ts


core/common/src/BackgroundMapSettings.ts
```diff
@@ -58,32 +58,6 @@ export interface BackgroundMapProps {
   providerData?: never;
 }
 
-/** Properties of [[PersistentBackgroundMapProps]] that have been deprecated, but are retained for backwards compatibility.
- * These properties are omitted from [[BackgroundMapProps]] as they are no longer part of the API, but are included in
- * [[PersistentBackgroundMapProps]] because they remain part of the persistence format.
- * @public
- * @extensions
- */
-export interface DeprecatedBackgroundMapProps {
```
MichaelSwigerAtBentley, line 67: @pmconne Interested in your opinion here: How should removing these Props look (if we want to remove them along with the API functions)? Since both Props are deprecated, the interface would be unnecessary. However, that would make PersistentBackgroundMapProps identical to BackgroundMapProps, correct? and make all of the persistentJSON() functions redundant. Those functions were not  flagged as deprecated, therefore we shouldn't remove them without warning first? Or do my initial changes here look correct?

core/common/src/rpc/IModelReadRpcInterface.ts
```diff
@@ -115,8 +115,6 @@ export abstract class IModelReadRpcInterface extends RpcInterface {
   public async getViewStateData(_iModelToken: IModelRpcProps, _viewDefinitionId: string, _options?: ViewStateLoadProps): Promise<ViewStateProps> { return this.forward(arguments); }
   public async readFontJson(_iModelToken: IModelRpcProps): Promise<FontMapProps> { return this.forward(arguments); } // eslint-disable-line @typescript-eslint/no-deprecated
   public async getToolTipMessage(_iModelToken: IModelRpcProps, _elementId: string): Promise<string[]> { return this.forward(arguments); }
-  /** @deprecated in 3.x use ViewStore apis. */
-  public async getViewThumbnail(_iModelToken: IModelRpcProps, _viewId: string): Promise<Uint8Array> { return this.forward(arguments); }
```
MichaelSwigerAtBentley, line 119: Is this something we want to remove as well? I will need to bump the interface version from 3.7.0 to 4.0.0 as this is breaking, correct?

pmconne: It tries to read a thumbnail out of the db, it's almost never going to find one except in very ancient iModels.
However do note that to my knowledge nobody's using ViewStore currently so you won't have any better luck finding thumbnails there either.
The saved views service does produce thumbnails.
BTW [the implementation](https://github.com/iTwin/itwinjs-core/blob/451a2ffad93a08de0b470dd81648075b262377a7/core/backend/src/rpc-impl/IModelReadRpcImpl.ts#L329) advises you to use a non-existent `queryViewThumbnail` method instead.

core/common/src/BackgroundMapSettings.ts
pmconne: PersistentBackgroundMapProps describes the shape of the data as stored in iModels, including the "deprecated" providerName and providerData properties. We have to support those persistent properties indefinitely. The intent was to deprecate and ultimately remove their usage in public APIs.

Persistent formats should always be accurately represented in the public API, so we can't remove those properties. But I don't know another way of conveying "you shouldn't use this property, it's just a wart on the persistent representation" in a way that will warn people automatically (linter) other than marking it deprecated.

Simplest solution is probably just to leave those two properties deprecated and don't remove them.

MichaelSwigerAtBentley: Makes sense, Ill revert this.

core/common/src/rpc/IModelReadRpcInterface.ts
MichaelSwigerAtBentley: @hl662 FYI while looking at core-backend deprecated methods

MichaelSwigerAtBentley:
There are also deprecated objects in rpc, such as HttpServerRequest and HttpServerResponse with the note, "The RPC system will be significantly refactored (or replaced) in the future.". These objects are used throughout the rest of the files. Are these objects we wish to no longer export, or make private?

core/common/src/BackendTypes.ts
```diff
@@ -1,26 +0,0 @@
...
-
-/*
-IMPORTANT:
-This is a temporary source file for backwards compatibility only.
-Do not add any new types to this file.
-All types here will be removed in 4.0
-*/
-
-/** @public @deprecated in 3.x. This type was mistakenly made public in the common scope. */
-export type BackendReadable = Readable;
```
MichaelSwigerAtBentley, line 20: Also, for these deprecated types, in files like RpcMarchaling.ts, should I replace them with their node:stream counterpart?

pmconne:
> deprecated objects in rpc

My strong opinion is to leave RPC to a separate PR.

MichaelSwigerAtBentley:
> > deprecated objects in rpc
> 
> My strong opinion is to leave RPC to a separate PR. It seems unlikely that most of the deprecated stuff will actually be removable in the 5.0 timeline.

Alright, I'll leave it out for now and create a new issue to track it separately.

core/common/src/RenderMaterial.ts
```diff
@@ -36,71 +35,3 @@ export abstract class RenderMaterial {
...
 }
-
-/** @public */
-export namespace RenderMaterial {
-  function clampToNormalizedRange(value: number): number {
-    return Math.max(0.0, Math.min(1.0, value));
-  }
-
-  /** @deprecated in 3.x. Use [CreateRenderMaterialArgs]($frontend). */
-  export class Params {
```
MichaelSwigerAtBentley, line 47: These params are also still in use, I'm going to try swapping them out for CreateRenderMaterialArgs

core/common/src/BackendTypes.ts
aruniverse: I think you should revert these changes and keep this coupled with the rpc changes in a separate pr that @GytisCepk and @paulius-valiunas will lead efforts on

MichaelSwigerAtBentley: Sounds good. Ill document what hasn't been removed yet in the PR description

core/common/src/RenderMaterial.ts
MichaelSwigerAtBentley: RenderTexture.Params is being using in multiple places in core\frontend

I was able to change one of the files, ParseImdlDoucment.ts, but I'm not sure if its implemented correctly as there is some logic around parsing colors. 

@aruniverse @ben-polinsky Who would be best to talk to for more info on how to proceed.

aruniverse: @pmconne will be your best starting point

pmconne: Looks like some of the core-frontend usage is being cleaned up in #7591.

Bear in mind that removing deprecated APIs just means making them unavailable to consumers of the public API. While generally we'll want to delete the APIs entirely, if it's too complicated to clean up our internal usage of a particular one, we can instead move it to /src/internal, and export it from cross-package.ts if it's used in other packages within itwinjs-core.

MichaelSwigerAtBentley:
Separate Issue for RPC: https://github.com/iTwin/itwinjs-backlog/issues/1374

core/common/src/RenderMaterial.ts
```diff
@@ -43,7 +43,9 @@ export namespace RenderMaterial {
     return Math.max(0.0, Math.min(1.0, value));
   }
 
-  /** @deprecated in 3.x. Use [CreateRenderMaterialArgs]($frontend). */
+  /** Params for use in old CreateMaterial functions. Use [CreateRenderMaterialArgs]($frontend) instead.
+   * @internal
+  */
   export class Params {
```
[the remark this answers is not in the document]
MichaelSwigerAtBentley: I have gone back and forth on replacing the usages of RenderMaterial.Params. I have had success with replacing one or two, and realized that most are not removable. I also thought about moving the RenderMaterial.Params into internal, but I wouldn't be able to extend the RenderMaterial Namespace from the internal folder right? If there is a way to keep Params inside the RenderMaterial namespace and have just params live in Internal, I think that is what we should do.

core/common/src/RenderTexture.ts
```diff
@@ -75,8 +75,8 @@ export namespace RenderTexture {
   }
 
   /** Parameters used to construct a [[RenderTexture]].
-   * @deprecated in 3.x. use RenderSystem.createTexture and CreateTextureArgs.
-   * @public
+   *  Use RenderSystem.createTexture and CreateTextureArgs.
+   *  @internal
```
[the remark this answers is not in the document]
MichaelSwigerAtBentley: For RenderTexture.Params, there is an addition enum 'Type' that is being exported as public in the namespace, so we wouldn't be able to set this namespace as internal like with RenderMaterial.Params

core/common/src/RenderMaterial.ts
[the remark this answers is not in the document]
pmconne: > most are not removable

Can you be more specific about what makes them not removable?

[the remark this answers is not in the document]
MichaelSwigerAtBentley: For RenderMaterial.Params:

[Material.ts](https://github.com/iTwin/itwinjs-core/blob/08d7378fc5cef73623f64a51148b9181b51bff51/core/frontend/src/render/webgl/Material.ts#L61) requires RenderMaterial.Params in its constructor. The replacement method [CreateRenderMaterial](https://github.com/iTwin/itwinjs-core/blob/08d7378fc5cef73623f64a51148b9181b51bff51/core/frontend/src/render/webgl/System.ts#L664) uses the createRenderMaterialArgs to create a Rendermaterial.Params to call Material constructor. This Material constructor would need to be updated to use createRenderMaterialArgs directly, then all places using Material would need an update. Is this something we want to do? If so I can take a stab at it.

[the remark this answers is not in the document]
MichaelSwigerAtBentley: For RenderTexture.Params: 

There are internal functions in use that don't seem to have alternatives, such as [createTextureFromElement](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/render/RenderSystem.ts#L755). Which is why I opted to update RenderTexture.Params to internal instead of removing it.

core/common/src/RenderTexture.ts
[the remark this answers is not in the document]
pmconne: > [Material.ts](https://github.com/iTwin/itwinjs-core/blob/08d7378fc5cef73623f64a51148b9181b51bff51/core/frontend/src/render/webgl/Material.ts#L61) requires RenderMaterial.Params in its constructor. The replacement method [createRenderMaterial](https://github.com/iTwin/itwinjs-core/blob/08d7378fc5cef73623f64a51148b9181b51bff51/core/frontend/src/render/webgl/System.ts#L664) uses the createRenderMaterialArgs to create a RenderMaterial.Params and just calls the Material constructor. This Material constructor would need to be updated to use createRenderMaterialArgs directly, then all places using Material would need an update. Is this something we want to do? If so I can take a stab at it.

That Material class is entirely internal. The `System.createRenderMaterial` implementation already receives a `CreateRenderMaterialArgs` that it must pointlessly convert to a `RenderMaterial.Params`, just so that the Material constructor can pull back out the information originally provided by the args.

core/common/src/RenderMaterial.ts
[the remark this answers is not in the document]
pmconne: > There are internal functions in use that don't seem to have alternatives, such as [createTextureFromElement](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/render/RenderSystem.ts#L755).

Look at `getTextureFromElement` to which the implementation of `System.createTextureFromElement` forwards. The only thing it uses from the params is the `type` field. It dumbly both accepts a `key` and also `params.key`. It should accept only key, iModel, type, and format, not `Params`.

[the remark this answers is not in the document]
pmconne: Argh, `RenderSystem.createTextureFromElement` is `@public` and not deprecated. Still, `RenderTexture.Params` is deprecated and we should remove it.

core/frontend/src/render/webgl/Material.ts
```diff
@@ -52,41 +51,62 @@ export class Material extends RenderMaterial {
...
 
     const scale = (value: number) => Math.floor(value * 255 + 0.5);
-    this.setInteger(scale(params.diffuse), scale(params.specular), 0);
 
-    const textureWeight = undefined !== this.textureMapping ? this.textureMapping.params.weight : 1.0;
-    const specularRgb = undefined !== params.specularColor ? params.specularColor : ColorDef.white;
-    const specularColors = specularRgb.colors;
-    this.setInteger(scale(textureWeight), specularColors.r, 1);
-    this.setInteger(specularColors.g, specularColors.b, 2);
+    const diffuseWeight = args.diffuse?.weight ?? 0.6;
```
MichaelSwigerAtBentley, line 95: As mentioned above, diffuse.weight can be undefined now, so I added the old default value from RenderMaterial.Params as a magic number here. Should we look to add the defaults somewhere else? If not, then why is 0.6 the defualt value, I would like to add a comment making sure its explained.

MichaelSwigerAtBentley:
@pmconne I refactored Material to use CreateRenderMaterialArgs directly, take a look to make sure I interpreted the CreateRenderMaterialArgs correctly. Is this what you had in mind?

core/frontend/src/render/webgl/Material.ts
```diff
@@ -39,8 +40,6 @@ import { SurfaceMaterial, SurfaceMaterialAtlas } from "../../common/internal/ren
  * @internal
  */
 export class Material extends RenderMaterial {
-  // eslint-disable-next-line @typescript-eslint/no-deprecated
-  public static readonly default: Material = new Material(RenderMaterial.Params.defaults);
```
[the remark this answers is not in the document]
pmconne: Not unless there's some compelling need for them.

core/frontend/src/render/webgl/Material.ts
pmconne: [By default] diffuse color contributes 60% to the final color and the other 40% comes from specular color.

core/common/src/internal/RenderMaterialParams.ts
MichaelSwigerAtBentley, line 1: So we want to keep this as internal instead of removing? I can revert my changes that remove it from Material, GraphicsDescription.test, and imdl files

MichaelSwigerAtBentley: I have 3 commits removing it from Imdl and tests that I have not pushed, but will now conflict

pmconne: I thought you were keeping it. I may have gotten this PR confused with the other one that was doing stuff with materials.

pmconne: Just keep it internal for now, the immediate priority is cleaning up the public APIs for 5.0. Internal cleanup can happen at our leisure.

MichaelSwigerAtBentley:
There seems to be an issue with Vite test setup detecting the exported internal classes. Talking with @hl662 about what could be happening.

pmconne (approved):

pmconne:
> There seems to be an issue with Vite test setup detecting the exported internal classes

What specific issue? The issue I have seen is when I am doing `import { Foo }` but I only use it as a type; switching to `import type { Foo }` fixes it.

core/common/src/RenderMaterial.ts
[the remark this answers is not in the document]
MichaelSwigerAtBentley: @aruniverse Both RenderMaterial.Params and RenderTexture.Params have now been moved to internal and exported through the cross-package

MichaelSwigerAtBentley:
> > There seems to be an issue with Vite test setup detecting the exported internal classes
> 
> What specific issue? The issue I have seen is when I am doing `import { Foo }` but I only use it as a type; switching to `import type { Foo }` fixes it.

It was an issue with vita not detecting RenderMaterialParams. 

```SyntaxError: The requested module '<core-common>' does not provide an export named 'RenderTextureParams'```

The vite's cache was not getting cleared, manually nuking node_modules and reinstalling and rebuilding did it. Rush clean/rebuild did not. 

@hl662 has created an issue to track this: https://github.com/iTwin/itwinjs-core/issues/7662

**Comment by MichaelSwigerAtBentley** [comment]
> i think it gets a little weird when we say an api is both intenral and deprecated, just things to clean up. but luckily most of that is surrounding rpc code which can get cleaned up later.
> 
> approving for now

Yes I agree, I left the deprecated tag on those functions as a indicator that they had not been addressed yet, and should be removed as a part of the RPC stuff.

**Comment by grigasp** [comment]
@MichaelSwigerAtBentley @aruniverse, `extract-api` on master fails for `core-common` because of changes in this PR:

```
Error: src/rpc/web/WebAppRpcProtocol.ts:26:1 - (ae-incompatible-release-tags) The symbol "HttpServerRequest" is marked as @public, but its signature references "BackendReadable" which is marked as @internal
Error: src/rpc/web/WebAppRpcProtocol.ts:57:1 - (ae-incompatible-release-tags) The symbol "HttpServerResponse" is marked as @public, but its signature references "BackendWritable" which is marked as @internal
```

MichaelSwigerAtBentley:
PR to resolve this: https://github.com/iTwin/itwinjs-core/pull/7670

PR #7573, opened 2025-01-19
Accudraw viewport UI

bbastings:
New platform AccuDraw UI for consideration to replace FrameworkAccuDraw from the appui package.
This PR also includes fixes that would be necessary to make FrameworkAccuDraw work properly.


bbastings:

core/frontend/src/tools/AccuDrawViewportUI.ts
```diff
@@ -0,0 +1,733 @@
...
+    if (!isDown)
+      return false;
+
+    itemField.value = currentValue.substring(0, selectionPrevious) + replacement + currentValue.substring(selectionStart);
+    itemField.selectionStart = itemField.selectionEnd = selectionPrevious + 1;
+
+    return false;
+  }
+
+  /** @internal */
```
pmconne, line 296: `doProcessOverrideKey` and `doProcessKey` are `@internal` but also `protected`, suggesting subclasses can be expected to override them. If they shouldn't be overridden, make them `private`. If they can be overridden by subclasses outside of itwinjs-core, make them `@public`. If they should only be overridden by subclasses inside itwinjs-core, rename them to use a `Symbol`.

core/frontend/src/AccuDraw.ts
```diff
@@ -1272,18 +1274,54 @@ export class AccuDraw {
...
+    switch (key) {
+      case ".":
+      case ",":
+        return true;
+
+      case "'": // feet/minutes...
+      case "\"": // inches/seconds...
+        return true;
+
+      case "°": // degrees...
```
pmconne, line 1291: Should this switch statement have a case for "^" for degrees? I can't tell.

core/frontend/src/tools/AccuDrawViewportUI.ts
pmconne: Same goes for all the other `@internal` `protected` members like `focusItem`, `doNavigate`, etc.

pmconne (approved):

core/frontend/src/tools/AccuDrawViewportUI.ts
```diff
@@ -0,0 +1,733 @@
...
+  protected doProcessOverrideKey(ev: KeyboardEvent, isDown: boolean, item: ItemField): boolean {
```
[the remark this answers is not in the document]
bbastings: They are all internal except doProcessUnhandledKey, which is documented. I see no reason to document the others?

core/frontend/src/AccuDraw.ts
bbastings: Not necessary, this method is mainly to support NSEW for bearing directions. I don't have an exhaustive list of what every parser might accept and just wanted to call out a few obvious things to early return before the number and letter checks. The "^" -> "°" replacement is only supported in the UI subclass as well...

core/frontend/src/tools/AccuDrawViewportUI.ts
bbastings: Done.

pmconne (approved):

PR #7577, opened 2025-01-21
ecschema-metadata integration in backend

rschili:
First step of cleaning up ClassRegistry, removing redundancy and make ecschema-metadata available in backend.

Story: https://github.com/iTwin/itwinjs-backlog/issues/1156

This is the first of several PRs, it only adds the dependency and adds some ways to access the data, it does not replace any existing logic yet.


core/ecschema-metadata/src/Metadata/Schema.ts
```diff
@@ -443,22 +443,225 @@ export class Schema implements CustomAttributeContainerProps {
...
+   * @param itemConstructor optional, the constructor of the item to return. If missing, the returned value cannot be type checked.
    */
-  public async getItem<T extends SchemaItem>(name: string): Promise<T | undefined> {
+  public async getItem<T extends SchemaItem>(name: string, itemConstructor?: SchemaItemConstructor<T>): Promise<T | undefined> {
     // this method exists so we can rewire it later when we load partial schemas, for now it is identical to the sync version
-    return this.getItemSync<T>(name);
+    return this.getItemSync<T>(name, itemConstructor);
   }
 
+  /**
```
ColinKerr, line 553: can we avoid adding these sync methods?  If we want to prioritize the async methods then the getItemSync method should be enough for callers

rschili: Yeah makes sense. Just seemed a bit strange since getting items within a schema is always sync at the moment and I'm not sure how likely we are to implement partially loaded schemas with the current class design.... So we could also do away with the async ones and leave that to ISchemaItemLocater. What do you think?

core/ecschema-metadata/src/Metadata/Schema.ts
```diff
@@ -443,22 +443,225 @@ export class Schema implements CustomAttributeContainerProps {
...
+
   /**
    * Gets an item from within this schema. To get by full name use lookupItem instead.
    * @param key the local (unqualified) name, lookup is case-insensitive
+   * @param itemConstructor optional, the constructor of the item to return. If missing, the returned value cannot be type checked.
    */
-  public getItemSync<T extends SchemaItem>(name: string): T | undefined {
-    // Case-insensitive search
-    return this._items.get(name.toUpperCase()) as T;
+  public getItemSync<T extends SchemaItem>(name: string, itemConstructor?: SchemaItemConstructor<T>): T | undefined {
```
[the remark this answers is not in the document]
rschili: Passing the constructor here is way I see mentioned all over the web. typeof in typescript on a class returns the type of the constructor of the class. That's not useful here, it's metainformation about the blueprint of the information that we need, not the actual information.

So it doesn't work like it would in C#.

Example:
```ts
// Use `typeof` to get the constructor type of EntityClass
type EntityClassConstructor = typeof EntityClass;

// Function that takes the constructor of EntityClass as a parameter
function createEntity(entityClass: EntityClassConstructor, name: string): EntityClass {
  return new entityClass(name);
}

// Usage
const entity = createEntity(EntityClass, "TestEntity");
```

core/ecschema-metadata/src/Metadata/Schema.ts
rschili: Hm, could we also consider the opposite? The items returned by these methods are all from within this schema, and the schema is always fully loaded, so accessing its items will always be sync, no?
If at some point we implement a partially loaded schema, we will probably do that in a separate class, not reusing the existing Schema class.
We do have ISchemaItemProvider which could serve that purpose.

core/ecschema-metadata/src/Metadata/Schema.ts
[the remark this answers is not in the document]
rschili: Mmh it doesn't quite work like that.  I tried so many variations now....

```ts
public getItemSync<T extends SchemaItem>(name: string, itemConstructor?: typeof SchemaItem): T | undefined {
```

I think this is what you had in mind. The problem with that is, that itemConstructor is not connected to T in this case. So you could call `getItemSync<EntityClass>("name", Unit);` which does not make any sense.

What DOES work is this:

```ts
  public getItemSync<T extends typeof SchemaItem>(name: string, itemConstructor?: T): InstanceType<T> | undefined {
    const value = this._items.get(name.toUpperCase());
    if (value === undefined)
      return value;

    if (itemConstructor !== undefined && (value.schemaItemType !== itemConstructor.schemaItemType))
      throw new ECObjectsError(ECObjectsStatus.InvalidSchemaItemType, `The item ${name} is not of the expected type.`);

    return value as InstanceType<T>;
  }
```

But that is a break, because now the generic parameter has to be `typeof EntityClass` so everybody calling the method with `getItemSync<EntityClass>` has to adjust it to `getItemSync<typeof EntityClass>`. I guess we could do that, but I'm hesitant...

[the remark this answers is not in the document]
rschili: I reworked it a bit. Strangely, when I put SchemaItemConstructor<T> into SchemaItem.ts and import it, the compiler would complain it cannot find the schemaItemType static property on the constructor. But IntelliSense didn't complain.
When I moved the type into the Schema.ts file, everything suddenly worked fine...

core/backend/package.json
```diff
@@ -101,6 +101,8 @@
   "dependencies": {
     "@bentley/imodeljs-native": "5.0.41",
     "@itwin/cloud-agnostic-core": "^2.2.4",
+    "@itwin/ecschema-locaters": "workspace:*",
```
aruniverse, line 104: doesnt seem to be used

core/backend/package.json
```diff
@@ -101,6 +101,8 @@
...
+    "@itwin/ecschema-metadata": "workspace:*",
```
aruniverse, line 105: should this be a peer instead of direct dep?

core/backend/package.json
rschili: Fair point, I just added this because I know we will need it eventually. This PR is supposed to be a minimal "get the foot in the door" one, and we will iterate on it, where we will make use of schema locaters.
Schema Registry for example uses global paths to schemas on the file system, which I have not rewired to use schema locaters yet, but I will eventually.
Can we still add it in this PR already? I feel those two packages sort of belong together.

core/backend/package.json
rschili: True, peer dependency seems like the better choice. I'll change it

core/ecschema-metadata/src/Metadata/Schema.ts
rschili: done

core/backend/package.json
rschili: done

core/backend/package.json
aruniverse: I would prefer to remove just in case we don't end up using it and accidentally release with it. But as long as we are 100% confident we will be using it, its fine.

rschili: Yeah sounds reasonable. I am not 100% confident we're going to use it, so I removed the dependency for now

core/backend/src/Entity.ts
```diff
@@ -92,6 +108,25 @@ export class Entity {
   /** Get the full BIS class name of this Entity in the form "schema:class". */
   public get classFullName(): string { return this._ctor.classFullName; }
 
+  public get schemaItemKey(): SchemaItemKey { return this._ctor.schemaItemKey; }
```
ColinKerr, line 111: lets tag everything we're adding as @beta during development so we can go through and review everything one last time before we make 5.0 final

core/backend/src/IModelDb.ts
```diff
@@ -1076,6 +1079,19 @@ export abstract class IModelDb extends IModel {
     return this._classMetaDataRegistry;
   }
 
+  public get schemaContext(): SchemaContext {
```
ColinKerr, line 1082: Add documentation now so we don't have a debt of docs to fill out at the end.  It will also help other people not familiar with ecschema-metadata understand the changes

core/backend/src/Schema.ts
```diff
@@ -21,6 +22,18 @@ export class Schema {
    */
   public static get schemaName(): string { throw new Error(`you must override static schemaName in ${this.name}`); }
 
+  private static _schemaKey?: SchemaKey;
+
+  /** Unique identifier for this schema, typed variant of [[schemaName]].
+   * @internal
+   */
+  public static get schemaKey(): SchemaKey {
```
ColinKerr, line 30: If this Schema object represents the version of the schema the domain handlers are written against you could make an argument that the version does matter.

I think what you have done is fine for now but our thinking here should be critical of the purpose of the Schema objects.  That may guide us to change this later.

core/ecschema-metadata/src/Metadata/Schema.ts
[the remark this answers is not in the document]
ColinKerr: This is cleaner but I'm concerned that this method could be suprising to callers.  If called without the constructor it will return undefined if no item is found, the item found cast to the template type regardless of the actual type.  If called with the constructor it will return undefined if no item is found, the item if found and of the right type or throw if an item of the wrong type is found.

I propose that we provide two options:
1. Explicit typesafe options that return the typed item if found and of the right type or undefined if not found or not of the right type
2. Explicitly generic option that returns a SchemaItem if found and leaves it up to the user to do the type checking

I suggest this because those two methods seem to cleanly fit the usecases we've seen over the years.  e.g. If I need to use a unit called smoot any other item called smoot won't do.  If I want to insert a unit called smoot I want to know if any other item exists regardless of type.

What do you think?

core/ecschema-metadata/src/SchemaKey.ts
```diff
@@ -19,6 +19,9 @@ export class ECVersion {
   private _write: number = 0;
   private _minor: number = 0;
 
+  // Define a static global instance representing zero
+  public static readonly ZERO = new ECVersion(0, 0, 0);
```
ColinKerr, line 23: Maybe it would be better to call this 'NoVersion' and document that a version with all zeros is an invalid version for a schema but using it in a key means no version has been specified.  As it stands this could mean anything and the docs don't help you figure what it should mean

ColinKerr (commented):
We need to keep track of the breaking changes we are making.  For example removing generics from ISchemaLocater.

This list should exist outside the scope of this PR so we can write up a nice doc for the 5 release.

core/backend/src/IModelDb.ts
rschili: done

core/backend/src/Entity.ts
rschili: Done

core/ecschema-metadata/src/SchemaKey.ts
rschili: done. Do we want to prevent schemas to be constructed with a NO_VERSION version? I don't think we have validation on that end now...
I'll keep the comment open to remind me to add something to the docs

core/backend/src/Schema.ts
ColinKerr: I'm ok resolving this comment, but we need some place to track things we want to ask consumers so we do not loose track of what we have to ask

ColinKerr (approved):

core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -196,8 +196,38 @@ export class SchemaContextEditor {
...
+    const schemaItem = await this.schemaContext.getSchemaItem(schemaItemKey);
+    if (!schemaItem) {
+      throw new SchemaEditingError(ECEditingStatus.SchemaItemNotFoundInContext, new SchemaItemId(schemaItemType, schemaItemKey));
+    }
+
+    if (!isItemType(schemaItem, itemConstructor))
+      throw new SchemaEditingError(ECEditingStatus.InvalidSchemaItemType, new SchemaItemId(schemaItemType, schemaItemKey));
+
+    return schemaItem;
+  }
```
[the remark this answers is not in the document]
rschili: It does seem quite redundant. I didn't want to make more changes than necessary to editing. I did talk with Stefan about the SchemaItemType argument, and he said they need it to differentiate between non-existing and mismatched items. Since Schema returns undefined in both cases, they need to check themselves.

As for schemaOrKey, that's true. Is it possible these are used to get a corresponding schema from a different context? @StefanApfel-Bentley  can you elaborate?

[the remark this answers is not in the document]
StefanApfel-Bentley: In the implementation before and after your changes, it does not make sense to have the two methods of lookupSchemaItem and getTypedSchemaItem. The only difference is that in lookupSchemaItem, the schemaOrKey variable is set to a key when a schema is passed. And then the variable is not used at all 🤨

I remember I had a talk with Chris when he introduced these two methods and one case did lookup the item on the schema directly instead of going through the locater. Unluckily I forgot the details why this has been done, but with the current implementations I don't see a need to have both methods.

core/ecschema-metadata/src/Metadata/Schema.ts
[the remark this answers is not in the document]
rschili: Done adjusting all spots

rschili:
> That comment is a little late, but after I've been using the api now a bit, I think it would be better to only have one method getItem which takes the key and optionally the type (aka the item class). That keeps the api a bit more compact.

I tried that, but the way typescript works makes this very tedious. It would look something like that:

```ts
  public async getItem(name: string): Promise<SchemaItem | undefined>
  public async getItem<T extends typeof SchemaItem>(name: string, itemConstructor?: T): Promise<InstanceType<T> | undefined> {
```

But then you could call it like getItem<typeof EntityClass>("name") and it would return an EntityClass without validation, because itemConstructor is optional and even though you provided the generic type parameter, the method cannot validate. Still typescript would think this type has been properly returned.

I wanted to make the itemConstructor parameter not optional so people who request a typed result are forced to do it correctly.

core/ecschema-metadata/src/Metadata/Schema.ts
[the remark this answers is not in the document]
StefanApfel-Bentley: I don't mind the idea passing the class as parameter, but since we only check for the schemaItemType, wouldn't it be easier, to pass the schemaItemType as 2nd argument. The comparison is the same (lookedup type vs expected type). We just need a way to return the actual item type based on the given schema type. But that could be done through a typescript map.

StefanApfel-Bentley:
> > That comment is a little late, but after I've been using the api now a bit, I think it would be better to only have one method getItem which takes the key and optionally the type (aka the item class). That keeps the api a bit more compact.
> 
> I tried that, but the way typescript works makes this very tedious. It would look something like that:
> 
> ```ts
>   public async getItem(name: string): Promise<SchemaItem | undefined>
>   public async getItem<T extends typeof SchemaItem>(name: string, itemConstructor?: T): Promise<InstanceType<T> | undefined> {
> ```
> 
> But then you could call it like getItem("name") and it would return an EntityClass without validation, because itemConstructor is optional and even though you provided the generic type parameter, the method cannot validate. Still typescript would think this type has been properly returned.
> 
> I wanted to make the itemConstructor parameter not optional so people who request a typed result are forced to do it correctly.

Yeah, you could add an overload with the parameter mandatory and the actual implementation would have it optional
image

Does not look great, but works. In Intellisense the third overload is hidden and you either look up by name and get a SchemaItem or you use the concrete constructor parameter overload and get that kind of item
image

rschili:
> Yeah, you could add an overload with the parameter mandatory and the actual implementation would have it optional image

I went ahead and implemented it like this, I like it. Thanks!

core/ecschema-editing/src/Editing/Editor.ts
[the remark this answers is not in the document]
rschili: The lookup methods and the item Type parameters have been removed

core/ecschema-metadata/src/Metadata/Schema.ts
```diff
@@ -443,29 +443,182 @@ export class Schema implements CustomAttributeContainerProps {
...
+    return refSchema.getItem(itemName);
   }
 
   /**
    * Attempts to find a schema item within this schema or a (directly) referenced schema
    * @param key The full name or a SchemaItemKey identifying the desired item.
+   * @param itemConstructor The constructor of the item to return.
    */
-  public async lookupItem<T extends SchemaItem>(key: Readonly<SchemaItemKey> | string): Promise<T | undefined> {
+  public async lookupTypedItem<T extends typeof SchemaItem>(key: Readonly<SchemaItemKey> | string, itemConstructor: T): Promise<InstanceType<T> | undefined> {
```
StefanApfel-Bentley, line 621: After we unified _getItem_ in one method with multiple overloads, we should probably do the same with lookupType either.

core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -286,3 +291,12 @@ export class SchemaContextEditor {
   }
 }
 
+function isItemType<T extends typeof SchemaItem>(schemaItem: SchemaItem, itemConstructor: T): schemaItem is InstanceType<T> {
```
rschili, line 294: @StefanApfel-Bentley This is sort-of redundant to logic in Schema.ts. I did not think we'd need this outside of there. Maybe we should expose this logic from the metadata package and use it here. Otherwise this will be pretty hard to spot if sometime in the future we expand our SchemaItem types.

core/ecschema-editing/src/Editing/Editor.ts
```diff
@@ -196,14 +198,17 @@ export class SchemaContextEditor {
...
-  public async getSchemaItem<T extends SchemaItem>(schemaItemKey: SchemaItemKey, schemaItemType: SchemaItemType): Promise<T> {
-    const schemaItem = await this.schemaContext.getSchemaItem<T>(schemaItemKey);
+  public async getSchemaItem(schemaItemKey: SchemaItemKey): Promise<SchemaItem>
+  public async getSchemaItem<T extends typeof SchemaItem>(schemaItemKey: SchemaItemKey,  itemConstructor: T): Promise<InstanceType<T>>
+  public async getSchemaItem<T extends typeof SchemaItem>(schemaItemKey: SchemaItemKey, itemConstructor?: T): Promise<SchemaItem | InstanceType<T>> {
+    const schemaItem = await this.schemaContext.getSchemaItem(schemaItemKey);
+    const schemaItemType = itemConstructor?.schemaItemType ?? AbstractSchemaItemType.SchemaItem;
     if (!schemaItem) {
-      throw new SchemaEditingError(ECEditingStatus.SchemaItemNotFoundInContext, new SchemaItemId(schemaItemType, schemaItemKey));
+      throw new SchemaEditingError(ECEditingStatus.SchemaItemNotFoundInContext, new SchemaItemId(schemaItemType as SchemaItemType, schemaItemKey));
```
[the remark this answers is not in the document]
christophermlawson: @rschili I guess we can just use SupportedSchemaItemType in place of SchemaItemType in the identifiers. In the context of the SchemaEditingError usage, this property is simple used to create error messages. So, in the case where the SupportedSchemaItemType is "Class" or "SchemaItem" the message would still be valid.  "SchemaItem FullNameOfSchemaItem could not be found in the schema context".  It will slightly more ambiguous since we don't have the exact type in the message. In what context would the AbstractSchemaItemType be used?  Just curious.

core/ecschema-editing/src/Editing/Editor.ts
StefanApfel-Bentley: The logic should probably live on the schema item as this is actually more a question of schemaItemType than schema.

rschili:
@aruniverse @FlyersPh9 @christophermlawson @StefanApfel-Bentley @naveedkhan8067 @grigasp This is ready to be merged. Colin has approved but given the large number of affected files, I need more approvals. Can you please review/approve? Thanks!

common/changes/@itwin/presentation-common/rob-metadata-dependency_2025-02-06-20-25.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-common",
+      "comment": "adjustments for metadata changes",
```
grigasp, line 5: These comments land onto the consumer-facing `CHANGELOG.md` file, so they should only mention stuff that affects consumers and should be properly formatted (start with upper case, etc.). In this case, the change has no effect to consumers, so I think it should be left empty.

```suggestion
      "comment": "",
```

grigasp (approved):
Approving the change in `presentation-common`. 

I like the addition of `IModelDb.schemaContext`. Can we expect the same on `IModelConnection`?

rschili:
@grigasp 
> Can we expect the same on `IModelConnection`?

Eventually yes, but for now we're limiting changes to backend first.

aruniverse (approved):

PR #7579, opened 2025-01-21
Use sas token from recently queried v2 checkpoint incase the one we currently have is expired

nick4598:
I believe fixes bug where:

1)Backend opens v2 checkpoint
2)Backend closes v2 checkpoint
3)time passes and token on that container expires
4) Backend opens v2 checkpoint on same container, gets 403 trying to get new manfiest because it uses the old token. Resulting in BE_SQLITE_CANTOPEN for newer checkpoints.


nick4598:

nick4598:

aruniverse (approved):

PR #7582, opened 2025-01-21
nuke core-telemetry

aruniverse:
The intent was to kill off bentley/telemetry-client during 3.0, but unfortunately due to lack of time the pkg had to be carried over. Telemetry should be handled by the top level app, not individual library pkgs.

Everything in core-telemetry is internal. And the only usage of it in core at all is IModelApp.telemetry which is also internal


docs/changehistory/NextVersion.md
```diff
@@ -215,10 +209,11 @@ The following APIs were re-exported from `@itwin/core-bentley` and have been rem
...
-| ------------------------------ | ---------------------------------------------------------------------------------------------------------------------- |
-| `@itwin/core-webpack-tools`    | We no longer recommend using [webpack](https://webpack.js.org/) and instead recommend using [Vite](https://vite.dev/). |
-| `@itwin/backend-webpack-tools` | We no longer recommend webpack-ing backends, which was previously recommended to shrink the size of backends.          |
+| Removed                        | Replacement                                                                                                                                                 |
+| ------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `@itwin/backend-webpack-tools` | We no longer recommend bundling backends via webpack, which was previously recommended to shrink the size of backends.                                      |
```
calebmshafer, line 214: Can we add our recommended way to do this now?

docs/changehistory/NextVersion.md
```diff
@@ -215,10 +209,11 @@ The following APIs were re-exported from `@itwin/core-bentley` and have been rem
...
+| `@itwin/core-telemetry`        | No consumable APIs were being published therefore this package has been removed, with no replacement available. Please implement your own telemetry client. |
+| `@itwin/core-webpack-tools`    | We no longer recommend using [webpack](https://webpack.js.org/) and instead recommend using [Vite](https://vite.dev/).                                      |
```
calebmshafer, line 216: Same here, do we have a recommendation for people that were relying on the tools?

calebmshafer (approved):

docs/changehistory/NextVersion.md
aruniverse: We will publish that in a separate PR once that work is done. Issue is being tracked internally

docs/changehistory/NextVersion.md
aruniverse: We dont recommend bundling backends at all, tried to make that more explicit now.

docs/changehistory/NextVersion.md
calebmshafer: sounds good.

aruniverse:
@tcobbs-bentley can you look at failing ios builds? I guess theres an issue with signing? Is this due to resetting the registered devices on our Enterprise iOS developer program after the new year?

Issue #7357, opened 2024-11-13
Report Changeset download progress for `CheckpointManager.downloadCheckpoint`


MindaugasButkus:
**Is your feature request related to a problem? Please describe.**
`CheckpointManager.downloadCheckpoint` accepts an `onProgress` callback which is only used for the Checkpoint itself. It is not propagated to changeset downloading if there are any in between the requested changeset and the latest available checkpoint.

**Describe the solution you'd like**
I would like to be able to get progress report on every changeset being downloaded. That is, not only `total` and `downloaded`, but also the changeset identifier (id, index, or the whole Changeset metadata object). Based on that progress report I would want to either restart the download, or even better - `CheckpointManager` should retry downloading changesets.

**Describe alternatives you've considered**
An alternative would be to manually check which changesets would be downloaded with `CheckpointManager.downloadCheckpoint` and download them before calling it, since it should not re-download what was already downloaded. By downloading changesets manually I could pass in the `onProgress` callback for the changeset download part.

**Additional context**
https://github.com/iTwin/imodels-clients/pull/275 This PR raised my interest. In the past I have seen iTwinJs hang within `CheckpointManager.downloadCheckpoint`. This happened very rarely, so it was pretty much impossible to debug. I now do have reason to believe that it was happening during changeset download.

This is the place where the `onProgress` callback is lost: https://github.com/iTwin/itwinjs-core/blob/master/core/backend/src/CheckpointManager.ts#L356

kabentley:
The whole concept of applying changesets when "downloading a checkpoint" is misguided and should never have been supported. Instead, checkpoints should be checkpoints - if someone wants to open a checkpoint that doesn't exist they should request that it be created and wait for that result. @nick4598

aruniverse:
> if someone wants to open a checkpoint that doesn't exist they should request that it be created and wait for that result. 

@nick4598 @amikoliunas do we have any numbers on how long it takes to create a checkpoint vs this "download an earlier checkpoint and apply changesets" workflow?

GytisCepk:
@MindaugasButkus, is per-changeset download progress reporting still needed on your side, or can this issue be closed ?

MindaugasButkus:
> [@MindaugasButkus](https://github.com/MindaugasButkus), is per-changeset download progress reporting still needed on your side, or can this issue be closed ?

[...]

PR #7587, opened 2025-01-22
Elevate BackendHubAccess to public, remove deprecated 3.x APIs in core-backend

hl662:
- Promote IModelHostOptions.hubAccess to `public`
- Promote `BackendHubAccess` interface to `public`
- Remove deprecated 3.x APIs in core-backend

Notes:
 - getViewThumbnail is [ignored](https://github.com/iTwin/itwinjs-core/pull/7572#issuecomment-2599204454) for now, until a later PR for RPC can address it
 - V1 Checkpoint related deprecated APIs will be ignored and tackled separately once it's tests are addressed

TODO:

- [x]  Run rush extract-api
- [x]  Run rush change
- [ ] Update NextVer.md


common/api/core-backend.api.md
```diff
@@ -346,16 +346,17 @@ export interface AzureBlobStorageCredentials {
...
 
-// @internal
+// @public
 export interface BackendHubAccess {
+    // @internal
     acquireLocks: (arg: BriefcaseDbArg, locks: LockMap) => Promise<void>;
     acquireNewBriefcaseId: (arg: AcquireNewBriefcaseIdArg) => Promise<BriefcaseId>;
     createNewIModel: (arg: CreateNewIModelProps) => Promise<GuidString>;
     deleteIModel: (arg: IModelIdArg & ITwinIdArg) => Promise<void>;
     downloadChangeset: (arg: DownloadChangesetArg) => Promise<ChangesetFileProps>;
```
hl662, line 356: @GytisCepk I ran extract-api and it threw warnings for 2 methods, saying it uses args (`DownloadChangesetArg` and `DownloadChangesetRangeArg`) that are marked as @beta for this API I just converted to @public...  [git blame](https://github.com/iTwin/itwinjs-core/commit/b7192169590e47cd7e09db972b24bdc73b104800#diff-296ede995d70f162e9ac5594f3a209a60cb11782c6e46f971121878a551d363f) showed you most recently worked on this, do you think those args are okay to move to public or should we move `BackendHubAccess` to @beta instead

GytisCepk: I expect `DownloadChangesetRangeArg` to break due to https://github.com/iTwin/itwinjs-core/issues/7357. So we might mark it `@public` to just deprecate it soon after.

That being said, if there is no other concerns about making `BackendHubAccess` public, I would go ahead and mark both `DownloadChangesetArg` and `DownloadChangesetRangeArg` public.

PR #7590, opened 2025-01-22
Update deprecated and outdated devDependencies

wgoehrig:
It's been a while since we've tried to update many of our devDependencies, and as a result the number of deprecation and unmet peer warnings has grown and grown.  Rather than attempting to update all devDependencies en masse, I've started with the deprecation warnings and worked backwards.  With these changes, the 50+ lines of warnings on `rush install` are now down to:

```
 WARN  7 deprecated subdependencies found: circular-json@0.3.3, glob@7.2.3, inflight@1.0.6, json-schema-ref-parser@5.1.3, postinstall-build@5.0.3, rimraf@3.0.2, uuid@3.4.0
...
 WARN  Issues with peer dependencies found
../../core/frontend
└─┬ @vitest/browser 2.1.0
  └─┬ @vitest/mocker 2.1.0
    └── ✕ unmet peer vite@^5.0.0: found 6.0.6

../../full-stack-tests/core
└─┬ @itwin/electron-authorization 0.19.8
  └── ✕ unmet peer electron@">=23.0.0 <34.0.0": found 34.0.0
```

The following dependencies are preventing us from updating/removing those last 7 deprecated packages:

- `typemoq`[^1] has [not been maintained at all](https://github.com/florinn/typemoq/commits/master/) in nearly **7 years**; we should migrate away from it ASAP.
- `json-schema-faker`[^2] has [newer versions available](https://github.com/json-schema-faker/json-schema-faker/tree/master), but I'm concerned that updating may break or introduce flakiness in presentation tests.
- `typescript-json-schema`[^3] is a dependency of `@itwin/presentation-common` and is now in ["maintenance mode"](https://github.com/YousefED/typescript-json-schema/commit/43e557d9c50dc1695dfba1ff83fc50f40f9376ca).  We should consider switching to an alternative package or opening a PR to help them update.
- `nyc`[^3][^4] already has a [PR open](https://www.github.com/istanbuljs/nyc/pull/1565) to update `glob` and `rimraf`
- `babel-plugin-istanbul`[^3] already has a [PR open](https://redirect.github.com/istanbuljs/babel-plugin-istanbul/pull/301) to update `test-exclude` 
- `azurite`[^4][^5] has a [PR open](https://redirect.github.com/Azure/Azurite/pull/2217) for `uuid`, but nothing currently for `rimraf` or `@azure/ms-rest-js` (although there is an [issue open for `@azure/ms-rest-js`](https://redirect.github.com/Azure/Azurite/issues/2470))

Updating `jsdom` led me to remove `jsdom-global` since [it breaks `atob` in jsdom v23](https://redirect.github.com/jsdom/jsdom/issues/3628#issuecomment-1850255716) (and also hasn't been updated in 8 years).  In a few cases I still had to stick some globals from `jsdom` into node's global scope but the [jsdom docs are very clear that this is an anti-pattern](https://github.com/jsdom/jsdom/wiki/Don%27t-stuff-jsdom-globals-onto-the-Node-global).  We should probably consider dropping `jsdom` entirely and moving those tests to vitest-browser.

[^1]: `typemoc` depends on both `circular-json@0.3.3` and `postinstall-build@5.0.3`
[^2]: `json-schema-faker` depends on `json-schema-ref-parser@5.1.3`
[^3]: `typescript-json-schema`, `nyc`, and `babel-plugin-istanbul` (via `test-exclude`) all depend on `glob@7.2.3` (which itself depends on `inflight@1.0.6`)
[^4]: `azurite` and `nyc` (both directly and via `spawn-wrap`) depend on `rimraf@3.0.2` (which itself depends on `glob@7.2.3` and, indirectly, `inflight@1.0.6`)
[^5]: `azurite` (both directly and via `@azure/ms-rest-js`) depends on `uuid@3.4.0`


common/config/rush/pnpm-config.json
```diff
@@ -10,7 +10,8 @@
     ]
   },
   "globalOverrides": {
-    "cross-spawn": "^7.0.5" // https://github.com/advisories/GHSA-3xgq-45jj-v275 npm-run-all>cross-spawn
+    "cross-spawn": "^7.0.5", // https://github.com/advisories/GHSA-3xgq-45jj-v275 npm-run-all>cross-spawn
```
aruniverse, line 13: do we still need this override?

common/config/rush/.pnpmfile.cjs
```diff
@@ -1,3 +1,10 @@
+function overrideToWorkspace(pkg, dependencies) {
+  for (const dep of dependencies) {
+    pkg.dependencies[dep] = "workspace:*";
+    pkg.peerDependencies[dep] = "workspace:*";
```
aruniverse, line 4: do you need to set the peer dep? i guess it doesnt really matter since itll be installed anyway but just curious if this was needed / used to just suppress peer warnings further

extensions/test-extension/package.json
```diff
@@ -33,13 +33,12 @@
...
     "eslint": "^9.13.0",
-    "rimraf": "^3.0.2",
+    "rimraf": "^6.0.1",
     "typescript": "~5.6.2"
   },
   "dependencies": {
     "@itwin/core-extension": "workspace:*",
-    "buffer": "^6.0.3",
-    "url": "^0.11.0"
+    "buffer": "^6.0.3"
```
aruniverse, line 41: i tunk we can nuke buffer too

aruniverse (approved):
thx for volunteering to do this Bill.
i hope we can drop nyc, instabul, typemoq as we move to vitest, and probably wouldnt want to spend any further time updating those deps

extensions/map-layers-formats/package.json
```diff
@@ -47,20 +47,19 @@
...
     "@types/sinon": "^17.0.2",
     "@types/sinon-chai": "^3.2.0",
     "chai": "^4.3.10",
     "chai-as-promised": "^7.1.1",
-    "cpx2": "^3.0.0",
+    "cpx2": "^8.0.0",
     "eslint": "^9.13.0",
-    "ignore-styles": "^5.0.1",
-    "jsdom": "^19.0.0",
-    "jsdom-global": "3.0.2",
```
saskliutas, line 59: Don't know if you seen but there is an up to date alternative called [`global-jsdom`](https://www.npmjs.com/package/global-jsdom). 😄

grigasp (approved):
Approving presentation changes.

For the 3 dependencies likely coming only from our packages:

> `typemoq`[^1] has [not been maintained at all](https://github.com/florinn/typemoq/commits/master/) in nearly **7 years**; we should migrate away from it ASAP.

While I agree we should avoid using it in all new tests (and we do), and it would be great to get rid of it, is having it under `devDependencies` really that big of a problem? We have a huge set of tests which are using the library, and switching to another library would be a huge undertaking. In addition, the tests are for code that's gradually being deprecated and will go away in future releases. With all that in mind, I'd hate to waste anyone's time on this, unless it's a real issue.

> `json-schema-faker`[^2] has [newer versions available](https://github.com/json-schema-faker/json-schema-faker/tree/master), but I'm concerned that updating may break or introduce flakiness in presentation tests.

I'm going to drop it, it's not necessary.

> `typescript-json-schema`[^3] is a dependency of `@itwin/presentation-common` and is now in ["maintenance mode"](https://github.com/YousefED/typescript-json-schema/commit/43e557d9c50dc1695dfba1ff83fc50f40f9376ca).  We should consider switching to an alternative package or opening a PR to help them update.

https://github.com/iTwin/itwinjs-core/issues/7659

extensions/map-layers-formats/package.json
wgoehrig: That's still [going agains jsdom's best practices](https://github.com/jsdom/jsdom/wiki/Don%27t-stuff-jsdom-globals-onto-the-Node-global), no?

saskliutas: I am not sure, just suggested a drop in replacement for deprecated package.

Looking over react testing library docs they suggest using `global-jsdom` when running with mocha. https://testing-library.com/docs/dom-testing-library/setup/#using-without-jest
I would guess that other test runners/frameworks does something similar too.

pmconne (approved):

extensions/map-layers-formats/package.json
wgoehrig: I realize a lot of test runners do something similar, but I think the `jsdom` docs make a good point that you wind up with a strange hybrid franken-environment that doesn't strictly match either a real-world node or browser runtime.

@hl662 would these tests be a good candidate for running in `vitest-browser` maybe?

common/config/rush/.pnpmfile.cjs
wgoehrig: Yeah this was mainly to avoid all those unmet peerDependency warnings, which are just noise

eringram (approved):

extensions/map-layers-formats/package.json
hl662: I would inspect the tests and see how they use the browser APIs before deciding if we need to move to using vitest-browser - I'm not too familiar with this package so I can't say if using a real browser would be better than mocking the browser APIs out.

common/config/rush/pnpm-config.json
aruniverse: ok ignoring for now then

extensions/test-extension/package.json
wgoehrig: nuked it

extensions/map-layers-formats/package.json
saskliutas: In my opinion until test aren't updated to run with vitest browser there is nothing wrong with using global-jsdom. At least we would use standart way to polute node globals instead of our own. 😄

wgoehrig:
@tcobbs-bentley looks like the build is failing due to missing support for a Unicode character class in our iOS environment.  Is ICU somehow disabled on mobile or something?  If not, any other ideas?

FWIW, we can revert the `sinon` update that caused this, but I'm concerned that it's a matter of time before another dependency runs into something like this - seems like we should expect regexes like this to be more common as more and more ReDoS vulnerabilities are found and fixed...

tcobbs-bentley:
@wgoehrig ICU is missing from the mobile builds of Node (both iOS and Android). Unfortunately, the way that we build Node makes it difficult to include. The V8 build is done completely separately from the rest of the build which include ICU4C, and it needs ICU support in order for Node to support ICU. Fixing this _may_ be possible, but I'm not sure how to do so in a reasonable way.

We have run into this problem before and made sure that the problematic code didn't get accessed in mobile builds. The problem with this specific instance is that Node throws an exception while parsing the JS due to the hard-coded RegEx using functionality that only works when ICU is loaded. So that code has to be completely excluded from the mobile build.

I think that a discussion is probably needed between multiple people who have a firm understanding of Bentley buildology if we want to get ICU support into our mobile Node builds.

PR #7599, opened 2025-01-24
Tests and documentation for wrong property tags issue

Abhijeet-Bentley:
imodel-native: https://github.com/iTwin/imodel-native/pull/970

- Added tests for checking behavior of wrong property tags in an ECXml schema.
- Added documentation clearly mentioning what behavior (in this scenario) to be expected across versions.


docs/learning/backend/HandlingUnknownAndWrongSchemaElements.md
```diff
@@ -1,10 +1,19 @@
-## Handling of newer schemas by older ECDb runtimes
+## Handling of schemas with unknown/wrong elements
 
 A schema with a newer ECXml version might contain elements not known to the ECDb runtime being used as it's older.
+Also, a schema might contain wrong elements, for e.g. an ECProperty tag used for denoting a struct inside an entity class.
 A read compatible but write incompatible schema (only ECXml minor version incremented) cannot be imported into the iModel.
-However, the schema can be deserialized and loaded into memory.
+However, any schema can be deserialized and loaded into memory but, the version 3.2 of ECXml schema.
 This schema is also then restricted from being serialized to a file or a string.
```
ben-polinsky, line 7:
```suggestion
A schema with a newer ECXml version might contain elements not recognized by the older ECDb runtime. Additionally, a schema might include incorrect elements, such as an ECProperty tag used to denote a struct inside an entity class. A read-compatible but write-incompatible schema (with only an ECXml minor version increment) cannot be imported into the iModel. However, any schema, except for version 3.2 of the ECXml schema, can be deserialized and loaded into memory. This schema version is also restricted from being serialized to a file or a string.
```

core/ecschema-metadata/src/test/Metadata/Schema.test.ts
```diff
@@ -1235,6 +1235,49 @@ describe("Schema", () => {
...
+
+          _schema = new Schema(_context);
+          try {
+            await _xmlReader.readSchema(_schema, new DOMParser().parseFromString(schemaXml));
+          } catch (err: any) {
+            assert.equal(err.message, `The provided primitive type, Test.PrimStruct, is not a valid PrimitiveType or Enumeration.`);
+          }
+      });
+
+      it.only("should throw an error for current version, when encountered an unknown type", async () => {
```
RohitPtnkr1996, line 1260: Remove .only

ben-polinsky (approved):

naveedkhan8067 (commented):
As you have renamed the `HandlingUnknownSchemaElements.md`, its references should also be updated. For example in `IModelDb.md`:

From:
```
To know more about how these unknown elements are handled, refer to [Handling Unknown Schema Elements](./HandlingUnknownSchemaElements.md)
```

To something like:
```
To know more about how these unknown elements are handled, refer to [Handling Unknown And Wrong Schema Elements For Newer Versions](./HandlingUnknownAndWrongSchemaElementsForNewerVersions.md)
```

core/ecschema-metadata/src/test/Metadata/Schema.test.ts
```diff
@@ -1235,6 +1235,49 @@ describe("Schema", () => {
         deserializeAndTestJSONSync(JSON.stringify(schemaJson), testRelationshipStrength);
         await testSerialization(_schema, false, unsupportedVersionError);
       });
+
+      it("should throw an error for version previous to current, when encountered an unknown type", async () => {
```
ColinKerr, line 1239: Tye typescript code only supports EC spec 3.2 and newer (so only 3.2 because there is no 3.3 yet).

These tests seem like they have the wrong description as well.  The first test is testing the current version but it says `version previous to current`. the second test says `current version` but tests previous version.

Since previous version cannot actually be tested right now and we do not need to handle the inconsistent cases of 2, 3, and 3.1 it should test EC 3.2, and current version plus 1.  Where 3.2 should not allow an unknown type name and current + 1 should.  Just like on the C++ side.

docs/learning/backend/HandlingUnknownAndWrongSchemaElements.md
```diff
@@ -0,0 +1,12 @@
+## Handling of schemas with unknown/wrong elements
```
ColinKerr, line 1: Why new file?

I would keep all of this in the same file for clarity.  Also we do not know if it's wrong or just unknown when the spec is newer.  So I would leave the file name the same.

core/ecschema-metadata/src/test/Metadata/Schema.test.ts
Abhijeet-Bentley: I have deleted the 3.1 test case and just kept the test case for 3.2 version. Rohit has got a test which already checks behavior for 3.3 version. Also, the description is now edited correctly.

docs/learning/backend/HandlingUnknownAndWrongSchemaElements.md
Abhijeet-Bentley: Rohit and I discussed, and we think having separate file for newer versions makes it easier since,
we document behavior for various versions in separate file. Although, taking your point into consideration,
I have changed the file name and removed "wrong" from file name.

ColinKerr: Here are some examples of how we combine both older/new version documentation together to make it more clear to the reader:
https://www.itwinjs.org/bis/guide/schema-evolution/schema-versioning-and-generations/#application-schema-compatibility-logic


This one is still WIP but it includes both forwards and backwards compatibility together:
https://github.com/iTwin/itwinjs-core/pull/7558/files

core/ecschema-metadata/src/test/Metadata/Schema.test.ts
ColinKerr: The test you kept is for the current version, not for EC 3.2 ... so it will change with the next version of EC spec we add.  I suggest hard coding this test to be on EC 3.2 because this is the version we want to start enforcing the rule on.

I don't see the +1 test for this specific unknown behavior.  These two tests should be co-located so someone will find one if they find the other.

Abhijeet-Bentley: Alright, I will hard code it 3.2 version. Rohit's test case : "Schema XML with property having an unknown primitive type", checks for defaulting behavior in newer versions.

docs/learning/backend/HandlingUnknownAndWrongSchemaElements.md
Abhijeet-Bentley: I have merged the two files into one.

rschili (approved):

core/backend/src/test/ecdb/ECSqlStatement.test.ts
```diff
@@ -3127,8 +3127,8 @@ describe("ECSqlStatement", () => {
           <ECProperty propertyName="booleanProperty" typeName="boolean"/>
           <ECProperty propertyName="blobProperty" typeName="binary"/>
           <ECProperty propertyName="doubleProperty" typeName="double"/>
-          <ECProperty propertyName="customIdProperty" typeName="id"/>
-          <ECProperty propertyName="customIdSetProperty" typeName="idSet"/>
```
rschili, line 3131: I'm confused about this one. Is this actual logic, or did the author of the test intend "string" and misspell? There is a string property alongside, so I want to understand what the point is of these two types...

ColinKerr: Here is the PR that added the test: https://github.com/iTwin/itwinjs-core/pull/5897/files

It looks like a mistake, a navigation property would be technically correct and I guess longs next so.  I'm not sure what if any validation the binders do on the type of the target, so maybe it doesn't matter if we get the types correct

rschili: Abhijeet asked me if we can close this. I'm actually not sure. If this is treated as string, then I don't think it adds value to this test and should be removed.
If we want to keep it to preserve testing if this fallback-scenario keeps working, it should probably be moved into a separate test.
I feel it's probably best to just remove those two..

RohitPtnkr1996: The test case is using those 2 ec props to test the bindId method of the QueryBinder for an insert and a select.

I thought the schema setup part of it was just to have a property to check the bindId against and served no purpose other than that.
Hence why the defaulting it to string until now also didn't cause any failures.

rschili: Ah, yeah makes sense, alright let's keep it, I'm closing this comment

ColinKerr (approved):

PR #7604, opened 2025-01-24
GoogleMaps support

mdastous-bentley:
New imagery format / proviter to support Google maps 2D tiles:
image

More details available in NextVersion.md.


core/common/src/MapLayerSettings.ts
```diff
@@ -19,6 +19,10 @@ export type ImageryMapLayerFormatId  = "ArcGIS" | "BingMaps" | "MapboxImagery" |
 /** @public */
 export type SubLayerId = string | number;
 
+/** @public */
+export type  PropertyBagArrayProperty = Array<number|string|boolean>;
+export interface PropertyBag { [key: string]: number | string | boolean | PropertyBagArrayProperty };
```
pmconne, line 24: See if you can avoid reinventing [JSONSchema](https://github.com/iTwin/itwinjs-core/blob/master/core/bentley/src/JsonSchema.ts).

mdastous-bentley:
@pmconne Part of this PR, I need to have an async version of `TileTreeReference.addLogoCards`, mainly because a request need to Google web API is required  to retrieve copyright strings associated with selected tiles.  Currently it's being invoked through` Viewport.forEachTileTreeRef`, which is not async. 

My only solution right now, would consist in created an `async` version of the forEach methods, but I feel like it will result in a ugly API.

Any thoughts?

core/common/src/MapLayerSettings.ts
mdastous-bentley: `export type JSONSchemaType = string | number | boolean | object | JSONSchemaType[];`

I need a type representing types that can be directly serialized and can be compared easily.  This is why I excluded `object`, and I didn't want to support array of arrays.

Is there a way to compose type from existing types with typescript ?

pmconne:
> My only solution right now, would consist in created an async version of the forEach methods, but I feel like it will result in a ugly API.

The `forEach` methods are a relic of the days when we were still learning how to do TypeScript.
I'd suggest supplanting them with synchronous method(s) that return `Iterable<TileTreeReference>`. Then the caller can worry about calling async methods on the references, halting iteration partway through, doing something with the return value of whatever he's calling on the references, etc etc.

core/common/src/MapLayerSettings.ts
pmconne: Do the google maps property bags definitely conform to those constraints? Do we anticipate other map providers that might not meet those constraints?

mdastous-bentley: Currently for google it only holds this data structure:
```
export interface CreateGoogleMapsSessionOptions {
  mapType: GoogleMapsMapType;
  language: string;     
  region: string;    
  orientation?: number;
  layerTypes?: string[];
};
```
My thinking is that MapLayerSettings's properties are meant to be simple list key,value pairs that can be passed down to the provider, it's not meant to pass complex data structures.  If nested objects need to be passed, it can always be simplified into a flat list of values and re-construct inside the provider.

core/bentley/src/Compare.ts
```diff
@@ -92,3 +92,47 @@ export function areEqualPossiblyUndefined<T, U>(t: T | undefined, u: U | undefin
...
+
+export function comparePrimitiveArrays (lhs?: Array<number|string|boolean>, rhs?: Array<number|string|boolean> ) {
+  if (undefined === lhs)
+    return undefined === rhs ? 0 : -1;
+  else if (undefined === rhs)
+    return 1;
+  else if (lhs.length === 0 && rhs.length === 0) {
+    return 0;
+  } else if (lhs.length !== rhs.length) {
+    return Math.abs(lhs.length - rhs.length);
```
pmconne, line 127: Why absolute value? It breaks the [strict weak constraint].

mdastous-bentley: right, I will remove it.

core/common/src/MapLayerSettings.ts
pmconne: Oh, so this property bag is not something supplied by the maps service, it's a way for the user of our API to configure specific map providers?

mdastous-bentley: Exactly.. sorry for the confusion, I should have better explained this.

pmconne: Thanks. I'd suggest using less generic names than "PropertyBag" and "PrimitiveType" (especially since EC already has *two* different "primitive type" types).

mdastous-bentley:
@pmconne @aruniverse @matmarchand  Looking for feedback on the actual implementation state:
1. The new maplayer format / provider was added to `@itwin/map-layers-formats` package.  Meaning, the package has to be installed in your iTwin application to get GoogleMaps support.
2. Also, there's no new value built-in value for `BackgroundMapProviderName` to use with `BaseMapLayerSettings.fromProvider`
3. Introduction of a new `properties` field on `MapLayerSettings` to pass provider specific properties, which will be part of the display style serialization.
4. An utility `GoogleMaps` object is provided (could also be merged on `GoogleImageryFormat` class), is provided to easily create MapLayerSettings from GoogleMaps API parameters (i.e. you don't have to provide the property bag or the service URL by hand).
5. We don't provided any GoogleMaps API key, each product will have to provider their own.

mdastous-bentley:
@pmconne An async version of `TileTreeReference.addLogoCards` will be needed, what strategy do you suggest to replace an existing public API?

core/common/src/MapLayerSettings.ts
mdastous-bentley: I renamed `PropertyBag` to `MapLayerProviderProperties`.  If you have better terminology, let me know.

 Also, I now refer `number|string|boolean` as `SimpleType`. I'm unsure if it deserve a type definition in `@itwin/core-bentley` though.

pmconne:
> An async version of TileTreeReference.addLogoCards will be needed, what strategy do you suggest to replace an existing public API?

The approach you've taken (deprecate it, add new `addAttributions` replacement) looks ok on the face of it but will introduce bugs. Existing implementations (outside of itiwnjs-core, where you've added the new method) will not have overridden the new method, so if you update callers of `addLogoCards` to call `addAttributions` instead, suddenly some attributions will go missing.

The simplest solution would be to change `addLogoCards`' return type to `void | Promise<void>` but that's a breaking change.

All but one of your `addAttributions` methods is just a copy-paste of `addLogoCards`. You could instead do something like:
```ts
class TileTreeReference {
  // Everyone who provides attributions today overrides this method.
  /** @deprecated in 5.0. Implementations should instead override [[addLogoCardsAsync]]. Callers should instead call [[addAttributions]]. */
  public addLogoCards(_cards: HTMLTableELement, _vp: ScreenViewport): void { }

  // People who need async attributions, or those reacting to deprecation warnings, would override this method instead of addLogoCards.
  protected addLogoCardsAsync(cards: HTMLTableElement, vp: ScreenViewport): Promise<void> {
    return Promise.resolve(this.addLogoCards(cards, vp);
  }

  public addAttributions(cards: HTMLTableElement, vp: ScreenViewport): Promise<void> {
    return this.addLogoCardsAsync ? this.addLogoCardsAsync(cards, vp) : this.addLogoCards(cards, vp);
  }
}
```

mdastous-bentley:
> > An async version of TileTreeReference.addLogoCards will be needed, what strategy do you suggest to replace an existing public API?
> 
> The approach you've taken (deprecate it, add new `addAttributions` replacement) looks ok on the face of it but will introduce bugs. Existing implementations (outside of itiwnjs-core, where you've added the new method) will not have overridden the new method, so if you update callers of `addLogoCards` to call `addAttributions` instead, suddenly some attributions will go missing.
> 
> The simplest solution would be to change `addLogoCards`' return type to `void | Promise<void>` but that's a breaking change.
> 
> All but one of your `addAttributions` methods is just a copy-paste of `addLogoCards`. You could instead do something like:
> 
> ```ts
> class TileTreeReference {
>   // Everyone who provides attributions today overrides this method.
>   /** @deprecated in 5.0. Use [[addAttributions]] instead. */
>   public addLogoCards(_cards: HTMLTableELement, _vp: ScreenViewport): void { }
> 
>   // People who need async attributions, or those reacting to deprecation warnings, would override this method instead of addLogoCards.
>   protected addAttributions(cards: HTMLTableElement, vp: ScreenViewport): Promise<void> {
>     return Promise.resolve(this.addLogoCards(cards, vp);
>   }
> }
> ```

Thanks for the suggestion, I totally agree too much code duplication in previous approach.

pmconne:
> Looking for feedback on the actual implementation state

I guess the user interface is going to have to expose the provider-specific config options (property bag)? Should the provider expose a schema describing the available properties and their types?

mdastous-bentley:
> > Looking for feedback on the actual implementation state
> 
> I guess the user interface is going to have to expose the provider-specific config options (property bag)? Should the provider expose a schema describing the available properties and their types?
> 
Currently it's the job of the `GoogleMaps` object: it provides an interfaced name `CreateGoogleMapsSessionOptions`, and utility methods such as `createPropertiesFromSessionOptions` and `createMapLayerSettings` so that you never have to deal directly with the properties bag.

It doesn't offer though, a set of abstract types that could be shared among future providers interested in using properties bag.

mdastous-bentley:
@pmconne There are function inside `GoogleMaps.ts` that I would to keep private, meaning it should not be discoverable in the typing.  Do we have any utilities in this repo to achieve that?

pmconne:
> @pmconne There are function inside `GoogleMaps.ts` that I would to keep private, meaning it should not be discoverable in the typing. Do we have any utilities in this repo to achieve that?

[Guidelines for internal APIs](https://www.itwinjs.org/learning/guidelines/release-tags-guidelines/#internal-apis). Is this what you mean?

mdastous-bentley:
> > @pmconne There are function inside `GoogleMaps.ts` that I would to keep private, meaning it should not be discoverable in the typing. Do we have any utilities in this repo to achieve that?
> 
> [Guidelines for internal APIs](https://www.itwinjs.org/learning/guidelines/release-tags-guidelines/#internal-apis). Is this what you mean?

Not quite, `@internal` items are still visible when consuming the API, which I find rather confusing for our API consumers.

pmconne:
> @internal items are still visible when consuming the API

Only if you export them from the package.
Be more specific about which APIs you are trying to keep private, and what you mean by "private" (only accessible in the same package? only in monorepo? only for other special repos?)

extensions/map-layers-formats/src/GoogleMaps/GoogleMaps.ts
```diff
@@ -0,0 +1,268 @@
...
+  createBaseLayerSettings: (opts?: CreateSessionOptions) => {
+    return BaseMapLayerSettings.fromJSON(_internal.createMapLayerProps("GoogleMaps", opts));
+  }
+};
+
+/**
+ * Internal function export for testing purposes, do not use.
+ * @internal
+ * */
+export const _internal = {
```
mdastous-bentley, line 262: @pmconne To keep the API as simple as simple, I went for this strategy..

pmconne: Which package(s) import and use this?

mdastous-bentley: @itwin/map-layers-formats's tests

pmconne: Put your _internal functions in a separate file under src/internal/. If the tests only use ESModules, import the functions directly from @itwin/map-layers-formats/lib/esm/whatever. If the tests use both ESModules and CommonJS, export your functions from src/internal/cross-package.ts.

mdastous-bentley: Actually, some of those "internal" functions are also being used inside GoogleMaps.ts. Does your suggestion still apply?

pmconne: Yes, import them into GoogleMaps.ts from the internal file.

pmconne: I pushed some adjustments to make your tests build now. The tests are in the same package as the internal APIs, so cross-package.ts isn't needed. If you intend to make additional changes that will consume those APIs from some other package, then we'll need cross-package.ts.

mdastous-bentley:
@aruniverse vitest failing `rush audit`, Im assuming someone is already on it ?

aruniverse:
> @aruniverse vitest failing `rush audit`, Im assuming someone is already on it ?

Yes #7710

mdastous-bentley:
```
 ❯  chromium  src/test/render/RealityMeshParams.test.ts (4 tests | 1 failed) 16089ms
   × RealityMeshParamsBuilder > supports 8-, 16-, and 32-bit indices 15852ms
     → Test timed out in 15000ms.

```
@aruniverse is this a know flacky test? Failed on Windows only

aruniverse:
> ```
>  ❯  chromium  src/test/render/RealityMeshParams.test.ts (4 tests | 1 failed) 16089ms
>    × RealityMeshParamsBuilder > supports 8-, 16-, and 32-bit indices 15852ms
>      → Test timed out in 15000ms.
> ```
> 
> @aruniverse is this a know flacky test? Failed on Windows only

Did you retry it?

mdastous-bentley:
> > ```
> >  ❯  chromium  src/test/render/RealityMeshParams.test.ts (4 tests | 1 failed) 16089ms
> >    × RealityMeshParamsBuilder > supports 8-, 16-, and 32-bit indices 15852ms
> >      → Test timed out in 15000ms.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @aruniverse is this a know flacky test? Failed on Windows only
> 
> Did you retry it?

yup, and it worked

mdastous-bentley:
@pmconne passing 'checks' now, can you have a second look pls?

docs/changehistory/NextVersion.md
```diff
@@ -94,6 +96,32 @@ The Presentation system is moving towards a more modular approach, with smaller
 
 The unified selection system has been part of `@itwin/presentation-frontend` for a long time, providing a way for apps to have a single source of truth of what's selected. This system is now deprecated in favor of the new [@itwin/unified-selection](https://www.npmjs.com/package/@itwin/unified-selection) package. See the [migration guide](https://github.com/iTwin/presentation/blob/master/packages/unified-selection/learning/MigrationGuide.md) for migration details.
 
+## Map layers
+
+### Google Maps 2D tiles API
+
+The `itwin\map-layers-formats` package now includes an API for consuming Google Maps 2D tiles.
```
pmconne, line 103:
```suggestion
The `itwin/map-layers-formats` package now includes an API for consuming Google Maps 2D tiles.
```

docs/changehistory/NextVersion.md
```diff
@@ -94,6 +96,32 @@ The Presentation system is moving towards a more modular approach, with smaller
...
+
+To enable it as a base map, it's simple as:
+
+ ```typescript
+import { GoogleMaps } from "@itwin/map-layers-formats";
+const ds = IModelApp.viewManager.selectedView.displayStyle;
```
pmconne, line 109: Please consider putting sample code into the example-code-snippets package, to ensure it compiles now and in the future and can be referenced from the API docs as well. (Your example here has a compilation error, for example). This also allows you to provide a more complete example for someone to quickly integrate Google Maps into their own app.

docs/changehistory/NextVersion.md
```diff
@@ -140,6 +168,10 @@ The unified selection system has been part of `@itwin/presentation-frontend` for
 
 - Deprecated [HiliteSet.setHilite]($core-frontend) - use `add`, `remove`, `replace` methods instead.
 
+- Deprecated [addLogoCards]($core-frontend)-related APIs:
```
pmconne, line 171:
```suggestion
- Deprecated synchronous [addLogoCards]($core-frontend)-related APIs in favor of new asynchronous ones:
```

common/api/core-common.api.md
```diff
@@ -5727,6 +5731,15 @@ export interface MapLayerKey {
 // @public
 export type MapLayerProps = ImageMapLayerProps | ModelMapLayerProps;
 
+// @beta (undocumented)
+export type MapLayerProviderArrayProperty = Array<number | string | boolean>;
+
+// @beta (undocumented)
```
pmconne, line 5737: `@beta` APIs require documentation, here and throughout.

common/api/map-layers-formats.api.md
```diff
@@ -56,6 +58,17 @@ export class ArcGisFeatureProvider extends ArcGISImageryProvider {
     get tileSize(): number;
 }
 
+// @beta
+export interface CreateSessionOptions {
```
pmconne, line 62: This interface and the LayerTypes and MapTypes types it references seem to be specific to GoogleMaps yet they all have very generic names, are they supposed to be applicable to other map imagery providers?

core/bentley/src/Compare.ts
```diff
@@ -92,3 +92,48 @@ export function areEqualPossiblyUndefined<T, U>(t: T | undefined, u: U | undefin
   else
     return areEqual(t, u);
 }
+
+/** @beta */
+export function compareSimpleTypes(lhs: number | string | boolean, rhs: number | string | boolean): number {
+  // Make sure the types are the same
+  if (typeof lhs !== typeof rhs) {
+    return 1;
```
pmconne, line 100: Violates strict-weak ordering, return the comparison of the strings instead to produce consistent ordering.

common/api/core-common.api.md
```diff
@@ -5727,6 +5731,15 @@ export interface MapLayerKey {
 // @public
 export type MapLayerProps = ImageMapLayerProps | ModelMapLayerProps;
 
+// @beta (undocumented)
+export type MapLayerProviderArrayProperty = Array<number | string | boolean>;
```
pmconne, line 5735: This type (used in MapLayerProviderProperties and compareSimpleArrays) - are you sure you didn't mean `number[] | string[] | boolean[]`? The way you have it now permits things like `{ array: ["abc", 123, true] }`.

core/common/src/MapLayerSettings.ts
```diff
@@ -311,6 +329,11 @@ export class ImageMapLayerSettings extends MapLayerSettings {
     if (props.queryParams) {
       this.savedQueryParams = {...props.queryParams};
     }
+
+    if (props.properties) {
+      this.properties = {...props.properties}
```
pmconne, line 334: Note: this does not deep-copy arrays, if you are using spread operator due to concerns about mutating the input.

core/frontend/src/tile/map/ImageryTileTree.ts
```diff
@@ -333,12 +347,46 @@ class ImageryMapLayerTreeSupplier implements TileTreeSupplier {
...
+                  const rhsKeysLength = Object.keys(rhs.settings.properties).length;
+
+                  if (lhsKeysLength !== rhsKeysLength) {
+                    cmp = lhsKeysLength - rhsKeysLength;
+                  } else {
+                    for (const key of Object.keys(lhs.settings.properties)) {
+                      const lhsProp = lhs.settings.properties[key];
+                      const rhsProp = rhs.settings.properties[key];
+                      if (typeof lhsProp !== typeof rhsProp) {
+                        cmp = 1;
```
pmconne, line 362: Strict-weak violation here and on lines 377 and 379.

core/frontend/src/tile/map/ImageryTileTree.ts
```diff
@@ -333,12 +347,46 @@ class ImageryMapLayerTreeSupplier implements TileTreeSupplier {
...
+                        break;
+                      }
+                      if (Array.isArray(lhsProp) || Array.isArray(rhsProp)) {
+                        cmp = compareSimpleArrays(lhsProp as (number | string | boolean)[], rhsProp as (number | string | boolean)[]);
```
pmconne, line 366: Your cast is to a homogeneous array but MapLayerProperties and compareSimpleArrays operate on heterogeneous arrays. Which did you intend?

extensions/map-layers-formats/src/GoogleMaps/GoogleMapDecorator.ts
```diff
@@ -0,0 +1,95 @@
+import { CanvasDecoration, DecorateContext, Decorator, IconSprites, IModelApp, Sprite } from "@itwin/core-frontend";
+import { Point3d } from "@itwin/core-geometry";
+import { MapTypes } from "./GoogleMaps";
+
+
+/** A simple decorator that show logo at the a given screen position.
+ * @internal
+ */
+export class LogoDecoration implements CanvasDecoration {
```
pmconne, line 9: Eventually we're going to have to deal with multiple logos trying to overlay on the same area of the viewport (e.g., the imagery provider and the terrain provider). It's fine for now.

extensions/map-layers-formats/src/GoogleMaps/GoogleMapDecorator.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /** Activate the logo based on the given map type. */
+  public async activate(mapType: MapTypes): Promise<boolean> {
+    // Pick the logo that is the most visible on the background map
+    const imageName = mapType === "roadmap" ?
+    "google_on_white_hdpi" :
+    "google_on_non_white_hdpi";
+
+    // We need to move the logo right after the 'i.js' button
+    this.logo.offset = new Point3d(45, 5);
```
pmconne, line 83: Fragile, hope we don't decide to move or resize the i.js button. Not sure what you can do about it now.

pmconne (commented):
Implementation itself looks good, I only have comments on smaller details.

pmconne (approved):

common/api/map-layers-formats.api.md
mdastous-bentley: It's meant to be GoogleMaps specific.  Initially it contained the 'GoogleMaps' prefix but found that long and redundant.  So better to have the GoogleMapsXYZ terminology ?

pmconne: Yes, the names are going into a global scope so they need to indicate they are specific to google maps.

mdastous-bentley: Using a namespace is not an option right?

core/common/src/MapLayerSettings.ts
mdastous-bentley: good catch, what about `structuredClone` ?

docs/changehistory/NextVersion.md
mdastous-bentley: Will do.
I have seen example-code being referenced in .md files, can you do the same with JSDoc's @example ?

common/api/map-layers-formats.api.md
pmconne: It doesn't solve your problem of names being too long. If you put them in a `GoogleMaps` namespace then the name becomes `GoogleMaps.LayerType` instead of `GoogleMapsLayerType`, not an improvement IMO and namespaces have other downsides.

core/common/src/MapLayerSettings.ts
pmconne: If you're truly concerned, sure. Or just document that people shouldn't modify the input after passing it to you.

docs/changehistory/NextVersion.md
pmconne: [Here's an example](https://github.com/iTwin/itwinjs-core/blob/7e3c03b4c5027e5a4085a5ed42b3d9f34d338cee/core/frontend/src/tile/RealityTileTree.ts#L140) of referencing example code from API docs.

docs/changehistory/NextVersion.md
```diff
@@ -95,6 +96,30 @@ The Presentation system is moving towards a more modular approach, with smaller
 
 The unified selection system has been part of `@itwin/presentation-frontend` for a long time, providing a way for apps to have a single source of truth of what's selected. This system is now deprecated in favor of the new [@itwin/unified-selection](https://www.npmjs.com/package/@itwin/unified-selection) package. See the [migration guide](https://github.com/iTwin/presentation/blob/master/packages/unified-selection/learning/MigrationGuide.md) for migration details.
 
+## Google Maps 2D tiles API
+
+The `itwin/map-layers-formats` package now includes an API for consuming Google Maps 2D tiles.
```
aruniverse, line 101:
```suggestion
The `@itwin/map-layers-formats` package now includes an API for consuming Google Maps 2D tiles.
```

docs/changehistory/NextVersion.md
mdastous-bentley: done

common/api/map-layers-formats.api.md
mdastous-bentley: done, added'GoogleMaps' prefix

common/api/core-common.api.md
mdastous-bentley: I prefer the strict approach, so I changed it to `number[] | string[] | boolean`

core/bentley/src/Compare.ts
mdastous-bentley: fixed

core/bentley/src/Compare.ts
```diff
@@ -92,3 +92,48 @@ export function areEqualPossiblyUndefined<T, U>(t: T | undefined, u: U | undefin
...
+      break;
+    case "string":
+      cmp = compareStrings(lhs, rhs as string);
+      break;
+    case "boolean":
+      cmp = compareBooleans(lhs, rhs as boolean);
+      break;
+  }
+  return cmp;
+}
```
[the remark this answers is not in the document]
mdastous-bentley: done

mdastous-bentley:
@aruniverse @eringram One of you need to approve I think

extensions/map-layers-formats/src/internal/GoogleMapsUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+  },
+
+  /**
+ * Creates a Google Maps layer props.
+ * @param name Name of the layer (Defaults to "GoogleMaps")
+ * @param opts Options to create the session  (Defaults to satellite map type, English language, US region, and roadmap layer type)
+ * @internal
+*/
+  createMapLayerProps: (name: string = "GoogleMaps", opts?: GoogleMapsCreateSessionOptions): ImageMapLayerProps => {
+    GoogleMapsUtils.registerFormatIfNeeded();
```
aruniverse, line 52: should we just throw instead if the format isnt registered? users should be calling MapLayersFormats.initialize right?

mdastous-bentley: yea we can just throw.  I was unsure about this one, wanted to make it easier for newcomers but probably not necessary.

aruniverse (approved):
newly added beta apis exposed to users seem fine to me in map-layers-format

i think a lot is left to be desired here for formatting of the code but trying to not be picky, and hope we come back to it later

test-apps/display-test-app/src/common/DtaConfiguration.ts
```diff
@@ -141,6 +143,9 @@ export const getConfig = (): DtaConfiguration => {
   if (undefined !== process.env.IMJS_BING_MAPS_KEY)
     configuration.bingMapsKey = process.env.IMJS_BING_MAPS_KEY;
 
+  if (undefined !== process.env.IMJS_GOOGLE_MAPS_KEY)
```
eringram, line 146: Could you add these environment variables to the DTA README?

eringram (approved):

test-apps/display-test-app/src/common/DtaConfiguration.ts
mdastous-bentley: done

mdastous-bentley:
> newly added beta apis exposed to users seem fine to me in map-layers-format
> 
> i think a lot is left to be desired here for formatting of the code but trying to not be picky, and hope we come back to it later

Any example of what you are referring to ?

test-apps/display-test-app/src/common/DtaConfiguration.ts
eringram: Thanks. Did you push it? I don't see it in the diff and the only new commits I see are merging master in

mdastous-bentley: push got blocked, should be good now

core/bentley/src/Compare.ts
mdastous-bentley: @pmconne only this comment is left unresolved

mdastous-bentley:

test-apps/display-test-app/README.md
```diff
@@ -190,6 +190,8 @@ You can use these environment variables to alter the default behavior of various
   * If defined, sets the MapBox key for the `MapLayerOptions` as an "access_token".
 * IMJS_BING_MAPS_KEY
   * If defined, sets a Bing Maps key within the `MapLayerOptions` as a "key" type.
+* IMJS_BING_MAPS_KEY
```
aruniverse, line 193: bad copy and paste,


```suggestion
* IMJS_GOOGLE_MAPS_KEY
```

PR #7623, opened 2025-01-28
Fix error when converting multiple units parallel with the same UnitConverter instance

StefanApfel-Bentley:
When running multiple unit conversions parallel using the same UnitConverter instance, the conversion failed because the UnitGraph used in the converter is returning an intermediate result when called again before a unit preperation has been completed. This results in incomplete unit graphs.


core/ecschema-metadata/src/UnitConversion/UnitConverter.ts
```diff
@@ -51,18 +48,19 @@ export class UnitConverter {
       });
     }
 
-    const from = await this._uGraph.resolveUnit(fromSchemaItemName, fromSchema);
-    const to = await this._uGraph.resolveUnit(toSchemaItemName, toSchema);
+    const unitGraph = new UnitGraph(this._context);
```
ColinKerr, line 51: On the surface the UnitGraph has checks to ensure adding the same unit twice doesn't break unit conversion.  What is actually happening that causes this to break.

Any performance benefit from reusing the UnitGraph probably isn't a big performance issue because the caller should be caching UnitConversions returned.  But recreating the UnitGraph on each call makes a UnitConverter object somewhat pointless.  

@rschili @hl662

StefanApfel-Bentley: Looks like when calling addUnit in concurrent setups, the addUnit leaves early when it finds the unit in it's graph already (UnitGraph.hasNode). It may occur though, that at this point, the graph is not fully build yet (aka promise that adds the unit and it subcomponents is still running), so the method was left early. Now, the subsequent reduce calls in the UnitConverter produce an incomplete map which makes the base unit check fail.

I pushed a fix, where we cache incomplete addUnit operations and if a unit is added again, it won't leave the method before the adding-in-progress-promise has been resolved.

ColinKerr (approved):

PR #7632, opened 2025-01-29
Update Android SDK

tcobbs-bentley:
Compile SDK: 34
Min SDK: 31
Gradle: 8.7

Also fix Android integration test and reenable running it in CI pipelines.


aruniverse (approved):

tcobbs-bentley:
@aruniverse FYI, I made a lot of changes to runAndroidEmulator.py since your original approval.

tcobbs-bentley:
After working out various issues with the new emulator to get things to the point where it does run display-test-app in the emulator, I have now discovered that it is crashing during the run as part of an OpenSSL operation. I am investigating that to work out the cause of the crash so I can fix it.

tcobbs-bentley:
The crash appears to be caused by an uninitialized variable in an Android-only patch that Bentley made to the OpenSSL being used by the native add-on. I have submitted a [PR](https://github.com/iTwin/imodel-native/pull/986) in the imodel-native repository to fix that uninitialized variable. Once that PR merges and another add-on build is released, hopefully it will fix the problem, and display-test-app will once again work in Android.

pmconne (approved):

tcobbs-bentley:
The source of the crash was tracked down to an uninitialized variable in an android-only code path in imodel-native: https://github.com/iTwin/imodel-native/pull/986. Now that this has been fixed, the CI should succeed.

PR #7637, opened 2025-01-31
Drape model map layers onto reality models

markschlosseratbentley:
This PR adds support to iTwin.js for draping model map layers onto reality data.

A new property titled `drapeTarget` has been added to [ModelMapLayerProps](https://www.itwinjs.org/reference/core-common/maplayers/modelmaplayerprops/) and [ModelMapLayerSettings](https://www.itwinjs.org/reference/core-common/maplayers/modelmaplayersettings/). When this property is specified as `ModelMapLayerDrapeTarget.RealityData`, the model map layer will be only draped onto all attached reality data. Supported reality data formats include glTF or b3dm. If `drapeTarget` is not specified in the properties, it will default to `ModelMapLayerDrapeTarget.BackgroundMap`, which will only drape the model map layer onto the background map.

TODO:
- [x] Performance testing - results look adequate. Consulted with @MarcNeely.
- [x] Run Image Testing - pixel diff should look good (preliminary results look good)
- [x] Add Image Tests test cases.
- [x] Add unit test.
- [x] NextVersion + screenshot
- [x] Fix broken unit test: https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3724311&view=logs&j=751a5ea6-e147-5a0b-b7c5-e28ecfcb00e2&t=742556c9-a44d-5542-8a03-cf20a9c5ac3e&l=11638
- [x] Fix broken unit test: https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3726877&view=logs&j=751a5ea6-e147-5a0b-b7c5-e28ecfcb00e2&t=742556c9-a44d-5542-8a03-cf20a9c5ac3e&l=11727

TODO (Future PRs):
- [ ] Drape onto iModels
- [ ] Highlighting / selecting draped models does not work when they are draped on reality data
- [ ] Saved view `pipes_draped_onto_reality_model_off` in `geolocated_pipes.bim` is not rendering the draping in DPTA (Image Tests). It is rendering in display-test-app locally.  Investigating.  The other draping saved views work in DPTA (Image Tests). Issue: https://github.com/iTwin/itwinjs-core/issues/7818

**Here is an example of draping a model from within an iModel (the piping in the air) onto some glTF reality data (the terrain underneath):**
image

**Here is an example of the same dataset as above with the model itself disabled, so you only see the draping onto the glTF reality data:**
image

To test locally, you can open display-test-app and:
- open an iModel
- attach a reality model located somewhere near the iModel
- utilize the keyin `fdt attach model maplayer reality` while an element is selected - the model associated with that element will drape onto the reality model.


markschlosseratbentley:
If this ends up being a generally correct direction, the `initializeLayers()` function (and related code) added to RealityModelTileTree.ts could probably be refactored so `MapTileTreeReference` and `RealityModelTileTree.Reference` inherit from their shared parent.

danielzhong-bentley:
We are working on making the reality tile refresh automatically whenever changes occur, such as draping a map layer or detaching/removing map layer textures. We found a function in `Tiles.ts` called `getTileTreeOwner`, which states that it creates a new dictionary whenever the ID changes. However, the ID for the map tile tree does not contain any information related to map layers. We are still unsure how map layer changes trigger the automatic reload of the map tile, leaving us uncertain if we are on the right track.
 
@pmconne Could you please point us to where the tree owner keeps track of map layer changes so that the tile updates/reloads when a change occurs? Thanks!

core/frontend/src/Viewport.ts
```diff
@@ -917,6 +919,60 @@ export abstract class Viewport implements Disposable, TileUser {
     return true;
   }
 
+  /** Refresh the Reality Tile Tree to reflect changes in the map layer. */
+  private refreshRealityTile(): void {
```
danielzhong-bentley, line 923: This helper function is designed to refresh the reality tile. Specifically four our case, we use it when the map layer changes.

core/frontend/src/render/webgl/RealityMesh.ts
```diff
@@ -300,11 +302,53 @@ export class RealityMeshGeometry extends IndexedGeometry implements RenderGeomet
...
+      layerClassifiers
+    };
+
+    const layerTextures: TerrainOrProjectedTexture[] = texture ? [texture] : [];
+
+    layerClassifiers?.forEach((layerClassifier, layerIndex) => layerTextures[layerIndex] = new ProjectedTexture(layerClassifier, realityMeshParams, realityMeshParams.tileRectangle));
 
-    return new RealityMeshGeometry({ realityMeshParams: params, textureParams: texture ? RealityTextureParams.create([texture]) : undefined, baseIsTransparent: false, isTerrain: false, disableTextureDisposal });
+    return new RealityMeshGeometry({ realityMeshParams: params, textureParams: layerTextures.length > 0 ? RealityTextureParams.create(layerTextures) : undefined, baseIsTransparent: false, isTerrain: false, disableTextureDisposal });
   }
```
danielzhong-bentley, line 352: Modified this function to support reality tile draping. 

**Structural Difference:**
This function was called only once when I3dmReader or B3dmReader loaded files.

In contrast, compared to the map tile's method (`createGraphic()` within the same file),
the `createGraphic()` function is invoked whenever the `viewport.ts` graphic provider changes or resets.

core/frontend/src/Viewport.ts
```diff
@@ -917,6 +919,60 @@ export abstract class Viewport implements Disposable, TileUser {
...
+    // Check if all model IDs in newModelIds exist in prevModelIds
+   for (let i = 0; i < prevModelIds.length; i++) {
+        if (prevModelIds[i] !== newModelIds[i]) {
+            return true;
+        }
+    }
+
+    return false;
+  }
+
```
danielzhong-bentley, line 975: These helper functions determine when to refresh the reality tile. Specifically, in our case, they are used when the map layer changes.

**Why is reality tile refresh logic handled differently from map tiles?**
Refer to the comments in `RealityMesh.ts` for details. The map tile's draping function automatically refreshes whenever the `graphic provider` resets. However, reality tiles do not have a `graphic provider`, so the draping logic function is only triggered when the `gltfReader` loads files.

core/frontend/src/Viewport.ts
```diff
@@ -1778,11 +1835,21 @@ export abstract class Viewport implements Disposable, TileUser {
...
+      this.target.reset();
+    }
 
     if (undefined !== prevView && prevView !== view) {
       this.onChangeView.raiseEvent(this, prevView);
       this._changeFlags.setViewState();
+
+      if (isMapLayerChanged) {
+        this.refreshRealityTile();
+      }
```
danielzhong-bentley, line 1852: This function is triggered when the save view feature is used. Instead of refreshing or resetting the reality tile every time the save view changes, we detect when the map layer changes. It then passes parameters into `reset()` in `target.ts` to determine whether to reset the correct planar classifier for draping textures. The goal is to make sure the classifier and other settings resets without affecting the draping map layer.  

For more details, refer to the comments in `target.ts`.

core/frontend/src/render/webgl/Target.ts
```diff
@@ -556,12 +556,26 @@ export abstract class Target extends RenderTarget implements RenderTargetDebugCo
...
+        // Keep "map layer" for reality tile not to refresh
+        const filteredClassifiers = new Map(
+            [...this._planarClassifiers.entries()]
+                .filter(([key]) => key.toLowerCase().includes("maplayer"))
+        );
+
+        // Update the planar classifiers to keep only the "maplayer" ones
+        this.changePlanarClassifiers(filteredClassifiers.size > 0 ? filteredClassifiers : undefined);
+    }
+
```
danielzhong-bentley, line 578: Instead of resetting all `planarClassifiers` when the save view changes, we aim to preserve the map layer's planar classifier textures and variables when they do not need to be refreshed. This ensures that if the scene remains unchanged and the reality tile does not require refreshing, unnecessary resets are avoided. Otherwise the texture will be empty/default for reality tile without refreshing.

pmconne:
How do reviewers know this does what it's supposed to?

markschlosseratbentley:
> How do reviewers know this does what it's supposed to?

@danielzhong-bentley updated the PR description with a screenshot.

To test it locally, you can also open display-test-app and:
- open an iModel
- attach a reality model located somewhere near the iModel
- utilize the keyin `fdt attach model maplayer reality` while an element is selected - the model associated with that element will drape onto the reality model.

pmconne:
> To test it locally, you can also open display-test-app and:

No unit tests nor image tests?

markschlosseratbentley:
> > To test it locally, you can also open display-test-app and:
> 
> No unit tests nor image tests?

We will add a test to Image Tests.  Added to the TODO in the description.

As far as unit tests, @danielzhong-bentley, can you take a look at adding some?

danielzhong-bentley:
> > To test it locally, you can also open display-test-app and:
> 
> No unit tests nor image tests?

I've just added a new unit test and pushed the changes. Please feel free to review it. Thanks!

pmconne:
Please fix the builds.

common/api/core-common.api.md
```diff
@@ -6025,10 +6025,17 @@ export interface ModelLoadProps {
     id?: Id64String;
 }
 
+// @public
```
pmconne, line 6028: Is `public` the appropriate release tag? Note that adding a new value to an `enum` is a breaking API change - do you expect to add new drape targets in the future (e.g., draping on iModel geometry)?

markschlosseratbentley: Yes, we expect new drape targets; thanks for the observation. Do you recommend @beta or @alpha?

pmconne: beta. Note that properties+functions inherit the release tags of their containing class/interface/namespace so you will need to add some @beta to those too.

markschlosseratbentley: upgraded to beta.

core/frontend/src/tile/LayerTileTree.ts
```diff
@@ -0,0 +1,65 @@
...
+/** Utility interface that ties an imagery tile tree to its corresponding map-layer settings object.
+ * @internal
+ */
+interface MapLayerTreeSetting {
+  tree: ImageryMapTileTree;
+  settings: MapLayerSettings;
+  baseImageryLayer: boolean;
+}
+
+export class LayerTileTree extends RealityTileTree {
```
pmconne, line 24: Missing release tag. It needs to be `@public`, because `MapTileTree` is an `@public` subclass.
You need to add a line to core-frontend.ts that re-exports symbols from this and any other new files containing non-internal APIs.
Then re-run extract-api and address any errors.

core/frontend/src/tile/LayerTileTreeReference.ts
```diff
@@ -0,0 +1,157 @@
...
+/** @packageDocumentation
+ * @module Tiles
+ */
+
+import { BaseLayerSettings, ColorDef, MapLayerSettings } from "@itwin/core-common";
+import { SceneContext } from "../ViewContext";
+import { createMapLayerTreeReference, ImageryMapLayerTreeReference, ImageryMapTileTree, LayerTileTree, MapLayerTileTreeReference, ModelMapLayerTileTreeReference, TileTreeLoadStatus, TileTreeReference } from "./internal";
+import { IModelConnection } from "../IModelConnection";
+
+export abstract class LayerTileTreeReference extends TileTreeReference {
```
pmconne, line 14: Same comment as for LayerTileTree.

core/common/src/MapLayerSettings.ts
```diff
@@ -190,13 +190,27 @@ export interface ImageMapLayerProps extends CommonMapLayerProps {
 
 }
 
+/** The target onto which to drape a model map layer.
+ * @beta
+ */
+export enum ModelMapLayerDrapeTarget {
+  /** Drape only onto the background map. */
+  BackgroundMap = 0,
+  /** Drape only onto all attached reality data. Supported reality data formats include glTF or b3dm. */
```
pmconne, line 199: Is "drape onto all reality models" the one and only use cases? Nobody wants to drape onto one or more specific reality models?

core/common/src/MapLayerSettings.ts
```diff
@@ -190,13 +190,27 @@ export interface ImageMapLayerProps extends CommonMapLayerProps {
 
 }
 
+/** The target onto which to drape a model map layer.
+ * @beta
+ */
+export enum ModelMapLayerDrapeTarget {
+  /** Drape only onto the background map. */
```
pmconne, line 197: Do you mean "ellipsoid" or "globe"? The "background map" is imagery; the ellipsoid is the geometry onto which the background map and/model map layers are draped.

core/common/src/MapLayerSettings.ts
```diff
@@ -190,13 +190,27 @@ export interface ImageMapLayerProps extends CommonMapLayerProps {
 
 }
 
+/** The target onto which to drape a model map layer.
```
pmconne, line 193: Is it not possible to drape a model onto both reality data and ellipsoid? Would you have to create two otherwise-identical model map layers with different drape targets?

core/common/src/MapLayerSettings.ts
markschlosseratbentley: It is possible users will want to drape onto specific reality models. For this initial implementation, we intended to keep it simple (the implementation already is quite involved), get something merged, and get more feedback, making incremental changes after this. (For example, draping onto iModels is next).

Perhaps we do need to consider the API so it is forward-facing in this respect.  Let me know if you have any preferences based on my below ideas.

### **API Right Now**

As it stands (in this PR), `ModelMapLayerProps` now has:
- `drapeTarget` of possible enum types `ModelMapLayerDrapeTarget.Globe` or `ModelMapLayerDrapeTarget.RealityData`

In a future PR, if necessary, `ModelMapLayerProps` could also include, alongside `drapeTarget`:
- `realityModelIds` array of reality models to drape onto if `drapeTarget` is `RealityData`.

### **Other API Considerations**

Alternatively, we could right now (in this PR), make `drapeTarget` become an interface type which contains the `ModelMapLayerDrapeTarget` enum as a property. In that interface, `realityModelIds` could reside in a future PR if necessary.

core/frontend/src/tile/LayerTileTree.ts
pmconne: Did you consider ways of doing this without an inheritance hierarchy?

danielzhong-bentley: > Did you consider ways of doing this without an inheritance hierarchy?

I think Option B is the way to go without relying on the inheritance hierarchy. It introduces a helper class to replace `LayerTileTree` and removes `LayerTileTree` using a utility module for the helper class. 

`LayerTileTreeReference` will remain for now since it doesn’t impact anything, but we can convert it into a helper class if needed. The `TileTreeReference` inheritance structure is much more straightforward.

core/frontend/src/tile/LayerTileTreeReference.ts
markschlosseratbentley: The PR has been refactored to not expand the tree hierarchy, so this comment is no longer relevant.

core/frontend/src/tile/LayerTileTree.ts
markschlosseratbentley: The PR has been refactored to not expand the tree hierarchy, so this comment is no longer relevant.

core/common/src/MapLayerSettings.ts
markschlosseratbentley: Since this development is occurring in phases, we plan to expand this functionality in a subsequent PR --- this current PR will be limited to draping onto all attached reality data.  The beta API can accommodate such expansion.  Once we add iModel draping in a subsequent PR, applying the draping only onto specific models will become even more important.

core/common/src/MapLayerSettings.ts
markschlosseratbentley: We have changed this enum type from "BackgroundMap" to "Globe".

pmconne:
> Supported reality data formats include glTF or b3dm

Are there unsupported formats? If so, which ones and why?

core/frontend-devtools/README.md
```diff
@@ -130,7 +130,7 @@ This package provides several keyins to control the display of background maps,
 
 * `fdt attach maplayer <name>` - Attach a background map layer from name within the map layer source list.  Partial names may be used.
 * `fdt attach mapoverlay <name>` - Attach an overlay map layer from name within the map layer source list.  Partial names may be used.
-* `fdt attach model maplayer <name>` - Attach a model map layer for each unique model of the currently selected elements.
+* `fdt attach model maplayer <globe|reality> [name]` - Attach a model map layer for each unique model of the currently selected elements to either the globe or attached reality models. If name is specified, use that.
```
pmconne, line 133: Key-ins are the primary API of this package, don't break them unnecessarily (e.g., make new arguments optional and append them).

core/frontend/src/Viewport.ts
```diff
@@ -927,6 +929,61 @@ export abstract class Viewport implements Disposable, TileUser {
...
+   * Compares the map layers of two view states, ensuring both the number of layers
+   * and their order remain unchanged.
+   * Returns true if the map layers differ in count, order, or model IDs; otherwise, returns false.
+   *
+   * @param prevView The previous view state.
+   * @param newView The new view state.
+   * @returns {boolean} True if there is any difference in the model layer configuration; false otherwise.
+   * @internal
+   */
+  public compareMapLayer(prevView: ViewState, newView: ViewState): boolean {
```
pmconne, line 952: No reason for this to be a `Viewport` method, make it a free function.

core/frontend/src/Viewport.ts
```diff
@@ -1788,11 +1846,21 @@ export abstract class Viewport implements Disposable, TileUser {
     this.updateChangeFlags(view);
     this.doSetupFromView(view);
     this.invalidateController();
-    this.target.reset();
+
+    const isMapLayerChanged = undefined !== prevView && this.compareMapLayer(prevView, view);
+    if (this.target.reset.length > 0) {
```
pmconne, line 1851: This looks gross, why is it needed?

core/frontend/src/common/internal/render/MeshParams.ts
```diff
@@ -22,4 +24,6 @@ export interface MeshParams {
   edges?: EdgeParams;
   isPlanar: boolean;
   auxChannels?: AuxChannelTable;
+  tile?: Tile;
```
pmconne, line 27: You're passing Tiles and textures all over the place, ultimately extracting a few bits of information (ecef transform, range) from them when creating the graphics.
Did you consider making a single interface that captures that subset of information and passing an optional instance of that around instead?
Tiles and RenderSystem are already too tightly coupled, leads to circular dependency errors at run-time and spaghetti code.

core/frontend/src/internal/render/webgl/ShaderBuilder.ts
```diff
@@ -905,6 +905,9 @@ export const enum FragmentShaderComponent {
   // (Optional) Apply planar classifier.
   // vec4 applyPlanarClassification(vec4)
   ApplyPlanarClassifier,
+  // (Optional) Apply planar classifier.
```
pmconne, line 908: Forgot to change the copy-pasted comment.

core/frontend/src/internal/render/webgl/SurfaceGeometry.ts
```diff
@@ -36,12 +40,73 @@ function wantLighting(vf: ViewFlags) {
...
-    return undefined !== indexBuffer ? new SurfaceGeometry(indexBuffer, indices.length, mesh) : undefined;
+
+    const tile = params.tile;
+
+    const layerClassifiers = tile?.tree?.layerHandler?.layerClassifiers;
+
+    if (!layerClassifiers?.size || !tile || undefined === layerClassifiers) return undefined !== indexBuffer ? new SurfaceGeometry(indexBuffer, indices.length, mesh, undefined) : undefined;
+
+    const transformECEF = tile.tree.iModel.getEcefTransform();
+    if (!transformECEF || !tile.range) return undefined !== indexBuffer ? new SurfaceGeometry(indexBuffer, indices.length, mesh, undefined) : undefined;
```
pmconne, line 59:
```suggestion
    if (!layerClassifiers?.size || !tile || undefined === layerClassifiers) {
      return undefined !== indexBuffer ? new SurfaceGeometry(indexBuffer, indices.length, mesh, undefined) : undefined;
    }

    const transformECEF = tile.tree.iModel.getEcefTransform();
    if (!transformECEF || !tile.range) {
      return undefined !== indexBuffer ? new SurfaceGeometry(indexBuffer, indices.length, mesh, undefined) : undefined;
    }
```
I hate nitpicking on whitespace (we should be using an automated code formatter) but please keep readability in mind.

core/frontend/src/internal/render/webgl/Target.ts
```diff
@@ -558,12 +558,26 @@ export abstract class Target extends RenderTarget implements RenderTargetDebugCo
...
     this.graphics[Symbol.dispose]();
     this._worldDecorations = dispose(this._worldDecorations);
     dispose(this.uniforms.thematic);
 
-    this.changePlanarClassifiers(undefined);
+    if (realityMapLayerChanged) {
+      // If reality map layer changed, dispose of all planar classifiers
+      this.changePlanarClassifiers(undefined);
+    } else if (this._planarClassifiers) {
+        // If reality map layer did NOT change, remove only non-"maplayer" classifiers (e.g., plannar classifier)
```
pmconne, line 570: Why? What happens if you remove all classifiers?

core/frontend/src/internal/tile/ImdlDecoder.ts
```diff
@@ -29,6 +29,7 @@ export interface ImdlDecodeArgs {
    * with [[GraphicBranch.groupNodeId]] set to the index of the group to which the model belongs.
    */
   modelGroups?: Id64Set[];
+  tile?: Tile;
```
pmconne, line 32: Passing around a Tile all over the place, I didn't find what you're using it for - can you explain?

markschlosseratbentley:
> > Supported reality data formats include glTF or b3dm
> 
> Are there unsupported formats? If so, which ones and why?

The remaining formats (CMPT and I3DM) are technically implemented but untested.  We eventually need to devise / get our hands on good datasets for these formats that are spatially aligned with some iModel geometry.  We did speak with @ggetz who gave us a link to some sample testcases but these will need to be matched up to some suitable iModels.

For now, we planned to explicitly advertise support for B3DM and glTF (the testcases we have).  We believe glTF will be the most common use case.

cc @danielzhong

pmconne:
> For now, we planned to explicitly advertise support for B3DM and glTF (the testcases we have). We believe glTF will be the most common use case.

I would just not mention formats and just refer to draping on reality models. i3dm isn't useful for reality models and cmpt is rare outside of tests.

core/frontend/src/internal/render/webgl/RealityMesh.ts
```diff
@@ -301,18 +163,60 @@ export class RealityMeshGeometry extends IndexedGeometry implements RenderGeomet
 
   public static createFromRealityMesh(realityMesh: RealityMeshParams, disableTextureDisposal = false): RealityMeshGeometry | undefined {
     const params = RealityMeshGeometryParams.fromRealityMesh(realityMesh);
-    if (!params)
-      return undefined;
-    const texture = realityMesh.texture ? new TerrainTexture(realityMesh.texture, realityMesh.featureID ?? 0, Vector2d.create(1.0, -1.0), Vector2d.create(0.0, 1.0), Range2d.createXYXY(0, 0, 1, 1), 0, 0) : undefined;
+    if (!params) return undefined;
+
+    const { texture: meshTexture, featureID } = realityMesh;
+    const tile = realityMesh.tile as RealityTile;
```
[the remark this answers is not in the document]
danielzhong-bentley: Indeed, fixed.

core/frontend/src/internal/render/webgl/ShaderBuilder.ts
danielzhong-bentley: Fixed

core/frontend/src/internal/render/webgl/Target.ts
danielzhong-bentley: I've updated the comment to clarify this. Removing all classifiers when the map layer hasn't changed would cause the draping effect to be lost without a refresh.

core/frontend/src/internal/tile/ImdlDecoder.ts
danielzhong-bentley: Just removed them—those were the codes for iModel draping that I forgot to remove. Later, it should pass a `LayerTileData` reference instead of the whole tile reference for iModel, similar to the changes I just made for reality tile draping.

core/frontend/src/common/internal/render/MeshParams.ts
danielzhong-bentley: Thanks for the suggestion! I just pushed the new version, where I created a `LayerTileData` interface to pass around instead of the entire tile reference.

core/frontend/src/internal/render/webgl/SurfaceGeometry.ts
```diff
@@ -49,15 +49,15 @@ export class SurfaceGeometry extends MeshGeometry {
...
 
     if (!layerClassifiers?.size || !tile || undefined === layerClassifiers) {
       return undefined !== indexBuffer ? new SurfaceGeometry(indexBuffer, indices.length, mesh, undefined) : undefined;
     }
 
-    const transformECEF = tile.tree.iModel.getEcefTransform();
+    const transformECEF = tile.ecefTransform;
     if (!transformECEF || !tile.range) {
       return undefined !== indexBuffer ? new SurfaceGeometry(indexBuffer, indices.length, mesh, undefined) : undefined;
     }
```
pmconne, line 63:
```suggestion
    const transformECEF = tile.ecefTransform;
    }
```
Neither of these can be `undefined`.

markschlosseratbentley:
> > For now, we planned to explicitly advertise support for B3DM and glTF (the testcases we have). We believe glTF will be the most common use case.
> 
> I would just not mention formats and just refer to draping on reality models. i3dm isn't useful for reality models and cmpt is rare outside of tests.

References to specific formats have been removed.

core/frontend-devtools/README.md
markschlosseratbentley: I changed the key-in as such, and pushed the change:

* `fdt attach model maplayer [name] [globe|reality]` - Attach a model map layer for each unique model of the currently selected elements to either the globe or attached reality models. If `globe` or `reality` is not specified, the attachment will default to `globe`. If `name` is specified, use that. If `name` is specified as `0`, that is considered `undefined` or unspecified.

core/frontend/src/ContextRealityModelState.ts
```diff
@@ -51,6 +53,9 @@ export class ContextRealityModelState extends ContextRealityModel {
         classifiers: this.classifiers,
         planarClipMask: this.planarClipMaskSettings,
         getDisplaySettings: () => this.displaySettings,
+        mapSettings,
```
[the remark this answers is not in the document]
markschlosseratbentley: I see your point. `mapSettings`, `backgroundBase`, and `backgroundLayers` are derived from the `DisplayStyleState` passed into the `ContextRealityModelState` constructor.  This `DisplayStyleState` already gets passed into the properties and is stored as `source` on the created `RealityTreeReference`.  From there, we could check for the latest `mapSettings`, `backgroundBase`, and `backgroundLayers`.

I do not believe we need a `getDisplayState()` method, since the state is passed in directly to this constructor. Instead, we can check these properties on the stored `source` in case the underlying properties on the display style change.  There is already precedence for this in `RealityTreeReference._wantWireMesh()`, for example, which does: 
```ts
return this._source.viewFlags.wiremesh;
```

I will attempt to change the code to use the passed-in display style to retrieve the map layer properties if no other concerns are raised.

core/frontend/src/internal/render/webgl/SurfaceGeometry.ts
danielzhong-bentley: Removed and pushed the code, because this commit suggestion was outdated.

core/frontend/src/Viewport.ts
danielzhong-bentley: Because during save view, we want to detect if it's the same view or if the map layer hasn't changed. If no change occurred, we don't want to refresh unnecessarily. Reality Tile Drape differs from map tiles because it doesn't have MapTiledGraphicsProvider.ts.

pmconne: My question was vague.
Why do you need to check `target.reset.length` and then perform a cast?

core/frontend/src/Viewport.ts
danielzhong-bentley: Fixed

core/frontend/src/Viewport.ts
danielzhong-bentley: Got it, just fixed!

core/frontend/src/Viewport.ts
```diff
@@ -927,6 +929,61 @@ export abstract class Viewport implements Disposable, TileUser {
     return true;
   }
 
+  /** Refresh the Reality Tile Tree to reflect changes in the map layer. */
+  private refreshLayerTiles(): void {
+    const { tiles } = this.iModel;
+    for (const { supplier, id, owner } of tiles) {
+      if (owner.tileTree instanceof RealityModelTileTree) {
+        tiles.resetTileTreeOwner(id, supplier);
```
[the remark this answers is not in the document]
pmconne: > I’d appreciate any suggestions on improving the approach for refreshing reality tiles or iModel tiles in the future.

Generally, a TileTreeReference computes the Id of the TileTree it should use every time it needs to access the tree, based on current state, which may change during the lifetime of the reference. Most TileTreeSuppliers have comparison routines that do things similar to your `compareMapLayer` function to determine if two tile tree Ids point to equivalent TileTrees. That's why I asked below about using a function to access the current map settings. Ideally the Id captures all relevant state so that no special "refresh"/"reset" logic is needed.

core/frontend/src/ContextRealityModelState.ts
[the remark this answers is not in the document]
danielzhong-bentley: Fixed

core/frontend/src/Viewport.ts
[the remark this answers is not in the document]
danielzhong-bentley: Thanks for the suggestion! I modified it to update the state as you recommended --- the tree ID now captures that state, and I removed the refresh function.

core/frontend/src/internal/tile/RealityModelTileTree.ts
```diff
@@ -665,12 +666,16 @@ export namespace RealityModelTileTree {
       }
       removals.push(context.viewport.onChangeView.addListener((vp, previousViewState) => {
         if(compareMapLayer(previousViewState, vp.view)){
-          const mapImagery = context.viewport.displayStyle.settings.mapImagery;
           this._layerRefHandler.setBaseLayerSettings(mapImagery.backgroundBase);
           this._layerRefHandler.setLayerSettings(mapImagery.backgroundLayers);
           this._layerRefHandler.clearLayers();
         }
       }));
+      removals.push(IModelApp.tileAdmin.onTileChildrenLoad.addListener(() => {
```
markschlosseratbentley, line 674: My understanding is that this solves https://github.com/iTwin/itwinjs-core/issues/7818 by ensuring that the tile tree which is being draped is fully loaded before the draping begins.

I believe this code you added will trigger when any children tile loads, not just the entire tree.

You said elsewhere that onTileTreeLoad does not work; what did you observe exactly?

pmconne (approved):

PR #7638, opened 2025-01-31
re enable oneAtATime test

### LINKED ISSUE #7200: Refactor core-bentley's OneAtATime test
Stemming from #7165, had to disable the test suite mentioned in the title. 

Need to revisit those tests and figure out how to solve them without using chaiAsPromised.

hl662:
Closes #7200 

Handling of vitest where:
1. We expect an error to be thrown
2. We expect those errors to be thrown as promise rejections (i.e run as non awaited promises) 

Can cause wonky behaviors, with vitest's runner identifying those errors to be 'uncaught'. We can listen to these uncaught exception events raised to suppress that behavior.

### DIFF
--- a/core/bentley/src/test/OneAtATime.test.ts
+++ b/core/bentley/src/test/OneAtATime.test.ts
@@ -2,7 +2,7 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, onTestFinished } from "vitest";
 import { AbandonedError, OneAtATimeAction } from "../OneAtATimeAction";
 import { BeDuration } from "../Time";
 
@@ -42,6 +42,8 @@ describe("OneAtATime test", () => {
     void expect(operation.request(200, "hello")).rejects.with.toEqual(abandonedError); // aborts previous, becomes pending
     count = await operation.request(200, "hello");
     expect(count).toBe(3);
-    process.removeListener("unhandledRejection", unhandledRejectionHandler);
+    onTestFinished(() => {
+      process.removeListener("unhandledRejection", unhandledRejectionHandler)
+    });
   });
 });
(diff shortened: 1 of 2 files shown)


core/bentley/src/test/OneAtATime.test.ts
```diff
@@ -2,51 +2,44 @@
...
-    expect(operation.request(200, "hello")).rejects.with.toEqual(abandonedError); // becomes pending, doesn't abort previous because its already started
-    expect(operation.request(200, "hello")).rejects.with.toEqual(abandonedError); // aborts previous, becomes pending
+    void expect(operation.request(10, "hello")).rejects; // try calling a function that throws
+    void expect(operation.request(10, "hello")).rejects.with.toEqual(cancelled); // try calling a function that throws
+    void expect(operation.request(200, "hello")).rejects.with.toEqual(abandonedError); // becomes pending, doesn't abort previous because its already started
+    void expect(operation.request(200, "hello")).rejects.with.toEqual(abandonedError); // aborts previous, becomes pending
     count = await operation.request(200, "hello");
-    assert.equal(count, 3);
+    expect(count).toBe(3);
+    process.removeListener("unhandledRejection", unhandledRejectionHandler);
```
ben-polinsky, line 43: What happens if the test exits early?

hl662: yeah, to ensure the test does not exit early, I added a [expect.assertions](https://vitest.dev/api/expect#expect-assertions) and if the test exits early, that assertion would fail

ben-polinsky: Will that ensure the listener is removed, though? 
It also requires us to track expectations should we add more... what about [onTestFinished](https://vitest.dev/api/#ontestfinished)?

hl662: that looks good, okay, I added that

ben-polinsky: Sorry, will it work at the bottom of the test? Should it be moved up?

hl662: good point, I've moved it up

ben-polinsky: thanks

ben-polinsky (approved):

Issue #7454, opened 2024-12-06
Transform applied to RenderInstances are incorrect when GraphicTemplate is created from a GLTF


mdastous-bentley:
**Describe the bug**
When creating instances using a template created from a GLTF, transforms appear to be applied based on the GLTF coordinate system instead of the iModelConnection's spatial coordinate system.

**To Reproduce**
1.  Get this [branch](https://github.com/iTwin/itwinjs-core/tree/micheld/gltfInstances)
2.  Run Display-Test-app
3. Click on the blank connection button in the tool bar.
4. Click on the smiling face icon in the tool bar.

**Expected behavior**
You should have 3 spheres in the view: red, blue, green.  Each sphere has it's center positioned  at points described by the structure returned by 
the `getCirclesData` function:
![Image](https://github.com/user-attachments/assets/f730daff-e633-41fe-9c5e-90bd2b7f2a80)

Each sphere should contains a model, creating from instancing a GLTF template:
![Image](https://github.com/user-attachments/assets/cb6f4719-cd48-4854-884d-f6fff9dfcf74)
![Image](https://github.com/user-attachments/assets/09164dcb-3407-4df0-a44f-8bf97a1f1fff)

You will notice, the green sphere has no gltf instance inside it, instead it got positioned incorrectly:
![Image](https://github.com/user-attachments/assets/024dc824-3996-426d-8354-65e345011229)

It should be placed at the center of the sphere, like the two others spheres.  The problem arise because the position of the green sphere has a delta value added on the Z-axis.  
Our assumption is that, the axis orientation is based on the GLTF coordinate system.

pmconne:
@markschlosseratbentley @danieliborra

andremig-bentley:
Here is an in progress PR addressing this issue -> https://github.com/iTwin/itwinjs-core/pull/7653

[...]

PR #7653, opened 2025-02-04
Add yAxisUp to ReadGltfGraphicsArgs

andremig-bentley:
This PR addresses this issue: https://github.com/iTwin/itwinjs-core/issues/7454

When calling `ReadGltf()`, `ReadGltfGraphics()`, or `ReadGltfTemplate()`, there is currently no option to specify the `yAxisUp` option of the `GltfGraphicsReader` when it is created as a part of these functions. Instead, `yAxisUp` is forced true. This leads to situations like the one outlined in the above issue, where a call to `ReadGltfTemplate()` may create misplaced graphics due to axis orientation issues. To solve this, an optional `yAxisUp` flag was added to `ReadGltfGraphicsArgs`, and can now be passed in to any of the above function calls if desired. If `ReadGltfGraphicsArgs.yAxisUp` is `undefined`, the flag defaults to true.

### DIFF
--- a/core/frontend/src/tile/GltfReader.ts
+++ b/core/frontend/src/tile/GltfReader.ts
@@ -2134,6 +2134,8 @@ export interface ReadGltfGraphicsArgs {
   hasChildren?: boolean;
   /** @internal */
   idMap?: BatchedTileIdMap;
+  /** Specify the axis orientation. If true, y will be up. */
+  yAxisUp?: boolean;
 }
 
 /** The output of [[readGltf]].
@@ -2178,7 +2180,7 @@ export async function readGltfGraphics(args: ReadGltfGraphicsArgs): Promise<Rend
  */
 export async function readGltfTemplate(args: ReadGltfGraphicsArgs): Promise<GltfTemplate | undefined> {
   const baseUrl = typeof args.baseUrl === "string" ? new URL(args.baseUrl) : args.baseUrl;
-  const props = GltfReaderProps.create(args.gltf, true, baseUrl); // glTF supports exactly one coordinate system with y axis up.
+  const props = GltfReaderProps.create(args.gltf, args.yAxisUp, baseUrl); // glTF supports exactly one coordinate system with y axis up.
   const reader = props ? new GltfGraphicsReader(props, args) : undefined;
   if (!reader)
     return undefined;
--- a/core/frontend/src/tile/GltfReader.ts
+++ b/core/frontend/src/tile/GltfReader.ts
@@ -2180,7 +2180,8 @@ export async function readGltfGraphics(args: ReadGltfGraphicsArgs): Promise<Rend
  */
 export async function readGltfTemplate(args: ReadGltfGraphicsArgs): Promise<GltfTemplate | undefined> {
   const baseUrl = typeof args.baseUrl === "string" ? new URL(args.baseUrl) : args.baseUrl;
-  const props = GltfReaderProps.create(args.gltf, args.yAxisUp, baseUrl); // glTF supports exactly one coordinate system with y axis up.
+  const yAxisUp = args.yAxisUp ?? true; // default to true
+  const props = GltfReaderProps.create(args.gltf, yAxisUp, baseUrl); // glTF supports exactly one coordinate system with y axis up.
   const reader = props ? new GltfGraphicsReader(props, args) : undefined;
   if (!reader)
     return undefined;
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -9041,6 +9041,7 @@ export interface ReadGltfGraphicsArgs {
     pickableOptions?: PickableGraphicOptions;
     // @alpha (undocumented)
     transform?: Transform;
+    yAxisUp?: boolean;
 }
 
 // @beta
--- a/common/changes/@itwin/core-frontend/andremig-render-instances_2025-02-04-17-04.json
+++ b/common/changes/@itwin/core-frontend/andremig-render-instances_2025-02-04-17-04.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "add yAxisUp to ReadGltfGraphicsArgs",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}


core/frontend/src/tile/GltfReader.ts
```diff
@@ -2178,7 +2180,8 @@ export async function readGltfGraphics(args: ReadGltfGraphicsArgs): Promise<Rend
  */
 export async function readGltfTemplate(args: ReadGltfGraphicsArgs): Promise<GltfTemplate | undefined> {
   const baseUrl = typeof args.baseUrl === "string" ? new URL(args.baseUrl) : args.baseUrl;
-  const props = GltfReaderProps.create(args.gltf, true, baseUrl); // glTF supports exactly one coordinate system with y axis up.
+  const yAxisUp = args.yAxisUp ?? true; // default to true
+  const props = GltfReaderProps.create(args.gltf, yAxisUp, baseUrl); // glTF supports exactly one coordinate system with y axis up.
```
pmconne, line 2184: Per the comment, [y is always up in glTF](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#coordinate-system-and-units). Presumably the problem is in our attempt to align it to our z-up coordinate system. Caller should not have to fix that for us.

pmconne (changes requested):

core/frontend/src/tile/GltfReader.ts
andremig-bentley: In the latest commit, this has been addressed. A new rotation has been added to the instance's translation which is handled in `createGraphicFromTeplate` in `System.ts`. It is handled there so that we can check whether or not the template from which the instance is created is a gltf template. This way, the caller does not have to specify any option to fix this for us.

andremig-bentley:
The addition of unit tests to `RenderInstances.test.ts` is WIP.

andremig-bentley:
Latest commit includes fixes to recreation of instances within `createGraphicsFromTemplate`, and the addition of a test in `RenderInstances.test.ts`.

core/frontend/src/internal/render/webgl/System.ts
```diff
@@ -966,4 +970,44 @@ export class System extends RenderSystem implements RenderSystemDebugControl, Re
...
+    // rotate transforms to z-up
+    for (let i = 0; i < count; i++) {
+      const instanceIdx = i * 12;
+      const instanceY = transforms[instanceIdx + 7];
+      const instanceZ = transforms[instanceIdx + 11];
+      transforms[instanceIdx + 7] = instanceZ;
+      transforms[instanceIdx + 11] = -instanceY;
+    }
+
+    // recreate instances
```
pmconne, line 988: Adding all these extra, redundant properties to RenderInstancesImpl so you can "recreate" the instances seems wrong. Can't you fix the transform earlier, and avoid having to recreate anything?

andremig-bentley: Unfortunately `instanceGltfModel()` is an example code snippet so I can't edit anything there to fix the issue. That being said, it does show the expected work flow of creating instances from a gltf template and it can show the issue I'm having here. Link to the function [here](https://github.com/iTwin/itwinjs-core/blob/2e08f6f7b670b5e6a333e8caab42bd2505b1e4ae/example-code/snippets/src/frontend/Instancing.ts#L11).

The expected work flow for creating these instances is essentially: 
1 - Obtain a Gltf template 
2 - Create `RenderInstances` (including their transforms, important to this PR)
3 - Call `createGraphicFromTemplate` passing in the template and the instances.

Now we generally want to avoid making the caller pass in any information to tell us when to rotate the transforms. If we want to avoid this, we need to know that what we are instancing is a gltf template. This means that the earliest possible point to know if we need to rotate the transforms is when we also have access to the gltf template, so in `createGraphicFromTemplate()` which is what is being done here. Unfortunately, this means that they need to be recreated since at this point the instances' transforms are held in a `BufferHandle` in a `readonly` property of `InstanceBuffersData`.

This leaves to me, three options: 
1 - Have the caller pass in a flag when creating the instances that tells us to rotate their transforms.
2 - Change `InstanceBuffersData.transforms` to be able to be written to, and then change the methods within `createGraphicFromTemplate()` to recreate that one `BufferHandle` and reassign it.
3 - Recreate the instances as we are doing here.

IMO, it makes the most sense to recreate the instances so that we can avoid relying on a caller to tell us to fix the issue, and so that we can avoid changing properties away from `readonly`. That being said, I think all three options could be valid solutions, and I'm open to any suggestions/discussion.

pmconne: 2 is not an option because the instances can be shared by multiple graphics - if you mutate them you'll affect other graphics that are also using them.

andremig-bentley: Do you think option 3, recreating the instances, will work then? It does seem to add some redundancies and it seems a little over-complicated but IMO it's still better for us to be able to rotate on our own without a caller needing to tell us to do it.

pmconne: Have you experimented with instancing a graphic template that has a different rotation+translation applied to it (not y-up-to-z-up)? Do the instances draw at the expected positions and orientations, or are you solving a specific instance of a general problem here? Did you check if we've got a reversed order of matrix multiplications somewhere?

andremig-bentley: I've been taking a deeper dive into this to answer these questions, and I think now that this solution might not be the best. 

1)  I think this is a general problem. Unless I am missing something, we simply do not rotate the translation portion of the RenderInstance's matrix, so any graphic template which requires a different rotation will also necessitate some form of translation rotation similar to this. glTF is an example of one where we know based on the template's type that it will need a specifically y-up-tp-z-up rotation, but in the scenario where there's a different rotation, I think we might need the caller to tell us what type of rotation is required (which is obviously not ideal).
2) I don't think there's a reversed order anywhere, instead I think we are just not rotating the translation at any point. I could be missing something, but I have looked everywhere that I can think of and found nothing.

A potential solution that might work better and would be agnostic of rotation would be to find the place where the rotation which is defined as a part of the `GltfReader` is actually applied to the instance's orientation, and apply the rotation to the instance's translation there as well. I need to do some more work to flesh this out, and I'm not sure exactly where this all happens. As of now, I'm looking into `BranchUniforms` because the rotation defined in the `GltfReader` gets added to a `GraphicBranch` as its passed down.

I'll update when I have more clear path forward, let me know if I need to re-explain anything here.

markschlosseratbentley: @andremig-bentley It sounds like you have investigated fairly thoroughly the matrix pipeline leading from the CPU to GPU. If there is nothing awry going on in that process, to me it seems likely okay to go with the currently implemented solution on this PR.

markschlosseratbentley (approved):
See my comment here: https://github.com/iTwin/itwinjs-core/pull/7653#discussion_r2033686863

andremig-bentley:
@markschlosseratbentley @pmconne 

Given the planned changes in our rendering architecture coming up, is this PR worth continuing to pursue? It seems to me that whatever issue we were having will be fixed with those changes, and this PR has now sat dormant for a while with no other complaints about the issue. Is it worth it to continue trying to push this through or should we just close and wait for the architectural changes?

markschlosseratbentley:
@andremig-bentley Priority, to me, is not high. I am okay deferring further work on this PR.

andremig-bentley:
Closing this PR as it's gone stale and the work will most likely not be necessary. I can reopen it if desired.

PR #7660, opened 2025-02-06
Presentation: Remove `3.x` deprecations from `presentation-common`

grigasp:
Closes https://github.com/iTwin/itwinjs-backlog/issues/1364.

Follow-up: https://github.com/iTwin/imodel-native/issues/992.


pmconne (approved):

full-stack-tests/presentation/.mocharc.json
```diff
@@ -0,0 +1,21 @@
+{
+  "require": [
+    "jsdom-global/register",
```
saskliutas, line 3:
```suggestion
    "global-jsdom/register",
```

presentation/frontend/.mocharc.json
```diff
@@ -1,6 +1,7 @@
 {
   "require": [
     "source-map-support/register",
+    "jsdom-global/register",
```
saskliutas, line 4: Should be switch to `global-jsdom` like in full-stack-tests?

grigasp: maybe in a separate pr?.. this is already way bigger than I'd like it to be

aruniverse (approved):
lgtm, do we need to update NextVer.md at all?

grigasp:
> lgtm, do we need to update NextVer.md at all?

good point, updated in https://github.com/iTwin/itwinjs-core/pull/7668

PR #7663, opened 2025-02-06
Improve error message in extract-api workflow to include local vs server version

rschili:
There is a common issue where local extract-api calls produce mismatching contents to what the server generates.
Tracking down those issues which are mostly ordering issues inside enums or generic parameters, is very tedious without the server giving feedback on the missing things.

This adds a "git diff" call which will print the diffs between the live and the server generated files.


DanRod1999 (approved):

ben-polinsky:
In my experience, this occurs when you haven't merged changes from master or run rush clean/rebuild before api-extract. You've found this happens other times?

rschili:
> In my experience, this occurs when you haven't merged changes from master or run `rush clean`/`rebuild` before `api-extract`. You've found this happens other times?

@ben-polinsky  Happens all the time to me. It's with the order of types/enum/generic values. Somehow my local extract-api sometimes arranges them in a different order than the file generated on the server. Then I have to guess and try/error to find the culprit. Affan said it's happening to him, too.

Note, I recently switched to linux for my development machine, and it happened before on windows, and still happens now. There may be more types prone to that in ecschema-metadata than in the other packages.

PR #7667, opened 2025-02-06
Remove top-level core-frontend rendering APIs from package exports

pmconne:
This just moves entire files or portions thereof from src/render/ to src/internal/render and adjusts imports accordingly.
Also moves a bunch of tests for disposing of rendering resources from core-full-stack-tests to core-frontend.


aruniverse (approved):

bbastings (approved):

markschlosseratbentley (approved):
LGTM

PR #7670, opened 2025-02-07
Revert BackendTypes to Public

MichaelSwigerAtBentley:
In https://github.com/iTwin/itwinjs-core/pull/7572#issuecomment-2642292035 BackendTypes referenced by `HttpServerRequest` were converted to internal. These types need to remain public to match `HttpServerRequest`.

Both the BackendTypes and `HttpServerRequest` function are deprecated, and should be removed by the upcoming RPC refactor.

### DIFF
--- a/core/common/src/internal/BackendTypes.ts
+++ b/core/common/src/internal/BackendTypes.ts
@@ -16,7 +16,7 @@ Do not add any new types to this file.
 All types here will be removed in 4.0
 */
 
-/** BackendReadable and BackendWritable are tagged public for depreciated public RPC APIs which reference these types. */
+/** BackendReadable and BackendWritable are tagged public for deprecated public RPC APIs which reference these types. */
 /** @public @deprecated in 3.x. This type was mistakenly made public in the common scope. */
 export type BackendReadable = Readable;
 
--- a/core/common/src/internal/BackendTypes.ts
+++ b/core/common/src/internal/BackendTypes.ts
@@ -16,10 +16,11 @@ Do not add any new types to this file.
 All types here will be removed in 4.0
 */
 
-/** @internal @deprecated in 3.x. This type was mistakenly made public in the common scope. */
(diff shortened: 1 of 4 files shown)


core/common/src/internal/BackendTypes.ts
```diff
@@ -16,10 +16,11 @@ Do not add any new types to this file.
 All types here will be removed in 4.0
 */
 
-/** @internal @deprecated in 3.x. This type was mistakenly made public in the common scope. */
+/** BackendReadable and BackendWritable are tagged public for depreciated public RPC APIs which reference these types. */
```
ben-polinsky, line 19:
```suggestion
/** BackendReadable and BackendWritable are tagged public for deprecated public RPC APIs which reference these types. */
```

core/common/src/internal/BackendTypes.ts
```diff
@@ -16,10 +16,11 @@ Do not add any new types to this file.
...
+/** @public @deprecated in 3.x. This type was mistakenly made public in the common scope. */
```
ben-polinsky, line 20: This comment is a bit confusing now.

aruniverse:
```suggestion
/** @deprecated in 3.x. This type was mistakenly made public in the common scope. */
```

MichaelSwigerAtBentley:
@aruniverse Removing the public tag also causes the rush extract-api to fail:

```Error: src/internal/BackendTypes.ts:21:1 - (ae-missing-release-tag) "BackendReadable" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)       
Error: src/internal/BackendTypes.ts:24:1 - (ae-missing-release-tag) "BackendWritable" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)       
```

aruniverse (approved):

PR #7671, opened 2025-02-07
Add limitations to quantity parsing logic

### LINKED ISSUE #7596: Default imperial units length parser does not accept string from default length formatter.
Using:

IModelApp.quantityFormatter.setActiveUnitSystem("imperial")).
IModelApp.quantityFormatter.findFormatterSpecByQuantityType(QuantityType.Length);
IModelApp.quantityFormatter.findParserSpecByQuantityType(QuantityType.Length);

Formatter can return a string like: 20'-6" and parser won't accept the "-". Must instead enter 20'6" or use a different separator like 20':6", these will be successfully formatted as 20'-6".

If possible fix should be backported to 4.11.x...

hl662:
Closes #7596 

- Improve tests, ensuring default parsing and formatting in core frontend's QuantityFormatter produces equivalent values between each other.
- Update the tokenization logic, adding limits and document them in scenarios where a spacer conflicts with a math operator. In particular:
  - Mathematical operations only apply when it is in between whitespace. So `-2FT 6IN + 6IN` is equal to `-2FT-6IN + 6IN`, and `-2FT-6IN - 6IN` is not equal to `-2FT-6IN-6IN`.
  - For a value like `2FT 6IN-0.5`, if the FormatProps has the spacer set to `-`, the `-` sign will be treated as a spacer and not subtraction. However, the `0.5` value will use the default unit conversion provided to the parser, because it's not a part of the composite unit when that composite is made up of only 2 units - `FT` and `IN`.


bbastings:
I only approve of the test changes to make sure the parser/formatter match. Don't know enough about the parsing code and potential separators to say whether this is a good solution. Not supporting * and / makes this a big who cares for me.

hl662:
> I only approve of the test changes to make sure the parser/formatter match. Don't know enough about the parsing code and potential separators to say whether this is a good solution. Not supporting * and / makes this a big who cares for me.

I think the best solution, is to take out math operations like you say and figure out a way to add math support **on top** of quantity, not bake it in, something similar to what you did with AccuDraw. But that shouldn't be done as part of this PR - we want to slot in an alternative the same time we drop this support. We can open an issue to address exactly this later on

 Immediately? I want to address the bug as is and the lack of tests for these cases.

aruniverse:
fyi @mathieu-fournier

core/frontend/src/test/QuantityFormatter.test.ts
```diff
@@ -79,6 +79,20 @@ describe("Quantity formatter", async () => {
...
+  it("Length default parser should handle format", async () => {
+    const numericVal = 6.2484; // 20'-6" in meters
+
+    await quantityFormatter.setActiveUnitSystem("imperial");
+    expect(quantityFormatter.activeUnitSystem).toBe("imperial");
+    const imperialParserSpec = await quantityFormatter.getParserSpecByQuantityType(QuantityType.Length);
+    const imperialFormatSpec = await quantityFormatter.getFormatterSpecByQuantityType(QuantityType.Length);
+    const stringVal = quantityFormatter.formatQuantity(numericVal, imperialFormatSpec);
+    expect(stringVal).toBe(`20'-6"`);
+    const parsedVal = quantityFormatter.parseToQuantityValue(`20'-6"`, imperialParserSpec);
```
tcobbs-bentley, line 91: While this works, I feel that it would be better to use `stringVal` instead of the hard-coded string. The previous line already makes sure that it has the expected value, and passing `stringValue` directly reiterates that the parser can correctly parse the output from the formatter.

```suggestion
    const parsedVal = quantityFormatter.parseToQuantityValue(stringVal, imperialParserSpec);
```

core/quantity/src/Parser.ts
```diff
@@ -333,22 +333,33 @@ export class Parser {
           }
         } else {
           // not processing a number
-          if (isOperator(charCode)) {
+          const isCharOperator = isOperator(charCode);
+          const isSpacer = charCode === format.spacerOrDefault.charCodeAt(0) && charCode !== QuantityConstants.CHAR_SPACE;
```
tcobbs-bentley, line 337: Just to check, are you intentionally disallowing space even if `spacer` is set to space? Because otherwise, I think the following would be better:

```suggestion
          const isSpacer = charCode === format.spacer.charCodeAt(0);
```

Note: `spacerOrDefault` returns `spacer` if it has a value, or `" "` if it is undefined. But if `spacer` is **set** to space, it seems like it should be allowed.

core/quantity/src/Parser.ts
```diff
@@ -333,22 +333,33 @@ export class Parser {
...
+          if (isSpacer && i > 0 && i < str.length - 1) {
+            const prevCharCode = str.charCodeAt(i - 1);
+            const nextCharCode = str.charCodeAt(i + 1);
+            if (isCharOperator && nextCharCode !== QuantityConstants.CHAR_SPACE && prevCharCode !== QuantityConstants.CHAR_SPACE) {
+              // ignore spacer if it's not at the start or end, not whitespace, and is not in between whitespace
+              continue;
+            }
+          }
+
+          if(wipToken.length === 0 && charCode === QuantityConstants.CHAR_SPACE){
```
tcobbs-bentley, line 348:
```suggestion
          if (wipToken.length === 0 && charCode === QuantityConstants.CHAR_SPACE) {
```

tcobbs-bentley: I just noticed that this file goes back and forth between whether or not it has space around the parentheses for `if` statements. I _believe_ that spaces are supposed to be there for all if/for/while, etc. statements, but I'm not 100% sure of that.

core/quantity/src/test/Parsing.test.ts
```diff
@@ -138,6 +138,52 @@ describe("Parsing tests:", () => {
...
+      uomSeparator: "",
+      allowMathematicOperations: false,
+    };
+    const format = new Format("test");
+    const unitsProvider = new TestUnitsProvider();
+    await format.fromJSON(unitsProvider, formatData).catch(() => { });
+
+    const tests = [
+      {input: "12'-6 1/2\"", expectedTokens: [{ value: 12 }, { value: "'" }, { value: 6.5 }, { value: '"' }]},
+      {input: "-2FT-6IN + 6IN", expectedTokens: [{value: "-", isOperand: true}, { value: 2 }, { value: "FT" }, { value: 6 }, { value: "IN" }, { value: "+", isOperand: true }, { value: 6 }, { value: "IN" }]},
```
tcobbs-bentley, line 175:
```suggestion
      { input: "12'-6 1/2\"", expectedTokens: [{ value: 12 }, { value: "'" }, { value: 6.5 }, { value: '"' }]},
      { input: "-2FT-6IN + 6IN", expectedTokens: [{ value: "-", isOperand: true }, { value: 2 }, { value: "FT" }, { value: 6 }, { value: "IN" }, { value: "+", isOperand: true }, { value: 6 }, { value: "IN" }]},
```

docs/learning/frontend/QuantityFormatting.md
```diff
@@ -336,6 +336,10 @@ parseResult.value // 7.5 (feet)
 #### Limitations
 Only plus(`+`) and minus(`-`) signs are supported for now.
 Other operators will end up returning a parsing error or an invalid input result.
+If a Format uses a spacer that conflicts with one of the operators above, additional restrictions will apply:
+
+- Mathematical operations only apply when it is in between whitespace. So `-2FT 6IN + 6IN` is equal to `-2FT-6IN + 6IN`, and `-2FT-6IN - 6IN` is not equal to `-2FT-6IN-6IN`.
```
tcobbs-bentley, line 341:
```suggestion
- Mathematical operations only apply when they are in between whitespace. So `-2FT 6IN + 6IN` is equal to `-2FT-6IN + 6IN`, and `-2FT-6IN - 6IN` is not equal to `-2FT-6IN-6IN`.
```

docs/learning/frontend/QuantityFormatting.md
```diff
@@ -336,6 +336,10 @@ parseResult.value // 7.5 (feet)
...
+- For a value like `2FT 6IN-0.5`, the `-` sign will be treated as a spacer and not subtraction. However, the `0.5` value will use the default unit conversion provided to the parser, because it's not a part of the composite unit when that composite is made up of only 2 units - `FT` and `IN`.
```
tcobbs-bentley, line 342: I feel that this is confusing. Can you provide examples of what the final result would be with different possible parser settings?

core/quantity/src/Parser.ts
tcobbs-bentley: Never mind. Even if `spacer` is set to space, you don't want `isSpacer` to be true.

core/quantity/src/Parser.ts
pmconne: We should be automating enforcement of any style rules that we care about (I advocate using [prettier](https://prettier.io/). Nitpicking over them on PRs is a waste of valuable developer time.

core/quantity/src/test/Parsing.test.ts
pmconne: Please, don't nitpick whitespace. File an issue to address lack of automated code formatting.

docs/learning/frontend/QuantityFormatting.md
hl662: I've added examples to each of the bullet points, encased in a markdown expandable block. I haven't seen the expandable blocks used in iTwinjs.org before, so @ben-polinsky lmk if it doesn't look like it'll work

docs/learning/frontend/QuantityFormatting.md
hl662: updated

core/frontend/src/test/QuantityFormatter.test.ts
hl662: gotcha, updated

core/frontend/src/test/QuantityFormatter.test.ts
```diff
@@ -535,9 +549,12 @@ describe("Test Formatted Quantities", async () => {
 
   async function testFormatting(type: QuantityTypeArg, magnitude: number, expectedValue: string) {
     const formatterSpec = await quantityFormatter.getFormatterSpecByQuantityType(type);
+    const parserSpec = await quantityFormatter.getParserSpecByQuantityType(type);
+
     const formattedValue = quantityFormatter.formatQuantity(magnitude, formatterSpec);
-    // console.log(`Type=${type} formatted value=${formattedValue}`); // eslint-disable-line no-console
+    const parsedValue = quantityFormatter.parseToQuantityValue(expectedValue, parserSpec);
     expect(formattedValue).toBe(expectedValue);
+    expect(withinTolerance((parsedValue as ParsedQuantity).value, magnitude, 0.01)).toBe(true);
```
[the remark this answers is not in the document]
hl662: i added a `expect(parsedValue.ok).toBe(true)` but it won't auto cast the interface to `ParsedQuantity` so I left the type cast here.

docs/learning/frontend/QuantityFormatting.md
ben-polinsky: See example [here](https://www.itwinjs.org/presentation/advanced/localization/#on-the-backend)

hl662: i checked the code for that .md file, the syntax is the same as my use here, i think we're good to go

bbastings:
@hl662 So what's up with this PR, has any decision been made? Also, ideally I'd like a fix backported to 4.11.x so I can remove my hack to make AccuDraw work...

  private makeParserHappy(value: string, isAngle: boolean): string {
    // TODO: Work around for default length parser not accepting output formatted with dash separator, ex. 20'-6"...
    const parserSpec = (isAngle ? undefined : this.getLengthParser());
    if (undefined === parserSpec)
      return value;
    return (FormatType.Fractional === parserSpec.format.type && -1 !== value.indexOf("'-") ? value.replaceAll("'-", "':") : value);
  }

core/quantity/src/Parser.ts
```diff
@@ -333,22 +333,33 @@ export class Parser {
...
-          if (isOperator(charCode)) {
+          const isCharOperator = isOperator(charCode);
+          const isSpacer = charCode === format.spacerOrDefault.charCodeAt(0) && charCode !== QuantityConstants.CHAR_SPACE;
+
+          if (isSpacer && i > 0 && i < str.length - 1) {
+            const prevCharCode = str.charCodeAt(i - 1);
+            const nextCharCode = str.charCodeAt(i + 1);
+            if (isCharOperator && nextCharCode !== QuantityConstants.CHAR_SPACE && prevCharCode !== QuantityConstants.CHAR_SPACE) {
+              // ignore spacer if it's not at the start or end, not whitespace, and is not in between whitespace
+              continue;
```
ColinKerr, line 344: I don't think we want to force a white space on either side of the operator.

If we are going to rely on spacing maybe a space before the operator is enough.

Another idea that is more complex is that we know how many segments there could be so we know the max number of separators ... if we see more it must be an operator.  If we did this we would still need to do something with spaces to handle cases where a value didn't need all segments.  e.g. exactly 12'.

I'd be happy to start simple and say there must be a space before if the spacer character matches the operator.

hl662: alrighty, i've edited the rule so just having a space before the operator is enough. no regression in our tests, I also added a few test cases where following the operator is a non white space character too.

ColinKerr (approved):

bbastings (approved):

hl662:

PR #7693, opened 2025-02-11
Adding api-documenter in sync with api-extractor to emit markdown documents with custom tags

pankhur94:
- Creating draft pr as API extractor currently supports custom tags only using tsdoc.json file. 
- Although, it allows us to declare custom tags, it doesn't allow us to add behavior to the custom tags. However, rush team are [working](https://github.com/microsoft/rushstack/issues/4954) on a possible plugin that can be applied to api-documenter to define custom tags behavior. 
- This way api-extractor would accept the tags(tsdoc.json) and then api-documenter plugin would implement any special processing.
- I have introduced api-documenter to frontend for now, this would help us to keep this documented and continue further when api-documenter has the plugin enabled.


ben-polinsky:
Do you have an example of the output?

tools/build/package.json
```diff
@@ -43,7 +43,8 @@
     "typedoc-plugin-merge-modules": "^6.0.3",
     "typescript": "~5.6.2",
     "wtfnode": "^0.9.1",
-    "yargs": "^17.4.0"
+    "yargs": "^17.4.0",
+    "@microsoft/api-documenter": "^7.26.7"
```
ben-polinsky, line 47: Please move to the top next to `@microsoft/api-extractor`

pankhur94:
`@microsoft/api-extractor` introduced a [patch](https://github.com/microsoft/rushstack/blob/main/apps/api-extractor/CHANGELOG.md#7523) to add support for custom TSDoc tags to appear in API reports. This helped us to include our custom `@preview` tag to our API reports which was merged [here](https://github.com/iTwin/itwinjs-core/pull/7979).

PR #7699, opened 2025-02-12
 Clean cache on every test run to make sure linked dependencies are up to date

### LINKED ISSUE #7662: Clear out vite's cache across vitest-enabled packages
When working across different packages within the monorepo, someone might want to change an API in one package, and consume that API in another package. They want to write a test for this integration, but they find issues where the changed API doesn't have their updated logic included. 

We use [dependency pre-bundling](https://vite.dev/guide/dep-pre-bundling#file-system-cache), set in our vitest configs, for linked packages in our monorepo . This tells vite to pre-bundle those dependencies, and to not go through `node_modules` to find those dependencies, and then those pre-bundled assets would be stored in a cache. The issue is, the standard [conditions](https://vite.dev/guide/dep-pre-bundling#file-system-cache) the cache would be updated don't apply to rush monorepos like ours, so issues like in the first paragraph would arise. We want to [add a config](https://vite.dev/config/dep-optimization-options.html#optimizedeps-force) to all our vitest configs to clean the cache out on every run, just to make sure linked dependencies are up to date. when working and testing across packages on our monorepo.

pachhaibipin:
**PR Description**
* closes #7662

**Fixes:**
* Added ` optimizeDeps: { force: true, }` to vitest config so as to force dependency pre-bundling, ignoring previously cached optimized dependencies.
* By default, linked packages not inside node_modules are not pre-bundled. Used [optimizeDeps.include](https://vite.dev/config/dep-optimization-options.html#optimizedeps-force) option on vitest config to enforce pre-bundling of peer Dependencies.


aruniverse:
It's not easily apparent how your changes you made resolves the issue. Can you add links to your PR description to the vitest config and explain how it works/ how it's fixed now

anmolshres98:
From my experience, this configuration breaks debugging sometimes. Could you check to verify this change doesn't affect debugging. Thanks

MichaelSwigerAtBentley:
> From my experience, this configuration breaks debugging sometimes. Could you check to verify this change doesn't affect debugging. Thanks

Curious about this, in what way does it break?

hl662 (approved):
@anmolshres98 's issue with debugging is not related to the issue this PR is tackling. They were more pointing out how HMR/watch mode for vite doesn't watch for linked workspace deps in apps, like DTA and DPTA. I've created a separate [issue](https://github.com/iTwin/itwinjs-core/issues/7773) for that, as an enhancement anyone can tackle, but it should not block this PR from going in, which we were able to reproduce the problem and verify the problem disappeared with this PR's change via @MichaelSwigerAtBentley

dassaf4 (approved):

MichaelSwigerAtBentley (approved):

saeeedtorabi (approved):

core/geometry/vitest.config.mts
```diff
@@ -41,5 +41,8 @@ export default defineConfig({
         },
       },
     },
+  },
+  optimizeDeps: {
```
aruniverse, line 45: @hl662 any reason we dont have an include array here listing core-bentley?
all the other ones seem include peer deps, but this pkg doesnt have peers and has a direct dep on core-bentley and should that be listed?

aruniverse (approved):

core/geometry/vitest.config.mts
aruniverse: ok, if @pachhaibipin is going to add this, lets do it programmitically not defining it directly. obtain the itwinjs core workspace pkgs defined in either the direct or peer deps from the package.json

markschlosseratbentley (approved):

core/geometry/vitest.config.mts
pachhaibipin: > ok, if @pachhaibipin is going to add this, lets do it programmatically not defining it directly. obtain the itwinjs core workspace pkgs defined in either the direct or peer deps from the package.json

just updated to include programmatically and added `rush change`.

core/common/vitest.config.mts
```diff
@@ -1,4 +1,16 @@
 import { coverageConfigDefaults, defineConfig } from 'vitest/config';
+import * as packageJson from "./package.json";
+
+const includePackages: string[] = [];
+
+Object.keys(packageJson.peerDependencies).forEach((pkgName) => {
+  if (pkgName.startsWith("@itwin") || pkgName.startsWith("@bentley")) {
```
hl662, line 7: i wouldn't use the name of the package... instead use the values of the package. We want any dep that has its value to be `workspace:*` to be in the includedPackages. There may be other @itwin or @bentley packages we use that aren't coming from within iTwin core

common/changes/@itwin/core-bentley/bp-clear-vite-cache_2025-02-27-18-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-bentley",
+      "comment": "Clean cache on every test run to make sure linked dependencies are up to date",
```
aruniverse, line 5: these change files are supposed to be user facing, there were no changes in our api that we need to call out here

```suggestion
      "comment": "",
```

common/changes/@itwin/core-common/bp-clear-vite-cache_2025-02-27-18-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
+      "comment": "Clean cache on every test run to make sure linked dependencies are up to date",
```
aruniverse, line 5:
```suggestion
      "comment": "",
```

common/changes/@itwin/core-frontend/bp-clear-vite-cache_2025-02-27-18-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Clean cache on every test run to make sure linked dependencies are up to date",
```
aruniverse, line 5:
```suggestion
      "comment": "",
```

common/changes/@itwin/core-geometry/bp-clear-vite-cache_2025-02-27-18-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-geometry",
+      "comment": "Clean cache on every test run to make sure linked dependencies are up to date",
```
aruniverse, line 5:
```suggestion
      "comment": "",
```

common/changes/@itwin/core-quantity/bp-clear-vite-cache_2025-02-27-18-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-quantity",
+      "comment": "Clean cache on every test run to make sure linked dependencies are up to date",
```
aruniverse, line 5:
```suggestion
      "comment": "",
```

core/common/vitest.config.mts
```diff
@@ -1,4 +1,16 @@
 import { coverageConfigDefaults, defineConfig } from 'vitest/config';
+import * as packageJson from "./package.json";
+
+const includePackages: string[] = [];
+
+Object.entries(packageJson.peerDependencies).forEach(([pkgName, version]) => {
+  if (version === "workspace:*") {
+    try {
```
aruniverse, line 8: any reason to wrap this with a try/catch?

core/common/vitest.config.mts
```diff
@@ -1,4 +1,16 @@
 import { coverageConfigDefaults, defineConfig } from 'vitest/config';
+import * as packageJson from "./package.json";
+
+const includePackages: string[] = [];
+
+Object.entries(packageJson.peerDependencies).forEach(([pkgName, version]) => {
```
aruniverse, line 6: instead of forEach you can clean this code up and use `.filter`

core/common/vitest.config.mts
pachhaibipin: > any reason to wrap this with a try/catch? this applies to all the files

nothing in particular. I was just taking https://github.com/iTwin/itwinjs-core/blob/f2c191a171fc2063ca48977bbed1b3f6acbcf3a1/test-apps/display-test-app/vite.config.mts#L22 as reference thinking I won't go out of practice. I will just drop it then?

PR #7700, opened 2025-02-12
Make BackendHubAccess interface public, IModelHost.hubAccess private

tcobbs-bentley:
Fix for https://github.com/iTwin/itwinjs-backlog/issues/1276

This marks `IModelHostOptions.hubAccess` as public, along with its type (`BackendHubAccess`). It also uses Symbol instance replacements for `hubAccess`, `setHubAccess`, and `getHubAccess` on `IModelHost` to make them internal.


core/backend/src/BackendHubAccess.ts
```diff
@@ -188,18 +188,27 @@ export interface CreateNewIModelProps extends IModelNameArg {
...
  * @note This interface is implemented in another repository. Any changes made to this interface must be validated against
  * the implementation found here: https://github.com/iTwin/imodels-clients/blob/main/itwin-platform-access/imodels-access-backend/src/BackendIModelsAccess.ts
- * @internal
+ * @public
  */
 export interface BackendHubAccess {
-  /** Download all the changesets in the specified range. */
+  /**
+   * Download all the changesets in the specified range.
+   * @internal
```
tcobbs-bentley, line 196: @pmconne Should we make `ChangesetFileProps` public instead of being forced to mark the three functions that use it internal?

core/backend/src/BackendHubAccess.ts
```diff
@@ -188,18 +188,27 @@ export interface CreateNewIModelProps extends IModelNameArg {
  * Generally direct access to these methods should not be required, since higher-level apis are provided.
  * @note This interface is implemented in another repository. Any changes made to this interface must be validated against
  * the implementation found here: https://github.com/iTwin/imodels-clients/blob/main/itwin-platform-access/imodels-access-backend/src/BackendIModelsAccess.ts
- * @internal
+ * @public
```
tcobbs-bentley, line 191: @pmconne I think public is good here, but it's possible that beta is more appropriate.

core/backend/src/internal/cross-package.ts
```diff
@@ -7,4 +7,5 @@ export { IModelJsNative, NativeCloudSqlite, NativeLoggerCategory } from "@bentle
 export { IModelNative } from "./NativePlatform";
 export {
   _nativeDb,
+  _hubAccess,
```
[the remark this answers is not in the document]
pmconne: cross-package.ts files only exist because some packages still consume both ESModules and CommonJS modules. core-full-stack-tests only consumes CommonJS, so it can consume internal APIs directly via `@itwin/package-name/lib/cjs/path/to/file.ts`. [Example](https://github.com/iTwin/itwinjs-core/blob/cd383b3d7962fbc8b8c04464cc28a660e03b7d9a/full-stack-tests/core/src/frontend/standalone/tile/TileIO.test.ts#L14).
I didn't check, but if example-code-snippets also uses only ESM or only CJS, it can do the same thing. Otherwise, exporting from cross-package.ts is appropriate.

core/backend/src/BackendHubAccess.ts
pmconne: Yes. A package outside of itwinjs-core implements BackendHubAccess, so all of its APIs must be @public and therefore subject to our API support policies.

core/backend/src/BackendHubAccess.ts
```diff
@@ -188,18 +188,27 @@ export interface CreateNewIModelProps extends IModelNameArg {
  * Generally direct access to these methods should not be required, since higher-level apis are provided.
  * @note This interface is implemented in another repository. Any changes made to this interface must be validated against
  * the implementation found here: https://github.com/iTwin/imodels-clients/blob/main/itwin-platform-access/imodels-access-backend/src/BackendIModelsAccess.ts
```
pmconne, line 190:
```suggestion
```

Remove the note. The idea that we could freely break this API at any time was misguided, because the implementation lives outside of repo defining the interface and follows different versioning cadence (and we can't control what versions of each get combined in a given app).

core/backend/src/BackendHubAccess.ts
tcobbs-bentley: OK, I'll make things public.

tcobbs-bentley: I assume that means that the other APIs in `BackendHubAccess` that were already explicitly marked as internal also need to be public now?

pmconne: Yes, except the ones marked as deprecated in 3.x - those can be deleted.

hl662 (approved):
LGTM, I'll defer to @pmconne on whether these APIs should be public immediately or beta first.

core/backend/src/BackendHubAccess.ts
tcobbs-bentley: OK, will do.

tcobbs-bentley: Because it is so involved, I feel that deleting `downloadV1Checkpoint` should be done in a separate PR.

core/backend/src/internal/cross-package.ts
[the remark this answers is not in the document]
tcobbs-bentley: The cjs import worked in `example-code-snippets`, so I removed this export.

core/backend/src/BackendHubAccess.ts
tcobbs-bentley: Everything except `downloadV1Checkpoint` is now public. That can be deleted in a separate PR.

core/backend/src/BackendHubAccess.ts
tcobbs-bentley: Based on replies in the other comment, I'm going to mark this one as resolved: public is appropriate.

core/backend/src/BackendHubAccess.ts
```diff
@@ -62,12 +62,15 @@ export interface V2CheckpointAccessProps {
   readonly storageType: string;
 }
 
-/** @internal */
+/**
+ * A Map from `Id64String` to `CommonLockState`.
```
pmconne, line 66:
```suggestion
 * Maps element Ids to their corresponding [LockState]($common)s.
```

example-code/snippets/src/backend/IModelBranchingOperations.test.ts
```diff
@@ -31,7 +32,7 @@ async function initializeBranch(myITwinId: string, masterIModelId: string, myAcc
   const masterDb = await BriefcaseDb.open({ fileName: masterDbProps.fileName });
 
   // create a duplicate of master as a good starting point for our branch
-  const branchIModelId = await IModelHost.hubAccess.createNewIModel({
+  const branchIModelId = await IModelHost[_hubAccess].createNewIModel({
```
pmconne, line 35: We can't use internal APIs like `IModelHost[_hubAccess]` in the published parts of example code snippets. It appears core-backend does not expose a public API to create a new iModel.

tcobbs-bentley: Should I add createNewIModel as a new public static function on IModelHost?

tcobbs-bentley: I added `createNewIModel` as a public static function on `IModelHost`.

pmconne (approved):

PR #7706, opened 2025-02-14
TextBlock margins

### LINKED ISSUE #7675: TextAnnotations - TextBlocks should have a property for margins

I would like to be able to specify margins around a `TextAnnotation` similar to how we describe the `width`. 

`TextBlock` should have a property called `margins` with the following attributes
* Either: `top`, `bottom`, `left`, `right`
* Or: `vertical`, `horizontal`
* Or both
* Also units should be meters

claudiaareneee:
**Changes**

* Creates a new interface called `TextBlockMargins`
* Adds a new attribute called `margins` to `TextBlock` of the type above
* Draws additional lines to the `debugAnchorPoint` geometry to show the margins
* Adds a `marginRange` to the `TextBlockLayout` class that expands the original `range` to include the specified margins.
* Does this in a method called `applyMargins`
* Adds a test to cover changes
* Adds keyins to set the margin in `display-test-app`
  * `dta text margin <'left', 'right', 'top', 'bottom'> <number>`

**Screenshots**

This was taken in `display-test-app`

keyins

**Associated issue** 
* Closes #7675

## Edit: 

After all the changes, this is what it looks like now:

https://github.com/user-attachments/assets/252cb7b3-3ce6-41d8-850f-a4f36485bc64


core/common/src/annotation/TextBlock.ts
```diff
@@ -422,6 +436,8 @@ export interface TextBlockProps extends TextBlockComponentProps {
   width?: number;
   /** The alignment of the document content. Default: "left". */
   justification?: TextBlockJustification;
+  /** The margins to surround the document content. Default: 0 margins on each sides */
```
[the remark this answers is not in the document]
tcobbs-bentley: This could also be "0 margin on each side".

tcobbs-bentley (approved):

pmconne:
Two questions.

1. How do the margins affect the bounding box of a text annotation element?

2. What happens when the user provides a negative margin value?

claudiaareneee:
@pmconne, to answer question 1, It should take the margins into account and be included in the bounding box. Right now, because I haven't made changes to it, I expect it to match the existing range.

I wrote a test to see what happens. I had some unexpected results. Not only does the bbox of the geometric element not match my new range, it doesn't match the original range. I wondered if there was some transformation or different geometry space that were being calculated. 

In my test, I console log these values. This is what I get. 

Note, `el1`, the element created,  xLength is 5.53271484375 when I expect it to be 7. yLength is 2 when I expect it to be 1. 

```tsx

TextAnnotation element
    persistence
Error | ECDb | ECClass 'bis.ChannelRootAspect' does not exist or could not be loaded. 
el0 bbox : 
        JSON: {"low":[1e+200,1e+200,1e+200],"high":[-1e+200,-1e+200,-1e+200]} 
        xLength: 0 
        yLength: 0 


el1 bbox : 
        JSON: {"low":[-2.766357421875,-1,0],"high":[2.766357421875,1,0]} 
        xLength: 5.53271484375 
        yLength: 2 


newLayout.range : 
        JSON: [[0,-1],[7,0]] 
        xLength: 7 
        yLength: 1 


transformed newLayout.range : 
        JSON: {"low":[-3.5,-0.5,0],"high":[3.5,0.5,0]} 
        xLength: 7 
        yLength: 1 
```

core/backend/src/test/annotations/TextAnnotation.test.ts
```diff
@@ -1241,5 +1283,77 @@ describe("TextAnnotation element", () => {
...
+        // Before my changes, this test should pass. After my changes, it should fail.
+        const range3d = Range3d.fromJSON(newLayout.range);
+        expect(el1.placement.bbox.isAlmostEqual(range3d)).to.be.true;
+
+        // After my changes, this should pass instead.
+        // const range3d = Range3d.fromJSON(newLayout.marginRange);
+        // expect(el1.placement.bbox.isAlmostEqual(range3d)).to.be.true;
+      };
+
+      // 0 margins
```
claudiaareneee, line 1351: Broken test I mentioned.

pmconne: > it doesn't match the original range

By "original range" are you referring to this null range?
> `JSON: {"low":[1e+200,1e+200,1e+200],"high":[-1e+200,-1e+200,-1e+200]}`

claudiaareneee: I mean that the `el1` range: `JSON: {"low":[-2.766357421875,-1,0],"high":[2.766357421875,1,0]}` does not match the `TextBlockLayout` range: `JSON: [[0,-1],[7,0]]`

I suppose the el0 range `isNull` because it hasn't been inserted into the model yet?

claudiaareneee: (I think I was calling the original range the `TextBlockLayout` range computed from `doLayout`)

pmconne: > I suppose the el0 range isNull because it hasn't been inserted into the model yet?

Right.

> the el1 range: JSON: {"low":[-2.766357421875,-1,0],"high":[2.766357421875,1,0]} does not match the TextBlockLayout range: JSON: [[0,-1],[7,0]]

Thanks for clarifying. The element's bounding box is computed from the bounding boxes of the `TextString`(s) inserted into its geometry stream, which are created from the `TextBlock`. Some [padding is added](https://github.com/iTwin/imodel-native/blob/5bae73c8872eea691dc6e1b36d0935293c366305/iModelCore/iModelPlatform/DgnCore/ElementGeometry.cpp#L313).

claudiaareneee: Okay, then this is as expected. I wasn't sure. So should bounding box be specifically for the geometry actually produced, and if the user wants something more explicit, use the `TextBlockLayout.range` or `TextBlockLayout.marginRange` depending on use-case? 

If so, I can leave a comment somewhere about that and delete this test.

pmconne: > should bounding box be specifically for the geometry actually produced

Yes. If at some point you start adding boxes or other kinds of frames, those should produce additional geometry that would expand the element's bounding box, accounting for the margins. When you're only producing `TextString` geometry, the size of the bounding box should be indifferent to the margins. The offsets of the `TextString`s as computed during layout would be affected by the margins.

claudiaareneee: Okay, sounds good. Whenever I do that, I can add these tests back for that specific use case. As it is, I'll remove this test for now.

pmconne (changes requested):
Need tests that verify that the generated TextString geometry is offset appropriately based on the margins.

claudiaareneee:
> Need tests that verify that the generated TextString geometry is offset appropriately based on the margins.

You're right. My current implementation doesn't move the text at all which is kind of the whole point. I'll add that. For clarification, this is what you mean right?

comment

For reference, this is what HTML does:
comment 2

pmconne:
> this is what you mean right?

Yes.

> For reference, this is what HTML does:

Do you intend to implement margin, border, and/or padding similarly to HTML?

I have been interpreting your "margin" concept as analogous to HTML's "padding" concept.
The HTML "margin" concept doesn't make sense for stand-alone text annotations (e.g., text elements), it only makes sense when there's layout happening with multiple annotations (e.g., a cell in a text table, maybe, but even then the cell already has a border shared with adjacent cells so only padding inside the border makes sense).

pmconne:
> I believe this behavior is correct for rotation and translation.

Can you explain why do you believe this?

To me it seems you've defined the margins as part of the text block. Therefore the text block's bounding box includes any extra padding added by margins. It's confusing if the anchor point is based on a second bounding box that excludes the margins. It also doesn't seem helpful, especially if in future you add (as you've indicated you want to) a visible border which presumably would encompass the padding (so, equal to the full bounding box). Example: I'd expect rotation about a middle-left anchor point to pivot about a point on the left edge of the full bounding box (top), not the text's inner range (bottom).

image

claudiaareneee:
Right, your top version is how I have it implemented. 

I agree that that the pivot point should include the padding of the margins. Basically, I think the `anchor` point as it's calculated by `calculateAnchorPoint` should include the margin as well. It wouldn't change any behavior to add them, but I think that naming is clearer. I wasn't sure about changing it if there was a use-case, so I wanted to make sure.

So should `calculateAnchorPoint` include the offset produced by the margins? (I think it should)

core/common/src/annotation/TextBlock.ts
```diff
@@ -411,6 +411,20 @@ export class Paragraph extends TextBlockComponent {
  */
 export type TextBlockJustification = "left" | "center" | "right";
 
+/** Describes the margins around the content inside a [[TextBlock]]. It's measured in meters.
+ * @beta
+ */
+export interface TextBlockMargins {
+  /** The left margin. Must be a positive integer >= 0. Negative values are disregarded */
```
pmconne, line 418: It's measured in meters, it doesn't need to be an integer.

core/common/src/annotation/TextBlock.ts
```diff
@@ -440,13 +456,17 @@ export class TextBlock extends TextBlockComponent {
...
+  public margins: TextBlockMargins;
   /** The ordered list of paragraphs within the document. */
   public readonly paragraphs: Paragraph[];
 
   private constructor(props: TextBlockProps) {
     super(props);
     this.width = props.width ?? 0;
     this.justification = props.justification ?? "left";
+    // Assign default margins if not provided
+    this.margins = { left: 0, right: 0, top: 0, bottom: 0, ...props.margins };
```
pmconne, line 469:
```suggestion
    this.margins = {
      left: props.margins.left ?? 0,
      right: props.margins.right ?? 0,
      top: props.margins.top ?? 0,
      bottom: props.margins.bottom ?? 0,
    };
```

If `props` is, e.g., `{ top: undefined }` then the spread operator will overwrite your `0` with `undefined` in `this.margins`.

test-apps/display-test-app/src/frontend/TextDecoration.ts
```diff
@@ -97,6 +97,10 @@ class TextEditor implements Decorator {
     this._textBlock.justification = justification;
   }
 
+  public setMargins(margins: Partial<TextBlockMargins>): void {
+    this._textBlock.margins = { ...this._textBlock.margins, ...margins };
```
pmconne, line 101: Spread operator problem again.

pmconne:
> I agree that that the pivot point should include the padding of the margins. Basically, the anchor point as it's calculated by calculateAnchorPoint does not include the margin offset.

The anchor point **is** the pivot point. This is where I am becoming confused.

core/common/src/annotation/TextBlock.ts
claudiaareneee: Should be fixed now

core/common/src/annotation/TextBlock.ts
claudiaareneee: Should be fixed now

claudiaareneee:
Okay, @pmconne, ready for a rereview

pmconne (approved):
Approving with comments, please also update NextVersion.md.

core/backend/src/TextAnnotationLayout.ts
```diff
@@ -497,6 +497,7 @@ export class LineLayout {
 export class TextBlockLayout {
   public source: TextBlock;
   public range = new Range2d();
+  public marginRange = new Range2d();
```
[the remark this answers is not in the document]
claudiaareneee: Yeah, that makes more sense. I'll do that.

claudiaareneee:
> Approving with comments, please also update NextVersion.md.

Okay, all updates should be made. Do you want to resolve your comments?

docs/changehistory/NextVersion.md
```diff
@@ -90,6 +91,16 @@ The new class [IndexedPolyfaceWalker]($core-geometry) has methods to complete th
 
 If a walker operation would advance outside the mesh (e.g., `edgeMate` of a boundary edge), it returns an invalid walker.
 
+### Text Block Margins
+
+A new enhancement has been made to the [TextBlock]($core-common) class to add support for margins. These margins are in meters, and are specified by using the [TextBlockMargin]($core-common) interface.
+
+The margin range is calculated in [TextAnnotationLayout.layoutTextBlock]($core-backend) and saved to the [TextBlockLayout.range]($core-backend) property. The range of just the text itself is stored in a property called [TextBlockLayout.textRange]($core-backend). In general, use of `range` is preferred over `textRange`.
```
pmconne, line 98: `textRange` is `@internal`, it should not be mentioned in public documentation.

docs/changehistory/NextVersion.md
```diff
@@ -90,6 +91,16 @@ The new class [IndexedPolyfaceWalker]($core-geometry) has methods to complete th
...
+
+The [TextBlockLayout.range]($core-backend) property will be passed into the [TextAnnotation.computeTransform]($core-common) method when producing the geometry. This will position the anchor point of the text relative to the [TextBlockLayout.range]($core-backend) property.
+
+When `debugAnchorPointAndRange` is true in [TextAnnotationGeometry.produceTextAnnotationGeometry]($core-backend), both ranges (margin and text) will be drawn in the geometry along with the anchor point.
```
pmconne, line 102: `produceTextAnnotationGeometry` is a free function, it's not inside a `TextAnnotationGeometry` object. @ben-polinsky why isn't that producing a docs build error?

I will rewrite the NextVersion.md docs.

docs/changehistory/NextVersion.md
pmconne, line 98: @ben-polinsky no docs build error despite no function called `TextAnnotationLayout.layoutTextBlock`. Is this some feature I'm not familiar with where you use the filename sans extension to disambiguate a free function's name?

docs/changehistory/NextVersion.md
pmconne: Resolving, but question for @ben-polinsky remains.

docs/changehistory/NextVersion.md
pmconne: Resolving, but question for @ben-polinsky remains.

Issue #7527, opened 2025-01-08
Presentation: Deprecate selection-related APIs


grigasp:
The APIs should be deprecated in favor of the ones in `@itwin/unified-selection` package.

PR #7707, opened 2025-02-17
Presentation: Deprecate unified selection related APIs

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/7527

https://github.com/iTwin/presentation/pull/875 should be completed first to make sure the "migration guide" link works.


saskliutas (approved):

aruniverse (approved):
Lgtm. I would like to pull docs from these new pkgs into the doc site but we can circle back to that

grigasp:
> Lgtm. I would like to pull docs from these new pkgs into the doc site but we can circle back to that

@aruniverse Those docs are mostly copy-paste from the doc site with some adjustments to match the new APIs. 

As for moving our non-itwinjs-core packages into the doc site, I think the doc site needs to be restructured to accommodate packages of different versions.

aruniverse:
> As for moving our non-itwinjs-core packages into the doc site, I think the doc site needs to be restructured to accommodate packages of different versions.

Agreed. fyi @ben-polinsky

PR #3366, opened 2022-03-16
Add "getMassPropertiesPerCandidate" RPC operation

MindaugasButkus:
This adds the ability to get mass properties for multiple elements separately with just a single request. The existing "getMassProperties" aggregates mass properties of all given candidates.

This PR is a copy of an earlier PR, but all the changes are applied to the master branch (merge to the previous branch/PR was problematic): #3361 

Q: I assume you were seeing issues with having this loop on the frontend? How many elements are you getting mass properties for?

A: It will be up to 1000 elements with each request. The reason for this whole PR is this article: https://www.itwinjs.org/learning/rpcinterface/#rpcinterface-performance ("RpcInterface methods must be "chunky" and not "chatty"").

This RPC method will be used by another backend application, so with each request to that application there would be up to 3000 requests (1000 elements all with 3 different MassPropertiesOperations) to the RPC getMassProperties.

The request round trip time adds up, so the performance degrades. Also, I did not want to put that much more load on the backend (handling 1 request instead of 3000 in a small amount of time).

I have tested the performance. Both the RpcInterface implementation and the "client" backend was running on the same machine, so network latency should have very little impact. The "client" backend implementation used caching for the getMassPropertiesPerCandidate method results, that is, get mass properties for 1000 elements and cache it in memory, so that for the next 999 elements you do not need to make additional RPC calls to get mass properties. That kind of caching does not make much sense for the older getMassProperties implementation, because it only returns mass properties for a single element, so for those tests the caching was removed. There were a total of 3438 elements processed, split up into 4 batches - 3x1000 + 438. Here are the numbers:

1st batch of 1000 elements with the new implementation was processed in 3440ms, while the old implementation took 20346ms
2nd batch of 1000 elements - 3705ms vs 20046ms
3rd batch of 1000 elements - 4418ms vs 19734ms
4th batch of 438 elements - 3746ms vs 10570ms

PR #7709, opened 2025-02-17
Deprecate `IModelConnection.getMassPropertiesPerCandidate`

GytisCepk:
`getMassPropertiesPerCandidate` was added in https://github.com/iTwin/itwinjs-core/pull/3366 to improve performance when requesting mass properties for a high number of elements individually. 

Currently, no one is actually using this method and, with us working on replacing RPC, it would be preferable to remove this method together with RPC.


core/frontend/src/IModelConnection.ts
```diff
@@ -345,15 +345,15 @@ export abstract class IModelConnection extends IModel {
...
-   */
+  /** Request element mass properties from the backend. */
   public async getMassProperties(requestProps: MassPropertiesRequestProps): Promise<MassPropertiesResponseProps> {
     return IModelReadRpcInterface.getClientForRouting(this.routingContext.token).getMassProperties(this.getRpcProps(), requestProps);
   }
 
-  /** Request mass properties for multiple elements from the backend. */
-  public async getMassPropertiesPerCandidate(requestProps: MassPropertiesPerCandidateRequestProps): Promise<MassPropertiesPerCandidateResponseProps[]> {
+  /** Request mass properties for multiple elements from the backend.
+   * @deprecated in 4.11. Use [[IModelConnection.getMassProperties]].
```
paulius-valiunas, line 354: Is it possible to replicate this functionality using `getMassProperties`?

bbastings:
@GytisCepk I don't have a problem with removing this, but how do you know it's not being used from another package? @MindaugasButkus Added this for a reason?

GytisCepk:
> @GytisCepk I don't have a problem with removing this, but how do you know it's not being used from another package? @MindaugasButkus Added this for a reason?

Over last 6 months we got 0 request to this endpoint on GPB (visualization service). Also, searched through all internal and public repos and only library, which uses this, is no longer supported.

If someone is using this in desktop application, they can simple add IPC method for this. That being said, it doesn't make sense using this over `getMassProperties` on desktop as network latency there is minimal.

core/frontend/src/IModelConnection.ts
MindaugasButkus: Definitely, if you don't mind the network traffic / latency that it adds to call that method over and over again to get mass properties for all of the elements that you need.

`getMassPropertiesPerCandidate` calculates mass properties separately for every candidate.
`getMassProperties` calculates the sum of mass properties of all provided candidates.

paulius-valiunas (approved):

tcobbs-bentley (approved):

aruniverse (approved):

GytisCepk:

PR #7720, opened 2025-02-18
Remove Node.js and DOM type references from core-common

paulius-valiunas:


core/common/src/rpc/web/WebAppRpcRequest.ts
```diff
@@ -189,7 +190,7 @@ export class WebAppRpcRequest extends RpcRequest {
     const requestClass = this.supplyRequest();
     const fetchFunction = this.supplyFetch();
 
-    const path = new URL(this.path, typeof (location) !== "undefined" ? location.origin : undefined);
+    const path = new URL(this.path, location?.origin ?? "undefined");
```
aruniverse, line 193: would prefer window?.location?.origin

core/common/src/rpc/core/RpcRequest.ts
```diff
@@ -41,7 +41,7 @@ export class ResponseLike implements Response {
   public get status() { return 200; }
   public get statusText() { return ""; }
   public get trailer(): Promise<Headers> { throw new IModelError(BentleyStatus.ERROR, "Not implemented."); }
-  public get type(): ResponseType { return "basic"; }
+  public get type(): "basic" | "cors" | "default" | "error" | "opaque" | "opaqueredirect" { return "basic"; }
```
aruniverse, line 44: do we need this union type when we are returning a literal?

core/common/src/rpc/web/WebAppRpcRequest.ts
paulius-valiunas: I wasn't sure if `window.location` is always the same as `location`... is it?

core/common/src/rpc/core/RpcRequest.ts
paulius-valiunas: technically narrowing the return type would be a breaking change (someone might have extended the class to return something else)

core/common/tsconfig.json
```diff
@@ -1,4 +1,8 @@
 {
   "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
-  "include": ["./src/**/*.ts"]
+  "include": ["./src/**/*.ts"],
+  "compilerOptions": {
+    "lib": ["ES2022", "ESNext"],
```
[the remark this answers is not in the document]
paulius-valiunas: ES2022 should be enough, I removed the other one

core/common/src/rpc/web/WebAppRpcRequest.ts
aruniverse: yes it should be, and its best to be safer
https://stackoverflow.com/questions/4709037/window-location-versus-just-location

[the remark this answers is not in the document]
ben-polinsky: I guess it depends on the value of the first argument... undefined can throw as well...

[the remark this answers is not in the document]
paulius-valiunas: oops, fixed that

ben-polinsky (approved):

paulius-valiunas:
@mergify queue

markschlosseratbentley (approved):

paulius-valiunas:
@tcobbs-bentley could you help with the failing Android build? I don't have the environment for that and I'm not sure what could be causing it... maybe just setting breakpoints at the lines I changed would show that some of the globals are undefined?

tcobbs-bentley:
> @tcobbs-bentley could you help with the failing Android build? I don't have the environment for that and I'm not sure what could be causing it... maybe just setting breakpoints at the lines I changed would show that some of the expected globals are undefined?

I will investigate. If it's something trivial, I'll probably push changes to this PR.

tcobbs-bentley:
When running in Android I get a run-time exception:
```
Uncaught (in promise) ReferenceError: global is not defined
    at get instance (index-Dm76TGMn.js:109130:25)
    at RpcManager.initializeInterface (index-Dm76TGMn.js:109400:17)
    at new MobileIpcTransport (index-Dm76TGMn.js:246995:16)
    at new MobileRpcProtocol (index-Dm76TGMn.js:247122:17)
    at <instance_members_initializer> (index-Dm76TGMn.js:247403:18)
    at new config (index-Dm76TGMn.js:247401:20)
    at RpcConfiguration.obtain (index-Dm76TGMn.js:108944:56)
    at MobileRpcManager.performInitialization (index-Dm76TGMn.js:247408:40)
    at MobileRpcManager.initializeClient (index-Dm76TGMn.js:247414:29)
    at MobileApp.startup (index-Dm76TGMn.js:247507:24)
```
I'll see if adding a declare const like for self and window fixes the problem.

core/common/src/rpc/core/RpcRegistry.ts
```diff
@@ -44,7 +47,7 @@ export class RpcRegistry {
 
   public static get instance() {
     if (!RpcRegistry._instance) {
-      const globalObj: any = typeof global !== "undefined" ? global : typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {};
+      const globalObj: any = global ?? self ?? window ?? {};
```
tcobbs-bentley, line 50: This is the line that is throwing an exception in Android. Updating it to be the following fixes the problem **_in Android_**, but I believe that you need to revert all of your changes that abandoned the old `typeof` checks, because the new code is fundamentally different (and definitely broken for `global` in Android, and probably broken for `self` and `window`):

```ts
      const globalObj: any = (typeof global !== "undefined" ? global : undefined) ?? self ?? window ?? {};
```

What I mean by that is that both `self` and `window` are defined in Android, but if they weren't, then my updated line above would not work, while the original code would. Adding the following doesn't do any good, because `declare const` just lets the TypeScript compiler know that a variable is supposed to exist; it does not actually create the variable:

```ts
declare const global: any;
```

Because of this, you _may_ need to remove the `declare const` statements that you added (in addition to switching back to the old `typeof x !== "undefined"` checks with ternary operators).

tcobbs-bentley: If you really don't like the old code, you could instead do something like this:

```ts
  private static get anyGlobal(): any {
    return typeof global !== "undefined" ? global : undefined;
  }
...
      const globalObj: any = this.anyGlobal ?? self ?? window ?? {};
```
Do the same for `anySelf` and `anyWindow`.

paulius-valiunas: Thanks for debugging this, I didn't realize it would result in a runtime error. I reverted this line.

tcobbs-bentley: If there are any platforms where `window` is undefined, your other changes that use `window?.` are going to result in the same run-time exception. Obviously `window` is defined while running on Android. It could well be that `window` is defined on all possible platforms, but if that is the case, it seems like the `?` is unneeded.

paulius-valiunas: I just reverted the other occurrences too, now it should be functionally identical (including the `typeof` checks) to the original lines.

paulius-valiunas:

Issue #7562, opened 2025-01-16
Presentation: Improve exceptional RPC responses


grigasp:
At the moment `PresentationRpcImpl` uses its own system to handle majority of exceptional cases:
 
- When a timeout occurs, it returns a specific HTTP 200 response, which tells the frontend to repeat the request. Instead, a proper HTTP response should be sent (408?)
 
- When a request processing error occurs, e.g. a bad argument or something like that, it returns an HTTP 200 response containing the error message. Instead, we should probably return HTTP 400.

aruniverse:
Fyi @paulius-valiunas @GytisCepk

PR #7723, opened 2025-02-18
Presentation: RPC improvements

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/7562.

### DIFF
--- a/common/api/presentation-common.api.md
+++ b/common/api/presentation-common.api.md
@@ -440,6 +440,12 @@ export enum ContentSpecificationTypes {
 // @public
 export type ContentUpdateInfo = typeof UPDATE_FULL;
 
+// @internal (undocumented)
+export function createCancellableTimeoutPromise(timeoutMs: number): {
+    promise: Promise<void>;
+    cancel: () => void;
+};
+
 // @public
 export function createFieldHierarchies(fields: Field[], ignoreCategories?: boolean): FieldHierarchy[];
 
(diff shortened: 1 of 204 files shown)


common/api/presentation-common.api.md
```diff
@@ -1827,6 +1833,7 @@ export interface PresentationRpcResponseData<TResult = undefined> {
 
 // @public
 export enum PresentationStatus {
+    // @deprecated
     BackendTimeout = 65543,
     Canceled = 1,
     Error = 65536,
```
paulius-valiunas, line 1839: Should this throw a runtime error (and result in HTTP status 500) instead?

grigasp: Can you clarify what you mean by "this"? `PresentationStatus.BackendTimeout` is not used anymore.

paulius-valiunas: `PresentationStatus.Error` (the last line in the snippet is the one I submitted the comment on)

grigasp: For errors that are likely bugs on our end, we throw generic `Error`, which ends up as a HTTP 500 response. 

For errors that are user input-related (e.g. your presentation rules contain a non-existing class), we throw `PresentationError` with `PresentationStatus.Error`, which ends up as HTTP 422 response.

I don't think we want to throw 5xx errors for something that's not server-related.

paulius-valiunas (approved):

saskliutas (approved):

PR #7733, opened 2025-02-19
Deprecate `GeoServiceStatus.Pending` and `GeoCoordStatus.Pending`

GytisCepk:
`GeoCoordStatus` and `GeoServiceStatus` seem to be only used to indicate coordinate conversion status. After going through the code it doesn't seem like `Pending` will be ever returned (last usage in core was removed in https://github.com/iTwin/itwinjs-core/pull/5098 and [`ReprojectStatus`](https://github.com/iTwin/imodel-native/blob/daa66a0a45de96f8e8153d4265edc7c153d8f5e1/iModelCore/GeomLibs/PublicAPI/Geom/GeoPoint.h#L49) in native does not contain `Pending` type either).

### DIFF
--- a/core/bentley/src/BentleyError.ts
+++ b/core/bentley/src/BentleyError.ts
@@ -288,6 +288,9 @@ export enum GeoServiceStatus {
   NoDatumConverter = GEOSERVICESTATUS_BASE + 3,
   VerticalDatumConvertError = GEOSERVICESTATUS_BASE + 4,
   CSMapError = GEOSERVICESTATUS_BASE + 5,
+  /**
+   * @deprecated in 5.0. Pending is no longer a valid status.
+   */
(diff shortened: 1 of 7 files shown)


core/bentley/src/BentleyError.ts
```diff
@@ -288,6 +288,9 @@ export enum GeoServiceStatus {
   NoDatumConverter = GEOSERVICESTATUS_BASE + 3,
   VerticalDatumConvertError = GEOSERVICESTATUS_BASE + 4,
   CSMapError = GEOSERVICESTATUS_BASE + 5,
+  /**
+   * @deprecated in 5.0. Pending is no longer a valid status.
```
pmconne, line 292:
```suggestion
   * @deprecated in 5.0. This status is never returned.
```

pmconne (approved):

PR #7740, opened 2025-02-19
ECSqlReader can see uncommitted changes on primary connection.

khanaffan:
imodel-native: https://github.com/iTwin/imodel-native/pull/1012

Editing application need to read changes as present on primary connection. `ECSqlReader` uses concurrent queries to read data which uses separate connection which does not see current changes on primary db.

This can be solved with `usePrimaryConn` option but we lose the concurrency in doing so.

An alternative method is to use `PRAGMA read_uncommitted=1` which only works if shared cache is on. So, this PR enables read_uncommitted from concurrent query and enables shared cache by default.

- [ ] Need to get more feedback on this fix from SQLite dev

*I sent out an email to sqlite dev asking them if enabling shared cache is clever idea and if there is alternative or they can implement something new for WAL mode.*

This is obsolete method and may cause concurrent queries to block each other on shared page cache.

### DIFF
--- a/core/backend/src/test/ecdb/ECDb.test.ts
+++ b/core/backend/src/test/ecdb/ECDb.test.ts
@@ -60,6 +60,36 @@ describe("ECDb", () => {
 
   });
 
+  it("ecsql reader can read uncommitted changes from primary connection", async () => {
+    const fileName = "schemaimport.ecdb";
+    const ecdbPath: string = path.join(outDir, fileName);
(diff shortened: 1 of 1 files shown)


khanaffan:
`PRAGMA read_uncommitted=1` is not really a good option as it see uncommitted changes and require shared page cache which is deprecated by sqlite. We settled with using `usePrimaryCon` for design application.

PR #7744, opened 2025-02-20
Forward error metadata to Frontend

pankhur94:
We currently do not send error metadata to the frontend via IPC. However, this change would enable error metadata to be sent for further processing.

- Added metadata property to IpcInvokeReturn.
- Modified test to now check for metadata.


core/backend/src/IpcHost.ts
```diff
@@ -176,7 +176,6 @@ export abstract class IpcHandler {
       } catch (err: any) {
         let ret: IpcInvokeReturn;
         if (ITwinError.isITwinError(err)) {
-          // TODO: Should metadata be left out? It is left out in the original error implementation..
           // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
pmconne, line 179:
```suggestion
```

All vars are used now, right?

pankhur94: yes you are right. I have removed it.

pankhur94:
Closing this pr. Creating new [pr](https://github.com/iTwin/itwinjs-core/pull/7769).

PR #7747, opened 2025-02-21
Account for changes for bearing format conventions

bbastings:
When the "angle" input field is being used with a bearing direction formatter/parser, update the logic that adjusts the angle (which is measured from the AccuDraw compass X axis) for the current bearing format conventions. Now design X wants 90 for "N90E" instead of 0, and design Y wants 0 for "N0E" instead of 90, which was the old convention.


bbastings:

pmconne (approved):

rschili (approved):

PR #7754, opened 2025-02-21
Decoration graphics use line color for planar region outline

### LINKED ISSUE #7751: Line color is ignored when producing decoration graphics
**Describe the bug**
GraphicParams permits specifying fill color independently from line color. Line color is supposed to be applied to lines and points, including edges of closed planar regions. But when providing a GraphicParams with differing fill and line colors to a GraphicAssembler, the output graphics always seem to use only the fill colors.

**To Reproduce**

1. Visit [this sandbox](https://www.itwinjs.org/sandboxes/MichelDastous/rosalia-the-lavender-tapir-827/).
1. Inspect Simple3dWidget.tsx, which sets fill color to green and line color to blue.
```ts
    decoratorState.setColor(ColorDef.blue);
    decoratorState.setLineThickness(15)
    decoratorState.setFillColor(ColorDef.green);
    decoratorState.setFill(true);
```
3. Observe the rectangle in the viewport displays with green outline and green fill.

**Expected behavior**
The green rectangle should have a blue outline.

**Additional context**

The same geometry if processed by the back-end tile generator would produce the expected behavior.

Reported by @mdastous-bentley.

pmconne:
Fixes #7751.

DisplayParams sets fillColor to lineColor if the geometry type is not a mesh. So points and polylines already use the correct (line) color.
But DisplayParams preserves them both for meshes. We need to choose line color when producing the outline. (If line and fill color are the same, we don't produce a separate outline, we just make an edge table using the fill color).


mdastous-bentley:
I got this branch [here](https://github.com/iTwin/itwinjs-core/tree/micheld/loopsoutlines) (that uses your fix), where I create a loop
with this symbology:
`
    builder.setSymbology(ColorDef.blue, ColorDef.fromTbgr(ColorDef.withTransparency(ColorDef.green.tbgr, 150)),  15);
`

and here is the result:
image

Also, I see the black outline only when Render mode is set to `Solid Fill`, is that expected?

pmconne:
> is this based on the "visible" edge flag being on?

Yes. Non-scene decorations draw in smooth mode with edge display turned off. You have to explicitly put them in a GraphicBranch with different view flags to override that.

bbastings:
@pmconne Ah okay, I retract everything I said, option should not be necessary. Thought you might have been trying to help with the table decorator needing to draw fill and outline separately.

mdastous-bentley:
> > is this based on the "visible" edge flag being on?
> 
> Yes. Non-scene decorations draw in smooth mode with edge display turned off. You have to explicitly put them in a GraphicBranch with different view flags to override that.

Those green squares are scene graphics, not decorations.
How should I change the black outline color ?

pmconne:
> Those green squares are scene graphics, not decorations. How should I change the black outline color ?

Explain the problem more specifically and show a small snippet of code that illustrates it - there's way too much code on your branch.
e.g., you said black outline only occurs in solid fill mode. Is there an outline in smooth mode with visible edges enabled and/or disabled? If so, what color is it?

bbastings (approved):

mdastous-bentley:
> > Those green squares are scene graphics, not decorations. How should I change the black outline color ?
> 
> Explain the problem more specifically and show a small snippet of code that illustrates it - there's way too much code on your branch.

I set the symbology in [GraphicCreator.ts](https://github.com/iTwin/itwinjs-core/blob/895c04aed43e693e1834c6be58d7b93f4c86ee33/test-apps/display-test-app/src/frontend/GraphicCreator.ts#L108)

If you want to run it, start display-test-app, open blank, click on the smiley face button in the toolbar.

> e.g., you said black outline only occurs in solid fill mode. Is there an outline in smooth mode with visible edges enabled and/or disabled? If so, what color is it?
Smooth mode looks like this (right symbology but no edge):
image

I didn't set any special edge options.  If you point me to some docs, I will look into it.

markschlosseratbentley (approved):

pmconne:
> If you want to run it, start display-test-app, open blank, click on the smiley face button in the toolbar.

No smiley face icon. Are all your changes pushed to your branch?

image

pmconne:
@mdastous-bentley

Smooth mode with visible edges turned on: blue outline, green rectangle as expected.
image

Solid fill chooses an edge color to contrast with the background color. In this case, the background color is white, so it chooses black. Obviously you can't see the view's background color because map and skybox are enabled, but that's how it works. You can override the edge color if you want:
image

mdastous-bentley:
> @mdastous-bentley
> 
> Smooth mode with visible edges turned on: blue outline, green rectangle as expected.
> 
> Solid fill chooses an edge color to contrast with the background color. In this case, the background color is white, so it chooses black. Obviously you can't see the view's background color because map and skybox are enabled, but that's how it works. You can override the edge color if you want: 
Can I enforce edge display at graphics level (i.e. not with view flags) ?

pmconne:
> Can I enforce edge display at graphics level (i.e. not with view flags) ?

Yes, you can put your graphics inside a [GraphicBranch](https://www.itwinjs.org/reference/core-frontend/rendering/graphicbranch/) and selectively override view flags for that branch.

mdastous-bentley (approved):

mdastous-bentley:
@pmconne Approved.  Thanks for the prompt fix, appreciated.

mdastous-bentley:
@pmconne Possible to backport it ?

pmconne:

PR #7768, opened 2025-02-24
Adding Channel Errors

pankhur94:


core/backend/src/internal/ChannelAdmin.ts
```diff
@@ -73,7 +73,7 @@ class ChannelAdmin implements ChannelControl {
 
     const deniedChannel = this._deniedModels.get(modelId);
     if (undefined !== deniedChannel)
-      throw new Error(`channel "${deniedChannel}" is not allowed`);
+      throw new IModelError(ChannelStatus.NotAllowed, `channel "${deniedChannel}" is not allowed`);
```
aruniverse, line 76: wonder if we should use ITwinError instead

kabentley: definitely. We should stop using IModelError entirely. Error numbers are not useful.

Issue #7571, opened 2025-01-17
Improve Editing Channel errors


karolis-zukauskas:
**Is your feature request related to a problem? Please describe.**
Attempting to delete an Element from a Channel that is not allowed results in an Error that has no "type" - errorNumber is not set - [error in question is thrown here](https://github.com/iTwin/itwinjs-core/blob/8518e5df9e4779e28ea54c3947d4024d6c693d74/core/backend/src/internal/ChannelAdmin.ts#L76).

Calling code cannot catch such error (without comparing against `message` field).

Bonus: if this error is passed to the Frontend (in case of an electron desktop app), the error object is of class `BackendError` (`IModelError`) and has an `errorNumber` of `0` - which is `IModelStatus.Success`.

![Image](https://github.com/user-attachments/assets/7c4e534d-bcb9-480e-b538-4cd7496a51a4)

**Describe the solution you'd like**
I'd like to be able to catch such error by checking its "type" and react to it e.g. by showing user a message that he cannot make changes to the Element(s).

Is there something I need to be calling before attempting to edit elements?

**Additional context**
iTwin.js: 4.10.2

Issue #7442, opened 2024-12-02
ITwinError changes


karolis-zukauskas:
**Is your feature request related to a problem? Please describe.**
[ITwinError](https://www.itwinjs.org/reference/core-common/imodels/itwinerror/) has been recently added (under beta tag), see https://github.com/iTwin/itwinjs-core/pull/6909.

While this new interface supports passing "information rich" errors from Backend to Frontend, there is now a problem of having two different interfaces for errors. One of the driving examples for improved errors was the "lock in use" error that originates in imodels-client package - detailed error information in the case of "use in locks" is lost when internal package error is converted to "platform flavour" `IModelError`. Now with the new `ITwinError`, imodels-client could start throwing it, but that cannot be easily done because it is a breaking change - major version update needed ... Concern here is that `ITwinError` will not be adopted and platform will have inconsistent errors.

**Describe the solution you'd like**
A backwards compatible solution. Perhaps a better approach would be to retain the one and only `BentleyError` class, don’t subclass it and implement `ITwinError` interface on it. Make the problematic `errorNumber` deprecated and have `namespace + errorKey` as the new way of checking error types _(we can keep the assert function approach introduced with ITwinError)_. Advantage is that we stay with one error interface and updates to it are backwards compatible – imodels-client can continue to throw IModelError (derived from BentleyError) just pass in the additional properties for the object. Over time, platform could deprecate existing BentleyError derived classes (like IModelError).

Simplified version of BentleyError expansion:
```ts
interface ITwinError {
    namespace: string;
    errorKey: string;
}

class BentleyError extends Error implements ITwinError {
    public errorNumber: number;
    public namespace: string;
    public errorKey: string;

    // New constructor
    public constructor(namespace: string, errorKey: string, message?: string);

    // Existing constructor (deprecated)
    public constructor(errorNumber: number, message?: string);

    public constructor(namespaceOrErrNumber?: number | string, errorKeyOrMessage?: string, message?: string) {
        super("TODO");
        this.namespace = "TODO";
        this.errorKey = "TODO";
        this.errorNumber = 0;
    }
}

function isBentleyError(error: unknown): error is BentleyError {
    return error instanceof BentleyError;
}

// "derived" error example:

interface FooError extends ITwinError {
    namespace: "myPackage";
    errorKey: "foo";
    extraDetail: number;
}

function throwFooError(message: string, extraDetail: number): never {
    const error = new BentleyError("myPackage", "foo", message);
[...]

PR #7774, opened 2025-02-25
Adding Channel Errors

pankhur94:
- fixes : #7571 & https://github.com/iTwin/itwinjs-backlog/issues/1402

- Refactored ITwin Error


pankhur94:
@karolis-zukauskas Could you please take a look and let me know your thoughts on the changes for [channel-error](https://github.com/iTwin/itwinjs-core/issues/7571) issue? Any feedback or suggestions would be greatly appreciated!

karolis-zukauskas:
@pankhur94,
The added error interfaces are fine, but I don't like that the implementation is so "boilerplatty". I like the `ITwinError` interface, but currently it is hard to use and hence adoption will be hard. I don't want to write an entire function (12 lines) just to throw a single type of error. We don't want every developer out there to write these constructors - they will make errors, things will be inconsistent.

This is out of scope for this specific PR/issue, but general feedback for `ITwinErrors` is that we need:
1) A convenient way to construct `ITwinError`

I suggest a function like:
```ts
namespace ITwinError {
  function make(namespace: string, errorKey: string, message: string, metadata?: LoginggMetaData): ITwinError { ... }
}
```
You could then define "constructors" like so: 
```ts
const makeFooError = ITwinError.make.bind(undefined, "namespace", "foo");
const makeBarError = ITwinError.make.bind(undefined, "namespace", "bar");
```
Note that I used "make" and not "throw" - I'd prefer a "constructor" function for the error that you later `throw` (see my [previous comment](https://github.com/iTwin/itwinjs-core/issues/7442#issuecomment-2662857170) on this).

Note that the above function works fine for `ITwinError`s that don't define any extra properties - one reason that we are switching to `ITwinError` is to be able to easily define new properties / details. So for that case I would also like a handy function, perhaps:
```ts
namespace ITwinError {
  function makeDetailed<T extends ITwinError>(namespace: string, errorKey: string, message: string, details: Omit<T, keyof ITwinError>, metadata?: LoggingMetaData): T { ... }
}

const makeInUseLocksError = (ITwinError.makeDetailed<InUseLocksError>).bind(undefined, "itwinjs-core", "in-use-locks");
```

2) A convenient way to type check specific `ITwinError` cases

I think we need a utility function to create partial functions for type assertions `ITwinError.bindIsError<T extends ITwinError>(namespace: string, errorKey: string): ITwinErrorAssertFn<T>` so usage would be
```ts
const isFooError = ITwinError.bindIsError<FooError>("myNamespace", "foo");

if (isFooError(e)) {
  e.fooDetails; // Compiler now knows e is of type FooError
}
```

FYI @nick4598

pankhur94:
@karolis-zukauskas Thank you for your feedback. Based on your suggestions, we have made adjustments to completely eliminate boilerplate code. We have introduced the following functions:
```javascript 
export function constructError(namespace: string, errorKey: string, message?: string, metadata?: LoggingMetaData): ITwinError {...}
``` 

````javascript
export function constructDetailedError<T extends ITwinError>(namespace: string, errorKey: string, details: Omit<T, keyof ITwinError>, message?: string, metadata?: LoggingMetaData): T {...}
````

for creating errors(iTwin and extended errors with details), while

````javascript
export function isValidError<T extends ITwinError>(namespace: string, errorKey: string): ITwinAssertFn<T> {....}
```` 

for type validating errors. I have taken liberty to create enums for namespaces and errorKeys(idea is the export these later) for now  as below.

````javascript
export enum ITwinErrorKeys {
  InUseLocks = "in-use-locks",
  Channel = "channel-error",
}
````

````javascript
export enum ITwinErrorNamespaces {
  ItwinJsCore = "itwinjs-core"
}
````
 Let me know of what you think.
 @aruniverse FYI

File: example-code/snippets/src/backend/ExampleCode.test.ts [hunk 1]
```diff
@@ -48,7 +48,8 @@ describe("Example Code", () => {
       try {
         await briefcaseDb.locks.acquireLocks({ exclusive: elementId });
       } catch (err) {
-        if (InUseLocksError.isInUseLocksError(err)) {
+        const isValidInUseError = ITwinError.isValidError<InUseLocksError>(ITwinErrorNamespaces.ItwinJsCore, ITwinErrorKeys.InUseLocks);
```
karolis-zukauskas, line 51: `isValidError` naming is misleading - the function checks if given `error` IS an `ITwinError`, it doesn't "validate" it. The `error` either IS or ISN'T of a certain type - I wouldn't say an error "is valid" or "is invalid".

Also the usage here is not great - you don't want to construct these assert functions every time you want to check if error is of certain type. We need:
1) A static assert function `ITwinError.isITwinError(error, namespace, errorKey)`
2) A higher order function used to define more convenient assert functions that only accept the error as parameter `ITwinError.createTypeAssertion<T>(namespace, errorKey): (error: unknown): error is T` _(not sure about the naming)_
This would be used to create an exported functions that is reused across the codebase:
```ts
// You could define this under some namespace if you wish, maybe the InUseLocksError namespace
export const isInUseLocksError = ITwinError.createTypeAssertion(ITwinErrorNamespaces.ITwinJsCore, ITwinErrorKeys.InUseLocks)`
```

core/backend/src/internal/ChannelAdmin.ts
```diff
@@ -86,11 +88,15 @@ class ChannelAdmin implements ChannelControl {
...
-      throw new Error("channels may not nest");
+    if (ChannelControl.sharedChannelName !== this.getChannelKey(args.elementId)) {
+      const error = ITwinError.constructError(ITwinErrorNamespaces.ItwinJsCore, ITwinErrorKeys.Channel, `Channel ${this.getChannelKey(args.elementId)} may not nest`);
+      throw error;
+    }
 
-    if (this.queryChannelRoot(args.channelKey) !== undefined)
-      throw new Error("a channel root for the specified key already exists");
+    if (this.queryChannelRoot(args.channelKey) !== undefined) {
+      const error = ITwinError.constructError(ITwinErrorNamespaces.ItwinJsCore, ITwinErrorKeys.Channel, `A channel root for ${args.channelKey} already exists`);
```
karolis-zukauskas, line 97: I'm not sure if it is a good idea to reuse the same errorKey for these errors ... In my application - I want to explicitly catch and gracefully handle the "channel is not allowed" case, but I don't want to handle the "channel already exists" case - that might indicate a different kind of problem/bug. To be safe, I would go for separate errorKeys for all of these throw statements.

Perhaps core developers have a stronger opinion here.

example-code/snippets/src/backend/ExampleCode.test.ts
karolis-zukauskas: Note, you could expand existing `ITwinError.isITwinError(error: unknown)` by adding optional namespace and errorKey  params. so these would be valid use cases:
```ts
ITwinError.isITwinError(e);
ITwinError.isITwinError(e, "namespace");
ITwinError.isITwinError(e, "namespace", "errorKey");
```

core/common/src/ITwinError.ts
```diff
@@ -56,49 +67,68 @@ export interface ITwinError {
...
   }
 
-}
-/** @beta */
-export namespace ITwinError {
-  /** type guard function that returns whether or not the passed in parameter is an [[ITwinError]] */
-  export function isITwinError(error: unknown): error is ITwinError {
-    return error !== undefined && error !== null && typeof error === "object" && "namespace" in error && "errorKey" in error && "message" in error;
+   /**
+ * A function which will be used to construct a details error for example [[ InUseLocksError ]] above.
```
hl662, line 101:
```suggestion
 * A function which constructs a detailed error, for example [[ InUseLocksError ]].
```

core/common/src/ITwinError.ts
```diff
@@ -56,49 +67,68 @@ export interface ITwinError {
...
+
+  /** type guard function that returns whether or not the passed in parameter is an extension of [[ITwinError]] */
+  export function isValidError<T extends ITwinError>(namespace: string, errorKey: string): ITwinAssertFn<T> {
+    return (error: unknown): error is T => {
+      return error !== undefined && error !== null && typeof error === "object" && "namespace" in error && "errorKey" in error && (error as ITwinError).namespace === namespace && (error as ITwinError).errorKey === errorKey;
+    }
   }
 
-  /** get the meta data associated with this ITwinError, if any. */
+  /** get the meta data associated with this ITwinErrorTest, if any. */
```
hl662, line 122: Shouldn't this be reverted?

example-code/snippets/src/backend/ExampleCode.test.ts
pankhur94: I have extended isItwinError function to accept three parameters now i.e. error, namespace(optional), errorKey(optional) and have used this method inside createTypeAsserter which will be reused across.

core/backend/src/internal/ChannelAdmin.ts
pankhur94: this sounds good to me and I think it helps in segregating different errors and making processing those errors easier at application end. I have added three distinct error keys for channel.

core/common/src/ITwinError.ts
pankhur94: thanks. this is done.

core/common/src/ITwinError.ts
pankhur94: yeah I missed this one. Thanks!

core/common/src/ITwinError.ts
```diff
@@ -56,49 +69,64 @@ export interface ITwinError {
...
-  namespace: "itwinjs-core";
-  errorKey: "in-use-locks";
   inUseLocks: InUseLock[];
 }
 
+type ITwinAssertFn<T extends ITwinError> = (error: unknown) => error is T;
+
 /** @beta */
-export namespace InUseLocksError {
+export namespace ITwinError {
```
ben-polinsky, line 78: What's the point of the namespace? Can we expose them from the module level i.e. `constructITwinError`?

core/common/src/ITwinError.ts
```diff
@@ -26,15 +26,28 @@ export interface InUseLock {
...
+ */
+export enum ITwinErrorNamespaces {
+  ItwinJsCore = "itwinjs-core"
+}
+
+/**
+ * An enum used to describe an error keys for a developer/application.
+ * @beta
+ */
+export enum ITwinErrorKeys {
```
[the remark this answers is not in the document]
pankhur94: Wouldn't using const object be a better approach in that case as below.
```javascript
export const ITwinErrorKeys = {
  InUseLocks: "in-use-locks",
  ChannelNest: "channel-may-not-nest",
  ChannelNotAllowed: "channel-not-allowed",
  ChannelRootExists: "channel-root-exists"
}
```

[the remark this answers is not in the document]
ben-polinsky: I'm split on that. If you do use an object literal, use `as const` to make it a bit more safe. I suppose since the error categories are used at runtime (or potentially w/o typescript) the object makes more sense.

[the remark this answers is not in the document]
pankhur94: changed from enum to const object.

core/common/src/ITwinError.ts
```diff
@@ -26,15 +26,28 @@ export interface InUseLock {
...
+ */
+export const ITwinErrorNamespaces = {
+  ItwinJsCore: "itwinjs-core"
+}
+
+/**
+ * error keys object used to describe an error keys for a developer/application.
+ * @beta
+ */
+export const ITwinErrorKeys = {
```
ben-polinsky, line 41: These objects should use `const assertions` (`as const`): https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

core/common/src/ITwinError.ts
```diff
@@ -26,15 +26,28 @@ export interface InUseLock {
   briefcaseIds: number[];
 }
 
+/**
+ * error namespaces object to describe namespaces for a developer/application.
+ * @beta
+ */
+export const ITwinErrorNamespaces = {
+  ItwinJsCore: "itwinjs-core"
```
aruniverse, line 34: we should be consistent with our naming, the key name should be ITwinJsCore not ItwinJsCore.

do we see other possibile namespace being used in the core repo?
im not a big fan of the namespace either so im for this change, but the namespace did allow for [merging of namespaces](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-namespaces), which isnt easily apparent so I cant imagine itd be used like that. if not this should just be a string literal

pankhur94: yes I would be changing it to object with const assertions as ben mentioned [here](https://github.com/iTwin/itwinjs-core/pull/7774/files#r1980066102), which can be used as a replacement for enums. However, do we have a strict rule for the casing in name. I mean since we would be declaring these as const assertions, lint wants them to follow camelCase or uppercase. Do you have any specific view on this?

pankhur94: for now, I have used Capital casing.

core/common/src/ITwinError.ts
pankhur94: done

core/common/src/ITwinError.ts
```diff
@@ -26,18 +26,33 @@ export interface InUseLock {
   briefcaseIds: number[];
 }
 
+/**
+ * error namespaces object to describe namespaces for a developer/application.
+ * @beta
+ */
+/* eslint-disable @typescript-eslint/naming-convention */
```
karolis-zukauskas, line 33: afaik `eslint-disable` comment will disable the rule for all lines afterwards until an `eslint-enable` is found. You want a `/* eslint-enable @typescript-eslint/naming-convention */` after the object is declared.

karolis-zukauskas: Though I wouldn't mind if keys started in lowercase .. would go with consistency here (i.e. go with what itwinjs core already does in such cases)

core/common/src/ITwinError.ts
ben-polinsky: Yes, if going the module-level route, we need `iTwin` in the name. My preference for staying away from namespaces (and enums) is mainly b/c of node's support for TS and the introduction of the [--eraseableSyntaxOnly](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8-beta/) flag. Also interested in @wgoehrig's thoughts

wgoehrig: I'm not as motivated by [--eraseableSyntaxOnly](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8-beta/)... I'm just not sure (a) whether the juice is worth the squeeze or (b) how realistic of a goal it is for a project like itwinjs-core given our longstanding and widespread use of namespaces and enums.  It's probably worth having a separate discussion (maybe a new issue?) on whether we want to avoid all of those constructs going forward.

I do think we tend to overuse namespaces tho.  IMO they're mostly a legacy feature from the pre-module era. With modules, everything is always implicitly namespaced to a module, so if there's any ambiguity you can always `import as` locally.  Granted, the way we re-export whole packages from barrel files means there is still probably a place for _some_ namespaces, but then you have to deal with the fact that TypeScript's [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) with namespaces isn't exactly intuitive.

Anyway, in this case I agree that a namespace here is overkill and `constructITwinError` is a clearer name.

ben-polinsky: > It's probably worth having a separate discussion (maybe a new issue?) 

I agree. There are other debates around enums and namespaces - so it'd be nice to come to a conclusion moving forward no matter the motivation.

core/common/src/ITwinError.ts
```diff
@@ -56,49 +71,71 @@ export interface ITwinError {
  * @beta
 */
 export interface InUseLocksError extends ITwinError {
-  namespace: "itwinjs-core";
-  errorKey: "in-use-locks";
```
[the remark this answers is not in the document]
pankhur94: since we will be using constructDetailedError to create errors extending ITwinError including "InUseLocksError", we will be passing namespace and errorKey during the its creation.

core/common/src/ITwinError.ts
```diff
@@ -56,49 +71,71 @@ export interface ITwinError {
...
-};
+/**
+ * type guard function that returns whether or not the passed in parameter is an [[ITwinError]]
+ * @param error The error which is to ve verified.
+ * @param namespace The namespace associated with the error.
+ * @param errorKey The errorKey associated with the error.
+ * @beta
+*/
+export function isITwinError(error: unknown, namespace?: string, errorKey?: string): error is ITwinError {
+  return error !== undefined && error !== null && typeof error === "object" && "namespace" in error && "errorKey" in error && "message" in error && (namespace === undefined || (error as ITwinError).namespace === namespace) && (errorKey === undefined || (error as ITwinError).errorKey === errorKey);
```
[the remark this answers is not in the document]
pankhur94: What do you think about this?
```javascript
export function isITwinError(error: unknown, namespace?: string, errorKey?: string): error is ITwinError {
  return isValidErrorObject(error) && hasRequiredProperties(error) && matchNamespace(error, namespace) && matchErrorKey(error, errorKey);
}

// Check if the error is a valid object
function isValidErrorObject(error: unknown): boolean {
  return error !== undefined && error !== null && typeof error === "object";
};

// Check if the error contains the required properties
function hasRequiredProperties(error: any): boolean {
  return "namespace" in error && "errorKey" in error && "message" in error;
};

// Check if the namespace is correct
function matchNamespace(error: any, namespace?: string): boolean {
  return (namespace === undefined || (error as ITwinError).namespace === namespace);
}

// Check if the errorKey is correct
function matchErrorKey(error: any, errorKey?: string): boolean {
  return (errorKey === undefined || (error as ITwinError).errorKey === errorKey);
}
```

core/common/src/ITwinError.ts
pankhur94: I noticed that we are using camelCase for const assertions, hence keeping in line with it, I have changed it to camelCase.

core/common/src/ITwinError.ts
```diff
@@ -56,49 +71,71 @@ export interface ITwinError {
  * @beta
 */
 export interface InUseLocksError extends ITwinError {
-  namespace: "itwinjs-core";
-  errorKey: "in-use-locks";
   inUseLocks: InUseLock[];
```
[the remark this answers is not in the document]
pankhur94: done

core/common/src/ITwinError.ts
```diff
@@ -56,49 +71,71 @@ export interface ITwinError {
...
+* @param errorKey The errorKey associated with the error.
+* @param details Other details associated with the error.
+* @param message The message associated with the error.
+* @param metadata Metadata associated with the error.
+* @beta
+*/
+export function constructDetailedError<T extends ITwinError>(namespace: string, errorKey: string, details: Omit<T, keyof ITwinError>, message?: string, metadata?: LoggingMetaData): T {
+  const baseError = constructError(namespace, errorKey, message, metadata);
+
+  return Object.assign(baseError, details) as T;
```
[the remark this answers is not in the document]
pankhur94: spread operator wouldn't preserve non-enumerable props i.e. stack. Hence Object.assign would be better in this case.

core/common/src/ITwinError.ts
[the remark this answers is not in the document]
pankhur94: done, split in multiple lines

core/common/src/ITwinError.ts
```diff
@@ -56,49 +71,71 @@ export interface ITwinError {
...
-export namespace InUseLocksError {
+/**
+* A function which will be used to construct an error.
+* @param namespace The namespace associated with the error.
+* @param errorKey The errorKey associated with the error.
+* @param message The message associated with the error.
+* @param metadata Metadata associated with the error.
+* @beta
+*/
+export function constructError(namespace: string, errorKey: string, message?: string, metadata?: LoggingMetaData): ITwinError {
```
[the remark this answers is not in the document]
pankhur94: done

core/common/src/ITwinError.ts
```diff
@@ -56,49 +71,71 @@ export interface ITwinError {
...
 export interface InUseLocksError extends ITwinError {
-  namespace: "itwinjs-core";
-  errorKey: "in-use-locks";
   inUseLocks: InUseLock[];
 }
 
-/** @beta */
-export namespace InUseLocksError {
+/**
+* A function which will be used to construct an error.
```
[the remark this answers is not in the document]
pankhur94: done

core/common/src/ITwinError.ts
```diff
@@ -56,49 +71,71 @@ export interface ITwinError {
...
+    errorKey,
+    message: message ?? `${errorKey} occurred`,
+    metadata,
+  };
 
-  /** throws an error which passes the [[InUseLocksError.isInUseLocksError]] type guard function */
-  export function throwInUseLocksError(inUseLocks: InUseLock[], message?: string, metadata?: LoggingMetaData): never {
-    const errorObject = new Error();
-    errorObject.name = "InUseLocksError"; // optional but makes it so that when the error is thrown and not caught we see InUseLocksError: 'message' instead of Error: 'message'
-    Error.captureStackTrace(errorObject, throwInUseLocksError); // optional: whether we want to hide throwInUseLocksError or not from the stack. not super important
```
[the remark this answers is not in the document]
pankhur94: thanks for pointing this out. I have added stack trace back by instantiating Error inside createItwinError.

full-stack-tests/core/src/frontend/standalone/Error.test.ts
```diff
@@ -24,16 +24,37 @@ if (ProcessDetector.isElectronAppFrontend) {
...
+      let caughtError = false;
+      try {
+        await coreFullStackTestIpc.throwITwinError(ITwinErrorNamespaces.ITwinJsCore, ITwinErrorKeys.ChannelNest, message, metadata);
+      } catch (err) {
+        caughtError = true;
+        const isInUseError = createTypeAsserter<ITwinError>(ITwinErrorNamespaces.ITwinJsCore, ITwinErrorKeys.ChannelNest);
+        expect(isInUseError(err)).to.be.true;
+        if (isInUseError(err)) {
+          // Even though we're on the frontend we should make sure our stack trace includes backend code.
+          expect(err.stack?.includes("core\\backend") || err.stack?.includes("core/backend"), `Expected ${err.stack} to have mention of 'core\\backend' or 'core/backend'`).to.be.true;
```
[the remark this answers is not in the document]
pankhur94: [same.](https://github.com/iTwin/itwinjs-core/pull/7774#discussion_r1982381477)

core/common/src/ITwinError.ts
aruniverse: I dont know if thats necessary, and I think that can probably be determined from the stack trace where the error was initially thrown. other namespaces i can think of are the shared packages layer or at the app layer, both which would be outside of the core repo

core/backend/src/internal/ChannelAdmin.ts
```diff
@@ -86,11 +88,15 @@ class ChannelAdmin implements ChannelControl {
   }
 
   public makeChannelRoot(args: { elementId: Id64String, channelKey: ChannelKey }) {
-    if (ChannelControl.sharedChannelName !== this.getChannelKey(args.elementId))
-      throw new Error("channels may not nest");
+    if (ChannelControl.sharedChannelName !== this.getChannelKey(args.elementId)) {
```
aruniverse, line 91: nit: extract `this.getChannelKey(args.elementId)` into a local variable that you use in the if stmt and the msg

core/backend/src/internal/ChannelAdmin.ts
```diff
@@ -72,8 +72,10 @@ class ChannelAdmin implements ChannelControl {
       return;
 
     const deniedChannel = this._deniedModels.get(modelId);
-    if (undefined !== deniedChannel)
-      throw new Error(`channel "${deniedChannel}" is not allowed`);
+    if (undefined !== deniedChannel) {
+      const error = constructError(ITwinErrorNamespaces.ITwinJsCore, ITwinErrorKeys.ChannelNotAllowed, `channel "${deniedChannel}" is not allowed`);
```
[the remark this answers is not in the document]
aruniverse: nice pass at it, its a good start. i dont know if we need the additional namespace separation and that just adds unnecessary complexity. 


playing around with some copilot suggestions and it spit out  the following which we can clean up and get us trending towards the right direction

```ts
type ErrorType = typeof iTwinErrorKeys[keyof typeof iTwinErrorKeys];
type ErrorMessageFn = (...args: any[]) => string;

export const ITwinErrorMessages: Record<ErrorType, ErrorMessageFn> = {
  'in-use-locks': () => `One or more objects are already locked by another briefcase`,
  'channel-may-not-nest': (id: number) => `channel ${id} may not nest`,
  'channel-not-allowed': (id: number) => `channel "${id}" is not allowed`,
  'channel-root-exists': (id: number) => `channel ${id} root doesn't exist`
}
```

pankhur94:
[sending metadata for bentley error pr](https://github.com/iTwin/itwinjs-core/pull/7795) closed in favour of this pr and changes moved.

core/common/src/ITwinError.ts
pankhur94: for now, I have changed it to a string literal, if need be we can modify it in the future.

core/backend/src/internal/ChannelAdmin.ts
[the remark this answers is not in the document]
pankhur94: this is a good idea. I have applied your suggestions.

core/backend/src/internal/ChannelAdmin.ts
pankhur94: done

aruniverse (approved):
thx @pankhur94 , looks much cleaner than your first pass and lgtm

@karolis-zukauskas if you can review as the filer of the original issue and see if this meets your needs thatd be great

core/common/src/ITwinError.ts
```diff
@@ -23,21 +24,44 @@ export interface InUseLock {
    */
   state: LockState;
   /** An array of Briefcase ids that hold this lock. */
-  briefcaseIds: number[];
+  briefcaseIds: BriefcaseId[];
 }
 
+/**
```
kabentley, line 30: I realize that this PR has already merged, but this is not the right place and not the right name for these channel error types. 

The type `iTwinErrorMessages` is definitely not a "record for all itwin error messages". Maybe it is a place for all iTwin errors thrown by "core-backend"? I'd move it to a file called "ChannelErrors" or something so it doesn't encourage other to think they need to add entries here.

PR #7775, opened 2025-02-25
Remove V1 checkpoints API

tcobbs-bentley:


core/backend/src/core-backend.ts
```diff
@@ -15,6 +15,7 @@ export * from "./ChannelControl";
 export * from "./CheckpointManager";
 export * from "./ClassRegistry";
 export * from "./CloudSqlite";
+export * from "./CloudSqliteMock";
```
aruniverse, line 18: do we need to export this from the main barrel?
should this internal api be imported from some sub path?

core/backend/src/CloudSqliteMock.ts
```diff
@@ -0,0 +1,62 @@
...
+import * as sinon from "sinon";
+import { CloudSqlite } from "./CloudSqlite";
+import { HubMock } from "./HubMock";
+import { CheckpointProps } from "./CheckpointManager";
+
+/**
+ * Mocks the CloudSqlite access needed for [[V2CheckpointManager.downloadCheckpoint]].
+ * @internal
+ */
+export class CloudSqliteMock {
```
aruniverse, line 15: i would move this class into the test dir since its a util, and we dont necessarily want to expose this as part of the pkg api

core/backend/src/core-backend.ts
tcobbs-bentley: I copied what is being done with `HubMock`, which is also marked as internal. Both classes are imported and used in full-stack-tests and example-code.

core/backend/src/CloudSqliteMock.ts
tcobbs-bentley: I put it next to HubMock.ts, which is used in the same way. I could move both files, but even though HubMock.ts doesn't use sinon, I do feel that both classes go together.

core/backend/src/core-backend.ts
tcobbs-bentley: I'll hide CloudSqliteMock and do its startup and shutdown inside the startup and shutdown for HubMock.

tcobbs-bentley: With startup/shutdown happening inside HubMock, I am no longer exporting CloudSqliteMock.

core/backend/src/CloudSqliteMock.ts
tcobbs-bentley: To be honest, I feel that HubMock.ts should probably be moved to the test dir, but I have left that alone for now. Now that CloudSqliteMock is only accessed from HubMock, I moved CloudSqliteMock.ts into the test directory.

core/backend/src/rpc-impl/RpcBriefcaseUtility.ts
```diff
@@ -58,8 +58,13 @@ export class RpcBriefcaseUtility {
           const fileName = resolver({ briefcaseId, iModelId });
           if (IModelJsFs.existsSync(fileName)) {
             const briefcaseDb = BriefcaseDb.findByFilename(fileName);
-            if (briefcaseDb !== undefined)
-              return briefcaseDb as BriefcaseDb;
+            if (briefcaseDb !== undefined) {
+              if (briefcaseDb instanceof BriefcaseDb) {
```
aruniverse, line 62: do you need the if on L61? would 62 be a stricter check that suffices?

core/backend/src/CheckpointManager.ts
```diff
@@ -198,6 +199,9 @@ export class V2CheckpointManager {
 
     try {
       const container = this.getContainer(v2props, checkpoint);
+      if (container instanceof CloudContainerMock) {
```
aruniverse, line 202: should we be leaking test code into the actual api?

aruniverse (commented):
looks fine, but would be good for @wgoehrig to review

core/backend/src/CheckpointManager.ts
tcobbs-bentley: I couldn't figure any other way to make it work. I don't really like this solution, but nothing else I tried worked, because running the subsequent code doesn't work. I'm open to suggestions.

core/backend/src/rpc-impl/RpcBriefcaseUtility.ts
tcobbs-bentley: Not really, because line 65 needs to run only if briefcaseDb is not undefined. I guess I could change else to an else if not undefined.

core/backend/src/test/standalone/IModelWrite.test.ts
```diff
@@ -209,7 +212,10 @@ describe("IModelWriteTest", () => {
     expect(nClosed).equal(1);
 
     bc.close();
-    sinon.restore();
+    // NOTE: Since HubMock.startup() is called in the before() block and not beforeEach(), we CANNOT
+    // call sinon.restore() here. This is because sinon.restore() will restore the stubs for that
```
wgoehrig, line 216:
```suggestion
    // call sinon.restore() here. This is because sinon.restore() will restore the stubs for CloudSqlite that
```

core/backend/src/test/standalone/IModelWrite.test.ts
```diff
@@ -278,7 +284,10 @@ describe("IModelWriteTest", () => {
...
```
wgoehrig, line 288:
```suggestion
    // call sinon.restore() here. This is because sinon.restore() will restore the stubs for CloudSqlite that
```

core/backend/src/rpc-impl/RpcBriefcaseUtility.ts
wgoehrig: I prefer the current flow, but we should probably still avoid `instanceof` where possible...
```suggestion
              if (briefcaseDb.isBriefcaseDb()) {
```

core/backend/src/CheckpointManager.ts
wgoehrig: Can you be more specific about what doesn't work?  Your `CloudContainerMock.connect()` already "downloads" the checkpoint, so can't we have `HubMock.queryV2Checkpoint()` return a `V2CheckpointAccessProps` with the full path set as `dbName`?

core/backend/src/rpc-impl/RpcBriefcaseUtility.ts
kabentley: This PR would be a good time to finally get rid of "RpcBriefcase" entirely. If you're using Rpc, you can't have a Briefcase. I think it only comes up for the now-deprecated `SyncMode.PullOnly` and `SyncMode.PullAndPush` use cases that are only used by tests, and should be eliminated.

core/backend/src/CheckpointManager.ts
kabentley: the whole reason that HubMock is in the package is so Sam could use it for his testing. Is there a better way to do that?

core/backend/src/test/standalone/IModelWrite.test.ts
tcobbs-bentley: Fixed in three places.

core/backend/src/CheckpointManager.ts
tcobbs-bentley: > Can you be more specific about what doesn't work? Your `CloudContainerMock.connect()` already "downloads" the checkpoint, so can't we have `HubMock.queryV2Checkpoint()` return a `V2CheckpointAccessProps` with the full path set as `dbName`?

This `attach` function is called by `SnapshotDb.attachAndOpenCheckpoint`. It then uses `dbName` and `container` from that return and passes them into `SnapshotDb.openFile`. In order for things to work, the `container` passed into `SnapshotDb.openFile` needs to be undefined. This function isn't actually allowed to have container be undefined in its return, so `CloudSqliteMock.attach` typecasts to pretend that it isn't undefined.

I couldn't figure out any way to make things work without the `container` returned from `attach` being undefined. And since that required the code to know about the mock, I did it this way.

tcobbs-bentley: Since `V2CheckpointManager` is internal, I updated the return type of `attach` to reflect the changes I made.

tcobbs-bentley: @aruniverse @wgoehrig I was able to get the tests to pass by updating the last failing one to use "ThirdVersion" instead of "SecondVersion" for the second checkpoint. It appears that "SecondVersion" doesn't have a V2 checkpoint, but "FirstVersion" and "ThirdVersion" do. That means that the pipeline now passes, and this open conversation and an appropriate reviewer are the only things preventing this PR from merging. Are my answers above sufficient, or are more changes required?

core/backend/src/test/CloudSqliteMock.ts
```diff
@@ -0,0 +1,95 @@
...
+      } else {
+        return origCreateCloudContainer(args);
+      }
+    }));
+    this._stubs.push(sinon.stub(CloudSqlite, "transferDb").callsFake(async (direction, container: any, props) => {
+      if (container instanceof CloudContainerMock) {
+        if (direction === "download") {
+          container.downloadCheckpoint(props.localFileName);
+        } else {
+          debugger; // eslint-disable-line no-debugger
```
aruniverse, line 74: id remove the debugger call

core/backend/src/CheckpointManager.ts
aruniverse: I would still prefer to not check if its instance of a Mock we created for testing in the actual checkpoint manager

core/backend/src/test/CloudSqliteMock.ts
tcobbs-bentley: Fixed.

core/backend/src/CheckpointManager.ts
tcobbs-bentley: > I would still prefer to not check if its instance of a Mock we created for testing in the actual checkpoint manager

The only alternative I can come up with is to stub out `CheckpointManager.attach`, then copy the first 11 lines of that function into the stub and then return `container.attach()` in the spot that the `instanceof` check is currently being performed. I can do that, but I feel that it is worse than what is there now.

tcobbs-bentley: I made the changes mentioned above. I don't personally believe that this is superior to the mock check in the projection code, but it removes that. I did have to make `V2CheckpointManager.getContainer` public in order to do the stub.

aruniverse: the class is internal, no one else should be using it so thats fine

aruniverse (approved):
lgtm but would like @wgoehrig  to have final approval before merging

tcobbs-bentley:
> looks fine, but would be good for @wgoehrig to review

@wgoehrig I'm waiting on your review.

wgoehrig (approved):

Issue #7618, opened 2025-01-27
ScreenViewport.waitForSceneCompletion does not resolve under certain conditions


thanhbinh-d:
**Describe the bug**
When a sheet contains a Drawing attachment with an incorrect X offset, the function `waitForSceneCompletion` does not resolve. This causes the application to hang indefinitely, preventing further actions.

**Steps to Reproduce:**
1. Open the iModel and switch to the Civil Frontstage. (Note: Contact Raphael Lemieux for permission to access the test model.)
3. In the Civil Frontstage, open the sheet "A001 cord22 - 606+50xxx" in the Sheet Viewer Widget.
4. Click the Print Preview button.
5. Observe that the function hangs because it uses the waitForSceneCompletion API before capturing the sheet image.

**Expected behavior**
The function `waitForSceneCompletion` should resolve even when there are incorrect Drawing attachments, ensuring that the viewport scene completes rendering regardless of attachment issues.

**Suggestion for Improvement:**
To make the API more robust and user-friendly, consider adding an optional timeout parameter to the waitForSceneCompletion function. This would allow developers to specify a maximum wait time, preventing indefinite hangs in edge cases like this. 
For example:
`await viewport.waitForSceneCompletion({ timeout: 100 }); // Timeout in milliseconds`

**Screenshots**

![Image](https://github.com/user-attachments/assets/c660f305-7541-4776-8982-767e71f1ee11)

pmconne:
Hi @thanhbinh-d, can you please provide steps to reproduce?

thanhbinh-d:
> Hi [@thanhbinh-d](https://github.com/thanhbinh-d), can you please provide steps to reproduce?

The bug description was updated.
Thanks,
Binh

pmconne:
> Contact Raphael Lemieux for permission to access the test model

@raplemie are you able to provide iTwin+iModel Id and grant read access to me and @markschlosseratbentley?

raplemie:
@markschlosseratbentley @pmconne added, this iModel is in QA:
iTwin 71f06c2a-d1a1-466f-8566-40bc85863c11
iModel 62e69c4c-86af-4d54-a22f-fee6c7c7b9c8
Changeset ce8e877b6c5ec33b4707e762abe189a4abadd625

(Mentioning changeset as later changeset modify the location of the additional graphics and do not exhibit the same issue)

markschlosseratbentley:
[...]

PR #7785, opened 2025-02-27
Add attachment out of view check to ViewAttachments.areAllTileTreesLoaded

andremig-bentley:
This PR is to address this issue: https://github.com/iTwin/itwinjs-core/issues/7618

Currently, it is possible for `Viewport.waitForSceneCompletion` to hang on the check for `areAllTileTreesLoaded()`. This happens in a `SheetViewState` if there is an `Attachment` outside of the view that has not been loaded yet once `waitForSceneCompletion` was called. When checking if each attachment tile tree was loaded, there was no check for the case where an attachment was outside of the current view and therefore might not have been loaded yet. In this case, the attachment out of view would not load by design, but `areAllTileTreesLoaded` would continually return `false`, causing the bug.

To fix this, a check has been added to `ViewAttachments.areAllTileTreesLoaded` which determines whether or not the attachment being checked is within the view. If it is not, then we skip checking if the associated tile tree is loaded. A test has also been added to `SheetView.test.ts` to recreate the issue.


markschlosseratbentley (approved):

markschlosseratbentley:
Might be a good idea to run Image Tests on this branch.

andremig-bentley:
> The bug appears to be that we are not detecting when an attempt to load a tile tree irrecoverably fails. Did you look into fixing that?

The most recent commit includes this fix. The bug occurs when an attachment is outside of the view and has therefore not been loaded by the time it is checked for whether or not its tile trees are loaded. This incorrect behavior is fixed by adding a check for whether or not the attachment is within the view and if it is not, we skip checking if the tile tree is loaded.

core/frontend/src/SheetViewState.ts
```diff
@@ -309,7 +309,16 @@ class ViewAttachments {
   }
 
   public get areAllTileTreesLoaded(): boolean {
-    return this._attachments.every((x) => x.areAllTileTreesLoaded);
+    return this._attachments.every((x) => {
+      const vp = IModelApp.viewManager.selectedView;
```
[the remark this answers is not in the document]
andremig-bentley: You're right that this needs to be independent of the selected view, but I don't think the `SheetViewState`'s `getViewedExtents` will work. `SheetViewState.getViewedExtents` returns `this._extents`. That is set [here](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/SheetViewState.ts#L416) to be equal to the sheets' size. It doesn't actually correspond to the extents of what is being viewed, and therefore doesn't change as the view changes, which is necessary for the check. Digging into this to find a different solution to uncouple from the selected viewport.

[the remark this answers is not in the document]
andremig-bentley: Most recent commit I believe solves this. Now, instead of getting the viewed extents from the selected view, the range is obtained from the `SheetViewState`'s frustum, which should mean we are uncoupled from the selected view.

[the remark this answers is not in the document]
pmconne: That should work for 2d views (for 3d views, the viewport can operate on an expanded frustum which might differ from the ViewState's frustum).

full-stack-tests/core/src/frontend/hub/SheetViewState.test.ts
```diff
@@ -168,4 +169,29 @@ describe("Sheet views (#integration)", () => {
     delete (clone.viewAttachmentProps[0] as any).attachedView;
     expect(clone.viewAttachmentProps).to.deep.equal(view.viewAttachmentProps);
   });
+
+  describe("areAllTileTreesLoaded", () => {
+    it("should return true when attachments are outside of the view", async () => {
```
pmconne, line 174: If you're trying to prove that `areAllTileTreesLoaded` will return `true` when an attachment's tile tree is not loaded because it's outside the view's frustum, make your test verify the latter condition too.
Then, fit the view. Expect `areAllTileTreesLoaded` to now be `false`, because the attachment is inside the frustum and hasn't had time to load its tile tree.
Then, verify the tile tree loads asynchronously and `areAllTileTreesLoaded` becomes `true` again.

andremig-bentley: Most recent commit begins to address this, but I'm having some trouble with the test, so this is an open request for help.

I'm having trouble getting the new attachment I create to actually attach to the sheetview. I am attempting to follow what I think is the right methodology laid out [here](https://github.com/iTwin/itwinjs-core/blob/e98d4e39fc2b4d46949d12ed42b89187041f0d01/core/frontend/src/SheetViewState.ts#L129) in the `ViewAttachmentsInfo` docs. I have tried following this order described here, I have tried to create a sort of fake response from the bakend to do the postLoad with (which is what's being done here), and I've tried creating test classes extending these classes which allow me to directly edit things like `SheetViewState._attachments`, but I've had no luck so far. The final expect statement on line 209 continues to fail because no attachment is being correctly attached to the `SheetViewState`. 

Can anyone see where I'm going wrong? Is there some other way to attach a new attachment to a SheetViewState, or am I sort of pigeon holed into creating a new model and uploading it to iModelHub (which would explain why these tests are here in the full stack integration tests and not somewhere else)?

andremig-bentley: I've been struggling to use the `FullStackTestIpc` backend to insert an element in the backend. I'm trying to do something similar to [this function](https://github.com/iTwin/itwinjs-core/blob/c6fb59374d8b8d9b4bcdfaba74fddcda0d9106f7/full-stack-tests/core/src/backend/backend.ts#L48) to insert a view attachment as an element. When I go to insert the element, I get back an error that simply says "Error inserting element [error=e9086000]" where that final string is some sort of id that changes every time the test is run. Putting my insert method into a try catch block and using `JSON.stringify()` on the error, I get: "{"errorNumber":588114720,"name":"Error"}". I see the spot creating the error is [here](https://github.com/iTwin/itwinjs-core/blob/c6fb59374d8b8d9b4bcdfaba74fddcda0d9106f7/core/backend/src/IModelDb.ts#L2108) in `IModelDb.ts`, but I'm not sure what to do to track down the true source of the error. Any suggestions?

full-stack-tests/core/src/backend/backend.ts
```diff
@@ -86,6 +88,116 @@ class FullStackTestIpcHandler extends IpcHandler implements FullStackTestIpc {
...
+    const newElement = standaloneModel.elements.createElement(newAttachmentProps);
+    const attachmentId =  standaloneModel.elements.insertElement(newElement.toJSON());
+
+    const displayStyle2dId = DisplayStyle2d.insert(standaloneModel, drawingDefinitionModelId, "DisplayStyle2d");
+    const drawingCategorySelectorId = CategorySelector.insert(standaloneModel, drawingDefinitionModelId, "DrawingCategories", [drawingCategoryId]);
+    const drawingViewRange = new Range2d(0, 0, 100, 100);
+    const drawingViewId = DrawingViewDefinition.insert(standaloneModel, drawingDefinitionModelId, "Drawing View", sheetModelId, drawingCategorySelectorId, displayStyle2dId, drawingViewRange);
+
+    //create new sheet view
+    const sheetViewProps = await standaloneModel.views.getViewStateProps(drawingViewId);
```
pmconne, line 186: `DrawingViewDefinition`s do not have `ViewAttachment`s, and their `baseModelId` must point to a drawing model (yours points to a sheet model here).
You are putting your `ViewAttachment`s into the sheet model, which is correct.
If you create a `SheetViewDefinition` viewing that sheet model then we will automatically populate the list of view attachments in `IModelDb.views.loadViewData`.

pmconne: However, I note that `SheetViewDefinition` has no public methods for instantiating it...lemme dig around.

pmconne: Pushed changes. Tried to run test, got the following:
```
j2 core: andremig/waitforscene 14:26:00 ~/c/j2/s/itwinjs/full-stack-tests/core 0> npm run test:chrome

> core-full-stack-tests@0.0.0 test:chrome
> run-p -r azurite:setup test:chrome:azurite


> core-full-stack-tests@0.0.0 azurite:setup
> mkdirp ./lib/cjs/test/azuriteStorage && azurite-blob --silent --loose --location ./lib/cjs/test/azuriteStorage


> core-full-stack-tests@0.0.0 test:chrome:azurite
> certa -r chrome --grep "#integration" --invert

Certa web frontend now listening on port 3010
Web backend for full-stack-tests listening on port 5010
Azurite Blob service is starting on 127.0.0.1:10000
Azurite Blob service successfully listens on http://127.0.0.1:10000


  0 passing (29ms)

ERROR: There were 0 passing tests.  That doesn't seem right.
If there are really no passing tests and no failures, then what was even the point?
It seems likely that tests were skipped by it.only, it.skip, or grep filters, so I'm going to fail now.
```

pmconne: Error is because these are `#integration` tests, skipped by `npm run test/cover`. Your test doesn't need to be an integration test because it doesn't talk to any cloud services.

andremig-bentley: Running `npm run test:integration` within the `full-stack-tests/core` dir should run it for now, but I'll move it to somewhere that makes more sense with next commits.

pmconne: Fixing it now...

pmconne: Pushed. With or without those latest changes, running the test produces the following - dunno if that's what you were seeing previously.
```
    ViewAttachments
      1) areAllTileTreesLoaded should return true when attachments are outside of the viewed extents


  0 passing (7s)
  1 failing

  1) Sheet views (#integration)
       ViewAttachments
         areAllTileTreesLoaded should return true when attachments are outside of the viewed extents:
     TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
      at Object.join (node:path:1244:7)
      at FullStackTestIpcHandler.insertViewAttachmentAndGetSheetViewProps (/home/paul/c/j2/s/itwinjs/full-stack-tests/core/src/backend/backend.ts:93:27)
      at /home/paul/c/j2/s/itwinjs/core/backend/src/IpcHost.ts:174:37
      at IpcWebSocketBackend.processMessages (/home/paul/c/j2/s/itwinjs/core/common/src/ipc/IpcWebSocket.ts:175:32)
      at IpcWebSocketBackend.dispatch (/home/paul/c/j2/s/itwinjs/core/common/src/ipc/IpcWebSocket.ts:159:16)
      at /home/paul/c/j2/s/itwinjs/core/common/src/ipc/IpcWebSocket.ts:138:53
      at WebSocket.<anonymous> (/home/paul/c/j2/s/itwinjs/core/backend/src/LocalhostIpcHost.ts:54:9)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

pmconne: I don't know what's supposed to automatically configure that path for me but manually setting it reveals test failure. I adjusted your test slightly to show that while `sheetViewStateProps.sheetAttachments.length` is `1`, `newSheetView.viewAttachmentProps.length` is `0`. Presumably when loading the view attachment it discovered the element was improperly configured.

full-stack-tests/core/src/backend/backend.ts
```diff
@@ -86,6 +88,124 @@ class FullStackTestIpcHandler extends IpcHandler implements FullStackTestIpc {
...
+    const codeProps = { spec: "", scope: "", value: "" };
+    sheetViewProps.sheetProps = {
+      model: "",
+      code: codeProps,
+      classFullName: "",
+      width: 100,
+      height: 100,
+      scale: 1,
+    };
+    sheetViewProps.sheetAttachments = [attachmentId];
```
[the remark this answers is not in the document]
pmconne: Removed. `getViewStateProps` is indeed setting `sheetAttachments` to `["0x17"].

full-stack-tests/core/src/backend/backend.ts
andremig-bentley: The path issue is probably a .env issue, I'm using the `IMODELJS_CORE_DIRNAME` env variable used in a few other full stack tests to get the local core path. I'm seeing the same failing based on `viewAttachmentProps.length` and have been for a bit. Digging in now to the loading to try and find out how it's failing.

full-stack-tests/core/src/frontend/hub/SheetViewState.test.ts
```diff
@@ -169,3 +174,94 @@ describe("Sheet views (#integration)", () => {
...
+
+    it("areAllTileTreesLoaded should return true when attachments are outside of the viewed extents", async () => {
+      if (!process.env.IMODELJS_CORE_DIRNAME)
+        throw new Error("IMODELJS_CORE_DIRNAME not set");
+
+      const vp = openBlankViewport({ height: 100, width: 100 });
+      const sheetViewStateProps = await coreFullStackTestIpc.insertViewAttachmentAndGetSheetViewProps();
+      expect(sheetViewStateProps.sheetAttachments).not.to.be.undefined;
+      expect(sheetViewStateProps.sheetAttachments!.length).to.equal(1);
+      const newSheetView = SheetViewState.createFromProps(sheetViewStateProps, await BriefcaseConnection.openStandalone(path.join(process.env.IMODELJS_CORE_DIRNAME, "core/backend/lib/cjs/test/assets/sheetViewTest.bim")));
```
pmconne, line 257: Your viewport is viewing a `BlankConnection`. Your `SheetViewState` is from a `BriefcaseConnection`. They have to be the same iModel. Let me try to fix it.

pmconne: Hacked that in.

full-stack-tests/core/src/frontend/hub/SheetViewState.test.ts
pmconne: Your remaining problem is that you're creating an invalid ViewAttachment. It has to provide the Id of a real drawing or spatial view - you're just throwing "0x99" in as the view Id.

full-stack-tests/core/src/backend/backend.ts
```diff
@@ -86,6 +88,118 @@ class FullStackTestIpcHandler extends IpcHandler implements FullStackTestIpc {
...
+    const drawingDefinitionModelId = DefinitionModel.insert(standaloneModel, jobSubjectId, "DrawingDefinition");
+    const drawingCategoryId = DrawingCategory.insert(standaloneModel, drawingDefinitionModelId, "DrawingCategory", new SubCategoryAppearance());
+    const sheetModelId = await insertSheet(standaloneModel, "sheet-1");
+
+    const newAttachmentProps: ViewAttachmentProps = {
+      classFullName: 'BisCore:ViewAttachment',
+      model: sheetModelId,
+      code: Code.createEmpty(),
+      jsonProperties: { displayPriority: 0},
+      view: { id: '0x99', relClassName: 'BisCore.ViewIsAttached' },
```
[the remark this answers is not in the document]
andremig-bentley: Most recent commit contains a fully functional test. Thank you so much for all the help

andremig-bentley:
> Might be a good idea to run Image Tests on this branch.

Image Tests have been run on this branch showing no concerns, so I believe this PR is ready for final approval steps.

andremig-bentley:
> The bug appears to be that we are not detecting when an attempt to load a tile tree irrecoverably fails. Did you look into fixing that?

This has been addressed, can it be resolved?

common/api/core-backend.api.md
```diff
@@ -1590,6 +1590,24 @@ export interface CreateNewWorkspaceDbVersionArgs {
     versionType: CloudSqlite.SemverIncrement;
 }
 
+// @public
+export interface CreateSheetViewDefinitionArgs {
```
pmconne, line 1594: public APIs must be documented.

core/frontend/src/SheetViewState.ts
```diff
@@ -455,6 +461,13 @@ export class SheetViewState extends ViewState2d {
     return this._viewedExtents;
   }
 
+  public getViewedExtentsFromFrustum(): Range3d {
```
pmconne, line 464: Don't call this "viewed extents" - "viewed extents" means the volume of space of interest to the view, not the current volume being displayed by the view. Consider deleting this tiny function that is only used in one place and just computing a range from the frustum in `areAllTileTreesLoaded`.

full-stack-tests/core/src/frontend/hub/SheetViewState.test.ts
```diff
@@ -169,3 +174,130 @@ describe("Sheet views (#integration)", () => {
...
+      parentDiv.style.width = wPx;
+
+      if (options?.position)
+        parentDiv.style.position = options.position;
+
+      document.body.appendChild(parentDiv);
+
+      const view = SpatialViewState.createBlank(iModel, { x: 0, y: 0, z: 0 }, { x: 1, y: 1, z: 1 });
+
+      class BlankViewport extends ScreenViewport {
```
pmconne, line 230: Looks like a whole bunch of code copied from core-frontend tests - did you look at `TestViewport.ts` in core-full-stack-tests?

full-stack-tests/core/src/frontend/hub/SheetViewState.test.ts
```diff
@@ -169,3 +174,130 @@ describe("Sheet views (#integration)", () => {
...
+      const origin = Point2d.fromJSON(attachment.placement?.origin);
+      expect(origin).not.to.be.undefined;
+
+      const bbox = Range2d.fromJSON(attachment.placement?.bbox);;
+      expect(bbox).not.to.be.undefined;
+      const attachmentRange = new Range2d(origin.x + bbox.xLow, origin.y + bbox.yLow, origin.x + bbox.xHigh, origin.y + bbox.yHigh);
+
+      // Load Tree Refs
+      for (const ref of newSheetView.getTileTreeRefs()) {
+        ref.treeOwner.load();
```
pmconne, line 281: Why should your test have to manually load each tree? waitForSceneCompletion should be indirectly causing that to happen. By manually doing it, your test no longer appears to simulate the actual bug you are trying to fix.

common/api/core-backend.api.md
andremig-bentley: docs added in latest commit

common/api/core-backend.api.md
```diff
@@ -5548,8 +5566,15 @@ export class SheetTemplate extends Document_2 {
...
 export class SheetViewDefinition extends ViewDefinition2d {
+    protected constructor(props: ViewDefinition2dProps, iModel: IModelDb);
     // (undocumented)
     static get className(): string;
+    // (undocumented)
+    static create(args: CreateSheetViewDefinitionArgs): SheetViewDefinition;
+    // (undocumented)
+    static fromJSON(props: Omit<ViewDefinition2dProps, "classFullName">, iModel: IModelDb): SheetViewDefinition;
+    // (undocumented)
+    static insert(args: CreateSheetViewDefinitionArgs): Id64String;
```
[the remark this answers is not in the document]
andremig-bentley: docs added in latest commit.

core/frontend/src/SheetViewState.ts
andremig-bentley: Viewed extents is now referred to as "displayedExtents", and this function is removed, moving the calculation to within `areAllTileTreesLoaded`.

full-stack-tests/core/src/frontend/hub/SheetViewState.test.ts
andremig-bentley: Removed copied code and switched to using `testOnScreenViewport` in latest commit.

full-stack-tests/core/src/frontend/hub/SheetViewState.test.ts
```diff
@@ -169,3 +174,130 @@ describe("Sheet views (#integration)", () => {
...
+      }
+
+      const viewport = BlankViewport.create(parentDiv, view) as BlankViewport;
+      if (undefined === options?.iModel)
+        viewport.ownedIModel = iModel;
+
+      return viewport;
+    }
+
+    it("areAllTileTreesLoaded should return true when attachments are outside of the viewed extents", async () => {
```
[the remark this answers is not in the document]
andremig-bentley: There was some difficulty in getting the attachments from the sheetview to determine whether or not the specific attachments are loaded. This has been resolved in the latest commit by adding `areAllAttachmentsLoaded` to both the `ViewAttachments` class and `SheetViewState`. Latest commit should now have the test confirming attachment loading as expected.

full-stack-tests/core/src/frontend/hub/SheetViewState.test.ts
andremig-bentley: Removed in latest commit

core/backend/src/ViewDefinition.ts
```diff
@@ -812,11 +812,64 @@ export class DrawingViewDefinition extends ViewDefinition2d {
...
+  name: string;
+  /** The Id of the Model displayed by this view */
+  baseModelId: Id64String;
+  /** The [[CategorySelector]] that this view should use */
+  categorySelectorId: Id64String;
+  /** The [[DisplayStyle2d]] that this view should use */
+  displayStyleId: Id64String;
+  /** Defines the view origin and extents */
+  range: Range2d;
+}
```
pmconne, line 833:
```suggestion
export interface CreateSheetViewDefinitionArgs {
  /** The iModel in which the sheet view will be created. */
  iModel: IModelDb;
  /** The  Id of the [[DefinitionModel]] into which the sheet view will be inserted. */
  definitionModelId: Id64String;
  /** The name to use as the view's Code value. */
  name: string;
  /** The Id of the sheet model whose contents will be displayed by this view. */
  baseModelId: Id64String;
  /** The [[CategorySelector]] that this view should use. */
  categorySelectorId: Id64String;
  /** The [[DisplayStyle2d]] that this view should use. */
  displayStyleId: Id64String;
  /** Defines the view origin and extents. */
  range: Range2d;
}
```

core/frontend/src/SheetViewState.ts
```diff
@@ -527,7 +538,21 @@ export class SheetViewState extends ViewState2d {
...
+    let displayedExtents = this._viewedExtents;
+    const frustum = this.calculateFrustum();
+    if (frustum) {
+      displayedExtents = frustum.toRange();
+    }
+
+    return super.areAllTileTreesLoaded && (!this._attachments || this._attachments.areAllTileTreesLoaded(displayedExtents));
+  }
+
+  /** Strictly for testing */
```
pmconne, line 550:
```suggestion
  /** @internal Strictly for testing */
```

full-stack-tests/core/src/frontend/hub/SheetViewState.test.ts
```diff
@@ -169,3 +172,69 @@ describe("Sheet views (#integration)", () => {
...
+        // Expect areAllTileTreesLoaded to be true when attachment is outside of the viewed extents and not yet loaded
+        expect(viewedExtents2d.intersectsRange(attachmentRange)).to.be.false;
+        expect(sheetView.areAllAttachmentsLoaded()).to.be.false;
+        expect(sheetView.areAllTileTreesLoaded).to.be.true;
+
+        //Fit view
+        vp.zoom(new Point3d(attachmentRange.xLow, attachmentRange.yLow, 0), 1)
+
+        // Expect attachments to not yet be loaded as the scene is not yet complete
+        expect(sheetView.areAllAttachmentsLoaded()).to.be.false;
```
pmconne, line 225:
```suggestion
        expect(sheetView.areAllAttachmentsLoaded()).to.be.false;
        expect(sheetView.areAllTileTreesLoaded).to.be.false;
```

pmconne (approved):

markschlosseratbentley:
👏

PR #7617, opened 2025-01-27
TileTreeReference iterables

mdastous-bentley:
Based on [this discussion](https://github.com/iTwin/itwinjs-core/pull/7604#issuecomment-2613381142) in #7604.

PR #7788, opened 2025-03-02
Revert TileTreeReference iterables (backport #7617)

pmconne:
@mdastous-bentley backported it as a preliminary to backporting support for Google Maps, but it appears we will not be backporting Google Maps to 4.11 after all. The iterables PR introduced breaking changes to @internal APIs and deprecates some @public APIs. We don't want to disrupt the APIs unnecessarily at the end of the 4.x development period.


bbastings (approved):

pmconne:
> it appears we will not be backporting Google Maps to 4.11

If this changes, just revert this PR.

PR #7805, opened 2025-03-06
Add general guidelines for the modeling of data as properties in BIS

diegoalexdiaz:


docs/bis/guide/fundamentals/attribution-guidelines.md
```diff
@@ -0,0 +1,71 @@
+# Attribution: Guidelines
```
ColinKerr, line 1: When I read the title of this page I thought 'assigning cause or credit' rather than 'defining a quality or characteristic'.  I know it can mean both but I associate it with the first definition more than the second.  Do we have a better way to describe this?

diegoalexdiaz: Getting inspiration from the title of this PR... what about "Modeling data as Attribution: Guidelines"

docs/bis/guide/fundamentals/attribution-guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+
+Note, however, that there may be attributions whose applicability dependend on the state of a more general concept on a specific Entity. Such attributions shall be implemented via a class-hierarchy based on `Element-Aspects`(./elementaspect-fundamentals.md). If the overall concept is considered *essential* for the associated Entity, the target multiplicity of its associated `ElementOwnsUniqueAspect` or `ElementOwnsMultiAspect` relationship shall be set accordingly (i.e. 1..1 or 1..* respectively). See [Sub-Properties and ElementAspects](./elementaspect-fundamentals.md#sub-properties-and-elementaspects) for more information and an example of this situation.
+
+The default behavior of all UX controls in iTwin.js is to show First-class properties as attribution of their associated Element-class, unless they are selectively hidden by using the `HiddenProperty` or `HiddenClass` *Custom-Attributes*.
```
ColinKerr, line 19: I think it would be better to keep this doc neutral and just say they should be shown unless the hidden CAs are applied

docs/bis/guide/fundamentals/attribution-guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Introduction
+
+Not every attribute of an Entity needs to be persisted into a BIS repository. Some attributes can be derived from persisted data, therefore saving them in a BIS repository turns derived attributes into effectively cached data that needs to be kept in-sync as the more fundamental attributes change. This situation effectively applies to iTwin Native apps that directly store their data in BIS repositories.
+
+## Persisted Attributes
+
+The following sections focus on attributes that are meant to be persisted in a BIS repository.
+
+### First-class Attributes
```
ColinKerr, line 13: We mention perspectives but I worry that it will be missed or overlooked by someone not familiar with modeling perspectives.  I'm reading this from the perspective of someone who's trying to figure out where their properties should go.  From that persona's point of view mentioning that the property might be better suited for another perspective and linking to more docs about perspectives would round out the options of where to look when adding a new property

docs/bis/guide/fundamentals/attribution-guidelines.md
ColinKerr: 'attribution' feels out of place in the context of the rest of the BIS docs because we generally talk about 'properties' instead of 'attributes'.  Searching the existing docs I only saw one or two instances where we used attributes for business properties.  All other uses of attribute were in EC for schema/schema items or Custom Attributes

diegoalexdiaz: I went with 'attribution' here because the article covers "Json data" - so at that point the term "property" felt more like the implementation of an 'attribution' in terms of either a first-class or an aspect ECProperty.

If we keep that mindset (abstract: attribution, implementation: property), then I should probably revise the entire guide to align to it.

ColinKerr: Attribution still doesn't seem like the right term even if we want to use 'attribute' over 'property'

ColinKerr: JSON Schemas used the term 'properties' similar to ECSchema: https://json-schema.org/learn/getting-started-step-by-step#define-properties

docs/bis/guide/fundamentals/properties-guidelines.md
```diff
@@ -0,0 +1,71 @@
+# Properties: Guidelines
+
+BIS offers multiple ways to capture properties of Entities being modeled. It is the job of the schema designer to choose the most appropiate strategy to follow in every case. This article aims to provide general guidelines for the choosing among those strategies.
+
+## Introduction
+
+Not every property of an Entity needs to be persisted into a BIS repository. Some properties can be derived from persisted data, therefore saving them in a BIS repository turns derived properties into effectively cached data that needs to be kept in-sync as the more fundamental properties change. This situation effectively applies to iTwin Native apps that directly store their data in BIS repositories.
```
ColinKerr, line 7: effectively seems like the wrong word here
```suggestion
Not every property of an Entity needs to be persisted into a BIS repository. Some properties can be derived from persisted data, therefore saving them in a BIS repository turns derived properties into effectively cached data that needs to be kept in-sync as the more fundamental properties change. This situation often applies to iTwin Native apps that directly store their data in BIS repositories.
```

docs/bis/guide/fundamentals/properties-guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+
+### First-class Properties
+
+A property of an Entity can be implemented as an [ECProperty](../../ec/ec-property.md) directly defined on the [Element-class](./data-classification.md#element-class) that represents it - also known as a `First-class property`. This is the strategy recommended to follow for properties considered as *essential* according to the semantics and [modeling perspective](../data-organization/modeling-perspectives.md) captured by the element-class modeling the associated Entity.
+
+Note, however, that there may be properties whose applicability dependend on the state of a more general concept on a specific Entity. Such properties shall be implemented via a class-hierarchy based on `Element-Aspects`(./elementaspect-fundamentals.md). If the overall concept is considered *essential* for the associated Entity, the target multiplicity of its associated `ElementOwnsUniqueAspect` or `ElementOwnsMultiAspect` relationship shall be set accordingly (i.e. 1..1 or 1..* respectively). See [Sub-Properties and ElementAspects](./elementaspect-fundamentals.md#sub-properties-and-elementaspects) for more information and an example of this situation.
```
ColinKerr, line 17: It's odd that this section talking about element aspects in this section instead of in `Optional Properties`.  It's also odd that it's so duplicative with the page on element aspects.  Discussion of mandatory aspects makes more sense as part of First-class Properties than as a part of Optional Properties but maybe the buckets you have defined are wrong.  `First-class Properties` and `Optional Properties` are much different buckets than `Properties in Link-table Relationships` and `JSON data`.  The first two are about the properties and the second two are about the persistence of the properties.

Maybe the correct way to handle this is to make the sections `Element Properties` and `Aspect Properties`.  in one way there is more BIS jargon to describe but I'd argue that `First-class` and `Optional` are just different jargon.  For example optional properties implicitly make it sound like the other option are mandatory and must be set.  That's not true, it's only true that the property definition is always there.

docs/bis/guide/fundamentals/elementaspect-fundamentals.md
```diff
@@ -38,11 +38,13 @@ For an ElementUniqueAspect, the combination of subclass and the Element identity
...
 
 ![ElementAspect hierarchy example](../media/elementaspect-shapeaspect-hierarchy.png)
 
-In this case, the `ShapeAspect` base-class should be used as the *key-class* to uniquely identify instances of the concrete aspects of this class-hierarchy. That is needed in this case in order to ensure an object only owns one aspect of any of the three concrete classes in the `ShapeAspect` hierarchy at any given time.
+In this case, the `ShapeAspect` base-class should be used as the *key-class* to uniquely identify instances of the concrete aspects of this class-hierarchy. That is needed in this case in order to ensure an object only owns one aspect of any of the three concrete classes in the `ShapeAspect` hierarchy at any given time. Furthermore, if the `Shape` attribution modeled by this class-hierarchy is considered essential (or [`First-class`](./properties-guidelines.md#first-class-attributes)) to the Entity whose Element-class representation owns the `Element-Aspect`, the target multiplicity of its associated `ElementOwnsUniqueAspect` or `ElementOwnsMultiAspect` relationship shall be set accordingly (i.e. 1..1 or 1..* respectively) in order to advertise that it is not optional.
```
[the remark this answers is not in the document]
diegoalexdiaz: That section is meant to be the one addressing "mandatory aspects" - so it is not tacking it on its last sentence, but explaining when it is needed. Making it hopefully clearer by renaming the section.

docs/bis/guide/fundamentals/properties-guidelines.md
diegoalexdiaz: I agree with the confusion that the word "Optional" brings. However, I prefer not to let BIS concepts define the buckets (i.e. Element vs. Aspect properties). I went with "Essential" and "Non-essential" **Element Properties**, which more clearly separate themselves from Relationship properties and ad-hoc JSON.

docs/bis/guide/fundamentals/properties-guidelines.md
```diff
@@ -0,0 +1,71 @@
+# Properties: Guidelines
+
+BIS offers multiple ways to capture properties of Entities being modeled. It is the job of the schema designer to choose the most appropiate strategy to follow in every case. This article aims to provide general guidelines for the choosing among those strategies.
+
+## Introduction
+
+Not every property of an Entity needs to be persisted into a BIS repository. Some properties can be derived from persisted data, therefore saving them in a BIS repository turns derived properties into effectively cached data that needs to be kept in-sync as the more fundamental properties change. This situation often applies to iTwin Native apps that directly store their data in BIS repositories.
```
MichaelSwigerAtBentley: This introduction focuses on derived properties, could we expand it to also mention persisted properties explicitly and a thesis statement on why/when one might want to pressist a property?

MichaelSwigerAtBentley: "This situation often applies to iTwin Native apps that directly store their data in BIS repositories."

What situation is 'this' referring to?

docs/bis/guide/fundamentals/properties-guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Persisted Properties
```
MichaelSwigerAtBentley, line 13: Could we add examples of what might be consisted an essential property?

docs/bis/guide/fundamentals/properties-guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+
+Link-table relationship properties shall not be included in any generic UX control by default.
+
+### JSON data
+
+Instance-specific *ad hoc* data on an Element can be captured in JSON format in the [`JsonProperties`](./element-fundamentals.md#jsonproperties) property of an `Element`. See [`Advantages and Disadvantages of JsonProperties`](./element-fundamentals.md#advantages-and-disadvantages-of-jsonproperties) for additional information on this topic.
+
+More carefully planned data can also be stored in JSON format. This is the case of [Authoring-focused attributes](#authoring-focused-attributes) in some cases, especially because such data typically only has one reader: the iTwin Native application that depends on it. In that case, it is recommended that a property different from the base `JsonProperties` is used to store it. Such kind of properties need to be defined as *string* [ECProperties](../../ec/ec-property.md) whose *extendedType* is set to *Json*.
+
+JSON-based properties shall not included in any generic UX control by default.
```
MichaelSwigerAtBentley:
```suggestion
JSON-based properties shall not be included in any generic UX control by default.
```

docs/bis/guide/fundamentals/properties-guidelines.md
```diff
@@ -0,0 +1,71 @@
...
+
+Properties of an Entity considered non-essential are implemented as [EC Properties](../../ec/ec-property.md) of an [Element-Aspect](./elementaspect-fundamentals.md). An `Element-Aspect` is a set of properties that can be attached to an `Element` instance.
+
+Element-Aspect properties shall be shown in UX controls as attribution of their associated Element-instance, unless they are selectively hidden by using the `HiddenProperty` or `HiddenClass` *Custom-Attributes*.
+
+### Properties in Link-table Relationships
+
+[Link-table](./relationship-fundamentals.md#link-table) relationships can contain properties if the associated property does not apply separately to any of the Entities involved, but it only applies to them when they are associated by the relationship instance.
+
+As an example, the base class of `Link-table` relationships - `ElementRefersToElements` - contains the `MemberPriority` property to enable data-writers to specify an order for the instances at the target-end of the relationship with respect to a common source-instance.
```
tigistmebrate11, line 39: I would love to see more example sections like `MemberPriority` under all of the other different property sections to help as a reference for the reader.

rschili (approved):
I am a worried that the properties guidelines are quite dense. A decision flowchart or table could make this easier to digest.

Maybe something like this? Approving anyways, but this could be useful.


## Quick-reference: Choosing a Property Strategy

| # | Ask yourself... | If yes → use this strategy | BIS mechanism |
|---|---|---|---|
| 1 | Can this property be computed from other already-persisted data? | **Derived Property** | [Calculated Properties](https://www.itwinjs.org/presentation/content/calculatedpropertiesspecification/) (ECExpression), [Custom Renderer](https://www.itwinjs.org/reference/presentation-common/presentationrules/customrendererspecification/), or [ECView](../../../learning/ECSqlReference/Views.md) (ECSQL) |
| 2 | Does the property only exist when two entities are related - not applicable to either one independently? | **Link-table Relationship Property** | ECProperty on a subclass of `ElementRefersToElements` |
| 3 | Is this ad-hoc, instance-specific data with no fixed schema, or authoring data specific to a single iTwin Native app? | **JSON data** | `JsonProperties` for truly ad-hoc data; a dedicated `string` ECProperty with `extendedType=Json` for structured app-specific data |
| 4 | Is the property essential to correctly describe the entity in this modeling perspective, and always applicable? | **Essential Element Property** | ECProperty directly on the Element class |
| 5 | Is the property essential, but its applicability depends on another essential property (e.g., _Diameter_ and _Height_ depend on _Shape_)? | **Mandatory ElementAspect** | Class hierarchy rooted at a key-class Aspect; `ElementOwnsUniqueAspect`/`ElementOwnsMultiAspect` with target multiplicity `1..1` or `1..*` |
| 6 | Is the property optional additional information - not required to correctly describe the entity in its modeling perspective? | **Non-essential Element Property** | ECProperty on a non-mandatory ElementAspect |

> Note: evaluate rows 1-3 first. If none apply, use rows 4-6 to decide where on the element the property belongs.

docs/bis/guide/fundamentals/properties-guidelines.md
```diff
@@ -0,0 +1,77 @@
+# Properties: Guidelines
+
+## Introduction
+
+BIS offers multiple ways to capture properties of Entities being modeled. It is the job of the schema designer to choose the most appropiate strategy to follow in every case. This article aims to provide general guidelines for the choosing among those strategies.
+
+Properties of Entities model their attributes relevant in light of a particular modeling perspective. Examples of properties include the _inner diameter_ of a _Pipe_ (Physical Perspective), the _roughness coefficient_ of a _Pipe_ (Hydraulic Analysis perspective) or the _tag number_ of a _Piping Segment_ (Functional Perspective).
+
+Note that while a schema designer is responsible to model all relevant properties of an Entity, not every property needs to be persisted in a BIS repository. Some properties can be derived from other information already persisted. This is specially true for iTwin Native apps offering data-editing functionality. Persisting data about properties that could be derived turns them into, effectively, caches that need to be kept in-sync as the more fundamental information change. In general, caches add significant implementation complexity.
+
```
hl662, line 10: Can we add a quick reference near the top here — either a small table or a flowchart — that answers something like:

1. computed from other persisted data → derived  
2. belongs to the association itself → relationship property  
3. app-private / ad hoc / weakly interoperable → JSON  
4. intrinsic + central to the element semantics → element property  
5. grouped / conditional / secondary / optionally applicable → element aspect

I think that would make the rest of the page much easier to follow, this little section can add as a cheat sheet for people who are revisiting this documentation

docs/bis/guide/fundamentals/properties-guidelines.md
```diff
@@ -0,0 +1,77 @@
...
+More carefully planned data can also be stored in JSON format. This is the case of [Authoring-focused attributes](#authoring-focused-attributes) in some cases, especially because such data typically only has one reader: the iTwin Native application that depends on it. In that case, it is recommended that a property different from the base `JsonProperties` is used to store it. Such kind of properties need to be defined as *string* [ECProperties](../../ec/ec-property.md) whose *extendedType* is set to *Json*. That strategy saves the data-writer application from having to define a namespace to avoid data collisions if it targeted the `JsonProperties` ECProperty instead.
+
+JSON-based properties shall not be included in any generic UX control by default.
+
+## Derived Properties
+
+There are two strategies that a schema designer can plan for in order to avoid storing derived data into a BIS repository:
```
hl662, line 53: Before diving into solutions and implementation of derived data, can we first explicitly answer: "when should I choose derived instead of persisted?" I guess this ties into the 'cheat sheet' I mentioned in my earlier PR comment

docs/bis/guide/fundamentals/properties-guidelines.md
diegoalexdiaz: Thanks. Added a summary table at the end of the article. That way, first-time readers won't get confused by seeing terms in a table that haven't been explained yet.

docs/bis/guide/fundamentals/properties-guidelines.md
diegoalexdiaz: Thanks. Added a paragraph addressing that question.

MichaelSwigerAtBentley (approved):

PR #7820, opened 2025-03-10
Add FormatSet interface to ecschema-metadata

### LINKED ISSUE #7808: Add FormatSet interface to ecschema-metadata as a beta API
The Pull Request will be the platform to solicit more opinions on the interface. Stand alone PR just dedicated to exposing this as an API, and also adding documentation on PFS.

hl662:
Closes #7808 
- Add section in learning docs on persistence
- Add entry to NextVer.md
- Add FormatSet interface as a beta API to ecschema-metadata


docs/learning/quantity/index.md
```diff
@@ -229,3 +231,62 @@ The quantity formatter supports parsing mathematical operations. The operation i
...
+
+```Typescript
+{
+  id: "metric",
+  name: "Metric",
+  // TODO: Should the unit system be included? Then, the id and name above doesn't have to conform to a unit system.
+  unitSystem: "metric",
+  formats: [ //  TODO: Array, or <key,FormatProps> to preserve uniqueness?
+    {
+      id: "Length", // Should there be more meaning/rules to the Id? If Koq, then should be <Schema>.<Name>
```
hl662, line 258: @diegoalexdiaz should these be tied to a KoQ's fullName?

diegoalexdiaz: I'm catching up with the implementation of _PresentationFormatSet_ so I'm not aware of all its details. However, its description sounds like it came from the perception that the concept of a KoQ only applies to cases defined in a schema so some parallel approach was needed for the cases where no schema is involved.

IMO, that perception is incorrect. I believe the Civil Review Tools did just that - create an in-memory KoQ by code associated to the formatting needed.

ColinKerr: I agree we should be looking at the possibility of using KOQ for everything instead of having two sources of keys.  Having an undescribed id field is confusing at both levels, I'd prefer we don't use the name 'id' if a more descriptive name could be used, if not then we should describe what the id means.

core/quantity/src/Formatter/Interfaces.ts
```diff
@@ -94,3 +94,18 @@ export interface CloneOptions {
...
 }
+
+/**
+ * This interface defines properties necessary to support persistence of a set of formats.
+ * @beta
+ */
+export interface PresentationFormatSet {
+  id: string;
+  name: string;
+  unitSystem: UnitSystemKey; // ????? Do we want this? Maybe this is an application level thing instead of in core?
```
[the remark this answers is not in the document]
hl662: I haven't thought much about whether a PFS needs a id/name yet, it was just assumed that there should be a way to uniquely identify a PFS, if there's an application that makes use of multiple/switching between different PFS's at runtime.

I feel the same with `unitSystem` - functionally, you could 'override' anything to be part of a unit System. A unit system for metric, a user could decide to use feet and inches temporarily. I'll remove it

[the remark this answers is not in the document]
hl662: I've removed references to unitSystem

docs/learning/quantity/index.md
hl662: I've updated the `formats` from an array to a key,value map, and updated the documentation to describe we are expecting a valid ECName for the map's key. That's just the expectation though, in reality anyone can put anything as the name, but of course that format won't be picked up by tools that use formatting, so it functionally wouldn't matter

core/quantity/src/Formatter/Interfaces.ts
```diff
@@ -94,3 +94,18 @@ export interface CloneOptions {
...
+  formats: (FormatProps & { // Similarities with SchemaItemProps
```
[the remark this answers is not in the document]
rschili: yes that makes sense.

docs/learning/quantity/index.md
```diff
@@ -305,3 +307,51 @@ If a Format uses a spacer that conflicts with one of the operators above, additi
...
+        "spacer": "",
+        "units": [{ "label": "m", "name": "Units.M" }]
+      },
+      "formatTraits": ["keepSingleZero", "showUnitLabel"],
+      "precision": 4,
+      "type": "Decimal",
+      "decimalSeparator": "."
+    },
+    "AecUnits.Angle": {
+      "id": "Angle",
```
ColinKerr, line 343: what us the purpose of this id property?

hl662: It represents a unique identifier for the format, separate from the identifier of a KindOfQuantity. This id shouldn't be user facing, and is relevant when working with formats retrieved from schemas, for example when overriding  the property of the format in session/runtime

hl662: in retrospect, I should not have named the id `angle` as an example, it should be more descriptive/ tells you it comes from a schema perhaps

core/quantity/src/Formatter/Interfaces.ts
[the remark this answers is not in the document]
hl662: After chatting with @rschili we clarified that it would make sense for SchemaItemFormatProps to extend from interfaces defined in core-quantity, not the other way around. Consolidating interfaces used between ecschema-metadata and core-quantity is a larger endeavor that'll be separate from this PR.

rschili (approved):

core/ecschema-metadata/src/Deserialization/JsonProps.ts
```diff
@@ -267,6 +267,16 @@ export interface ConstantProps extends SchemaItemProps {
...
 
+/**
+ * This interface defines properties necessary to support persistence of a set of formats.
+ * @beta
+ */
+export interface FormatSet {
+  name: string;
+  label: string;
+  formats: { [kindOfQuantityId: string]: SchemaItemFormatProps }
+}
```
grigasp, line 278: Assuming we'll have a number of these FormatSets coming from different sources, should we have some "priority" attribute to know which ones takes precedence?

I believe we'd have the following order: user format sets > imodel format sets > itwin format sets > organization format sets > app default format sets? > koqs in schema.

grigasp: sounds good, just wanted to make sure this is thought through

ColinKerr (approved):

PR #7822, opened 2025-03-10
Augment JSON format for `Sphere` and `Cone`

### LINKED ISSUE #7749: loadProps return empty geometry data for ellipsoilds with three different semi-axes
**Describe the bug**
I use **Sphere.createFromAxesAndScales** to create an ellipsoild with three different semi-axes (e.g. radiusX=1, radiusY=2, radiusZ=3). After that I use the **loadProps** api (wantGeometry: true) to retrieve its geometry,  it returns empty geometry data(as the following image shown). Also I find that it's not possible to create such ellipsoild using json format.

Image

### LINKED ISSUE #7761: ConeProps is not sufficient for describing irregular cones​
**Describe the bug**
Currently we are able to create irregular cones whose start and end sections are full ellipses (as the image shown). However, the **ConeProps** data returned by loadProps is not sufficient for describing irregular cones because it contains only one startRadius and endRadius, and the x/y vectors are both normalized.

Image

**To Reproduce**
We can use the following code to create an irregular cone:
`
Cone.createBaseAndTarget(Point3d.create(0.3, 0, 0), Point3d.create(0.3, 0, 0.1), Vector3d.create(0.1, 0.1, 0), Vector3d.create(-0.2, 0.2, 0), 0.1, 0.2, true)
`

dassaf4:
E.g., to allow for elliptical cross sections.

These `SolidPrimitive`s previously specified their axes in JSON format with an `AxesProps` interpreted during deserialization as a rigid Transform.

Now, if a `Sphere` or `Cone` has a non-rigid local frame, it is serialized to JSON with an _additional_ property `AxesProps.xyzVectors`. This is a new optional 3-vector AxesProps which JSON deserialization now prefers to the rigid variants, allowing for exact Transform reproduction.

New JSON representations of `Sphere` and `Cone` are backwards compatible with older readers, and old JSON representations deserialize the same in newer readers. The FlatBuffer representation for `Sphere` and `Cone` is unchanged, as it already encodes the full local->world Transform.

Spheres with non-orthogonal axes and cones with non-circular cross sections have limited portability. Constructors are provided on both classes to ensure portability to e.g., DGN.

Fixes #7749
Fixes #7761


bbastings:
> Spheres with non-orthogonal axes and cones with non-circular cross sections

Who asked for this? Probably won't convert these to Parasolid correctly, etc.?

bbastings (approved):

saeeedtorabi (approved):

PR #7823, opened 2025-03-11
WIP - Improve replacements for deprecated schema metadata methods in backend

RohitPtnkr1996:
** WIP **


core/ecschema-metadata/src/Context.ts
```diff
@@ -449,6 +427,35 @@ export class SchemaContext implements ISchemaItemLocater {
...
+   *
+   * @example
+   * ```typescript
+   * await schemaContext.getSchemaItemByName("BisCore", "Element");
+   * await schemaContext.getSchemaItemByName("BisCore", "Element", EntityClass);
+   * await schemaContext.getSchemaItemByName("Units", "THOUSAND_SQ_FT");
+   * await schemaContext.getSchemaItemByName("Units", "THOUSAND_SQ_FT", Unit);
+   * ```
+   */
+  public async getSchemaItemByName(schemaName: string, itemName: string): Promise<SchemaItem | undefined>
```
rschili, line 446: I have been thinking about this.
Even though I despise the typescript method overload syntax, I believe we should try and make everything into a single method, and give it a very short name, so it's easiest for callers.

Something like this should work:

```ts
  public async getItem(schemaItemKey: SchemaItemKey): Promise<SchemaItem | undefined>
  public async getItem<T extends typeof SchemaItem>(schemaItemKey: SchemaItemKey, itemConstructor: T): Promise<InstanceType<T> | undefined>
  public async getItem(schemaName: string, itemName: string): Promise<SchemaItem | undefined>
  public async getItem<T extends typeof SchemaItem>(schemaName: string, itemName: string, itemConstructor: T): Promise<InstanceType<T> | undefined>
  public async getItem<T extends typeof SchemaItem>(nameOrKey: SchemaItemKey | string, nameOrCtor?: T | string, ctor?: T): Promise<SchemaItem | InstanceType<T> | undefined> {
    let schema: Schema | undefined;
    if(typeof nameOrKey === "string") {
      schema = await this.getSchema(new SchemaKey(nameOrKey), SchemaMatchType.Latest);
    } else {
      schema = await this.getSchema(nameOrKey.schemaKey, SchemaMatchType.Latest);
    }

    if (undefined === schema)
      return undefined;

    if(typeof nameOrCtor === "string") {
      if(ctor === undefined)
        return schema.getItem(nameOrCtor);
      return schema.getItem(nameOrCtor, ctor);
    } else {
      if (typeof nameOrKey !== "string") {
        if (nameOrCtor === undefined)
          return schema.getItem(nameOrKey.name);
        return schema.getItem(nameOrKey.name, nameOrCtor);
      }
    }

    return undefined; // or throw error? This should never happen
  }
```

core/ecschema-metadata/src/Context.ts
```diff
@@ -489,4 +525,73 @@ export class SchemaContext implements ISchemaItemLocater {
...
+   * @throws [[ECObjectsError]] if the metadata cannot be found or loaded.
+   *
+   * @example
+   * ```typescript
+   * const entityClass = schemaContext.getSchemaItemMetaData("BisCore:Element", EntityClass);
+   * const customAttributeClass = schemaContext.getSchemaItemMetaData("BisCore:SchemaHasBehavior", CustomAttributeClass);
+   * const relationshipClass = schemaContext.getSchemaItemMetaData("BisCore:ElementOwnsChildElements", RelationshipClass);
+   * ```
+   */
+  public getSchemaItemMetaData<T extends typeof SchemaItem>(classFullName: string, itemConstructor: T): InstanceType<T> {
```
rschili, line 544: Hmm, I do not like this one. It feels like clutter.
I'd rather not have a fullname overload and instead have a nice public method which splits a fullname into two strings.
So instead of
`x.getSchemaItemMetaData("myschema:myclass");`
people could call
`x.getItem(...parseFullName("mySchema:myClass");`

Assuming the parseFullName method looks like this:
```ts
  public parseFullName(input: string): [string, string] {
    const [first, second] = input.split(":");
    return [first, second];
  }
The parseFullName is just a sketch. The real one needs to support colon, dot and make sure there are only 2 results...
```
Or if we absolutely need to stick with this, better call it getItemByFullName(). Everything here is MetaData, so that previous name is ambiguous.

rschili (approved):

PR #7850, opened 2025-03-17
docs: mislabeled RotationalSweep

ben-polinsky:
Fix docs which [incorrectly refers to](https://www.itwinjs.org/reference/core-geometry/solid/rotationalsweep/) RotationalSweep as LinearSweep.


bbastings (approved):

dassaf4 (approved):
If we are supposed to use "-" instead of "*" in TypeScript documentation to indicate a bullet list, then we're going to need to change a lot of code. This is kind of a thing in core-geometry.

Issue #7807, opened 2025-03-06
Switch the typing used to retrieve formatProps for Quantity Formatting


hl662:
[QuantityType](https://github.com/iTwin/itwinjs-core/blob/c84df3a7fd6b0cdeaa1068920d1a913299d967f2/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L26) is a numeric enum exposed in core-frontend. A concept like formatting quantities should not have to be dependent on a frontend library. Let's make this a **stack-agnostic** concept. We also should not be doing [whatever this is](https://github.com/iTwin/itwinjs-core/blob/c84df3a7fd6b0cdeaa1068920d1a913299d967f2/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L51C13-L51C28), a union between numbers and strings.  Below are some of the things to tackle relevant to this:
1. Deprecate QuantityType interface in core-frontend
~2. Add a new `QuantityType` **string** enum in core-quantity.~
~3. The 2 combined above should warrant deprecating some other types/[functions](https://github.com/iTwin/itwinjs-core/blob/c84df3a7fd6b0cdeaa1068920d1a913299d967f2/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L96) in core-frontend's QuantityFormatter because they aren't needed anymore. ~

Update May 6 2025:
We will move towards replacing QuantityType with `KindOfQuantity`, documenting a list of replacements for each QuantityType enumeration, made available on the itwinjs.org site. Tools and other frontend components will prioritize retrieving formatProps associated with that KindOfQuantity via a `formatsProvider`, falling back to using a QuantityType if a `formatsProvider` isn't provided. This fallback will in place through iTwin 5 lifecycle to provide users time to adopt, before deprecating QuantityType in 6.x.

aruniverse:
@ColinKerr  @rschili do you agree with adding a `QuantityType` to core-quantity like the one found in core-frontend? Is this the right way to reconcile the concepts in both pkgs? Is there a better alternative?

rschili:
QuantityType seems to have a large overlap with Phenomena which are dynamically defined in the units schema.
https://github.com/iTwin/bis-schemas/blob/master/System/Units.ecschema.xml

| QuantityType      | Matching Phenomenon |
|-------------------|---------------------|
| Length            | LENGTH              |
| Angle             | ANGLE               |
| Area              | AREA                |
| Volume            | VOLUME              |
[...]

PR #7858, opened 2025-03-18
Add FormatsProvider, SchemaFormatsProvider to iTwin.js 

### LINKED ISSUE #7811: Add a FormatsProvider interface to ecschema-metadata
To update the quantity formatting of values displayed on a frontend app, currently you would have to[ override formats](https://github.com/iTwin/itwinjs-core/blob/ab90bd29ed274298cfa3a0b689f01f1d455f515d/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L466) used within the QuantityFormatter in core-frontend. Then, tools and frontend components that use formats through the QuantityFormatter would be able to reflect the changed formats. However there are cases where some components might not use formats via QuantityFormatter.

From @grigasp:

> Regarding quantity format overrides... For some reason the first thing I thought of was unit system, which should work. But format overrides is a different beast - there are 2 ways to format property values:
1.Using quantity formatter on IModelApp.quantityFormatter
2.Using formats that we get from schema
Since all properties in property grid are based on schema, we use schema-based formatting, which means we don't take into account any overrides set on IModelApp.quantityFormatter. 
When @saskliutas started working on editors, we had some discussions with schema guys about having a single API for finding unit formats, taking into account: ECProperty kind of quantity, default formats, format overrides, etc.

@diegoalexdiaz on formats from schemas:
> During schema development, our mindset regarding format settings captured in KoQs defined at schema-level is that those are **"defaults"** / fallback settings. The decision at schema-level regarding KoQs is "what properties are formatted together". Our expectation all along has been that users (not getting into what personas in this statement) will be able to change (override) those default format-settings per KoQ without having to change any schema.

There may be a need for a FormatsProvider, just like a [UnitsProvider](https://github.com/iTwin/itwinjs-core/blob/ab90bd29ed274298cfa3a0b689f01f1d455f515d/core/quantity/src/Interfaces.ts#L102), that let's users decide what type of formats to use, and allow a way to replace formats used during runtime, separate from what's currently done in QuantityFormatter with format overrides. 

One goal this interface would achieve, is that formats loaded from a FormatsProvider would be independent from format overrides, hence should not be affected when [clearing format overrides](https://github.com/iTwin/itwinjs-core/blob/ab90bd29ed274298cfa3a0b689f01f1d455f515d/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L496)

TBD on methods to expose in this new interface

hl662:
Closes #7811 
- Add FormatsProvider interface
- Add MutableFormatsProvider interface
- Add SchemaFormatsProvider class
- Add documentation

TODOs:
- [x] More tests
- [x] Figure out events
- [x] Flesh out CRUD functionality of the provider.
- ~~[ ] Integrate into DTA~~ Will implement in separate PR utilizing a FormatSet introduced in #7820 
- [x] Update NextVer.md
- [x] Update quantity docs


core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,110 @@
...
+      this._context = contextOrLocater;
+    } else {
+      this._context = new SchemaContext();
+      this._context.addLocater(contextOrLocater);
+    }
+  }
+
+  /**
+   * Try retrieving the format from schemas. If the format is not part of the base Formats schema, but from a KindOfQuantity, the format will be constructed
+   * from the default presentation format of that KindOfQuantity.
```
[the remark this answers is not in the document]
grigasp: We're using `IModelApp.quantityFormatter.activeUnitSystem` to determine active unit system and `IModelApp.quantityFormatter.onActiveFormattingUnitSystemChanged` event to listen to its changes.

Our implementation to find format from koq based on unit system is here: https://github.com/iTwin/presentation/blob/f38653e109bdc2c3c1fdf505b27026670e656dd2/packages/core-interop/src/core-interop/Formatting.ts#L121.

pmconne (approved):

core/ecschema-metadata/src/SchemaFormatsProvider.ts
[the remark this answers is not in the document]
hl662: I've ported the logic you shared into my implementation - I wrote it so unitSystem is an optional argument, and if it's not provided, when getting a default presentation format from a KoQ, we'll just get the first entry from the array of presentation formats. We're just handling fallbacks, the formats from an EC KoQ is just a default, and our vision for applications is they eventually won't retrieve formats from the schema, but through the cache of the formatsProvider.

core/ecschema-metadata/src/Interfaces.ts
```diff
@@ -94,3 +96,16 @@ export interface HasMixins {
...
+ */
+export interface FormatsProvider {
+  getFormat(name: string, unitSystem?: string): Promise<SchemaItemFormatProps | undefined>;
+  addFormat(name: string, formatProps: SchemaItemFormatProps, unitSystem?: string): Promise<void>;
+  removeFormat(name: string): Promise<void>;
+
+  onFormatUpdated: BeUiEvent<string>;
+  onFormatRemoved: BeUiEvent<string>;
+  onCacheCleared: BeUiEvent<void>;
+}
```
grigasp, line 111: Should a "provider" type of interface define "add", "remove" methods and events like "on removed" and "on cache cleared"? It looks like these members are implementation detail. Can't a provider be read-only without ability to add / remove (in fact, it seems `SchemaFormatsProvider` should be read-only)? What if provider doesn't have a cache?

IMO, it's enough to have a "get" method and one of these:
- `onFormatChanged` event for consumers to know that specific format has changed,
- `onFormatsChanged` event for consumers to know that any of the provided formats changed.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  /**
+   * Retrieves a Format from the cache or from the schema. If retrieving from the schema, and the format is part of a KindOfQuantity,
+   * an optional UnitSystemKey can be provided to get the first presentation format in the KindOfQuantity that matches the unit system.
+   * @param id The full name of the Format.
+   * @returns
+   */
+  public async getFormat(id: string, unitSystem?: UnitSystemKey): Promise<SchemaItemFormatProps | undefined> {
+    if (this._formatCache.has(id)) {
```
grigasp, line 107: I think it would be great to normalize the given id for caching purposes. You could be given all kinds of different variations of the same thing: `MySchema.MyKoQ`, `MySchema:MyKoQ`, `myschema.mykoq` and so on.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+      return this._formatCache.get(id);
+    }
+    return this.getFormatFromSchema(id, unitSystem);
+  }
+
+  /**
+   * Adds a format to the provider's cache. If the cache already has a format with the same name, override the format.
+   */
+  public async addFormat(name: string, formatProps: SchemaItemFormatProps): Promise<void> {
```
grigasp, line 116: `getFormat` takes and `id` argument, but `addFormat` and `removeFormat` take `name`. I think they need to be made consistent.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (!defaultFormat) {
+      return undefined;
+    }
+    return getFormatProps(defaultFormat);
+  }
+
+  /**
+   * Retrieves a Format from the cache or from the schema. If retrieving from the schema, and the format is part of a KindOfQuantity,
+   * an optional UnitSystemKey can be provided to get the first presentation format in the KindOfQuantity that matches the unit system.
+   * @param id The full name of the Format.
```
grigasp, line 103: looks like it could also be full name of `KindOfQuantity`?

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+   * @returns
+   */
+  public async getFormat(id: string, unitSystem?: UnitSystemKey): Promise<SchemaItemFormatProps | undefined> {
```
grigasp, line 106: This method takes two arguments, but the second one is only used in special situations - when the `id` is not found in cache and means full name of `KindOfQuantity`. This is a sign of less than ideal API. Maybe the method could take a props in the following format?

```ts
getFormatFromSchema(props: {
  formatName: string;
} | {
  koqName: string;
  unitSystem: UnitSystemKey;
})
```

On the other hand, what meaning does unit system have in format overrides' case? Maybe unit system shouldn't be part of the "get" call at all, and instead should be a member of the schema formats provider class? And the provider should be re-created with a different unit system setting when the active unit system changes?

core/ecschema-metadata/src/Interfaces.ts
```diff
@@ -94,3 +96,16 @@ export interface HasMixins {
   mixins: LazyLoadedMixin[];
   getMixinsSync(): Iterable<Mixin>;
 }
+
+/** This interface is implemented by a class that would provide formats for use in formatting quantities.
+ * @beta
+ */
+export interface FormatsProvider {
+  getFormat(name: string, unitSystem?: string): Promise<SchemaItemFormatProps | undefined>;
```
grigasp, line 104: Is this an interface for generic formats provider, including for formats coming from sources other than schema? If yes:
- Why does it expose `SchemaItemFormatProps`? What would non-schema based providers return?
- Why is it defined in `eschema-metadata` package? Shouldn't it be placed in `core-quantity`?

core/ecschema-metadata/src/SchemaFormatsProvider.ts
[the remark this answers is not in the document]
grigasp: > our vision for applications is they eventually won't retrieve formats from the schema, but through the cache of the formatsProvider

Would it make more sense to have a number of providers handling each source individually + one combining provider that takes a list of providers and takes care of prioritization? I think trying to squeeze everything into `SchemaFormatsProvider` class will make it overly complex, especially when you start introducing other sources of formats (e.g. how is your cache going to handle situation when different sources define the same format and you need to prioritize them?)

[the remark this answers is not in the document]
hl662: This is the extent of the SchemaFormatsProvider, and when it comes to prioritization, it should be an implementation outside of iTwin.js - How that looks is going to be an implementation detail, my opinionated first thought was multiple providers following a decorator pattern (i.e each provider would have a member instance of a lower level provider, and if their own cache or way to retrieve formats don't find anything, the outer provider calls the getFormat() of the member formatsProvider)

core/ecschema-metadata/src/Interfaces.ts
grigasp: > SchemaItemFormatProps is a union between SchemaItemProps, and FormatProps (in core-quantity). Non based schema providers would return the same interface, but without properties defined as part of SchemaItemProps. That interface's properties are all optional, and the implementation of those providers can ignore that side and still return a valid value.
> 
> So this return valid is the technical reason why it's placed in ecschema-metadata

IMO, just because you can technically omit all the properties of `SchemaItemProps`, merely having it in type definition is already a problem - it suggests that `FormatsProvider` should return schema items. In addition, that forces you to define `FormatsProvider` in `ecschema-metadata`, which in my opinion is not the right place for it.

> One other reason is the formatProps in core-quantity is incomplete compared to an ECFormat. There is nothing in the formatProps to help identify that format, no description label. A third-party user, or a first-party user, could implement their own FormatsProvider, and they might want to extend the FormatProps type to include an identifier. How would they name that identifier? If it's `formatId` or `key`, then it won't be compatible with ECFormats, which uses `name` or `fullName`. 

I'm not sure I like mixing formatting-related props defined in `FormatProps` with other attributes not related to formatting (e.g. label of the format). Ideally, I think you'd have a type `{ label: string; format: FormatProps }` or similar.

On the other hand, if you want your type to match the native `ECFormat` class -  that's also fine, and you don't need to explicitly introduce relationship with EC by returning `SchemaItemFormatProps` from your provider. You could have something like this instead:

```ts
// defined in core-quantity
interface FormatDefinition extends FormatProps {
    readonly name?: string;
    readonly label?: string;
    readonly description?: string;
}

// properties of FormatDefinition fully match SchemaItemFormatProps, so you can still do this:
const x: SchemaItemFormatProps = ...
const y FormatDefinition = x;
```

I just don't think that all format providers should be aware of schemas, and implementing a format provider should require me to have an `ecschema-metadata` dependency.

Having said that, all of this is my recommendation and by no means an obligation for you do this way. It's up to you and your team to decide what's right here.

core/ecschema-metadata/src/Interfaces.ts
hl662: this makes sense, I've modified the events naming and dropped the add/remove methods from the generic interface.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: thanks, I've updated the doccomment to mention that

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: I missed that - I've renamed the arg for getFormat to be `name`

core/ecschema-metadata/src/Interfaces.ts
grigasp: I was suggesting to add one of the two events. Why do we need both? Especially since they're not raised consistently (`onFormatChanged` is raised on add/remove and `onFormatsChanged` is called on clear).

core/ecschema-metadata/src/SchemaFormatsProvider.ts
grigasp: > because the new provider would not have the old provider's formatCache (i.e state)

I think I mentioned this elsewhere - IMO the formats' cache doesn't belong to this class. It already has a cache in the form of `SchemaContext` - we can quickly find koq and pick the right format based on unit system that is active at that time.

> When the active unit system changes, the current provider could just update it's unitSystem member, after all, it's just used in special situations...

Not sure I understand you. Do you mean we can ignore the provider returning invalid data in those special situations?

> With all this in mind, I do want to ask what happens with regards to presentation APIs. Do you imagine registering an implementation of FormatsProvider, or SchemaFormatsProvider for the different measurement tools and formatting utils of presentation? If we drop unitSystem from getFormat(), the former wouldn't understand changing its unitSystem member, and the latter would mean any application developer implemented FormatsProvider should be extended from SchemaFormatsProvider, because the application should be registering the SchemaFormatsProvider for the different presentation APIs

Here's how I see it:
- Either some core package (e.g. `core-quantities`), or the application maintains the active unit system.
- Core delivers a few `FormatProvider` implementations:
  - `SchemaFormatsProvider` loads formats from schema based on active unit system. So we need one per iModel. It also needs to know the active unit system, so whoever creates it also has to either re-create on active unit system change, or at least set the new system on existing provider. When that happens, the provider should trigger "formats changed" event.
  - Some "service formats provider(s)" load user/imodel/itwin/organization format overrides.
  - Some "combined provider" takes a prioritized list of providers and uses them to get the active format.
- App builds the "combined provider" from the sub-providers it needs. We may provide the default setup, but I can see applications wanting to inject their own application-level provider into the prioritized list.
- From here I see two paths:
  1. App maintains the provider and feeds it to any API that needs formatting functionality. This requires more work but doesn't use static APIs and is generally cleaner.
  2. App sets the provider on some global, e.g. `IModelApp`.

Presentation library (or any other component) gets the format provider either from app as parameter or global `IModelApp` and simply calls `getFormat("xxx.yyy")`. All format priorities and unit systems are handled in the underlying providers. The component can also subscribe to format changes to re-format displayed values when something changes (e.g. new format override is set or active unit system is changed).

core/ecschema-metadata/src/Interfaces.ts
hl662: Oops, I've removed `onFormatChanged`, keeping `onFormatsChanged` instead.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: You make a good point on the cache: I've removed the formatsCache in SchemaFormatsProvider, so this provider will be read only now. I added a MutableFormatsProvider interface instead, that exposes adding/removing formats. With this, an application could combine a SchemaFormatsProvider and an implementation of the MutableFormatsProvider.

Implementations of a MutableFormatsProvider could be one that integrates local persistence of formats, one that saves persistence to the cloud, or just a runtime cache of formats.

This way, components that require editing formats, like the existing QuantityFormatSettingsPage, could have a MutableFormatsProvider passed into that component, while readonly/visualization components could just pass in the FormatsProvider interface, separating the two from each other.

I added a unitSystem property to the SchemaFormatsProvider, which will change what units are used to format quantities when retrieving formats from KoQs.

I think I'd prefer it if presentation library takes in a formatProvider via a parameter, rather than a global property like IModelApp.

core/ecschema-metadata/src/Interfaces.ts
hl662: There are 2 further reasons I put the generic `FormatsProvider` inside `ecschema-metadata`:

1. I want to align `ecschema-metadata` and `core-quantity` more closely. The APIs in core-quantity, related to formatting and parsing, aren't fully fleshed out enough to be independent from concepts like KoQs, Units, Phenomenons. We also have some duplication between serialization/deserialization of a SchemaUnit (ecschema-metadata) and a Unit (core-quantity) that could be improved upon. I don't want to put a new interface inside of `core-quantity`, when long term I'd like to move everything over to `echschema-metadata`.
2. [Getting rid of usage of QuantityType](https://github.com/iTwin/itwinjs-core/issues/7807). To remove it, an alternative should be proposed for getting formats, and that would be a formatsProvider. Except, the closest thing to a QuantityType is a KindOfQuantity, so it feels like this interface should be in ecschema-metadata.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+export class SchemaFormatsProvider implements FormatsProvider {
+  private _context: SchemaContext;
+  private _unitSystem: UnitSystemKey;
+  public onFormatsChanged = new BeUiEvent<string[]>();
+  /**
+   *
+   * @param contextOrLocater The SchemaContext or a different ISchemaLocater implementation used to retrieve the schema. The SchemaContext
+   * class implements the ISchemaLocater interface. If the provided locater is not a SchemaContext instance a new SchemaContext will be
+   * created and the locater will be added.
+   * @param _unitExtraData Additional data like alternate display label not found in Units Schema to match with Units; Defaults to empty array.
```
grigasp, line 33: no such param

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+    this._unitSystem = unitSystem;
+  }
+
+  public get context() { return this._context; }
+  public get unitSystem() { return this._unitSystem; }
+
+  public set unitSystem(unitSystem: UnitSystemKey) {
+    this._unitSystem = unitSystem;
+    this.onFormatsChanged.raiseEvent([this._unitSystem]);
```
grigasp, line 50: This surprised me - when looking at the definition of `FormatsProvider.onFormatsChanged`, I was expecting it to be called with names of formats that changed. I think this needs a change, but not sure what - some ideas, in the order of what's good for consumers to worst:
- Raise the event with affected format names.
- Remove the `string[]` argument from the event and simply raise it whenever anything changes.
- Add documentation to the event, explaining its arguments. This is probably needed no matter if or what changes are made to the event.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+          if (!unit) {
+            continue;
+          }
+          const currentUnitSystem = await unit.unitSystem;
+          if (currentUnitSystem && currentUnitSystem.name.toUpperCase() === system) {
+            return getFormatProps(format);
+          }
+        }
+      }
+      return undefined;
```
grigasp, line 75: should we really return here? this is different from [what we have in Presentation packages](https://github.com/iTwin/presentation/blob/master/packages/core-interop/src/core-interop/Formatting.ts), where we:
- fall back to persistence unit format props if their unit system matches requested one,
- fall back to default presentation format even if it doesn't match requested unit system.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (!defaultFormat) {
+      return undefined;
+    }
+    return getFormatProps(defaultFormat);
+  }
+
+
+  /**
+   * Retrieves a Format from a SchemaContext. If the format is part of a KindOfQuantity, the first presentation format in the KindOfQuantity that matches the current unit system will be retrieved.
+   * @param name The full name of the Format or KindOfQuantity.
```
grigasp, line 87: I think docs of this argument should be added to the `FormatsProvider` interface.

docs/learning/quantity/index.md
```diff
@@ -43,6 +45,14 @@ For a detailed description of all the setting supported by FormatProp see the EC
 
 ### Concepts
 
+#### Formats Provider
+
+A [FormatsProvider]($ecschema-metadata) interface helps provide all the necessary formats for displaying formatted quantity values, while also enabling users to add formats of their own.
+
+A [MutableFormatsProvider]($ecschema-metadata) interface extends the read-only FormatsProvider above by allowing adding or removing Formats to the provider.
```
grigasp, line 52:
```suggestion
A [MutableFormatsProvider]($ecschema-metadata) interface extends the read-only [FormatsProvider]($ecschema-metadata) above by allowing adding or removing Formats to the provider.
```

docs/learning/quantity/index.md
```diff
@@ -43,6 +45,14 @@ For a detailed description of all the setting supported by FormatProp see the EC
...
+
+The [SchemaFormatsProvider]($ecschema-metadata) takes in a [SchemaContext]($ecschema-metadata), to provide Formats coming from schemas. The SchemaFormatsProvider also requires a [UnitSystemKey]($quantity) passed in to filter the formatProps returned, according to the current unit system set in the SchemaFormatsProvider.
```
grigasp, line 54: 1. When referencing an API, it should be either a link (e.g. [SchemaFormatsProvider]($ecschema-metadata)), or in backticks (e.g. `SchemaFormatsProvider`).
2. In this section, "format" is used in consistently - in some places it starts with uppercase, in some others - not.
3. Terms like "formatProps" should be avoided. If this is a type, then it should be referenced appropriately (see 1st point). If it's a concept, then we should use general language.

docs/learning/quantity/index.md
```diff
@@ -219,3 +150,117 @@ Each Format defined in a FormatSet need to be mapped to a valid [ECName](../../b
 ```
 
 </details>
+
+## Examples of Usage
+
+### Numeric Format
+
+  The example below uses a simple numeric format and generates a formatted string with 4 decimal place precision. For numeric formats there is no conversion to other units; the unit passed in is the unit returned with the unit label appended if "showUnitLabel" trait is set.
```
grigasp, line 158:
```suggestion
The example below uses a simple numeric format and generates a formatted string with 4 decimal place precision. For numeric formats there is no conversion to other units; the unit passed in is the unit returned with the unit label appended if `showUnitLabel` trait is set.
```

docs/learning/quantity/index.md
```diff
@@ -219,3 +150,117 @@ Each Format defined in a FormatSet need to be mapped to a valid [ECName](../../b
...
+```ts
+[[include:Quantity_Formatting.Composite]]
+```
+
+</details>
+
+### Parsing Values
+
+<details>
+  <summary>Example Code:</summary>
```
grigasp, line 184: do we need a colon here? it's not used in other similar places

docs/learning/quantity/index.md
```diff
@@ -219,3 +150,117 @@ Each Format defined in a FormatSet need to be mapped to a valid [ECName](../../b
...
+
+```ts
+[[include:Quantity_Formatting.Simple_Parsing]]
+```
+
+</details>
+
+### Using a FormatsProvider
+
+The example below uses the SchemaFormatsProvider, an implementation of a FormatsProvider, found in `ecschema-metadata` to format values associated with the length of an object.
```
grigasp, line 194: APIs should be referenced appropriately (below as well)

core/ecschema-metadata/src/SchemaFormatsProvider.ts
grigasp: > I think I'd prefer it if presentation library takes in a formatProvider via a parameter, rather than a global property like IModelApp.

me too

docs/learning/quantity/index.md
hl662: I've wrapped APIs in backticks

docs/learning/quantity/index.md
hl662: valid point, I've removed the colon

docs/learning/quantity/index.md
hl662: I've went through the whole document, and modified things such that each section, when mentioning a API for the first time, would use a hyperlink. Subsequent mentions of those APIs would use backticks, so should be good

docs/learning/quantity/index.md
hl662: I've wrapped the term in backtick instead, because the term already uses a link in the relevant section

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: added

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: gotcha, I've removed the return statement here, so we should fall back to the default presentation format (if there is one)

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: thanks, removed

core/ecschema-metadata/src/SchemaFormatsProvider.ts
grigasp: do you think the fallback to persistence unit is not necessary anymore?

docs/learning/quantity/index.md
grigasp: I see 4 more occasions of `:` inside the `<summary>` block

core/ecschema-metadata/src/SchemaFormatsProvider.ts
grigasp: I think you chose the right path. 

One small detail - I think `FormatsChangedArgs` needs some improvement. Currently both of the arguments are optional, so you can pass neither of them or both:
```ts
// what does this mean? nothing changed?
const args1: FormatsChangedArgs = {};

// did everything change or just "x"?
const args2: FormatsChangedArgs = {
  allFormatsChanged: true,
  formatsChanged: ["x"],
};
```

Instead, I suggest this:
```ts
export interface FormatsChangedArgs {
  /**
   * If `all` - all formats within the `FormatsProvider` have changed. 
   * If array, the array items list the names of formats that were changed or removed.
   */
  formatsChanged: "all" | string[];
}
```

On the other hand, I think we should also consider what happens when you combine multiple providers and one of them raises the "all" event. The topmost provider will have to re-raise it as "all" too, causing listeners to reload formats from all providers, even the ones where nothing changed. So I still think it would be great to always say what formats changed... Maybe `SchemaFormatsProvider` could hold a list of formats that were requested and raise the event with that list?..

core/ecschema-metadata/src/Interfaces.ts
grigasp: FWIW, I'll add one more reason I think it would be better to have formatting code kept away from EC stuff. With Cesium acquisition, there are a lot of talks about how we can converge our functionality, and one of the items there is loading non-graphical data in various formats (trees, tables, property grids, etc.) - all of that requires ability to format. And we don't want to force Cesium consumers to have dependencies on EC packages if they're not going to use iTwins/iModels, right?

ColinKerr: At the moment the format system is entangled with core-frontend for QuantityType, unit definitions and the format provider that brings them together, users can opt into using the EC defined units and formats.  The multiple options have proven complex for consuming code and the 'base' providers do not provide support for all units used in an iModel.  So when we discussed the options with several users they wanted to drop the partial implementation in core-frontend and focus on the EC based implementation.

Our current goal is to simplify use for iTwin apps and set up the EC based implementation as the default.  I think it's worthy to consider how the formatting library could be used outside the context of iTwin.js but core-quantity is heavily dependent on an external source for unit conversion.  Today ecschema-metadata has the logic to generate unit conversions and the unit library is stored as an ECSchema.  So if you're not using EC you might consider using one of the other Open Source unit and formatting libraries.  That is in fact something we need to do some more investigation on ourselves.

Our core-quantity formatting library has some industry specific formatting options not available in the alternatives that we have previously investigated, e.g. station offset, but the other libraries provide significantly better math options.  More investigation is needed to see if we can extend these libraries to meet our industry specific needs so we can use them over a home grown solution.

I think it would take significant rework to make core-quantity worth using outside the scope of iTwin.js but I agree we want to re-evaluate our plans if that is our goal.  Taking @grigasp suggestion would at least avoid adding another dependency that would have to be undone if we choose to use core-quantity outside of iTwin.js and without ecschema-metadata.  I think it's reasonable so long as we do not negatively impact our goals for iTwin.js apps:
- iTwin.js apps all use EC based unit conversion and the new format sets
- Use is seamless and the default for users of core-frontend
- We can use without core-frontend for things like generating reports or other operations that may not be done on the frontend.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: I added a private formatsRetrieved set to the SchemaFormatsProvider, one thing I was uncertain about was clearing that set when the unit system is changed. I figured it wouldn't be a big hassle for the set to be repopulated though. I did want to make sure that the keys added to that set would always be the schemaItemKey downstream apps would call (not any of the fallback keys)

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: I was a bit hesitant to use it, in particular defaulting to using DefaultRealU and such. What if the KindOfQuantity was targeting non decimal formatTypes and etc. I added that fallback for now, and if we find the scenario above to be a problem we could remove it, or rearrange the order of the fallbacks.

docs/learning/quantity/index.md
hl662: missed it, thanks

core/ecschema-metadata/src/Interfaces.ts
hl662: Okay, I've moved `FormatsProvider` and `MutableFormatsProvider` to `core-quantity`, and added a `FormatDefinition` interface that extends the `FormatProps` interface. Modified the hyperlinks of the docs as well.

core/quantity/src/test/FormatsProvider.test.ts
grigasp, line 1: Why do you need these tests?

common/changes/@itwin/core-quantity/nam-formats-provider_2025-04-11-16-02.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-quantity",
+      "comment": "Add FormatDefinition, FormatsProvider interfaces to core-quantity ",
```
grigasp, line 5: - Code in backticks.
- No need to name the package since this is package-specific change file.

```suggestion
      "comment": "Add `FormatDefinition`, `FormatsProvider` and `MutableFormatsProvider` interfaces.",
```

core/ecschema-metadata/src/Interfaces.ts
```diff
@@ -6,7 +6,9 @@
  * @module Metadata
  */
 
+import { BeUiEvent } from "@itwin/core-bentley";
 import { DelayedPromise } from "./DelayedPromise";
+import { SchemaItemFormatProps } from "./Deserialization/JsonProps";
```
grigasp, line 11: I believe these imports are not used anymore

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -0,0 +1,159 @@
...
+    this.clear();
+  }
+
+  private clear(): void {
+    const formatsChanged = Array.from(this._formatsRetrieved);
+    this._formatsRetrieved.clear();
+    this.onFormatsChanged.raiseEvent({ formatsChanged });
+  }
+
+  private async getKindOfQuantityFormatFromSchema(itemKey: SchemaItemKey, unitSystem: UnitSystemKey): Promise<SchemaItemFormatProps | undefined> {
```
grigasp, line 60: I there a point to have `unitSystem` as argument for this function, when it can access `this._unitSystem`?

core/ecschema-metadata/src/test/SchemaFormatsProvider.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    const schemaXml = fs.readFileSync(schemaFile, "utf-8");
+    deserializeXmlSync(schemaXml, context);
+
+    const aecSchemaFile = path.join(__dirname, "..", "..", "..", "node_modules", "@bentley", "aec-units-schema", "AecUnits.ecschema.xml");
+    const aecSchemaXml = fs.readFileSync(aecSchemaFile, "utf-8");
+    deserializeXmlSync(aecSchemaXml, context);
+
+
+
+  });
```
grigasp, line 50:
```suggestion
    deserializeXmlSync(aecSchemaXml, context);
  });
```

core/ecschema-metadata/src/test/SchemaFormatsProvider.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  beforeEach(() => {
+    formatsProvider = new SchemaFormatsProvider(context, "metric");
+    unitsProvider = new SchemaUnitProvider(context, UNIT_EXTRA_DATA);
+  });
+
+  it("should throw an error when format doesn't follow valid name convention", async () => {
+    await expect(formatsProvider.getFormat("nonExistentFormat")).to.be.rejected;
+  });
+  it("should return undefined when format is not found", async () => {
```
grigasp, line 60:
```suggestion
  });

  it("should return undefined when format is not found", async () => {
```

example-code/snippets/src/quantity/FormatsProvider.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    // __PUBLISH_EXTRACT_START__ Quantity_Formatting.Mutable_Formats_Provider_Adding_A_Format
+    const formatsProvider = new ExampleFormatProvider();
+    const format: FormatDefinition = {
+      label: "NewFormat",
+      type: "Fractional",
+      precision: 8,
+      formatTraits: ["keepSingleZero", "showUnitLabel"],
+      uomSeparator: "",
+    };
+    await formatsProvider.addFormat("AecUnits.LENGTH", format); // Add a format with the name "AecUnits.LENGTH". The name could be anything as long as it is unique.
```
grigasp, line 114: Should we say that "the name could be anything", when one of our providers - `SchemaFormatsProvider` - throws if it receives a name in a format other than full-schema-item name?

core/ecschema-metadata/src/test/SchemaFormatsProvider.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    expect(spy.calledWith({ formatsChanged: [] })).to.be.true;
+    const formatProps = await formatsProvider.getFormat("AecUnits.LENGTH_LONG");
+    expect(formatProps).not.to.be.undefined;
+    const format = await Format.createFromJSON("testFormat", unitsProvider, formatProps!);
+    const persistenceUnit = await unitsProvider.findUnitByName("Units.M"); // or unitsProvider.findUnit("m");
+    const formatSpec = await FormatterSpec.create("TestSpec", format, unitsProvider, persistenceUnit);
+
+    const result = formatSpec.applyFormatting(50);
+    expect(result).to.equal("164'0 1/2\"");
+  });
```
grigasp, line 102: Doesn't look like these tests belong here. You should be testing `SchemaFormatsProvider` rather than `FormatterSpec` and how it formats values.

common/changes/@itwin/ecschema-metadata/nam-formats-provider_2025-03-25-19-09.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-metadata",
+      "comment": "Add SchemaFormatsProvider to ecschema-metadata",
```
grigasp, line 5:
```suggestion
      "comment": "Add `SchemaFormatsProvider` class that implements `FormatsProvider` from `@itwin/core-quantity` and looks for formats in schemas using given `SchemaContext`.",
```

core/quantity/src/test/FormatsProvider.test.ts
hl662: yeah, I wanted example code in the tests for user reference, but I've now removed this file, and added this example in example-code-snippets, and added a section in the learning docs of using a `MutableFormatsProvider`

common/changes/@itwin/core-quantity/nam-formats-provider_2025-04-11-16-02.json
hl662: done

core/ecschema-metadata/src/Interfaces.ts
hl662: thanks, removed

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: this was a holdover from a previous implementation, I've removed this argument now that we can use `this._unitSystem` directly.

example-code/snippets/src/quantity/FormatsProvider.test.ts
hl662: yeah... I've removed that section of the sentence, added a section for SchemaFormatsProvider in learning docs about throwing an error if not a ec full name when getting a format.

core/ecschema-metadata/src/test/SchemaFormatsProvider.test.ts
hl662: yeah these tests were more integration focused, but they aren't needed anymore now that example-code-snippets have the same test cases. Removed these tests

common/changes/@itwin/ecschema-metadata/nam-formats-provider_2025-03-25-19-09.json
hl662: done

grigasp (approved):
LGTM

core/quantity/src/Formatter/Interfaces.ts
```diff
@@ -94,3 +95,43 @@ export interface CloneOptions {
...
+/** This interface is implemented by a class that would provide formats for use in formatting quantities.
+ * @beta
+ */
+export interface FormatsProvider {
+  /**
+   * @param name The full name of the Format or KindOfQuantity.
+   */
+  getFormat(name: string): Promise<FormatDefinition | undefined>;
+
+  onFormatsChanged: BeUiEvent<FormatsChangedArgs>;
```
aruniverse, line 128: do we need `BeUiEvent` or can we just use `BeEvent`?

example-code/snippets/src/quantity/Parsing.test.ts
```diff
@@ -140,4 +140,5 @@ describe("Parsing examples", () => {
     assert.equal((result as ParsedQuantity).value, 0.9144);
     assert.equal((result as ParsedQuantity).value,(result2 as ParsedQuantity).value);
   });
+
```
aruniverse, line 143: revert files where theres only white space changes

aruniverse (approved):

example-code/snippets/src/quantity/Parsing.test.ts
hl662: done

core/quantity/src/Formatter/Interfaces.ts
hl662: Certainly, I switched over to `BeEvent`, took minimal changes

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -7,125 +7,121 @@
...
-  public onCacheCleared = new BeUiEvent<void>();
+  private _unitSystem: UnitSystemKey;
+  private _formatsRetrieved: Set<string> = new Set();
+  public onFormatsChanged = new BeEvent<(args: FormatsChangedArgs) => void>();
   /**
    *
    * @param contextOrLocater The SchemaContext or a different ISchemaLocater implementation used to retrieve the schema. The SchemaContext
    * class implements the ISchemaLocater interface. If the provided locater is not a SchemaContext instance a new SchemaContext will be
    * created and the locater will be added.
-   * @param _unitExtraData Additional data like alternate display label not found in Units Schema to match with Units; Defaults to empty array.
```
ColinKerr: add doc for unitSystem parameter

ColinKerr: OK, I see where you describe it.  Not sure if that is the best place but maybe improving it is best to a second pass that will work on documentation

ColinKerr (approved):

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: Done, I added a @param field for unitSystem.

PR #7866, opened 2025-03-19
Supporting esm in backend packages

pankhur94:
- Support esm for backend packages
- update sinon import in core-backend


pankhur94:
We are using some utils classes from core-backend in example-code snippets. Since now we have an export map in place, we would have to export it(which I have done). But I am not sure if we should be exporting these? Is there a more better way to do it rather than exposing these from api?

presentation/backend/package.json
```diff
@@ -21,11 +21,26 @@
...
+  },
   "scripts": {
-    "build": "npm run -s build:cjs",
-    "build:cjs": "tsc 1>&2",
-    "build:watch": "npm run -s build:cjs -- -w",
+    "build": "npm run -s build:cjs && npm run -s build:esm",
+    "build:cjs": "tsc 1>&2 -p tsconfig.cjs.json",
+    "build:esm": "tsc 1>&2 -p tsconfig.esm.json",
+    "build:watch:cjs": "npm run -s build:cjs --w",
+    "build:watch:esm": "npm run -s build:esm --w",
```
grigasp, line 43: `--w` won't work (you need `-- -w`), but I don't think we need these `watch` scripts anyway - I'd rather they're removed.

presentation/backend/tsconfig.esm.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "ESNext",
+    "moduleResolution": "node",
```
grigasp, line 5: why not the following?

```
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
```

presentation/backend/package.json
pankhur94: removed

presentation/backend/tsconfig.esm.json
pankhur94: keeping in line with the repo, I have used ES2022.

example-code/snippets/src/backend/AttachDb.test.ts
aruniverse, line 1: no actual changes to this, id revert this file; as well as all the other .test.ts files touched by this pr in example-code & full stack tests

pankhur94: done

aruniverse (approved):

aruniverse:
@tcobbs-bentley @toddsouthenbentley can you guys take a look at the ios failures?

presentation/backend/tsconfig.esm.json
grigasp: From https://www.typescriptlang.org/tsconfig/#module:

> You very likely want "nodenext" for modern Node.js projects

Regarding `moduleResolution` - I see you removed it from this config, which means we take it from `tsconfig-base.json` in build tools, and it's set to `node` there. Here's what https://www.typescriptlang.org/tsconfig/#moduleResolution says about `node` resolution mode:

> 'node10' (previously called 'node') for Node.js versions older than v10, which only support CommonJS require. You probably won’t need to use node10 in modern code.

And here: https://www.typescriptlang.org/docs/handbook/modules/reference.html#node10-formerly-known-as-node:

> --moduleResolution node was renamed to node10 (keeping node as an alias for backward compatibility) in TypeScript 5.0. It reflects the CommonJS module resolution algorithm as it existed in Node.js versions earlier than v12. It should no longer be used.

pankhur94: I discussed this with arun, and we think that it is fine to have node module resolution for now, it has worked fine with frontend packages. However, we will be testing these changes post publish and observe if we have some issues. @aruniverse.

tcobbs-bentley:
@pankhur94, @aruniverse Both calls are from the `before` hook in TestUtils.ts. The first call is from the esm version of that hook, and the second is from the cjs version of that same hook. So it looks like both cjs and esm versions are trying to run, and since the backend is already initialized at the time of the cjs run, the cjs version fails. Obviously having both sets of code loaded and running at the same time is a bad thing. I don't know why it only happens on the mobile test and not `rush cover`.

tcobbs-bentley:
@pankhur94 @aruniverse The iOS tests still run using webpack. I'm guessing that the glob pattern for the webpack needs to be updated.

tcobbs-bentley:
I pushed what I believe to be the fix: https://github.com/iTwin/itwinjs-core/pull/7866/commits/ae4ef6bb93a1f0241ab44b2f328a082f98d8046b

aruniverse (approved):

pmconne (approved):

presentation/backend/tsconfig.esm.json
pankhur94: I tried replacing es2022 with nodenext, although nodenext is used for both esm and cjs outputs, its output is determined by either the module system of the package(type field in package.json) defaults to commonjs, or the file extensions, which in case of typescript would be  .ts -> cjs, .mts -> esm. If we decide to change type to module instead of current commonjs, we would have to add .js in all the imports statements, which would make this pr heavy. Had a conversation with arun, and I believe it would be better for us to have this pr merged, published and pull this up for testing and may be look at nodenext in a separate pr.

grigasp: Right, Node.js relies on `type` attribute in `package.json` to tell if the package is ESM of CommonJS, with the latter being the default. So even though this PR adds ESM build output, that output is not usable in Node.js applications, because it lacks that attribute and Node.js doesn't know it's ESM.

I'm in the process of adding real support for ESM in presentation packages, including running tests on ESM to confirm that's the case. That's going to be in a separate PR. In the meantime, feel free to resolve this comment to complete this PR.

grigasp: Here's my shot at this: https://github.com/iTwin/itwinjs-core/pull/7885

grigasp (approved):

PR #7867, opened 2025-03-20
Rewrite how aspect can be used in relationships to be less confusing

### LINKED ISSUE #7802: Documentation Issue: Sentence in ElementAspect explanation needs clarification
The sentence, shown below, without context, contradicts itself (i.e. ElementAspect can't be in a relationship, but it can be in a relationship).

Image

Found on page https://www.itwinjs.org/bis/guide/intro/modeling-with-bis - "Modeling with BIS"

ColinKerr:
Closes #7802

### DIFF
--- a/docs/bis/guide/intro/modeling-with-bis.md
+++ b/docs/bis/guide/intro/modeling-with-bis.md
@@ -78,7 +78,7 @@ These two rules imply that a given class of Element cannot be both sub-modeled *
 
 ### ElementAspect
 
-[ElementAspects](../references/glossary.md#elementaspect) are a flexible way to augment the properties of an Element. They are sets of properties that typically hold information needed only in certain contexts, e.g. during the construction phase or when we have a link to information about the modeled Entity in a different repository. ElementAspects are not individually identifiable (thus relationships cannot point to them), but they may be the "source" of a relationship pointing to an Element.
+[ElementAspects](../references/glossary.md#elementaspect) are a flexible way to augment the properties of an Element. They are sets of properties that typically hold information needed only in certain contexts, e.g. during the construction phase or when we have a link to information about the modeled Entity in a different repository. ElementAspects are not individually identifiable so they can only be used in relationships where a [Navigation Property](../../ec/ec-property.md#ecnavigationproperty) exists on the aspect pointing to an Element.  For example every Aspect has a Navigation Property pointing to the Element that owns it.
 
 ### Identifiers
 
@@ -94,5 +94,6 @@ The [**FederationGuid**](../references/glossary.md#federationguid) is optional b
 **UserLabel** is an optional property of an Element that can be used as an informal name in the GUI, but it does not have to be unique. In some GUIs, if the UserLabel is null, the CodeValue will be used as a display label.
 
 ---
+
 | Next: [BIS Organization](./bis-organization.md)
 |:---


docs/bis/guide/intro/modeling-with-bis.md
```diff
@@ -78,7 +78,7 @@ These two rules imply that a given class of Element cannot be both sub-modeled *
 
 ### ElementAspect
 
-[ElementAspects](../references/glossary.md#elementaspect) are a flexible way to augment the properties of an Element. They are sets of properties that typically hold information needed only in certain contexts, e.g. during the construction phase or when we have a link to information about the modeled Entity in a different repository. ElementAspects are not individually identifiable (thus relationships cannot point to them), but they may be the "source" of a relationship pointing to an Element.
+[ElementAspects](../references/glossary.md#elementaspect) are a flexible way to augment the properties of an Element. They are sets of properties that typically hold information needed only in certain contexts, e.g. during the construction phase or when we have a link to information about the modeled Entity in a different repository. ElementAspects are not individually identifiable so they can only be used in relationships where a [Navigation Property](../../ec/ec-property.md#ecnavigationproperty) exists on the aspect pointing to an Element.  For example every Aspect has a Navigation Property pointing to the Element that owns it.
```
diegoalexdiaz, line 81: I think the sentence "_ElementAspects are not individually identifiable_" still generates confusion. It is true that BIS doesn't expect them to be individually identifiable, but their implementation certainly assigns them an identifier (ECInstanceId).

How about:

An ElementAspect may be associated to a relationship pointing to an Element only when such relationship backs up a [Navigation property](../../ec/ec-property.md#ecnavigationproperty) defined within it. No link-table relationship can reference an ElementAspect, nor relationships among two ElementAspects are allowed.

ColinKerr: While they may technically have an ECInstanceId in an iModel that Id is not guaranteed to be stable over the lifetime of the aspect and we want to be explicit that it does not exist in a BIS repository.  So it is correct to say that aspects are not individually identifiable or addressable, we could have a better way of saying it, but I do think we need to make that clear.

I explicitly tried to avoid talking about link tables or FKs because I felt that moved away from BIS and towards how we implement BIS in an iModel.

I think your description leaves open the possibility that some other type of relationship is possible because your first sentence applies to relationships pointing to an element and the second applies to link table relationships  and relationships between aspects.  That may cover the possibilities but knowing that requires the larger context about what could be done if aspects were individually addressable.

dovydasjaru (approved):

docs/bis/guide/intro/modeling-with-bis.md
diegoalexdiaz: I'm generally fine with your arguments and proposed description.

Side comment: BIS, independent of iModel-implementation, really has two kinds of relationships: the ones backing up navigation-properties (which BIS doesn't offer a base class for) and "the rest" (unrelated to properties, but BIS does offer base-classes for them that they must derive from). The BIS guide refers to the latter as "link-table relationships". BTW, I just found a related inaccuracy in the associated [article](https://www.itwinjs.org/bis/guide/fundamentals/relationship-fundamentals/#link-table): "_there is a single link table (ElementRefersToElements)_". ElementDrivesElements is another link-table and it doesn't inherit from ElementRefersToElements.

This conceptual distinction of relationships in BIS is certainly regretful since at a lower-level (i.e. ECObjects), all relationships are treated equal - which surfaces in the ECSQL language.

diegoalexdiaz (approved):

PR #7870, opened 2025-03-20
Add updateInstance Functions

MichaelSwigerAtBentley:
Adds:
updateElementWithHandlers()
updateModelWithHandlers()
updateAspectWithHandlers()
Updates:
updateLinkTableRelationship()

Merging between dev branchs, setting up a PR to collect feedback


MichaelSwigerAtBentley:
@khanaffan

core/backend/src/NativeInstaceHandlers.ts
```diff
@@ -123,4 +122,118 @@ export function insertAspectWithHandlers(iModel: IModelDb, aspectProps: ElementA
   iModel.models.getModel(element.model).update();
 
   return aspectProps.id;
+}
+
+// TODO: Could potentially be used to enforce that the id is required in the ElementProps, Do we want to do this?
+// export type ElementPropsWithId = Partial<ElementProps> & Required<Pick<ElementProps, "id">>;
```
MichaelSwigerAtBentley, line 128: Currently, iModelDb.updateElement takes a Partial ElementProps with no check to ensure id is included, would this work to enforce id inclusion?

khanaffan: This reminds me that we need one of the following to present. I will add native function to findElement(id|fed|guid) and will return {id classFullName } that you can append to elementProp
1. id
2. code
3. fedGuid

khanaffan: Ah wait looking at native code it does expect id otherwise it throw. So your code is correct.

```
void JsInterop::UpdateElement(DgnDbR dgndb, Napi::Object obj) {
    BeJsValue elProps(obj);
    DgnElementId eid = elProps[DgnElement::json_id()].GetId64<DgnElementId>();
    if (!eid.IsValid())
        throwInvalidId();
```

khanaffan: My previous comment only apply to getElement()

PR #7871, opened 2025-03-21
add symbols to main pkgs with singletons & rm deprecated in 3.x apis

aruniverse:
add symbols to the following pkgs to prevent multiple copies:
- core-backend
- presentation-frontend
- presentation-backend

rm deprecated apis in 3.x from core-backend


grigasp (commented):
Regarding the symbol check:
- I think the addition is worth mentioning in the `NextVersion.md`.
- Why is it added only to the three packages, but not, for example, `@itwin/core-frontend`?

core/backend/src/core-backend.ts
```diff
@@ -86,6 +85,10 @@ export * from "./ChangesetECAdaptor";
 
 export * from "./internal/cross-package";
 
+const globalSymbol = Symbol.for("itwin.core.backend.globals");
+if ((globalThis as any)[globalSymbol])
+  throw new Error("Multiple @itwin/core-backend imports detected!");
```
pmconne, line 90: Can you make these errors clearer so the developer understands what happened and how they can fix it?

pmconne (commented):
Any chance you can split this into 2 PRs? The deprecated API removals seem completely unrelated to the symbols.

aruniverse:
> Any chance you can split this into 2 PRs? The deprecated API removals seem completely unrelated to the symbols.

https://github.com/iTwin/itwinjs-core/pull/7875

anmolshres98:

core/backend/src/core-backend.ts
aruniverse: Going to handle this in a separate pr where were adding docs for consumers and developers for understanding dependencies to our docs site

pmconne: I must be dumb, how do these 3 lines of code actually work?

aruniverse: [Symbol.for](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for) checks if the specific symbol has been registered globally and returns it if so, and will create one if not. Since this check has been added to the top level barrel file of the package, the symbol will be created and registered the first time someone imports from the package as a side effect. If you were to import the same package (eg core-backend) with multiple instances in your lockfile/node-modules, it would try to re-register the symbol.

Without an else clause setting the symbol on `globalThis`, it seems like this wouldn't work correctly. However, there's a subtle behavior of the JavaScript module system that allows this to function:

In Node.js and modern JavaScript environments, modules are cached after they're first loaded. When you import a module multiple times in an application, the module's code only executes once, and subsequent imports receive the same exported objects from cache.

Here's how this mechanism works:

1. The first time `@itwin/core-backend` is imported, this code runs and checks if the symbol exists (it doesn't)
1. No error is thrown and execution continues
1. The exports are cached
1. When a second module tries to import `@itwin/core-backend`, the JavaScript module system returns the cached exports without re-executing the module code

The real protection comes from the JavaScript module system itself, not from setting the symbol on `globalThis`. The error would only be thrown if someone somehow circumvented the module cache and forced the module code to execute twice.

That said, adding an else clause to set the symbol would create a more robust guard that doesn't rely on module caching behavior, which would be a good improvement to make the code's intent clearer and protect against edge cases. Fix incoming.

aruniverse: My apologies, so scrolling all the way down to https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/extension/ExtensionRuntime.ts#L550-L552, we do set the globalThis with the symbol in `core-frontend`, but not in the other pkgs so it wouldnt have worked. Will also update

aruniverse: Pushed some changes, and tried to provide some additional context for consumers when debugging.

core/backend/src/core-backend.ts
```diff
@@ -82,6 +82,31 @@ export * from "./ChangesetECAdaptor";
...
+    "Multiple @itwin/core-backend imports detected! This may happen if:\n" +
+      "- You have multiple versions of the package installed\n" +
+      "- Your bundling configuration is incorrect\n" +
+      "- You're importing from both ESM and CommonJS versions"
+  );
+
+  /* eslint-disable no-console */
+  console.error("Duplicate @itwin/core-backend import:", error);
+  console.error("First import occurred at:", firstLoadStack);
+  console.error("Current import occurred at:", new Error().stack);
```
tcobbs-bentley:
```suggestion
  console.error("Current import occurred at:", error.stack);
```

PR #7873, opened 2025-03-21
Incremental schema loading

StefanApfel-Bentley:
Added incremental schema loading:

To locate schemas incrementally, the IncrementalSchemaLocater was added. Clients that want to load their schemas incrementally should add this locater to their context.

The actual schema loading happens in an abstract IncrementalSchemaLoader class. Derived classes may add their own handling to fetch the data before the loader loads them.

Also included:
- SchemaInfo now also holds the schemas alias. This affected a couple of tests that has to be adjusted.
- In SchemaReadHelper a few methods went from private to protected, that they can be overridden when needed.


StefanApfel-Bentley:
This PR can be closed, the code has been merged into master in PR #8247

PR #7874, opened 2025-03-21
Rework ITwinError

kabentley:
The legacy `BentleyError` classes are problematic for two reasons:

1. There's no way for the "errorNumber" value to be unique across the library, let alone applications.
2. Subclasses of BentleyError cannot survive across inter-process marshalling.

This PR introduces a new namespace called `ITwinError` that identifies errors with strings rather than numbers. It also only uses interfaces and static functions to throw/identify exceptions. It retroactively adds those concepts to `BentleyError` so that existing errors may be marshalled across processes (albeit using the new static method `BentleyError.isError`).

The previously committed implementation of ITwinError does not support legacy exceptions (even though it purports to work for ConflictingLocksError but doesn't).


kabentley:
@pankhur94 @karolis-zukauskas please review

pmconne:
See # #7876

full-stack-tests/core/src/frontend/standalone/ITwinError.test.ts
```diff
@@ -0,0 +1,73 @@
+import { BentleyError, IModelHubStatus, ProcessDetector } from "@itwin/core-bentley";
+import { BackendError, ChannelError, ConflictingLock, ConflictingLocksError, LockState } from "@itwin/core-common";
+import { expect } from "chai";
+import { coreFullStackTestIpc } from "../Editing";
+import { TestUtility } from "../TestUtility";
+
+if (ProcessDetector.isElectronAppFrontend) {
+
+  describe("ITwinError exceptions from backend", async () => {
```
aruniverse, line 9:
```suggestion
  describe("ITwinError exceptions from backend (#integration)", async () => {
```

aruniverse (commented):
Thanks for the PR! small nit re namespaces which might change implementation of the other errors

PR #7876, opened 2025-03-21
Rework ITwinError

pmconne:
Contributed by @kabentley 

The legacy `BentleyError` classes are problematic for two reasons:

1. There's no way for the "errorNumber" value to be unique across the library, let alone applications.
2. Subclasses of BentleyError cannot survive across inter-process marshalling.

This PR introduces a new namespace called `ITwinError` that identifies errors with strings rather than numbers. It also only uses interfaces and static functions to throw/identify exceptions. It retroactively adds those concepts to `BentleyError` so that existing errors may be marshalled across processes (albeit using the new static method `BentleyError.isError`).

The previously committed implementation of ITwinError does not support legacy exceptions (even though it purports to work for ConflictingLocksError but doesn't).


core/bentley/src/BentleyError.ts
```diff
@@ -8,6 +8,69 @@
...
 import { RepositoryStatus } from "./internal/RepositoryStatus";
+import { Optional } from "./UtilityTypes";
+
+type BaseError = Error;
+
+/**
+ * Namespace for handling all exceptions thrown by iTwin.js
+ * @beta
+ */
+export namespace ITwinError {
```
aruniverse, line 19: nit:

Do we really need a namespace here? I know our codebase is littered with namespaces so it will be impossible to completely get away from it anytime soon, but we have been trying to move away from adding any new namespaces.

We could move all the new additions to a new module called "ITwinError.ts". We could then also eliminate the type alias BaseError, and the interface currently defined as Error can be renamed to ITwinError. Then all utility functions (createError, throwError, isError) could be renamed to include ITwin in it; while a bit more verbose it is (imo) a cleaner way than using a namespace. The isObject type guard could be extracted out of the namespace, and could be used more generically than just in the context of ITwinErrors

core/bentley/src/BentleyError.ts
```diff
@@ -315,20 +378,37 @@ export type GetMetaDataFunction = () => object | undefined;
...
+  loggingMetadata?: object;
+}
+
+/**
+ * Base exception class for legacy iTwin.js errors.
+ * For backwards compatibility only. Do not create new subclasses of BentleyError. Instead use [[ITwinError]].
  * @public
  */
-export class BentleyError extends Error {
+export class BentleyError extends Error { // note: this class implements LegacyITwinErrorWithNumber but can't be declared as such because that interface is @beta.
```
aruniverse, line 410: wondering if we should promote LegacyITwinErrorWithNumber to public but also marked as @deprecated to be more explicit

full-stack-tests/core/src/frontend/standalone/ITwinError.test.ts
```diff
@@ -0,0 +1,73 @@
+import { BentleyError, IModelHubStatus, ProcessDetector } from "@itwin/core-bentley";
+import { BackendError, ChannelError, ConflictingLock, ConflictingLocksError, LockState } from "@itwin/core-common";
+import { expect } from "chai";
+import { coreFullStackTestIpc } from "../Editing";
+import { TestUtility } from "../TestUtility";
+
+if (ProcessDetector.isElectronAppFrontend) {
+
+  describe("ITwinError exceptions from backend", async () => {
```
aruniverse, line 9:
```suggestion
  describe("ITwinError exceptions from backend  (#integration)", async () => {
```

aruniverse (commented):
Thanks for the PR! small nit re namespaces which might change implementation of the other errors

(copying review over from the original pr: https://github.com/iTwin/itwinjs-core/pull/7874#pullrequestreview-2707096260)

core/bentley/src/BentleyError.ts
pmconne: Namespaces impede tree-shaking. OTOH, free functions are less discoverable than a group of related functions aggregated into a namespace. Do you have other reasons to dislike namespaces?

core/common/src/ITwinCoreErrors.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+import { ITwinError } from "@itwin/core-bentley";
+
+/**
+ * Errors associated with the `ChannelControl` interface. This interface exists in the common package so
+ * they may be handled on both the frontend and backend.
+ * @beta
+ */
+export namespace ChannelError {
```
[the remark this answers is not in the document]
pmconne: I don't necessarily object to the namespace. I do think you should define `ChannelError.Error` as `interface ChannelError` and let TypeScript merge the interface and namespace declarations; same for `ITwinError.Error` - that's noisy and unnecessary vs `ITwinError`.

core/bentley/src/BentleyError.ts
kabentley: This new interface shouldn't be deprecated - it will have to live forever since we can't get rid of BentleyError.

core/bentley/src/BentleyError.ts
kabentley: With namespaces, we'll have:
- `ITwinError.isError`
- `BentleyError.isError`
- `ChannelError.isError`
- etc.

When others create their own set of Errors, they'll create a namespace for it following the pattern in `ChannelError`. Seems very clean to me. Otherwise there'll be just lots of loose top-level types related only by some prefix. I get tree shaking, but that's true of every class too,

core/common/src/ITwinCoreErrors.ts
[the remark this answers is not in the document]
kabentley: I didn't know you could do that. Unfortunately we can't do that for BentleyError.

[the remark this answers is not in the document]
kabentley: unfortunately I can't push to this branch

core/bentley/src/BentleyError.ts
```diff
@@ -315,20 +378,37 @@ export type GetMetaDataFunction = () => object | undefined;
...
+ * An `ITwinError.Error` instance that also supplies an `errorNumber`.
+ * @note this interface exists *only* for legacy errors derived from `BentleyError`. The concept of "error number" is
+ * problematic since it is impossible to enforce across the iTwin.js library, let alone across applications. New code should
+ * use `ITwinError` and identify errors with strings instead.
+ * @beta */
+export interface LegacyITwinErrorWithNumber extends ITwinError.Error {
+  /** a number to identify the error. */
+  readonly errorNumber: number;
+
+  /** Logging metadata
```
karolis-zukauskas, line 398: We use `Logger.logException` to log errors, but it doesn't accept metadata. So now that `ITwinError` doesn't have metadata - how do we log it?

We try to avoid `Logger.logError` as it only logs error `message` - it will not include the callstack unless the caller explicitly "formats" it into the message (never happens). I like `Logger.logException` because internally it uses `Logger.getExceptionMessage` which adds callstack to the message and usage is very convenient, just `Logger.logException("cat", e)` - no additional calls to "extract" the message or check if `e` is really an error - all this boilerplate is done inside logException implementation.

It was quite convenient to attach additional context about an error to the metadata for logging purposes as it was travelling down the callstack. I guess we can now attach that same info directly to the `message` .. but it will be harder for external systems like Sentry to group errors by type if the message will vary. iTwin Studio automatically sends errors to Sentry when `Logger.logException` is called - it adds metadata as additional context object in Sentry.

I don't really understand why we are getting rid of `metadata` from `ITwinError`. Seems like a downgrade.

core/common/src/ITwinCoreErrors.ts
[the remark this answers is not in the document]
karolis-zukauskas: > I do think you should define ChannelError.Error as a top-level interface ChannelError (outside the namespace)

agree

core/bentley/src/BentleyError.ts
karolis-zukauskas: In OpenSite+ we defined a `Errors.withDetails(e, message, metadata?)` utility to "enrich" errors in travel as often the original error isn't that helpful e.g. "Element not found" / "Error inserting element". Callstacks in iTwin Studio app context are also quite often not very helpful. See my [previous comment on this](https://github.com/iTwin/itwinjs-core/issues/7442#issuecomment-2684663251).

karolis-zukauskas: On a side note, I dislike that there is `Logger.logError` and `Logger.logException` .. there really should only be one preferably one that accepts error object as parameter, not a message.

Crazy idea:
What if itwinjs deprecates `Logger.logException` and makes `Logger.logError` accept `string | object` as second param? Logger.logError accepts `metadata` as third param, so it would solve the issue I described above. There would also be less confusion in usage - _"which function should I use to log? How does error differ from exception"_

pmconne: > How does error differ from exception

logError logs a message at log level "error". An exception may or may not have occurred.

kabentley: > It was quite convenient to attach additional context about an error to the metadata for logging purposes as it was travelling down the callstack. 

Wouldn't it be better to create a new interface for your error and add that data explicitly rather than in a hidden/untyped blob in metadata? That way people could actually use it beyond logging.

kabentley: BTW, an argument could be made that `Logger.logException` should log all of the members of the exception.

karolis-zukauskas: @kabentley 
That's right, I could just add a new property to the error object and rethrow it without changing anything else. That is essentially what I was doing before with the `metadata` field on BentleyError.

> an argument could be made that Logger.logException should log all of the members of the exception.

If `metadata` is to be removed from base error type, then I say we need `Logger.logException` to log all members of the error object. Otherwise we are loosing information by calling `Logger.logException` and the alternative `Logger.logError` is no better.

kabentley:
@pmconne I suggest this change to `logException`:

```ts
  /** Log the specified exception. The special "ExceptionType" property will be added as metadata.
   * @param category  The category of the message.
   * @param err  The exception object.
   * @param log The logger output function to use - defaults to Logger.logError
   */
  public static logException(category: string, err: any, log: LogFunction = (_category, message, metaData) => Logger.logError(_category, message, metaData)): void {
    log(category, Logger.getExceptionMessage(err), () => {
      // For backwards compatibility, log BentleyError old way
      if (BentleyError.isError(err))
        return { ...BentleyError.getErrorMetadata(err), exceptionType: err?.constructor?.name ?? "<Unknown>" };

      // return a copy of the error, with non-enumerable members `message` and `stack` removed, as "metadata" for log.
      return { ...err };
    });
  }
```

Here's an update to the test:

```ts
  it("should log exceptions", () => {
    Logger.initialize(
      (c, m, d) => outerr = [c, m, BentleyError.getMetaData(d)],
      (c, m, d) => outwarn = [c, m, BentleyError.getMetaData(d)],
      (c, m, d) => outinfo = [c, m, BentleyError.getMetaData(d)],
      (c, m, d) => outtrace = [c, m, BentleyError.getMetaData(d)]);
    Logger.setLevel("testcat", LogLevel.Error);

    clearOutlets();
    try {
      throw new Error("error message");
    } catch (err: any) {
      Logger.logException("testcat", err);
    }
    checkOutlets(["testcat", "Error: error message", { ExceptionType: "Error" }], [], [], []);
    const m1 = { a: 200, b: "b" };
    const e1 = new BentleyError(IModelStatus.AlreadyLoaded, "test message", m1);
    clearOutlets();
    Logger.logException("testcat", e1);
    expect(outerr[0]).equal("testcat");
    expect(outerr[1]).equal("Already Loaded: test message")
    expect(outerr[2]).deep.equal({ ...m1, exceptionType: "BentleyError" });

    clearOutlets();
    const e2 = ITwinError.createError<ITwinError.Error & { val1: number }>({ iTwinErrorId: { key: "key1", scope: "scope1" }, message: "itwin error", val1: 200 });
    Logger.logException("testcat", e2);
    expect(outerr[0]).equal("testcat");
    expect(outerr[1]).equal("key1: itwin error");
    expect(outerr[2]).deep.equal({ ...e2 }); // message and stack should be stripped off 

    clearOutlets();
    expect(() => Logger.logException("testcat", undefined)).to.not.throw("undefined exception");
    checkOutlets(["testcat", "Error: err is undefined.", { ExceptionType: "Error" }], [], [], []);

    clearOutlets();
    expect(() => Logger.logException("testcat", null)).to.not.throw("null exception");
    checkOutlets(["testcat", "Error: err is null.", { ExceptionType: "Error" }], [], [], []);

  });
```

core/bentley/src/BentleyError.ts
pmconne: I pushed @kabentley's suggested change. Are you satisfied @karolis-zukauskas?

core/bentley/src/BentleyError.ts
pmconne: @aruniverse I pulled the interfaces out of the namespaces. Do you have further objections/suggestions?

core/bentley/src/BentleyError.ts
pmconne: @aruniverse satisfied?

core/common/src/ITwinCoreErrors.ts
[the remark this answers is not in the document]
pmconne: ITwinError and ChannelError are now both namespaces and interfaces.

core/bentley/src/Logger.ts
```diff
@@ -286,7 +286,12 @@ export class Logger {
    */
   public static logException(category: string, err: any, log: LogFunction = (_category, message, metaData) => Logger.logError(_category, message, metaData)): void {
```
kabentley, line 287: The comment for this method should be updated to say "For legacy BentleyError exceptions, the special "exceptionType" property will be added as metadata. Otherwise all enumerable members of the exception are logged as metadata."

pmconne: updated.

core/bentley/src/BentleyError.ts
```diff
@@ -315,20 +376,37 @@ export type GetMetaDataFunction = () => object | undefined;
...
 
 interface ErrorProps {
   message: string;
   stack?: string;
   metadata?: object;
 }
 
-/** Base exception class for iTwin.js exceptions.
+/**
+ * An `ITwinError.Error` instance that also supplies an `errorNumber`.
```
kabentley, line 387:
```suggestion
 * An `ITwinError` instance that also supplies an `errorNumber`.
```

full-stack-tests/core/src/frontend/standalone/ITwinError.test.ts
aruniverse: Ah good catch, just saw the original test file had the integration test; not applicable here. closing

core/bentley/src/BentleyError.ts
pmconne: I would expect that we eventually deprecate BentleyError in favor of ITwinError. However, ITwinError is being introduced as beta, so BentleyError shouldn't be deprecated in this PR.

Your original suggestion of introducing and immediately deprecating LegacyITwinErrorWithNumber seems weird to me. Why introduce a brand-new public API that nobody is supposed to use?

kabentley: I'm not sure how we can make BentleyError deprecated. The problem is that it should be an interface not a class. 

I suppose we could start by making its constructor deprecated and adding a `throwError` method to it. The main thing we need to do is stop people from creating subclasses and doing `instanceOf` tests for it, and I don't know how to do that.

pmconne: Regardless of whether/how we do that, I don't think such changes belong in this PR.

aruniverse (approved):

pmconne:
@pankhur94 any chance that #7892 which switched the domain tests to use ESM have caused those tests to start timing out?

Analytical domain timed out on Windows:
```
2025-03-28T18:25:13.9137616Z   1) AnalyticalSchema
2025-03-28T18:25:13.9137983Z        should import Analytical schema:
2025-03-28T18:25:13.9138685Z      Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\agent\_work\1\s\domains\analytical\backend\lib\esm\test\AnalyticalSchema.test.js)
```


On re-run, linear referencing domain timed out instead:
```
2025-03-28T19:05:40.3064735Z   1) LinearReferencing Domain
2025-03-28T19:05:40.3065088Z        should create elements exercising the LinearReferencing domain:
2025-03-28T19:05:40.3065683Z      Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\agent\_work\1\s\domains\linear-referencing\backend\lib\esm\test\LinearReferencingSchema.test.js)
```

PR #7885, opened 2025-03-25
Presentation: ES modules

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/7878.

- Full ESM support
- Run unit tests on ESM
- Run full stack tests on ESM 

Had to bump `i18next-http-backend` to version `3`, as the `1.x` we were using doesn't support ESM.


full-stack-tests/presentation/scripts/setup-tests.cjs
saskliutas, line 1: Should we add TODO comment on line 27 to replace `cjs` with `esm` when esm support for other backend package added?

grigasp: That'll keep working until we drop `cjs` completely, at which point the tests should start failing and we can update then. Not sure a TODO comment will help us in any way.

saskliutas: I was thinking about switching `cjs` to `esm` as soon as it is available so we won't have to get back to this when dropping `cjs`. However, now I am thinking that `assets` dir should be placed in `lib` instead of `lib/cjs`/`lib/esm` as it will be duplicated when esm supports is added. 😕

grigasp:
@ben-polinsky, docs build shows this error:
```
Invalid Link: "[ITwinLocalization.constructor]($i18n)" in file learning/frontend/localization/index.html"
```

I don't see how my changes could've affected this... Maybe because of `typedoc` bump? Any ideas?

full-stack-tests/presentation/scripts/setup-tests.cjs
grigasp: Good point. I changed `presentation-common` to place them under `lib` rather than having duplicated under `lib/cjs` and `lib/esm`, and adjusted the `setup-tests` script. Please have a look.

saskliutas (approved):

ben-polinsky:
> I don't see how my changes could've affected this... Maybe because of `typedoc` bump? Any ideas?

Most likely, yes. Typedoc is in 0.x semver mode and considers 0.x changes to be possibly breaking. Working to support latest version, but I don't have the bandwidth right now.

presentation/backend/package.json
```diff
@@ -20,22 +20,30 @@
     "name": "Bentley Systems, Inc.",
     "url": "http://www.bentley.com"
   },
+  "type": "module",
   "main": "lib/cjs/presentation-backend.js",
-  "typings": "lib/cjs/presentation-backend",
+  "types": "lib/cjs/presentation-backend",
+  "module": "lib/esm/presentation-backend.js",
```
pankhur94, line 26: are we using any bundlers like webpack for presentation? If not, it might be a good idea to get rid of legacy main and module entry points.

pmconne (approved):

presentation/common/package.json
```diff
@@ -18,31 +18,47 @@
...
+    },
+    "./package.json": "./package.json"
+  },
   "scripts": {
     "build": "npm run -s ruleset-json-schema && npm run -s build:assets && npm run -s build:public && npm run -s build:cjs && npm run -s build:esm",
-    "build:cjs": "tsc -p tsconfig.cjs.json",
+    "build:cjs": "internal-tools create-cjs-package-json -o ./lib/cjs && tsc -p tsconfig.cjs.json",
     "build:esm": "tsc -p tsconfig.esm.json",
-    "build:assets": "cpx \"./assets/**/*\" ./lib/cjs/assets && npm run -s pseudolocalize:assets",
+    "build:assets": "cpx \"./assets/**/*\" ./lib/assets && npm run -s pseudolocalize:assets",
```
aruniverse, line 44: should we call this out in nextver.md?
do you think there will people who are trying to manually resolve your assets from here?

presentation/backend/package.json
aruniverse: we are unlikely to be using a bundler on the backend, frontend yes

aruniverse (approved):
lgtm, thx for taking this on!

presentation/backend/src/presentation-backend/PresentationRpcImpl.ts
```diff
@@ -57,14 +57,17 @@ import {
...
+import { PresentationBackendLoggerCategory } from "./BackendLoggerCategory.js";
+import { Presentation } from "./Presentation.js";
+import { PresentationManager } from "./PresentationManager.js";
+import { TemporaryStorage } from "./TemporaryStorage.js";
+import { getRulesetIdObject } from "./PresentationManagerDetail.js";
+// @ts-expect-error TS complains about `with` in CJS builds; The path is fine at runtime, but not at compile time
+// eslint-disable-next-line @itwin/import-within-package
+import packageJson from "../../../package.json" with { type: "json" };
+
+// const packageJsonVersion = require("../../../package.json").version;
```
hl662, line 69: remove commented out line?

presentation/backend/package.json
grigasp: I feel like we should give one major version of legacy + non-legacy attributes before we remove the legacy ones

tools/internal/scripts/create-cjs-package-json.js
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+const path = require("path");
+const fs = require("fs");
+
+exports.command = "create-cjs-package-json";
```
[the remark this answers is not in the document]
grigasp: I'm hesitant... From one side - yes, dual build (cjs + esm) packages from other repos could use this. From another - this is a very small script and only needed for dual build packages. Having in mind that we're moving towards esm-only, this script will very soon be unnecessary. So not sure, you decide 😆

presentation/common/package.json
grigasp: I would be surprised if anyone is importing them besides `presentation-backend`. But added a section to be safe.

presentation/backend/src/presentation-backend/PresentationRpcImpl.ts
grigasp: yes, thanks for catching this - removed.

grigasp:
> > I don't see how my changes could've affected this... Maybe because of `typedoc` bump? Any ideas?
> 
> Most likely, yes. Typedoc is in 0.x semver mode and considers 0.x changes to be possibly breaking. Working to support latest version, but I don't have the bandwidth right now.
> 
> @grigasp perhaps try typedoc@0.27 which was the first version with ESM. Either way, will try to make some time to look into it.

Turns out all I needed was ability to specify the tsconfig file and I didn't need the `typedoc` bump at all - reverted it back to `0.26`.

PR #7891, opened 2025-03-26
Fix bug with draping onto reality data

danielzhong-bentley:
After merging into master yesterday, some point cloud data failed image tests and rendered nothing. This PR fixes the issue by ensuring files without ECEF location data no longer fail to render.


pmconne:
> ensuring files without ECEF location data no longer fail to render.

What specifically caused them to fail to render? Did something expect the ecef transform to be non-null, and threw an exception when it was null? If so, can you change the types so that the compiler would have caught the error?

danielzhong-bentley:
> > ensuring files without ECEF location data no longer fail to render.
> 
> What specifically caused them to fail to render? Did something expect the ecef transform to be non-null, and threw an exception when it was null? If so, can you change the types so that the compiler would have caught the error?

I believe some of the files simply aren’t geolocated, so they don’t have an associated ECEF location. My original code didn’t handle for that properly.

To answer your question—yes, the draping logic needs a transform input, but we don’t assume one will always be present. If it’s missing, we don’t throw an exception—we just fall back to using an identity transform. There's nothing inherently wrong; it's just that some files don’t include location data.

PR #7900, opened 2025-03-30
Drape onto imodel

danielzhong-bentley:
This PR adds support in iTwin.js for:
1. draping model map layers onto user-selected iModel data.
2. Fixed Batched Tile Tree won't load at the first time

TODO (before merging):
- [x] Add Image Tests testcase (below screenshot)
- [x] Pass Image Tests
- [x] DPTA / Image Tests does not appear to do draping onto imodels - even when the model being draped is still enabled. Try to reproduce locally (using DPTA + config file) and solve it.
- [x] Fix failing test: https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3809990&view=logs&j=8673a5cb-f235-5320-c755-005ead49efbb&t=ae0ee70f-cf19-52a3-7a8c-c857212779af&l=9077
- [x] Edit screenshot so it shows new "drape onto all models" approach.
- [x] Resolve BatchedTileTree items in this comment: https://github.com/iTwin/itwinjs-core/pull/7900#issuecomment-2783665960
- [x] https://github.com/iTwin/itwin-graphics-backlog/issues/636
- [x] https://github.com/iTwin/itwin-graphics-backlog/issues/634

The newly-added `drapeTarget` property (see PR https://github.com/iTwin/itwinjs-core/pull/7637) on [ModelMapLayerProps](https://www.itwinjs.org/reference/core-common/maplayers/modelmaplayerprops/) and [ModelMapLayerSettings](https://www.itwinjs.org/reference/core-common/maplayers/modelmaplayersettings/) now can be specified as `ModelMapLayerDrapeTarget.IModel`. When specified as such, the model map layer will be only draped onto the array of iModel models specified in the new `ModelMapLayerProps.drapeModels` property.

Here is a sample screenshot of draping a model from within an iModel onto all models from that same iModel. The red piping model is draping onto all models (even itself); all of these models are contained within the same iModel.

models-draping-onto-models

[Test dataset found here](https://github.com/iTwin/itwinjs-core/pull/7900#issuecomment-2769378659).


core/frontend-devtools/README.md
markschlosseratbentley, line 1: Using display-test-app, when I recall a saved view where I have imodel-draping set up, the draping does not happen until I do some more model map layer attachments.

I have attached a test dataset to this PR ([comment here](https://github.com/iTwin/itwinjs-core/pull/7900#issuecomment-2769378659)).

To reproduce, open the .bim in DTA and recall the draping saved view in the ESV json file.

markschlosseratbentley, line 1: Once, when recalling saved views, I managed to get the "bad texture" black and white checkboard pattern overlaid on the iModel instead of good draping.

markschlosseratbentley (dismissed):

pmconne (commented):
What is the plan for making this draping work on `BatchedTileTree`s (the ones produced by mesh export service)? Currently you have a `LayerTreeHandler` per single-model tile tree, and multiple checks against model Id, neither of which will work for tile trees that batch multiple models together.

markschlosseratbentley:
> What is the plan for making this draping work on `BatchedTileTree`s (the ones produced by mesh export service)? Currently you have a `LayerTreeHandler` per single-model tile tree, and multiple checks against model Id, neither of which will work for tile trees that batch multiple models together.

You are right - `BatchedTileTree` seems to be a shortcoming of this approach. @danielzhong-bentley and I will need to dig in further to this topic. Do you have any recommendations or immediate suggestions for achieving similar support there?

pmconne:
> You are right - BatchedTileTree seems to be a shortcoming of this approach. @danielzhong-bentley and I will need to dig in further to this topic. Do you have any recommendations or immediate suggestions for achieving similar support there?

Not really...I think you're going to have to conditionally apply the drape in the shader (sounds expensive...)

markschlosseratbentley:
I was thinking more this morning about imodel draping larger plans.
 
Short term:
- Maybe this PR should not allow draping onto any arbitrary model.
- Maybe it should just "drape onto all models" except itself.
- Perhaps that will be easier to achieve support in both BatchedTileTree and V1 tile trees.  We should explore if this assumption could be true.
 
I think we can come to a decision once we figure out how to get BatchedTileTree draping to at least do something as @danielzhong-bentley and I have discussed.

pmconne:
> I was thinking more this morning about imodel draping larger plans.

Your short-term plans sound reasonable.

If you simplify (perhaps temporarily) to just drape a single model on all other models except itself, the logic for batched tiles could potentially be easy. Feature Ids for batched tiles are allocated sequentially. e.g., feature Ids 1-20 might be for model X, feature Ids 21-25 for model Y, etc. So you could give the shader a pair of uniforms specifying the min and max feature Ids that should **not** have the drape applied (because they are associated with the model being draped).

Dunno how any of your plans work if there is more than one draped model map layer though...do you?

markschlosseratbentley:
Current plan to achieve support in both V1 tiles and BatchedTileTree:

- [x] Remove `drapeModels` from the API, documentation, and functionality. Not being concerned about filtering the draping to particularl models will make it much easier to achieve support in BatchedTileTree for draping. The previous work to achieve `drapeModels` support is preserved on branch `drape-onto-imodel-old`. Models draping onto themselves is not a concern right now, since the source model can be disabled from view.
- [ ] Implement the layer handler functionality that BatchedTileTree needs.
- [ ] Verify support in V4 tiles and do testing.

markschlosseratbentley:
> @markschlosseratbentley @danielzhong I see the drape models have been removed, is this now in a final state ready for review?

No, sorry, BatchedTileTree still needs work. See comment: https://github.com/iTwin/itwinjs-core/pull/7900#issuecomment-2783665960

I will mark this PR back to draft for now.

markschlosseratbentley:
@danielzhong-bentley BatchedTileTree support looks very promising!  One difference I found between V1 and V4 (batched tile trees) can be seen below:

V1 tiles:
image

V4 tiles:
image

markschlosseratbentley:
> @danielzhong-bentley BatchedTileTree support looks very promising! One difference I found between V1 and V4 (batched tile trees) can be seen below. Note how the red curving pipe is not receiving much draping from the blue slab, as it should be. (Everything drapes onto everything now).
> 
> V1 tiles: image
> 
> V4 tiles: image

Draping is now consistent between V1 and V4 tiles.

Basically, V1 still had the model not drape onto itself --- @danielzhong-bentley pushed a change to make this consistent; thanks Daniel.

danielzhong-bentley:
The PR is published and ready to review again!

core/frontend-devtools/README.md
```diff
@@ -130,7 +130,7 @@ This package provides several keyins to control the display of background maps,
...
 * `fdt attach mapoverlay <name>` - Attach an overlay map layer from name within the map layer source list.  Partial names may be used.
-* `fdt attach model maplayer [name] [globe|reality]` - Attach a model map layer for each unique model of the currently selected elements to either the globe or attached reality models. If `globe` or `reality` is not specified, the attachment will default to `globe`. If `name` is specified, use that. If `name` is specified as `0`, that is considered `undefined` or unspecified.
+* `fdt attach model maplayer [name] [globe|reality|imodel] [modelId0,modelId1,...modelIdN]` - Attach a model map layer for each unique model of the currently selected elements to either the globe, attached reality models, or iModel. If `globe`, `reality`, or `imodel` is not specified, the attachment will default to `globe`. If `name` is specified, use that. If `name` is specified as `0`, that is considered `undefined` or unspecified. `modelIds` is a list of models within an iModel to drape onto, given that `imodel` is specified as the drape target; the list is separated by commas and must have no spaces.
```
eringram, line 133: What are the correct arguments to run this command with the test dataset you provided? I am running `fdt attach model maplayer imodel 80,82,84` and not seeing anything (I got these model IDs by turning on tooltips when viewing the iModel, and converting them to decimal). And are there any other steps to reproduce the draping, like selecting models first?

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -425,6 +430,8 @@ export class IModelTileTree extends TileTree {
   public draw(args: TileDrawArgs): void {
     const tiles = this.selectTiles(args);
     this._rootTile.draw(args, tiles, this._numStaticTilesSelected);
+    if (!(args instanceof GraphicsCollectorDrawArgs))
```
pmconne, line 433: Why is this cast necessary?

core/frontend/src/core-frontend.ts
```diff
@@ -131,7 +131,7 @@ export {
   TileGraphicType, TileTreeReference,
   type TileTreeSupplier,
   TileTreeLoadStatus, TileTree,
-  Tile, TileLoadStatus, TileVisibility, TileLoadPriority, TileBoundingBoxes,
+  Tile, TileLoadStatus, TileVisibility, TileLoadPriority, TileBoundingBoxes, LayerTileTreeHandler, GraphicsCollectorDrawArgs, type MapLayerTreeSetting, LayerTileTreeReferenceHandler, compareMapLayer,
```
pmconne, line 134: core-frontend.ts defines the package's **public** API. None of these new exports you've added are part of that public API - they are all (or should all be) internal. If a given type is only used inside core-frontend, have the importing file import it from tile/internal. If a given type is used by another package in itwinjs-core (e.g., frontend-tiles), export it from cross-package.ts instead.

pmconne: Remove any release tags from these internal-only types like LayerTileTreeHandler to ensure they do not get accidentally leaked into the public API. Also put any files containing only internal types into the `core/frontend/src/internal/` directory or subdirectory (e.g., `tile/`) thereof.

core/frontend-devtools/README.md
markschlosseratbentley: `fdt attach model maplayer 0 imodel` while an element in the model you want to drape is selected.

You need to specify the [name] argument if you are specifying the globe|reality|imodel argument. `0` means `undefined`.

Based on your comment, I realized I neglected to update the documentation here -- because draping now occurs onto _**all**_ models, not a list of drape targets.  I have removed that from this doc file.

core/common/src/MapLayerSettings.ts
```diff
@@ -198,6 +198,8 @@ export enum ModelMapLayerDrapeTarget {
   Globe = 0,
   /** Drape only onto all attached reality data. */
   RealityData = 1,
+  /** Drape only onto all models within the iModel. */
```
pmconne, line 201: Did you consider making this `enum` a set of bit flags so that in the future people can say they want the model map layer draped on, e.g., both the globe and the reality data? You'd have to change `Globe` to a non-zero value (`4`, presumably) for that to work.

core/frontend/src/internal/tile/IModelTileTree.ts
danielzhong-bentley: Thanks for pointing out, just removed

core/frontend-devtools/README.md
eringram: Thanks, with that command I was able to reproduce.

I know this was in the original docs for this key-in before this PR, but the `name` argument refers to a name in the map layer source list which is to be attached, right? The wording for it is confusing. So is there any case where you would use `fdt attach model maplayer [name] imodel` with a name other than 0, or would that not make sense because you can't specify the "name" of the element you want to drape?

core/frontend/src/Viewport.ts
```diff
@@ -762,14 +762,19 @@ export abstract class Viewport implements Disposable, TileUser {
    * - Always/never drawn elements
    * - Symbology overrides.
    */
+  private _firstTimeLoaded = true;
```
pmconne, line 765: Comment above this property is not about this property.

extensions/frontend-tiles/src/BatchedTileTree.ts
```diff
@@ -86,10 +89,22 @@ export class BatchedTileTree extends TileTree {
...
 
   public override prune(): void {
     const olderThan = BeTimePoint.now().minus(this.expirationTime);
     this.rootTile.prune(olderThan);
   }
+
+  protected collectClassifierGraphics(args: TileDrawArgs, selectedTiles: Tile[]): void {
+    const classifier = args.context.planarClassifiers.get(this.modelId);
+    if (classifier)
+      classifier.collectGraphics(args.context, { modelId: this.modelId, tiles: selectedTiles, location: args.location, isPointCloud: this.isPointCloud });
```
pmconne, line 103: Was this copied from RealityTileTree.ts? Not relevant for iModel tiles.

core/frontend/src/internal/tile/PrimaryTileTree.ts
```diff
@@ -140,9 +142,31 @@ class PrimaryTreeReference extends TileTreeReference {
   private readonly _sectionClip?: StringifiedClipVector;
   private readonly _sectionCutAppearanceProvider?: FeatureAppearanceProvider;
   protected readonly _animationTransformNodeId?: number;
+  protected _classifier?: SpatialClassifierTileTreeReference;
```
pmconne, line 145: This property is never initialized. Is it here only because you copy-pasted some code from reality tiles into `addToScene`?

core/frontend/src/internal/cross-package.ts
```diff
@@ -48,6 +48,7 @@ export {
  SpatialTileTreeReferences,
  type WGS84Extent,
  WmsUtilities,
+ LayerTileTreeHandler, GraphicsCollectorDrawArgs, type MapLayerTreeSetting, LayerTileTreeReferenceHandler, compareMapLayer,
```
pmconne, line 51: I suspect now that you're removed your cast to `GraphicsCollectorDrawsArgs` from frontend-tiles, you no longer need to export that type.

common/api/core-frontend.api.md
```diff
@@ -4326,6 +4329,20 @@ export interface GraphicPrimitive2d {
     zDepth: number;
 }
 
+// @public (undocumented)
```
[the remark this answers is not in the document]
danielzhong-bentley: Updated to @internal, thanks!

core/frontend/src/Viewport.ts
```diff
@@ -762,14 +762,19 @@ export abstract class Viewport implements Disposable, TileUser {
...
   private maybeInvalidateScene(): void {
     // When shadows are being displayed and the set of displayed categories changes, we must invalidate the scene so that shadows will be regenerated.
     // Same occurs when changing feature symbology overrides (e.g., always/never-drawn element sets, transparency override)
     if (!this._sceneValid)
       return;
 
-    if (this.view.displayStyle.wantShadows || this.view.isSheetView())
+    if (this.view.displayStyle.wantShadows || this.view.isSheetView()) {
       this.invalidateScene();
+    } else if (this._firstTimeLoaded) {
```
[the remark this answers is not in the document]
danielzhong-bentley: Initially, a v4 tile is loaded as a `ProxySpatialTileTreeReference`, and once the loading completes, it is replaced with a `BatchedTileTree`. However, there was a bug where the tree would not reload correctly after it was successfully loaded.

In this function, as you can see, reloading (through `invalidateScene()`) is only triggered when `this.view.displayStyle.wantShadows` or `this.view.isSheetView()` is true. As a result, when the ProxySpatialTileTreeReference switches to a BatchedTileTree for the first time, it would not trigger a scene update, causing the tiles not to render immediately after loading.

This change fixes the issue where the Batched Tile Tree would not load and render correctly the first time.

core/frontend/src/internal/cross-package.ts
danielzhong-bentley: You are right, removed, thanks

core/frontend/src/core-frontend.ts
danielzhong-bentley: fixed

core/frontend/src/Viewport.ts
danielzhong-bentley: fixed

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -450,4 +456,13 @@ export class IModelTileTree extends TileTree {
   public get containsTransformNodes(): boolean {
     return undefined !== this._transformNodeRanges;
   }
-}
+
+  /** @internal */
+  protected collectClassifierGraphics(args: TileDrawArgs, selectedTiles: Tile[]) {
+    const classifier = args.context.planarClassifiers.get(this.modelId);
+    if (classifier)
+      classifier.collectGraphics(args.context, { modelId: this.modelId, tiles: selectedTiles, location: args.location, isPointCloud: this.isPointCloud });
```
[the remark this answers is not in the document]
danielzhong-bentley: Removed

extensions/frontend-tiles/src/BatchedTileTree.ts
danielzhong-bentley: Fixed

core/frontend/src/core-frontend.ts
pmconne: Types like LayerTileTreeHandler are still shown in the extract-api summary (did you re-run extract-api?) and still have `@internal` release tags (please remove them all, that way extract-api will complain about missing release tags if you accidentally leak them into the public API).

core/frontend/src/Viewport.ts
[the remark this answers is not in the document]
pmconne: I don't see any relationship between the calls to `maybeInvalidateScene` and the loading of a `BatchedTileTree`. It sounds like perhaps `BatchedTileTree` is supposed to be calling `viewport.invalidateScene` after it loads...but that's supposed to happen every time any tile or tile tree loads. Please investigate to find the root cause.

core/frontend/src/internal/tile/PrimaryTileTree.ts
danielzhong-bentley: Removed, thanks!

core/common/src/MapLayerSettings.ts
danielzhong-bentley: Fixed

core/frontend/src/core-frontend.ts
markschlosseratbentley: If we remove `@internal` release tags from those things, extract-api complains about missing release tags, because these things are now included in `cross-package.ts`.  Everything in `cross-package.ts` is [exported in `core-frontend.ts`](https://github.com/iTwin/itwinjs-core/blob/13f8570ae530ddcf3cb29b29ca0c18e207a714c6/core/frontend/src/core-frontend.ts#L180).  Suggestions for best path forward?

core/frontend-devtools/README.md
markschlosseratbentley: Yeah, the `name` argument was not really documented in the original version of this tool (before this PR).  [Here is how it is used in the code](https://github.com/iTwin/itwinjs-core/blob/13f8570ae530ddcf3cb29b29ca0c18e207a714c6/core/frontend-devtools/src/tools/MapLayerTool.ts#L92).  [According to this documentation](https://github.com/iTwin/itwinjs-core/blob/13f8570ae530ddcf3cb29b29ca0c18e207a714c6/core/common/src/MapLayerSettings.ts#L145), `name` is a "user-friendly name for the layer."  Is that a full enough description sense according to your experience with map layers?

We probably should call this out in the keyin, whatever the best description is.

eringram: Ah ok, I was misunderstanding what that argument is for. Then I guess I would suggest changing the sentence to:
"If `name` is specified, use that as the user-friendly name for the new map layer." just for clarity's sake (this is unrelated to model draping specifically).

core/frontend/src/Viewport.ts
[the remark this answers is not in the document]
danielzhong-bentley: I updated ProxySpatialTileTreeReferences to explicitly call invalidateScene() after replacing the proxy with the real tile tree.
This makes sure the scene reloads properly when the BatchedTileTree loads, fixing the issue without relying on maybeInvalidateScene() conditions.

eringram:
Some observations/questions from manually testing this:

1. With V1 tiles (simply opening the imodel-draping.bim posted above) when you drape some element, it has some flickering lines/triangles where the element is draped onto itself. In this screenshot, you can see the blue box draped onto the red tube looks good, but the blue box draped on itself causes lines on the box's side. In the same screenshot you can see where the red tube is draped on itself there are small triangles. Is this caused by z-fighting and possible to fix, or is that out of the scope of this PR?

Screenshot 2025-04-09 114606

image

2. Selecting an element also selects its drape/projection. Is this intentional? I can see the case for why, but was surprised by it

Screenshot 2025-04-09 114747

markschlosseratbentley:
@eringram --

> 1. With V1 tiles (simply opening the imodel-draping.bim posted above) when you drape some element, it has some flickering lines/triangles where the element is draped onto itself. In this screenshot, you can see the blue box draped onto the red tube looks good, but the blue box draped on itself causes lines on the box's side. In the same screenshot you can see where the red tube is draped on itself there are small triangles. Is this caused by z-fighting and possible to fix, or is that out of the scope of this PR?

As described in [this comment](https://github.com/iTwin/itwinjs-core/pull/7900#issuecomment-2783665960), we decided to allow self-draping (models drape onto themselves) in this PR.  This allowed us to support draping in BatchedTileTree without major complexity.  We could revisit this in the future.  The current rationale is that the user can disable the "model being draped" to exclude it from the scene (and any self-draping).  The typical use case will be "projecting" a model onto the rest of the imodel --- so the model vanishing from its original location and appearing on everything else is reasonable for now, especially considering the technical limitations of supporting both V1 and V4 tiles.

> 2. Selecting an element also selects its drape/projection. Is this intentional? I can see the case for why, but was surprised by it

This is the way attached model map layers already behaved when draping onto the globe (background map).

core/common/src/MapLayerSettings.ts
markschlosseratbentley: Yes, and I updated the _ESV.json file.  Tests are running right now.

markschlosseratbentley (changes requested):
Hold PR until Image Tests finishes.

eringram (approved):

common/api/core-common.api.md
```diff
@@ -6006,8 +6006,9 @@ export interface ModelLoadProps {
```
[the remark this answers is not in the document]
markschlosseratbentley: Image Tests has finished a full run for this branch. Most everything looks good. Draping is working in the majority of the expected views (minus known issues that @danielzhong-bentley is investigating in a separate PR).

I see no new major differences in non-draping views.

The only major problem is that the new imodel_draping saved view is not showing up with any draping in the "full run" --- because of some weird issue in Image Tests that we have been unable to reproduce locally. Running a MapLayer subset results in the draping showing up in that same saved view on the Image Tests machine.

Barring other ideas for this issue, I propose moving forward with the merge, especially since no pixel differences were detected in other situations, and the draping fully works locally with DPTA and on Image Tests in a subset test.  We should later try running the ImageTests from @mathieu-lemuzic and see if that same issue happens there or not.

markschlosseratbentley (approved):

PR #7903, opened 2025-03-31
Mark ecschema-metadata as public preview (was: beta)

rschili:
We intend to make this API stable for the 5.0 timeframe.


rschili:
@aruniverse Colin instructed me to use @public @preview here, which I did. How can we verify that it is working as intended? extract-api seems to make just public out of it...
We want this to go into the 5.0 RC, after everybody has reviewed.

aruniverse:
> @aruniverse Colin instructed me to use @public @Preview here, which I did. How can we verify that it is working as intended? extract-api seems to make just public out of it... We want this to go into the 5.0 RC, after everybody has reviewed.

We have yet to and need to update `tools/build/scripts/extract-api-summary.js` to add support for this new `@preview` tag which will add it to the csv file. We need to make sure the `*.csv` files get updated, often times I've seen only the `*.md` files are updated

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -25,7 +25,7 @@ import { ECSpecVersion, SchemaReadHelper } from "../Deserialization/Helper";
 
 /**
  * A common abstract class for all of the ECClass types.
- * @beta
+ * @public @preview
  */
 export abstract class ECClass extends SchemaItem implements CustomAttributeContainerProps {
   public static override get schemaItemType(): SupportedSchemaItemType { return AbstractSchemaItemType.Class; } // need this so getItem("name", ECClass) in schema works
```
ColinKerr, line 31: Does the static schemaItemType need to be public?  Don't we just use it internally?  Can it be marked @internal?

ColinKerr: I ended up marking all protected things internal.  We were very inconsistent with how we marked protected things.

Next question should we mark SchemaItem constructors as internal? I think so because they should only ever be called through the editor right?

rschili: Sounds reasonable. Those protected members should never be touched by anyone other than our editing API.

core/ecschema-metadata/src/Interfaces.ts
```diff
@@ -28,67 +28,67 @@ import { Unit } from "./Metadata/Unit";
...
 export type LazyLoadedConstant = LazyLoadedSchemaItem<Constant>;
-/** @beta */
+/** @public @preview */
 export type LazyLoadedPhenomenon = LazyLoadedSchemaItem<Phenomenon>;
-/** @beta */
+/** @public @preview */
 export type LazyLoadedUnitSystem = LazyLoadedSchemaItem<UnitSystem>;
-/** @beta */
+/** @public @preview */
 export type LazyLoadedFormat = LazyLoadedSchemaItem<Format>;
```
StefanApfel-Bentley, line 67: I wonder if we shouldn't change the LazyXX types to be a intersection of the payload & Promise.
Example: ```ts LazyLoadedSchema = SchemaKey & Promise<Schema>; ```

That would also allow us to keep the DelayedPromise internal and an implementation detail (in fact DelayedPromise is more or less a wrapper around a promise)

core/ecschema-metadata/src/Interfaces.ts
```diff
@@ -28,67 +28,67 @@ import { Unit } from "./Metadata/Unit";
...
 export interface WithSchemaKey {
   schemaKey: SchemaKey;
 }
 
 /** This is needed to break a circular dependency between Class and EntityClass.
- * @beta
+ * @public @preview
  */
 export interface HasMixins {
   mixins: LazyLoadedMixin[];
```
StefanApfel-Bentley, line 94: I assume we want the properties here readonly as well, in case of the references and mixin maybe also the array type.

core/ecschema-metadata/src/Metadata/CustomAttributeClass.ts
```diff
@@ -16,11 +16,12 @@ import { SchemaItem } from "./SchemaItem";
...
 /**
  * A Typescript class representation of an ECCustomAttributeClass.
- * @beta
+ * @public @preview
  */
 export class CustomAttributeClass extends ECClass {
   public override readonly schemaItemType = CustomAttributeClass.schemaItemType;
   public static override get schemaItemType() { return SchemaItemType.CustomAttributeClass; }
+  /** @internal */
   protected _appliesTo?: CustomAttributeContainerType;
```
StefanApfel-Bentley, line 25: We could make this property private. In can be accessed through the mixinClass.appliesTo getter and is set through the setAppliesTo setter method. I haven't any of our code using the field directly outside this class.

core/ecschema-metadata/src/Metadata/EntityClass.ts
```diff
@@ -21,11 +21,12 @@ import { SchemaItem } from "./SchemaItem";
...
 /**
  * A Typescript class representation of an ECEntityClass.
- * @beta
+ * @public @preview
  */
 export class EntityClass extends ECClass implements HasMixins {
   public override readonly schemaItemType = EntityClass.schemaItemType;
   public static override get schemaItemType() { return SchemaItemType.EntityClass; }
+  /** @internal */
   protected _mixins?: LazyLoadedMixin[];
```
StefanApfel-Bentley, line 30: We could make this property private. In can be accessed through the entityClass.mixins getter and is set through the addMixin setter method. I haven't any of our code using the field directly outside this class.

StefanApfel-Bentley: This applies to a couple of more schema items. 
I don't want to repeat the same message again.

core/ecschema-metadata/src/Metadata/KindOfQuantity.ts
```diff
@@ -34,7 +37,9 @@ export class KindOfQuantity extends SchemaItem {
   /** A list of presentation formats. */
   public get presentationFormats(): Array<Format | OverrideFormat> { return this._presentationFormats; }
 
+  /** Persistence unit */
   public get persistenceUnit(): LazyLoadedUnit | LazyLoadedInvertedUnit | undefined { return this._persistenceUnit; }
+  /** @internal */
   protected set persistenceUnit(value: LazyLoadedUnit | LazyLoadedInvertedUnit | undefined) { this._persistenceUnit = value; }
```
StefanApfel-Bentley, line 43: Our usual pattern is to have a setSomething method instead of a setter property. I guess we can apply that pattern here which then also allow us to mark the set method as internal and the getter as public.

ColinKerr: That sounds good to me

core/ecschema-metadata/src/Interfaces.ts
ColinKerr: I have not thought through the implications of this but I don't really see anything wrong with your suggestion.  I'd like to hear what other people think

core/ecschema-locaters/src/SchemaXml.ts
```diff
@@ -13,7 +13,7 @@ import { Schema } from "@itwin/ecschema-metadata";
 
 /**
  * Utility class to assist in creating serialized EC Schemas on the file system.
- * @beta
+ * @public @preview
```
rschili, line 16: this seems unused other than in our tests. the Xml Locaters do not use it?

StefanApfel-Bentley: The seem to use the FileLocater for finding schemas per filename. The SchemaXml namespace seem to be only used for writing. I'd suggest to make it internal.

core/ecschema-metadata/src/Interfaces.ts
rschili: I like that change.
BTW I just found out LazyLoadedFormat is not used once. Nowhere in the entire repo.

I would prefer we get rid of all this LazyLoadedXXX stuff. I know it seems nice to call myEntity.baseClass.Name, while myEntity.baseClass is at the same time a promise that can get it. But this is totally overengineered. I did not find a single spot in our current code where this promise is actually a promise. I'll have a call with Stefan to better understand if our incremental loading makes use of this...

core/ecschema-locaters/src/SchemaXml.ts
StefanApfel-Bentley: I made the namespace internal

core/ecschema-metadata/src/Interfaces.ts
StefanApfel-Bentley: I removed the DelayedPromises from the LazyLoaded definition and tagged the DelayedPromise types internal.
Also added an issue to address that items get properly loaded asynchronously:
https://github.com/iTwin/itwinjs-backlog/issues/1467

core/ecschema-metadata/src/ECObjects.ts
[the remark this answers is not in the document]
ColinKerr: We can technically remove the deprecated methods 'early' because it was always in beta so we don't have to leave it for any fixed time period.

I saw some of these methods used outside of itwinjs-core and I thought it needed more investigation before we could say for sure they could be made internal.  schemaItemTypeToXmlString was not used outside so I marked it internal.

https://dev.azure.com/bentleycs/iModelTechnologies/_git/design-app-studio-integration?path=/itwin-services/src/services/schema-editor/ecPropertyConversion.ts&_a=contents&version=GBmaster

core/ecschema-metadata/src/ECObjects.ts
```diff
@@ -468,27 +486,27 @@ export function strengthDirectionToString(direction: StrengthDirection): string
...
     case StrengthDirection.Forward: return "Forward";
     case StrengthDirection.Backward: return "Backward";
-    default: throw new ECObjectsError(ECObjectsStatus.InvalidStrengthDirection, `An invalid StrengthDirection has been provided.`);
+    default:
+      throw new ECObjectsError(ECObjectsStatus.InvalidStrengthDirection, `An invalid StrengthDirection has been provided.`);
   }
 }
 
 /** Compares a SchemaItemType against supported type.
  * @beta
```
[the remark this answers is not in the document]
ColinKerr: as noted in the other comment I saw them used so it needs further investigation ... I made them beta so we could investigate more

core/ecschema-metadata/src/Exception.ts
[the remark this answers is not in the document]
aruniverse: Who outside of the core pkgs would be using this error?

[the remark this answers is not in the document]
ColinKerr: I dont' think ECObjectsError adds any value being public.  It's kind of grubby because

core/ecschema-metadata/src/Exception.ts
```diff
@@ -8,7 +8,7 @@
 
 import { BentleyError } from "@itwin/core-bentley";
 
-/** @beta */
+/** @public @preview */
 export enum ECObjectsStatus {
```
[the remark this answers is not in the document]
ColinKerr: Renamed and made public

ColinKerr (approved):

StefanApfel-Bentley (approved):

aruniverse (approved):
approving since its promotion to public preview;
we will need to make sure to get the tag support for preview in before official release so they show up properly in the extract-api files, but thats on my team @pankhur94

PR #7911, opened 2025-04-01
Fixed RotationalSweep.isAlmostEqual

saeeedtorabi:


core/geometry/src/solid/RotationalSweep.ts
```diff
@@ -41,14 +41,15 @@ export class RotationalSweep extends SolidPrimitive {
     super(capped);
     this._contour = contour;
     this._normalizedAxis = normalizedAxis;
-    this.capped = capped;
     this._sweepAngle = sweepAngle;
+    this.capped = capped;
```
bbastings, line 45: this._capped will already have been set by the call to super...

bbastings (approved):

core/geometry/src/solid/RotationalSweep.ts
saeeedtorabi: correct. I will remove this line

dassaf4 (approved):

PR #7913, opened 2025-04-01
Getproperty include inherited

christophermlawson:


core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -135,15 +135,15 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
    * Searches, case-insensitive, for a local ECProperty with the name provided.
    * @param name
    */
-  public async getProperty(name: string, includeInherited: boolean = false): Promise<Property | undefined> {
```
ColinKerr, line 138: We need to update the docuemntation here to explicitly call out the behavior of the flag and the default value.

Additionally we need to update the next version md file to include this in the list of breaking changes

rschili (approved):

core/ecschema-metadata/src/Metadata/Class.ts
christophermlawson: Update the code docs and nextversion.md

StefanApfel-Bentley (approved):

core/ecschema-editing/src/Editing/Properties.ts
```diff
@@ -38,7 +38,7 @@ export class Properties {
         throw new SchemaEditingError(ECEditingStatus.SetPropertyName, new PropertyId(this.ecClassType, classKey, propertyName), e);
       });
 
-    const baseProperty = await existingProperty.class.getProperty(newPropertyName, true) as MutableProperty;
+    const baseProperty = await existingProperty.class.getProperty(newPropertyName) as MutableProperty;
     if (baseProperty) {
```
ColinKerr, line 42: baseProperty is a very confusing name here.  This is a property with the new name, either from this class or from a base class.

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -133,7 +133,8 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
 
   /**
    * Searches, case-insensitive, for a local ECProperty with the name provided.
```
ColinKerr, line 135: 'local ECProperty' implies one defined on this class and not one inherited ... but the default is to get the property local or inherited.

core/ecschema-editing/src/Editing/Properties.ts
christophermlawson: Yeah, that was confusing. Fixed.

ColinKerr (approved):

PR #7917, opened 2025-04-02
Core Backend : ES Modules

pankhur94:


pankhur94:
We changed our core-bentley package to module system, while re building the repo, Bentley package builds successfully, however, some depenedent package fail to build with the following issue
I tried providing different types file for both cjs and esm build using export map, 
"exports": {
    ".": {
      "require": {
        "types": "./lib/cjs/core-bentley",
        "default": "./lib/cjs/core-bentley.js"
      },
      "import": {
        "types": "./lib/esm/core-bentley",
        "default": "./lib/esm/core-bentley.js"
      }
    }
  }, but the issue still persists.
Anyone have any idea what could this be?

mayank99:
I'm not familiar with this codebase, but it looks like a TypeScript issue. The packages that are consuming `@itwin/core-bentley` are expecting some types (likely CJS) that don't match what is being imported (likely ESM).

grigasp:
> We changed our core-bentley package to module system, while re building the repo, Bentley package builds successfully, however, some depenedent package fail <..>

I'm not seeing this issue with `presentation-backend`, but I do with `presentation-frontend` package, and it's also related to `BeEvent` as yours:
```
src/test/PresentationManager.test.ts:97:63 - error TS2322: Type 'import("/home/grigas/itwinjs-core/core/bentley/lib/esm/BeEvent", { with: { "resolution-mode": "import" } }).BeEvent<(_imodel: import("/home/grigas/itwinjs-core/core/frontend/lib/cjs/IModelConnection").IModelConnection) => void>' is not assignable to type 'import("/home/grigas/itwinjs-core/core/bentley/lib/cjs/BeEvent").BeEvent<(_imodel: import("/home/grigas/itwinjs-core/core/frontend/lib/cjs/IModelConnection").IModelConnection) => void>'.
  Types have separate declarations of a private property '_listeners'.
```

From the paths in the error you can clearly see that `BeEvent` that tests use is from `esm`, but `IModelConnection` and its `onClose` `BeEvent` type are coming from `cjs`. And the tests are trying to assign an instance of ESM `BeEvent` to a property that is CJS `BeEvent`. From TS perspective these are 2 different types, and it rightfully complains.

I think this is happening because, with your changes, `core-bentley` package now properly supports ESM, and that makes our tests use ESM instead of CJS. On the other hand, `core-frontend` doesn't have proper ESM support, and the tests still use its CJS version, and all its dependencies are also from CJS, including `BeEvent` from `core-bentley`.

So I think you'll have to start from the top and then move deeper into the stack. E.g. `core-bentley` can be fixed only when you fix all packages that have dependency on it.

pankhur94:
> > We changed our core-bentley package to module system, while re building the repo, Bentley package builds successfully, however, some depenedent package fail <..>
> 
> I'm not seeing this issue with `presentation-backend`, but I do with `presentation-frontend` package, and it's also related to `BeEvent` as yours:
> 
> ```
> src/test/PresentationManager.test.ts:97:63 - error TS2322: Type 'import("/home/grigas/itwinjs-core/core/bentley/lib/esm/BeEvent", { with: { "resolution-mode": "import" } }).BeEvent<(_imodel: import("/home/grigas/itwinjs-core/core/frontend/lib/cjs/IModelConnection").IModelConnection) => void>' is not assignable to type 'import("/home/grigas/itwinjs-core/core/bentley/lib/cjs/BeEvent").BeEvent<(_imodel: import("/home/grigas/itwinjs-core/core/frontend/lib/cjs/IModelConnection").IModelConnection) => void>'.
>   Types have separate declarations of a private property '_listeners'.
> ```
> 
> From the paths in the error you can clearly see that `BeEvent` that tests use is from `esm`, but `IModelConnection` and its `onClose` `BeEvent` type are coming from `cjs`. And the tests are trying to assign an instance of ESM `BeEvent` to a property that is CJS `BeEvent`. From TS perspective these are 2 different types, and it rightfully complains.
> 
> I think this is happening because, with your changes, `core-bentley` package now properly supports ESM, and that makes our tests use ESM instead of CJS. On the other hand, `core-frontend` doesn't have proper ESM support, and the tests still use its CJS version, and all its dependencies are also from CJS, including `BeEvent` from `core-bentley`.
> 
> So I think you'll have to start from the top and then move deeper into the stack. E.g. `core-bentley` can be fixed only when you fix all packages that have dependency on it.

We are actually moving the other way round, for ex, we start with the package with least dependencies i.e. core-bentley, core-geometry and upwards. But yes, I do agree that the reason you explained is correct. As soon as well convert all the dependencies of presentation-backend i.e. core-bentley, core-backend etc to esm, the issue is resolved.

pankhur94:
completed in https://github.com/iTwin/itwinjs-core/pull/7866

PR #7928, opened 2025-04-04
Move HubMock to internal and delete CloudSqliteMock

tcobbs-bentley:
Move HubMock to the internal directory and stop exporting it from core-backend. iTwin Studio will now need to do a deep import in their tests:

```ts
import { HubMock } from "@itwin/core-backend/lib/cjs/internal/HubMock";
```

Refactor CheckpointManager and HubMock such that CloudSqliteMock.ts is no longer needed.


kabentley:
Frankly, neither one should be part of the core package. SamW just added HubMock because he needed it and didn't have any other way. I think there should be a "backend-test" package or something.

tcobbs-bentley:
> Frankly, neither one should be part of the core package. SamW just added HubMock because he needed it and didn't have any other way. I think there should be a "backend-test" package or something.

@aruniverse feels that moving HubMock into the `test` directory (which isn't published) and updating our tests to import it with a deep link would be a better solution.

kabentley:
> @aruniverse feels that moving HubMock into the test directory (which isn't published) and updating our tests to import it with a deep link would be a better solution.

But it is used by other projects for *their* tests.

pmconne:
Please strongly consider putting it in src/internal/.
(We should also put HubMock.ts there and stop exporting it from core-backend.ts - these mocks should not be part of the package's public API).

kabentley:
> Please strongly consider putting it in src/internal/.

That's equivalent to where it started. Then, when people started working on ITwinStudio apps, they needed a way to use these Mocks to test *their* apps. Hence the current mess.

tcobbs-bentley:
> > Please strongly consider putting it in src/internal/.
> 
> That's equivalent to where it started. Then, when people started working on ITwinStudio apps, they needed a way to use these Mocks to test _their_ apps. Hence the current mess. I don't know the solution, because if it is a separate package it will depend on core-backend, but it is used by core-backend's build.

Given that HubMock _is_ internal, why is requiring iTwin Studio apps to do a deep import from internal unacceptable?

kabentley:
> Given that HubMock is internal, why is requiring iTwin Studio apps to do a deep import from internal unacceptable

I confess I don't even know what that means, but having iTwinStudio depend on something internal from core doesn't sound right to me. 

I just want to make everyone aware that HubMock is used by iTS (but only by iTS, to be clear), so it can't "just be part of core-backend's internal test framework".

tcobbs-bentley:
> > Given that HubMock is internal, why is requiring iTwin Studio apps to do a deep import from internal unacceptable
> 
> I confess I don't even know what that means, but having iTwinStudio depend on something internal from core doesn't sound right to me.
> 
> I just want to make everyone aware that HubMock is used by iTS (but only by iTS, to be clear), so it can't "just be part of core-backend's internal test framework".

I mean that instead of importing HubMock directly from core-backend, it would instead be imported like this:

```ts
import { HubMock } from "@itwin/core-backend/lib/cjs/internal/HubMock"
```

kabentley:
> import { HubMock } from "@itwin/core-backend/lib/cjs/internal/HubMock"

Yuk.

My point is that HubMock is only useful for tests, it should *not* be delivered with the core-backend package. Sam added it to the public api for convenience over my objections so iTwinStudio could use it.

I don't understand what "CloudSqliteMock" is, nor why HubMock should depend on it (it looks to me like it should be called "CheckpointMock"?)  But it clearly can not be a part of the core package since it depends on "sinon". There should be a new package "backend-tests" that can be included with a dev dependency. The tricky part is that the tests in core-backend use HubMock, hence it's original location in its test directory.

tcobbs-bentley:
> Please strongly consider putting it in src/internal/. (We should also put HubMock.ts there and stop exporting it from core-backend.ts - these mocks should not be part of the package's public API).

I have done as you suggest, and moved both into `internal`. I also updated `CloudSqliteMock` so that it only loads `sinon` at run-time if you call `CloudSqliteMock.startup` (called from `HubMock.startup`), so it should only depend on `sinon` in cases where `sinon` is already present.

kabentley:
@tcobbs-bentley I just pushed a change to eliminate CloudSqliteMock and the use of sinon in core-backend. I still would prefer to find a way to move HubMock out of core-backend entirely, but that would take more work.

BTW, since there is no way to upload checkpoints, I don't see any way tests can use the "openCheckpoint" path at all other than for V0, and I don't see any tests doing that. Did you have a way to test it?

tcobbs-bentley:
> @tcobbs-bentley I just pushed a change to eliminate CloudSqliteMock and the use of sinon in core-backend. I still would prefer to find a way to move HubMock out of core-backend entirely, but that would take more work.

Thanks. That's obviously a better solution.

> BTW, since there is no way to upload checkpoints, I don't see any way tests can use the "openCheckpoint" path at all other than for V0, and I don't see any tests doing that. Did you have a way to test it?

I'm not sure what you mean by this. `full-stack-tests/backend/src/integration/Checkpoints.test.ts` has four calls to `SnapshotDb.openCheckpoint` (which pass in this branch), and I can't find any other classes that have an `openCheckpoint` function.

kabentley:
> full-stack-tests/backend/src/integration/Checkpoints.test.ts has four calls to SnapshotDb.openCheckpoint

But those tests don't use HubMock, right?

kabentley:
HubMock changes the implementation of `V2CheckpointManager.attach` (via _mockCheckpointAttach). That's only used by `SnapshotDb.openCheckpoint`. No tests that use HubMock ever attempts to open a checkpoint (of course that was also true of the previous implementation before you removed V1 checkpoints).

That's ok, I really don't think anyone using HubMock can do much with Checkpoints since there's no way to upload them other than with V0. But, there could be a test that tries opening V0.

aruniverse (approved):

kabentley:
But that test doesn't use HubMock either (even though for some reason it stubs in HubMock IModelHost.hubAccess.)

It never calls HubMock.startup/shutdown. I can't tell why your previous implementation would have be called, but it certainly doesn't use this implementation. It "fakes" a SnapshotDb and uses nonsense for the ChangesetId so HubMock wouldn't know what to do with it.

PR #7933, opened 2025-04-07
Add CatalogIModels api

kabentley:
Adds support for creating, versioning, editing, and opening CatalogIModels.
Could use some help creating a learning article for this. I'm not going to do that.
Also moved semver versioning from workspace API to `CloudSqlite`.
Also changed a bunch of error types into `ITwinError`s.


calebmshafer:
thanks for the PR @kabentley! I haven't had a chance most of this week to review yet - I'll take some dedicated time this afternoon to review.

core/backend/src/CatalogIModel.ts
```diff
@@ -0,0 +1,327 @@
...
+    if (args.prefetch)
+      CloudSqlite.startCloudPrefetch(container, dbFullName);
+
+    return EditableCatalogDb.openFile(dbFullName, OpenMode.ReadWrite, { container, ...args }) as EditableCatalogDb;
+  }
+
+  /**
+   * Open a CatalogIModel for read access.
+   */
+  export async function openReadonly(args: CatalogIModelTypes.OpenArgs): Promise<StandaloneDb & ReadCatalog> {
```
pmconne, line 267: I'm curious about the motivation for exposing this as `StandaloneDb & ReadCatalog` vs just exposing `CatalogDb` as a subclass of `StandaloneDb`. Can somebody do something useful with a `ReadCatalog` by itself?

kabentley: I didn't want to export `CatalogDb` so nobody could create instances of it (or `EditableCatalogDb`) directly.

pmconne: If that's important on the back-end, why does the front-end just provide `CatalogConnection` as a subclass of `BriefcaseConnection`?
It looks like if I opened a `CatalogConnection` as read-only and called its `updateCatalogManifest` method, I would get a very confusing error from the backend (via `CatalogIModelHandler`) about `updateCatalogManifest` being `undefined`. (`findCatalogByKey` asserts the result is an `EditCatalog` but it's only a `CatalogDb` if opened read-only).

pmconne: See if you like [this](https://github.com/iTwin/itwinjs-core/pull/7933/commits/c60094e9dbf4c618d503ce1ae14de4a3d0db54cf) better for backend. `CatalogIModel` extends *only* the non-static *interface* of `StandaloneDb` and adds catalog-specific methods.

pmconne: Made similar changes on front-end.

pmconne (approved):

pmconne:
@kabentley what remains to be done for this to merge? @Josh-Schifter @williamkbentley have you had a chance to take a look?

core/frontend/src/NativeApp.ts
```diff
@@ -66,6 +66,7 @@ export class NativeApp {
 
   /** A Proxy to call one of the [NativeAppFunctions]($common) functions via IPC. */
   public static nativeAppIpc = IpcApp.makeIpcProxy<NativeAppFunctions>(nativeAppIpcStrings.channelName);
+  public static catalogIpc = IpcApp.makeIpcProxy<CatalogIModel.IpcMethods, CatalogIModel.IpcChannel>("catalogIModel/ipc");
```
williamkbentley, line 69: should prob be `CatalogIModel.IpcChannel` from common

full-stack-tests/core/src/frontend/standalone/CatalogConnection.test.ts
```diff
@@ -0,0 +1,192 @@
...
+      await expect(CatalogConnection.createNewContainer({ localCatalogFile, version: "1.0.0", iTwinId, manifest, metadata })).rejectedWith("only admins may create containers");
+
+      await coreFullStackTestIpc.setAzTestUser("admin");
+      // verify that illegal dbNames are rejected
+      await expect(CatalogConnection.createNewContainer({ localCatalogFile, dbName: "a:b", version: "1.0.0", iTwinId, manifest, metadata })).rejectedWith("invalid dbName");
+      // verify that an illegal initial version is rejected
+      await expect(CatalogConnection.createNewContainer({ localCatalogFile, version: "not a version", iTwinId, manifest, metadata })).rejectedWith("invalid version specification");
+
+      // create a container for our tests. The initial version is supplied as "1.0.0"
+      const newContainer = await CatalogConnection.createNewContainer({ localCatalogFile, version: "1.0.0", iTwinId, manifest, metadata });
```
williamkbentley, line 67: Will this create a similar problem to workspaceDbs where now v1.0.0 is both empty and immutable?

core/backend/src/CatalogDb.ts
```diff
@@ -0,0 +1,333 @@
...
+      nativeDb.resetBriefcaseId(BriefcaseIdValue.Unassigned); // catalogs should always be unassigned
+      nativeDb.saveChanges(); // save change to briefcaseId
+      nativeDb.vacuum();
+      nativeDb.closeFile();
+    } catch (e: unknown) {
+      CatalogError.throwError("invalid-seed-catalog", { message: "Illegal seed catalog", ...args, cause: e });
+    }
+
+    const userToken = await IModelHost.getAccessToken();
+    // create tne new container from the blob service, requires "admin" authorization
```
williamkbentley, line 194: typo

williamkbentley (commented):
So CatalogIModels can only live in the BlobContainer.ContainerService such as the Storage Container Service?

kabentley:
> So CatalogIModels can only live in the BlobContainer.ContainerService such as the Storage Container Service?

Yes. Or in local files (mainly for testing), just like WorkspaceDbs.

core/frontend/src/NativeApp.ts
kabentley: `CatalogIModel.IpcChannel` is a type, not a string. It is used as the second argument to `makeIpcProxy`. This is a new approach to avoid adding objects to the common package.

full-stack-tests/core/src/frontend/standalone/CatalogConnection.test.ts
kabentley: When you call `createNewContainer` to create a new container for a catalog, you upload the initial version of it. That doesn't have to be empty. 

Then, you can supply the initial version number. The default is 0.0.0, which is a "prerelease" version. Like WorkspacesDbs, prerelease versions *are* editable.

This test supplies 1.0.0 to verify that you can override the default.

williamkbentley: ok cool

PR #7937, opened 2025-04-08
Deprecate `DialogModuleMethod` in core-electron

GytisCepk:
All methods using this were deprecated in https://github.com/iTwin/itwinjs-core/pull/4533 and removed recently.

### DIFF
--- a/core/electron/src/common/ElectronIpcInterface.ts
+++ b/core/electron/src/common/ElectronIpcInterface.ts
@@ -15,5 +15,6 @@ export const electronIpcStrings = {
 
 /** Asynchronous methods of dialog module in an Electron app.
  * @beta
+ * @deprecated in 5.0. Use `Electron.Dialog` from `electron` instead.
  */
 export type DialogModuleMethod = AsyncMethodsOf<Electron.Dialog>;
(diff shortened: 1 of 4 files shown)


ben-polinsky (approved):

core/electron/src/common/ElectronIpcInterface.ts
```diff
@@ -15,5 +15,6 @@ export const electronIpcStrings = {
 
 /** Asynchronous methods of dialog module in an Electron app.
  * @beta
+ * @deprecated in 5.0. Use `Electron.Dialog` from `electron` instead.
```
aruniverse, line 18: Since the API is beta, should we just remove it?

GytisCepk: I don't think anyone is using this, so no one would notice it's removal. 

That being said, I would like to follow our deprecated -> remove process as much as possible. As this is just a TS type, there is no maintenance required here.

aruniverse: Ok thats fine, but just keep in mind the deprecation policy is targeted mainly towards public apis

PR #7951, opened 2025-04-10
Fix image names in integration-validation.yaml

GytisCepk:
Node 22 Windows and Linux were incorrectly run on MacOS machines.

### DIFF
--- a/common/config/azure-pipelines/integration-validation.yaml
+++ b/common/config/azure-pipelines/integration-validation.yaml
@@ -31,10 +31,10 @@ jobs:
           imageName: macos-latest
           nodeVersion: 20.x
         Linux_node_22_x:
-          imageName: macos-latest
+          imageName: ubuntu-latest
           nodeVersion: 22.x
         Windows_node_22_x:
-          imageName: macos-latest
+          imageName: windows-latest
           nodeVersion: 22.x
         MacOS_node_22_x:
           imageName: macos-latest
--- a/common/config/azure-pipelines/integration-validation.yaml
+++ b/common/config/azure-pipelines/integration-validation.yaml
@@ -23,23 +23,30 @@ jobs:
       matrix:
         Linux_node_20_x:
           imageName: ubuntu-latest
+          poolName:
           nodeVersion: 20.x
         Windows_node_20_x:
           imageName: windows-latest
+          poolName:
           nodeVersion: 20.x
         MacOS_node_20_x:
           imageName: macos-latest
+          poolName: "iModelTechMacArm"
           nodeVersion: 20.x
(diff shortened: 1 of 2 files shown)


GytisCepk:

GytisCepk:
Run failures on MacOS are due to pipeline still using x64 architecture. @DanRod1999, could look into that ?

common/config/azure-pipelines/integration-validation.yaml
```diff
@@ -31,10 +31,10 @@ jobs:
...
         Linux_node_22_x:
-          imageName: macos-latest
+          imageName: ubuntu-latest
           nodeVersion: 22.x
         Windows_node_22_x:
-          imageName: macos-latest
+          imageName: windows-latest
           nodeVersion: 22.x
         MacOS_node_22_x:
           imageName: macos-latest
```
DanRod1999, line 40: Right now i believe we cannot use vm images for mac as arm processors are not part of the microsoft hosted agents I think
https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#use-a-microsoft-hosted-agent

So we will probably need to restructure this pipeline a little

DanRod1999: saw this in docs so trying it out so we can still use matrix with different builds
https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/jobs-job-strategy?view=azure-pipelines#build-on-multiple-platforms-using-self-hosted-and-microsoft-hosted-agents

DanRod1999: I think it worked, but my test run is failing on most jobs right now. Don't think its from me but not sure
https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3835863&view=results

tcobbs-bentley: Given that things seemed to work before, and this PR is intended to fix the Ubuntu update that breaks all PRs, I feel that the Mac fixes should be held off on. (Fixing the Linux and Windows jobs to not use macOS seems fine.)

tcobbs-bentley (approved):

common/config/azure-pipelines/integration-validation.yaml
DanRod1999: all integration tests passed
https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3836400&view=results
appears to be some flakey tests though, so we should look at that closer

DanRod1999 (approved):

Issue #7919, opened 2025-04-03
`ECClass.getProperties()` returns empty list


grigasp:
**Describe the bug**

Then `ECClass.getProperties()` is called multiple times without waiting for the first call to resolve, the subsequent calls return empty list.

**To Reproduce**

1. Apply this diff on https://github.com/iTwin/itwinjs-core/commit/2ce423e03127f15636c096d454dd9a9ba3e704ea:
   ```diff
   diff --git a/core/ecschema-metadata/src/test/Metadata/Class.test.ts b/core/ecschema-metadata/src/test/Metadata/Class.test.ts
   index 2f74cc8891..afb6cd1e29 100644
   --- a/core/ecschema-metadata/src/test/Metadata/Class.test.ts
   +++ b/core/ecschema-metadata/src/test/Metadata/Class.test.ts
   @@ -44,7 +44,7 @@ describe("ECClass", () => {
          schema = new Schema(new SchemaContext(), "TestSchema", "ts", 1, 0, 0);
        });
   
   -    it("inherited properties from base class", async () => {
   +    it.only("inherited properties from base class", async () => {
          const baseClass = new EntityClass(schema, "TestBase");
          const basePrimProp = await (baseClass as ECClass as MutableClass).createPrimitiveProperty("BasePrimProp");
   
   @@ -52,6 +52,10 @@ describe("ECClass", () => {
          await (entityClass as ECClass as MutableClass).createPrimitiveProperty("PrimProp");
          entityClass.baseClass = new DelayedPromiseWithProps(baseClass.key, async () => baseClass);
   
   +      const [res1, res2] = await Promise.all([entityClass.getProperties().then((x) => [...x]), entityClass.getProperties().then((x) => [...x])]);
   +      expect(res1).to.have.lengthOf(2);
   +      expect(res2).to.have.lengthOf(2);
   +
          expect(await entityClass.getProperty("BasePrimProp")).to.be.undefined;
          expect(await entityClass.getProperty("BasePrimProp", false)).to.be.undefined;
          expect(await entityClass.getProperty("BasePrimProp", true)).equal(basePrimProp);
   ```
 
2. Run `ecschema-metadata` tests.
3. Notice the test fails with:
   ```
  1) ECClass
       get properties
         inherited properties from base class:

      AssertionError: expected [] to have a length of 2 but got +0
      + expected - actual

      -0
      +2

      at Context.<anonymous> (src\test\Metadata\Class.test.ts:57:28)
   ```

[...]

PR #7961, opened 2025-04-11
Fix for ECClass.getProperties() which returns empty list when called multiple times [release/4.11.x]

Abhijeet-Bentley:
Issue Link: https://github.com/iTwin/itwinjs-core/issues/7919
This is already fixed on master by an independent change

- Introduced a _propertyCachePromise field to track ongoing cache builds
- buildPropertyCache() now builds the merged property list in isolation and returns it instead of mutating _mergedPropertyCache
- Final result is only assigned to _mergedPropertyCache after a successful build
- Ensures that only one build processes run at a time

### DIFF
--- a/core/ecschema-metadata/src/test/Metadata/Class.test.ts
+++ b/core/ecschema-metadata/src/test/Metadata/Class.test.ts
@@ -19,7 +19,6 @@ import { SchemaKey } from "../../SchemaKey";
 import { createSchemaJsonWithItems } from "../TestUtils/DeserializationHelpers";
 import { createEmptyXmlDocument, getElementChildren, getElementChildrenByTagName } from "../TestUtils/SerializationHelper";
 import { StrengthDirection } from "../../ECObjects";
-import { Property } from "../../ecschema-metadata";
 
 /* eslint-disable @typescript-eslint/naming-convention */
 
@@ -60,7 +59,7 @@ describe("ECClass", () => {
       expect(await entityClass.getInheritedProperty("PrimProp")).to.be.undefined;
     });
 
-    it.only("successfully executes multiples calls to getProperties", async() => {
+    it("successfully executes multiples calls to getProperties", async() => {
       const schemaJson = {
         $schema: "https://dev.bentley.com/json_schemas/ec/32/ecschema",
         name: "TestSchema",


core/ecschema-metadata/src/test/Metadata/Class.test.ts
```diff
@@ -59,6 +59,31 @@ describe("ECClass", () => {
...
+      await (baseClass as ECClass as MutableClass).createPrimitiveProperty("ParentPrimProp2");
+      await (baseClass as ECClass as MutableClass).createPrimitiveProperty("ParentPrimProp3");
+
+      const entityClass = new EntityClass(schema, "Child");
+      await (entityClass as ECClass as MutableClass).createPrimitiveProperty("ChildPrimProp");
+      await (entityClass as ECClass as MutableClass).createPrimitiveProperty("ChildPrimProp2");
+      await (entityClass as ECClass as MutableClass).createPrimitiveProperty("ChildPrimProp3");
+      entityClass.baseClass = new DelayedPromiseWithProps(baseClass.key, async () => baseClass);
+
+      const [res1,res2] = await Promise.all([entityClass.getProperties().then((x)=>[...x]), entityClass.getProperties().then((x)=>[...x]), entityClass.getProperties().then((x)=>[...x])]);
```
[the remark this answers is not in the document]
Abhijeet-Bentley: That is a blunder on my part. I have fixed it.

common/api/ecschema-metadata.api.md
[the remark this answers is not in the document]
Abhijeet-Bentley: Yes, it is.

common/changes/@itwin/ecschema-metadata/abhi-ecclass-getProperties-issue_2025-04-11-12-07.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-metadata",
+      "comment": "The method buildPropertyCache() now return a Property[] instead of being void and no longer has result: Property[] and existingValues: Map<string,number> as params.",
```
grigasp, line 5:
```suggestion
      "comment": "**BREAKING:** The signature of protected method `ECClass.buildPropertyCache()` has been changed to take less arguments and return a `Promise<Property[]>`.",
```

grigasp (approved):

core/ecschema-metadata/src/test/Metadata/Class.test.ts
```diff
@@ -59,6 +59,34 @@ describe("ECClass", () => {
       expect(await entityClass.getInheritedProperty("PrimProp")).to.be.undefined;
     });
 
+    it("successfully executes multiples calls to getProperties", async() => {
+      const baseClass = new EntityClass(schema, "Parent");
```
rschili, line 63: We discussed in our daily call that construction of a schema is better done from JSON instead of doing it like this. Please change this test even if there are others in the same file which do this.

rschili:
Hold on a second. You want to merge this into release/4.11.x, but the bug was reported for 4.10.2, which will be unaffected by the change.
The normal workflow is to implement fixes on the master branch and once those are merged, we use mergify to backport them to 4.11 and potentially 4.10. At that stage management often decides whether we want to backport and how far.

Directly fixing something on 4.11 without a fix on master is not right. Please fix this on master first and then backport + adjust fixes for the older releases. This needs to be coordinated with Colin and Arun.

rschili (approved):

grigasp:
> You want to merge this into release/4.11.x, but the bug was reported for 4.10.2, which will be unaffected by the change.

Just to make this clear, I'm fine with the fix landing on `4.11` rather than `4.10`.

common/changes/@itwin/ecschema-metadata/abhi-ecclass-getProperties-issue_2025-04-11-12-07.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-metadata",
+      "comment": "**BREAKING:** The signature of protected method `ECClass.buildPropertyCache()` has been changed to take less arguments and return a `Promise<Property[]>`.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/ecschema-metadata"
+}
```
StefanApfel-Bentley, line 10: Since this is a protected method that we marked as internal api (in the other PR), I don't think we should add a message for that to the changelog. I don't think we want to allow callers outside this package to re-build the property cache.

grigasp:
@StefanApfel-Bentley, please note this is targeting the `release/4.11.x` branch.

PR #7962, opened 2025-04-11
Changed schema item members on Format and KindOfQuantity to be lazy loaded

StefanApfel-Bentley:
Formats, OverrideFormats and KindOfQuantities still had some members that referred to other schema items which were not lazy loadable. In the two formats this affected only the units, kind of quantites has the presentationformats that are not lazy loadable.

This PR changes the type in the api and adjusted the callers to provide lazy loadable items.
Note: It basically only updates the public api and uses the implementation as it is now, there is upcoming work, covered in this issue: https://github.com/iTwin/itwinjs-backlog/issues/1467


grigasp (approved):
approving the change in `presentation-full-stack-tests`

ColinKerr (approved):

PR #7963, opened 2025-04-11
Fix Integration Pipeline to Support Node 18 Tests

DanRod1999:
While backporting [Fix image names in integration-validation.yaml](https://github.com/iTwin/itwinjs-core/pull/7957/files)
Forgot to include changes to Node 18 runs which are removed in master branch for 5.0

Pipeline was broken one this branch, but is now running as intended
https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3839252&view=results

### DIFF
--- a/common/config/azure-pipelines/integration-validation.yaml
+++ b/common/config/azure-pipelines/integration-validation.yaml
@@ -23,38 +23,38 @@ jobs:
       matrix:
         Linux_node_18_x:
           imageName: ubuntu-latest
-          poolName: # Leaving blank defaults to image. Excluding field will cause errors in yml
+          poolName: # Leaving blank will default to image. Excluding field will cause errors in yml
           nodeVersion: 18.x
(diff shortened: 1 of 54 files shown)


common/config/azure-pipelines/integration-validation.yaml
```diff
@@ -23,35 +23,38 @@ jobs:
       matrix:
         Linux_node_18_x:
           imageName: ubuntu-latest
+          poolName: # Leaving blank defaults to image. Excluding field will cause errors in yml
```
DanRod1999, line 26: adding this to every empty field is ugly. Does only adding it to first field seem sufficient?

hl662: yeah, the first occurrence of imageName and poolName should be enough

hl662 (approved):

aruniverse (approved):

PR #7967, opened 2025-04-14
Use readInstance() and Deserialize Entity Json Directly in TS

MichaelSwigerAtBentley:


MichaelSwigerAtBentley:
Note: 
While testing, there are some cases during the core backend tests where the old call to getElement() would return very large bbox values (1e200) that where not stored in the ECSql Instance. These large bbox values are not returned when reading the instance with readInstance() unlike other bbox values. The absence of these values do not result in any test failures, but it is inconsistent.

core/backend/src/test/imodel/IModel.test.ts
```diff
@@ -2668,7 +2668,7 @@ describe("iModel", () => {
     expect(elProps.code.scope).equal(element.code.scope);
     expect(elProps.code.spec).equal(element.code.spec);
     expect(elProps.federationGuid).equal(element.federationGuid);
-    expect((elProps as any).isPrivate).undefined;
+    expect((elProps as any).isPrivate).to.be.false;
```
MichaelSwigerAtBentley, line 2671: I noticed that for boolean values like 'isPrivate', the value is only returned if it is true, otherwise it is just never set (so not even undefined is returned even if isPrivate is set to false). However, for ViewDefinition, DefinitionElement and the classes that extend it, 'isPrivate' is returned as true and false. Except in this one test case, where it expected it to be undefined, even though it is false.

What makes ViewDefinition and DefinitionElement 'isPrivate' the exeception? Is there some additional information that is not being returned in the instance? Are there any more that might not be caught by our tests?

rschili:
> @rschili @khanaffan @ColinKerr Does anyone know why these values would be returned by the original native getElement() call but not present on the readInstance() instance itself?

It's probably some special treatment in the depths of the Dgn Platform native layer which was removed. If there is time left today I can pull the main branch and try to fix the piece of code responsible

rschili:
Just posting this as a reminder. The changes and examples from this PR description should probably go into NextVersion.md once we merge this.

common/config/rush/browser-approved-packages.json
```diff
@@ -374,6 +374,10 @@
       "name": "@svgdotjs/svg.js",
       "allowedCategories": [ "frontend" ]
     },
+    {
+      "name": "@type/json-diff",
```
MichaelSwigerAtBentley, line 378: This is left over from testing, does this need to be removed?

core/backend/src/TxnManager.ts
```diff
@@ -481,6 +481,8 @@ export class TxnManager {
   protected _onAfterUndoRedo(isUndo: boolean) {
     this.touchWatchFile();
     this.onAfterUndoRedo.raiseEvent(isUndo);
+    this._iModel.elements.cache.clear();
```
MichaelSwigerAtBentley, line 484: @khanaffan I see that the caches are cleared on these specific Txn events, why do we not clear the cache on the _onGeometryGuidsChanged or _onCommited events as well? Wouldn't the model be changing in those cases too?

common/config/rush/browser-approved-packages.json
ColinKerr: It allows the package dot be consumed by other packages in the 'allowedCategories'.  So if this is not there rush will complain if it finds this in the backend's package.json.  I thought it differentiated between dev and normal dependencies, e.g. dev dependencies don't need to be approved, but I could be wrong about that.

core/backend/src/Entity.ts
```diff
@@ -7,10 +7,37 @@
...
+}
+
+/** Set of properties that are used to deserialize an [[EntityProps]] from an ECSqlRow.
+ * @beta */
+export interface InstanceProps {
+  readonly row: ECSqlRow;
+  readonly iModel: IModelDb;
+  readonly options?: {
+    readonly element?: ElementLoadOptions;
+  }
```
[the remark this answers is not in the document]
MichaelSwigerAtBentley: In the case of CustomHandledProperty, those props should never be changed, instances should not be created, and should ideally be defined as a part of the custom class. So read only makes sense as an enforcement mech.

For InstanceProps, its more preferential. Once they are returned from readInstance() they shouldnt need to be modified, so its best to prevent it.

pmconne:
> Adds LRU caching for EntityProps in the TS layer, no caching is done on the native layer any more

You don't appear to have removed caching logic from the native side. Are you trying to convey that you're not loading elements in native any more, therefore the native caches will no longer be used/useful?

common/api/core-backend.api.md
```diff
@@ -669,11 +669,17 @@ export class Category extends DefinitionElement {
     protected constructor(props: CategoryProps, iModel: IModelDb);
     // (undocumented)
     static get className(): string;
+    // @beta (undocumented)
```
pmconne, line 672: Beta APIs need documentation. It looks like all of the ones added in this PR are undocumented.

core/backend/src/Element.ts
```diff
@@ -144,6 +143,43 @@ export class Element extends Entity {
     this.jsonProperties = { ...props.jsonProperties }; // make sure we have our own copy
   }
 
+  /** @beta */
```
pmconne, line 146: Document it.

khanaffan:
> > Adds LRU caching for EntityProps in the TS layer, no caching is done on the native layer any more
> 
> You don't appear to have removed caching logic from the native side. Are you trying to convey that you're not loading elements in native any more, therefore the native caches will no longer be used/useful?

That is correct. We have not removed any native caching it just not used during getElement()/getModel(). There are other code paths that does use native caching as they are implemented in native-addon. We need to move each of those one at a time to typescript before we are in position to remove native cache for element.

core/backend/src/LRUCaches.ts
```diff
@@ -0,0 +1,171 @@
...
+export class ElementLRUCache {
+  public static readonly DEFAULT_CAPACITY = 2000;
+  private _elementCache = new Map<Id64String, CachedElement>();
+  private _cacheByCode = new Map<Id64String, Id64String>();
+  private _cacheByFederationGuid = new Map<string, Id64String>();
+
+  private static makeCodeKey(code: CodeProps): string {
+    if (code.value)
+      return `${code.scope}:${code.value}:${code.value}`;
+    return `${code.scope}:${code.value}`;
```
pmconne, line 75: If code.value is truthy, you encode it twice in the key; otherwise you encode it once. You don't encode the spec at all.

core/backend/src/LRUCaches.ts
```diff
@@ -0,0 +1,171 @@
...
+ * A LRU cache for entities. Cache contains the ElementProps and the load options used to load it.
+ * Cache can be searched by id, code or federationGuid.
+ * @internal */
+export class ElementLRUCache {
+  public static readonly DEFAULT_CAPACITY = 2000;
+  private _elementCache = new Map<Id64String, CachedElement>();
+  private _cacheByCode = new Map<Id64String, Id64String>();
+  private _cacheByFederationGuid = new Map<string, Id64String>();
+
+  private static makeCodeKey(code: CodeProps): string {
```
pmconne, line 72: While unlikely, it is possible for this function to produce the same key for two different CodeProps. spec and value can both be arbitrary strings, which can include (for example) colon characters. Did you consider using a `Dictionary<CodeProps>` instead?

core/backend/src/Entity.ts
```diff
@@ -7,10 +7,37 @@
...
+}
+
+/** A property that are needs to be custom handled during deserialization and serialization.
```
pmconne, line 33:
```suggestion
/** A property that needs to be custom handled during deserialization and serialization.
```

core/backend/src/Entity.ts
```diff
@@ -7,10 +7,37 @@
...
+/** Represents a row returned by an ECSql query. The row is returned as a map of property names to values.
+ * ECSqlRows are returned by [[readInstance]] and represent raw instances Entities before they are deserialized during a read.
+ * @beta */
+export interface ECSqlRow {
+  [key: string]: any
+}
+
+/** Set of properties that are used to deserialize an [[EntityProps]] from an ECSqlRow.
+ * @beta */
+export interface InstanceProps {
```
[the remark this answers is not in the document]
pmconne: It's only ever used as input to the `deserialize` function of `Entity` and its subclasses. Follow our pattern used elsewhere and call it `DeserializeEntityArgs`?

core/backend/src/Entity.ts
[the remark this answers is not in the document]
pmconne: It's passed a function argument. Why restrict callers from being able to do, e.g.,
```ts
  const props: InstanceProps = { row, iModel };
  if (whatever) {
    props.options = { element: "blah" },
  }
```
Am I overlooking some place where this type is used as an output?

core/backend/src/test/LruCache.test.ts
```diff
@@ -0,0 +1,44 @@
+import { assert, expect } from "chai";
+import { LruCache } from "../LRUCaches";
+
+describe('LruCache', () => {
```
pmconne, line 4: Please write tests for your ElementLruCache.

core/backend/src/IModelDb.ts
```diff
@@ -1702,6 +1720,9 @@ export namespace IModelDb {
    * @public
    */
   export class Models {
+    /** @internal */
+    public readonly cache = new LruCache<Id64String, ModelProps>(500);
```
pmconne, line 1724: Any evidence to suggest caching models on the backend is useful?

core/backend/src/IModelHost.ts
```diff
@@ -231,6 +235,8 @@ export class IModelHostConfiguration implements IModelHostOptions {
   public logTileSizeThreshold = IModelHostConfiguration.defaultLogTileSizeThreshold;
   /** @internal */
   public crashReportingConfig?: CrashReportingConfig;
+  /** @beta */
```
pmconne, line 238: Beta APIs need documentation.

core/backend/src/IModelHost.ts
```diff
@@ -196,6 +196,10 @@ export interface IModelHostOptions {
    */
   allowSharedChannel?: boolean;
 
+  /**
+   * Use native instance functions for element, model, and aspect CRUD operations.
```
pmconne, line 200: As an app developer, when and why should I set this to true? What will happen if I do?

core/backend/src/IModelDb.ts
```diff
@@ -1929,6 +1964,9 @@ export namespace IModelDb {
    * @public
    */
   export class Elements implements GuidMapper {
+    /** @internal */
+    public cache = new ElementLRUCache();
```
pmconne, line 1968: why `public` and mutable?

core/backend/src/IModelDb.ts
```diff
@@ -1997,13 +2024,30 @@ export namespace IModelDb {
...
+          }
+          const options = { ...props, useJsNames: true };
+          const instanceKey = this.resolveElementKey(props);
+          const rawInstance = this._iModel[_nativeDb].readInstance(instanceKey, options) as ECSqlRow;
+          const classDef = this._iModel.getJsClass<typeof Element>(rawInstance.classFullName);
+          const elementProps = classDef.deserialize({ row: rawInstance, iModel: this._iModel, options: { element: props } }) as T;
+          this.cache.set({ elProps: elementProps, loadOptions: props });
+          return elementProps;
+        }
+        return this._iModel[_nativeDb].getElement(props) as T;
```
pmconne, line 2047:
```suggestion
        if (!IModelHost.configuration?.enableThinnedNativeInstanceWorkflow) {
          return this._iModel[_nativeDb].getElement(props) as T;
        }
        
        const cachedElm = this.cache.get(props);
        if (cachedElm) {
          return cachedElm.elProps as T;
        }

        const options = { ...props, useJsNames: true };
        const instanceKey = this.resolveElementKey(props);
        const rawInstance = this._iModel[_nativeDb].readInstance(instanceKey, options) as ECSqlRow;
        const classDef = this._iModel.getJsClass<typeof Element>(rawInstance.classFullName);
        const elementProps = classDef.deserialize({ row: rawInstance, iModel: this._iModel, options: { element: props } }) as T;
        this.cache.set({ elProps: elementProps, loadOptions: props });
        return elementProps;
```

Nit but this is way easier to understand IMO.

core/backend/src/LRUCaches.ts
MichaelSwigerAtBentley: Corrected by using JSON.stringify(code) as key

core/backend/src/LRUCaches.ts
MichaelSwigerAtBentley: Swapped to use JSON.stringify(code), added a test set for ElementLRUCache that tests this case

core/backend/src/test/LruCache.test.ts
MichaelSwigerAtBentley: Added

core/backend/src/LRUCaches.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    if (el.elProps.federationGuid) {
+      this._cacheByFederationGuid.set(el.elProps.federationGuid, el.elProps.id);
+    }
+
+    this._cacheByCode.set(ElementLRUCache.makeCodeKey(el.elProps.code), el.elProps.id);
+    if (this._elementCache.size > this.capacity) {
+      const oldestKey = this._elementCache.keys().next().value as Id64String;
+      this.delete({ id: oldestKey });
```
pmconne, line 162: When purging the oldest entry you don't also purge corresponding entries from code and guid maps?

core/backend/src/IModelDb.ts
pmconne: Ok I see why various other glue code needs to access `cache`. Use a `Symbol` for its name and make it `readonly`.

full-stack-tests/core/src/frontend/hub/IModelConnection.test.ts
```diff
@@ -180,22 +180,6 @@ describe("IModelConnection (#integration)", () => {
     expect(rootTile.isLeaf).to.be.false;
   });
 
-  it("ECSQL with BLOB", async () => {
```
pmconne, line 183: why deleted?

core/backend/src/IModelDb.ts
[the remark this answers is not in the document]
MichaelSwigerAtBentley: It is public to allow for clearing of the cache from places like TxnManager. The clear could be exposed by method however

core/backend/src/LRUCaches.ts
```diff
@@ -0,0 +1,171 @@
+import { Id64String } from "@itwin/core-bentley";
```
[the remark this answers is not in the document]
MichaelSwigerAtBentley: Moved to internal

full-stack-tests/core/src/frontend/standalone/WatchForChanges.test.ts
```diff
@@ -90,13 +90,24 @@ for (const watchForChanges of [false, true]) {
...
 
     async function expectModelChanges(func: () => Promise<void>): Promise<void> {
-      const promise = new Promise<void>((resolve) => {
-        roConn.onBufferedModelChanges.addOnce((modelIds) => {
-          expect(modelIds.size).to.equal(1);
-          expect(modelIds.has(modelId)).to.be.true;
-          resolve();
-        });
-      });
+      const promise = Promise.race([
```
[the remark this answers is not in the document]
MichaelSwigerAtBentley: We had an issue where this promise never returned and this test would hang until the test timeout would be hit, causing the test suite to not finish properly. We corrected the issue and the test passes again, but I added this to ensure that the test suite would continue even if this test fails due to a similar issue in the future.

core/backend/src/LRUCaches.ts
```diff
@@ -0,0 +1,169 @@
...
+ * Cache can be searched by id, code or federationGuid.
+ * @internal */
+export class ElementLRUCache {
+  public static readonly DEFAULT_CAPACITY = 2000;
+  private _elementCache = new Map<Id64String, CachedElement>();
+  private _cacheByCode = new Map<Id64String, Id64String>();
+  private _cacheByFederationGuid = new Map<string, Id64String>();
+
+  private static makeCodeKey(code: CodeProps): string {
+    return JSON.stringify(code);
```
pmconne: Don't wanna accidentally include arbitrary other properties of the object some caller handed you nor rely on unspecified order of object keys. Maybe something more like:
```suggestion
    const keys = [code.scope, code.spec];
    if (code.value) {
      keys.push([code.value];
    }
    
    return JSON.stringify(keys);
```

core/backend/src/test/LruCache.test.ts
```diff
@@ -0,0 +1,287 @@
...
+      federationGuid: "testFedGuid5"
+    },
+  }
+
+  it('should store and retrieve a valid element', () => {
+    const cache = new ElementLRUCache(3);
+    cache.set(testElem1);
+    const retrievedElem = cache.get({ id: testElem1.elProps.id });
+    expect(retrievedElem).to.not.be.undefined;
+    expect(retrievedElem).to.deep.equal(testElem1);
```
pmconne, line 134: Why deep equal here and elsewhere? Don't you expect to get back the exact same object?

core/backend/src/LRUCaches.ts
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+    return cachedElement;
+  }
+  public set(el: CachedElement): this {
+    if (!el.elProps.id)
+      throw new Error("Element must have an id");
+
+    // do not cache this element as geom need to be rerender into geom props if any part changes
+    // TODO: find a way to handle caching geometry
```
pmconne: Is an issue filed? This is a serious limitation vs the native cache, lots of backend operations are going to want the geometry, which can be quite large, which has performance implications if we're constantly reloading it, and also cache size considerations if we end up caching too many large geometry streams.

core/backend/src/IModelDb.ts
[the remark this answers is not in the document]
pmconne: If a bunch of different places need access to it then it's fine as public, just make it inaccessible outside the package by giving it a Symbol name.

core/backend/src/IModelDb.ts
pmconne: Needs Symbol name.

core/backend/src/Entity.ts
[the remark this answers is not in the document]
MichaelSwigerAtBentley: Removed readonly from `DeserializeEntityArgs` (InstanceProps)

core/backend/src/IModelDb.ts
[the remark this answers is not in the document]
MichaelSwigerAtBentley: Swapped to Symbol name `[_cache]`

core/backend/src/IModelDb.ts
MichaelSwigerAtBentley: Swapped to Symbol name [_cache]

core/backend/src/IModelDb.ts
MichaelSwigerAtBentley: Committing, changing getModel() to reflect this as well

MichaelSwigerAtBentley: applied

core/backend/src/LRUCaches.ts
MichaelSwigerAtBentley: this.delete() removes the entry from _elementCache, _cacheByCode, and _cacheByFederationGuid

MichaelSwigerAtBentley: Ah, I see, only if its passed in. nvm

MichaelSwigerAtBentley: Corrected.

core/backend/src/test/LruCache.test.ts
MichaelSwigerAtBentley: Yes, that is correct, corrected.

pmconne (approved):

core/backend/src/IModelHost.ts
MichaelSwigerAtBentley: Added comment

core/backend/src/IModelHost.ts
```diff
@@ -197,7 +197,11 @@ export interface IModelHostOptions {
   allowSharedChannel?: boolean;
 
   /**
-   * Use native instance functions for element, model, and aspect CRUD operations.
+   * Use native instance functions for element, model, and aspect CRUD operations. While set to true, the getElement(), getModel() and getAspect() functions will use
+   * a thinned down native workflow to read the entities from the database. This workflow performs work previously done on the native side in the TS side, resulting in performance
+   * improvements. Elements will be cached in the TS side as opposed to the native side. Setting this to false will revert to the previous behavior of using the native side for all
+   * CRUD operations.
+   * Will default to `true` in 5.0.
```
pmconne, line 204: Is this PR being backported to 4.x?

pmconne: Sorry edited my original comment while you were replying. Why do we need this - as a safety valve in case we want to turn it off after problems are detected? If we do need it, consider flipping to "enable" to "disable" - people find it surprising when boolean options default to true.

core/backend/src/internal/LRUCaches.ts
```diff
@@ -0,0 +1,173 @@
+import { Id64String } from "@itwin/core-bentley";
+import { CodeProps, ElementLoadOptions, ElementLoadProps, ElementProps } from "@itwin/core-common";
+
+export class LruCache<K, V>{
```
aruniverse, line 4: does this need to be exported? we already have an LruCache in core-bentley. could that be improved or if we need this custom version do we need to expose it? maybe rename this module(file) as `ElementLRUCache`?

common/api/core-backend.api.md
```diff
@@ -669,11 +669,17 @@ export class Category extends DefinitionElement {
...
+    protected static readonly _customHandledProps: CustomHandledProperty[];
```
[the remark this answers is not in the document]
MichaelSwigerAtBentley: Sweet, Ill look into it

core/backend/src/IModelHost.ts
MichaelSwigerAtBentley: Swapped to disable and removed note about 5.0

core/backend/src/internal/LRUCaches.ts
MichaelSwigerAtBentley: Renamed file to ElementLRUCache

common/api/core-backend.api.md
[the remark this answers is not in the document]
MichaelSwigerAtBentley: The inheritdoc tag works for intilisense, and will display all parent comments as well, but does not seem to work for extract-api

core/backend/src/internal/LRUCaches.ts
pmconne: Is `LRUMap` insufficient for your purposes?

MichaelSwigerAtBentley: Was able to swap the two out, removed the LruCache we added

common/api/core-backend.api.md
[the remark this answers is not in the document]
aruniverse: gah thats annoying. will close comment

aruniverse:

aruniverse (approved):
Approving addon changes

PR #7975, opened 2025-04-16
Fix flaky Electron test on MacOS

GytisCepk:
Changes to test:
- Increased time to wait for `maximize` and `unmaximize` event to fire from 100ms to 250ms
- Changed point to which window is moved (15, 25 -> 50, 75), since MacOS did not allow to change window position to smaller than 38 in Y axis locally.


GytisCepk:

aruniverse (approved):

GytisCepk:
> Could you write in the PR description why the x and y values had to go up for the Mac pipeline runs to pass?

Updated. However, I didn't find a single failure due to this in the pipeline runs. It is possible that this is reproducible only locally.

core/electron/src/test/backend/ElectronHost.test.ts
```diff
@@ -194,7 +198,7 @@ async function testWindowSizeSettings() {
   if (isXvfbRunning)
     window.emit("maximize"); // "maximize" event is not emitted when running with xvfb (linux)
   else
-    await BeDuration.wait(100); // "maximize" event is not always emitted immediately
+    await BeDuration.wait(250); // "maximize" event is not always emitted immediately
```
ben-polinsky, line 201: Did you consider listening on the [window.on("maximize") event](https://www.electronjs.org/docs/latest/api/browser-window/#event-maximize)? Supposedly it is emitted once maximization is complete? Promisify the event - await it, and assert.

GytisCepk: There is 2 problems with awaiting for events:
- Events are thrown at different times on different OS. In my testing, I noticed that on Windows event is throw immediately after calling `window.maximize` and on MacOS it takes a bit. Not that big of a problem because we can still handle it something like this:
```ts
  const maximizePromise = new Promise((resolve) => window.once("maximize", () => resolve(undefined)));
  window.maximize();
  ...
  await maximizePromise;
```
- Bigger issue is that calling something like `window.setSize` immediately after "maximize" or "unmaximize" event is thrown, causes window to not resize (at least on Windows). For that reason, we can't replace `wait` with await for promise, only add it as addition.

ben-polinsky (approved):

PR #7977, opened 2025-04-16
Add SchemaContext to IModelConnection in core-frontend.

hl662:
- Add `SchemaContext` to `IModelConnection` in core-frontend. The schema context will add a fallback `ECSchemaRpcLocater` by default using the `IModelRpcProps` in the `IModelConnection`
- Consequently, make ecschema-metadata, ecschema-rpcinterface-common a peerDependency to core-frontend.
- Add integration tests against schemaContext for BriefcaseConnection, SnapshotConnection, BlankConnection. (BlankConnection is expected to throw an error when calling something on SchemaContext)


core/frontend/src/IModelConnection.ts
```diff
@@ -612,6 +615,20 @@ export abstract class IModelConnection extends IModel {
     }
     return ("number" === typeof this._projectCenterAltitude) ? this._projectCenterAltitude : undefined;
   }
+
+  /**
+   * Gets the context that allows accessing the metadata (ecschema-metadata package) of this iModel.
+   * The context is created lazily when this property is accessed for the first time, without any locater registered.
+   */
+  public get schemaContext(): SchemaContext {
```
aruniverse, line 623: i would make this beta or public preview

docs/changehistory/NextVersion.md
```diff
@@ -328,6 +328,8 @@ const metaData: KindOfQuantity | undefined = await imodelDb.schemaContext.getSch
 
 - [IModelConnection.fontMap]($frontend) caches potentially-stale mappings of [FontId]($common)s to font names. If you need access to font Ids on the front-end for some reason, implement an [Ipc method](../learning/IpcInterface.md) that uses [IModelDb.fonts]($backend).
 
+- Added dependency to `@itwin/ecschema-metadata` to allow exposing schema metadata to our frontend APIs. Added a `SchemaContext` property to `IModelConnection`.
```
grigasp, line 331:
```suggestion
- Added peer dependency to `@itwin/ecschema-metadata` to allow exposing schema metadata to our frontend APIs. Added a `SchemaContext` property to `IModelConnection`.
```

common/changes/@itwin/core-frontend/nam-ec-metadata-peer-dep_2025-04-15-21-07.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Make ecschema-metadata a peerDep of core-frontend. Add `SchemaContext` to `IModelConnection`",
```
grigasp, line 5:
```suggestion
      "comment": "Add `@itwin/ecschema-metadata` peer dependency. Make iModel-specific `SchemaContext` available on `IModelConnection` through the new `schemaContext` getter.",
```

core/frontend/src/IModelConnection.ts
```diff
@@ -612,6 +617,23 @@ export abstract class IModelConnection extends IModel {
     }
     return ("number" === typeof this._projectCenterAltitude) ? this._projectCenterAltitude : undefined;
   }
+
+  /**
+   * Gets the context that allows accessing the metadata (ecschema-metadata package) of this iModel.
+   * The context is created lazily when this property is accessed for the first time, with an [[ECSchemaRpcLocater]] registered.
```
grigasp, line 623: Should it also mention the necessity for backend to have `ECSchemaRpcImpl.register()` call? Or maybe that should now be done automatically by `IModelHost`?

hl662: I have limited knowledge of the backend side, I'm not sure if `IModelHost` should call `ECSchemaRpcImpl.register()` by default on startup. Personally I don't see why not, but that's why we should ask someone else :D @wgoehrig @aruniverse ?

Also, @GytisCepk would `ElectronRpcManager` always need to initialize an `ECSchemaRpcInterface` too?

aruniverse: We are moving away from core registering rpcs by default. this will be handled by the meta framework / app that we are consolidating to on web and desktop

GytisCepk: > Also, @GytisCepk would ElectronRpcManager always need to initialize an ECSchemaRpcInterface too?

Yes. However, `ElectronRpcManager` is internal, so user would need to:
- Pass `ECSchemaRpcInterface` to `ElectronApp.startup`
- Pass `ECSchemaRpcInterface` to `ElectronHost.startup`
- Call `ECSchemaRpcImpl.register` in Electron backend

Regarding web, GPB currently already registers `ECSchemaRpcImpl`, so user only needs to register `ECSchemaRpcInterface` in the frontend

hl662: Okay, I updated the doc comment for this method to provide a link to configuring RPC for client applications. I also added a snippet about configuring RPC for server-side, in cases of Electron backends, but maybe I should be clearer and point out Electron backends? Or just toss the snippet on server side stuff completely

Also added a note that BlankConnection's schemaContext will throw an error when trying to call any methods through the schemaContext.

ColinKerr (approved):

core/frontend/src/IModelConnection.ts
```diff
@@ -612,6 +617,27 @@ export abstract class IModelConnection extends IModel {
     }
     return ("number" === typeof this._projectCenterAltitude) ? this._projectCenterAltitude : undefined;
   }
+
+  /**
+   * Gets the context that allows accessing the metadata (ecschema-metadata package) of this iModel.
```
grigasp, line 622:
```suggestion
   * Gets the context that allows accessing the metadata (see `@itwin/ecschema-metadata` package) of this iModel.
```

core/frontend/src/IModelConnection.ts
```diff
@@ -612,6 +617,27 @@ export abstract class IModelConnection extends IModel {
...
+   * The context is created lazily when this property is accessed for the first time, with an `ECSchemaRpcLocater` registered.
+   * This means to correctly access schema context, client-side applications must register ECSchemaRpcInterface following instructions for [RPC configuration](https://www.itwinjs.org/learning/rpcinterface/#client-side-configuration).
+   * Server-side applications would also [configure RPC](https://www.itwinjs.org/learning/rpcinterface/#server-side-configuration) as needed.
```
grigasp, line 625:
```suggestion
   * This means to correctly access schema context, client-side applications must register `ECSchemaRpcInterface` following instructions for [RPC configuration]($docs/learning/rpcinterface/#client-side-configuration).
   * Server-side applications would also [configure RPC]($docs/learning/rpcinterface/#server-side-configuration) as needed.
```

core/frontend/src/IModelConnection.ts
```diff
@@ -612,6 +617,27 @@ export abstract class IModelConnection extends IModel {
...
+   *
+   * @note `BlankConnection` `schemaContext` will throw an error when trying to access schema metadata.
```
grigasp, line 627:
```suggestion
   * @note The getter throws an error when called on `BlankConnection`.
```

core/frontend/src/IModelConnection.ts
```diff
@@ -612,6 +617,27 @@ export abstract class IModelConnection extends IModel {
...
+   * @beta
+   */
+  public get schemaContext(): SchemaContext {
+    if (this._schemaContext === undefined)
+      {
+        const context = new SchemaContext();
+        const locater = new ECSchemaRpcLocater(this._getRpcProps());
+        context.addLocater(locater);
+        this._schemaContext = context;
+      }
```
grigasp, line 637: formatting needs fixing

core/frontend/src/IModelConnection.ts
hl662: thanks, edited

core/frontend/src/IModelConnection.ts
hl662: thanks, edited

core/frontend/src/IModelConnection.ts
hl662: The getter doesn't throw an error, my note might not have been clear enough. I edited it to mention blankConnection does return a valid schemacontext, but has an invalid locater registered and will throw errors when calling its methods

core/frontend/src/IModelConnection.ts
hl662: fixed thanks

core/frontend/package.json
```diff
@@ -89,6 +91,7 @@
     "@itwin/object-storage-core": "^2.3.0",
     "@itwin/core-i18n": "workspace:*",
     "@itwin/webgl-compatibility": "workspace:*",
+    "@itwin/ecschema-rpcinterface-common": "workspace:*",
```
aruniverse, line 94: can this really be a direct dep?

hl662: It's use case is to register a default schema rpc locater for iModelConnections. It's internal-facing to me. If there's an application that wants to consume a different interface version for locating schemas, hence having a different dep of this pkg, they could retrieve the schemaContext and register this different locater.

core/frontend/src/IModelConnection.ts
```diff
@@ -612,6 +617,26 @@ export abstract class IModelConnection extends IModel {
...
   }
+
+  /**
+   * Gets the context that allows accessing the metadata (see `@itwin/ecschema-metadata` package) of this iModel.
+   * The context is created lazily when this property is accessed for the first time, with an `ECSchemaRpcLocater` registered as a fallback locater, enabling users to register their own locater that'd take more priority.
+   * This means to correctly access schema context, client-side applications must register `ECSchemaRpcInterface` following instructions for [RPC configuration]($docs/learning/rpcinterface/#client-side-configuration).
+   * Server-side applications would also [configure RPC]($docs/learning/rpcinterface/#server-side-configuration) as needed.
+   *
+   * @note While a `BlankConnection` returns a valid `schemaContext`, it has an invalid locater registered by default, and will throw an error when trying to call it's methods.
+   * @beta
```
hl662, line 628: @aruniverse do we want to switch from @beta to @public @preview ?

aruniverse: no we can leave as is for now

core/frontend/package.json
aruniverse: doesn't this intrinsically depend on ECSchemaRpcInterface which is static? and thereforce would cause us to make this a peer?

hl662: I've switched it over to a peerDep

aruniverse (approved):

Issue #7969, opened 2025-04-14
Expose an API to identify if a given `ContextRealityModel` represents a Cesium Ion Building assets


mdastous-bentley:
When implementing a handler for `ContextRealityModels.onChanged` there is curently no API allowing the handler to determine if the modified `ContextRealityModel` objects represent a Cesium IO building assets.

We used to compare the `url' property with the string provided by `getCesiumOSMBuildingsUrl()`, but this function is internal.

markschlosseratbentley:
`getCesiumAssetUrl()` is public -- is that one good enough to accomplish your goal?  (I realize you would need to pass in the cesium ion key and the osm building asset id; not ideal).

matmarchand:
@markschlosseratbentley I would prefer that we do not hardcode that Id on our side as well.

markschlosseratbentley:
> We used to compare the url' property with the string provided by getCesiumOSMBuildingsUrl()`, but this function is internal.

@matmarchand @mdastous-bentley can you share / link me to the code that used to do the comparison?

markschlosseratbentley:
@matmarchand @mdastous-bentley To solve your issue, I originally was thinking we could add a method to `ContextRealityModel` - something like `ContextRealityModel.isCesiumOSMBuildingsAsset`. However, the `ContextRealityModel` class exists within `common` and depends on methods and structures within `frontend`.  For this reason, we cannot implement such a function there.

In [this draft PR code](https://github.com/iTwin/itwinjs-core/pull/7978/files), I solved this the only other way I could think of --- adding a helper method in `core/frontend/src/tile/map/CesiumTerrainProvider.ts`.  Basically, you would call `isCesiumOSMBuildingsRealityModel(contextRealityModel);` to figure out if your context reality model represents a Cesium OSM Buildings asset.

An alternative would be to just promote `getCesiumOSMBuildingsUrl()` in that same file from `internal` to `beta` and then the caller code you mentioned can just remain the same.

What do you think?

@aruniverse @pmconne interested in your opinion here too.

mdastous-bentley:
I think you could simply expose a constant representing the `AssetId` for cesium building, and then we call `getCesiumAssetUrl()`.

markschlosseratbentley:
> I think you could simply expose a constant representing the `AssetId` for cesium building, and then we call `getCesiumAssetUrl()`.

[...]

PR #7978, opened 2025-04-16
Add CesiumIonAssetId enum API

markschlosseratbentley:
This work aims to help resolve some of the issues brought up in https://github.com/iTwin/itwinjs-core/issues/7969

Essentially, a caller of some iTwin.js APIs would like to have access to the Cesium ion asset ID for OSM Buildings.  Previously, this asset ID was baked into iTwin.js internally.

There is precedent for other Cesium ion asset IDs to be exposed via enum, [as seen here](https://www.itwinjs.org/reference/core-common/displaystyles/cesiumterrainassetid/).  That is terrain-specific, however.

This PR creates a new enum titled `CesiumIonAssetId` which includes asset IDs for cases outside the `CesiumTerrainAssetId` umbrella, namely just Cesium OSM Buildings for now.

Internally, the `CesiumTerrainAssetId` enum has been moved into a new file where it resides alongside the new `CesiumIonAssetId` enum.


markschlosseratbentley:
> FYI core-common exposes some [known ion asset Ids](https://github.com/iTwin/itwinjs-core/blob/1fc628a6376d53c7997bc1af52f8f92c326030fe/core/common/src/TerrainSettings.ts#L17) already.

Those seem specific for "terrain", but I think we should probably do something similar in an `enum` if we go this route.

pmconne:
Frankly I don't think a special case for OSM buildings should ever have been baked into the core APIs but that ship has long sailed.

pmconne (commented):
I believe "ion" branding is supposed to be all lower-case.

markschlosseratbentley:
> I believe "ion" branding is supposed to be all lower-case.

Looks like you are correct -- changed the comments to follow that pattern.

pmconne (approved):

markschlosseratbentley:
Waiting for Image Tests.

PR #7980, opened 2025-04-16
Fix shadow map rendering bug

danielzhong-bentley:
## Bug:
One of the models on ImageTests appears to have lost its shadows. There are 2 view (same view with different settings) looks like the shadows are missing. They're not really missing, they're just so spread out that everything on the model is just a bit darker. 

Model: DelawareRIverWaterfrontRail.bim

The 2 views in question are:
Station masked
Station without mask

More Details:
[582](https://github.com/iTwin/itwin-graphics-backlog/issues/582)

## Root Cause:
When previous commit changed from using `forEachTileTreeRef()` to `getTileTreeRefs()`, the logic became based on a **lazy iterator** rather than immediate function execution.

So, using `return` inside the loop over `getTileTreeRefs()` caused the shadow map setup to exit early when encountering a non-shadow-casting tree. This skipped important tile trees and caused the scene to render too dark in some views.

## Solution
Replaced `return` with `continue` to skip over irrelevant trees while still processing the rest.


pmconne (approved):
Dumb bug, good fix. Run `rush change`.

markschlosseratbentley:
Great job, thanks @danielzhong-bentley

markschlosseratbentley (approved):

PR #7986, opened 2025-04-20
readPixels includes contour line information

pmconne:
As described in https://github.com/iTwin/platform-bentley-community/issues/103, applications want to provide tooltips and/or decorations displaying the elevations of contour lines in the view. `HitDetail` and `Viewport.readPixels` now provide the following information about a pixel that was drawn as a contour line:
- Whether it is a major or minor contour;
- To which `ContourGroup` it belongs; and
- The elevation of the contour line in world coordinates.

Note that the elevation is a multiple of the contour line interval specified by the `ContourGroup` (with floating point fuzz), **not** the Z of the pixel in world coordinates (you can get that from `HitDetail.hitPoint.z`). So every pixel in a contiguous contour line will have the same elevation, along with the same `ContourGroup` and major/minor flag. Note, however, that we can only tell you about individual pixels; if an app wants to reconstruct contour **lines** it will have to analyze and compare adjacent pixels to reconstruct them, and that can only ever be an approximation - dash patterns, occluding geometry, portions of contours being off-screen, etc can all break up otherwise-contiguous contour lines, and multiple contour lines from the same group with the same elevation can appear in a single scene depending on the topology.

To make this work, I had to add a fourth color attachment to which the fragment shader can output the encoded information. I considered simply adding a couple of new `RenderOrder` values for major and minor contours instead of adding the new color attachment, but then applications would have to reconstruct the approximate elevation and would not know which group the contour came from. It adds a little extra unnecessary computation to the surface fragment shaders in the case where contours exist but we're not reading pixels. If no contours exist, that should get optimized away; in that case we also don't bother reading back the contour frame buffer (it'll all be zeroed).

I updated `fdt tooltips` to output contour info.

#7205 introduced contour lines, but included zero unit tests to verify the display output, and none have been added since. I don't know how much effort it will take to add some, and I may decide not to invest the time. We really need to stop approving PRs that lack unit tests.

I added some unit tests (we need more) and added missing comparison routines to the core-common contour APIs.

### TODO
- [x] Run ImageTests
- [x] Make sure it works with MSAA (need different render buffers?) and ambient occlusion
- [x] Add `Pixel.Selector.Contours`?
- [x] Make sure it computes correct negative elevations
- [x] Unit tests if feasible
- [x] Re-round to nearest interval multiple on CPU (higher precision)?


pmconne:
Need to run ImageTests.

bbastings (approved):

bbastings (commented):
Approving HitDetail changes. Rest looks good to me too, but should be reviewed by display team.

ben-polinsky (approved):
Approving API changes

common/api/core-frontend.api.md
```diff
@@ -2074,6 +2075,13 @@ export enum ContextRotationId {
     View = 6
 }
 
+// @beta
```
markschlosseratbentley, line 2078: Perhaps consider a NextVersion.md entry.

core/frontend/src/internal/render/webgl/glsl/Contours.ts
```diff
@@ -87,8 +104,13 @@ const applyContours = `
   uint msk = 1u << uint(offset);
   contourAlpha *= (patterns[(lineCodeWt / 16) & 0xf] & msk) > 0u ? 1.0 : 0.0;
   contourAlpha = min(contourAlpha, 1.0);
+
```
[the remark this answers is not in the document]
pmconne: If it does then all of the pre-existing computations preceding it is impacting performance too. I'm not sure whether various WebGL implementations are clever enough to optimize out everything following the "no contours" conditional at the top of `applyContours` - that conditional isn't directly based on a uniform. If you wanted to find out, you could add a boolean `u_haveContours` uniform to test against instead and see if it affects performance.

core/frontend/src/Viewport.ts
```diff
@@ -1334,6 +1334,7 @@ export abstract class Viewport implements Disposable, TileUser {
       removals.push(settings.onAmbientOcclusionSettingsChanged.addListener(displayStyleChanged));
       removals.push(settings.onEnvironmentChanged.addListener(displayStyleChanged));
       removals.push(settings.onPlanProjectionSettingsChanged.addListener(displayStyleChanged));
+      removals.push(settings.onContoursChanged.addListener(displayStyleChanged));
```
pmconne, line 1337: @markschlosseratbentley my tests weren't working because this line was missing. After this PR completes please make sure someone adds some proper basic unit tests.

core/frontend/src/internal/render/webgl/glsl/Contours.ts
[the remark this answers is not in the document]
pmconne: Do you intend to measure it?

core/frontend/src/Viewport.ts
markschlosseratbentley: I filed an item: https://github.com/iTwin/itwin-graphics-backlog/issues/655

pmconne:
`readPixels` produces accurate contour information when MSAA and/or AO are enabled. However, enabling MSAA causes the skybox to disappear. That doesn't happen on master.

core/frontend/src/internal/render/webgl/glsl/Contours.ts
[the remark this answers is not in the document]
markschlosseratbentley: I ran a medium-sized suite of datasets using DPTA locally on my Windows machine but have some inconsistent timings. I believe several runs will need to be averaged to get good data.

@MarcNeely was considering running these on some test machines he has, but I will continue on my end trying to stabilize the numbers with some more runs as my machine is free.

pmconne:
[Adding](https://github.com/iTwin/itwinjs-core/pull/7986/commits/b19ae93d92570f3d449619c790e34fde0b233b5e) a multi-sample buffer for contours made no difference.
If I [don't clear](https://github.com/iTwin/itwinjs-core/pull/7986/commits/11535753f9349f0ae50fdc4f075f10ad84e5605e) the fourth fragment shader output, the background does not turn black.

pmconne:
> > readPixels produces accurate contour information when MSAA and/or AO are enabled. However, enabling MSAA causes the skybox to disappear. That doesn't happen on master. The sheet border also disappears when MSAA is enabled, as does the background map if it's drawn without depth (it still renders for readPixels though - you can see the tooltips on the blank background where the map should be).
> > All of these use SceneCompositor.getBackgroundFbo when drawing to the screen.
> > This behavior began with the very first commit on this branch.
> 
> [Adding](https://github.com/iTwin/itwinjs-core/pull/7986/commits/b19ae93d92570f3d449619c790e34fde0b233b5e) a multi-sample buffer for contours made no difference. If I [don't clear](https://github.com/iTwin/itwinjs-core/pull/7986/commits/11535753f9349f0ae50fdc4f075f10ad84e5605e) the fourth fragment shader output, the background does not turn black. However, if I do clear all four outputs and turn off the skybox, the background does display in the view's background color, not black.

@MarcNeely do you spot what I'm doing wrong?

core/frontend/src/internal/render/webgl/SceneCompositor.ts
```diff
@@ -320,16 +337,31 @@ class FrameBuffers implements WebGLDisposable {
...
+      assert(
+        undefined !== textures.colorMsBuff &&
+        undefined !== textures.featureIdMsBuff &&
+        undefined !== textures.featureIdMsBuffHidden &&
+        undefined !== textures.contoursMsBuff &&
+        undefined !== textures.depthAndOrderMsBuff &&
+        undefined !== textures.depthAndOrderMsBuffHidden
+      );
+      const colorAndPickMsBuffs = [textures.colorMsBuff, textures.featureIdMsBuff, textures.depthAndOrderMsBuff, textures.contoursMsBuff];
       const colorAndPickFilters = [GL.MultiSampling.Filter.Linear, GL.MultiSampling.Filter.Nearest, GL.MultiSampling.Filter.Nearest];
```
MarcNeely, line 365: Looks like colorAndPickFilters should have a 4th entry in it for the new render target (nearest filter).

pmconne: D'oh. Thank you!

core/frontend/src/internal/render/webgl/glsl/Contours.ts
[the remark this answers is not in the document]
pmconne: Thanks Marc. Mark please resolve comment if satisfied.

common/api/core-frontend.api.md
pmconne: I was about to add one but I see master is now targeting 5.1.0. This will need to be backported to 5.0.0, so any changelogs belong on that branch's 5.0.0.md, not in NextVersion.md.

pmconne:

core/frontend/src/HitDetail.ts
```diff
@@ -148,6 +148,22 @@ export interface HitPath {
   sectionDrawingAttachment?: SectionDrawingAttachmentHitInfo;
 };
 
+/** Information about a [contour line]($docs/learning/display/ContourDisplay.md) that generated a [[HitDetail]] or [[Pixel]].
+ * @see [[HitDetail.contour]]
+ * @see [[Pixel.Data.contour]]
+ * @beta
+ */
+export interface ContourHit {
+  /** The contour group that generated the contour line, as specified by [ContourDisplay.groups]]. */
```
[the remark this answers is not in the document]
pmconne: Thanks @eringram.

core/frontend/src/internal/render/webgl/glsl/Contours.ts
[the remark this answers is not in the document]
markschlosseratbentley: Yes, this is great to know. Resolving.

common/api/core-frontend.api.md
markschlosseratbentley: Sounds good.

Issue #7855, opened 2025-03-18
`@internal` `CustomAttributeSet` referenced by public / beta APIs


grigasp:
I can get to `CustomAttributeSet` through public / beta APIs (e.g. `await Property.getCustomAttributes()`), but can't use it due to it being `@internal`. Shouldn't it be public?

https://github.com/iTwin/itwinjs-core/blob/97c5147c898c4de106c3f6f3ac467a360062a5b2/core/ecschema-metadata/src/Metadata/CustomAttribute.ts#L17-L29

PR #7988, opened 2025-04-22
Fix public API exposing internal CustomAttributeSet

rushikesh-bentley:
Issue Link: https://github.com/iTwin/itwinjs-core/pull/7903
fixes: https://github.com/iTwin/itwinjs-core/issues/7855

Changes made:
- Marked `CustomAttributeSet` as `@public preview`


core/ecschema-metadata/src/Metadata/CustomAttribute.ts
```diff
@@ -14,7 +14,7 @@ export interface CustomAttribute {
   [propName: string]: any;
 }
 
-/** @internal */
+/** @public @preview */
 export interface CustomAttributeSet {
   [Symbol.iterator]: () => IterableIterator<[string, CustomAttribute]>;
   has(className: string): boolean;
```
rschili, line 20: If we make it public, it needs proper documentation. Add something, especially indicating what the string on the iterator is, and how className is formatted (casing, if it supports fullname, if it does, which separator does it use etc..)

rschili (approved):

core/ecschema-metadata/src/Metadata/CustomAttribute.ts
rushikesh-bentley: documentation added

ColinKerr (approved):

StefanApfel-Bentley (approved):

PR #7991, opened 2025-04-22
Fix imodel draping bug

danielzhong-bentley:
This PR fixes a bug that certain iModels not being draped correctly.

More detail:
https://github.com/orgs/iTwin/projects/37/views/26?sliceBy%5Bvalue%5D=danielzhong-bentley&pane=issue&itemId=107680059&issue=iTwin%7Citwin-graphics-backlog%7C660

https://github.com/iTwin/itwin-graphics-backlog/issues/659

Before:
image

After:
image


core/frontend/src/internal/tile/IModelTile.ts
```diff
@@ -108,6 +108,14 @@ export class IModelTile extends Tile {
...
 
     const sizeMultiplier = this.hasSizeMultiplier ? this.sizeMultiplier : undefined;
+
+    let ecefTransform: Transform;
+    try {
+      ecefTransform = this.tree.iModel.getEcefTransform();
+    } catch {
+      ecefTransform = Transform.createIdentity();
+    }
+
```
pmconne, line 118:
```suggestion
    const ecefTransform = this.tree.iModel.isGeoLocated ? this.tree.iModel.getEcefTransform() : Transform.createIdentity();
```
Just check the condition that ensures it won't throw, instead of catching a predictable exception.

core/frontend/src/internal/tile/RealityTileLoader.ts
```diff
@@ -99,8 +99,16 @@ export abstract class RealityTileLoader {
...
     let reader: GltfReader | ImdlReader | undefined;
 
+    let ecefTransform: Transform;
+
+    try {
+      ecefTransform = tile.tree.iModel.getEcefTransform();
+    } catch {
+      ecefTransform = Transform.createIdentity();
+    }
+
```
pmconne, line 109:
```suggestion
    const ecefTransform = tile.tree.iModel.isGeoLocated ? tile.tree.iModel.getEcefTransform() : Transform.createIdentity();
```
Just check the condition that ensures it won't throw, instead of catching a predictable exception.

pmconne:
Why would this test only be **sporadically** timing out, if you've identified the cause? Wouldn't an uncaught exception cause it to **always** time out or fail?

danielzhong-bentley:
> Why would this test only be **sporadically** timing out, if you've identified the cause? Wouldn't an uncaught exception cause it to **always** time out or fail?

I'm guessing, the test timed out sporadically because the issue happened inside async tile loading code. When `ecefLocation` existed but was invalid, calling `.getTransform()` could throw, and since the error wasn’t caught, the async function hung without resolving. That didn’t cause the test to immediately fail — it just caused `waitForAllTilesToRender()` to never complete, leading to a timeout. 

It was sporadic because the bad code path probably only triggered under certain timing conditions (e.g., cache, etc.). Switching to `isGeoLocated ? getEcefTransform() :` allow us access the transform safely, preventing the hang.

danielzhong-bentley:
> Why would this test only be **sporadically** timing out, if you've identified the cause? Wouldn't an uncaught exception cause it to **always** time out or fail?

I'm guessing, the test timed out sporadically previously because the issue happened inside **async tile loading code**. When `ecefLocation` existed but was invalid (e.g. ecefLocation is loaded at runtime, and sometimes it's not fully initialized before tiles start using it — especially in async), then calling .getTransform() returned a corrupt value. That didn’t throw immediately, but when the returned transform was later used in tile processing, it could cause failures or hangs inside async logic. Since the error wasn’t caught, the async function hung without resolving. That didn’t cause the test to immediately fail — it just caused `waitForAllTilesToRender()` to slower complete? Leading to a timeout.

So I was thinking `iModel.getEcefTransform()` is safer with throw error. I was using this function before, it had no issue at all. And I tried this function again instead of `ecefLocation.getTransform()` — I have started CI test multiple times, they all passed.

danielzhong-bentley:
> Why would this test only be **sporadically** timing out, if you've identified the cause? Wouldn't an uncaught exception cause it to **always** time out or fail?

I'm still running more CI tests right now to make sure it won't sporadically time out. I checked the CI test history, and it looks like the first failure happened during the PR that only included this line change.

I'm guessing, the test timed out sporadically previously because the issue happened inside async tile loading code. When ecefLocation existed but was invalid (e.g. ecefLocation is loaded at runtime, and sometimes it's not fully initialized before tiles start using it, especially in async), then calling .getTransform() returned a corrupt value. That didn’t throw immediately, but when the returned transform was later used in tile processing, it could cause failures or hangs inside async logic. Since the error wasn’t caught, the async function hung without resolving. That didn’t cause the test to immediately fail — it just caused waitForAllTilesToRender() to slower complete? Leading to a timeout.

So I was thinking iModel.getEcefTransform() is safer with throw error. I was using this function before, it had no issue at all. And I tried this function again instead of ecefLocation.getTransform(). Then I started CI test multiple times, they all passed.

danielzhong-bentley:
**Update:**

I ran the tests four times, one of them just failed again.

pmconne:
> May I know the earliest date you noticed these tests failing?

I don't recall - sorry.

> My direction is probably wrong

Your changes are worth keeping even if they don't solve the timeout problem.

danielzhong-bentley:
I've updated this PR to fix the draping bug and improve the code. I'll create a separate PR later to address the timeout issue.

common/api/core-frontend.api.md
```diff
@@ -4359,6 +4359,20 @@ export interface GraphicPrimitive2d {
     zDepth: number;
```
markschlosseratbentley, line 4359: Using the dataset in this item (https://github.com/iTwin/itwin-graphics-backlog/issues/659):

1) Open DTA
2) Recall saved view `no_drape_0` from _ESV.json
3) Recall saved view `drape_0` from _ESV.json

Notice you need to zoom in and out in order to make tiles / geometry show up.

3) Reload DTA and reverse the order of recalling the saved views (recall `drape_0` then recall `no_drape_0`).

Notice that only the blue support beam geometry shows up.  You need to zoom in and out in order to make the ground show up.

danielzhong-bentley: https://github.com/orgs/iTwin/projects/37/views/26?sliceBy%5Bvalue%5D=danielzhong-bentley&pane=issue&itemId=107849852&issue=iTwin%7Citwin-graphics-backlog%7C664

Let's create a new PR to address the other bugs, since this is a different issue. I believe the bug in the current PR is more critical for this feature.

markschlosseratbentley: Sounds good.  Resolving.

danielzhong-bentley:
This PR is ready for review, thanks!

FYI: @pmconne @markschlosseratbentley

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -430,7 +430,8 @@ export class IModelTileTree extends TileTree {
   public draw(args: TileDrawArgs): void {
     const tiles = this.selectTiles(args);
     this._rootTile.draw(args, tiles, this._numStaticTilesSelected);
-    this._layerHandler.collectClassifierGraphics(args, tiles);
+    if(!(args instanceof GraphicsCollectorDrawArgs))
```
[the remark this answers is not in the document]
danielzhong-bentley: Thanks for the feedback, just did it.

extensions/frontend-tiles/src/BatchedTileTree.ts
```diff
@@ -89,7 +89,8 @@ export class BatchedTileTree extends TileTree {
       tile.drawGraphics(args);
 
     args.drawGraphics();
-    this._layerHandler.collectClassifierGraphics(args, tiles);
+    if(!(args instanceof GraphicsCollectorDrawArgs))
```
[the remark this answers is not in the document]
pmconne: Don't forget to replace this one too.

pmconne (approved):

extensions/frontend-tiles/src/BatchedTileTree.ts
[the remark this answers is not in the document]
danielzhong-bentley: Fixed. Also updated the corresponding logic in RealityTileTree.

aruniverse:
does this need to be bp'd?

markschlosseratbentley:
> does this need to be bp'd?

yes to 5

aruniverse:

PR #7995, opened 2025-04-23
Don't fail test when Android emulator output includes invalid UTF-8

tcobbs-bentley:
While I don't know what causes invalid UTF-8 to sometimes show up in the emulator output, we don't really care about it, so instead of generating an error, replace the invalid output with the Unicode Replacement Character (�).


pmconne (approved):

hl662:
backport to release/5.0?

tcobbs-bentley:

PR #8001, opened 2025-04-24
Implementing new caching strategy for getProperties method

Abhijeet-Bentley:
Issue Link: https://github.com/iTwin/itwinjs-backlog/issues/1441

The caching strategy utilises a map for our _mergedPropertyCache instead of an array. This greatly improves our lookup speed.
Below are some numbers which supports this claim: 
image

- buildPropertyCache and buildPropertyCacheSync now internally build the cache and then return it, instead of updating the input parameter.
- We no longer require mergeProperties method and hence, it is removed.
- getProperty() method now also uses a cache. During the performance tests it was overserved that it can reach more than 2X speeds compared to our current implementation. But, in order to compensate for the time require to populate the map, we required at least 4 getProperty() calls.


core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -705,11 +723,11 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
     }
 
     if (!this._mergedPropertyCache) {
-      this._mergedPropertyCache = [];
-      this.buildPropertyCacheSync(this._mergedPropertyCache, undefined);
+      this._mergedPropertyCache = new Map<string, Property>();
+      this.buildPropertyCacheSync(this._mergedPropertyCache);
```
rschili, line 727: this is a similar problem that we recently had. Wait until the cache is completely generated and then assign the field. Doing it like this, you open a chance that an incomplete cache may be used

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -146,18 +146,27 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
    * @param excludeInherited If true, excludes inherited properties from the results. Defaults to false.
    */
   public async getProperty(name: string, excludeInherited: boolean = false): Promise<Property | undefined> {
+    const upperKey = name.toUpperCase();
+
+    if(this._mergedPropertyCache?.has(upperKey)) {
```
rschili, line 151: This is probably a bug if excludeInherited is set to true, because the cache will also return inherited properties

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -138,46 +138,67 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
     if (excludeInherited) {
       return undefined;
     }
 
-    return this.getInheritedProperty(name);
+    if (!this._mergedPropertyCache) {
+      this._mergedPropertyCache = await this.buildPropertyCache();
+    }
+
+    if(this._mergedPropertyCache?.has(upperKey)) {
```
rschili, line 166: could just return cache.get() here, like we do on line 152

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -612,76 +633,54 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
-
-    if (this.baseClass) {
-      const baseClass = await this.baseClass;
-      if (baseClass) {
-        ECClass.mergeProperties(result, existingValues, await baseClass.getProperties(), false);
+protected async buildPropertyCache(): Promise<Map<string, Property>> {
+  const cache = new Map<string, Property>();
+  const baseClass = await this.baseClass;
+  if (baseClass) {
+    Array.from(baseClass.getPropertiesSync()).forEach(property => {
```
rschili, line 645: Array.from creates a copy of the properties returned, I don't see a reason to do that here. You can just iterate using a for...of loop

core/ecschema-metadata/src/Metadata/EntityClass.ts
```diff
@@ -106,51 +106,69 @@ export class EntityClass extends ECClass implements HasMixins {
...
+          cache.set(property.name.toUpperCase(), property);
+      }
     }
 
-    const localProps = await this.getProperties();
-    ECClass.mergeProperties(result, existingValues, localProps, true);
-  }
+    const localProps = this.getPropertiesSync(true);
+    if (localProps) {
+      Array.from(localProps).forEach(property => {
```
rschili, line 135: same here, Array.from is not needed

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -612,76 +633,54 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
-      return;
-
-    ECClass.mergeProperties(result, existingValues, [...this._properties.values()], true);
+    if (this._properties) {
+      this._properties.forEach(property => {
+        cache.set(property.name.toUpperCase(), property);
+      });
+    }
+    return cache;
   }
```
StefanApfel-Bentley, line 684: Why do we have two methods here, doing the same thing? The only difference is that the base class is awaited in the in the async method and in the sync, we use the sync lookup method. 

I guess, we could use the sync method for both cases, or we pass the resolved baseclass into the method (but I'm not sure whether I like that approach)

core/ecschema-metadata/src/test/Metadata/Class.test.ts
```diff
@@ -31,6 +31,82 @@ describe("ECClass", () => {
       schema = new Schema(new SchemaContext(), "TestSchema", "ts", 1, 0, 0);
     });
 
+    it("measures caching performance for getProperties", async () => {
```
rschili, line 34: We should not mix performance tests into regular tests. Move this to a better place, like the tests in full-stack-tests/backend/src/perftest/CloseIModalPerformance.test.ts
I see there is a test helper in core/ecschema-metadata/src/test/TestUtils/PerfTestHelper.ts but I cannot find any callers.
We don't seem to have a common structure for performance tests, but let's at least replicate what some existing tests to, like the backend perftest, or core/geometry/src/test/misc/performance.test.ts which puts them into a separate file

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -612,76 +633,54 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
-    if (!existingValues) {
-      existingValues = new Map<string, number>();
-    }
-
+  protected buildPropertyCacheSync(): Map<string, Property> {
+    const cache = new Map<string, Property>();
     const baseClass = this.getBaseClassSync();
     if (baseClass) {
-      ECClass.mergeProperties(result, existingValues, baseClass.getPropertiesSync(), false);
+      Array.from(baseClass.getPropertiesSync()).forEach(property => {
```
christophermlawson, line 671: Based on Robert's comment above, don't forget this Array.from

StefanApfel-Bentley: getProperties/Sync returns an iterable, to call forEach, we converted it in an array, but I don't see a reason why not use a for loop on the iterable directly `for(const property of baseClass.getProperitiesSync()) {`

Abhijeet-Bentley: made necessary changes.

core/ecschema-metadata/src/Metadata/Class.ts
Abhijeet-Bentley: using a for loop now

core/ecschema-metadata/src/Metadata/Class.ts
Abhijeet-Bentley: done.

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -166,18 +175,27 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
 
-    if (excludeInherited) {
-      return undefined;
+    if (!property && !excludeInherited) {
+      property = this.getInheritedPropertySync(name);
     }
 
-    return this.getInheritedPropertySync(name);
+    if (property) {
+      this._mergedPropertyCache?.set(upperKey, property);
```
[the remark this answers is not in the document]
Abhijeet-Bentley: updated logic and tested in our test.

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -138,46 +138,65 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
     if (excludeInherited) {
       return undefined;
     }
 
-    return this.getInheritedPropertySync(name);
+    if (!this._mergedPropertyCache) {
+      this._mergedPropertyCache = this.buildPropertyCacheSync();
+    }
+
+    if(this._mergedPropertyCache?.has(upperKey)) {
```
rschili, line 195: instead of the if and .has you can just return the result from get(), it will be undefined if not found

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -138,46 +138,65 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
     }
 
-    return this.getInheritedProperty(name);
+    if (!this._mergedPropertyCache) {
+      this._mergedPropertyCache = await this.buildPropertyCache();
+    }
+
+    property = this._mergedPropertyCache.get(upperKey);
+
+    return property;
```
rschili, line 168:
```suggestion
    return this._mergedPropertyCache.get(upperKey);
```

core/ecschema-metadata/src/test/Metadata/Performance.test.ts
```diff
@@ -0,0 +1,97 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import { expect } from "chai";
+import { SchemaContext } from "../../Context";
+import { ECClass } from "../../Metadata/Class";
+import { Schema } from "../../Metadata/Schema";
+import { ECSchemaNamespaceUris } from "../../Constants";
+
+/* eslint-disable @typescript-eslint/naming-convention */
+
+describe("ECClass", () => {
```
rschili, line 13: Let's call them "Performance Tests" instead of just ECClass. Also, how long do these run? It may be fine to run them during our normal test suite if they take less than 5 seconds, anything longer and I suggest we filter them out by default and add a separate "test:performance" task to the package.json which runs only these tests

core/ecschema-metadata/src/Metadata/Class.ts
StefanApfel-Bentley: And  since line 191 checks already that the cache gets created if it doesn't exists yet, the conditional ? in line 195, should also be not needed anymore.

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -612,76 +631,54 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
-      if (existing !== undefined) {
-        if (overwriteExisting) {
-          target[existing] = property;
-        }
-      } else {
-        existingValues.set(upperCaseName, target.length);
-        target.push(property);
-      }
-    }
-  }
```
StefanApfel-Bentley, line 635: Since we removed that method, we also lost the ability to override properties. I don't know if that has been discussed before, so I'll point it out here. I' assume if an class overrides a property of it's baseclass, we want the cache to hold the actual classes property in cache, not the baseclasses. 

Also, I see some benefits, in having a method that adds the properties to the cache, as it reduces the repetitive code fragements (loop, cache.set, uppercaseify propertyname, possible checks...). But instead of mergeProperties, it could be something like addPropertiesToCache, could be a class-method to have access to the _mergePropertyCache field and just needs the property-iterable to be added.

Abhijeet-Bentley: nope, we do override properties in our buildPropertyCache methods.

core/ecschema-metadata/src/test/Metadata/Performance.test.ts
Abhijeet-Bentley: They run for less than 1200 ms.

core/ecschema-metadata/src/Metadata/Class.ts
Abhijeet-Bentley: I have committed the said changes.

core/ecschema-metadata/src/Metadata/Class.ts
Abhijeet-Bentley: Changes are committed.

rschili (approved):

core/ecschema-metadata/src/test/Metadata/Class.test.ts
```diff
@@ -31,6 +31,51 @@ describe("ECClass", () => {
...
+          },
+        },
+      };
+      schema = await Schema.fromJson(schemaJson, new SchemaContext());
+      const testClass = await schema.getItem("TestClass", EntityClass);
+      const testBase = await schema.getItem("TestBase", EntityClass);
+      assert.isDefined(testClass);
+      assert.isDefined(testBase);
+      const testClassPrimProp = await testClass!.getProperty("PrimProp", true);
+      const testBasePrimProp = await testBase!.getProperty("PrimProp", true);
```
StefanApfel-Bentley, line 72: I don't know if that test setup is any helpful for us:
Since both classes have a property with that name, it would successfully look it up from the class __property_ field and the cache wouldn't even be build. Also if we exclude the inherited properties (2nd argument), the property cache wouldn't even be built either as it would return undefined if the property can't be looked up in the first check.

I guess what we have to do is to add another class which doesn't have own properties, this derives from the _TestClass_, so if we lookup for our _PrimProp_ (including inherited properties), it's forced to build a property cache and we can see whether all works as designed.

StefanApfel-Bentley: Since we have _sinon_ available in this package, I'd also propose to add a spy on the _buildPropertyCache_ method, to ensure it has been called.

StefanApfel-Bentley (approved):

Issue #7426, opened 2024-11-25
Graphics created with instancing have a null range.


matmarchand:
**Describe the bug**
Some type of graphic will not produce a valid range when using instancing. I tested PointString which was fine but not Solid(Sphere) and Gltf.

**To Reproduce**

1) Checkout branch : https://github.com/iTwin/itwinjs-core/tree/MathieuM/GraphicInstancesRange
2) Go to \itwinjs-core\core\frontend
3) npm run test RenderInstances
4) Test will fail

**Expected behavior**
I expect graphics to have a non null range so I can use it in a TileTree as the root tile.

danieliborra:
With iTwinExperience moving into Cesium, is this still relevant?

matmarchand:
I think yes, we will need to support both viewer for a while. For someone knowledgeable of that code, I would think that this is not a major fix?

danieliborra:
> I think yes, we will need to support both viewer for a while.

Ok

andremig-bentley:
Associated PR for this bug is now in the approval process -> https://github.com/iTwin/itwinjs-core/pull/8003

andremig-bentley:
PR has been completed and merged in. I'm moving this issue to "Done", but I'll keep it open for now in case any further discussion needs to happen.

markschlosseratbentley:
Reopening but keeping in Done column so we can track it until release.

PR #8003, opened 2025-04-24
Create Render Instance Graphics with Non-Null Range

andremig-bentley:
This PR is to address this issue: https://github.com/iTwin/itwinjs-core/issues/7426
 
Currently there is a bug where graphics created with instancing are created with a null range, not allowing them to be used in a TileTree as the root tile. This is due to a bug in `MeshGraphic.UnionRange()` where the null range being passed into the function was being extended to itself, resulting in a subsequent null range. This PR fixes the bug by extending the passed in range to `MeshGraphic._instances.range`. A check was added to a test in `RenderInstances.test.ts` which checks that a newly created graphic will not be created with a null range.

### DIFF
--- a/core/frontend/src/internal/render/webgl/Mesh.ts
+++ b/core/frontend/src/internal/render/webgl/Mesh.ts
@@ -154,9 +154,9 @@ export class MeshGraphic extends Graphic {
   }
 
   public override unionRange(range: Range3d) {
-    // if (this._instances)
-    //   range.extendRange(range);
-    // else
+    if (this._instances)
+      range.extendRange(this._instances.range);
+    else
       range.extendRange(this._meshRange);
   }
 
--- a/core/frontend/src/internal/render/webgl/Mesh.ts
+++ b/core/frontend/src/internal/render/webgl/Mesh.ts
@@ -154,9 +154,9 @@ export class MeshGraphic extends Graphic {
   }
 
   public override unionRange(range: Range3d) {
-    if (this._instances)
-      range.extendRange(range);
-    else
+    // if (this._instances)
+    //   range.extendRange(range);
+    // else
       range.extendRange(this._meshRange);
   }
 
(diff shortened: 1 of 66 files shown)


core/frontend/src/internal/render/webgl/Mesh.ts
```diff
@@ -154,9 +154,9 @@ export class MeshGraphic extends Graphic {
   }
 
   public override unionRange(range: Range3d) {
-    if (this._instances)
-      range.extendRange(range);
-    else
+    // if (this._instances)
```
andremig-bentley, line 157: Following the creation of these graphics, if we attempt to get the graphic's range using `UnionRange()` (like what's being done in the test below), we see a string of `UnionRange()` calls which end with this one in `Mesh.ts`, here. In this call, we see that if the `MeshGraphic` has `_instances`, then the range being passed down will be extended to encompass itself, effectively doing nothing with that range. This seems to be causing the issue, as the range being passed down here is an empty newly created range in all examples I have seen of it being used with instancing. 
 
With this check in place, if I'm understanding it right, then calling `UnionRange()` like this on instanced graphics will always return a null range. Removing the check and always extending the range to encompass `this._meshRange` fixes the problem and creates graphics with a non null range.
 
My question (and why this PR is still in draft) is if anyone knows the point of this check? I have been trying to dig in to find why we would not want to extend this range specifically for instances, but I have not yet been able to find an answer. If there is a good reason for the check, then obviously simply removing it will not work, but otherwise, a simple solution might be to just remove or somehow change this check.

pmconne (approved):

markschlosseratbentley (approved):

core/frontend/src/test/render/RenderInstances.test.ts
```diff
@@ -314,6 +310,16 @@ describe("RenderInstances", () => {
     expect(instances).toBeDefined();
 
     let graphic = IModelApp.renderSystem.createGraphicFromTemplate({ template, instances });
+
+    const range = new Range3d();
+    graphic.unionRange(range);
+    expect(range.isNull).toBe(false);
+
+    //Using isAlmostEqual due to floating point errors
+    expect(range.low.isAlmostEqual(Point3d.create(-25, -25, 0)));
```
pmconne, line 318: Does this do the same thing as `expect(range.low.isAlmostEqual(Point3d.create(-25, -25, 0))).to.be.true;`?

andremig-bentley: Nope I don't think it does, that was a mistake. Fixed in latest commit

PR #8006, opened 2025-04-24
Remove Function to fix Content Security Policy

### LINKED ISSUE #7990: LoadExtensionScript `Function` string eval is blocked by our Content Security Policy
**Describe the bug**

Not really a bug, but [the `Function ("..")` string evaluation](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/extension/providers/ExtensionLoadScript.ts#L22) that [was added to avoid webpack warnings](https://github.com/iTwin/itwinjs-core/pull/3616) is in conflict with our Content Security Policy, namely `script-src` blocks `unsafe-eval` calls.

I suggest reverting that change, and just accepting the webpack warning `Critical dependency: the request of a dependency is an expression`. The warning only happens when using cjs.

**Additional context**

`EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src"`

pankhur94:
fixes #7990

### DIFF
--- a/core/frontend/src/extension/providers/ExtensionLoadScript.ts
+++ b/core/frontend/src/extension/providers/ExtensionLoadScript.ts
@@ -14,12 +14,7 @@
  * @internal
  */
 export async function loadScript(jsUrl: string): Promise<any> {
-  // const module = await import(/* webpackIgnore: true */jsUrl);
-  // Webpack gives a warning:
-  // "Critical dependency: the request of a dependency is an expression"
-  // Because tsc transpiles "await import" to "require" (when compiled to is CommonJS).
-  // So use FunctionConstructor to avoid tsc.
-  const module = await Function("x","return import(x)")(jsUrl);
+  const module = await import(/* @vite-ignore */ jsUrl);
   return execute(module);
 }
 


core/frontend/src/extension/providers/ExtensionLoadScript.ts
```diff
@@ -14,12 +14,7 @@
  * @internal
  */
 export async function loadScript(jsUrl: string): Promise<any> {
-  // const module = await import(/* webpackIgnore: true */jsUrl);
-  // Webpack gives a warning:
-  // "Critical dependency: the request of a dependency is an expression"
-  // Because tsc transpiles "await import" to "require" (when compiled to is CommonJS).
-  // So use FunctionConstructor to avoid tsc.
-  const module = await Function("x","return import(x)")(jsUrl);
+  const module = await import(/* @vite-ignore */ jsUrl);
```
aruniverse, line 17: Does the vite ignore magic comment actually do anything?

aruniverse (commented):
@johnnyd710 can you verify this works for you via pnpm patch?

core/frontend/src/extension/providers/ExtensionLoadScript.ts
aruniverse: What is said warning?

pankhur94: image

johnnyd710:
> @johnnyd710 can you verify this works for you via pnpm patch?

Works! I can see my extension loaded

aruniverse (approved):

aruniverse:

Issue #7972, opened 2025-04-15
Planar clip mask integration tests sporadically time out


pmconne:
From core-full-stack-tests.
```
  1) Planar clip mask (#integration)
       is masked by specific model:
     Error: Timeout of 240000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  

  2) Planar clip mask (#integration)
       is masked by DesignModel priority:
     Error: Timeout of 240000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

240 seconds seems like more than enough time to do whatever they're doing.

danielzhong-bentley:
**Update:**

This test has run successfully multiple times locally. I’ll try again more and keep an eye on it to see if it fails again.

<img width="557" alt="Image" src="https://github.com/user-attachments/assets/b9ec0a01-dfa9-422b-8e2d-eabdd8242dea" />

pmconne:
Try doing what the CI scripts do (they probably run `rush cover` - your output suggests you only ran these 7 tests?). Look for something in the tests that might conflict with another test (e.g., using the same .bim file or other file path). Next time they fail in CI, inspect the raw log for clues.

danielzhong-bentley:
I suspect this PR: [#7891](https://github.com/iTwin/itwinjs-core/pull/7891) may be the reason the test is consistently failing.

I'll follow up with a fix in a separate PR shortly.

danielzhong-bentley:
**Update:**

1. The CI tests are run using `npm run -s test:integration:chrome && npm run -s test:integration:electron`, which is equivalent to running `pnpm test:integration` locally. However, local machines are significantly faster than the CI virtual machines.

2. The tests run in two environments: Chrome and Electron. The consistent failures are occurring in the Electron tests. Electron generally runs slower than Chrome, but both are subject to the same timeout limit of 240,000ms.

3. After conducting several test runs, I noticed that the draping feature has increased execution time. Previously, the average execution time for the Electron planar projection test on the virtual machine was around 122,907ms. Post draping feature, the average has risen to 226,250ms, which is very close to the timeout limit of 240,000ms.

Before Draping:
 
![Image](https://github.com/user-attachments/assets/846303b4-09ca-411e-a966-8784e77c58da)

![Image](https://github.com/user-attachments/assets/cf16c47d-6930-40d9-8d3f-1b678602eb8f)

After:

[...]

PR #8010, opened 2025-04-25
Change Planar Clip Mask test time out

danielzhong-bentley:
This PR doubles the planar clip mask test timeout from the default 240,000 ms to 480,000 ms to reolve this issue: https://github.com/iTwin/itwinjs-core/issues/7972

### DIFF
--- a/full-stack-tests/core/src/frontend/map/PlanarClipMask.test.ts
+++ b/full-stack-tests/core/src/frontend/map/PlanarClipMask.test.ts
@@ -16,6 +16,10 @@ import { TestSnapshotConnection } from "../TestSnapshotConnection";
 describe("Planar clip mask (#integration)", () => {
   let imodel: IModelConnection;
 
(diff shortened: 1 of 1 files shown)


danielzhong-bentley:
I ran 5 CI tests, and the timeout for the planar clip mask test was successfully doubled.
The test duration varied depending on the virtual machine's load, ranging between (**190,000 ms - 368,927 ms**).

pmconne (approved):

aruniverse:
Is doubling the timeout the right answer here? Are these tests supposed to take a long time?

danielzhong-bentley:
> Is doubling the timeout the right answer here? Are these tests supposed to take a long time?

Yes, currently the draping feature causes the planar classifier to take longer during testing, as they share a similar code path. Doubling the test time is expected under the current conditions. We will work on optimizing it soon, if possible..

PR #8019, opened 2025-04-29
Adjust GoogleMaps 2D tiles design to support Curated content service

mdastous-bentley:


mdastous-bentley:
@aruniverse

test-apps/display-test-app/src/frontend/CuratedGoogleMapsSessionClient.ts
```diff
@@ -0,0 +1,87 @@
+/*---------------------------------------------------------------------------------------------
```
mdastous-bentley, line 1: @aruniverse Obviously this file should not committed but that's shows what the curated content version will look like.

markschlosseratbentley:
Probably a good idea to run Image Tests on this.

mdastous-bentley:
> Probably a good idea to run Image Tests on this.

job started, waiting for result

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
```diff
@@ -144,18 +144,52 @@ export class MapLayerFormatRegistry {
...
+  /**
+  * Returns the active session client for a given format.
+   *  @beta */
   public getAccessClient(formatId: string): MapLayerAccessClient | undefined {
     if (formatId.length === 0)
       return undefined;
 
-    return this._formats.get(formatId)?.accessClient;
+    const accessClient = this._formats.get(formatId)?.client;
+    if (accessClient && typeof((accessClient as any).getAccessToken) === "function") {
```
ben-polinsky, line 161: When would accessClient not be a function? Your type is `getAccessToken(params: MapLayerAccessTokenParams): Promise<MapLayerAccessToken | undefined>;
`

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
```diff
@@ -144,18 +144,52 @@ export class MapLayerFormatRegistry {
...
+  /**
+  * Returns the active session client for a given format.
+  * @beta
+  */
+  public getSessionClient(formatId: string): MapLayerSessionClient | undefined {
+    if (formatId.length === 0)
+      return undefined;
+
+    const client = this._formats.get(formatId)?.client;
+    if (client && typeof((client as any).getSessionManager) === "function") {
```
ben-polinsky, line 176: Same as previous comment, when will `getSessionManager` not be a function?

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
mdastous-bentley: `MapLayerFormatEntry.client`'s type if `MapLayerAccessClient|MapLayerSessionClient`.  Those 2 sub-types are interfaces, and we need to determine which of the types is being active it.  To achieve it, I check if a mandatory function is available.

So to answer your first question,  `MapLayerFormatEntry.client.getAccessToken` will either be `function` OR `undefined`

ben-polinsky: Can't you check if it's undefined, then?

core/frontend/src/tile/map/MapLayerAuthentication.ts
```diff
@@ -35,11 +35,26 @@ export interface MapLayerAccessTokenParams {
...
   invalidateToken?(token: MapLayerAccessToken): boolean;
 
   onOAuthProcessEnd?: BeEvent<Listener>;
 }
 
+/**
+ *  Represents a map layer session manager.
+ * @beta
+ **/
+export interface MapLayerSessionManager {
```
pmconne, line 50: Documentation comment merely repeats the interface name. What is this interface for and what does its single property `type` mean?

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
mdastous-bentley: I could, but why not be more restrictive and also check its a function ?

core/frontend/src/tile/map/MapLayerAuthentication.ts
```diff
@@ -35,11 +35,26 @@ export interface MapLayerAccessTokenParams {
...
+  type: string;
+}
+
+/**
+ *  Represents a client handling map layer sessions.
```
pmconne, line 55: Can you elaborate?

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
ben-polinsky: Because you know it's a function (or undefined), right? Aren't we the ones setting this? We can trust our types within our own code boundary.

core/frontend/src/tile/map/MapLayerAuthentication.ts
mdastous-bentley: done

mdastous-bentley:
> > Probably a good idea to run Image Tests on this.
> 
> job started, waiting for result

`[Main_2025-05-02_08_07_MichelD_googleCuratedContent]` completed successfully

core/frontend/src/tile/map/MapLayerFormatRegistry.ts
mdastous-bentley: Alright, changed it to check `undefined` instead.

core/frontend/src/tile/map/MapLayerAuthentication.ts
mdastous-bentley: This type needs to be sub-classed by our consumers, is `type` sufficient ?

markschlosseratbentley:
> > Probably a good idea to run Image Tests on this.
> 
> job started, waiting for result

Image Tests results look fine to me.

core/frontend/src/tile/map/MapLayerAuthentication.ts
pmconne: I don't know because I have no idea what `type` is supposed to mean (it's undocumented). `type` is not even documented. What's it for, what are valid values, why isn't it read-only? Does ever may layer format's session manager actually require a `type` field? Is `MapLayerSessionManager` just some opaque container for name-value pairs specific to the private implementation of a format? If there are answers to all these question please ensure the documentation makes them obvious.

mdastous-bentley: @pmconne 
I need an abstraction that will let "external" implementations pass a  Google maps session manager (or any other session manager) down to the provider.

If you have better design suggestions I'm all ears.

pmconne: I don't understand why any of this session manager stuff needs a generic API in core-frontend, vs just letting GoogleMapsImageryProvider (or any other provider) handle it themselves. core-frontend doesn't call `getSessionManager` itself. No one at all calls `setSessionClent`. Only `GoogleMapsImageryProvider.getSessionManager` calls `getSessionClient`, and does a bunch of custom logic with it. No one looks at the `type` field. Why can't the provider just store the client itself?

mdastous-bentley: @pmconne I completely removed dependencies to `MapLayerFormatRegistry`.  Please have another look.

pmconne (approved):

aruniverse (approved):

aruniverse:

Issue #5017, opened 2023-02-02
Flaky test - Handle leak detected


grigasp:
#### Build Info

- Repo: `itwinjs-core`
- Build Name: iTwin.js
- OS: Windows x64, Linux x64
- Test Suite Name: `core-i18n`, `core-markup`, `map-layers-formats`, `linear-referencing-backend`, `analytical-backend`

#### Error Detail

The specific cause is unclear - all tests succeed and error is thrown after the whole test suite is run.

```
==[ FAILURE: 2 operations ]====================================================

--[ FAILURE: @itwin/core-i18n ]-----------------------------[ 28.73 seconds ]--

WARNING: Tests attempted to load missing asset: "/locales/en-US/NotExist.json"
WARNING: Tests attempted to load missing asset: "/locales/en/NotExist.json"
WARNING: Tests attempted to load missing asset: "/locales/en-US/test1.json"
WARNING: Tests attempted to load missing asset: "/locales/en/test1.json"
WARNING: Tests attempted to load missing asset: "/locales/en-US/test2.json"
WARNING: Tests attempted to load missing asset: "/locales/en/test2.json"
WARNING: Tests attempted to load missing asset: "/locales/en-US/initial1.json"
WARNING: Tests attempted to load missing asset: "/locales/en/initial1.json"
WARNING: Tests attempted to load missing asset: "/locales/en-US/initial2.json"
WARNING: Tests attempted to load missing asset: "/locales/en/initial2.json"
##[error]Handle leak detected. Node was still running 5 seconds after tests completed.
Try running with the DEBUG_LEAKS env var set to see open handles.

Operations failed.

--[ FAILURE: @itwin/core-markup ]---------------------------[ 58.14 seconds ]--

##[error]Handle leak detected. Node was still running 5 seconds after tests completed.
Try running with the DEBUG_LEAKS env var set to see open handles.
```

#### Links to failing builds

-  **Jan 31, Windows:** https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=1931113&view=logs&j=9f3a3068-7880-5994-433d-b92c1f9d379b&t=c78dc741-fe12-5d6a-c099-5134f70fdb0c

[...]

PR #8020, opened 2025-04-29
[certa] Fix handle leak detection

pankhur94:
- explicitly closing browser even in case of failed tests.
- don't install handle leak detection timeout until after tests have fully completed
- attempted updating playwright version to 1.50.0 but it caused i18n test to fail in this [run](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3886244&view=logs&j=751a5ea6-e147-5a0b-b7c5-e28ecfcb00e2&t=742556c9-a44d-5542-8a03-cf20a9c5ac3e&l=12362) so reverted back to 1.47.1


tools/build/src/mocha-reporter/index.ts
```diff
@@ -69,30 +69,21 @@ class BentleyMochaReporter extends Spec {
...
-        logBuildError("There were 0 passing tests.  That doesn't seem right."
-          + "\nIf there are really no passing tests and no failures, then what was even the point?"
-          + "\nIt seems likely that tests were skipped by it.only, it.skip, or grep filters, so I'm going to fail now.");
-        failBuild();
-      }
-    }
 
     // Detect hangs caused by tests that leave timers/other handles open - not possible in electron frontends.
     if (!("electron" in process.versions)) {
+      process.on("chrome-test-runner-done", () => {
```
wgoehrig, line 75: This will only work for tests run in certa - we still need to handle tests that just run in node...

tools/build/src/mocha-reporter/index.ts
```diff
@@ -69,30 +69,21 @@ class BentleyMochaReporter extends Spec {
...
+        console.warn("WARNING: chrome-test-runner-done event received");
```
wgoehrig, line 76: Why is this a warning?  Won't this just add noise to our test output?

tcobbs-bentley: That is a mistake (committing debug code). Will fix.

tcobbs-bentley: Fixed.

tools/certa/src/runners/chrome/ChromeTestRunner.ts
```diff
@@ -65,6 +65,7 @@ export class ChromeTestRunner {
       writeCoverageData(coverage);
 
     process.exitCode = failures;
+    (process as any).emit("chrome-test-runner-done");
```
wgoehrig, line 68: Why does this work but starting a timer when writing the test epilogue doesn't?

I mean, I know doing the handle leak detection in our reporter is pretty hacky in the first place, but I'm confused why we'd get to here 10s of seconds after the tests finished.  Is chrome really taking that long to shut down?  And if so, how do we know that we're not ultimately causing that with a handle leak?

tcobbs-bentley: I have no idea why the timer didn't work from epilog, but I got consistent failure when running on macOS. (As an aside, the timer had recently been bumped to 30 seconds, and I dropped it back to 10 after moving the timeout.) My understanding is that multiple modules are being tested concurrently during `rush cover`, and that seems to contribute, as I don't think I ever saw the failure when just testing one module. But I still have no idea why it takes so long.

tools/build/src/mocha-reporter/index.ts
tcobbs-bentley: I updated the code to run the leak test (with 30 second timeout) from `epilogue` when not running in Electron and not running in Certa Chrome.

tools/certa/src/runners/chrome/ChromeTestRunner.ts
wgoehrig: OK, I still think this warrants more investigation, but I understand this will help us unblock builds.

Let's merge this PR but keep #5017 open until we understand the root cause here..

wgoehrig (approved):

Issue #8029, opened 2025-05-01
Can't delete a category


FelixGirard:
**Describe the bug**
When deleting a category, I get an error: 

![Image](https://github.com/user-attachments/assets/965dea08-9b89-4684-a373-518df5239bd9)

**To Reproduce**
Try to delete a category element.

**Expected behavior**
I would expect the category to be deleted, if there are elements associated with it I would like to see this in the message. I tested without associated elements though and still get the error.

**Desktop (please complete the applicable information):**

- OS: Windows
- Browser Electron
- Version
- iTwin.js Version 4.11

aruniverse:
Thanks for reporting Felix!

We need some triage to figure out the actual problem and even if it's WAD (like sub categories need to be deleted first) our error messages need to be improved

khanaffan:
Deleting category is not allowed unless as part of purging operation.

FelixGirard:
In our use case we are letting user create Category with the Rank set to User.
I think it make sense to allow deleting a user category if there are no elements assigned to it.

I don't know what a purge is

khanaffan:
In simple words definition elements can only be deleted if it's not referenced by anything

You should call [deleteDefinitionElements](https://www.itwinjs.org/reference/core-backend/imodels/imodeldb/imodeldb.elements/deletedefinitionelements/)

FelixGirard:
That worked! Thank you! 

I guess it just wasn't clear by the error, but at least it's working now.

aruniverse:
> I guess it just wasn't clear by the error, but at least it's working now.

We should provide a clearer error message. Reactivating issue to handle that

khanaffan:
Ah I thought it was a small fix so I fixed it.

FelixGirard:
Seems like this is solved. Closing.

PR #8030, opened 2025-05-01
Improve error message when deleting element that is a definition element

khanaffan:
closes https://github.com/iTwin/itwinjs-core/issues/8029

This pull request improves the error message when attempting to delete an element that is a definition element. Instead of a generic error, a specific error message is thrown indicating that definition elements cannot be deleted directly and that the `deleteDefinitionElements()` method should be used instead. This provides clearer guidance to developers and helps prevent accidental deletions of important elements.

### DIFF
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -2205,8 +2205,14 @@ export namespace IModelDb {
     public deleteElement(ids: Id64Arg): void {
       const iModel = this._iModel;
       Id64.toIdSet(ids).forEach((id) => {
+        this[_cache].delete({ id });
+        const key = iModel[_nativeDb].resolveInstanceKey({ partialKey: { id, baseClassName: "BisCore:Element" } });
+        const isDefinitionElement = iModel[_nativeDb].isSubClassOf(key.classFullName, "BisCore:DefinitionElement");
+        if (isDefinitionElement) {
+          throw new IModelError(IModelStatus.DeletionProhibited, "DefinitionElements cannot be deleted directly. Use deleteDefinitionElements() instead.");
+        }
+        
         try {
-          this[_cache].delete({ id });
           iModel[_nativeDb].deleteElement(id);
         } catch (err: any) {
           err.message = `Error deleting element [${err.message}], id: ${id}`;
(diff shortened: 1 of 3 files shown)


pmconne (approved):

core/backend/src/IModelDb.ts
```diff
@@ -2205,8 +2205,14 @@ export namespace IModelDb {
     public deleteElement(ids: Id64Arg): void {
       const iModel = this._iModel;
       Id64.toIdSet(ids).forEach((id) => {
+        this[_cache].delete({ id });
+        const key = iModel[_nativeDb].resolveInstanceKey({ partialKey: { id, baseClassName: "BisCore:Element" } });
```
MichaelSwigerAtBentley, line 2209: Do we know the overhead of calling into native for the instance key here? I suppose this extra overhead is offset by the fail-early behavior.

core/backend/src/IModelDb.ts
```diff
@@ -2205,8 +2205,14 @@ export namespace IModelDb {
...
+        const isDefinitionElement = iModel[_nativeDb].isSubClassOf(key.classFullName, "BisCore:DefinitionElement");
+        if (isDefinitionElement) {
+          throw new IModelError(IModelStatus.DeletionProhibited, "DefinitionElements cannot be deleted directly. Use deleteDefinitionElements() instead.");
```
MichaelSwigerAtBentley, line 2212: Maybe we should mention why it cant be deleted directly, IE:

"DefinitionElements cannot be deleted directly if referenced by another entity. Use deleteDefinitionElements() instead."

khanaffan: It cannot be deleted via deleteElement() independent of if it's referenced or not. Finding references is complex task and done by deleteDefinitionElements()

core/backend/src/IModelDb.ts
khanaffan: Delete is more expensive than querying. From CRUD performance test we know you can read around 30K whole element/sec but you can delete like 800/sec. In case of just doing `SELECT ECClassId FROM Bis.Element WHERE ECINstanceId=?` is like the query SQLite is optimized around as it stores table as btree. And btree is indexed on ROWID which is ECInstanceId in this case.

kabentley: I don't know why you would do this here. The error message should be improved in native code.

kabentley: That would be first choice. Alternatively, this test could be moved to the `catch` clause so it doesn't slow down normal deletes.

PR #8033, opened 2025-05-01
Reduce Element and Model Cache Size

MichaelSwigerAtBentley:
Reduces the element cache to 50 elements and the model cache to 10 models.

### DIFF
--- a/common/changes/@itwin/core-backend/mike-reduce-cache-size_2025-05-01-19-28.json
+++ b/common/changes/@itwin/core-backend/mike-reduce-cache-size_2025-05-01-19-28.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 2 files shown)


aruniverse:
but why?

MichaelSwigerAtBentley:
> but why?

@khanaffan Asked that we reduce it. I assume it is to save on memory

khanaffan:
> > but why?
> 
> 
> 
> @khanaffan Asked that we reduce it. I assume it is to save on memory

Correct. Cache needs to be effective and not too large. We can come back to this if it is a problem but starting with large number ensure more memory will be used for not much gain as typically app work with small number of elements and a single model. Other application like transformation only glance over million of elements without going back to same element again.

aruniverse (approved):

PR #8035, opened 2025-05-02
Improve error message when deleting element that is a definition element

khanaffan:
closes https://github.com/iTwin/itwinjs-core/issues/8029

This pull request enhances the error message displayed when trying to delete a definition element. Instead of a generic error, it now provides a specific message indicating that definition elements cannot be deleted directly. Developers are advised to use the `deleteDefinitionElements()` method instead. This change offers clearer guidance and helps prevent accidental deletions of crucial elements.

### DIFF
--- a/common/changes/@itwin/core-backend/affank-fix-errmsg-delete-def-el_2025-05-02-16-29.json
+++ b/common/changes/@itwin/core-backend/affank-fix-errmsg-delete-def-el_2025-05-02-16-29.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Add test for improved error message for when deleting definition element",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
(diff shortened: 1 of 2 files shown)


common/changes/@itwin/core-backend/affank-fix-errmsg-delete-def-el_2025-05-02-16-29.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Add test for improved error message for when deleting definition element",
```
aruniverse:
```suggestion
      "comment": "",
```

aruniverse (approved):

kabentley:
This change is fine and useful. 

However, in the context of deleting Categories, have you considered what should happen to the CodeIndex? Category names must be reserved before you can use them. If you delete a Category, should the name be "unreserved" from the CodeIndex? Probably. It makes me think there should be a `deleteCategory` method somewhere.

ColinKerr (approved):

PR #8037, opened 2025-05-02
Optionally return world CRS

ben-polinsky:
imodel-native: https://github.com/iTwin/imodel-native/pull/1103

There are a number of critical "world" CRS ("EPSG:4326" for example) which are not available from [getAvailableCoordinateReferenceSystems](https://github.com/iTwin/itwinjs-core/blob/06313ff2dc8671959008c377d92437e3bbe2f57a/core/backend/src/GeographicCRSServices.ts#L48). This is because [GeoServicesInterop::GetListOfGCS](https://github.com/iTwin/imodel-native/blob/644642cf7a86d9926c7c64aee51f4491045cca97/iModelJsNodeAddon/GeoServicesInterop.cpp#L75) filters out CRS without extents or those with extents spanning the globe.

This PR exposes an optional includeWorld argument to circumvent the above check.

### DIFF
--- a/core/backend/src/GeographicCRSServices.ts
+++ b/core/backend/src/GeographicCRSServices.ts
@@ -38,6 +38,9 @@ export interface GetAvailableCoordinateReferenceSystemsArgs {
    * Maximum longitude and latitude correspond to extent.high.x and extent.high.y, respectively.
    */
   extent?: Range2dProps;
+  /** If true, returns additional CRS with extents spanning the entire Earth's surface.
+   * @default false
+   */
   includeWorld?: boolean
 }
 
--- a/core/backend/src/GeographicCRSServices.ts
+++ b/core/backend/src/GeographicCRSServices.ts
@@ -38,6 +38,7 @@ export interface GetAvailableCoordinateReferenceSystemsArgs {
    * Maximum longitude and latitude correspond to extent.high.x and extent.high.y, respectively.
    */
   extent?: Range2dProps;
+  includeWorld?: boolean
 }
 
 /** Get a list of Geographic Coordinate Reference Systems.
@@ -46,5 +47,5 @@ export interface GetAvailableCoordinateReferenceSystemsArgs {
  * @beta
  */
 export async function getAvailableCoordinateReferenceSystems(args: GetAvailableCoordinateReferenceSystemsArgs): Promise<AvailableCoordinateReferenceSystemProps[]> {
-  return IModelNative.platform.GeoServices.getListOfCRS(args.extent);
+  return IModelNative.platform.GeoServices.getListOfCRS(args.extent, args.includeWorld);
 }
--- a/core/backend/src/GeographicCRSServices.ts
+++ b/core/backend/src/GeographicCRSServices.ts
@@ -9,7 +9,7 @@
 import { Range2dProps } from "@itwin/core-geometry";
 import { IModelNative } from "./internal/NativePlatform";
 
-/** Describes a geographic coordinate reference system produced by [[getAvailableCoordinateReferenceSystems]].
+/** Describes a coordinate reference system produced by [[getAvailableCoordinateReferenceSystems]].
  * @beta
  */
 export interface AvailableCoordinateReferenceSystemProps {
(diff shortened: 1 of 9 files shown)


bbastings (approved):

core/backend/src/GeographicCRSServices.ts
```diff
@@ -38,6 +38,10 @@ export interface GetAvailableCoordinateReferenceSystemsArgs {
    * Maximum longitude and latitude correspond to extent.high.x and extent.high.y, respectively.
    */
   extent?: Range2dProps;
+  /** If true, returns additional CRS with extents spanning the entire Earth's surface.
```
tcobbs-bentley, line 41: I don't like using "CRS" in a way where it is supposed to be plural. The `extent` comment above uses "CRS's", which I also don't particularly like, but I feel that it is better than just "CRS". (Grammatically, I don't think "apostrophe s" is ever allowed to indicate plural, but with acronyms, it might be accepted.) Whatever is used should be consistent. I feel that the only real options are "CRS's" and "Coordinate Reference Systems" in both locations.

core/backend/src/test/misc/GeoServices.test.ts
```diff
@@ -809,23 +809,31 @@ describe("GeoServices", () => {
...
+      }
 
-        // These fields should always be present
-        assert.isTrue(crs.name !== undefined);
-        assert.isTrue(crs.description !== undefined);
-        assert.isTrue(crs.deprecated === true || crs.deprecated === false);
+      // These fields should always be present
+      assert.isTrue(crs.name !== undefined);
+      assert.isTrue(crs.description !== undefined);
+      assert.isTrue(crs.deprecated === true || crs.deprecated === false);
```
tcobbs-bentley, line 836: I believe that this unindent was unintentional.

core/backend/src/test/misc/GeoServices.test.ts
```diff
@@ -840,26 +848,33 @@ describe("GeoServices", () => {
...
-      await validateCRSList(67, validationRangeSmall, extent);
+      await validateCRSList({expectedCount: 67, allowedRange: validationRangeSmall, extent});
+      await validateCRSList({expectedCount: 424, allowedRange: validationRangeSmall, extent, includeWorld: true});
     });
 
     it("should return CRS that are in the specified range (3)", async () => {
       const extent: Range2dProps = { low: { x: 0.3, y: 2.4 }, high: { x: 1.6, y: 3.77 } };
-      await validateCRSList(62, validationRangeSmall, extent);
+      await validateCRSList({expectedCount: 62, allowedRange: validationRangeSmall, extent});
+      await validateCRSList({expectedCount: 422, allowedRange: validationRangeSmall, extent, includeWorld: true});
```
tcobbs-bentley, line 873: Why do the tests at line 861 and 867 have 357 new systems, while this test has 360 new systems? Shouldn't the number of world systems be the same, no matter the allowed range?

ben-polinsky: That's probably a mistake, checking...

core/backend/src/GeographicCRSServices.ts
ben-polinsky: > don't think "apostrophe s" is ever allowed to indicate plural, but with acronyms, it might be accepted.

Seems to be bending toward allowing in some cases, but I agree. Looking it up now, it seems CRSs might be the best idea, which I don't hate.

tcobbs-bentley: I'm ok with CRSs as long as you update the `extent` comment to be consistent.

core/backend/src/test/misc/GeoServices.test.ts
ben-polinsky: It wasn't, the formatting is better here. But for the sake of a cleaner history, I'll fix.

tcobbs-bentley: Lines 826 through 836 can't be at the same indent level as the `for` statement on line 825.

core/backend/src/test/misc/GeoServices.test.ts
ben-polinsky: The numbers are correct...

tcobbs-bentley: Any idea why this one has 360 extras while the other two only have 357? It might indicate that something is wrong somewhere, or there could be some other logical explanation.

ben-polinsky: Ah good - I read the debugger wrong, it is 419: 
image

Phew :)

core/backend/src/test/misc/GeoServices.test.ts
ben-polinsky: Should be fine now

core/backend/src/GeographicCRSServices.ts
ben-polinsky: I've moved to coordinate reference systems in that file for consistency... we're all over the place across the repo.

ben-polinsky:

PR #8038, opened 2025-05-05
Deprecate `getSchemaSync` method in `ECSchemaRpcLocater` and enforce async usage

akshay05n:
**What’s Changed:**

- Deprecated the `getSchemaSync` method in `ECSchemaRpcLocater`.
- Updated its implementation to always throw an error, clearly stating that synchronous schema loading over RPC is not supported.

**New Behavior:**

Calling `getSchemaSync` will now immediately throw:
```typescript
Error: getSchemaSync is not supported. Use the asynchronous getSchema method instead.
```

### DIFF
--- a/common/api/ecschema-metadata.api.md
+++ b/common/api/ecschema-metadata.api.md
@@ -1443,7 +1443,7 @@ export enum PropertyType {
 }
 
 // @internal (undocumented)
-export function propertyTypeToString(type: PropertyType): "PrimitiveProperty" | "StructProperty" | "StructArrayProperty" | "NavigationProperty" | "PrimitiveArrayProperty";
+export function propertyTypeToString(type: PropertyType): "PrimitiveArrayProperty" | "PrimitiveProperty" | "StructArrayProperty" | "StructProperty" | "NavigationProperty";
 
 // @internal (undocumented)
 export namespace PropertyTypeUtils {
--- a/common/api/ecschema-metadata.api.md
+++ b/common/api/ecschema-metadata.api.md
@@ -1443,7 +1443,7 @@ export enum PropertyType {
 }
 
 // @internal (undocumented)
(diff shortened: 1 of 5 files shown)


common/api/ecschema-metadata.api.md
```diff
@@ -1443,7 +1443,7 @@ export enum PropertyType {
 }
 
 // @internal (undocumented)
-export function propertyTypeToString(type: PropertyType): "PrimitiveProperty" | "StructProperty" | "StructArrayProperty" | "NavigationProperty" | "PrimitiveArrayProperty";
```
rschili, line 1446: This broke the API build, revert it to the original order

full-stack-tests/core/src/frontend/standalone/SchemaLocator.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    await TestUtility.shutdownFrontend();
+  });
+
+  it("locate valid schema Async", async () => {
+    const schemaKey = new EC.SchemaKey("Gist", 1, 0, 0);
+    const schema = await context.getSchema(schemaKey, EC.SchemaMatchType.Exact);
+
+    assert.isDefined(schema);
+    assert.strictEqual(schema!.schemaKey.name, "Gist");
+    assert.strictEqual(schema!.schemaKey.version.toString(), "01.00.00");
```
rschili, line 38: I suggest a few more tests:
After calling getSchema() we should do a getSchemaSync, which due to caching is probably successful. This behavior seems inconsistent to me, but at the very least we should cover it with a test so we have it covered.

And another test that uses getSchemaItem/getSchemaItemSync on the context, instead of getSchema.

rschili (approved):

full-stack-tests/core/src/frontend/standalone/SchemaLocator.test.ts
akshay05n: Added new test cases

ColinKerr (approved):

aruniverse (approved):

aruniverse:

PR #8046, opened 2025-05-05
Add formatsProvider prop to IModelApp, update MeasureTool

### LINKED ISSUE #7807: Switch the typing used to retrieve formatProps for Quantity Formatting

hl662:
Closes #7807 
- Add non breaking, optional formatsProvider property to IModelApp, defaulting to `QuantityTypeFormatsProvider`
- Update core-frontend MeasureTool to use IModelApp.formatsProvider, with a fallback on quantityFormatter API.
- Add LocalFormatsProvider class to display test app, functions to load a .json file of a FormatSet locally.
- Update display test app to register a SchemaFormatsProvider on IModelConnection open.
- Add method overloads to `QuantityFormatter.formatString` and `QuantityFormatter.parseToQuantityValue` to ease formatting usage.

TODOs:
- [x] rush change
- [x] rush extract-api
- [x] add section in quantity docs on migrating from `QuantityType` to a respective `KindOfQuantity`
- [x] Add a test against ImodelApp.formatsProvider
- [ ] (In backport PR) add section in NextVersion.md


core/frontend/src/tools/MeasureTool.ts
```diff
@@ -24,11 +24,24 @@ import { PrimitiveTool } from "./PrimitiveTool";
...
+async function getFormatterSpecByKoQAndPersistenceUnit(koq: string, persistenceUnitName: string, fallbackQuantity: QuantityType): Promise<FormatterSpec | undefined> {
+  if (!IModelApp.formatsProvider) {
+    return IModelApp.quantityFormatter.findFormatterSpecByQuantityType(fallbackQuantity);
+  }
+  const formatProps = await IModelApp.formatsProvider.getFormat(koq);
+  if (undefined === formatProps)
+    return undefined;
+  const persistenceUnitProps = await IModelApp.quantityFormatter.unitsProvider.findUnitByName(persistenceUnitName);
+  const format = await Format.createFromJSON(koq, IModelApp.quantityFormatter.unitsProvider, formatProps)
+  return FormatterSpec.create("bleh", format, IModelApp.quantityFormatter.unitsProvider, persistenceUnitProps); // Name it better than bleh?
```
hl662, line 42: todo: update this

bbastings (commented):
What we really need to address are the FormattedQuantityDescription sub-classes used by tool settings/AbstractUI. Like LengthDescription. Measure tool is not at all representative of how the quantify parsers/formatters are normally used.

```
export class LengthDescription extends FormattedQuantityDescription {
  constructor(name?: string, displayLabel?: string, iconSpec?: string) {
    const defaultName = "length";
    super(
      name ? name : defaultName,
      displayLabel ? displayLabel : IModelApp.localization.getLocalizedString("iModelJs:Properties.Length"),
      iconSpec,
    );
  }

  public get formatterQuantityType(): QuantityType { return QuantityType.Length; }
  public get quantityType(): string { return "Length"; }

  public get parseError(): string { return IModelApp.localization.getLocalizedString("iModelJs:Properties.UnableToParseLength"); }
}
```

hl662:
> What we really need to address are the FormattedQuantityDescription sub-classes used by tool settings/AbstractUI. Like LengthDescription. Measure tool is not at all representative of how the quantify parsers/formatters are normally used.

Thanks for bringing this to my attention. 

Looking at your code snippet, quantityType() should be modified to not return a fixed value. We can rework the constructor to pass in an optional quantityType, with the default value being one of the KoQs part of [this table](https://github.com/iTwin/itwinjs-core/pull/8046/files#diff-66edb1fd7a647c360573e6f0a2c9538d674e7373c59562df7afa35636eec2f4cR30-R40). The [doccomment](https://github.com/iTwin/itwinjs-core/blob/master/ui/appui-abstract/src/appui-abstract/properties/Description.ts#L86-L90) for the quantityType property is encouraging, seems like it's already planned to possibly be a EC full name for a KindOfQuantity. If we're worried about bloating the number of args passed into the constructor, we'll just define an interface for `BaseQuantityDescriptionProps`

When creating the Parser/FormatterSpecs we can rely on `IModelApp.formatsProvider`, that `quantityType`, a `persistenceUnitProps` that is retrieved via `IModelApp.quantityFormatter.unitsProvider` (as for the persistenceUnit in question, each sub class Description might need to have a default like `Units.M`). 

We eventually get rid of the `formatterQuantityType` getter, as that'd be a duplicate of the `quantityType` getter. But for now we can keep that, to use as a fallback to generate a parser/formatterSpec for the Description subclasses in case IModelApp.formatsProvider is undefined.

Overall it doesn't seem like there's anything in these classes that worry me, one thing might be that the formatterSpec/parserSpec getters are how the specs are set by the classes, and those aren't async...

I would prefer reworking these classes in a separate PR after this one is merged in, I can open a separate GH issue.

bbastings:
> I would prefer reworking these classes in a separate PR after this one is merged in, I can open a separate GH issue.

Sounds good.

hl662:
GH issue opened: https://github.com/iTwin/itwinjs-core/issues/8056

docs/learning/frontend/QuantityFormatting.md
```diff
@@ -21,7 +21,31 @@ There are nine built-in quantity types (see [QuantityType]($frontend)). The Quan
...
+| Coordinate | AecUnits.LENGTH |
+| Stationing | RoadRailUnits.STATION |
+| LengthSurvey | RoadRailUnits.LENGTH |
+| LengthEngineering | AecUnits.LENGTH |
+
+[AecUnits](../../bis/domains/AecUnits.ecschema.md) is a Common layer schema that will be present in many iModels. [RoadRailUnits](../../bis/domains/RoadRailUnits.ecschema.md), a Discipline-Physical layer schema, contains Kind of Quantities used by Road & Rail schemas. More information on schemas and their different layers can be found in [Bis Organization](../../bis/guide/intro/bis-organization.md).
+
+iModels might not have AecUnits or RoadRailUnits schemas included, in such cases developers can address this through integrating their tools/components to use a [FormatsProvider]($quantity), and add the missing KindOfQuantity (and associated [FormatProps]($quantity)) through that FormatsProvider, independent from schemas coming from iModels. More information on `FormatsProvider` can be found in this learnings [section](../quantity/index.md/#formats-provider).
+
+We plan to deprecate `QuantityType` in iTwin.js 6.x.
```
grigasp, line 46: why not 5.0?

docs/learning/frontend/QuantityFormatting.md
```diff
@@ -21,7 +21,31 @@ There are nine built-in quantity types (see [QuantityType]($frontend)). The Quan
 
  Custom quantity types that implement the [CustomQuantityTypeDefinition]($frontend) interface may also be registered with the QuantityFormatter, see method `registerQuantityType`. See example implementation of a custom type [here](https://github.com/iTwin/itwinjs-core/blob/5905cb1a48c4d790b5389d7f0ea141bc3ce95f23/test-apps/ui-test-app/src/frontend/api/BearingQuantityType.ts).
 
-#### Overriding Default Formats
+#### Migrating from QuantityType to KindOfQuantity
+
+Starting in iTwin.js 5.0, we encourage developers to move away from using `QuantityType`, and to instead use `KindOfQuantity` [EC full names](https://www.itwinjs.org/bis/ec/ec-name/#full-name).
+
+Here is a table of replacements for each `QuantityType`:
```
grigasp, line 28: Should we also highlight that the new way provides access to much more formats and not just the ones listed in the table? That would also explain why the deprecation of `QuantityType` is being done in the first place.

core/frontend/src/tools/MeasureTool.ts
```diff
@@ -24,11 +24,24 @@ import { PrimitiveTool } from "./PrimitiveTool";
...
+import { Format, FormatterSpec } from "@itwin/core-quantity";
 
 function translateBold(key: string) {
   return `<b>${CoreTools.translate(`Measure.Labels.${key}`)}:</b> `;
 }
 
+async function getFormatterSpecByKoQAndPersistenceUnit(koq: string, persistenceUnitName: string, fallbackQuantity: QuantityType): Promise<FormatterSpec | undefined> {
+  if (!IModelApp.formatsProvider) {
+    return IModelApp.quantityFormatter.findFormatterSpecByQuantityType(fallbackQuantity);
+  }
```
grigasp, line 36: Could `IModelApp` have a default fallback formats provider, that maps given known koq name to `QuantityType` and calls quantity formatter to get the format? That would help consumers move away from quantity types towards using `IModelApp.formatsProvider`, since it would always be available.

core/frontend/src/tools/MeasureTool.ts
```diff
@@ -24,11 +24,24 @@ import { PrimitiveTool } from "./PrimitiveTool";
...
+  const formatProps = await IModelApp.formatsProvider.getFormat(koq);
+  if (undefined === formatProps)
+    return undefined;
+  const persistenceUnitProps = await IModelApp.quantityFormatter.unitsProvider.findUnitByName(persistenceUnitName);
+  const format = await Format.createFromJSON(koq, IModelApp.quantityFormatter.unitsProvider, formatProps)
+  return FormatterSpec.create(`${koq}_format_spec`, format, IModelApp.quantityFormatter.unitsProvider, persistenceUnitProps);
```
grigasp, line 42: It looks like this would need to be done by every possible consumer who wants to format value from persistence units to some format. Maybe quantity formatter could have a function that does this? E.g.:

```ts
async createFormatterSpec(
  props: {
    persistenceUnitName: string;
    formatProps: FormatProps;
    formatName?: string; // is this needed for a temporary format? maybe just assign some arbitrary string or Guid.createValue()?
  },
): Promise<FormatterSpec>;
```

core/frontend/src/tools/MeasureTool.ts
```diff
@@ -259,15 +272,15 @@ export class MeasureDistanceTool extends PrimitiveTool {
...
       totalDistance += adjustedPoints[i].distance(adjustedPoints[i + 1]);
     if (0.0 === totalDistance)
       return;
 
-    const formatterSpec = IModelApp.quantityFormatter.findFormatterSpecByQuantityType(QuantityType.Length);
-    if (undefined === formatterSpec)
+    const formatterSpec = await getFormatterSpecByKoQAndPersistenceUnit("AecUnits.LENGTH", "Units.M", QuantityType.Length);
+    if (formatterSpec === undefined)
       return;
     const formattedTotalDistance = IModelApp.quantityFormatter.formatQuantity(totalDistance, formatterSpec);
```
grigasp, line 285: It looks like requesting the formatter spec and then calling `formatQuantity` is going to be a very common practice. Maybe we could make this easier on consumers by moving formatter spec retrieval into a `formatQuantity` overload? E.g.:

```ts
async formatQuantity(
  props: {
    value: number;
    valueUnitName: string; // should we call it "persistence" unit, when it's not always related to persistence per se, e.g. in this case it's a dynamically generated value
    targetKindOfQuantityName: string;
  }
): Promise<string>;
formatQuantity(
  props: {
    value: number;
    formatSpec: FormatterSpec;
  }
): string;
// I'd deprecate this overload and only keep the above
formatQuantity(value: number, spec: FormatterSpec): string;
```

ColinKerr: I agree persistence might not be the right term here.  Value seems ok so long as we're consistent with naming and always call the input number as the value

docs/learning/frontend/QuantityFormatting.md
aruniverse: Mybe we cant deprecate for 5.0, but it shouldnt stop us from deprecating in say 5.x; i would prefer to avoid pushing it as far out as 6

core/frontend/src/tools/MeasureTool.ts
hl662: I don't know if we can deprecate the existing way to pass in args, `formatQuantity` (and by extension, `QuantityFormatter`) is marked as public?

core/frontend/src/tools/MeasureTool.ts
hl662: we could... some QuantityTypes will be converged though, Length, Cooridnate, and LengthEngineering are all mapped to AecUnits.Length, but that's a minor trade off

docs/learning/frontend/QuantityFormatting.md
hl662: updated the docs to say we'll deprecate during the 5.x lifecycle

docs/learning/frontend/QuantityFormatting.md
hl662: I added a paragraph explaining why we are moving away from `QuantityType`, it mentions being able to access more formats and make it easier

core/frontend/src/tools/MeasureTool.ts
aruniverse: we cant deprecate the current method and provide a new overload?

core/frontend/src/tools/MeasureTool.ts
hl662: I like it. I've added a `createFormatterSpec` and `createParserSpec` helper methods inside `QuantityFormatter` that reflects the suggestion you made. formatName is optional in the interface I made for the args. This way we can also control the unitsProvider used will be the one inside `QuantityFormatter`.

core/frontend/src/tools/MeasureTool.ts
hl662: oh, I also don't think we can do this overload, generating the specs needs to be an async method, so the alternative is to create a new method called `formatQuantityAsync` if that's alright with you

grigasp: you can have overloads returning different types, e.g.:
```ts
function x(arg: string): string;
function x(arg: number): Promise<string>;
function x(arg: string | number): string | Promise<string> {
  if (typeof arg === "string") {
    return arg;
  }
  return Promise.resolve(arg.toFixed(2));
}
```

core/frontend/src/tools/MeasureTool.ts
grigasp: I think having formats provider always available is a very important aspect. If it can be `undefined`, then consumers of the API will always have to think about fallbacks, possibly causing them to use quantity types or some other alternatives.

hl662: I added a BasicFormatsProvider that's init by default if no formatsProvider is passed into IModelApp.startup. It contains a limited set of formats mapped to the KoQs we agreed on, using QuantityType's formatProps under the hood. It's marked internal, we don't want people to use this as a replacement for proper impls of FormatsProviders

core/frontend/src/tools/MeasureTool.ts
hl662: I'm not able to make the function return a union between a non promise and a promise value....

Doing something like the below, makes the linter throw an error saying functions that return promises must have `async` declared after `public`

```ts
  public formatQuantity(props: {
    value: number;
    valueUnitName: string;
    kindOfQuantityName: string;
  }): Promise<string>;
  public formatQuantity(magnitude: number, formatSpec?: FormatterSpec): string;
  public formatQuantity(args: number | object, spec?: FormatterSpec): string | Promise<string> {
    if (typeof args === "number") {
      /** Format a quantity value. Default FormatterSpec implementation uses Formatter.formatQuantity. */
      const magnitude = args;
      if (spec)
        return spec.applyFormatting(magnitude);
      return magnitude.toString();
    }

    return Promise.resolve(this.formatQuantityAsync(args as {
      value: number;
      valueUnitName: string;
      kindOfQuantityName: string;
    }))
  }
```

core/frontend/src/IModelApp.ts
```diff
@@ -763,6 +763,16 @@ export class IModelApp {
...
 
+  /**
+   * Resets the formatsProvider back to the default [[BasicFormatsProvider]].
+   * @beta
+   */
+  public static resetFormatsProvider() {
+    const oldFormatsProvider = this.formatsProvider;
+    this.formatsProvider = new BasicFormatsProvider();
+    oldFormatsProvider.onFormatsChanged.raiseEvent({ formatsChanged: "all"})
+  }
```
grigasp, line 774: This showcased an interesting situation.

Let's say property widget, which relies on formatting, registers an `onFormatsChanged` listener. Then, at some point, application using the widget decides to change or reset the formats provider - widget would be listening on the stale provider. How do we solve that?

One idea I have is to always have the same formats provider on `IModelApp` - it would be used to re-raise events from and forward calls to underlying implementation of the actual formats provider - the "basic" one or app-supplied. So instead of setting them on the `IModelApp`, you'd set them on the IModelApp's formats provider. What do you think?

core/frontend/src/tools/MeasureTool.ts
grigasp: That's a lint rule problem, that was fixed with `@typescript-eslint/eslint-plugin` version `8.17.0` (we're using `8.11.0` in `itwinjs-core`). The fix is here: https://github.com/typescript-eslint/typescript-eslint/pull/10304. The situation is also explained in [docs](https://typescript-eslint.io/rules/promise-function-async/):

> When functions return unions of Promise and non-Promise types implicitly, it is usually a mistake—this rule flags those cases. If it is intentional, make the return type explicitly to allow the rule to pass.

There's also an example of this situation there.

I think you can safely add a lint ignore comment there for now, until someone bumps the plugin dep version.

core/frontend/src/IModelApp.ts
hl662: I see that yes... without modifying the interface of a formatsProvider further to tackle this, I think the root of the solution is having some sort of 'manager' in IModelApp to handle this. Maybe we can repurpose the QuantityFormatter to do so: add a setFormatsProvider(), a BeEvent for onFormatsProvider changed (or onFormatsChanged that raises the FormatsChangedArg defined in core-quantity). We keep the formatsProvider opt in IModelAppOptions, and if user supplied one then we set quantityFormatter.formatsProvider during startup.

Users would get formats through IModelApp.quantityFormatter.formatsProvider.

grigasp: I don't mind formats provider being on quantity formatter rather than `IModelApp`, but I'd hate to see `onFormatsProviderChanged` or another `onFormatsChanged` event.

- The former would mean that, in my property widget example, I have to listen to two events: `QuantityFormatter.onFormatsProviderChanged` and `FormatsProvider.onFormatsChanged`.
- The latter would duplicate the `onFormatsChanged` event and I think it wouldn't be clear which one to use for consumers.

My initial suggestion doesn't require changing the `FormatsProvider` interface, but instead make the implementation act as the manager that you mention. We don't need to expose the implementation - it's enough that `IModelApp` or `QuantityFormatter` knows it.

core/frontend/src/tools/MeasureTool.ts
hl662: I ignored the lint rule, and added the overload. Also added a similar funciton overload for `parseToQuantityValue`

core/frontend/src/tools/MeasureTool.ts
hl662: Gotcha, renamed the class

core/frontend/src/IModelApp.ts
```diff
@@ -298,11 +298,17 @@ export class IModelApp {
...
   /** The [[FormatsProvider]] for this session.
-   * @param provider The provider to use for formatting quantities. If not provided, the provider will be cleared.
+   * @param provider The provider to use for formatting quantities.
    * @beta
    */
   public static get formatsProvider(): FormatsProvider { return this._formatsProvider; }
-  public static set formatsProvider(provider: FormatsProvider) { this._formatsProvider = provider }
+  public static set formatsProvider(provider: FormatsProvider) {
+    this._formatsProvider.formatsProvider = provider;
+    this._formatsProvider.formatsProvider.onFormatsChanged.addListener((args: FormatsChangedArgs) => {
```
grigasp, line 307: why not do this in `FormatsProviderManager.formatsProvider` setter?

hl662: I was 50/50 on this yeah. I've moved the events logic into FormatsProviderManager

grigasp (approved):

hl662:

aruniverse (approved):

core/frontend/src/tools/MeasureTool.ts
```diff
@@ -354,7 +365,7 @@ export class MeasureDistanceTool extends PrimitiveTool {
       builderDynHid.addLineString(tmpPoints);
 
       context.addDecorationFromBuilder(builderDynHid);
-      this.displayDynamicDistance(context, tmpPoints, tmpAdjustedPoints);
+      void this.displayDynamicDistance(context, tmpPoints, tmpAdjustedPoints);
```
bbastings, line 368: Don't like that you are ignoring the getFormatterSpecByKoQAndPersistenceUnit promise here...

hl662: I've added internal `formatterSpec` members to MeasureDistanceTool as we discussed internally, and added listeners to onPostInstall and respective clean up logic for the new listener.

bbastings: Much nicer!

bbastings (approved):

PR #8048, opened 2025-05-06
Add documentation for metadata package

rschili:
Fixes: https://github.com/iTwin/itwinjs-backlog/issues/1437


docs/learning/metadata/index.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Packages
+
+The metadata packages implement the abstract concepts of EC in typescript [EC overview](../../bis/ec/index.md)
+The following packages exist:
+
+- ($ecschema-metadata) is the basic package that exposes metadata objects.
+- ($ecschema-editing) Since the basic package only implements an API for understanding schemas, this package adds editing on top of it.
+- ($ecschema-locaters) contains classes for locating and loading EC schema files from the file system.
+- ($ecschema-rpc) provides a schema RPC interface implementation
```
hl662, line 11: should this line be split into 2? the 2 npm packages users would want are `ecschema-rpcinterface-impl` and `ecschema-rpcinterface-common`

rschili: Ah, I never used these so I didn't know. I still need to make sure those variables resolve correctly, the documentation process could use some documentation. I think it's in `docs/config/docSites.json` but I'm not sure yet

rschili: I ended up removing those packages as I noticed we do not publish API docs for them. It may be more confusing than good to have them listed here.

ColinKerr:
Update existing docs to point to the new way of doing things:
- Snippet `IModelSchemas.loadFromDb` and associated docs.  docs/learning/serializing-xml-schemas.md
- Snippet `DynamicSchema-syncDynamicSchema.example-code` ... this one is connector focused but still we should update the docs to use the shared context.
- `SchemaLoader` docs should be updated to point people to the contexts on IModelDb and IModelConnection core/ecschema-metadata/src/SchemaLoader.ts

docs/learning/metadata/index.md
```diff
@@ -0,0 +1,73 @@
...
+``` ts
+[[include:Metadata.entitiesFromIModelDb]]
+```
+
+The provided [SchemaMatchType]($ecschema-metadata) specifies what schemas are acceptable, for example if the caller only cares that the returned schema is read compatible with the requested version.
+
+## Items in a schema
+Like schemas are identified by [SchemaKey]($ecschema-metadata), items inside a schema are identified by [SchemaItemKey]($ecschema-metadata).
+
+The methods for getting items inside a schema, like [SchemaContext.getSchemaItem]($ecschema-metadata) or [Schema.getItem]($ecschema-metadata) follow a pattern where you can either get all items, or filter for a specific item type. The latter is done by passing the type of the desired item to the method like with EntityClass in the example in section above.
```
ben-polinsky, line 27:
```suggestion
The methods for getting items inside a schema, like [SchemaContext.getSchemaItem]($ecschema-metadata) or [Schema.getItem]($ecschema-metadata), follow a pattern where you can either get all items, or filter for a specific item type. The latter is done by passing the type of the desired item to the method like with EntityClass in the example in section above.
```

docs/learning/serializing-xml-schemas.md
```diff
@@ -50,7 +50,7 @@ We can now use the file system API to write the xml to file.
 
 ## Schemas from an iModel
 
-First use the [SchemaLoader]($ecschema-metadata) to create a SchemaContextEditor that knows about the schemas in the iModel
+The suggested way for this is to use [IModelDb.schemaContext] to access the schemas and to create a SchemaContextEditor that knows about the schemas in the iModel
```
ben-polinsky, line 53:
```suggestion
Use [IModelDb.schemaContext] to access the schemas and to create a SchemaContextEditor that knows about the schemas in the iModel:
```

ben-polinsky (approved):
Few nits, but looks good.

docs/bis/ec/ec-custom-attributes.md
```diff
@@ -36,3 +36,5 @@ For example:
 ```
 
 The custom attribute concept allows any custom metadata to be applied because they are defined using a special type of ECClass, see [ECCustomAttributeClass](./ec-custom-attribute-class.md) for more details
+
+Note: The order of custom attributes within its parent container is not guaranteed. Our APIs will return a consistent order, but it may not always be the same as the order in which they were defined in the schema. This should not be a problem since the same CA may only be applied once.
```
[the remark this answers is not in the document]
rschili: you're right, I changed it to spell out

docs/learning/serializing-xml-schemas.md
rschili: done

docs/learning/metadata/index.md
rschili: done

docs/bis/ec/ec-custom-attributes.md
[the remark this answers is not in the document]
ben-polinsky: thanks

ColinKerr (approved):

PR #8049, opened 2025-05-06
Shared lock on parent implies shared lock on children

pmconne:
`ServerBasedLocks.holdsSharedLock` claims otherwise, but holding a shared lock on an owner element prevents anyone else from acquiring an exclusive lock on any of its owned elements, which is the definition of a shared lock. Nevertheless `holdsSharedLock` returns false unless the element itself is locked or one of its owners is **exclusively** locked.

@laurynas111 found this confusing and surprising.

This is a proposed change. @kabentley what am I misunderstanding?

The tests in ServerBasedLocks.test.ts fail with this change (they are counting invocations of private methods and querying the lock db directly).

### DIFF
--- a/core/backend/src/internal/ServerBasedLocks.ts
+++ b/core/backend/src/internal/ServerBasedLocks.ts
@@ -68,11 +68,12 @@ export class ServerBasedLocks implements LockControl {
     });
   }
 
-  private getLockState(id?: Id64String): LockState | undefined {
-    return (id === undefined || !Id64.isValid(id)) ? undefined : this.lockDb.withPreparedSqliteStatement("SELECT state FROM locks WHERE id=?", (stmt) => {
+  private getLockState(id?: Id64String): LockState {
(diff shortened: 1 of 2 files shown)


kabentley:
> ServerBasedLocks.holdsSharedLock claims otherwise, but holding a shared lock on an owner element prevents anyone else from acquiring an exclusive lock on any of its owned elements, which is the definition of a shared lock.

No, I don't think that's right. You cannot obtain an exclusive lock on an element unless you first hold a shared lock on its parents. However, multiple people may hold shared locks on an element (that's why it's called shared) and each may acquire exclusive locks on children (though of course not on the same child.) The (sole, as far as I know) purpose of a shared lock is to disallow deleting a parent while one of its children is being edited in another briefcase.

pmconne:
> I don't think that's right.

It's not - thanks. @laurynas111, make sense?

laurynas111:
Got it.
I'll switch to using Exclusive locks then for locking full branch as this seems to be more effective than locking each Element with a shared lock in the hierarchy. The branch that I'm locking is under application channel and stores only application rank categories.

PR #8051, opened 2025-05-06
Increase timeout for dynamic geometry clip mask tests

danielzhong-bentley:
Added timeout to two additional tests in the planar clip mask suite to account for shader compilation time related to draping. These tests were missed in the previous PR, as CI failures only showed the first two tests at that time.


full-stack-tests/core/src/frontend/map/PlanarClipMask.test.ts
```diff
@@ -206,7 +207,8 @@ describe("Planar clip mask (#integration)", () => {
     });
   });
 
-  it("is masked by priority by dynamic geometry", async () => {
+  it("is masked by priority by dynamic geometry", async function () {
+    this.timeout(480000);
```
pmconne, line 211: Put a comment explaining why, so we can come back and revert when/if we fix whatever caused the tests to start timing out more frequently.

pmconne (approved):

Issue #6793, opened 2024-06-05
Provide a text annotation layout API for use on front-end


Alfonso-Martello:
**Is your feature request related to a problem? Please describe.**
I am working on creating a caret API for drawing production. A caret is a marker used to indicate a location within a text block for performing actions (like insert). The caret API needs to know its location in the layout, but the layout function/classes are marked as internal.

**Describe the solution you'd like**
Please explicitly export `layoutTextBlock` and `TextBlockLayout` and remove the `@internal` tag. 

**Describe alternatives you've considered**
Normally, I could access this internal code for development in the short-term. However, I am working in iTwin Studio, and iTwin Studio has a build time check that disallows deep iTwin.js imports such as “import X from @itwin/core-backend/lib/cjs/TextAnnotationLayout”.

Alfonso-Martello:
@pmconne, I assume this is a simple change of removing the `@internal` tag and adding the mentioned classes/functions to the export list. If you do not have any objections to or questions about making the above class and function available, I will submit a PR making the necessary changes.

pmconne:
Yeah I kept them internal in part because they are messier than they should be for public APIs, and in part because I suspect you will want your caret to reside on the front-end whereas these are backend-only APIs.

I briefly had a public API in core-common that would provide a representation of the layout usable on the frontend but I had no use for it at that time so removed it to reduce API surface area.

Why don't we talk about what your API will look like and how it will use the layout.

pmconne:
Yeah that's basically what the core-common representation of the layout was, removed in [this commit](https://github.com/iTwin/itwinjs-core/pull/6572/commits/99114eb49044457e185789486049b10a995dc58c#diff-82d7b519d69272ac46ac8c86a821bb55f9a680b72b27db4016d4664aaac2ad5c).

Please make sure you have read [this](https://dmitripavlutin.com/what-every-javascript-developer-should-know-about-unicode/) and are routinely testing with emojis, Asian characters, combining characters, etc.

Alfonso-Martello:
That commit looks close to what I need. Currently, `layoutTextBlock` is `@internal` and the results are not persisted. I would need that function exposed or the calculated result persisted/returned somewhere not internal.

That article was a very informative read, thank you for sharing it. I will account for special characters and make sure to test with various tricky characters.

Alfonso-Martello:
Yes.

[...]

PR #8055, opened 2025-05-07
Text Annotation Frames

### LINKED ISSUE #7677: TextAnnotations - `fillColor` property

I would like to be able to specify a `fillColor` around a `TextAnnotation` similar to how we describe the `width`. 

The most common use-case is that a user has a label overlapping geometry. They want to be able to specify a fill (usually just the bg color) to make the text more legible.

`TextBlock` should have a property called `fillColor` with the following attributes
* Default should be transparent
* When defined, a rectangle should be drawn behind the text with the given bg color. This rectangle should fill to the margins
* Optionally, it could be helpful for a user to be able to pass in some geometry themselves (say they wanted an oval instead). Default should be rectangle.

claudiaareneee:
## Description

I've been working on the text annotation wishlist items like frames and fillcolor. 

Core changes:
- Created a new folder in the backend for `annotations` and moved related files into it. 
  - `FrameGeometry.ts` (new)
    - I had fun reliving my highschool geometry class for this. But this file is just a bunch of helper methods for calculating the curve primitives for different types of frames.
  - `TextAnnotationElement.ts`
    - Removed `updateAnnotation`.
    - Users should still call `setAnnotation` to change just the annotation of the element. But the geometry is no longer generated here. Instead, geometry is calculated using `TextAnnotationGeometry.appendTextAnnotationGeometry`.
    - Currently out of covenience, I'm calling the `TextAnnotationGeometry.appendTextAnnotationGeometry` in the `toJson` methods of each element, but I could be convinced that this is misleading/wrong and that the user should call this method explicitly.
  - `TextAnnotationGeometry.ts` (new)
    - creates and exposes new methods for producing text annotation geometry by appending the frame geometry and textblock geometry to an ElementGeometry.Builder that's passed in.
    - The debug geomtry logic was moved here.
  - `TextBlockGeometry.ts`(formerly `TextAnnotationGeometry`)
    - removed `produceTextAnnotationGeometry`
    - exposes `produceTextBlockGeometry`
    - moves all the debug logic to `TextAnnotationGeometry`
    - now this file is more concise
  - `TextBlockLayout.ts` (formerly `TextAnnotationLayout`)
    - renamed to be more explicit
    - Fixes a bug 
- `core/backend/src/core-backend.ts`: exposes the new files along with `layoutTextBlock`
- `core/common/src/annotation/TextAnnotation.ts`: Adding properties for frame and fill. These should probably be in the TextStyles, but that's a whole other discussion
- `core/common/src/geometry/ElementGeometry.ts`: adding param for overrides to be reused

Tests
- `core/backend/src/test/annotations/TextAnnotation.test.ts`: tests roundtripping, element encoding, etc.
- `core/backend/src/test/annotations/TextBlock.test.ts`: tests the produceTextBlockGeometry method

Display test app changes:
- `backend/Backend.ts`
- `common/DtaRpcInterface.ts`
- `frontend/TextDecoration.ts`

## Todos / questions

- [x] Comments -> extract API
- [x] Documentation

## Link to Issue

* closes #7677


core/backend/src/strokers/TextAnnotationStroker.ts
```diff
@@ -0,0 +1,174 @@
...
+      this.debugSnapPoints(annotation.frame, dimensions, transform);
+
+    if (debugAnchorPoint)
+      this.debugAnchorPoint(annotation, layout, transform);
+
+    return { format: "flatbuffer", data: this._builder.entries };
+  }
+
+  public override async computeLayoutResults(props: TextAnnotationStrokerProps, requestProps: StrokerGraphicsRequestProps, options?: TextAnnotationStrokerOptions): Promise<TextAnnotationStrokerResults> {
+    // Todo: Should this be IModelError?
```
pmconne, line 76: It should be `ITwinError`.

core/common/src/annotation/FrameGeometry.ts
claudiaareneee, line 1: I had fun reliving my highschool geometry class for this. But this file is just a bunch of helper methods for calculating the curve primitives for different types of frames. It's mostly commented but I need to add a comment about how the computePolygon method doesn't frame as tightly as some of the individual compute methods.

core/common/src/annotation/TextAnnotation.ts
claudiaareneee, line 1: Adding properties for frame and fill. These should probably be in the TextStyles, but that's a whole other discussion

test-apps/display-test-app/src/backend/Backend.ts
claudiaareneee, line 1: The dta changes are just introducing a new RPC endpoint for the stroker, removing the old endpoint, and adding keyins for the new variables

core/backend/src/annotations/FrameGeometry.ts
```diff
@@ -0,0 +1,299 @@
...
+    } else if (frame.fill !== "subcategory") {
+      params.fillColor = ColorDef.fromJSON(frame.fill);
+      params.lineColor = params.fillColor;
+      params.fillDisplay = FillDisplay.Always;
+    }
+
+    if (frame.border !== "subcategory") {
+      params.lineColor = ColorDef.fromJSON(frame.border);
+      params.weight = frame.borderWeight;
+    }
```
claudiaareneee, line 47: @pmconne, I remember you had feelings about using `FillDisplay.Always` here vs `BlankingFill`, but I can't remember what they are. I'm using `Always` (in combo with `params.elmPriority`) because `Blanking` is applying to all geometry and not just the geometry internal to my builder. I'm up for suggestions. 

Also I'm having an issue with using the `elmPriority` to ensure the text is drawn over the background. Currently I'm having to stradle the `elemPriority` accross `FrameGeometry` and `ElementGeometry` to make sure the params apply propperly. Instead I just want to have it in my `TextAnnotationGeometry` file. I'm still investigating this, but if you have any opinions on this, that would be helpful.

claudiaareneee: Maybe this will help. These are the different configurations I've tried:

https://github.com/user-attachments/assets/a66de461-6160-4ea1-ad10-e300d7592457

claudiaareneee: If not element priority, what should I be using instead?

claudiaareneee: @pmconne, thoughts?

pmconne: Supply .bim file(s) containing the problematic text.

claudiaareneee: Zip: [house.bim.zip](https://github.com/user-attachments/files/20214154/house.bim.zip)

Copy and pastable keyins:
```txt
dta text init
dta text center
dta text height 0.25 // Text size (text height)
dta text width 6 // Width of the text box
dta text color #FDFCDC
dta text text "You were the chosen one! It was said that you would destroy the Sith, not join them. You were to bring balance to the Force, not leave it in darkness."
dta text break
dta text break
dta text text "— Obi-wan Kenobi, "
dta text underline
dta text text "Revenge of the Sith"
dta text margin all 0.25
dta text frame shape capsule
dta text frame border #0081A7
dta text frame borderWeight 10
dta text frame fill #00AFB9"
```

pmconne: Ok so you are not actually putting elements into the iModel. I wanted to see the geometry streams. Do I have to copy-patse each line one at a time into display-test-app's key-in field?

pmconne: nvm `dta macro` exists.

pmconne: The tile generator does not produce edges for shapes with blanking fill. Blanking fill controls order of display within a single element. Element priority controls order of display between different elements. You shouldn't hard-code any specific priority values in your geometry generation code. Your app probably should provide tools to change priorities of 2d elements relative to one another (a la PowerPoint's "move to front/back", "move backward/forward", etc.

I pushed fixes. The frame now draws in front of the fill.

I failed to reproduce your issue with the building geometry displaying in front of the fill. If that persists, please start a separate comment thread with details and repro steps.

pmconne:
> @pmconne, thank you for your suggestions, but I still have an issue with blanking fill. Here's a quick video (with audio). I'm using the same bim file from earlier:

To summarize: with latest changes, open Elevation-Front Views - View 1 and run the key-ins. The house geometry appears in front of the text fill. Something different happens in 3d which we will ignore for now.

I can't copy key-ins from a video, can you please provide them?

If you replace `Blanking` with `Always` and change nothing else, this problem vanishes?

claudiaareneee:
Just changing blanking fill in `FrameGeometry` L36 & L37 to 'Always' produces this: 
image

claudiaareneee:
@pmconne, it's possible that this view is just built weird. I tested on a different iModel and it seemed to be working fine

pmconne:
> @pmconne, it's possible that this iModel is just built weird. I tested on a different iModel and it seemed to be working fine

Nope, there's a bug with blanking fill. It's been broken forever. I'll fix it in a separate PR.

pmconne:
@claudiaareneee for your 3d weirdness, see if toggling Force Surface Discard ON changes things.
image

claudiaareneee:
@pmconne, I'll check and see what happens with that. 

However, at the moment I'm stuck on tests. I can't get it to insert the text element in the `core/backend/src/test/annotations/TextAnnotation.test.ts` persistance tests. I have one suite for `TextAnnotation2d` and another for `TextAnnotation3d`. (The original code only tested the 3d, so I've added another set for 2d). Both test suites are failing at the same place. 

It keeps failing on the `el0.insert()` call.  I believe I've isolated the error to the geometry builder's `appendTextString` method but I could be wrong. The failures are only happening when I have a textblock with content to insert.

✅ <span style="color:green">roundtrips an empty annotation</span>: passes
✅ <span style="color:green">roundtrips an annotation with a style</span>: passes
❌ <span style="color:red">roundtrips an annotation with a textBlock</span>: fails

The error I'm getting is:

```
Error inserting element [BuildGeometryStream failed]
    at Elements.insertElement (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/core/backend/src/IModelDb.ts:782:7)
    at TextAnnotation2d.insert (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/core/backend/lib/cjs/Element.js:183:104)
    at test (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/core/backend/src/test/annotations/TextAnnotation.test.ts:286:22)
    at Context.<anonymous> (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/core/backend/src/test/annotations/TextAnnotation.test.ts:311:75)
    at callFn (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runnable.js:364:21)
    at Test.Runnable.run (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runnable.js:352:5)
    at Runner.runTest (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runner.js:677:10)
    at /Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runner.js:800:12
    at next (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runner.js:592:14)
    at /Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runner.js:602:7
    at next (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runner.js:485:14)
    at Immediate._onImmediate (/Users/claudianelson/Documents/dp-cospace/repos/itwinjs-core/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runner.js:570:5)
    at processImmediate (node:internal/timers:478:21) {
  errorNumber: 65539,
  metadata: {
    elProps: {
      classFullName: 'BisCore:TextAnnotation2d',
      code: [Code],
      model: '0x24',
      jsonProperties: [Object],
      category: '0x16',
      placement: [Placement2d],
      elementGeometryBuilderParams: [Object]
    }
  }
}
        1) roundtrips an annotation with a textBlock
```

I believe the error I'm getting is [IModelStatus.BadArg](https://github.com/iTwin/itwinjs-core/blob/ba8d02fbf9a42c1f6260ff169486c88b03daaae7/common/api/core-bentley.api.md?plain=1#L840) unless I'm missreading that. 

I went ahead and pushed these tests. Looking at lines 311 (test that's failing), 286 (the line that's failing).

As a test, I added a backend ipc method to display-test-app to make sure that I can insert the TextAnnotation2d element. That's working. So something is wrong with the tests. Any thoughts?

I think I must not be mocking something that should be? But I don't know...

pmconne:
> Error inserting element [BuildGeometryStream failed]

That's [this function](https://github.com/iTwin/imodel-native/blob/df376a539940576f918bb64dcc2fe923f727fd81/iModelCore/iModelPlatform/DgnCore/ElementGeometry.cpp#L3305). It doesn't like your element geometry builder params.
Show what `elementGeometryBuilderParams` looks like.

> As a test, I added a backend ipc method to display-test-app to make sure that I can insert the TextAnnotation2d element. 

Using the exact same seed iModel that you are creating a copy of in your test?
Check what version of the BisCore schema that seed iModel contains.
Try changing your test to create a brand-new iModel from scratch instead of creating one from a seed (usually preferable for many reasons, including ensuring you have up to date versions of schemas).

claudiaareneee:
@pmconne, I did not do my manual test with the same imodel, so let me see what happens there. Do you know of any tests that create an iModel that I could use as an example?

To me the elementGeometryBuilderParams seem normal, but I can't parse them. This is what they look like:

```
Claudia Test >  elementGeometryBuilderParams [
  {
    opcode: 4,
    data: Uint8Array(12) [
      8, 0, 0, 0, 4,
      0, 4, 0, 4, 0,
      0, 0
    ]
  },
  {
    opcode: 22,
    data: Uint8Array(200) [
       20,   0,   0,   0,   0,   0,  14,   0, 116,   0, 115,   0,
        0,   0, 108,   0, 104,   0,   4,   0,  14,   0,   0,   0,
      138, 175, 245, 151, 192, 254, 239,  63,  39, 220, 247, 201,
       88, 222, 145, 191,  39, 101, 234,   3,  29, 246,  51, 191,
      202,   3, 132, 200,  39,  77,  48, 192,  30, 221, 137,  43,
       11, 223, 145,  63, 180, 130,  95,  60, 129, 253, 239,  63,
       39, 220, 247, 201,  88, 222, 145,  63, 143, 120,  92, 236,
       94,  35,  23, 192,   0,   0,   0,   0,   0,   0,   0,   0,
       30, 221, 137,  43,
      ... 100 more items
    ]
  },
  {
    opcode: 22,
    data: Uint8Array(192) [
       20,   0,   0,   0,   0,   0,  14,   0, 116,   0, 115,   0,
        0,   0, 108,   0, 104,   0,   4,   0,  14,   0,   0,   0,
      138, 175, 245, 151, 192, 254, 239,  63,  39, 220, 247, 201,
       88, 222, 145, 191,  39, 101, 234,   3,  29, 246,  51, 191,
      138,  51, 133, 250, 231,  85,  33, 192,  30, 221, 137,  43,
       11, 223, 145,  63, 180, 130,  95,  60, 129, 253, 239,  63,
       39, 220, 247, 201,  88, 222, 145,  63, 128, 144, 148, 137,
      237, 154,  22, 192,   0,   0,   0,   0,   0,   0,   0,   0,
       30, 221, 137,  43,
      ... 92 more items
    ]
  },
  {
    opcode: 22,
    data: Uint8Array(200) [
       20,   0,   0,   0,   0,   0,  14,   0, 116,   0, 115,   0,
        0,   0, 108,   0, 104,   0,   4,   0,  14,   0,   0,   0,
      138, 175, 245, 151, 192, 254, 239,  63,  39, 220, 247, 201,
       88, 222, 145, 191,  39, 101, 234,   3,  29, 246,  51, 191,
      182, 173,  46, 151,  24,  49,  16, 192,  30, 221, 137,  43,
       11, 223, 145,  63, 180, 130,  95,  60, 129, 253, 239,  63,
       39, 220, 247, 201,  88, 222, 145,  63, 107, 101, 209,  37,
       39, 149,  19, 192,   0,   0,   0,   0,   0,   0,   0,   0,
       30, 221, 137,  43,
      ... 100 more items
    ]
  },
  {
    opcode: 10,
    data: Uint8Array(88) [
       98, 103,  48,  48,  48,  49, 102,  98,  12,   0,   0,   0,
        8,   0,  10,   0,   9,   0,   4,   0,   8,   0,   0,   0,
       12,   0,   0,   0,   0,   1,   6,   0,   4,   0,   4,   0,
        6,   0,   0,   0, 240, 187, 113,  20, 248,  45,  16, 192,
       19,  73, 240,  95,  76,  72,  20, 192,   0,   0,   0,   0,
        0,   0,  89,  64,   0,   0,   0,   0,   0, 240, 226,  60,
      255, 255, 255, 255, 255, 255,  19, 192,   0,   0,   0,   0,
        0,   0,  89,  64
    ]
  },
  {
    opcode: 22,
    data: Uint8Array(192) [
       20,   0,   0,   0,   0,   0,  14,   0, 116,   0, 115,   0,
        0,   0, 108,   0, 104,   0,   4,   0,  14,   0,   0,   0,
      138, 175, 245, 151, 192, 254, 239,  63,  39, 220, 247, 201,
       88, 222, 145, 191,  39, 101, 234,   3,  29, 246,  51, 191,
       33, 213,  42,  75,  91,  31,  10, 192,  30, 221, 137,  43,
       11, 223, 145,  63, 180, 130,  95,  60, 129, 253, 239,  63,
       39, 220, 247, 201,  88, 222, 145,  63, 142, 209,  24,  55,
       93, 186,  23, 192,   0,   0,   0,   0,   0,   0,   0,   0,
       30, 221, 137,  43,
      ... 92 more items
    ]
  }
]
```

pmconne:
> Do you know of any tests that create an iModel that I could use as an example?

See the tests added in #7785 (you can steal parts of the backend code).

claudiaareneee:
Cool cool, I'll do that. It probably is a schema issue. Running the manual test I got a "unable to enter graphical editing scope: version too old" error. I'm not sure of the actual schema number yet though

claudiaareneee:
Heyyo, my tests work now so I'm publishing this PR

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -63,8 +57,8 @@ export class TextAnnotation2d extends AnnotationElement2d {
    * @param subCategory If specified, the subcategory on which to define the geometry; otherwise, the default subcategory of the element's category is used.
```
claudiaareneee, line 57: TODO: fix these comments, I forgot to remove subcategory.

pmconne:
> Heyyo, my tests work now so I'm publishing this PR

Can you make sure the PR description is up to date? I see a lot of unchecked TODOs.

claudiaareneee:
@pmconne, I found this discussion: https://github.com/iTwin/itwinjs-core/issues/6793#issuecomment-2150832935.

Essentially because of the refactoring of the text annotation geometry, I now need `TextBlockLayout` in more places, and the `extract-api` command was failing. I ended up making some of the classes in this file `@beta` for this reason. 

If you think this API needs to be cleaned up, could I do that in a separate PR? This one is already really large and I don't want to oversaturate it.

core/backend/src/annotations/FrameGeometry.ts
```diff
@@ -0,0 +1,324 @@
...
+  /**
+   * Based on the frame style, this method will construct and append [[GeometryParams]] (for line style) and a [[Loop]] (for the frame shape) to the builder.
+   * @param builder that will be appended to in place
+   * @param frame
+   * @param range to enclose with the frame
+   * @param transform that transform the range to world coordinates
+   * @returns `true` if any geometry was appended to the builder
+  * @beta
+   */
+  export const appendFrameToBuilder = (builder: ElementGeometry.Builder, frame: TextFrameStyleProps, range: Range2d, transform: Transform, geomParams?: GeometryParams): boolean => {
```
pmconne, line 29: @ben-polinsky will documentation generator produce correct docs for functions declared like this instead of e.g. `export function(whatever)`?
@claudiaareneee was this intended to prevent weirdos from doing `FrameGeometry.appendFrameToBuilder = () => throw new Error("mwuhaha");`?
(IMO the namespace is just API clutter and you could remove it).

core/backend/src/annotations/FrameGeometry.ts
```diff
@@ -0,0 +1,324 @@
...
+    if (frame.shape === "none") {
+      return false;
+    }
+
+    const params = geomParams?.clone() ?? new GeometryParams(Id64.invalid);
```
pmconne, line 34: Why `Id64.invalid`? Does anyone actually pass in `undefined` for `geomParams`?

core/backend/src/annotations/FrameGeometry.ts
```diff
@@ -0,0 +1,324 @@
...
+    frame: Exclude<TextAnnotationFrameShape, "none">;
+    /** Range to be enclosed */
+    range: Range2d;
+    /** Transform that translates and rotates the range to world coordinates */
+    transform: Transform;
+  }
+
+  /**
+   * Computes the frame geometry based on the provided frame shape and range.
+   * @returns a [[Loop]] or [[Path]](if it's just a line) that represents the frame geometry
```
pmconne, line 84:
```suggestion
   * @returns a [Loop]($geometry) or [Path]($geometry) (if it's just a line) that represents the frame geometry
```
`[[SomeThing]]` is short-hand for `[SomeThing]($package-containing-this-file)`. It doesn't work for types defined outside of this package.
@ben-polinsky why isn't that producing a documentation build error about `Loop` and `Path` not being defined in core-backend?

core/backend/src/annotations/FrameGeometry.ts
```diff
@@ -0,0 +1,324 @@
...
+      const end = curve instanceof Arc3d ? arcIntervalFactor : lineIntervalFactor;
+      for (let interval = 0; interval <= 1; interval += end) {
+        points.push(curve.fractionToPoint(interval));
+      }
+    });
+    return points;
+  }
+
+  /** Line - currently just adds an underline. Once we have leaders, this method may change. */
+  const computeLine = (range: Range2d, transform: Transform): Path => {
```
pmconne, line 135: If you're going to use a namespace, move all the non-exported stuff outside of it so it's clear at a glance what's being exported vs local implementation detail.

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -35,6 +36,31 @@ export interface TextAnnotationAnchor {
...
+ */
+export type TextAnnotationFillColor = TextStyleColor | "background";
+
+/**
+ * Describes how to draw the frame around a [[TextBlock]].
+ * The frame can be a simple line, a filled shape, or both.
+ * @beta
+ */
+export interface TextFrameStyleProps {
+  shape: TextAnnotationFrameShape;
```
pmconne, line 58: Did you consider defaulting to "rectangle"?

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -115,6 +148,7 @@ export class TextAnnotation {
       orientation: props?.orientation ? YawPitchRollAngles.fromJSON(props.orientation) : undefined,
       textBlock: props?.textBlock ? TextBlock.create(props.textBlock) : undefined,
       anchor: props?.anchor ? { ...props.anchor } : undefined,
+      frame: props?.frame,
```
pmconne, line 151: You should probably clone this to prevent unwanted unintentional modifications via the input object.

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -140,6 +174,8 @@ export class TextAnnotation {
       props.anchor = { ...this.anchor };
     }
 
+    props.frame = this.frame;
```
pmconne, line 177: You should definitely clone this so that the caller can modify the returned JSON object without accidentally mutating the source object.

pmconne (approved):
Comments are mostly nitpickery.

core/backend/src/annotations/FrameGeometry.ts
ben-polinsky: I'm not sure, [I'll have to investigate](https://github.com/iTwin/itwinjs-core/issues/8159).

claudiaareneee: > `[[SomeThing]]` is short-hand for `[SomeThing]($package-containing-this-file)`

Good to know! I did not know that

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -84,7 +85,7 @@ export function layoutTextBlock(args: LayoutTextBlockArgs): TextBlockLayout {
 /**
  * Gets the result of laying out the the contents of a TextBlock into a series of lines containing runs.
  * The visual layout accounts for the [TextStyle]($common)s, fonts, and [TextBlock.width]($common). It applies word-wrapping if needed.
- * The layout returned matches the visual layout of the geometry produced by [[produceTextAnnotationGeometry]].
+ * The layout returned matches the visual layout of the geometry produced by [[TextAnnotationGeometry]].
```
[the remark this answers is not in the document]
claudiaareneee: Would you rather I un-`namespace` everything? I thought that looked cleaner to me, but I'm not attached to it. If you think that it would be better to not wrap things in a `namespace`, I can get rid of them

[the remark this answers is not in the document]
pmconne: Namespaces are another thing that mostly help the writer of the code (they don't have to type the namespace inside the namespace) and not the callers/readers (they have to type the namespace all over the place).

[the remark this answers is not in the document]
claudiaareneee: Okay, no problem. Away they go ⛵️

core/common/src/annotation/TextAnnotation.ts
claudiaareneee: done

core/common/src/annotation/TextAnnotation.ts
claudiaareneee: done

core/backend/src/annotations/TextAnnotationGeometry.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Id64 } from "@itwin/core-bentley";
+import { produceTextBlockGeometry } from "./TextBlockGeometry";
+import { appendFrameToBuilder, computeIntervalPoints } from "./FrameGeometry";
+
+/**
+ * Properties required to compute the geometry of a text annotation.
+ * @beta
+ * @see [[appendTextAnnotationGeometry]] to append the geometry to an [[ElementGeometry.Builder]].
+ */
+export interface RequestProps {
```
pmconne: Give this a less vague name. Since its only purpose is to define the arguments to `appendTextAnnotationGeometry`, I recommend `AppendTextAnnotationGeometryArgs`.

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -35,6 +36,31 @@ export interface TextAnnotationAnchor {
   horizontal: "left" | "center" | "right";
 }
 
+/** Set of predefined shapes that can be computed and drawn around the margins of a [[TextBlock]]
+ * @beta
+*/
+export type TextAnnotationFrameShape = "none" | "line" | "rectangle" | "circle" | "equilateralTriangle" | "diamond" | "square" | "pentagon" | "hexagon" | "octagon" | "capsule" | "roundedRectangle";
```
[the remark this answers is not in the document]
claudiaareneee: Made an attempt at that. The linter doens't seem to like it. 

We don't have any plans to add more. If we did it would not be for a long time. I would think we would only add more and not take away. I don't think the change would be very breaking. 

Though I understand the concerns for future proofing. I could turn this into an enum and then us the `| string` to expand it. Though I liked the cohesion of this type being declared like the anchor ones were. 

@pmconne, what's your vote? enum or leave it as is?

image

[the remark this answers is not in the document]
pmconne: > I would think we would only add more and not take away. I don't think the change would be very breaking.

Unfortunately tsc is strict about exhaustiveness, so adding a new enum member can cause existing code to fail to compile. That's a general problem that we don't have a solution for yet. Since these APIs are currently `@beta` let's leave it as-is for now.

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -35,6 +36,31 @@ export interface TextAnnotationAnchor {
...
+/**
+ * Describes how to draw the frame around a [[TextBlock]].
+ * The frame can be a simple line, a filled shape, or both.
+ * @beta
+ */
+export interface TextFrameStyleProps {
+  shape: TextAnnotationFrameShape;
+  fill?: TextAnnotationFillColor;
+  border?: TextStyleColor;
+  borderWeight?: number;
```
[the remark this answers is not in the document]
claudiaareneee: I considered that, but I found this approach to be similar.  

> Weight is meaningful only if border is defined. Did you consider

It's more like:
Weight is meaningful only if **shape** is defined.

There is a default color (black) so the user can still define borderWeight meaningfully.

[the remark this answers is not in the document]
pmconne: Can you please document all of these fields and their defaults? If I want text with a solid blue rectangle behind it (no border), must I set borderColor and color to the same color (blue)? EDIT: Thanks.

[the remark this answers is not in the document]
claudiaareneee: done! sorry, I was in the process of committing those changes when I sent that last message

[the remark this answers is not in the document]
claudiaareneee: (Also, I haven't decided on the defaulting to rectangle bit, I'm still messing around with that)

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -35,6 +36,35 @@ export interface TextAnnotationAnchor {
...
+ * Describes how to draw the frame around a [[TextBlock]].
+ * The frame can be a simple line, a filled shape, or both.
+ * @beta
+ */
+export interface TextFrameStyleProps {
+  /** Shape of the frame. */
+  shape: TextAnnotationFrameShape;
+  /** The color to fill the shape of the text frame. This fill will is applied using [[FillDisplay.Blanking]]. Default: no fill */
+  fill?: TextAnnotationFillColor;
+  /** The color of the text frame. Default: black */
```
pmconne:
```suggestion
  /** The color of the text frame's outline. Default: black */
```

core/backend/src/annotations/FrameGeometry.ts
claudiaareneee: done - removed namespaces

core/backend/src/annotations/TextAnnotationGeometry.ts
claudiaareneee: My bad, was left over from the namespace. Changed!

core/backend/src/annotations/FrameGeometry.ts
claudiaareneee: I can't remember where I got that from. Should I be using the element's category id?

core/backend/src/annotations/FrameGeometry.ts
claudiaareneee: Since it's broken, I'll change it to export them as functions.

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -290,7 +291,7 @@ function split(source: string): Segment[] {
   return segments;
 }
 
-/** @internal */
+/** @beta */
```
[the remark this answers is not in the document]
claudiaareneee: Okay I just added comments to the tops of each of these classes. Do you need me to go more granularly for each public method in the classes too? Could I do that in a separate PR? 

(I'm planning on adding new runs (TabRun, ListRun, etc) so I'll likely be editing this file)

core/common/src/annotation/TextAnnotation.ts
claudiaareneee: Okay, defaulted to rectangle. I also added a test in common on TextAnnotation.test.ts for it.

core/backend/src/annotations/FrameGeometry.ts
claudiaareneee: Okay I remember where I got that from. The ElementGeometry.builder and GeometryStreamBuilder both use Id64.Invalid for the lines in the fractions bars.

Maybe the frame geom should be a different category though. Thoughts @pmconne?

core/backend/src/annotations/FrameGeometry.ts
claudiaareneee: Done

core/backend/src/annotations/FrameGeometry.ts
pmconne: That's for your app and/or the user to decide. Either way, I think your API here should require the caller to configure all of that - you certainly can't decide here what category to put the frame on. If you want to continue just cloning the input GeometryParams, I think you should make them non-optional so you can eliminate Id64.invalid here.

claudiaareneee: Done! 

I think that should be the last comment to address!

PR #8058, opened 2025-05-08
Checks for proper behavior of xml serializer and deserializer in case of CAs

Abhijeet-Bentley:
1. Issue Link: https://github.com/iTwin/itwinjs-backlog/issues/1439
2. Issue Link: https://github.com/iTwin/itwinjs-backlog/issues/1438

[Note]: This PR solves both issues listed above.

-Added test as requested in the story
-Fix was simple, instead of looping over local properties we now loop over getProperties which, cover both local and inherited properties.
-Refactored the previous "XmlSerializationUtils" namespace to "XmlCustomAttributesUtils" since, it only entails CA serialization.


rschili (approved):
rush change is still missing at the moment

pankhur94:
it would be great to use development tab(issue) in case of pvt gh issues to link issue with pr. Non members don't have access to backlog issues and won't be able to access it using the link in description anyway.

core/ecschema-metadata/src/Deserialization/XmlCustomAttributesUtils.ts
```diff
@@ -14,7 +14,7 @@ import { Schema } from "../Metadata/Schema";
  * Namespace holding utility functions for serializing EC types to the EC XML format.
  * @internal
  */
-export namespace XmlSerializationUtils {
+export namespace XmlCustomAttributesUtils {
```
rschili, line 17: I think here has been some confusion. I remember Stefan suggested you rename this class, but I see a lot of callers to createXmlTypedName() in code which has nothing to do with custom attributes. This method should remain in the old namespace, moving it into XmlCustomAttributeUtils makes it confusing

core/ecschema-metadata/src/test/Metadata/CustomAttributeClass.test.ts
```diff
@@ -315,6 +315,87 @@ describe("CustomAttributeClass", () => {
...
+          },
+        },
+      };
+
+      const ecschema = await Schema.fromJson(schema, new SchemaContext());
+      assert.isDefined(ecschema);
+
+      const serialized = await ecschema.toXml(newDom);
+      assert.isDefined(serialized);
+      const elements = serialized.lastChild;
```
rschili, line 386: Even though this works, I don't like this code for asserting. It's very hard to read, especially since there is no XML snippet, understanding what the lastElement/firstElement refers to is not easy.
Imagine somebody getting a failure in this test because structProps.childNode does not have length 4. Troubleshooting this seems like a nightmare.

I suggest to load the xml back into a schema. Should be simple as it has no dependencies. Then you can use our API to get the custom attribute and check the values.

core/ecschema-metadata/src/test/Metadata/CustomAttributeClass.test.ts
```diff
@@ -315,6 +315,87 @@ describe("CustomAttributeClass", () => {
...
+            {
+              type: "PrimitiveProperty",
+              typeName: "string",
+              name: "hexColor",
+            },
+            ],
+          },
+          RgbColor: {
+            schemaItemType: "StructClass",
+            baseClass: "TestSchema.HexColor",
```
rschili, line 348: Like we discussed, the schema seems fine to me, I just think the example can be better. e.g. "An RGBColor IS a HexColor" seems confusing. Use Animal->Bird or Vehicle->Car or any common example for inheritance. Initially I think I suggested ColorRGB->ColorARGB (where ARGB just adds one alpha property)

core/ecschema-metadata/src/Deserialization/XmlCustomAttributesUtils.ts
StefanApfel-Bentley: Or we export it as single function. I don't see much value of a namespace with just one function in it.

core/ecschema-metadata/src/test/Metadata/CustomAttributeClass.test.ts
StefanApfel-Bentley: Or serialize the schema to xml and compare it with an expected value. That's even easier to understand and easier to spot if something is missing.

core/ecschema-metadata/src/Deserialization/XmlCustomAttributesUtils.ts
Abhijeet-Bentley: Should I just change it back or export a single function

PR #8062, opened 2025-05-08
Fix Imodel Tile Missing Child Tiles After clearLayers()

danielzhong-bentley:
This PR fixes an issue where iModel child tiles fail to reload after `clearLayers()` is called, particularly when parent tiles are cleared but their children are needed for refinement due to screen-space resolution. Previously, `clearLayers()` did not reset `_childrenLoadStatus`, causing the system to believe the children were still loaded and skip calling `loadChildren()`. As a result, when the camera is very close to geometry and the tile system attempts to refine to more detailed tiles, those child tiles were never reloaded—leading to missing graphics.

Unlike `RealityTile`, which explicitly manages child loading during traversal, its `selectRealityTiles()` method always explicitly calls `loadChildren()` during selection. `IModelTile` relies on `_childrenLoadStatus` to trigger loading. If a tile is too coarse and lacks graphics, it defers to children. But if `_childrenLoadStatus` wasn't reset after clearing, the system assumes children are already loaded and skips `loadChildren()`, leading to missing tiles.

More Detail:
https://github.com/orgs/iTwin/projects/37/views/26?sliceBy%5Bvalue%5D=danielzhong-bentley&pane=issue&itemId=107849852&issue=iTwin%7Citwin-graphics-backlog%7C664


core/frontend/src/internal/tile/IModelTile.ts
```diff
@@ -332,4 +332,11 @@ export class IModelTile extends Tile {
 
     return this.isParentDisplayable ? SelectParent.Yes : SelectParent.No;
   }
+
+  public override clearLayers(): void {
+    super.clearLayers();
+    if (this.isDisplayable) {
+      this._childrenLoadStatus = TileTreeLoadStatus.NotLoaded;
```
pmconne, line 339: Should it actually unload the children?

danielzhong-bentley: It's already disposed the children in `super.clearLayers()`.

danielzhong-bentley: Fixed, thanks for the suggestion! I updated iModelTile to call disposeChildren(), since RealityTiles don’t seem affected.

markschlosseratbentley:
> If a tile is too coarse and lacks graphics, it defers to children. But if `_childrenLoadStatus` wasn't reset after clearing, the system assumes children are already loaded and skips `loadChildren()`, leading to missing tiles.

This makes a lot of sense, thanks for researching.

PR #8077, opened 2025-05-12
Handle leak fixes

tcobbs-bentley:
* Call `unref` on long-running CloudSqlite token refresh timeout.
* Remove registered IpcAppNotifyHandler in `IpcApp.shutdown`.
Hopefully this will fix the handle leaks that are occurring on the Mac CI pipeline.


ben-polinsky (approved):
Can no longer reproduce locally

PR #8081, opened 2025-05-13
Time-based deprecation policy

paulius-valiunas:
- Changed the deprecation policy in documentation
- Unified the format of existing deprecation comments

Further changes will be required:
- In https://github.com/iTwin/eslint-plugin/ - to accommodate the new comment format and to add the ability to add/remove dates from deprecation comments automatically.
- In this repo - to consume the updated eslint rule and update deprecation comments automatically during the package publishing process. This will insert the correct date exactly one year from the current release before publishing the output to npm. It will also remove dates that are already in the past and replace them with a more generic note that the API might be removed in next major version.


dassaf4 (approved):

saeeedtorabi (approved):

grigasp (changes requested):
The deprecation message changes under `presentation` dir look fine. But can you please revert formatting changes?

paulius-valiunas:
> The deprecation message changes under `presentation` dir look fine. But can you please revert formatting changes?

I just ran the `prettier:fix` script, it should at least be consistent with the prettier config now

ben-polinsky (approved):

paulius-valiunas:
here's the associated update to eslint-plugin: https://github.com/iTwin/eslint-plugin/pull/102

pmconne (approved):

christophermlawson (approved):

aruniverse:
> @kckst8 @jason-crow @jjbeckman13 please take a look

@kckst8 @jason-crow @jjbeckman13 @simnorm

docs/learning/api-support-policies.md
```diff
@@ -69,7 +69,93 @@ Rare exceptions may be made to this policy when a breaking API change is require
...
 
-Before a "public" API can be removed it must first be marked as "deprecated" in version `N` of the package. It must remain stable for the duration of version `N+1`. Finally, in version `N+2` it _may_ be removed. This means that, for example, an API deprecated in version 4.1 of a package cannot be removed until version 6.0 at the earliest.
+Before a "public" API can be removed it must first be marked as "deprecated", with a clear upgrade path described in the deprecation note and the earliest date when it might be removed. A grace period of one year is given for consumers to update their code. After the grace period, deprecated APIs may be removed in any following major release. We recommend using [typescript-eslint](https://typescript-eslint.io/) with the [no-deprecated rule](https://typescript-eslint.io/rules/no-deprecated/) enabled so that you get timely warnings when you use deprecated APIs.
+
+The same policy applies to "preview" APIs, with a 3 month grace period instead of a year.
```
jjbeckman13, line 74: The way I understand it is that a Preview API could be added and removed within the same major version. Can someone confirm?

Also, I am assuming here we won't remove a Preview API in a patch?

paulius-valiunas: I always thought that "preview" was supposed to be more stable than "beta". So while beta can be removed anytime, preview still has to wait until a major version.

paulius-valiunas: @aruniverse if beta APIs can be removed anytime (I was under the impression that they also had to wait for major release when writing this), do we need soft-deprecation (and deprecation comments) for preview APIs at all?

jjbeckman13 (approved):

docs/learning/api-support-policies.md
```diff
@@ -69,7 +69,93 @@ Rare exceptions may be made to this policy when a breaking API change is require
...
 
-Before a "public" API can be removed it must first be marked as "deprecated" in version `N` of the package. It must remain stable for the duration of version `N+1`. Finally, in version `N+2` it _may_ be removed. This means that, for example, an API deprecated in version 4.1 of a package cannot be removed until version 6.0 at the earliest.
+Before a "public" API can be removed it must first be marked as "deprecated", with a clear upgrade path described in the deprecation note and the earliest date when it might be removed. A grace period of one year is given for consumers to update their code. After the grace period, deprecated APIs may be removed in any following major release. We recommend using [typescript-eslint](https://typescript-eslint.io/) with the [no-deprecated rule](https://typescript-eslint.io/rules/no-deprecated/) enabled so that you get timely warnings when you use deprecated APIs.
```
raplemie, line 72: Could there be a mention that upgrade paths for a public API, unless simply removed, should be replaced by something that is public ? Or is this not part of the goal ? (This happened in the past that the replacement was beta, so we would hold on a deprecated API because moving to something that could be removed anytime was even more worrying.)

aruniverse: This is a fair point, and I think we need to better document the API promotion process from `@beta` to `@public @preview`. Will handle in a separate PR/Issue

PR #8082, opened 2025-05-13
Replace PropertyDescription quantityType with kindOfQuantityName

### LINKED ISSUE #8056: Refactor FormattedQuantityDescription sub-classes in core-frontend
As a result of #7807, [FormattedQuantityDescription](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/properties/FormattedQuantityDescription.ts) will want to be refactored, centered around usage of `QuantityType`.

Similarly, extended subclasses of `FormattedQuantityDescription` will have to be updated as well, found in [LengthDescription.ts](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/properties/LengthDescription.ts) and [AngleDescription.ts](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/properties/AngleDescription.ts).

On ideas of how to refactor, I'm forwarding a comment I made in #8046: 

> Looking at the code, quantityType() should be modified to not return a fixed value. We can rework the constructor to pass in an optional quantityType, with the default value being one of the KoQs part of [this table](https://github.com/iTwin/itwinjs-core/pull/8046/files#diff-66edb1fd7a647c360573e6f0a2c9538d674e7373c59562df7afa35636eec2f4cR30-R40). The [doccomment](https://github.com/iTwin/itwinjs-core/blob/master/ui/appui-abstract/src/appui-abstract/properties/Description.ts?rgh-link-date=2025-05-07T15%3A37%3A15Z#L86-L90) for the quantityType property is encouraging, seems like it's already planned to possibly be a EC full name for a KindOfQuantity. If we're worried about bloating the number of args passed into the constructor, we'll just define an interface for BaseQuantityDescriptionProps

> When creating the Parser/FormatterSpecs we can rely on IModelApp.formatsProvider, that quantityType, a persistenceUnitProps that is retrieved via IModelApp.quantityFormatter.unitsProvider (as for the persistenceUnit in question, each sub class Description might need to have a default like Units.M).

> We eventually get rid of the formatterQuantityType getter, as that'd be a duplicate of the quantityType getter. But for now we can keep that, to use as a fallback to generate a parser/formatterSpec for the Description subclasses in case IModelApp.formatsProvider is undefined.

> Overall it doesn't seem like there's anything in these classes that worry me, one thing might be that the formatterSpec/parserSpec getters are how the specs are set by the classes, and those aren't async...

hl662:
Closes #8056

In ecschema-metadata:
- Update `SchemaFormatsProvider` to handle errors when calling getSchema/getSchemaItem through SchemaContext.

In appui-abstract:
- Add a new optional `kindOfQuantityName` argument for `BaseQuantityDescription`. Make it accessible to get/set on runtime
- Deprecate `quantityType` in favor of `kindOfQuantityName` in `PropertyDescription`

In core-frontend:
- Deprecate `quantityType` getters for subclasses of `FormattedQuantityDescription`, keep the `formatterQuantityType()` for now and deprecate once `QuantityType` interface is deprecated
- Update constructor for `FormattedQuantityDescription` to take in a `kindOfQuantityName`, also add a new constructor overload
- Update SetupCameraTool to retrieve a formatterSpec from `QuantityFormatter` instead of the `LengthDescription`'s `.format()` 
- Add async `addFormattingSpecsToRegistry` and non async `getSpecsByName` to `QuantityFormatter`. On init(), quantityFormatter now listens to formatsChanged events `IModelApp.formatsProvider`. If a format was added to the QuantityFormatter through `addFormattingSpecsToRegistry`, the specs will be updated every time formatsProvider changes that format.

There are 2 ways to retrieve formats through the new `kindOfQuantityName`:
1. For async eligible workflows, you can query `await IModelApp.formatsProvider.getFormat(koqName)` to get a FormatProps object, and construct the formatting specs ([new helper methods](https://github.com/iTwin/itwinjs-core/blob/815d5ae7f2eda50149769ace9c524b96fdd9999b/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L1087-L1109)). There is this concern about having to pass in a valid persistenceUnit name though, but a koqName can be used to lookup the KoQ schemaItem and retrieve that. But if this workflow is too expensive, and you're concerned about time it takes to get things on demand, you can go for option 2.
2. For sync only workflows, use [getSpecsByName](https://github.com/iTwin/itwinjs-core/blob/815d5ae7f2eda50149769ace9c524b96fdd9999b/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L1111-L1117). The quantityFormatter acts as a cache for the specs and listens to formatsChanged events. Applications or tools can call [addFormattingSpecsToRegistry](https://github.com/iTwin/itwinjs-core/blob/815d5ae7f2eda50149769ace9c524b96fdd9999b/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L1126-L1141) on startup()/ahead of time.


core/frontend/src/tools/ViewTool.ts
```diff
@@ -4043,15 +4044,36 @@ export class SetupCameraTool extends PrimitiveTool {
...
+  protected _lengthFormatterSpec?: FormatterSpec;
+  /** @internal */
+  protected _removeFormatterListener?: () => void;
 
   public override isCompatibleViewport(vp: Viewport | undefined, isSelectedViewChange: boolean): boolean { return (super.isCompatibleViewport(vp, isSelectedViewChange) && undefined !== vp && vp.view.allow3dManipulations()); }
   public override isValidLocation(_ev: BeButtonEvent, _isButtonEvent: boolean): boolean { return true; }
   public override requireWriteableTarget(): boolean { return false; }
   public override async onPostInstall() {
     await super.onPostInstall();
+    this._lengthFormatterSpec = await this.getFormatterSpecByKoQAndPersistenceUnit("AecUnits.LENGTH", "Units.M");
```
bbastings, line 4057: This is a pretty significant change to require of every single tool that is currently using a FormattedQuantityDescription. Could we move this "cache" some place common so that every single tool isn't required to do this? There are a lot of tool in other packages outside core that would need to change.

core/frontend/src/tools/ViewTool.ts
```diff
@@ -4251,7 +4285,7 @@ export class SetupCameraTool extends PrimitiveTool {
   private _cameraHeightProperty: DialogProperty<number> | undefined;
   public get cameraHeightProperty() {
     if (!this._cameraHeightProperty)
-      this._cameraHeightProperty = new DialogProperty<number>(new LengthDescription("cameraHeight", ViewTool.translate("SetupCamera.Labels.CameraHeight")),
+      this._cameraHeightProperty = new DialogProperty<number>(new LengthDescription("cameraHeight", ViewTool.translate("SetupCamera.Labels.CameraHeight"), undefined, "AecUnits.LENGTH"),
```
bbastings, line 4288: I'm fine with this level of change (assuming this is sufficient for appui is still able to format and parse values).

Isn't "AecUnits.LENGTH" the default for LengthDescription though, did the caller need to be changed?

export class LengthDescription extends FormattedQuantityDescription {
  constructor(name?: string, displayLabel?: string, iconSpec?: string, kindOfQuantityName?: string) {
    const defaultName = "length";
    super({
      name: name ? name : defaultName,
      displayLabel: displayLabel ? displayLabel : IModelApp.localization.getLocalizedString("iModelJs:Properties.Length"),
      kindOfQuantityName: kindOfQuantityName ? kindOfQuantityName : "AecUnits.LENGTH",
      iconSpec,
    });
  }

core/frontend/src/tools/ViewTool.ts
hl662: We could repurpose the `QuantityFormatter`. It currently acts as a cache of formatter and parserSpecs for `QuantityType`. If we have APIs that are unable to use async based workflows to get formats from the `formatsProvider`, maybe they can call something like `IModelApp.quantityFormatter.getSpecsByKindOfQuantity()`, that is synchronous.

I'll then expose an `async` method in QuantityFormatter that populates this new cache with new Format/ParserSpecs, taking in a koqName, and a persistenceUnitName. Under the hood this method will ask the IModelApp.formatsProvider() and unitsProvider() to check for the KoQ and lookup the persistenceUnit. If it can't find any, the specs won't be populated into the cache. Perhaps the method can return true or false depending on if the cache was populated or not.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -1051,8 +1107,41 @@ export class QuantityFormatter implements UnitsProvider {
...
+   * @beta
+   * @param name The key used to identify the formatter and parser spec
+   * @param persistenceUnitName The name of the persistence unit
+   * @param formatProps If not supplied, tries to retrieve the [[FormatProps]] from [[IModelApp.formatsProvider]]
+   */
+  public async addFormattingSpecsToRegistry(name: string, persistenceUnitName: string, formatProps?: FormatProps): Promise<void> {
+    if (!formatProps) {
+      formatProps = await IModelApp.formatsProvider.getFormat(name);
+    }
+    if (formatProps) {
```
grigasp, line 1130: do we want it to silently fail? maybe it makes more sense to throw an error when we fail to find format props?

core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -586,6 +608,40 @@ export class QuantityFormatter implements UnitsProvider {
...
+          } else {
+            this._formatSpecsRegistry.delete(name); // clear the specs if format is removed
+          }
+        });
+      } else {
+        for (const name of args.formatsChanged) {
+          if (this._formatSpecsRegistry.has(name)) {
+            const formatProps = await IModelApp.formatsProvider.getFormat(name);
+            if (formatProps) {
+              const persistenceUnitName = this._formatSpecsRegistry.get(name)!.formatterSpec.persistenceUnit.name;
```
grigasp, line 635: Not sure you should use `!` here, since there's no guarantee you're going to find an entry there. For example, if the event is fired with `["x", "x"]`, and `getFormat("x")` returns no format props, the first iteration will remove `x` from `_formatSpecsRegistry` and the second one will throw an error due to use of `undefined`

core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -428,6 +440,16 @@ export class QuantityFormatter implements UnitsProvider {
     }
   }
 
+  /**
+   * @internal
+   */
+  public shutdown(): void {
```
grigasp, line 446: I find it a bit confusing, that we have a public class, whose initialization and shutdown methods are marked `@internal`, making it somewhat useless. In any case, I think the common practice is to name methods like this `[Symbol.dispose]`.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -1051,8 +1107,41 @@ export class QuantityFormatter implements UnitsProvider {
     const format = await Format.createFromJSON(formatName ?? "temp", this._unitsProvider, formatProps);
     return ParserSpec.create(format, this._unitsProvider, persistenceUnitProps);
   }
-}
 
+  /**
+   * @beta
+   * Returns a [[FormattingSpecEntry]] for a given name, typically a KindOfQuantity full name.
+   */
+  public getSpecsByName(name: string): FormattingSpecEntry | undefined {
```
grigasp, line 1115: public method returns internal type

ui/appui-abstract/src/appui-abstract/quantity/BaseQuantityDescription.ts
```diff
@@ -20,10 +20,13 @@ export abstract class BaseQuantityDescription implements PropertyDescription {
   public displayLabel: string;
   public typename: string;
   public editor: PropertyEditorInfo;
+  public kindOfQuantityName: string;
 
-  constructor(name: string, displayLabel: string, iconSpec?: string) {
+
+  constructor(name: string, displayLabel: string, kindOfQuantityName: string, iconSpec?: string) {
```
grigasp, line 26: This breaks the API - the third argument was `iconSpec`, and now it's `kindOfQuantityName`. And I would say it breaks it in a very bad way, because the types of the two arguments are the same, so if anyone was passing a value for `iconSpec` - there will be no build error, but the value will now be set on `kindOfQuantityName` rather than what it was previously set on.

Same with changes in `FormattedQuantityDescription`.

core/frontend/src/tools/ViewTool.ts
```diff
@@ -4279,13 +4279,15 @@ export class SetupCameraTool extends PrimitiveTool {
   public set targetHeight(value: number) { this.targetHeightProperty.value = value; }
 
   private syncCameraHeightState(): void {
-    this.cameraHeightProperty.displayValue = (this.cameraHeightProperty.description as LengthDescription).format(this.cameraHeight);
+    const specs = IModelApp.quantityFormatter.getSpecsByName(this.cameraHeightProperty.description.kindOfQuantityName ?? "AecUnits.LENGTH")
```
grigasp, line 4282: As commented in our private discussion, I don't see why tools need to do property formatting, when they already set all the necessary information to do that on UI layer, displaying the values.

ui/appui-abstract/src/appui-abstract/properties/Description.ts
```diff
@@ -83,11 +83,10 @@ export interface PropertyDescription {
...
   /** Information for a property converter */
   converter?: PropertyConverterInfo;
-  /** Quantity type key used to look up formatting and parsing specs. This is typically either the name of a quantity type used by a tool
-   *  or the full name of a KOQ (schema:koq).
-   * @alpha
+  /** KindOfQuantity full name used to look up [[FormatProps]]
+   * @beta
    */
-  quantityType?: string;
+  kindOfQuantityName?: string;
```
GerardasB, line 89: This `@alpha` API was added many years ago, would it be safer to deprecate instead?

core/frontend/src/tools/ViewTool.ts
hl662: I looked across the repo to see if this was done in any other tools that uses `PropertyDescription` and this was the only tool to do so. This tool has existed for many years now so it might have not been updated to **not** do formatting on its own. We can revisit this tool in the future to drop this, I figured this change should be enough for this PR

bbastings: Yeah, this is just old code. Typically if you wanted to actually change the value you just set the new one and sync.

private syncTargetHeightState(): void {
  this.targetHeight = /* SOME NEW VALUE */;
  this.targetHeightProperty.isDisabled = !this.useTargetHeight;
  this.syncToolSettingsProperties([this.targetHeightProperty.syncItem]);
}

Since this code isn't actually changing the value (this.targetHeight is current value) these method are really only responsible to update the disabled state based on the associated toggle and could be changed to this:

  private syncCameraHeightState(): void {
    this.cameraHeightProperty.isDisabled = !this.useCameraHeight;
    this.syncToolSettingsProperties([this.cameraHeightProperty.syncItem]);
  }

  private syncTargetHeightState(): void {
    this.targetHeightProperty.isDisabled = !this.useTargetHeight;
    this.syncToolSettingsProperties([this.targetHeightProperty.syncItem]);
  }

This is completely unnecessary too:

  public override async applyToolSettingPropertyChange(updatedValue: DialogPropertySyncItem): Promise<boolean> {
    if (updatedValue.propertyName === this.useCameraHeightProperty.name) {
-      this.useCameraHeight = updatedValue.value.value as boolean; // <- Value was already updated, which is why this method was called...
      IModelApp.toolAdmin.toolSettingsState.saveToolSettingProperty(this.toolId, this.useCameraHeightProperty.item);
      this.syncCameraHeightState();
    } else if (updatedValue.propertyName === this.useTargetHeightProperty.name) {
...

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: I've updated the interface to be beta

ui/appui-abstract/src/appui-abstract/properties/Description.ts
hl662: Makes sense - I've kept quantityType back on, but marked it as deprecated.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: Thanks, I've updated to use [Symbol.dispose]

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: I've refactored the code to handle this. .forEach also doesn't work with async functions either so had to use for .. of

core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -608,32 +607,35 @@ export class QuantityFormatter implements UnitsProvider {
...
       } else {
         for (const name of args.formatsChanged) {
           if (this._formatSpecsRegistry.has(name)) {
             const formatProps = await IModelApp.formatsProvider.getFormat(name);
             if (formatProps) {
-              const persistenceUnitName = this._formatSpecsRegistry.get(name)!.formatterSpec.persistenceUnit.name;
-              await this.addFormattingSpecsToRegistry(name, persistenceUnitName, formatProps);
+              const existingEntry = this._formatSpecsRegistry.get(name);
+              if (existingEntry) {
+                const persistenceUnitName = this._formatSpecsRegistry.get(name)!.formatterSpec.persistenceUnit.name;
```
grigasp, line 636:
```suggestion
                const persistenceUnitName = existingEntry.formatterSpec.persistenceUnit.name;
```

common/changes/@itwin/ecschema-metadata/nam-quantity-description-change_2025-05-14-14-20.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-metadata",
+      "comment": "Gracefully handle errors, including rpc, when retrieving schema items in SchemaFormatsProvider",
```
grigasp, line 5:
```suggestion
      "comment": "Gracefully handle errors, including rpc, when retrieving schema items in `SchemaFormatsProvider`",
```

grigasp (approved):

bbastings (approved):

hl662:

ui/appui-abstract/src/appui-abstract/properties/Description.ts
```diff
@@ -86,8 +86,13 @@ export interface PropertyDescription {
   /** Quantity type key used to look up formatting and parsing specs. This is typically either the name of a quantity type used by a tool
    *  or the full name of a KOQ (schema:koq).
    * @alpha
+   * @deprecated in 5.0. Use `kindOfQuantityName` instead.
    */
   quantityType?: string;
```
GerardasB, line 91: Should we deprecate `quantityType` getters of `BaseQuantityDescription` and subclasses: `FormattedQuantityDescription`, `AngleDescription`, `EngineeringLengthDescription`, `LengthDescription`, `SurveyLengthDescription`? Or all of these classes will be deprecated later on instead?

ui/appui-abstract/src/appui-abstract/quantity/BaseQuantityDescription.ts
```diff
@@ -20,10 +20,13 @@ export abstract class BaseQuantityDescription implements PropertyDescription {
...
   public typename: string;
   public editor: PropertyEditorInfo;
+  public kindOfQuantityName?: string;
 
-  constructor(name: string, displayLabel: string, iconSpec?: string) {
+
+  constructor(name: string, displayLabel: string, iconSpec?: string, kindOfQuantityName?: string) {
     this.name = name;
     this.displayLabel = displayLabel;
+    this.kindOfQuantityName = kindOfQuantityName;
```
GerardasB, line 29: Should subclasses, like `AngleDescription` set `kindOfQuantityName`?

ui/appui-abstract/src/appui-abstract/quantity/BaseQuantityDescription.ts
```diff
@@ -20,10 +20,13 @@ export abstract class BaseQuantityDescription implements PropertyDescription {
   public displayLabel: string;
   public typename: string;
   public editor: PropertyEditorInfo;
+  public kindOfQuantityName?: string;
 
-  constructor(name: string, displayLabel: string, iconSpec?: string) {
+
+  constructor(name: string, displayLabel: string, iconSpec?: string, kindOfQuantityName?: string) {
```
GerardasB, line 26: Do we intend to deprecate `BaseQuantityDescription`? If so - asking consumers to provide a `kindOfQuantityName` arg will lead to one more change in the near future. I.w. would a helper to create `PropertyDescription` work better?

GerardasB (approved):

ui/appui-abstract/src/appui-abstract/properties/Description.ts
hl662: Later, we'll have to start by deprecating `QuantityType` in core-frontend, and we've [documented](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/frontend/QuantityFormatting.md#migrating-from-quantitytype-to-kindofquantity) that plan to be done during 5.x lifecycle. We'll promote the new way of retrieving formats through the upcoming first party application, to encourage other products to do the same

ui/appui-abstract/src/appui-abstract/quantity/BaseQuantityDescription.ts
hl662: This warrants more discussion later on, I believe it'll impact many of the tools that are using this class. This class feels like the most basic implementation of the `PropertyDescription` interface, I'm not sure a helper for `PropertyDescription` is worth it when it's just an interface. A helper to create an instance of `BaseQuantityDescription` might help? I would leave this to you and the presentation side.

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -59,7 +60,13 @@ export class SchemaFormatsProvider implements FormatsProvider {
   }
 
   private async getKindOfQuantityFormatFromSchema(itemKey: SchemaItemKey): Promise<SchemaItemFormatProps | undefined> {
-    const kindOfQuantity = await this._context.getSchemaItem(itemKey, KindOfQuantity);
+    let kindOfQuantity: KindOfQuantity | undefined;
+    try {
+      kindOfQuantity = await this._context.getSchemaItem(itemKey, KindOfQuantity);
```
christophermlawson, line 65: getSchemaItem will return undefined. What errors are we catching here?

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -110,7 +117,13 @@ export class SchemaFormatsProvider implements FormatsProvider {
   public async getFormat(name: string): Promise<SchemaItemFormatProps | undefined> {
     const [schemaName, schemaItemName] = SchemaItem.parseFullName(name);
     const schemaKey = new SchemaKey(schemaName);
-    const schema = await this._context.getSchema(schemaKey);
+    let schema: Schema | undefined;
+    try {
+      schema = await this._context.getSchema(schemaKey);
```
christophermlawson, line 122: Likewise, getSchema will return undefined. What errors are we catching here?

core/ecschema-metadata/src/SchemaFormatsProvider.ts
christophermlawson, line 1: Online 133 there is another getSchemaItem call. Should this be wrapped in a try/catch as well?

hl662: thanks, I missed this, just added

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: When the formatsProvider uses a SchemaContext that has an ECSchemaRpcLocater registered, the rpc requests for getting a schema/schemaItem will throw if the item isn't found. A new workflow for retrieving a schemaContext is through the [IModelConnection](https://github.com/iTwin/itwinjs-core/blob/675bd9e1e323f1281ba8133722b2e7fd83be9f4f/core/frontend/src/IModelConnection.ts#L630-L636), so this error will popup and be thrown through `RpcRequest`

core/ecschema-metadata/src/SchemaFormatsProvider.ts
hl662: same explanation as in the reply above

christophermlawson (approved):

PR #8083, opened 2025-05-13
add webpack magic comment to suppress csp

aruniverse:
for bundlers based like webpack and rspack


aruniverse:

PR #8087, opened 2025-05-14
Added test for the pragma validate_ecsql_inserts

RohitPtnkr1996:
imodel-native: https://github.com/iTwin/imodel-native/pull/1110


core/backend/src/test/ecdb/ECSqlStatement.test.ts
```diff
@@ -3338,4 +3338,108 @@ describe("ECSqlStatement", () => {
...
+    assert.isTrue(await reader.step());
+    const parentClassId: Id64String = reader.current.ECInstanceId;
+    assert.isTrue(Id64.isValidId64(parentClassId));
+
+    // When the ECSql insert validation is set to true, the invalid relClassId is detected and an error is thrown.
+    reader = ecdb.createQueryReader("PRAGMA validate_ecsql_inserts=true");
+    assert.isTrue(await reader.step());
+
+    try {
+      ecdb.withCachedWriteStatement(`INSERT INTO test.Child(Parent.Id, Parent.RelECClassId) VALUES(1, ${parentClassId})`, () => {});
```
rschili, line 3374: Just pasting this as a comment from teams discussion
I think it would be good to have another test that uses insertElement and updateElement instead of a low level sql statement. Not sure if you can force an invalid class id in those though..
To make sure those methods fail gracefully if bind fails, as expected.

aruniverse:
@khanaffan @rschili plz review

PR #8090, opened 2025-05-14
Add Viewport.backgroundMapTileTreeReference

pmconne:
`backgroundMap` is @internal and returns `MapTileTreeReference` which is @internal.

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -12909,6 +12909,7 @@ export abstract class Viewport implements Disposable, TileUser {
     get backgroundMapGeometry(): BackgroundMapGeometry | undefined;
     get backgroundMapSettings(): BackgroundMapSettings;
     set backgroundMapSettings(settings: BackgroundMapSettings);
+    get backgroundMapTileTreeReference(): TileTreeReference | undefined;
     changeBackgroundMapProps(props: BackgroundMapProps): void;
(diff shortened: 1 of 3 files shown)


a-gagnon (approved):

ben-polinsky (approved):

aruniverse:

a-gagnon:
@pmconne is this something that can be backported to latest 4.x branch? We're trying to support both itwinjs 4 and 5 if possible.

markschlosseratbentley (approved):

raplemie:
> @pmconne is this something that can be backported to latest 4.x branch? We're trying to support both itwinjs 4 and 5 if possible.
> 
> The alternative is that in our code we have something like `vp.backgroundMapTileTreeReference ?? vp.backgroundMap` that would support both until we entirely drop support for 4.x. Your call.

Even if this is backported to 4.11, our peerDependencies states we support ^4.8.5, so we would need this hack to support earlier versions anyway. Bumping our peerDependencies would be a breaking change, and would require us to deliver a new Major, so while not perfect, I think it would be a viable solution.

pmconne:
Gonna have to let @aruniverse opine on backporting to 4.x.x.

aruniverse:
Spoke to @raplemie , they do not need the backport

PR #8095, opened 2025-05-14
Utilize `PolyfaceData.edgeMateIndex` to speed up some `Polyface` methods

dassaf4:
This PR will need backporting to 5.0.

Improvements to efficiency and usability of `Polyface` API for Booster:
* Export `IndexedEdgeMatcher`, `SortableEdge`, `SortableEdgeCluster`.
* Speed up the following methods if the `PolyfaceData.edgeMateIndex` array (topo cache) is detected. Performance boost depends on the mesh but can be quite substantial. Some input combinations do not admit speedup, e.g., if the visitor is a subset visitor.
  * `PolyfaceQuery.isPolyfaceManifold`
  * `PolyfaceQuery.announceBoundaryEdges`
  * `PolyfaceQuery.boundaryOfVisibleSubset`
* Expand input facet type union to include `PolyfaceVisitor` for many methods.
* `PolyfaceQuery.sumVolumeBetweenFacetsAndPlane` - break out loop contents into separate method `PolygonOps.volumeBetweenPolygonAndPlane` so that callers can implement their own parallelization algorithms.  Add input argument `skipMoments` to speed up calls when only volume is required.
* `SweepLineStringToFacetsOptions` - add optional searcher range tree object to greatly speed up linestring drape for the most general draper `PolyfaceQuery.sweepLineStringToFacets`.

Improvements for iTwin Editing:
* `ConsolidateAdjacentCurvePrimitivesOptions.disableLinearCompression` allows for chain consolidation without linestring compression.
* `CurveCollection.findParentOfDescendant` to locate intermediate primitives in a collection hierarchy.

Also:
* API changes (with justifications):
  * `IndexedPolyfaceWalker.clone` - remove confusing `edgeIndex` arg, add `result` arg, remove unnecessary `undefined` return type. This method is new in 5.0, so change should be fine.
  * `ExportGraphicsMeshVisitor.getPolyface`: add `undefined` to return type since it always returns `undefined`.
  * `PolyfaceQuery.boundaryEdges`, `PolyfaceQuery.sumFacetAreas`: remove unnecessary `undefined` from the input mesh type union. This just caused the method to immediately return without doing anything.

* Other API changes:
  * Remove unnecessary `undefined` from return type: `IndexedPolyfaceWalker.polyface`.
  * Expand input arg type union: `AnnounceDrapePanel`, the `isAlmostEqual` structured array comparators in PointHelpers.ts, the sweep linestring methods of `PolyfaceQuery`, 
  * Add optional `Transform` and/or `result` args to `IndexedXYZCollection.getRange` and overrides.
  * Add optional `tolerance` arg: `Point4d.isAlmostEqual, Point4d.isAlmostEqualXYZW, Point4dArray.isAlmostEqual`.
  * Remove `any` return for method that doesn't return anything: `PolyfaceQuery.announceSweepLinestringToConvexPolyfaceXY`.
  * `PolyfaceQuery.sumVolumeBetweenFacetsAndPlane` expand input plane type to `Plane3d`.
  * Visitor constructor `numWrap` argument is now optional, defaulting to zero.

* New API:
  * `PolyfaceVisitor.createSubsetVisitor` - optional interface method to specify the creation of a visitor over a subset of a visitor's facets. Implement for `IndexedPolyfaceVisitor` and `ExportGraphicsMeshVisitor`. Add class `ExportGraphicsMeshSubsetVisitor` and new method `PolyfaceQuery.isSubsetVisitor`.
  * `Plane3d.projectXYZToPlane` - new method in abstract class, new overrides in subclasses.
  * `GrowableXYZArray.mapPoint` - mutate the instance using a callback to modify each point.
  * `IndexedPolyface.hasEdgeMateIndex` - helper function when utilizing the topo cache.
  * `Transform.createRigidFromOriginAndVector` - helper method to create a localToWorld transform from a single vector. Useful for generating the worldToLocal transformation for rotating planar geometry so that z-coords can be ignored.

* Minor refactoring:
  * `IndexedMeshWalker.buildEdgeMateIndices` (topo cache constructor) now sets a null edge's edgeMate as its own edgeIndex, not as `undefined`. This allows mesh traversal past an edge of zero length, such as naturally occur at each pole of a uv-faceted sphere. `IndexedMeshWalker.isNull` returns whether the walker is at such an edge.
  * `PolyfaceVisitor` implementations no longer call `reset` in the constructor, as this is superfluous. Many callers already invoke `reset` before iterating facets, while those that don't, call `moveToNextFacet` or `moveToReadIndex`, which do the same job.
  * `SortableEdge` no longer subclasses Float64Array; now it just stores the vertex indices and another number.

* Tests for new functionality.
* Many doc improvements.


core/backend/src/ExportGraphics.ts
```diff
@@ -464,12 +464,93 @@ export class ExportGraphicsMeshVisitor extends PolyfaceData implements PolyfaceV
...
+  public override moveToNextFacet(): boolean {
+    if (this._nextSubsetIndex !== this._currentSubsetIndex)
+      return this.moveToReadIndex(this._nextSubsetIndex);
+    this._nextSubsetIndex++;
+    return true;
+  }
+  /** Call this before iterating facets with [[moveToNextFacet]]. */
+  public override reset(): void {
+    this.moveToReadIndex(0);
+    this._nextSubsetIndex = 0; // so immediate moveToNextFacet stays here.
```
saeeedtorabi, line 540: For one line in code comments I dont put `.` at the end
```suggestion
    this._nextSubsetIndex = 0; // so immediate moveToNextFacet stays here
```

core/backend/src/test/standalone/ExportGraphics.test.ts
```diff
@@ -1176,4 +1179,56 @@ describe("exportGraphics", () => {
     assert.strictEqual(infos[0].mesh.indices.length, 6);
   });
 
+  it("verifies subset visitor", () => {
```
saeeedtorabi, line 1182: How about `VerifySubsetVisitor` to keep test name consistency

core/geometry/src/polyface/IndexedPolyfaceVisitor.ts
```diff
@@ -202,49 +201,53 @@ export class IndexedPolyfaceVisitor extends PolyfaceData implements PolyfaceVisi
...
  * * The subset is defined by an array of facet indices provided when this visitor is created.
  * * Input indices (e.g., for `moveToReadIndex`) are understood to be indices into the subset array.
  * @public
  */
 export class IndexedPolyfaceSubsetVisitor extends IndexedPolyfaceVisitor {
-  private _parentFacetIndices?: number[]; // only undefined during super constructor!
-  private _currentActiveIndex: number;    // index within _parentFacetIndices, or -1 after construction
-  private _nextActiveIndex: number;       // index within _parentFacetIndices
+  private _facetIndices: number[];
+  private _currentSubsetIndex: number;    // index within _facetIndices, or -1 after construction
```
saeeedtorabi, line 218: I think the space was there to align with previous comment but now you removed the previous comment so you can remove space as well

```suggestion
  private _currentSubsetIndex: number; // index within _facetIndices, or -1 after construction
```

core/geometry/src/polyface/IndexedPolyfaceVisitor.ts
```diff
@@ -202,49 +201,53 @@ export class IndexedPolyfaceVisitor extends PolyfaceData implements PolyfaceVisi
...
+  private _nextSubsetIndex: number;       // index within _facetIndices
```
saeeedtorabi, line 219:
```suggestion
  private _nextSubsetIndex: number; // index within _facetIndices
```

saeeedtorabi (approved):

core/geometry/src/test/curve/CurveCollection.test.ts
```diff
@@ -346,6 +397,40 @@ describe("ConsolidateAdjacentPrimitives", () => {
     ck.testExactNumber(1, singlePointPathB.children.length, "Single point path consolidates to stub");
     expect(ck.getNumErrors()).toBe(0);
   });
+
+  it("CompressionOption", () => {
```
saeeedtorabi, line 401: nice test!

core/geometry/src/test/polyface/PolyfaceQuery.test.ts
```diff
@@ -1415,3 +1425,265 @@ describe("FacetShape", () => {
...
+        ck.testLE(elapsedDrapeWithSearcher, elapsedDrape, "searcher speeds up drape");
+        ck.testLE(elapsedDrapeXY, elapsedDrape, "drapeXY speeds up vertical drape");
+
+        const hole0 = BagOfCurves.create(...sweptEdges0).collectCurvePrimitives(undefined, true, true) as LineSegment3d[];
+        const hole1 = BagOfCurves.create(...sweptEdges1).collectCurvePrimitives(undefined, true, true) as LineSegment3d[];
+        const hole2 = BagOfCurves.create(...sweptEdges2).collectCurvePrimitives(undefined, true, true) as LineSegment3d[];
+        ck.testTrue(compareLineSegmentArrays(hole0, hole1, true), "drape compares to drape with searcher");
+        ck.testTrue(compareLineSegmentArrays(hole1, hole2, true), "drape with searcher compares to drapeXY");
+
+        GeometryCoreTestIO.captureCloneGeometry(allGeometry, hole0, x0, 0, z0 += 2);
```
saeeedtorabi, line 1516: any reason for not capturing `hole2`?

dassaf4: Yes, because `hole0` and `hole2` represent existing logic with equivalent output. And I already capture `hole0`. I am just testing here that the output of the accelerated branch of `sweepLineStringToFacets` agrees with both its non-accelerated branch, and with the output of `sweepLineStringToFacetsXY`.

core/backend/src/test/standalone/ExportGraphics.test.ts
dassaf4: I used the test naming convention prevalent in this file.

core/geometry/src/geometry3d/Ray3d.ts
```diff
@@ -397,41 +397,29 @@ export class Ray3d implements BeJSONFunctions {
...
-     * u = [(rayOrigin - v0).(rayVector x (v2-v0))] / [(v1-v0).(rayVector x (v2-v0))]
-     * v = [-rayVector.((rayOrigin - v0) x (v1-v0))] / [(v1-v0).(rayVector x (v2-v0))]
-     *
-     * Note that we should verify 0 <= u,v,w <= 1. To do so we only need to check 0 <= u <= 1, 0 <= v, and u+v <= 1.
-     * That's because w = 1-(u+v) and if we have those 4 checks, it's guaranteed that v <= 1 and 0 <= u+v and so
-     * 0 <= w <= 1.
-     *
-     * More info be found at
-     * https://en.wikipedia.org/wiki/M%C3%B6ller%E2%80%93Trumbore_intersection_algorithm
+     * More info be found at https://en.wikipedia.org/wiki/M%C3%B6ller%E2%80%93Trumbore_intersection_algorithm.
```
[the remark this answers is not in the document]
dassaf4: Nice!

common/api/core-backend.api.md
```diff
@@ -2680,18 +2680,30 @@ export interface ExportGraphicsMesh {
...
+    getVisitableFacetCount(): number;
+    moveToNextFacet(): boolean;
+    moveToReadIndex(subsetIndex: number): boolean;
+    parentFacetIndex(subsetIndex?: number): number | undefined;
+    reset(): void;
+}
+
 // @public
 export class ExportGraphicsMeshVisitor extends PolyfaceData implements PolyfaceVisitor {
+    protected constructor(facets: ExportGraphicsMesh, numWrap: number);
```
pmconne, line 2695: Adding a protected constructor replaces the former default public no-arg constructor. That's a breaking change to a `@public` API.

common/api/core-backend.api.md
```diff
@@ -2680,18 +2680,30 @@ export interface ExportGraphicsMesh {
...
     clearArrays(): void;
     clientAuxIndex(_i: number): number;
     clientColorIndex(_i: number): number;
     clientNormalIndex(i: number): number;
     clientParamIndex(i: number): number;
     clientPointIndex(i: number): number;
-    clientPolyface(): Polyface;
-    static create(polyface: ExportGraphicsMesh, numWrap: number): ExportGraphicsMeshVisitor;
+    clientPolyface(): Polyface | undefined;
```
pmconne, line 2702: This changes the return type of a `@public` method in a breaking way.

dassaf4: Even when we always return `undefined` typecast away?

aruniverse:

cc @bsy-nicholasw

common/api/core-backend.api.md
dassaf4: Currently, this constructor is private; is there a public no-arg constructor? I elevated it to protected so I could subclass.

common/api/core-backend.api.md
pmconne: Definition of a breaking API change = previously valid code produces a compilation error when I upgrade to new version.

My existing code compiles and works:
```ts
  console.log(meshVisitor.clientPolyface.twoSided);
```

When I upgrade to the new version I'll suddenly get an error because `meshVisitor.clientPolyface` may now be undefined.

dassaf4: I guess I don't understand since this visitor always returns `undefined` for `clientPolyface`. I admit the double cast to `any` and `Polyface` is mysterious. But it's `undefined` under the hood. How does your code run?

common/api/core-backend.api.md
pmconne: Never mind, extract-api doesn't include private constructor in its output so it looked like it was newly added, not changed.

common/api/core-backend.api.md
pmconne: > `public clientPolyface(): Polyface { return (undefined as unknown) as Polyface; }`

Wow, what the heck was the point of that.

I am reviewing the api.md files without looking at implementation. Yes, in this case the old API could never have worked, so breaking it = fixing it.

pmconne (approved):

PR #8103, opened 2025-05-15
Bug fix: Google Maps would only use the supplied session manager if an API key was also configured.

mdastous-bentley:
The session manager should be used, no matter if an API key is set or not in the map layer format registry.


aruniverse (commented):
can we also modify this pkg to not auto register the google maps formats? or add an option to opt in/out of the current behavior

aruniverse:

mdastous-bentley:
> can we also modify this pkg to not auto register the google maps formats? or add an option to opt in/out of the current behavior

What would be the benefit ?  I feel like it's going to make the whole thing even more complicated to consume (if we change the default behavior to not auto-register)

aruniverse (approved):

Issue #8036, opened 2025-05-02
Add support for Google Photorealistic 3d Tiles


aruniverse:
As of https://github.com/iTwin/itwinjs-core/pull/7604 we support [Google Maps 2D Tiles](https://developers.google.com/maps/documentation/tile/2d-tiles-overview). We should look to add support for [Google Photorealistic 3D Tiles](https://developers.google.com/maps/documentation/tile/3d-tiles).

Areas of concern we need to investigate is how to work with terrain and draped imagery / reality data / etc

dkbraig:
@aruniverse can we break this down into a series of task to review the areas of concern and how we might address them.  Given that we will want to add support for GP3DT as we move towards providing a consistent view / geospatial context across our applications.

matmarchand:
I wanted to see if we support GP3DT dataset in iTwin.js so similar to what can be done in Cesium:
` Cesium.Cesium3DTileset.fromUrl("https://tile.googleapis.com/v1/3dtiles/root.json?key=XXX");`

I attached GP3DT in iTwin.js:
`viewport.displayStyle.attachRealityModel(
{ tilesetUrl: "https://tile.googleapis.com/v1/3dtiles/root.json?key=XXX", name: "googleMap3dTiles" });`

The root will load fine but it will fail to load sub-tiles. 
Example:
'`https://tile.googleapis.com/v1/3dtiles//v1/3dtiles/datasets/CgA/files/UlRPVEYuYnVsa21ldGFkYXRhLnBsYW5ldG9pZD1lYXJ0aCxidWxrX21ldGFkYXRhX2Vwb2NoPTk5NyxwYXRoPSxjYWNoZV92ZXJzaW9uPTYsYWxpZ25tZW50X3ZlcnNpb249Uk9DS1RSRUVfOTk3X0dPT0dMRV9EQVRVTV8yMDI1MDIwNlQxNDUzWl9nZW5lcmF0ZWRfYXRfMjAyNTA0MThUMTIwOFo.json?session=CJ2fmueU_ILr_wEQpIjvwAY`

Problems:
1) The API key was not propagated.
2) Sub tile URL was not properly constructed.  ...googleapis.com/**v1/3dtiles//v1/3dtiles**/datasets/CgA/files...

So on top of handling the Google curated API authentication, terrain and drape imagery... we also need to look into supporting this specific dataset in our Cesium3dTiles implementation.

markschlosseratbentley:
[...]

PR #8104, opened 2025-05-15
Support Google Photorealistic 3D Tiles in iTwin.js

markschlosseratbentley:
Related to https://github.com/iTwin/itwinjs-core/issues/8036.

iTwin.js can already attach tilesets from a variety of sources. [Google Photorealistic 3d Tiles](https://developers.google.com/maps/documentation/tile/3d-tiles) are in the common 3D Tiles format so in theory should work in iTwin.js. However, a few hurdles were found which this PR works around.

This PR adds a new `RealityDataSourceG3DTImpl` implementation of `RealityDataSource` which can be used to serve up Google Photorealistic 3D Tiles if you have a valid key to access them. This `RealityDataSourceG3DTImpl` implementation does the following:
- implements `usesGeometricError` as `true` which tells iTwin.js to use the geometric errors found inside the G3DT tileset, which is crucial for good performance.
- fixes some base URL handling for G3DT tilesets that `RealityDataSourceTilesetUrlImpl` did not handle properly.
- fixes query param propagation so they get passed down to children tiles from child json files. This could be fixed more thoroughly.

To use this implementation, attach a reality model to your `ViewState` in the following manner:

```ts
const url = `https://tile.googleapis.com/v1/3dtiles/root.json?key=YOUR_GOOGLE_3D_TILES_KEY_GOES_HERE`;
viewState.displayStyle.attachRealityModel({
  tilesetUrl: url,
  name: "googleMap3dTiles",
  rdSourceKey: {
      provider: "G3DT",
      format: "ThreeDTile",
      id: url,
    },
  });
```

TODO (maybe in this PR, maybe not):
- [ ] This PR, in the code, uses the acronym `G3DT` for Google Photorealistic 3d Tiles.  Should we use `GP3DT` instead? I have seen both forms used.
- [ ] Should the base URL handling improvements also end up on `RealityDataSourceTilesetUrlImpl`?
- [ ] How can we do the query param propagation in a way that actually considers the entire tree structure, rather than just checking for json files? (see code for how it works, and comments).

### DIFF
--- a/core/common/src/ContextRealityModel.ts
+++ b/core/common/src/ContextRealityModel.ts
@@ -51,6 +51,10 @@ export enum RealityDataProvider {
    * Will provide Open Street Map Building (OSM) from Cesium Ion (in 3dTile format)
    */
   CesiumIonAsset = "CesiumIonAsset",
+  /**
+   * Will provide Google Photorealistic 3D Tiles (G3DT) from Google (in 3dTile format)
+   */
+  G3DT = "G3DT",
 }
 
 /** Identify the Reality Data storage format
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-frontend",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "iTwin.js frontend components",
   "main": "lib/cjs/core-frontend.js",
   "module": "lib/esm/core-frontend.js",
--- a/core/frontend/src/RealityDataSource.ts
+++ b/core/frontend/src/RealityDataSource.ts
@@ -12,6 +12,7 @@ import { CesiumIonAssetProvider, ContextShareProvider, getCesiumAssetUrl } from 
 import { RealityDataSourceTilesetUrlImpl } from "./RealityDataSourceTilesetUrlImpl";
 import { RealityDataSourceContextShareImpl } from "./RealityDataSourceContextShareImpl";
 import { RealityDataSourceCesiumIonAssetImpl } from "./RealityDataSourceCesiumIonAssetImpl";
+import { RealityDataSourceG3DTImpl } from "./RealityDataSourceG3DTImpl";
 import { IModelApp } from "./IModelApp";
 import { Range3d } from "@itwin/core-geometry";
 
@@ -252,6 +253,9 @@ export class RealityDataSourceProviderRegistry {
     this.register(RealityDataProvider.OrbitGtBlob, {
       // ###TODO separate TilesetUrlImpl
       createRealityDataSource: async (key, iTwinId) => RealityDataSourceTilesetUrlImpl.createFromKey(key, iTwinId),
+    });
+    this.register(RealityDataProvider.G3DT, {
+      createRealityDataSource: async (key, iTwinId) => RealityDataSourceG3DTImpl.createFromKey(key, iTwinId),
     });
   }
 
--- a/core/frontend/src/RealityDataSource.ts
+++ b/core/frontend/src/RealityDataSource.ts
@@ -214,6 +214,10 @@ export namespace RealityDataSource {
 
     return provider.createRealityDataSource(key, iTwinId);
   }
+
+  export function setBaseUrl(id: string) {
+    throw new Error(`Function not implemented. rdSourceId: ${id}}`);
+  }
 }
 
 /** A named supplier of [RealityDataSource]]s.
--- a/common/changes/@itwin/core-frontend/webpack-magic-comment_2025-05-13-17-35.json
+++ b/common/changes/@itwin/core-frontend/webpack-magic-comment_2025-05-13-17-35.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/common/config/rush/version-policies.json
+++ b/common/config/rush/version-policies.json
@@ -2,7 +2,7 @@
   {
     "policyName": "prerelease-monorepo-lockStep",
     "definitionName": "lockStepVersion",
-    "version": "5.1.0-dev.14",
+    "version": "5.1.0-dev.15",
     "nextBump": "prerelease"
   }
 ]
--- a/core/backend/package.json
+++ b/core/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-backend",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "iTwin.js backend components",
   "main": "lib/cjs/core-backend.js",
   "module": "lib/esm/core-backend.js",
--- a/core/bentley/package.json
+++ b/core/bentley/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-bentley",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "Bentley JavaScript core components",
   "main": "lib/cjs/core-bentley.js",
   "module": "lib/esm/core-bentley.js",
--- a/core/common/package.json
+++ b/core/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-common",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "iTwin.js components common to frontend and backend",
   "main": "lib/cjs/core-common.js",
   "module": "lib/esm/core-common.js",
--- a/core/ecschema-editing/package.json
+++ b/core/ecschema-editing/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-editing",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "ECSchema editing and validation API",
   "license": "MIT",
   "main": "lib/cjs/ecschema-editing.js",
--- a/core/ecschema-locaters/package.json
+++ b/core/ecschema-locaters/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-locaters",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "EC Schema file locaters",
   "license": "MIT",
   "main": "lib/cjs/ecschema-locaters.js",
--- a/core/ecschema-metadata/package.json
+++ b/core/ecschema-metadata/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-metadata",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "ECObjects core concepts in typescript",
   "license": "MIT",
   "main": "lib/cjs/ecschema-metadata.js",
--- a/core/ecschema-rpc/common/package.json
+++ b/core/ecschema-rpc/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-rpcinterface-common",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "Schema RPC Interface common interface",
   "main": "lib/cjs/ecschema-rpc-interface.js",
   "module": "lib/esm/ecschema-rpc-interface.js",
--- a/core/ecschema-rpc/impl/package.json
+++ b/core/ecschema-rpc/impl/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-rpcinterface-impl",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "Schema RPC Interface backend implementation",
   "main": "lib/cjs/ecschema-rpc-impl.js",
   "module": "lib/esm/ecschema-rpc-impl.js",
--- a/core/ecsql/common/package.json
+++ b/core/ecsql/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecsql-common",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "ECSql component that can be reference on backend and frontend",
   "main": "lib/cjs/ecsql-common.js",
   "module": "lib/esm/ecsql-common.js",
--- a/core/electron/package.json
+++ b/core/electron/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-electron",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "iTwin.js ElectronHost and ElectronApp",
   "license": "MIT",
   "engines": {
--- a/core/express-server/package.json
+++ b/core/express-server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/express-server",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "iTwin.js express utilities",
   "main": "lib/cjs/express-server.js",
   "module": "lib/esm/express-server.js",
--- a/core/extension/package.json
+++ b/core/extension/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-extension",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "iTwin.js Extension API",
   "type": "module",
   "typings": "index.d.ts",
--- a/core/frontend-devtools/package.json
+++ b/core/frontend-devtools/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/frontend-devtools",
-  "version": "5.1.0-dev.14",
+  "version": "5.1.0-dev.15",
   "description": "Debug menu and supporting UI widgets",
   "main": "lib/cjs/frontend-devtools.js",
   "module": "lib/esm/frontend-devtools.js",
--- a/core/frontend/src/RealityDataSourceG3DTImpl.ts
+++ b/core/frontend/src/RealityDataSourceG3DTImpl.ts
@@ -1,0 +1,196 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Tiles
+ */
+import { assert, BentleyStatus, GuidString } from "@itwin/core-bentley";
+import { IModelError, RealityData, RealityDataFormat, RealityDataProvider, RealityDataSourceKey, RealityDataSourceProps } from "@itwin/core-common";
+
+import { request } from "./request/Request";
+import { PublisherProductInfo, RealityDataSource, SpatialLocationAndExtents } from "./RealityDataSource";
+import { ThreeDTileFormatInterpreter } from "./tile/internal";
+
+/** This class provides access to the reality data provider services.
+ * It encapsulates access to a reality data from the Google Photorealistic 3D Tiles service.
+ * The caller must pass in a valid GP3D key as a parameter of the URL they specify when attaching reality data that uses this data source.
+* @internal
+*/
+export class RealityDataSourceG3DTImpl implements RealityDataSource {
+  public readonly key: RealityDataSourceKey;
+  /** The URL that supplies the 3d tiles for displaying the G3DT tileset. */
+  private _tilesetUrl: string | undefined;
+  /** Base URL of the G3DT tileset. Includes trailing subdirectories. */
+  private _baseUrlWithSubDirs: string = "";
+  /** Same as _baseUrlWithSubDirs, but without any trailing subdirectories (the true website root). */
+  private _baseUrlNoSubDirs: string = "";
+
+  /** Need to be passed down to child tile requests */
+  private _searchParams?: URLSearchParams;
+
+  /** This is necessary for G3DT tilesets! This tells the iTwin.js tiling system to use the geometric error specified in the G3DT tileset rather than any of our own. */
+  public readonly usesGeometricError = true;
+
+  /** Construct a new reality data source.
+   * @param props JSON representation of the reality data source
+   */
+  protected constructor(props: RealityDataSourceProps) {
+    assert(props.sourceKey.provider === RealityDataProvider.G3DT);
+    this.key = props.sourceKey;
+    this._tilesetUrl = this.key.id;
+  }
+
+  /**
+   * Create an instance of this class from a source key and iTwin context.
+   */
+  public static async createFromKey(sourceKey: RealityDataSourceKey, _iTwinId: GuidString | undefined): Promise<RealityDataSource | undefined> {
+    if (sourceKey.provider !== RealityDataProvider.G3DT)
+      return undefined;
+    const rdSource = new RealityDataSourceG3DTImpl({ sourceKey });
+    return rdSource;
+  }
+
+  public get isContextShare(): boolean {
+    return false;
+  }
+  /**
+   * Returns Reality Data if available
+  */
+  public get realityData(): RealityData | undefined {
+    return undefined;
+  }
+  public get realityDataId(): string | undefined {
+    return undefined;
+  }
+  /**
+   * Returns Reality Data type if available
+   */
+  public get realityDataType(): string | undefined {
+    return undefined;
+  }
+
+  public getTilesetUrl(): string | undefined {
+    return this._tilesetUrl;
+  }
+
+  private setBaseUrl(url: string): void {
+    const urlParts = url.split("/");
+    const newUrl = new URL(url);
+    this._searchParams = newUrl.searchParams
+    urlParts.pop();
+    if (urlParts.length === 0) {
+      this._baseUrlWithSubDirs = "";
+      this._baseUrlNoSubDirs = "";
+    } else {
+      this._baseUrlWithSubDirs = `${urlParts.join("/")}/`;
+      this._baseUrlNoSubDirs = `${urlParts[0]}/${urlParts[1]}/${urlParts[2]}/`;
+    }
+  }
+
+  /**
+   * This method returns the URL to access the actual 3d tiles from the service provider.
+   * @returns string containing the URL to reality data.
+   */
+  public async getServiceUrl(_iTwinId: GuidString | undefined): Promise<string | undefined> {
+    return this._tilesetUrl;
+  }
+
+  public async getRootDocument(iTwinId: GuidString | undefined): Promise<any> {
+    const url = await this.getServiceUrl(iTwinId);
+    if (!url)
+      throw new IModelError(BentleyStatus.ERROR, "Unable to get service url");
+
+    this.setBaseUrl(url);
+    return request(url, "json");
+  }
+
+  /** Returns the tile URL relative to the base URL.
+   * If the tile path is a relative URL, the base URL is prepended to it.
+   * For both absolute and relative tile path URLs, the search parameters are checked. If the search params are empty, the base URL's search params are appended to the tile path.
+   */
+  private getTileUrl(tilePath: string): string {
+    //&&MM Notes
+    // Spec: https://docs.ogc.org/cs/18-053r2/18-053r2.html 6.4 and 7.8.2
+    //  query params:
+    //    - no mention about propagation/preservation of query params.
+    //    - not sure if absolute url should receive params from base url?
+    // - "When the URI is relative, its base is always relative to the referring tileset JSON file."
+    //     - We are always using the base URL of the root document but we need to use the referring tileset which is nested in the tree structure. In case of google they add the session id.
+    // - What are we breaking by using the URL API instead of a simple concatenation of the base URL and tile path?
+
+    // TODO absolute vs relative URL handling?
+
+    const baseUrl = tilePath.startsWith("/") ? this._baseUrlWithSubDirs : this._baseUrlNoSubDirs;
+
+    // Build a relative URL where tilePath params will win over base URL params.
+
+    const url = new URL(tilePath, baseUrl);
+
+    // maybe use getTileContentType and check for tilset.json?
+    // If URL is to json file, store search params
+    const isJson = url.pathname.toLowerCase().endsWith("json");
+    if (isJson && url.searchParams.size !== 0) {
+      for (const [key, value] of url.searchParams.entries()) {
+        this._searchParams?.append(key, value);
+      }
+    }
+
+    // console.log("search params:");
+    // if (this._searchParams !== undefined) {
+    //   for (const [key, value] of this._searchParams.entries()) {
+    //     console.log(`${key}, ${value}`);
+    //   }
+    // }
+
+    if (this._searchParams === undefined || this._searchParams.size === 0)
+      return url.toString();
+
+    // append the base URL params to the tile path URL
+    const result = url.searchParams.size === 0 ? `${url}?${this._searchParams}` :`${url}&${this._searchParams}`;
+    return result;
+  }
+
+  /**
+   * Returns the tile content. The path to the tile is relative to the base url of present reality data whatever the type.
+   */
+  public async getTileContent(name: string): Promise<ArrayBuffer> {
+    return request(this.getTileUrl(name), "arraybuffer");
+  }
+
+  /**
+   * Returns the tile content in json format. The path to the tile is relative to the base url of present reality data whatever the type.
+   */
+  public async getTileJson(name: string): Promise<any> {
+    return request(this.getTileUrl(name), "json");
+  }
+
+  public getTileContentType(url: string): "tile" | "tileset" {
+    return new URL(url, "https://localhost/").pathname.toLowerCase().endsWith("json") ? "tileset" : "tile";
+  }
+
+  /**
+   * Gets spatial location and extents of this reality data source
+   * @returns spatial location and extents
+   * @internal
+   */
+  public async getSpatialLocationAndExtents(): Promise<SpatialLocationAndExtents | undefined> {
+    let spatialLocation: SpatialLocationAndExtents | undefined;
+    if (this.key.format === RealityDataFormat.ThreeDTile) {
+      const rootDocument = await this.getRootDocument(undefined);
+      spatialLocation = ThreeDTileFormatInterpreter.getSpatialLocationAndExtents(rootDocument);
+    }
+    return spatialLocation;
+  }
+  /**
+   * Gets information to identify the product and engine that create this reality data
+   * Will return undefined if cannot be resolved
+   * @returns information to identify the product and engine that create this reality data
+   * @alpha
+   */
+  public async getPublisherProductInfo(): Promise<PublisherProductInfo | undefined> {
+    let publisherInfo: PublisherProductInfo | undefined;
+    return publisherInfo;
+  }
+}
+
--- a/core/frontend/src/RealityDataSourceG3DTImpl.ts
+++ b/core/frontend/src/RealityDataSourceG3DTImpl.ts
@@ -21,11 +21,8 @@ export class RealityDataSourceG3DTImpl implements RealityDataSource {
   public readonly key: RealityDataSourceKey;
   /** The URL that supplies the 3d tiles for displaying the G3DT tileset. */
   private _tilesetUrl: string | undefined;
-  /** Base URL of the G3DT tileset. Includes trailing subdirectories. */
-  private _baseUrlWithSubDirs: string = "";
-  /** Same as _baseUrlWithSubDirs, but without any trailing subdirectories (the true website root). */
-  private _baseUrlNoSubDirs: string = "";
-
+  /** Base URL of the G3DT tileset. Does not include trailing subdirectories. */
+  private _baseUrl: string = ""
   /** Need to be passed down to child tile requests */
   private _searchParams?: URLSearchParams;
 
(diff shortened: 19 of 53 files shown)


matmarchand:
Thanks for your effort on this.
I understand the careful approach of having a separate source implementation but I feel like `RealityDataSourceTilesetUrlImpl` and `RealityDataSourceG3DTImpl` should merge since they are based on the same standard. In that sense, maybe we should avoid exposing new APIs and hide that complexity internally by enabling Google behavior based on the url?

Unrelated to this PR but I'm surprised that [GeometricError](https://github.com/CesiumGS/3d-tiles/blob/main/specification/README.adoc#geometric-error) is not enabled by default.  Would we get a performance improvement for reality mesh as well?

pmconne:
> Unrelated to this PR but I'm surprised that [GeometricError](https://github.com/CesiumGS/3d-tiles/blob/main/specification/README.adoc#geometric-error) is not enabled by default. Would we get a performance improvement for reality mesh as well?

Yes, at the expense of drawing much lower-resolution tiles for the vast majority of reality models used with iTwins.
Apparently most reality models produced by Bentley have been created with a target error of 1 versus Cesium's 16. So they render with appropriate LOD in iTwin.js and need to override the default error to render properly in CesiumJS.
I think some changes were made in that area recently - can you check with that team?

markschlosseratbentley:
> Thanks for your effort on this. I understand the careful approach of having a separate source implementation but I feel like `RealityDataSourceTilesetUrlImpl` and `RealityDataSourceG3DTImpl` should merge since they are based on the same standard. In that sense, maybe we should avoid exposing new APIs and hide that complexity internally by enabling Google behavior based on the url?

Doing everything in `RealityDataSourceTilesetUrlImpl` was the original approach, but when we found we needed to define different behavior for `usesGeometricError` for Google Photorealistic 3D Tiles, we made a separate source implementation.

We could just put all of this code in `RealityDataSourceTilesetUrlImpl` and conditionally trigger it based on the URL, but what if the GP3DT URL changes in the future?

I also hesitate to throw all of this into `RealityDataSourceTilesetUrlImpl` because [this comment](https://github.com/iTwin/itwinjs-core/blob/50ccc964b14a4fa4a6f91c9db3892dda127fe020/core/common/src/ContextRealityModel.ts#L31) indicates that `RealityDataProvider.TilesetUrl` is considered legacy.

I generally agree that the proper handling of all of this stuff should eventually also / and go into other source implementations that parse 3d Tiles format tiles, probably getting combined into one implementation.

Should that work be in followup PR? (This current fix is not 100% proper, it just gets GP3DT up and running).

pmconne:
> ```ts
> const url = `https://tile.googleapis.com/v1/3dtiles/root.json?key=YOUR_GOOGLE_3D_TILES_KEY_GOES_HERE`;
> viewState.displayStyle.attachRealityModel({
>   tilesetUrl: url,
> ```

The url gets stored in the display style JSON. People will not want to store their Google 3D Tiles API key in there. The primary job of the various map and reality data "provider" objects is to handle authentication and format URLs. Shouldn't your provider implement those similarly to `RealityDataSourceCesiumIonAssetImpl`?

markschlosseratbentley:
> > ```ts
> > const url = `https://tile.googleapis.com/v1/3dtiles/root.json?key=YOUR_GOOGLE_3D_TILES_KEY_GOES_HERE`;
> > viewState.displayStyle.attachRealityModel({
> >   tilesetUrl: url,
> > ```
> 
> The url gets stored in the display style JSON. People will not want to store their Google 3D Tiles API key in there. The primary job of the various map and reality data "provider" objects is to handle authentication and format URLs. Shouldn't your provider implement those similarly to `RealityDataSourceCesiumIonAssetImpl`?

Agree.

eringram:
@markschlosseratbentley I added attributions [ordered by number of occurrences](https://developers.google.com/maps/documentation/tile/create-renderer#display-attributions). If there are copyright properties in the tiles but the tree provider is not `RealityDataProvider.G3DT`, it still displays them but without the Google header and icon. If there are no copyrights there will be no new logo card added to this list.

image

I also started working on adding the logo as a decorator on top of the view itself. `GoogleMapsImageryProvider` uses `GoogleMapsDecorator` to achieve this which I think we should reuse, but it's in the map-layer-formats package which we don't want to import into core-frontend, so maybe that class can be moved (it is internal). I'll try that tomorrow morning

mdastous-bentley (approved):

core/frontend/src/RealityDataFormatRegistry.ts
```diff
@@ -0,0 +1,47 @@
...
+export interface RealityDataOptions {
+  /** Access key for Google Photorealistic 3D Tiles in the format `{ key: "key", value: "your-gp3dt-key" }`. */
+  gp3dt?: RealityDataKey;
+}
+
+/**
+ * A registry of RealityDataFormats identified by their unique format IDs. The registry can be accessed via [[IModelApp.realityDataFormatRegistry]].
+ * @alpha
+ */
+export class RealityDataFormatRegistry {
```
pmconne, line 37: This should probably be a *provider* registry not *format* - google's tiles are in the same format as ION's, Context Capture's, etc - what differs is the provider logic (authentication, url formatting, etc). That would match with `RealityDataSourceKey.provider`.

markschlosseratbentley: My takeaways from this feedback:
1) Make this be `RealityDataProviderRegistry`.
2) Add something [like this](https://github.com/iTwin/itwinjs-core/blob/c9958802ae321dbf54dd7f4eea74c429fad6da51/core/frontend/src/tile/map/MapLayerFormatRegistry.ts#L113) to `RealityDataOptions`.

Is there something further you are thinking of?  It does just basically wrap an auth key right now.

pmconne: No, the crucial thing your registry is missing compared to `MapLayerFormatRegistry` is the `register` function. The "config" thing with key-value pairs IMO is not great; it'd make more sense to me if somebody instantiates a `Google3DTilesProvider`, passing it the API key and whatever else it needs, and then registers that with your registry.

common/api/core-frontend.api.md
```diff
@@ -8069,6 +8100,26 @@ export class RealityDataError extends BentleyError {
...
 
+// @alpha
+export class RealityDataFormatRegistry {
+    constructor(opts?: RealityDataOptions);
+    // (undocumented)
+    get configOptions(): RealityDataOptions;
+}
+
+// @alpha
+export interface RealityDataKey {
```
matmarchand, line 8111: For our [Google Curated API](https://developer.bentley.com/apis/google-curated-content/operations/access-3d-tiles/), authorization will be provided via authorization header instead of api key.  Could we add support for authorization header as well?

core/frontend/src/GoogleMapsDecorator.ts
aruniverse, line 1: you would need to copy the assets this decorator needs from map-layers-fromats into core-frontend

core/frontend/src/tile/Tile.ts
aruniverse, line 1: revert whitespace polluting diff

core/frontend/src/RealityDataProviderRegistry.ts
```diff
@@ -0,0 +1,49 @@
...
+  value: string;
+}
+
+/** Options supplied at startup via [[IModelAppOptions.realityDataOptions]] to specify access keys for various reality data formats.
+ * `gp3dt` must have its key value set to `key`
+ * @alpha
+ */
+export interface RealityDataProviderOptions {
+  /** Access key for Google Photorealistic 3D Tiles in the format `{ key: "key", value: "your-gp3dt-key" }`. */
+  gp3dt?: RealityDataProviderKey;
```
aruniverse, line 30: should we be hardcoding these in core-frontend?
also take into account those who will not be providing an access token/key, but has to interface with a service to obtain a session token. you should look at what was done for google 2d tiles

core/frontend/src/RealityDataSource.ts
```diff
@@ -221,6 +229,11 @@ export namespace RealityDataSource {
  * is invoked to produce the reality data source.
  * @alpha
  */
+
+// TODO extend this for GP3DT
```
aruniverse, line 233: how applicable are these todos still?

core/frontend/src/internal/tile/RealityModelTileTree.ts
```diff
@@ -979,9 +1001,48 @@ export class RealityTreeReference extends RealityModelTileTree.Reference {
...
+            copyrightMap.set(copyright, currentCount ? currentCount + 1 : 1);
+          }
+        }
+      }
+    }
+
+    // Only add another logo card if the tiles have copyright
+    if (copyrightMap.size > 0) {
+      // Order by most occurances to least
+      // See https://developers.google.com/maps/documentation/tile/create-renderer#display-attributions
```
aruniverse, line 1021: based off of this, i wonder if we need an additional / clickable `data attributions` thats always visible on the vp and when clicked also shows the logo card

aruniverse (commented):
not sure if this pr is actually ready for review, and still needs to be cleaned up to not cut corners like paul mentioned elsewhere. adding a few minor comments.

if not ready, please draft

core/frontend/src/tile/Tile.ts
eringram: Done

core/frontend/src/GoogleMapsDecorator.ts
eringram: Fixed, thanks, good catch. I guess this didn't show up as an error DTA because the assets were also included from map-layers-formats?

core/frontend/src/RealityDataSource.ts
eringram: `RealityDataSourceGP3DTProvider` is still technically WIP so I reverted this to draft

core/frontend/src/RealityDataProviderRegistry.ts
markschlosseratbentley: Based on earlier feedback (and now your feedback as well), we are in the process of changing the way this works. New approach:

1. User can create a GP3DT provider (new API class)
2. User can register that provider themselves if they want to use it.
3. User will be responsible for passing in a valid key, if desired.
4. An alternate using a session token / [authorization header](https://github.com/iTwin/itwinjs-core/pull/8104#discussion_r2103521100) will be available - but we need to develop this.

Stay tuned for further code updates in these areas (the first three of these are done; we need to clean up old classes still).

core/frontend/src/RealityDataSource.ts
markschlosseratbentley: Thanks for reverting to draft, @eringram. Thanks @aruniverse for the feedback - we were unsure how fast we would try to push this through until we had some more discussions.

core/frontend/src/internal/tile/RealityModelTileTree.ts
markschlosseratbentley: For reference; more details about attribution requirements are discussed here: https://github.com/iTwin/itwinjs-core/issues/8036#issuecomment-2895776072

common/api/core-frontend.api.md
eringram: What kind of authorization - like a bearer token, and from what service (Bentley IMS/other)? And can this replace the API key? From what I understand, the API key is [required for all the Google 3D Tiles requests](https://developers.google.com/maps/documentation/tile/create-renderer#requests-and-responses), so I'm not sure how they would be accessed without the user providing one.

Maybe this is also relevant to @aruniverse's comment below. I don't think the Google Photorealistic 3D Tiles allow you to use a session token in the same way as the 2D Map Tiles. I don't see a mention of sessions in the 3D Tiles docs, and see [this ](https://developers.google.com/maps/documentation/tile/session_tokens)in the 2D Tiles:
>Note: You need to use session tokens for getting 2D Tiles and Street View Tiles, but not for getting 3D Tiles.

Let me know if we're missing a puzzle piece here though, and there's another service that can provide auth/sessions in this context

matmarchand: Yes a Bearer token in the authorization header. ref: https://developers.google.com/maps/documentation/tile/oauth_token

eringram: Thanks for clarifying. In [this commit](https://github.com/iTwin/itwinjs-core/pull/8104/commits/25950bda1635986f6cca9e3a3a60baca7fbf773e) I added an option so when the provider is created, the user can pass in a function that returns an authorization token. That function is called to provide the token for the authorization header when the google tiles are requested.

Working on cleaning this up to make the API key optional (I assume it won't be required in the auth header case?) and add documentation

common/api/core-frontend.api.md
```diff
@@ -8115,11 +8146,24 @@ export namespace RealityDataSource {
     export function createOrbitGtBlobPropsFromKey(rdSourceKey: RealityDataSourceKey): OrbitGtBlobProps | undefined;
     // @alpha
     export function fromKey(key: RealityDataSourceKey, iTwinId: GuidString | undefined): Promise<RealityDataSource | undefined>;
+    // @alpha
+    export function setBaseUrl(id: string): void;
+}
+
+// @alpha
+export class RealityDataSourceGP3DTProvider implements RealityDataSourceProvider {
+    constructor(options: RealityDataSourceGP3DTProviderOptions);
```
markschlosseratbentley: Am I missing it, or did RealityDataSourceGP3DTProviderOptions not show up in extract-api?

eringram: You're right. pretty sure that's because I forgot to export it, oops

core/frontend/src/GoogleMapsDecorator.ts
```diff
@@ -0,0 +1,158 @@
...
+    "google_on_non_white";
+
+    // We need to move the logo right after the 'i.js' button
+    this.logo.offset = new Point3d(45, 10);
+
+    return this.logo.activate(IconSprites.getSpriteFromUrl(`${IModelApp.publicPath}images/${imageName}.png`));
+  };
+
+  /** Decorate implementation */
+  public decorate = (context: DecorateContext) => {
```
markschlosseratbentley: It might be worth investigating enabling `useCachedDecorations` for this scenario - it seems a good candidate.
Check out docs here: https://www.itwinjs.org/reference/core-frontend/views/viewportdecorator/usecacheddecorations/
Also see `CachedDecorations.tests.ts`, `IncidentMarkerDemo.ts` for examples of usage.

eringram:
This is ready for review again and the description has been updated to summarize the latest version.

core/frontend/src/internal/tile/RealityModelTileTree.ts
eringram: This has been added, see the [changes](https://github.com/iTwin/itwinjs-core/pull/8104/files#diff-89617e1a0dc7e63c0dc2d7d8f91a0f733c2133154df28e78493e728896d547ffR117-R135) to `GoogleMapsDecorator.decorate()`. The flag for showing these attributions is `true` by default when you create a `RealityDataSourceGP3DTProvider`, but otherwise undefined (see [here](https://github.com/iTwin/itwinjs-core/pull/8104/files#diff-589682f22153f4dd9eef15882ee1221bc3e81b7a065efa579ad0a5c1e15572d5R325)). So when the decorator is used by the map-layers-formats GoogleMapsProvider for 2D map imagery, no attributions will be shown, but they are on by default for the 3D photorealistic tiles

common/api/core-frontend.api.md
markschlosseratbentley: I see it has been added now; thanks. Resolving.

common/api/core-frontend.api.md
```diff
@@ -8115,11 +8146,34 @@ export namespace RealityDataSource {
     export function createOrbitGtBlobPropsFromKey(rdSourceKey: RealityDataSourceKey): OrbitGtBlobProps | undefined;
     // @alpha
     export function fromKey(key: RealityDataSourceKey, iTwinId: GuidString | undefined): Promise<RealityDataSource | undefined>;
+    // @alpha
+    export function setBaseUrl(id: string): void;
+}
+
+// @alpha
+export class RealityDataSourceGP3DTProvider implements RealityDataSourceProvider {
```
markschlosseratbentley: Let's make a NextVersion.md entry with content similar to the description of this PR:
- How to use new RealityDataSourceGP3DTProvider to display GP3DT tiles
- Make sure any included screenshots are small enough, jpeg, and include proper attribution in the viewable portion (bottom left).

core/frontend/src/GoogleMapsDecorator.ts
eringram: Added in [this commit](https://github.com/iTwin/itwinjs-core/pull/8104/commits/bc689988d3cc9adf1207d2e1ae23d223a1744eca).

The cached decorations weren't working when I just set `useCachedDecorations` to true on the `GoogleMapsDecorator`. When I debugged it, the decorator that actually had its `useCachedDecorations` checked was `RealityTreeReference`. So my solution is to set `RealityTreeReference.useCachedDecorations` based on its `RealityDataSourceProvider`, and make `RealityDataSourceGP3DTProvider.useCachedDecorations = true`.

core/common/src/ContextRealityModel.ts
```diff
@@ -28,22 +28,22 @@ export interface OrbitGtBlobProps {
  */
 export enum RealityDataProvider {
   /**
-   * This is the legacy mode where the access to the 3d tiles is harcoded in ContextRealityModelProps.tilesetUrl property.
-   * It was use to support RealityMesh3DTiles, Terrain3DTiles, Cesium3DTiles
-   * You should use other mode when possible
+   * This is the legacy mode where the access to the 3d tiles is hardcoded in ContextRealityModelProps.tilesetUrl property.
```
ben-polinsky, line 31: If it is legacy, do we want to deprecate?
`You should use other modes when possible` - to users experienced with tiles, is this clear enough for them to determine what other modes to use?

core/frontend/src/internal/tile/RealityModelTileTree.ts
```diff
@@ -979,9 +983,17 @@ export class RealityTreeReference extends RealityModelTileTree.Reference {
     }
   }
 
-  public override async addAttributions(cards: HTMLTableElement): Promise<void> {
-    // eslint-disable-next-line @typescript-eslint/no-deprecated
-    return Promise.resolve(this.addLogoCards(cards));
+  public override async addAttributions(cards: HTMLTableElement, vp: ScreenViewport): Promise<void> {
+    const provider = IModelApp.realityDataSourceProviders.find(this._rdSourceKey.provider);
+    if (provider && provider.addAttributions) {
```
ben-polinsky:
```suggestion
    if (provider?.addAttributions) {
```

core/frontend/src/internal/tile/RealityModelTileTree.ts
```diff
@@ -979,9 +983,17 @@ export class RealityTreeReference extends RealityModelTileTree.Reference {
...
+      await provider.addAttributions(cards, vp);
+    }
   }
-}
 
+  public override decorate(_context: DecorateContext): void {
+    const provider = IModelApp.realityDataSourceProviders.find(this._rdSourceKey.provider);
+    if (provider && provider.decorate) {
+      provider.decorate(_context);
```
ben-polinsky:
```suggestion
    if (provider?.decorate) {
      provider.decorate(_context);
```

core/frontend/src/GoogleMapsDecorator.ts
```diff
@@ -0,0 +1,158 @@
...
+/** Get copyrights from tiles currently in the viewport.
+ * @internal
+ */
+export function getCopyrights(vp: ScreenViewport): Map<string, number> {
+  const tiles = IModelApp.tileAdmin.getTilesForUser(vp)?.selected;
+  const copyrightMap = new Map<string, number>();
+  if (tiles) {
+    for (const tile of tiles as Set<RealityTile>) {
+      if (tile.copyright) {
+        for (const copyright of tile.copyright?.split(";")) {
```
ben-polinsky: nit: if you're checking for `tile.copyright` why do you need optional chaining on the next line?

core/frontend/src/internal/tile/RealityModelTileTree.ts
ben-polinsky: (nit)

core/frontend/src/internal/tile/RealityModelTileTree.ts
ben-polinsky: (nit)

core/frontend/src/GoogleMapsDecorator.ts
markschlosseratbentley: Thanks!

core/common/src/ContextRealityModel.ts
markschlosseratbentley: All we did for this line was fix a typo in the comments.  `harcoded` => `hardcoded`.  Do we want further changes? I think the comment says it is a legacy mode, but I think many, many cases still use it.

ben-polinsky: Ah indeed, thanks. Let's leave it.

ben-polinsky (approved):

core/frontend/src/GoogleMapsDecorator.ts
eringram: Yeah it's unnecessary, fixed

core/frontend/src/internal/tile/RealityModelTileTree.ts
eringram: Done

core/frontend/src/internal/tile/RealityModelTileTree.ts
eringram: Done

common/api/core-frontend.api.md
eringram: Pushed the version we discussed, with the logos easier to see in the screenshots.

markschlosseratbentley: Looks good, thanks.

docs/changehistory/NextVersion.md
```diff
@@ -15,6 +16,46 @@ Table of contents:
 
 In addition to [already supported Electron versions](../learning/SupportedPlatforms.md#electron), iTwin.js now supports [Electron 36](https://www.electronjs.org/blog/electron-36-0).
 
+## Google Photorealistic 3D Tiles support
+
+iTwin.js now supports displaying Google Photorealistic 3D Tiles (GP3DT) via the new `RealityDataSourceGP3DTProvider`. This provider handles authentication, tile loading, and attribution display for GP3DT.
```
hl662: This seems like a valuable feature users will want to revisit after the release. Can we move this into one of the docs/learning/frontend .md files? The entry for NextVer.md would be a hyperlink pointing to that section of the .md file.
That way, to onboard new users we can point them to the iTwinjs.org docs site instead of looking at version releases.

If there isn't a fitting existing file, you could add a new file that covers google api support.

markschlosseratbentley: Good suggestion, @hl662. @eringram, in addition to a learning article, it could be helpful to put any example code in the snippets directory in iTwin.js, so any future code changes which make the example code obsolete would get caught.

[Here is an example of some snippet code for contour display](https://github.com/iTwin/itwinjs-core/blob/master/example-code/snippets/src/frontend/ContourDisplay.ts).

[And here is an example of the contour learning article linking to that same snippet code](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/display/ContourDisplay.md).

eringram: Added a new markdown file to `docs/learning/frontend/` that references code snippets, and the changelog now links to that learning article. 

Side note, are any `alpha` APIs captured in the documentation website? Because the new `RealityDataSourceGP3DTProvider` is `alpha`, and I linked to it in the new learning article because I assumed it would be in the itwinjs.org auto-generated docs. But I see that the way you register it, with `IModelApp.realityDataSourceProviders.register()`, is `alpha` and not visible in the docs. To me it doesn't make much sense to have a learning article referencing an undocumented feature like this, so we should either mark the RealityDataSourceProvider-related APIs as `beta`, or make their docs visible despite staying `alpha`.

hl662: no, alpha apis are not published onto the docs site. If you have added doccomments for these new apis perhaps they should be beta?

markschlosseratbentley: I agree, we might as well make these APIs `beta`.

eringram: I promoted the new APIs in this PR to beta, as well as `IModelApp.realityDataSourceProviders`

core/frontend/src/GoogleMapsDecorator.ts
markschlosseratbentley: @aruniverse Can this one be resolved now?

core/frontend/src/tile/Tile.ts
markschlosseratbentley: @aruniverse Can this one be resolved now?

core/frontend/src/RealityDataProviderRegistry.ts
markschlosseratbentley: @aruniverse Can this one be resolved now? We have a new structure for handling this stuff.

core/frontend/src/RealityDataSource.ts
markschlosseratbentley: @aruniverse Can this one be resolved now? I believe all these TODOs are taken care of and removed.

core/frontend/src/internal/tile/RealityModelTileTree.ts
markschlosseratbentley: @aruniverse Can this one be resolved now?  See Erin's screenshots in the PR description for this VP attribution in action, in the lower left hand corner.

common/api/core-frontend.api.md
eringram: @matmarchand Let us know if the changes related to the authorization header look good and this comment can be resolved. This [example code snippet](https://github.com/iTwin/itwinjs-core/pull/8104/files#diff-0e13ea2bb64bedd980aabefda3df89d45c401fa908fb207338ced194cca05c16R35-R44) shows how to use the `getAuthToken` option.

matmarchand: Looks good thx!

common/api/core-frontend.api.md
```diff
@@ -3764,9 +3764,15 @@ export function getCesiumAssetUrl(osmAssetId: number, requestKey: string): strin
...
 export function getCompressedJpegFromCanvas(canvas: HTMLCanvasElement, maxBytes?: number, minCompressionQuality?: number): string | undefined;
 
+// @internal
+export function getCopyrights(vp: ScreenViewport): Map<string, number>;
+
 // @internal (undocumented)
 export function getFrustumPlaneIntersectionDepthRange(frustum: Frustum, plane: Plane3dByOriginAndUnitNormal): Range1d;
 
+// @beta
+export function getGooglePhotorealistic3DTilesURL(): string;
```
pmconne: nit: inconsistent naming. "GooglePhotorealistic3DTiles" here, "GP3DT" elsewhere. The former is a mouthful, the latter is cryptic. IMO somebody searching our docs for Google 3d tile support should be able to find it by typing "google", which won't work for GP3DT.
I suggest using "Google3dTiles" everywhere.

core/frontend/src/GoogleMapsDecorator.ts
```diff
@@ -0,0 +1,158 @@
+/*---------------------------------------------------------------------------------------------
```
pmconne: This file contains nothing but internal APIs. Please move it to src/internal and export it from src/internal/cross-package.ts, not from core-frontend.ts.

core/frontend/src/RealityDataSource.ts
```diff
@@ -213,27 +217,43 @@ export namespace RealityDataSource {
 
     return provider.createRealityDataSource(key, iTwinId);
   }
+
+  /** Implement this function in order to provide the base URL for the specified reality data source.
+   * @alpha
+   */
+  export function setBaseUrl(id: string) {
+    throw new Error(`Function not implemented. rdSourceId: ${id}}`);
```
pmconne: What is this for?

core/frontend/src/RealityDataSource.ts
```diff
@@ -265,3 +285,87 @@ export class RealityDataSourceProviderRegistry {
...
+  showCreditsOnScreen?: boolean;
+}
+
+/**
+ * Will provide Google Photorealistic 3D Tiles (GP3DT) from Google (in 3dTile format).
+ * A valid GP3DT API key or getAuthToken fuction must be supplied when creating this provider.
+ * To use this provider, you must register it with [[IModelApp.realityDataSourceProviders]].
+ * @beta
+ */
+export class RealityDataSourceGP3DTProvider implements RealityDataSourceProvider {
```
pmconne: Rename it "Google3dTilesProvider" for improved brevity and clarity.

core/frontend/src/RealityDataSource.ts
```diff
@@ -265,3 +285,87 @@ export class RealityDataSourceProviderRegistry {
     return this._providers.get(name);
   }
 }
+
+/**
+ * Options for creating a Google Photorealistic 3D Tiles (GP3DT) reality data source provider.
+ * @beta
+ */
+export interface RealityDataSourceGP3DTProviderOptions {
```
pmconne: The caller needs to provide *either* an API key *or* a `getAuthToken` function - never both and never neither - right?
Otherwise, e.g., your provider will produce URLs with `?key=undefined`.
Can you adjust this interface to require exactly one to be specified?

core/frontend/src/RealityDataSource.ts
```diff
@@ -265,3 +285,87 @@ export class RealityDataSourceProviderRegistry {
...
+  /** Google Map Tiles API Key used to access GP3DT. */
+  apiKey?: string;
+  /** Function that returns an OAuth token for authenticating with GP3DT. This token is expected to not contain the "Bearer" prefix. */
+  getAuthToken?: () => Promise<string | undefined>;
```
pmconne: Why is the function permitted to return an undefined auth token and what do you do in that situation?

example-code/snippets/src/frontend/GooglePhotorealistic3DTiles.ts
```diff
@@ -0,0 +1,48 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { getGooglePhotorealistic3DTilesURL, IModelApp, RealityDataSourceGP3DTProvider, Viewport } from "@itwin/core-frontend";
+
+// __PUBLISH_EXTRACT_START__ GooglePhotorealistic3DTiles_providerApiKey
+export async function setUpGP3DTProviderWithApiKey(viewport: Viewport, apiKey: string) {
```
pmconne: It's good that you linked to these snippets from NextVersion.md but it's even more important/useful to link to them from the API docs.
Or, link to your GooglePhotorealistic3DTiles.md from the API docs, which includes these snippets.

core/frontend/src/RealityDataSourceGP3DTImpl.ts
```diff
@@ -0,0 +1,227 @@
...
+ * @beta
+ */
+export function getGooglePhotorealistic3DTilesURL() {
+  return "https://tile.googleapis.com/v1/3dtiles/root.json";
+}
+
+/** This class provides access to the reality data provider services.
+ * It encapsulates access to a reality data from the Google Photorealistic 3D Tiles service.
+ * A valid GP3DT authentication key must be configured for this provider to work (provide the key in the [[RealityDataSourceGP3DTImpl.createFromKey]] method).
+* @internal
```
pmconne: Move this to src/internal and export from cross-package.ts.

eringram: Moved this. Also moved the one `beta` function in this file, `getGoogle3dTilesUrl()`, to RealityDataSource.ts. `getGoogle3dTilesUrl()` isn't related to the `Impl` class, and with it removed this file could be moved to src/internal and nothing from it needs to be exported in cross-package.ts.

Side note, other files like RealityDataSourceCesiumIonAssetImpl.ts and RealityDataSourceContextShareImpl.ts are similar with only one `internal` class and could probably be moved in the future.

common/api/core-frontend.api.md
eringram: Good point about the search for "google". Renamed everything to use "Google3dTiles", and "GooglePhotorealistic3dTiles" only in documentation filenames.

core/frontend/src/GoogleMapsDecorator.ts
eringram: Moved

core/frontend/src/RealityDataSource.ts
eringram: Renamed this and related classes.

core/frontend/src/RealityDataSourceGP3DTImpl.ts
pmconne: Yeah, we introduced the src/internal directory and the cross-package barrels after many of these internal APIs already existed, haven't gotten around to cleaning all of them up yet.

eringram:
@pmconne By the way, we noticed a side effect here from your recent maximumScreenSpaceError-related PR. Previously, we set `usesGeometricError = true` for the Google 3D Tiles, but did not set the `maximumScreenSpaceError`. It used to default to 16 I think due to [this line in TileDrawArgs](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/tile/TileDrawArgs.ts#L279). After your PR it defaults to 1, so we have to set it explicitly.

This might not be an issue as it didn't show up in the reality data cases in ImageTests, and if you set `usesGeometricError = true` maybe you also *should* set max SSE, but it is a change in the default behavior for this specific case so we wanted to let you know.

Max SSE change was added in 1e74630, so I was able to confirm it still defaulted to 16 in 840bdfa.

pmconne:
> @pmconne By the way, we noticed a side effect here from your recent maximumScreenSpaceError-related PR. Previously, we set `usesGeometricError = true` for the Google 3D Tiles, but did not set the `maximumScreenSpaceError`. It used to default to 16 I think due to [this line in TileDrawArgs](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/tile/TileDrawArgs.ts#L279). After your PR it defaults to 1, so we have to set it explicitly.

Please fix it to default to 16 again.

example-code/snippets/src/frontend/GooglePhotorealistic3DTiles.ts
eringram: Added one of these snippets and linked to the learning markdown file from the Google3dTilesProvider docs.

core/frontend/src/RealityDataSource.ts
eringram: Yes that's correct, just changed this to a union type that requires either one of these but not both.

core/frontend/src/RealityDataSource.ts
eringram: This function is allowed to return undefined because of the possibility that if a user calls an external service within their `getAuthToken`, that service could fail for some reason and not produce a token. If they are relying on an external service for the token, I'm not sure what they would do if their request for it fails but they must return a string - I guess then they would be responsible for their own error handling. Does that sound preferable?

Currently if it returns an undefined token, no authorization header is added to the request for Google 3D Tiles, so the tileset.json request fails with a 400 status, but no error is thrown in the code.

pmconne:
@aruniverse please revisit your comments.

core/frontend/package.json
```diff
@@ -82,7 +82,9 @@
     "vitest": "^3.0.6",
     "vite-multiple-assets": "^1.3.1",
     "vite-plugin-static-copy": "2.2.0",
-    "webpack": "^5.97.1"
+    "webpack": "^5.97.1",
+    "sinon": "^17.0.2",
+    "@types/sinon": "^17.0.2"
```
aruniverse: alphabetize

core/frontend/src/RealityDataSource.ts
```diff
@@ -265,3 +278,104 @@ export class RealityDataSourceProviderRegistry {
...
+ * Options for creating a Google Photorealistic 3D Tiles reality data source provider.
+ * The caller must provide either an API key or a function that returns an auth token.
+ * @beta
+ */
+export type Google3dTilesProviderOptions = {
+  /** Google Map Tiles API Key used to access Google 3D Tiles. */
+  apiKey: string;
+  getAuthToken?: never;
+  /** If true, the data attributions/copyrights from the Google 3D Tiles will be displayed on screen. The Google Maps logo will always be displayed. Defaults to `true`. */
+  showCreditsOnScreen?: boolean
```
aruniverse, line 292: isnt allowing users to opt out of this a liability for us? 
when would this be ok?

aruniverse (approved):
great work

core/frontend/src/RealityDataSource.ts
markschlosseratbentley: We followed the pattern discussed in this comment: https://github.com/iTwin/itwinjs-core/issues/8036#issuecomment-2895776072
Tagging @ggetz as well.

core/frontend/package.json
eringram: Done

core/frontend/src/RealityDataSource.ts
markschlosseratbentley: @ggetz Any feedback / thoughts if a user should be able to disable this feature? (`showCreditsOnScreen`)

aruniverse: Lets go ahead and remove this and always show credits. If a user requests to turn this off later we can add this flag back, but I imagine that will be unlikely

ggetz: @aruniverse I don't think there is any harm in exposing this option. The default here is more conservative than what we have in CesiumJS. The opt-in behavior is more or less equivalent to the default CesiumJS behavior.

aruniverse: My only hesitation comes from the legal side, but as long as our bases are covered I'm ok. Resolving.

PR #8105, opened 2025-05-15
Refresh schedule-affected tiles without full tree reload

danielzhong-bentley:
https://github.com/orgs/iTwin/projects/37/views/26?sliceBy%5Bvalue%5D=danielzhong-bentley&pane=issue&itemId=106823705&issue=iTwin%7Citwin-graphics-backlog%7C653

Previously, when a Schedule Script changed, it would trigger a **full tile tree reload** by creating a new tile tree owner.

This PR attempts to **refresh only the affected tiles**, avoiding the need to reload the entire tile tree.

## Before:
`vp.displayStyle.scheduleScript = this._script;`

## After:
Start/Update editing mode:
`vp.displayStyle.setScheduleEditing(this._script);`
Finish and exit editing mode, everything refresh once:
`vp.displayStyle.commitScheduleEditing();`

- [x] Image test
- [x] Unit test
- [x] Clean up temp testing API

Example:
```
// Build initial schedule script
const script = new RenderSchedule.ScriptBuilder();
const timeline = script.addModelTimeline("0x123"); // modelId
const elem = timeline.addElementTimeline(["0xabc"]); // elementId

// First color change: user starts editing
elem.addColor(Date.now(), new RgbColor(255, 0, 0));
vp.displayStyle.setScheduleEditing(RenderSchedule.Script.fromJSON(script.finish()));
// 👆 Enters editing mode — uses dynamic tiles for quick preview, no full tile tree refresh

// Second change: update color again during edit
elem.addColor(Date.now() + 3000, new RgbColor(0, 255, 0));
vp.displayStyle.setScheduleEditing(RenderSchedule.Script.fromJSON(script.finish()));
// 👆 Still editing — dynamic tile updates continue without full reload

// Done editing: commit to apply changes and refresh whole tile tree once
vp.displayStyle.commitScheduleEditing();
// 👆 Exits editing mode, triggers full refresh to apply all changes cleanly
```

* Please check the backlog for more details and video result
https://github.com/orgs/iTwin/projects/37/views/26?sliceBy%5Bvalue%5D=danielzhong-bentley&pane=issue&itemId=106823705&issue=iTwin%7Citwin-graphics-backlog%7C653


danielzhong-bentley:
Please share any feedback. This is a very early version, and I want to make sure I'm heading in the right direction.

pmconne:
> A new helper method isAffectedByScheduleScript() is added to the IModelTile class. It checks whether the tile is affected by a ScheduleScript, based on matching the model ID and element ID against the script's defined timelines.

> When the tile content is detected affected, it will "refresh".

To be clear, this will not solve the problem. The use case involves a schedule script that affects most of the elements. The user is interactively, incrementally adjusting the script by e.g. adding individual elements to it. The approach you're describing will effectively refresh every tile after every little change, just like it currently does, because just about every tile is going to include at least one element that is affected by the script. What the user wants is for only those tiles that are affected by a **change** to the script to be refreshed. Do you have a plan for addressing that problem?

danielzhong-bentley:
> > A new helper method isAffectedByScheduleScript() is added to the IModelTile class. It checks whether the tile is affected by a ScheduleScript, based on matching the model ID and element ID against the script's defined timelines.
> 
> > When the tile content is detected affected, it will "refresh".
> 
> To be clear, this will not solve the problem. The use case involves a schedule script that affects most of the elements. The user is interactively, incrementally adjusting the script by e.g. adding individual elements to it. The approach you're describing will effectively refresh every tile after every little change, just like it currently does, because just about every tile is going to include at least one element that is affected by the script. What the user wants is for only those tiles that are affected by a **change** to the script to be refreshed. Do you have a plan for addressing that problem?

Got it. I'm planning to compare the previous and current schedule scripts to compute a delta of changed element IDs. This will allow us to identify only the elements that were added or removed from the script.

To support this, I’ll update the `pruneAnimatedTiles` method to accept a Set<Id64String> representing the changed element IDs. During traversal, the method will skip tiles that don't contain any of the affected elements, so we only refresh tiles that are actually impacted by the script changes.

danielzhong-bentley:
Updated, the tiles refresh only when their corresponding elements change (e.g., visibility, color, transform). The new logic detects which elements in the schedule script need updates and refreshes only their tiles.

danielzhong-bentley:
I’ve tested the model provided by the Synchro 4D team. It’s essentially the same as the standard v1 tile .bim file, but with more models in the scene.

The main issue with the current approach is that, even though the logic is implemented at the tile class level, each model’s tile tree typically has only one large root tile. 

If a model had two or more root tiles, optimization would be significantly improved. For example, if an element on the right side changes, only the right-side tiles would need to refresh. If both schedule scripts remain unchanged, then no refresh would be needed at all.

However, because most models only have a single large root tile, detecting a change at the root level ends up same as triggering a refresh of the entire tile tree/tree owner. All models within that tile are refreshed unnecessarily.

I'm trying to implement a method that detects only the current camera level, so the root tile is refreshed only when zooming out. If the camera isn't zoomed out, only the child tiles at the current level would be refreshed.

However, if the camera is positioned far away, the same issue mentioned earlier still occurs — the root tile ends up being refreshed, causing the everything reload unnecessarily(if the tile tree only has 1 root tile).

gedkek:
> I'm trying to implement a method that detects only the current camera level, so the root tile is refreshed only when zooming out. If the camera isn't zoomed out, only the child tiles at the current level would be refreshed.

So just to be clear, all tiles on screen get refreshed anyway because of single root? If zoomed in, the entire tree won't get refreshed, but tiles on current zoom level (all tiles on screen) would refresh then, and if zoomed out, then same effect basically, but since entire tree is visible on screen, so root and all children get refreshed?

If I understand correctly then, while a potential gain in performance when zoomed in, would there be visual difference to flickering (loading tiles) while refreshing those tiles?

pmconne:
> If I understand correctly then, while a potential gain in performance when zoomed in, would there be visual difference to flickering (loading tiles) while refreshing those tiles?

It could be made to limit visible flickering/refresh but only to a point. With this approach, any tile that does contain at least one element affected by a change to the timeline is going to have to update the content for all of the elements it contains.

Your problem is similar to that faced by authoring apps: the user is making interactive, incremental changes that affect individual elements' visualization and want to see the visualization update in real time. Authoring apps achieve that using [GraphicalEditingScope](https://www.itwinjs.org/reference/core-frontend/imodelconnection/graphicaleditingscope/). I think it would be productive to revisit a previous suggestion I made:

1. Permit the `timeline` on the display style and tile tree Id  to be an object representing a "timeline under construction" that emits notifications when the contents of the timeline change.
2. Leverage [dynamic tiles](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/internal/tile/DynamicIModelTile.ts) to hide elements affected by timeline changes in the tiles and draw them using temporary graphics instead.
3. Ensure those graphics obey the timeline (transform, clip, color, visibility).
4. Permit the application to notify us when modifications to the timeline are complete, so we can discard the temporary graphics and repair the tiles. This can be important for performance when the user accumulates changes to a very large number of elements.

How much implementation could be shared between this approach and `GraphicalEditingScope` is TBD. `GraphicalEditingScope` involves coordination between backend and frontend whereas timeline modification is frontend-only and doesn't require a writable iModel. You could imagine a scenario where an app is both modifying elements and modifying the timeline, where the two would need to cooperate, but I don't know if @gedkek has that use case.

danielzhong-bentley:
Ready for another round of feedback.

Currently, the implementation follows the idea described above:
The tile tree owner creates a new tile tree only after the timeline editing is finished. While editing, it transitions to the Dynamic tile and hides the static tile.

It works perfectly for the script contains transform. However, there’s a problem when changes are purely non-transformational (such as color, visibility, or cutting plane changes). In these cases, transitioning to the Dynamic tile hides the static tile, but nothing appears in its place.

As a temporary workaround, I simply manually added this line for schedule script:
`elementTimeline1.addTransform(now, Transform.createIdentity());`

danielzhong-bentley:
Ready to Review. Everything is working as intended, including both transform and non-transform changes. Looking for feedback on how to refactor the code better

danielzhong-bentley:
# Editing Mode with color chage:

https://github.com/user-attachments/assets/a0571c65-9d46-4a39-8410-a8ba89e043ab

# Editing Mode with Transform change:

https://github.com/user-attachments/assets/67c40a43-70fb-41c2-ae06-9a69ec67fba3

# Finish editing (everything refresh once for final)

https://github.com/user-attachments/assets/d044d9c7-8149-449d-8073-30e8abfcaaa1

pmconne:
> Everything is working as intended, including both transform and non-transform changes. Looking for feedback on how to refactor the code better

Number one priority: add unit tests to **prove** everything works as intended, today and into the future.

danielzhong-bentley:
Added unit tests and confirmed image test is working. Will extract the API after the review, because temporary APIs were added to simplify testing the schedule script, they will be removed.

pmconne:
Please make the builds pass so we can verify tests etc.

danielzhong-bentley:
> Does this work with both V1 and V4 tiles?

Not yet, it currently only supports v1 tiles.

full-stack-tests/core/src/frontend/hub/ScheduleScript.test.ts
```diff
@@ -282,4 +282,72 @@ describe("Schedule script (#integration)", () => {
     await style.load();
     expect(style.scheduleScript).not.to.be.undefined;
   });
+
+  // it("sets schedule script in editing mode without triggering tile tree refresh", async () => {
```
pmconne: Why are these tests commented out?

pmconne (commented):
Mark all new APIs `@beta`.

pmconne (commented):
Please provide a code snippet (in example-code-snippets) illustrating how someone would use this API to make multiple incremental changes to a schedule script and then commit the finished script.

full-stack-tests/core/src/frontend/hub/ScheduleScript.test.ts
danielzhong-bentley: The iTwin.js (Build Linux_Node_22) tests weren't passing earlier, possibly due to a timeout issue. I commented them out temporarily for testing purposes. I’ll reopen the PR shortly.

core/frontend/src/ScriptUtils.ts
```diff
@@ -0,0 +1,135 @@
+import { Id64String } from "@itwin/core-bentley";
+import { RenderSchedule } from "@itwin/core-common";
+
+/** @beta */
+export function getScriptDelta(prev: RenderSchedule.Script | undefined | null, next: RenderSchedule.Script): Set<Id64String> {
```
pmconne: Is this function useful to anybody outside the core-frontend package? If so, you need to export it from the core-frontend.ts barrel file. If (as I suspect) not, you should move this file to src/internal and remove the release tag.

core/frontend/src/DisplayStyleState.ts
```diff
@@ -63,6 +64,16 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
 
   private _scriptReference?: RenderSchedule.ScriptReference;
 
+  /** Event raised when schedule script edits are made, providing changed element IDs and the editing scope. */
+  public readonly onScheduleEditingChanged = new BeEvent<
```
pmconne: This event's payload doesn't convey enough information to be useful. For example, it doesn't tell us which models are affected, so your event listener ends up invalidating every single TileTree instead of just the one(s) affected by a given change.

core/frontend/src/ScriptUtils.ts
danielzhong-bentley: Got it, moved to src/internal

core/frontend/src/internal/ScriptUtils.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      // Check Transform
+      const tf1 = a.getAnimationTransform(t);
+      const tf2 = b.getAnimationTransform(t);
+      if (!tf1.isAlmostEqual(tf2)) return false;
+
+      // Check Color
+      const c1 = a.getColor(t);
+      const c2 = b.getColor(t);
+      if ((c1 && !c2) || (!c1 && c2) || (c1 && c2 && !c1.equals(c2))) return false;
```
eringram: Currently this continues and doesn't return false if both `c1` and `c2` are undefined, is that intended?

core/frontend/src/DisplayStyleState.ts
danielzhong-bentley: Thanks for the suggestion, added.

danielzhong-bentley:
> Please provide a code snippet (in example-code-snippets) illustrating how someone would use this API to make multiple incremental changes to a schedule script and then commit the finished script.

Updated the description to include code snippets.

core/frontend/src/DisplayStyleState.ts
pmconne: > `{ changedElementIds: Set<Id64String>, affectedModelIds: Set<Id64String> }`

This doesn't provide a way to correlate affected elements with models. Consider something like `Array<{ timeline: ModelTimeline, elements: Id64Set }>`. Then change your event listeners to only apply to tile trees associated with the affected model(s).

core/frontend/src/internal/ScriptUtils.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      // Cutting Plane (CLIPPING)
+      const clip1 = a.getCuttingPlane?.(t);
+      const clip2 = b.getCuttingPlane?.(t);
+
+      const bothUndefined = clip1 === undefined && clip2 === undefined;
+      const oneUndefined = (clip1 === undefined) !== (clip2 === undefined);
+
+      if (oneUndefined) return false;
+      if (!bothUndefined && clip1 && clip2) {
```
eringram:
```suggestion
      if (clip1 && clip2) {
```
Pretty sure you can simplify this condition by removing the `bothUndefined` variable entirely

core/frontend/src/DisplayStyleState.ts
danielzhong-bentley: Please review my updates to confirm if I understood you correctly. Thank you!

core/frontend/src/internal/ScriptUtils.ts
```diff
@@ -0,0 +1,135 @@
...
+      // Check Color
+      const c1 = a.getColor(t);
+      const c2 = b.getColor(t);
+      if ((c1 && !c2) || (!c1 && c2) || (c1 && c2 && !c1.equals(c2))) return false;
+
+      // Cutting Plane (CLIPPING)
+      const clip1 = a.getCuttingPlane?.(t);
+      const clip2 = b.getCuttingPlane?.(t);
+
+      const bothUndefined = clip1 === undefined && clip2 === undefined;
```
eringram:
```suggestion
```
With the if statement changed, this is not used anywhere else

core/frontend/src/internal/ScriptUtils.ts
danielzhong-bentley: Fixed, thanks!

eringram: I don't think your latest edit changes the behavior if both are undefined -- just so I understand better, what should happen if both are undefined? That means they are equal, so it should continue right?

danielzhong-bentley: Yes, that's exactly the intended behavior, if both c1 and c2 are undefined, we consider them equal and continue.

common/api/core-frontend.api.md
```diff
@@ -2648,6 +2649,11 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
     moveMapLayerToTop(mapLayerIndex: MapLayerIndex): void;
     get name(): string;
     readonly onOSMBuildingDisplayChanged: BeEvent<(osmBuildingDisplayEnabled: boolean) => void>;
+    readonly onScheduleEditingChanged: BeEvent<(change: Array<{
```
pmconne: Mark all new APIs beta.

One major reason is that I'd prefer in the future to see an interface that represents a "schedule under construction" so that people can modify a potentially very large schedule without having to copy it and without you having to exhaustively analyze it to discover what element(s) changed. Instead it would keep track internally as changes are made.

core/common/src/RenderSchedule.ts
```diff
@@ -568,6 +568,8 @@ export namespace RenderSchedule {
     public readonly cuttingPlane?: TimelineEntryList<CuttingPlaneEntry, CuttingPlaneEntryProps, CuttingPlane | undefined>;
     /** The total time period represented by this timeline. */
     public readonly duration: Range1d;
+    /** Indicates whether the schedule editing session has been finalized and is no longer active. */
+    public isEditingCommitted: boolean = false;
```
pmconne, line 575: The public API shouldn't permit people to modify this property. It's weird anyway that every timeline has a separate `isEditingCommitted` property when really it's a property of the entire script. I suggest making it `@internal`.

core/frontend/src/Viewport.ts
```diff
@@ -1304,6 +1304,32 @@ export abstract class Viewport implements Disposable, TileUser {
...
     removals.push(style.onScheduleScriptChanged.addListener(scriptChanged));
 
+
+    const scheduleEditingChanged = (
+      changes: Array<{ timeline: RenderSchedule.ModelTimeline, elements: Set<Id64String> }>
+    ) => {
+      for (const { timeline, elements } of changes) {
+        for (const ref of this.getTileTreeRefs()) {
+          const tree = ref.treeOwner.tileTree;
+          if (tree instanceof IModelTileTree && tree.modelId === timeline.modelId) {
```
pmconne: Add this type to `RenderSchedule` namespace:
```
export interface EditingChanges {
  timeline: RenderSchedule.ModelTimeline;
  elements: Set<Id64String>;
}
```
Use that instead of inlining the type everywhere.

Define new `@internal` method `TileTree.onScheduleEditingChanged(changes: RenderSchedule.EditingChanges[]): void` that does nothing by default. Call it here instead of looping and casting. Override it in `IModelTileTree`.

Then you will have a path forward to make V4 tiles support this workflow.

core/frontend/src/Viewport.ts
```diff
@@ -1304,6 +1304,32 @@ export abstract class Viewport implements Disposable, TileUser {
...
+            tree.onScheduleEditingChanged({ changedElementIds: elements });
+          }
+        }
+      }
+    };
+
+    const scheduleEditingCommitted = () => {
+      for (const ref of this.getTileTreeRefs()) {
+        const tree = ref.treeOwner.tileTree;
+        if (tree instanceof IModelTileTree) {
```
pmconne: Similar advice as for `scheduleEditingChanged` - new @internal method `TileTree.onScheduleEditingCommitted`.

pmconne:
Have you asked the people who want this enhancement whether or not they are using or plan to use V4 tiles?

danielzhong-bentley:
> Have you asked the people who want this enhancement whether or not they are using or plan to use V4 tiles?

They weren’t familiar with the V1 or V4 tiles the last time I reached out, but I’ll double check with them.

core/frontend/src/DisplayStyleState.ts
```diff
@@ -63,6 +64,18 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
 
   private _scriptReference?: RenderSchedule.ScriptReference;
 
+  /** Event raised when schedule script edits are made, providing changed element IDs and the editing scope. */
+  /** @beta */
```
pmconne:
```suggestion
  /** Event raised when schedule script edits are made, providing changed element IDs and the editing scope.
   * @beta
   */
```

core/frontend/src/DisplayStyleState.ts
```diff
@@ -63,6 +64,18 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
...
+  public readonly onScheduleEditingChanged = new BeEvent<
+      (change: Array<{ timeline: RenderSchedule.ModelTimeline, elements: Set<Id64String> }>) => void
+  >();
+
+  /** Event raised when schedule script edits are committed (finalized). */
+  /** @beta */
```
pmconne:
```suggestion
  /** Event raised when schedule script edits are committed (finalized).
   * @beta
   */
```

core/common/src/RenderSchedule.ts
```diff
@@ -568,6 +568,9 @@ export namespace RenderSchedule {
     public readonly cuttingPlane?: TimelineEntryList<CuttingPlaneEntry, CuttingPlaneEntryProps, CuttingPlane | undefined>;
     /** The total time period represented by this timeline. */
     public readonly duration: Range1d;
+    /** Indicates whether the schedule editing session has been finalized and is no longer active. */
+    /** @internal */
```
pmconne:
```suggestion
    /** Indicates whether the schedule editing session has been finalized and is no longer active.
     * @internal
     */

```

core/frontend/src/Viewport.ts
danielzhong-bentley: Added, thanks!

core/frontend/src/DisplayStyleState.ts
```diff
@@ -63,6 +64,20 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
 
   private _scriptReference?: RenderSchedule.ScriptReference;
 
+  /** Event raised when schedule script edits are made, providing changed element IDs and the editing scope.
+   * @beta
+   */
+  public readonly onScheduleEditingChanged = new BeEvent<
+      (change: Array<{ timeline: RenderSchedule.ModelTimeline, elements: Set<Id64String> }>) => void
+  >();
```
pmconne:
```suggestion
  public readonly onScheduleEditingChanged = new BeEvent<(changes: RenderSchedule.EditingChanges[]) => void>();
```

core/frontend/src/DisplayStyleState.ts
```diff
@@ -295,6 +310,44 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
...
+
+    this.scheduleScript = newScript;
+    this.onScheduleEditingChanged.raiseEvent(changes);
+
+    for (const modelTimeline of this.scheduleScript.modelTimelines) {
+      modelTimeline.isEditingCommitted = false;
+    }
+  }
+
+  /** Commit the editing session. */
```
pmconne:
```suggestion
  /** Commit the editing session.
   * @beta
   */
```

core/frontend/src/DisplayStyleState.ts
```diff
@@ -295,6 +310,44 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
       await this._queryRenderTimelinePropsPromise;
   }
 
+  /** Start/update the script in the current editing session. Fires onScheduleEditingChanged. */
```
pmconne:
```suggestion
  /** Start/update the script in the current editing session. Fires onScheduleEditingChanged.
   * @beta
   */
```

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -457,4 +472,46 @@ export class IModelTileTree extends TileTree {
   public get containsTransformNodes(): boolean {
     return undefined !== this._transformNodeRanges;
   }
+
+  public override onScheduleEditingChanged(changes: RenderSchedule.EditingChanges[]): void {
+    const displayStyle = IModelApp.viewManager.selectedView?.displayStyle;
```
pmconne, line 477: Search `changes` for an entry with a `ModelTimeline` with the same model Id as this tile tree. Then only process the elements in that entry. The rest of the changes are not relevant for this tile tree.

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -457,4 +472,46 @@ export class IModelTileTree extends TileTree {
...
+      this._options.timeline = scriptInfo.timeline;
+    }
+
+    const changedElements: ElementGeometryChange[] = [];
+    for (const change of changes) {
+      for (const id of change.elements) {
+        changedElements.push({
+          id,
+          type: DbOpcode.Update,
+          range: new Range3d(),
```
pmconne: Null range seems bad, culling won't work properly.

core/frontend/src/DisplayStyleState.ts
pmconne: Expand your documentation comment to:
- explain what a script editing session is.
- link to the example code snippet you added to your PR description, which you should put in the example-code-snippets package.
- specify that the user must invoke `commitScheduleEditing` when finished.
- warn that you cannot currently use this API at the same time as a GraphicalEditingScope.

core/frontend/src/internal/tile/IModelTileTree.ts
danielzhong-bentley: I'm struggling to determine the appropriate range to use here. I initially tried `this._rootTile.range.clone()`, but that’s clearly incorrect. I'm not sure what the correct range should be in this context. From what I can tell, `DynamicModelTile` typically relies on the `ElementGeometryChange` type, but it seems I don’t have access to instances of `ElementGeometryChange`. I tried hardcoding a range, and while it works, it doesn’t feel like the right solution. I’ll continue investigating a proper way to handle this.

pmconne: The correct range is the bounding box of the element. You'd have to query the backend to obtain that.

core/frontend/src/DisplayStyleState.ts
danielzhong-bentley: Expanded, thanks for the suggestion!

core/frontend/src/DisplayStyleState.ts
```diff
@@ -295,6 +308,72 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
...
+    *
+    * Calling this method multiple times will update the current editing session with new script changes.
+    * When all edits are complete, you must invoke @see [[commitScheduleEditing]] to finalize the session and
+    * trigger a full tile tree refresh with the committed script.
+    *
+    * **Note:** You cannot use schedule script editing while a @see [[GraphicalEditingScope]] is active.
+    * Example:
+    * ```ts
+    * [[include:ScheduleScript_setEditingMode]]
+    * ```
```
pmconne:
```suggestion
    * Begins or updates a schedule script editing session for the current display style.
    * During an editing session, changes to the schedule script are applied incrementally
    * using temporary dynamic tiles, allowing for interactive preview of visual changes like color,
    * transforms, visibility, and cutting planes — without requiring a full tile tree reload.
    *
    * Calling this method multiple times will update the current editing session with new script changes.
    * When all edits are complete, you must invoke [[commitScheduleEditing]] to finalize the session and
    * trigger a full tile tree refresh with the committed script.
    *
    * @note You cannot use schedule script editing while a @see [[GraphicalEditingScope]] is active.
    *
    * Example:
    * ```ts
    * [[include:ScheduleScript_setEditingMode]]
    * ```
```

docs/changehistory/NextVersion.md
```diff
@@ -40,3 +40,11 @@ iTwin.js now supports displaying Google Photorealistic 3D Tiles via the new `Goo
...
 - The `PresentationManagerProps.schemaContextProvider` property has been deprecated. Starting with `5.0` release, `SchemaContext` is always available on [IModelConnection]($core-frontend), so this prop is no longer needed. If supplied, it will still be preferred over the iModel's schema context, until the property is removed completely in a future release.
 - The `PresentationManagerProps.defaultFormats` property has been deprecated in favor of the `FormatsProvider` now being available on [IModelApp.formatsProvider]($core-frontend).
+
+## Schedule Script Editing Mode
+
+A new editing mode has been added to `DisplayStyle3dState` for working with `RenderSchedule.Script`. This allows dynamic editing of schedule scripts in real time, with immediate visual feedback using dynamic tiles, without requiring a full tile tree refresh for every change.
+
+Use `setScheduleEditing(script)` to apply a schedule script in editing mode. Once all edits are complete, call `commitScheduleEditing()` to finalize the changes and trigger a full refresh.
+
+See the [Schedule Script Editing example](../../example-code/snippets/src/frontend/ScheduleScriptEditing.ts) for more details.
```
pmconne:
```suggestion
[DisplayStyle3dState]($frontend) now offers an interactive editing mode for [RenderSchedule.Script]($common)s, allowing interactive, dynamic modification of the style's schedule script with immediate visual feedback. Previously, every modification of the script would trigger a refresh of all of the graphics; the new mode instead draws temporary graphics just for the elements affected by the script changes.

Use [DisplayStyle3dState.setScheduleEditing]($frontend) to enter this mode, or to update the script again after entering the mode. Once all edits are complete, call [DisplayStyle3dState.commitScheduleEditing]($frontend) to finalize the changes a trigger a full refresh of the graphics. Example:

```ts
[[include:ScheduleScript_editingMode]]
```
```

File: example-code/snippets/src/frontend/ScheduleScriptEditing.ts [hunk 21]
```diff
@@ -0,0 +1,60 @@
...
+    direction: { x: 0, y: 0, z: 1 },
+  });
+  elementTimeline.addCuttingPlane(now + 3000, {
+    position: { x: 0, y: 0, z: 2 },
+    direction: { x: 0, y: 0, z: 1 },
+  });
+
+  // Commit changes — triggers full tile tree refresh to finalize edits
+  vp.displayStyle.commitScheduleEditing();
+}
```
pmconne:
```suggestion
/** A helper function that extracts a schedule script from a builder. */
function finishScript(builder: RenderSchedule.ScriptBuilder): RenderSchedule.Script {
  const json = builder.finish();
  const script = json ? RenderSchedule.Script.fromJSON(json) : undefined;
  if (!script) {
    throw new Error("invalid script!");
  }
  
  return script;
}

/** Applies color and visibility changes to a specified element using interactive
 * schedule script editing.
 */
export function editElementTimeline(vp: Viewport, elementId: Id64String, modelId: Id64String): void {
  const builder = new RenderSchedule.ScriptBuilder();
  const now = Date.now();

  const modelTimeline = builder.addModelTimeline(modelId);
  const elementTimeline = modelTimeline.addElementTimeline([elementId]);

  // Change color over time
  elementTimeline.addColor(now, new RgbColor(255, 0, 0));
  elementTimeline.addColor(now + 3000, new RgbColor(0, 255, 0));

  // Enter interactive script editing mode
  vp.displayStyle.setScheduleEditing(finishScript(builder));  

  // Change visibility over time
  elementTimeline.addVisibility(now, 0);       // fully transparent
  elementTimeline.addVisibility(now + 3000, 100); // fully visible

  // Update the script under construction
  vp.displayStyle.setScheduleEditing(finishScript(builder));
  
  // ...make some other changes to the script if desired...

  // Commit changes, triggering full graphical refresh.
  vp.displayStyle.commitScheduleEditing();
}
```

core/frontend/src/DisplayStyleState.ts
```diff
@@ -295,6 +308,73 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
...
+    *
+    * Calling this method multiple times will update the current editing session with new script changes.
+    * When all edits are complete, you must invoke [[commitScheduleEditing]] to finalize the session and
+    * trigger a full tile tree refresh with the committed script.
+    *
+    * @note You cannot use schedule script editing while a @see [[GraphicalEditingScope]] is active.
+    *
+    * Example:
+    * ```ts
+    * [[include:ScheduleScript_setEditingMode]]
```
pmconne:
```suggestion
    * [[include:ScheduleScript_editingMode]]
```

core/frontend/src/DisplayStyleState.ts
```diff
@@ -295,6 +308,73 @@ export abstract class DisplayStyleState extends ElementState implements DisplayS
...
+  /**
+  * Finalizes a script editing session previously started with [[setScheduleEditing]].
+  * This applies all pending script changes and triggers a full tile tree reload to reflect them in the viewport.
+  * After this call, the schedule script is considered committed and editing mode ends.
+  *
+  * @see [[setScheduleEditing]] to begin a schedule script editing session.
+  * Example:
+  * ```ts
+  * [[include:ScheduleScript_commitEditing]]
+  * ```
```
pmconne:
```suggestion
```

danielzhong-bentley:
> Have you asked the people who want this enhancement whether or not they are using or plan to use V4 tiles?

They mentioned they’re unsure about any plans to use V4 tiles and will go with whatever is set as the default.

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -457,4 +473,66 @@ export class IModelTileTree extends TileTree {
...
+
+    const relevantChange = changes.find((c) => c.timeline.modelId === this.modelId);
+    const changedElements: ElementGeometryChange[] = [];
+
+    if (relevantChange) {
+      for (const id of relevantChange.elements) {
+        changedElements.push({
+          id,
+          type: DbOpcode.Update,
+          range: await this.getElementRange(id),
```
pmconne: Instead of awaiting each element's range one at a time, create a list of all the relevant elements first. Then request all of their ranges in a single request. Finally, process the results to populate `changedElements`.

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -457,4 +473,66 @@ export class IModelTileTree extends TileTree {
   public get containsTransformNodes(): boolean {
     return undefined !== this._transformNodeRanges;
   }
+
+  private async getElementRange(elementId: Id64String): Promise<Range3d> {
+    const [element] = await this.iModel.elements.getProps(elementId) as GeometricElementProps[];
```
pmconne: You're requesting far more data than you actually need (every property of the element). Instead, execute an ECSql query to obtain just the bounding box. You can do that using `IModelConnection.Elements.getPlacements`. That also will allow you to obtain the placements for any number of elements in a single backend request.

core/frontend/src/internal/tile/IModelTileTree.ts
danielzhong-bentley: Fixed, thanks!

core/frontend/src/internal/tile/IModelTileTree.ts
danielzhong-bentley: Fixed, thanks!

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -513,21 +495,52 @@ export class IModelTileTree extends TileTree {
...
-    if (relevantChange) {
-      for (const id of relevantChange.elements) {
-        changedElements.push({
-          id,
-          type: DbOpcode.Update,
-          range: await this.getElementRange(id),
-        });
+    for (let i = 0; i < elementIds.length; i++) {
+      const id = elementIds[i];
+      const placement = placements[i];
```
pmconne: Look at the documentation of `getPlacements`. It doesn't guarantee the same number of entries or ordering thereof as the input array of element Ids. Each entry *has* an element Id. So you should be able to just do `for (const placementAndId of placements)` here and remove your whole try-catch thing above.

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -513,21 +495,52 @@ export class IModelTileTree extends TileTree {
...
+    if (!relevantChange || relevantChange.elements.size === 0)
+      return;
+
+    const elementIds = Array.from(relevantChange.elements);
+    if (elementIds.length === 0)
+      return;
+
+    let placements: (Placement2d | Placement3d | undefined)[] = [];
+    try {
+      placements = await this.iModel.elements.getPlacements(elementIds);
```
pmconne: Pass the type ("2d" or "3d") here based on whether the model is 2d or 3d (it's almost certainly 3d).

core/frontend/src/internal/tile/IModelTileTree.ts
```diff
@@ -513,21 +495,52 @@ export class IModelTileTree extends TileTree {
...
+        }
+      } catch {
+        range = Range3d.createNull();
       }
+
+      changedElements.push({
+        id,
+        type: DbOpcode.Update,
+        range,
+      });
```
pmconne: `placements` is already an array of `Placement`s (plus element Id), you don't need to do `fromJSON`. Something like this:

```suggestion
    changedElements = placements.map((x) => {
      return {
        id: x.elementId,
        type: DbOpcode.Update,
        rance: x.calculateRange(),
      };
    });
```

core/frontend/src/internal/tile/IModelTileTree.ts
pmconne:
```suggestion
      placements = await this.iModel.elements.getPlacements(elementIds, {
        type: this.is3d ? "3d" : "2d",
      });
```

core/frontend/src/internal/tile/IModelTileTree.ts
danielzhong-bentley: image

It will have this error

pmconne: I pushed this change, no such error. (There was a typo `rance` for `range` though).

danielzhong-bentley: That makes perfect sense, I was destructuring with `map(({ elementId, ...placement }) => ...)`, which ended up stripping away the class type, so placement became a plain object and lost access to methods like `calculateRange`. This was super helpful!

core/frontend/src/internal/tile/IModelTileTree.ts
danielzhong-bentley: Added

core/frontend/src/internal/tile/IModelTileTree.ts
danielzhong-bentley: Removed, thanks

pmconne (approved):

PR #8107, opened 2025-05-15
Fix blanking fill

### LINKED ISSUE #8101: Blanking fill draws behind other elements
**Describe the bug**
Blanking fill is supposed to make a filled planar shape draw behind other geometry belonging to the same element. The principal use case is for text background color. Instead, it can also draw behind geometry belonging to lots of other elements in the same view.

**To Reproduce**

1. Follow the steps in [this comment](https://github.com/iTwin/itwinjs-core/pull/8055#issuecomment-2884309037).
2. Observe the house geometry draws in front of the text background.

Image

**Expected behavior**
The text background should draw in front of the house geometry, but behind the text border and text itself.

**Additional context**

We [adjust the depth](https://github.com/iTwin/itwinjs-core/blob/f44eba5a86131cff3cf045d67d83052a0fa2fa51/core/frontend/src/internal/render/webgl/glsl/Surface.ts#L244) of blanking regions backward slightly to put it behind other geometry, but that puts it behind geometry belonging to any element that happens to have a similar non-adjusted depth.
We also draw both the text and the fill in the same render pass (opaque planar), which gives us no opportunity to discard the fill in favor of the text.

We should stop adjusting the depth, and just draw the blanking fill in the non-planar render pass. Then it will be discarded in favor of the element's other geometry.

pmconne:
Fixes #8101.
Dunno why we ever tried to do this using the old-fashioned polygon offset thing when we have fragment shader logic for controlling display priority.
Unit tests proved infeasible.

- [ ] Run ImageTests

### DIFF
--- a/common/changes/@itwin/core-frontend/pmc-fix-blanking-fill_2025-05-15-20-53.json
+++ b/common/changes/@itwin/core-frontend/pmc-fix-blanking-fill_2025-05-15-20-53.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Fix blanking regions inappropriately drawing behind unrelated geometry.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/core/frontend/src/internal/render/webgl/RenderCommands.ts
+++ b/core/frontend/src/internal/render/webgl/RenderCommands.ts
@@ -230,9 +230,9 @@ export class RenderCommands implements Iterable<DrawCommands> {
       switch (command.renderOrder) {
         case RenderOrder.PlanarLitSurface:
         case RenderOrder.PlanarUnlitSurface:
-        case RenderOrder.BlankingRegion:
           pass = "opaque-planar";
           break;
+        case RenderOrder.BlankingRegion: // draw behind other planar stuff
         case RenderOrder.LitSurface:
         case RenderOrder.UnlitSurface:
           pass = "opaque";
@@ -253,9 +253,9 @@ export class RenderCommands implements Iterable<DrawCommands> {
         switch (command.renderOrder) {
           case RenderOrder.PlanarLitSurface:
           case RenderOrder.PlanarUnlitSurface:
-          case RenderOrder.BlankingRegion:
             opaquePass = RenderPass.OpaquePlanar;
             break;
+          case RenderOrder.BlankingRegion: // draw behind other planar stuff
           case RenderOrder.LitSurface:
           case RenderOrder.UnlitSurface:
             opaquePass = RenderPass.OpaqueGeneral;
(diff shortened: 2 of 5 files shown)


pmconne:
Area patterns now z-fight with their background shapes. The shapes draw with blanking fill. We test for "same element" [in shader](https://github.com/iTwin/itwinjs-core/blob/cb21d3455affacb6e0aa8ea4c182700eb990ea7f/core/frontend/src/internal/render/webgl/glsl/FeatureSymbology.ts#L454) by comparing feature Ids. But feature Ids will differ because the pattern geometry is of GeometryClass.Pattern. Same would technically occur if e.g. blanking region was construction geometry with geometry of class primary intended to draw in front of it.

pmconne:
Fixing this requires the fragment shader to be able to answer the question "am I drawing the same element as the one that was previously written to the pick buffer?" This is only relevant for blanking fill - edges always have the exact same feature Id as their surfaces, so the current logic works; and planar regions draw on top of non-planar surfaces regardless of feature/element.

### Proposed solution

If a Batch contains any blanking regions, then we need to populate an extra texture that correlates feature index to "element index" (just some unique integer that is the same for all features belonging to the same element - e.g., the index of that element in a list of unique elements in the batch, or the index of the first feature associated with that element).
When drawing the linear and planar passes, the fragment shader looks up the element index and outputs it to the pick buffer.

When drawing the non-planar pass (in which planar blanking regions are now drawn), the fragment shader looks up the element index of the current feature and compares it to the one in the pick buffer. If and only if they match, the blanking region is discarded in favor of the previously-drawn fragment.

A couple of limitations:
- WebGL 2 guarantees a minimum of 16 texture units, no more. We recently allocated all 6 of the previously-free units for draping map layers onto iModel geometry. I propose we take one of those back to be used for the element index lookup texture.
- WebGL 2 guarantees a minimum of 4 fragment shader outputs. We recently allocated the 4th one for contour line information. It is extremely unlikely that blanking fill and contour lines will both be relevant for a given fragment, and in any case, contour lines recolor the fragment, making blanking fill unimportant. I propose that if a fragment is *not* a contour line, then we output the element index to the fragment shader output that otherwise would hold contour line info. We would need to ensure the contour buffer gets ping-ponged like the other pick buffer outputs.

@markschlosseratbentley I'm open to other ideas.

# Priority

I'm deferring implementation temporarily as we have more pressing problems in the area of drawing production. I suspect that in general users try to avoid having text with background fill overlap other geometry in their section drawings. @Josh-Schifter what is your take on the urgency?

Josh-Schifter:
> I suspect that in general users try to avoid having text with background fill overlap other geometry in their section drawings.

I suspect that users _try_ to avoid overlapping, but based on what I'm seeing from example site design sheets it is often unavoidable.  Here's some screenshots.

image

image

image

image

pmconne:
> I suspect that users try to avoid overlapping, but based on what I'm seeing from example site design sheets it is often unavoidable. Here's some screenshots.

Thanks for the examples. In that case I think @claudiaareneee should proceed with #8055 with the assumption that we will address the blanking fill issue separately. @markschlosseratbentley if no one else picks it up in the meanwhile I should be able to get back to it in a week or so.

pmconne:
A problem remains: raster text draws in the translucent pass, after the blanking regions have drawn in the opaque pass. Now that we no longer move blanking regions backward in view-Z, the transparent text may fail the depth test and be hidden by the blanking region.

Example:
image
If you zoom in far enough to make the text draw as polyfaces, it will draw in front of the blanking fill.

PR #8108, opened 2025-05-15
Rename "toggle reality bounds" key-in

eringram:
I noticed the [frontend-devtools readme](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend-devtools/README.md) documents the key-in `fdt toggle reality bounds`, when really the command was `fdt reality bounds`. This PR simply updates the command to match the documentation.

### DIFF
--- a/core/frontend-devtools/public/locales/en/FrontendDevTools.json
+++ b/core/frontend-devtools/public/locales/en/FrontendDevTools.json
@@ -82,7 +82,7 @@
       "keyin": "fdt visibility"
     },
     "ToggleRealityTileBounds": {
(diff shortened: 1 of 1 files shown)


pmconne:
Fix the README instead, so you don't break people who are using the key-in.

eringram:
> Fix the README instead, so you don't break people who are using the key-in.

Ok good point, I was thinking it would be better to fix the command to be more in line with the rest of the `fdt toggle ...` but people may be relying on the old command even though it's not captured by api contracts

pmconne (approved):

markschlosseratbentley (approved):

PR #8123, opened 2025-05-20
`RegionOps.constructAllXYRegionLoops` fails to compute outer boundary

dassaf4:
...due to numerical slop in the arc-segment outer chain. The solution is to explicitly look for the endpoint intersections in `CurveCurveIntersectXY.dispatchSegmentArc`.

This is for Booster, so probably needs to be backported to 5.0 branch.

Also for Booster (via iTwin Editing):
* `RegionOps.regionBooleanXY` now returns the region with simplest type. Previously it was always a `UnionRegion` though the return type is `AnyRegion` and callers should not have been assuming this. Now this method can return a `Loop`, a `ParityRegion`, or a `UnionRegion`. This simplification logic is encapsulated in the new helper method is `RegionOps.simplifyRegionType`.
* `AngleSweep.fractionToSignedPeriodicFraction` now takes `undefined` for input arg `toNegativeFraction` to convert an exterior sweep fraction to its period shift _closest_ to [0,1] in fractional space.

Other fixes:
* `Point3d.plus2Scaled` - fix non-breaking input type typo.


bbastings (approved):

core/geometry/src/test/topology/RegionOps.test.ts
```diff
@@ -1011,7 +1017,89 @@ describe("RegionOps", () => {
...
   });
+
+  it("RegionBooleanMerge", () => {
+    const ck = new Checker();
+    const allGeometry: GeometryQuery[] = [];
+    let isArray = false, hasLoopEntries = false;
+    const arrayConsistsOfLoops = (a: Array<any>) => a.every((value: any) => value instanceof Loop);
+    const regions = IModelJson.Reader.parse(JSON.parse(fs.readFileSync("./src/test/data/curve/areaBoolean/unionRegionWithOverlappingLoops.imjs", "utf8")));
+    if (ck.testDefined(regions, "read regions from file")) {
+      if ((isArray = Array.isArray(regions)) && ck.testTrue(isArray, "regions is an array")) {
```
saeeedtorabi, line 1028: What's the benefit of 
```ts
​ if ((isArray = Array.isArray(regions)) && ck.testTrue(isArray, "regions is an array")) {
```
over
```ts
isArray = Array.isArray(regions);
​if (ck.testTrue(isArray, "regions is an array")) {
```
I know it makes it 1 liner but it's a bit confusing as it looks `isArray = Array.isArray(regions)` is checking a condition while it's just assigning `isArray`

core/geometry/src/test/topology/RegionOps.test.ts
```diff
@@ -1011,7 +1017,89 @@ describe("RegionOps", () => {
...
+              GeometryCoreTestIO.captureCloneGeometry(allGeometry, merged);
+              if (ck.testType(merged, UnionRegion, "merge operation results in a UnionRegion")) {
+                if (ck.testExactNumber(3, merged.children.length, "merged region has 3 children")) {
+                  if ((hasLoopEntries = arrayConsistsOfLoops(merged.children)) && ck.testTrue(hasLoopEntries, "merged region children are Loops")) {
+                      let totalMergedArea = 0;
+                      for (const loop of merged.children) {
+                        const area = RegionOps.computeXYArea(loop);
+                        if (ck.testDefined(area, "area computed for loop")) {
+                          totalMergedArea += area;
+                          GeometryCoreTestIO.captureCloneGeometry(allGeometry, loop, 0, 0, 50);
```
saeeedtorabi, line 1042: Rather than stacking the 4 images on top of each other (via `dz`), isnt it easier to see if you stack them side by side (via `dx`)?

saeeedtorabi (approved):
I pulled your PR and reviewed it.

Src code changes are straightforward. I also checked tests.

Looks like no new test in `ClipUtilities.test.ts` and you just moved tests around.

I saw 2 new tests in `RegionOps.test.ts` and ran them. Added 2 optional comments.

core/geometry/src/test/topology/RegionOps.test.ts
dassaf4: Both statements are necessary because 1) I want this test to verify that `regions` is indeed an array, but 2) `ck.testTrue` swallows the `Array.isArray` type guard that suppresses Typescript complaining about using `regions.length` below.  I could split it into two lines, but `if (a = foo(b))` is a common idiom testing for truthiness of `a`, and core-geometry tests are rife with compound if statements that look like `if (ck.testFoo() && type_guard)` because most Checker tests aren't type guards. Ideally we should write a `ck.testArray` that serves as a type guard so all we'd need is `if (ck.testArray(regions, "regions is an array"))` and Typescript would understand that in the subsequent scope, `regions` is an `Array`.

core/geometry/src/test/topology/RegionOps.test.ts
dassaf4: I want to see them stacked on top of each other to xy operations.

dassaf4:

PR #8133, opened 2025-05-22
Adjust error message and add more details assertions to test

rschili:
imodel-native: https://github.com/iTwin/imodel-native/pull/1125

Today, when somebody tries to import a schema with changes but no version increment, we log an error saying "Schema import failed", but we actually continue and import the schema.

Except if it's the only schema being imported, in which case our DgnPlatform layer will strip it from the import and possibly skip the import altogether if no other schemas remain.

The behavior is very inconsistent. This current change is just to make the log message consistent and not suggest a fail when it actually procceeds.


ColinKerr (approved):

aruniverse (approved):

DanRod1999:

PR #8135, opened 2025-05-22
WIP: Move some test functionality into new test-support package

tcobbs-bentley:
The `HubMock` class is used outside itwinjs-core, even though it is marked as internal and not exported (requiring a deep import to access it). This PR creates a new `@itwin/test-support` package that contains `HubMock` in preparation for it being made public (or more likely beta in the short term). Classes that depend on `HubMock` were also put into `test-support`.

Unfortunately, removing these items from `core/backend/src/test` would result in failed core-backend unit tests, since those tests rely on this functionality, and `test-support` depends on `@itwin/core-backend`. I verified that everything outside `core-backend` is now using `test-support` by temporarily deleting `core/backend/src/test`.

One solution is to create a new `core-backend-tests` package and move everything from `core/backend/src/tests` into there, deleting the functionality that is now in `test-support`. `core-backend-tests` would then be built after `test-support` and could thus depend on it. I'm not actually sure if that's possible, or if some backend tests rely on being part of the backend package. But the current state, where everything in `test-support` is duplicated in `core/backend/src/test` is obviously not ok.

Further investigation will be performed to try to determine the best course of action.


core/backend-tests/README.md
```diff
@@ -0,0 +1,11 @@
+# @itwin/core-backend
```
ben-polinsky: Still contains core-backend readme file

ben-polinsky: Could you add a note in `core-backend` to write tests in this package?

tcobbs-bentley: Fixed; let me know if you think more is warranted.

pmconne:
Why does this require a new package versus simply exporting relevant APIs from core-backend?

tcobbs-bentley:
> Why does this require a new package versus simply exporting relevant APIs from core-backend?

I did this based on talks I had with Keith a few months ago, but I don't remember all of the exact details, so it could be that doing it this way is the wrong way to go.

The thought was that test-only functionality should be kept separate from core-backend, and in the future we will add things with (for example) mocha dependencies to the new package. Those would then not introduce mocha dependencies in core-backend. But looking at what is in there now, and looking at the package.json, I see that all it has are dev dependencies to chai, which already existed in core-backend. (As an aside, if this PR stands, then core-backend's dev dependencies to chai can be removed.)

pmconne:
I'll defer to others on the virtues of separate test packages. But if you do go that route, please rename the tools/test-support package to indicate it's only usable for **backend** tests - everything in it except for AdvancedEquals depends on core-backend.

tcobbs-bentley:
> But if you do go that route, please rename the tools/test-support package to indicate it's only usable for **backend** tests - everything in it except for AdvancedEquals depends on core-backend.

The hope was to move other non-backend test support functionality in there as well. That obviously hasn't been done yet, and I guess it's possible that it won't work. I can rename it to backend-test-support.

tcobbs-bentley:
I renamed the package to `@itwin/backend-test-support`. I talked to @wgoehrig, and he agreed that while this new package isn't an ideal solution, it's the best one available.

PR #8140, opened 2025-05-23
Added tangent API to curve collection

saeeedtorabi:
3 new APIs: emitTangents, allTangents, and closestTangent


dassaf4 (approved):
This is right on the border of "too much duplicated code", but I don't see immediately how to avoid it without introducing a circular reference.

Issue #8120, opened 2025-05-19
Window Select only selects what is visible


DRHoltman:
We currently have items with multiple elements that we are trying to convert and consolidate.
When we select multiple elements using the window select tool

![Image](https://github.com/user-attachments/assets/f7e83ea9-2ed1-4799-a833-2abe5db243f1)

It selects the items visible to the user

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the applicable information):**

- iTwin.js Version [e.g. 2.5.3]

**Additional context**

aruniverse:
Have you tried modifying the [Selection Scopes](https://www.itwinjs.org/presentation/unified-selection//#selection-scopes) ?

pmconne:
> Is this by design? Or is this a bug?

It is by design.

> Or is there another way to do this?

You could use a [spatial query](https://www.itwinjs.org/learning/spatialqueries/) to find elements that intersect your rectangle. @bbastings might have advice or examples.

bbastings:
Volume query [example](https://www.itwinjs.org/sandboxes/iTwinPlatform/Volume%20Query/)

> Have you tried modifying the [Selection Scopes](https://www.itwinjs.org/presentation/unified-selection//#selection-scopes) ?

Selection scope could help here if you are creating these elements as part of an assembly. Note: display-test-app does not support selection scope.

DRHoltman:
The screen shots above are using the element selection scope.
We will look into adding the spatial query as Paul mentioned above.
Thanks Guys.

markschlosseratbentley:
> @bbastings Does your response mean that spatial query will not work for them?

@markschlosseratbentley I wasn't able to figure out why my query didn't work in their file, it does in other files I tested (need query expert). The hope was that people that wanted volume selection could just set ToolSettings.enableVolumeSelection to true and the select tool along will all other tools that support drag box selection would just work.

markschlosseratbentley:
> I wasn't able to figure out why my query didn't work in their file, it does in other files I tested (need query expert).

@aruniverse do you know who is a query expert we can loop into this conversation?

aruniverse:
> > I wasn't able to figure out why my query didn't work in their file, it does in other files I tested (need query expert).
> 
> [@aruniverse](https://github.com/aruniverse) do you know who is a query expert we can loop into this conversation?

@khanaffan

khanaffan:
[...]

PR #8143, opened 2025-05-25
Support for volume selection.

bbastings:
Addresses this issue #8120 which has been brought up several times now.

Share crossing line and drag box selection code between SelectTool and ElementSetTool.
Breaking api change to SelectTool to make protected selectByPointsProcess and selectByPointsEnd methods async.
Promise calling public processSelection method was already being ignored which didn't allow it to be overridden properly.
Doubt many (if any) apps have overridden these SelectTool methods, best to break now before 5.0 is released.


bbastings:

PR #8158, opened 2025-05-28
Prevent extraneous hole during Polyface clip

dassaf4:
Failure to account for parallel, overlapping segments during exterior shard cleanup in `PolyfaceClip.clipPolyfaceUnionOfConvexClipPlaneSetToBuilders` occasionally caused an extraneous hole in the surviving mesh.

Also: various improvements for RegionOps methods.

### DIFF
--- a/common/changes/@itwin/core-backend/2025-06-02-15-46.json
+++ b/common/changes/@itwin/core-backend/2025-06-02-15-46.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -17,8 +17,8 @@ importers:
   ../../core/backend:
     dependencies:
       '@bentley/imodeljs-native':
-        specifier: 5.1.35
-        version: 5.1.35
+        specifier: 5.1.36
+        version: 5.1.36
       '@itwin/cloud-agnostic-core':
         specifier: ^2.2.4
         version: 2.2.4(inversify@6.0.1)(reflect-metadata@0.1.13)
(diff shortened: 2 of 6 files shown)


pmconne (approved):

markschlosseratbentley (approved):

core/geometry/src/curve/RegionOps.ts
```diff
@@ -345,6 +346,36 @@ export class RegionOps {
...
+  /**
+   * Validate the region's parent/child hierarchy in place:
+   * * Regions with exactly one child are simplified as per [[simplifyRegionType]].
+   * * Regions without children are removed.
+   * * No Boolean operations are performed.
+   * @param region region to validate in place
+   * @returns reference to the updated input region
+   * @see [[simplifyRegionType]]
+   */
+  public static validateRegion(region: AnyRegion): AnyRegion | undefined {
```
saeeedtorabi: when I see the name `validateRegion` I expect the function to validate a region and return a boolean. This function is kind of simplifying the region (not validating it) and returning a region. Isnt `simplifyRegion` a better name?

core/geometry/src/geometry3d/Range.ts
```diff
@@ -726,17 +726,45 @@ export class Range3d extends RangeBase implements LowAndHighXYZ, BeJSONFunctions
...
+   * @param other the other range.
+   * @param margin optional signed distance by which to expand/shrink `other` in all directions.
+   */
+  public intersectsRange(other: Range3d, margin?: number): boolean {
+    if (margin) {
+      return !(
+        this.low.x > other.high.x + margin
+        || this.low.y > other.high.y + margin
+        || this.low.z > other.high.z + margin
+        || other.low.x - margin > this.high.x
```
saeeedtorabi: how about moving `margin` to right for consistency?
```ts
        || other.low.x > this.high.x + margin
        || other.low.y > this.high.y + margin
        || other.low.z > this.high.z + margin

File: core/geometry/src/geometry3d/Range.ts [hunk 3]
```diff
@@ -726,17 +726,45 @@ export class Range3d extends RangeBase implements LowAndHighXYZ, BeJSONFunctions
...
+   * Test if there is any intersection with the other range, ignoring z.
+   * @param other the other range.
+   * @param margin optional signed distance by which to expand/shrink `other` in all xy-directions.
+   */
+  public intersectsRangeXY(other: Range3d, margin?: number): boolean {
+    if (margin) {
+      return !(
+        this.low.x > other.high.x + margin
+        || this.low.y > other.high.y + margin
+        || other.low.x - margin > this.high.x
```
saeeedtorabi: same

core/geometry/src/test/clipping/PolyfaceClip.test.ts
```diff
@@ -1502,4 +1502,39 @@ describe("PolyfaceClip", () => {
     GeometryCoreTestIO.saveGeometry(allGeometry, "PolyfaceClip", "DeckBuilder");
     expect(ck.getNumErrors()).toBe(0);
   });
+
+  it("ExtraneousHoleAfterClip", () => {
```
saeeedtorabi, line 1506: `ExtraneousHoleAfterClip` is not testing a new functionality so what bug fix is it testing? The extraneous hole during Polyface clip?

dassaf4: yes

core/geometry/src/curve/RegionOps.ts
dassaf4: Can't return `boolean` because output is not necessarily input. For example, for an input `UnionRegion` with only one `Loop`, the `Loop` is returned. I'll rename it.

core/geometry/src/test/clipping/PolyfaceClip.test.ts
dassaf4: You can see the hole by copying the test to your master branch and running it.  This will give you an idea of the kinds of problems that can result from clipping meshes.

saeeedtorabi: Yes, Indeed I ran it. It was a simple geom with a hole in it so I did not expect a bug there!

dassaf4: Maybe you ran it on the branch, which has the fix. If you copy and run the test on your master branch (without the rest of the PR changes), you should see a small quad hole near the expected hole.

Issue #8070, opened 2025-05-09
Documentation: KaTeX broken


ben-polinsky:
https://www.itwinjs.org/reference/core-geometry/cartesiangeometry/matrix3d/

<img width="452" alt="Image" src="https://github.com/user-attachments/assets/e53c112d-7c14-46fd-a2cd-860e269ab30c" />

PR #8166, opened 2025-05-30
Docs: adjust KaTeX example

ben-polinsky:
The latest [KaTeX](https://katex.org/) integration with our docs requires macros to be prefaced with `\\`.
Part of #8070


ben-polinsky:

tcobbs-bentley:
@ben-polinsky I realize that this is already merged, but is it expected that the `equation` example above the one you edited does not get KaTeX formatted?

image

ben-polinsky:
> Member



> @ben-polinsky I realize that this is already merged, but is it expected that the `equation` example above the one you edited does not get KaTeX formatted?
> 
> image

No, they both should be formatted. But this won't work in github md preview if that is where the image is from.

I now see [this](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions). It would make sense to be compatible with gh markdown if possible.

tcobbs-bentley:
> No, they both should be formatted. But this won't work in github md preview if that is where the image is from.
> 
> I now see [this](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions). It would make sense to be compatible with gh markdown if possible.

The image is from GitHub preview, and the one using `math` seems to be formatted (although probably not correctly, since it apparently doesn't look at the settings json).

ben-polinsky:
Yeah looks like `math` works but equation is our own impl.

I see this a low-priority, however. We rarely use this. We can circle back next quarter if/when we dedicate some time to revising our docs.

PR #8173, opened 2025-06-04
Fix: Add validation for bearing inputs including unexpected negative values

rushikesh-bentley:
Issue Link: https://github.com/iTwin/itwinjs-backlog/issues/1410

- Added formatter test for negative values like -45° and -270°

- Added parser test with valid and invalid bearing strings

- Covered valid cases like N-45:00:00E → 315°

- Rejected quadrant overbound inputs like N200E, S361:00:00E

### DIFF
--- a/common/changes/@itwin/presentation-backend/master_2025-06-06-07-55.json
+++ b/common/changes/@itwin/presentation-backend/master_2025-06-06-07-55.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-backend",
+      "comment": "Fix `getDisplayLabelDefinitions` implementation not taking into account different formats and casings of full class name.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/presentation-backend"
+}
--- a/common/changes/@itwin/presentation-common/master_2025-06-06-07-55.json
+++ b/common/changes/@itwin/presentation-common/master_2025-06-06-07-55.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-common",
+      "comment": "Fix `InstanceKey.compare` implementation not taking into account different formats and casings of full class name.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/presentation-common"
+}
--- a/common/config/rush/browser-approved-packages.json
+++ b/common/config/rush/browser-approved-packages.json
@@ -331,6 +331,10 @@
       "allowedCategories": [ "integration-testing", "internal" ]
     },
     {
+      "name": "@itwin/presentation-shared",
+      "allowedCategories": [ "backend", "common" ]
+    },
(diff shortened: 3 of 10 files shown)


core/quantity/src/Parser.ts
```diff
@@ -786,11 +787,18 @@ export class Parser {
...
     const revolution = this.getRevolution(spec);
-    magnitude = this.normalizeAngle(magnitude, revolution);
     const quarterRevolution = revolution / 4;
+
+    if (parsedResult.value > quarterRevolution) {
+      return { ok: false, error: ParseError.BearingAngleOutOfRange };
+    }
+
+    let magnitude = parsedResult.value;
+    if (magnitude < -quarterRevolution || magnitude > quarterRevolution) {
```
hl662, line 794: couldn't you have the left side of the OR in line 793? The RHS seems like it's already checked in line 793 as well

package.json
```diff
@@ -29,5 +29,6 @@
     "url": "http://www.bentley.com"
   },
   "//dependencies": "NOTE: dependencies will be empty here and instead specified in each package.",
-  "//devDependencies": "NOTE: devDependencies will be empty here and instead specified in each package."
-}
\ No newline at end of file
+  "//devDependencies": "NOTE: devDependencies will be empty here and instead specified in each package.",
+  "packageManager": "pnpm@10.10.0+sha512.d615db246fe70f25dcfea6d8d73dee782ce23e2245e3c4f6f888249fb568149318637dca73c2c5c8ef2a4ca0d5657fb9567188bfab47f566d1ee6ce987815c39"
```
hl662: can you revert your change to the root package.json?

core/quantity/src/test/BearingAzimuth.test.ts
```diff
@@ -218,6 +217,119 @@ describe("Bearing format tests:", () => {
...
+        ],
+      },
+    };
+
+    const bearingFormat = new Format("bearing-parser");
+    await bearingFormat.fromJSON(unitsProvider, bearingFormatProps);
+    const bearingParser = await ParserSpec.create(bearingFormat, unitsProvider, degree);
+    // Valid inputs
+    const validTestData = [
+      { input: "N45:00:00E", expected: 45.0 },
```
rschili, line 284: what happens if one of the other composite units is negative? Like N90:-44:-23E

rushikesh-bentley: I added similar inputs like "N90:-44:-23E", "S-90:-00:-01W", and "N-70:20:-5E" to the invalid test cases. They are correctly rejected as expected.

core/quantity/src/Parser.ts
rushikesh-bentley: I updated the logic to avoid the redundant check and moved the condition to line 793 as suggested.

rschili (approved):

hl662 (approved):

hl662:

PR #8191, opened 2025-06-09
Clear Caches on abandonChanges

MichaelSwigerAtBentley:
Clears the iModelDb caches on abandonChanges
Adds a unit test 

It was possible that a discarded element could be returned by the cache if a new element was later created with the same Id. By clearing the iModelDb Element and Models cache on abandonChanges, this is corrected.


aruniverse:

Rock2000 (approved):

PR #8195, opened 2025-06-09
Expand parsing flexibility for Bearing input strings.

### LINKED ISSUE #8175: Improve bearing and angle parsing/formatter to be more flexible
We want to improve our handling for bearing and angle inputs as requested by a first party product.

This should be implemented in the iTwin quantity package, but any changes made to the formatter should be made to the native quantity formatter as well as presentation still uses it in some scenarios. The parser changes are fine to only make in itwin

hl662:
Closes #8175

### DIFF
--- a/core/quantity/src/Parser.ts
+++ b/core/quantity/src/Parser.ts
@@ -283,15 +283,15 @@ export class Parser {
                 processingNumber = false;
                 wipToken = "";
               }
-             else if (wipToken.length > 0) {
-               if (signToken.length > 0) {
-                wipToken = signToken + wipToken;
-                signToken = "";
-               }
+              else if (format?.type === "Bearing" && wipToken.length > 0) {
+                if (signToken.length > 0) {
+                  wipToken = signToken + wipToken;
+                  signToken = "";
+                }
              tokens.push(new ParseToken(parseFloat(wipToken))); // Create token for the current number
              processingNumber = false;
              wipToken = "";
-            }
+              }
               continue;
             } else {
               // an "E" or "e" may signify scientific notation
--- a/core/quantity/src/Parser.ts
+++ b/core/quantity/src/Parser.ts
@@ -391,6 +391,22 @@ export class Parser {
         tokens.push(new ParseToken(wipToken));
       }
     }
+    if (tokens.length === 1 && typeof tokens[0].value === "number" && format.type === "Bearing") {
+      // Example: 65.4545 → 65 (degrees), 45 (minutes), 45 (seconds)
+      const value = tokens[0].value;
+      const degrees = Math.floor(value);
+      const remaining = (value - degrees) * 100;
+      const minutes = Math.floor(remaining);
+      const seconds = (remaining - minutes) * 100;
+      // Remove the original compact decimal token (e.g., 65.4545)
+      // before pushing split DMS tokens (degrees, minutes, seconds)
+      tokens.pop();
+      tokens.push(
+      new ParseToken(degrees),
+      new ParseToken(minutes),
+      new ParseToken(seconds)
+      );
+    }
 
     return tokens;
   }
--- a/core/quantity/src/Parser.ts
+++ b/core/quantity/src/Parser.ts
@@ -52,8 +52,8 @@ export interface ParsedQuantity {
 enum Operator {
   addition = "+",
   subtraction = "-",
-  multiplication = "*",
-  division = "/"
+  multiplication = "*", // unsupported but we recognize it during parsing
+  division = "/" // unsupported but we recognize it during parsing
 }
 
 function isOperator(char: number | string): boolean {
--- a/core/quantity/src/Parser.ts
+++ b/core/quantity/src/Parser.ts
@@ -285,7 +285,7 @@ export class Parser {
                 processingNumber = false;
                 wipToken = "";
               }
-              else if (format?.type === "Bearing" && wipToken.length > 0) {
+              else if (format.type === FormatType.Bearing && wipToken.length > 0) {
                 if (signToken.length > 0) {
                   wipToken = signToken + wipToken;
                   signToken = "";
--- a/core/quantity/src/Parser.ts
+++ b/core/quantity/src/Parser.ts
@@ -283,6 +283,15 @@ export class Parser {
                 processingNumber = false;
                 wipToken = "";
               }
+             else if (wipToken.length > 0) {
+               if (signToken.length > 0) {
+                wipToken = signToken + wipToken;
+                signToken = "";
+               }
+             tokens.push(new ParseToken(parseFloat(wipToken))); // Create token for the current number
+             processingNumber = false;
+             wipToken = "";
+            }
               continue;
             } else {
               // an "E" or "e" may signify scientific notation
--- a/core/quantity/src/Parser.ts
+++ b/core/quantity/src/Parser.ts
@@ -52,6 +52,8 @@ export interface ParsedQuantity {
 enum Operator {
   addition = "+",
   subtraction = "-",
+  multiplication = "*",
+  division = "/"
 }
 
 function isOperator(char: number | string): boolean {
--- a/core/quantity/src/Parser.ts
+++ b/core/quantity/src/Parser.ts
@@ -759,6 +759,17 @@ export class Parser {
     return this.parseAndProcessTokens(inString, format, unitsConversions);
   }
 
+  /** The value of special direction is always in degrees, try to find the unit conversion for that. */
+  private static processSpecialBearingDirection(mag: number, spec: ParserSpec): QuantityParseResult {
+    const specialDirUnit = spec.unitConversions.find((unitConversion) => unitConversion.name === "Units.ARC_DEG");
+    if (!specialDirUnit)
+      return { ok: false, error: ParseError.UnknownUnit };
+    const preferredUnit = spec.outUnit;
+    const conversion = this.tryFindUnitConversion(specialDirUnit.label, spec.unitConversions, preferredUnit);
+    if (!conversion) return { ok: true, value: mag };
+    return { ok: true, value: applyConversion(mag, conversion) };
+  }
+
   private static parseBearingFormat(inString: string, spec: ParserSpec): QuantityParseResult {
     const specialDirections: Record<string, number> = {
       n: 0,
@@ -805,7 +816,7 @@ export class Parser {
       const suffix = matchedSuffix?.toLowerCase() || "";
       const specialDirection = specialDirections[`${prefix}${suffix}`];
       if (specialDirection !== undefined)
-        return { ok: true, value: specialDirection };
+        return this.processSpecialBearingDirection(specialDirection, spec);
     }
 
     if (matchedPrefix === null || matchedSuffix === null) {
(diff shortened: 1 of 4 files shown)


core/quantity/src/test/BearingAzimuth.test.ts
```diff
@@ -176,6 +176,44 @@ describe("Bearing format tests:", () => {
...
+      },
+    };
+    const bearingDMSWithLabel = new Format("BearingDMSWithLabel");
+    await bearingDMSWithLabel.fromJSON(unitsProvider, bearingDMSWithLabelJson);
+    const bearingDMSParser = await ParserSpec.create(bearingDMSWithLabel, unitsProvider, degree);
+    const testData = [
+      { input: "s45d00m00se", expected:  135.0 },
+      { input: "s 45 00 00 e", expected:  135.0 },
+      { input: "n35 45 45.101e", expected:  35.7625 },
+      { input: "s45d45m45.0se", expected:  135.7625 },
```
hl662: @ckjackso this input, the `s` right before the last `e`, it stands for seconds?

ckjackso: @hl662 Yes. That is correct.

core/quantity/src/Parser.ts
```diff
@@ -391,6 +391,22 @@ export class Parser {
         tokens.push(new ParseToken(wipToken));
       }
     }
+    if (tokens.length === 1 && typeof tokens[0].value === "number" && format.type === "Bearing") {
```
hl662: this is a nitpick from me, but you can make the `format.type === "Bearing"` check be first in the if statement, that way if it's not a bearing the if statement won't bother checking the rest of the other segments and just return false

core/quantity/src/Parser.ts
```diff
@@ -283,6 +285,15 @@ export class Parser {
                 processingNumber = false;
                 wipToken = "";
               }
+              else if (format?.type === "Bearing" && wipToken.length > 0) {
```
rschili:
```suggestion
              else if (format.type === FormatType.Bearing && wipToken.length > 0) {
```

core/quantity/src/Parser.ts
```diff
@@ -381,6 +392,23 @@ export class Parser {
       }
     }
 
+    // Handle compact decimal input in Bearing format (e.g., 65.4545 → 65°45′45″)
+    if (format.type === "Bearing" && tokens.length === 1 && typeof tokens[0].value === "number" ) {
```
rschili:
```suggestion
    if (format.type === FormatType.Bearing && tokens.length === 1 && typeof tokens[0].value === "number" ) {
```

core/quantity/src/Parser.ts
```diff
@@ -381,6 +392,23 @@ export class Parser {
...
+      const value = tokens[0].value;
```
rschili: We cannot do it hard-wired like this. Check the BearingAzimuth.test.ts
The format specifies what units we use and how we arrange them. Granted, the consumers here seem to be pretty Focused on DMS (degrees minutes seconds), but you can use Bearing with radians, degrees, or as DM. The format gives you the flexibility to define any valid composite. This code is assuming that we are working with degrees, minutes and seconds, and it will produce invalid results when somebody uses a different format like RAD.

core/quantity/src/Parser.ts
```diff
@@ -52,6 +52,8 @@ export interface ParsedQuantity {
 enum Operator {
   addition = "+",
   subtraction = "-",
+  multiplication = "*",
+  division = "/"
```
rschili:
```suggestion
  multiplication = "*", // unsupported but we recognize it during parsing
  division = "/" // unsupported but we recognize it during parsing
```

core/quantity/src/test/BearingAzimuth.test.ts
```diff
@@ -176,6 +176,105 @@ describe("Bearing format tests:", () => {
...
+      { input: "n85 60 60e", expected: 86.0167 },
+      { input: "n85 45 65e", expected: 85.76275 },
+      { input: "s65:40:00w", expected: 245.667 },
+      { input: "s45e", expected: 135 },
+      { input: "S45E", expected: 135 },
+  //  { input: "s45,45e", expected: 245.667 }, // directly fail
+      { input: "s45d45m45se", expected: 134.2375 },
+      { input: "s45d45m45.0se", expected: 134.2375 },
+      { input: "s45d45m45.0e", expected: 134.2375 },
+     ];
```
rschili:
```suggestion
      { input: "NE", expected: 45.0 },
      { input: "SE", expected: 135.0 },
      { input: "SW", expected: 225.0 },
      { input: "NW", expected: 315.0 },
      { input: "N45E", expected: 45.0 },
      { input: "S45W", expected: 225.0 },
      { input: "N45W", expected: 315.0 },
      { input: "S45E", expected: 135.0 },
      { input: "nE", expected: 45.0 },
      { input: "sW", expected: 225.0 },
      { input: "nW", expected: 315.0 },
      { input: "sE", expected: 135.0 },
      { input: "n45e", expected: 45.0 },
      { input: "s45w", expected: 225.0 },
      { input: "n45w", expected: 315.0 },
      { input: "s45e", expected: 135.0 },
      { input: "N", expected: 0.0 },
      { input: "S", expected: 180.0 },
      { input: "n", expected: 0.0 },
      { input: "s", expected: 180.0 },
```

core/quantity/src/test/BearingAzimuth.test.ts
```diff
@@ -176,6 +176,168 @@ describe("Bearing format tests:", () => {
...
+          { name: "Units.ARC_SECOND", label: "\"" },
+        ],
+      },
+    };
+
+    const bearingFormat = new Format("bearing-flexible-parser");
+    await bearingFormat.fromJSON(unitsProvider, bearingFormatProps);
+    const bearingParser = await ParserSpec.create(bearingFormat, unitsProvider, degree);
+
+     const validTestData = [
```
hl662: @ckjackso can you go over the test cases we've added here, is there anything you think is missing?

ckjackso: @hl662 Nam,  For everything I see it looks like all the cases are covered.  Great job.

hl662: Let's add roundtrip testing to each of these inputs to be safe. So converting from the numeric value back to a formatting string. They don't have to be the same format as the input. `N45 45 45E` turns into 45.7625, which comes out as `N45°45'45"E`, adhering to the bearingFormatProps we defined at the top

core/quantity/src/Parser.ts
rushikesh-bentley: I’ve updated the logic and added checks to ensure it only applies to DMS formats — RAD and other cases are now correctly handled.

core/quantity/src/Parser.ts
```diff
@@ -605,39 +615,58 @@ export class Parser {
     const defaultUnit = format.units && format.units.length > 0 ? format.units[0][0] : undefined;
     defaultUnitConversion = defaultUnitConversion ? defaultUnitConversion : Parser.getDefaultUnitConversion(tokens, unitsConversions, defaultUnit);
 
+    if (format.type === FormatType.Bearing && format.units !== undefined && format.units.length > 0) {
```
rschili, line 618: @rushikesh-bentley I refactored this so it's independent of units. We cannot hard code degrees minutes and seconds or unit names into the parser. This solution is still sub optimal, but at least it does not hard-code units anymore. Please have a look.

core/quantity/src/test/BearingAzimuth.test.ts
```diff
@@ -176,6 +176,168 @@ describe("Bearing format tests:", () => {
...
+    const bearingRad = new Format("BearingRad");
+    await bearingRad.fromJSON(unitsProvider, bearingRadJson);
+    expect(bearingRad.hasUnits).to.be.true;
+
+    const bearingRadParser = await ParserSpec.create(bearingRad, unitsProvider, rad);
+
+    const testData = [
+      { input: "N0.785398E", expected: 0.785398 }, // 45 degrees in radians
+      { input: "S0.785398E", expected: 2.35619 }, // 135 degrees in radians
+      { input: "S", expected: 180 }, // 180 degrees in radians
```
hl662: this doesn't make sense, the bearingRadParser is using radian as the output. Shouldn't the expected value be 3.14159?

core/quantity/src/test/BearingAzimuth.test.ts
rushikesh-bentley: Added a roundtrip test for input n 45 45 45 e to ensure correct formatting back as per FormatProps.

hl662: 2 things:
1. Unless the comments are GitHub code suggestions, are very trivial, or the reviewer has replied that it's good, don't resolve comments made by another reviewer, just let them know you've addressed it and let them take a look and they'll resolve it if it's good.
2. I asked for roundtrip testing to **each** input, not just the one I gave as an example. You can extend each object of `validTestData` to also include an `expectedText`. Make sure there's a bearingFormatter, not just a bearingParser. I pushed a commit that does what I'm asking for, you'll have to do the same for each object in `validTestData`. 
I saw that you used degree as the outputUnit for bearingParser. You should use radians, the value we store in our iModels is in radians I believe, and the user who wants this feature will also be parsing/formatting using radians

core/quantity/src/Parser.ts
rushikesh-bentley: Okay, I checked it. Looks good now, thanks!

core/quantity/src/test/BearingAzimuth.test.ts
hl662: Digging deeper, this is something that was handled in a past PR, [it's a side-effect](https://github.com/iTwin/itwinjs-core/blob/master/core/quantity/src/Parser.ts#L774-L779) of handling special cases, we should probably revisit this case outside of this PR. @ckjackso is an input of "S" valid? Probably not?

hl662:

core/quantity/src/test/BearingAzimuth.test.ts
ColinKerr: If the parser says the persistence unit is radian than the output always has to be rad.  We are always going to want to parse to rad for persistence ... maybe it makes sense to have two expected values, one rad one degrees and try parsing with two different specs, one for rad and one for deg.

hl662: I've updated the specialDirectionParsing to always return the value based on the output Unit - since the special Direction cases always use ARC_DEG, we can try to find the unitConversion for ARC_DEG, and convert from that to the outUnit.

core/quantity/src/test/BearingAzimuth.test.ts
rushikesh-bentley: Thanks for the feedback, I'll keep that in mind going forward and make sure not to resolve reviewer comments unless they're trivial or confirmed as resolved.
 
Regarding the roundtrip tests, I updated the validTestData by adding expectedText for each input. I also removed the separate roundtrip test I pushed yesterday and incorporated everything into the current test as per your suggestion

rschili (approved):

ColinKerr (approved):

PR #8200, opened 2025-06-10
Presentation: Make `KoqPropertyValueFormatter` use `FormatsProvider` for finding format props

grigasp:
Part of: https://github.com/iTwin/viewer-components-react/issues/1340

@aruniverse, this is based on your request to use the new formatting APIs in the properties widget. I'm targeting `master` branch here, do you think we need to backport these changes to 5.0?

### DIFF
--- a/presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
+++ b/presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
@@ -6,24 +6,17 @@
  * @module Core
  */
 
-import { Format, FormatProps, FormatterSpec, ParserSpec, UnitsProvider, UnitSystemKey } from "@itwin/core-quantity";
-import {
-  getFormatProps,
-  InvertedUnit,
-  KindOfQuantity,
-  SchemaContext,
-  SchemaKey,
-  SchemaMatchType,
-  SchemaUnitProvider,
-  Unit,
-  UnitSystem,
-} from "@itwin/ecschema-metadata";
+import { assert, BeEvent } from "@itwin/core-bentley";
+import { Format, FormatProps, FormatsChangedArgs, FormatsProvider, FormatterSpec, ParserSpec, UnitsProvider, UnitSystemKey } from "@itwin/core-quantity";
+import { InvertedUnit, KindOfQuantity, SchemaContext, SchemaFormatsProvider, SchemaKey, SchemaMatchType, SchemaUnitProvider } from "@itwin/ecschema-metadata";
 
 /**
  * A data structure that associates unit systems with property value formatting props. The associations are used for
  * assigning formatting props for specific phenomenon and unit system combinations (see [[FormatsMap]]).
  *
  * @public
+ *
+ * @deprecated in 5.1. `FormatsMap` and related APIs have been deprecated in favor of [FormatsProvider]($core-quantity).
  */
 export interface UnitSystemFormat {
   unitSystems: UnitSystemKey[];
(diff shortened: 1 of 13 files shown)


presentation/common/src/test/KoqPropertyValueFormatter.test.ts
```diff
@@ -20,163 +20,224 @@ import { KoqPropertyValueFormatter } from "../presentation-common/KoqPropertyVal
...
       });
-      expect(formatted).to.be.eq(`1,5 ${usSurveyUnit.label}`);
-    });
 
-    it("formats value using 'Metric' system when KoQ supports metric and SI", async () => {
-      const formatted = await formatter.format(1.5, {
-        koqName: "TestSchema:TestKoqMetricAndSi",
-        unitSystem: "metric",
+      // FIXME: Fails because `SchemaFormatsProvider` doesn't have these changes: https://github.com/iTwin/itwinjs-core/pull/6262
+      it.skip("formats value using 'Metric' system when KoQ supports metric and SI", async () => {
```
grigasp: @hl662, I believe this test found a bug in `SchemaFormatsProvider`. Can you have a look?

presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
```diff
@@ -113,37 +144,44 @@ export class KoqPropertyValueFormatter {
...
+
+    const formatsProvider = this._formatsProvider ?? new SchemaFormatsProvider(this._schemaContext, unitSystem);
+    const formatProps = await formatsProvider.getFormat(koqName);
+
+    // `SchemaFormatsProvider` will fall back to default presentation format, but we want to fall back
+    // to default formats' map first, and only then to the default presentation format. All of this can
+    // be removed with the removal of default formats map.
+    if (this._defaultFormats && (!formatProps || (await getUnitSystemKey(this._unitsProvider, formatProps)) !== unitSystem)) {
+      const defaultFormatsProvider = createFormatsProviderFromDefaultFormats(this._schemaContext, this._defaultFormats, unitSystem);
+      const defaultFormatProps = await defaultFormatsProvider.getFormat(koqName);
```
saskliutas: Why `createFormatsProviderFromDefaultFormats` is needed? It looks like `createFormatsProviderFromDefaultFormats` can be replaced by `getFormat` it returns? Something like:

```suggestion
      const defaultFormatProps = await getFormatPropsFromDefaultFormats(this._schemaContext, this._defaultFormats, unitSystem, koqName);
```

presentation/common/src/test/KoqPropertyValueFormatter.test.ts
hl662: Thank you, I found this during testing my PR last month, I thought it was a feature, not a bug 😢 Opened a PR: https://github.com/iTwin/itwinjs-core/pull/8201

hl662: PR has been merged to master, this test can be re enabled

aruniverse:
Spoke to @rbreukelaar and @simnorm today; It should be ok if this slips into 5.1. Would be nice to have in 5.0, but with the team out this week and pending release, don't want to further delay it.

presentation/common/src/test/KoqPropertyValueFormatter.test.ts
grigasp: confirmed it passes after merge, thanks!

presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
grigasp: agreed and changed

PR #8201, opened 2025-06-10
Handle SI and Metric together in SchemaFormatsProvider

hl662:
Aligning with https://github.com/iTwin/itwinjs-core/pull/6262

SI and Metric unit systems should be treated as one unit system, when retrieving presentation formats from a KindOfQuantity.

- Bump some schema dependencies used in full stack tests


hl662:

hl662:
> Can you create tests for each unitSystem/group combination? i.e usCustomary/Finance. Looks like we never had tests that cover the each Case in the Switch block in the getUnitSystemGroupNames function. Also maybe rename this method? Something like getUnitSystemGroupMatchers?

Good suggestions @christophermlawson . I've added test cases for each combination, and renamed the method to getUnitSystemGroupMatchers. Had to add a couple schemas that use FiNANCE, or USSURVEY units and KoQs.

aruniverse (approved):

PR #8210, opened 2025-06-11
Add resolveInstanceId Cache

MichaelSwigerAtBentley:
Adds a cache for resolveInstanceId.
Adds a set of unit tests for the cache.
Swaps QueryElementIdByCode to use resolveInstanceId (and thus the cache).

Stores the result from resolveInstanceId and the args used to get it in a bi-directional map. Three separate maps are used to track the args (id, fedGuid, or code), meaning any combination of the 3 can be used to query the cache. All three are deleted when one is deleted.

Cache clearing is done on model or element delete or during a larger full cache clear. Cache is not cleared on updates, as id, fedGuid, and code should not change. 

-------------------------------

<h2>Performance Test Results:</h2>

<h3>resolveElementKey()</h3>

For performance testing, I performed 500,000 queries to `iModel.elements.resolveElementKey()` which calls `resolveInstanceKey()` and recorded the time elapsed in milliseconds to perform these queries on my local machine. I ran the test 5 times and averaged the results. I recorded these results with no cache, with more elements than the cache could hold (ensuring the cache would always be full and would never hit - worst case), and with less elements than the cache could hold (ensuring the cache would always hit after the first query - best case). The default cache size is 50. 

Below is the time in milliseconds for each test:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<body link="#467886" vlink="#96607D">

<br></br> | No Cache | Cache - All Miss | Cache - All Hit
-- | -- | -- | --
Test 1 | 2300 ms | 2417 ms | 39 ms
Test 2 | 2255 ms | 2424 ms | 42 ms
Test 3 | 2297 ms | 2398 ms | 38 ms
Test 4 | 2323 ms | 2395 ms | 41 ms
Test 5 | 2472 ms | 2358 ms | 39 ms
-- | -- | -- | --
Averages | 2329.4 ms | 2398.4 ms | 39.8 ms

</body>

<h3>queryElementIdByCode()</h3>

For testing, I queried a single element (the same element each time) by code 500,000 times using `iModel.elements.queryElementIdByCode()`. I recorded the time elapsed in milliseconds to perform these queries on my local machine. I ran each test 5 times and averaged the results. I recorded the old `queryElementIdByCode()` which used `withPreparedStatement`, the new `queryElementIdByCode()` which uses `resolveInstanceKey()` without a cache, and the new `queryElementIdByCode()` with the new cache.

Below is the time in milliseconds for each test:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<body link="#467886" vlink="#96607D">

<br></br> | Old withPreparedStatement | ResolveInstanceKey - No Cache | ResolveInstanceKey - Cache
-- | -- | -- | --
Test 1 | 4552 ms | 3052 ms | 168 ms
Test 2 | 4622 ms | 3118 ms | 168 ms
Test 3 | 4547 ms | 3066 ms | 172 ms
Test 4 | 4534 ms | 3177 ms | 165 ms
Test 5 | 4578 ms | 3029 ms | 170 ms
-- | -- | -- | --
Averages | 4566.6 ms | 3088.4 ms | 168.8 ms

</body>


core/backend/src/internal/ElementLRUCache.ts
```diff
@@ -120,3 +121,181 @@ export class ElementLRUCache {
...
+  id?: Id64String;
+  code?: string;
+  federationGuid?: Id64String;
+}
+
+/**
+ * A map to store instance keys based on different optional arguments like id, code, or federationGuid.
+ * This allows the cache to be searched by any combination of arguments.
+ */
+class ArgumentsToResultMap {
```
pmconne, line 136: Did you consider simplifying this to use a single [Dictionary](https://www.itwinjs.org/reference/core-bentley/collections/dictionary/) instead of 3 maps?

pmconne:
The linked issue claims this will help to address performance issues in a specific application. @khanaffan  asserts:
> We are currently doing performance testing on this new code, but one hotspot will be resolveInstanceKey and we need to cache its result in map on native side

I would like to see before and after measurements illustrating the degree of improvement achieved by this PR.

core/backend/src/internal/ElementLRUCache.ts
MichaelSwigerAtBentley: I thought about it, with the issue being cases where the key stored in the dictionary is more complete than the key used in the query. IE: a case where {id, code, and fedGuid} are stored as the key in the dictionary, but then only {fedGuid} is passed in as a key. We would want the cache to hit in that case. When using a Dictionary, the solution would be a dictionary where the key is an object that consists of id, code, and fedGuid, then instead of using get() I would need to define a custom function that iterates through all of the keys and checks if a part of the key is a match, not the whole key itself. 

I was worried about the overhead of performing that many comparisons. With 3 maps, I only need to iterate through the relevant key, and can use the default implementation of get(). The downside being memory, as it is potentially redundant to store the same result (consisting of id and fullClassName) 3 times in 3 different maps. 

Do you think the dictionary approach would be better? Is there a better way to query through the dictionary for partial key matches?

MichaelSwigerAtBentley:
> The linked issue claims this will help to address performance issues in a specific application. @khanaffan asserts:
> 
> > We are currently doing performance testing on this new code, but one hotspot will be resolveInstanceKey and we need to cache its result in map on native side
> 
> I would like to see before and after measurements illustrating the degree of improvement achieved by this PR.

@pmconne @khanaffan

I've added some basic performance results to the PR description. 
Let me know if if you want me to test anything else out

pmconne (approved):

PR #8213, opened 2025-06-11
remove extract api from native builds

DanRod1999:
remove extract api check in native builds because it does not break addon, and extract api is checked during the core PR validation as a GH workflow

### DIFF
--- a/common/config/azure-pipelines/templates/core-build.yaml
+++ b/common/config/azure-pipelines/templates/core-build.yaml
@@ -134,8 +134,3 @@ steps:
   - script: node common/scripts/install-run-rush.js docs
     displayName: rush docs
     workingDirectory: ${{ parameters.workingDir }}
-
-  - script: node common/scripts/install-run-rush.js extract-api
-    displayName: rush extract-api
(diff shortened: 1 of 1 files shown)


common/config/azure-pipelines/templates/core-build.yaml
```diff
@@ -134,8 +134,3 @@ steps:
   - script: node common/scripts/install-run-rush.js docs
     displayName: rush docs
     workingDirectory: ${{ parameters.workingDir }}
-
-  - script: node common/scripts/install-run-rush.js extract-api
```
aruniverse, line 138: doesnt this remove it from the js pipelines?

DanRod1999: itwinjs-core pipeline runs extract api in a [workflow](https://github.com/iTwin/itwinjs-core/actions/runs/15591983308/workflow?pr=8213). this is only ran when core-build is called by a native PR, you can see in the condition below this line it only runs when nativePR is true

aruniverse (approved):

PR #8220, opened 2025-06-13
add deprecation dates

paulius-valiunas:


core/common/src/BackgroundMapSettings.ts
```diff
@@ -67,11 +67,11 @@ export interface BackgroundMapProps {
 export interface DeprecatedBackgroundMapProps {
   /** Identifies the source of the map tiles. Currently supported providers are "BingProvider" and "MapBoxProvider".
    * Default value: "BingProvider"
-   * @deprecated in 3.x. use MapImageryProps.backgroundBase.
+   * @deprecated in 3.x - might be removed in next major version. use MapImageryProps.backgroundBase.
```
MichaelSwigerAtBentley, line 70: Should we just say "will be removed in next major version"? Even if we don't get around to it, isn't it better to give a hard warning instead of something that might be misinterpreted. Someone could see "might" and get the wrong impression

Edit: I see that "might" is more inline with our listed policy.

hl662:
Should there be a NextVer.md entry?

hl662 (approved):

docs/changehistory/NextVersion.md
```diff
@@ -451,6 +451,14 @@ const schema = await context.getSchema(schemaKey, SchemaMatchType.Exact);
...
+
+We have introduced the concept of a new `@preview` tag to be used for APIs that are stable within the current major version but may change in the next major version. This tag is intended to help developers identify APIs that are not yet fully stable (`@public`) and may require adjustments in future releases.
+
+Deprecation notices now include specific dates that indicate the earliest time a deprecated APIs can be removed. This is to provide developers with a clear timeline for when they need to migrate away from deprecated APIs. Specific grace periods are now defined as **one year** for `@public` APIs and **three months** for `@preview` APIs. This is a departure from the previous policy of waiting until major version `N+2` for removal of an API deprecated in major version `N`; which was often too long and led to confusion about when APIs would actually be removed.
```
MichaelSwigerAtBentley: `indicate the earliest time a deprecated APIs can be removed`
should be `a deprecated API can be removed`

aruniverse:
```suggestion
Deprecation notices now include specific dates that indicate the earliest time a deprecated API can be removed. This is to provide developers with a clear timeline for when they need to migrate away from deprecated APIs. Specific grace periods are now defined as **one year** for `@public` APIs and **three months** for `@preview` APIs. This is a departure from the previous policy of waiting until major version `N+2` for removal of an API deprecated in major version `N`; which was often too long and led to confusion about when APIs would actually be removed.
```

aruniverse:
@paulius-valiunas on monday can you cherry pick these dates over into master?

paulius-valiunas:
https://github.com/iTwin/itwinjs-core/pull/8235

PR #8224, opened 2025-06-13
Clean up NextVer.md for 5.0 publish

aruniverse:


docs/changehistory/NextVersion.md
anmolshres98, line 1: Under `### Reworked @itwin/ecschema-metadata package`
Add the following:
```md
- Removed `Schema.getClasses()` - use `Schema.getItems()` instead with `ECClass.isECClass()` check for each item returned
 ```

Issue #8141, opened 2025-05-23
Contour lines do not show on SectionDrawings


Ellord207:
**Describe the bug**
When creating a SectionDrawing using a Spatial View with Contour Display, the SectionDrawing does not show effects from Contour Display.

**To Reproduce**

1. Open iModel in Display Test App
2. Enabled Apply Contour Display in the view
3. Create a Section Drawing from view (I used the `dta create section drawing` key-in)
4. Observe there are no affects from Contour Display in the section drawing.

**Expected behavior**
The SectionDrawing should have the Contour Display applied as it is in the Spatial View

**Screenshots**
Spatial View in the viewport
![Image](https://github.com/user-attachments/assets/208c48ec-0736-47e9-8014-378a940aba17)

SectionDrawing created from that spatial view
![Image](https://github.com/user-attachments/assets/b2874e33-9e81-4c94-9287-f45d6181c354)

**Desktop (please complete the applicable information):**

- OS: Windows
- Electron
- iTwin.js Version [5.1.0-dev.22]
- Node: 22.15.0

**Additional context**

Ellord207:
We are needing to investigate if this approach will work for our requirements.  Could you please prioritize this issue?

andremig-bentley:
Here is a draft PR addressing this issue - https://github.com/iTwin/itwinjs-core/pull/8229

This seems to be a two part problem. One, contour lines are currently only able to be drawn on 3d display styles. This PR addresses that issue. Two, contour line data associated with a spatial view does not get passed down into a SectionDrawing created from that view. This problem is still under investigation, the PR description goes into more detail. 

Below is a screenshot of a separately defined contour lines group being shown on a section drawing.

![Image](https://github.com/user-attachments/assets/8019cc1b-880a-45d3-8056-6e2982cdf9ef)

Ellord207:
Thanks for getting a fixed merged.  Once the version is available in our environment, I'll test and validate there that Contours in Sections will satisfy our needs.  Closing this issue as the reported bug is resolved in core.

markschlosseratbentley:
Reopening until released in iTwin.

aruniverse:
@Ellord207 this was released in 5.1.x; can you verify this is working?

Ellord207:
I did a quick initial test and it didn't work for me in our test app.  I need to do a deeper investigation make sure I didn't do anything incorrect and compare with display-test-app.  Will report back about this, hopefully later this week.

[...]

PR #8229, opened 2025-06-13
Allow Contour Lines on Section Drawings

andremig-bentley:
This PR is to address this issue: https://github.com/iTwin/itwinjs-core/issues/8141

Currently, a SectionDrawing created from a Spatial View does not show the contour lines associated with the spatial view. This seems to be a two part problem.

1 - Currently contour lines are not able to be drawn at all on SectionDrawings as they are limited to 3D display styles only. This is fixed by removing the `is3d()` check in `createRenderPlanFromViewport`, and moving all contour logic from `DisplayStyle3dSettings` into `DisplayStyleSettings`. 

2 - While with the above changes a new group of contour lines can be added to an existing section drawing (see image below), when the SectionDrawing is created from a spatial view, the existing contour data is not passed along into the new SectionDrawing. 

For clarity, if a contour line group is defined with specific colors, line style, etc. on a spatial view, and then a SectionDrawing is created from that view, the resulting SectionDrawing has no defined contour groups. A new group can be defined to match the group defined in the spatial view, but it is not created with the SectionDrawing. It seems to me that the expected behavior would be that the contour data defined within the spatial view would be passed along into the SectionDrawing as it's created. 

This problem is still under investigation as to the best way to solve it, which is why this PR is still in draft.

Below is a previously impossible section drawing with a contour lines group defined.

sectionWithContours

### DIFF
--- a/core/common/src/DisplayStyleSettings.ts
+++ b/core/common/src/DisplayStyleSettings.ts
@@ -150,6 +150,8 @@ export interface DisplayStyleSettingsProps {
   planarClipOvr?: DisplayStylePlanarClipMaskProps[];
   /** See [[DisplayStyleSettings.whiteOnWhiteReversal]]. */
   whiteOnWhiteReversal?: WhiteOnWhiteReversalProps;
+  /** See [[DisplayStyle3dSettings.contours]]. */
+  contours?: ContourDisplayProps;
 }
 
 /** JSON representation of [[DisplayStyle3dSettings]]  associated with a [[DisplayStyle3dProps]].
@@ -442,6 +444,7 @@ export class DisplayStyleSettings {
   private _clipStyle: ClipStyle;
   private readonly _contextRealityModels: ContextRealityModels;
   private _whiteOnWhiteReversal: WhiteOnWhiteReversalSettings;
+  private _contours: ContourDisplay;
 
   /** Returns true if this is a [[DisplayStyle3dSettings]]. */
   public is3d(): this is DisplayStyle3dSettings {
@@ -592,6 +595,8 @@ export class DisplayStyleSettings {
       createContextRealityModel: options?.createContextRealityModel,
       deferPopulating: options?.deferContextRealityModels,
     });
+
+    this._contours = ContourDisplay.fromJSON(this._json.contours);
   }
 
   /** Flags controlling various aspects of the display style. */
@@ -975,6 +980,8 @@ export class DisplayStyleSettings {
       props.excludedElements = this._excludedElements.ids;
     }
 
+    props.contours = this.contours.toJSON();
+
     return props;
   }
 
@@ -995,6 +1002,9 @@ export class DisplayStyleSettings {
   public applyOverrides(overrides: DisplayStyleSettingsProps): void {
     this._applyOverrides(overrides);
     this.onOverridesApplied.raiseEvent(overrides);
+    if (overrides.contours)
+      this.contours = ContourDisplay.fromJSON(overrides.contours);
+
   }
 
   /** @internal */
@@ -1067,6 +1077,17 @@ export class DisplayStyleSettings {
 
     this.onOverridesApplied.raiseEvent(overrides);
   }
+
+  /** The settings that control contour display. */
+  public get contours(): ContourDisplay { return this._contours; }
+  public set contours(contours: ContourDisplay) {
+    if (contours.equals(this.contours))
+      return;
+
+    this.onContoursChanged.raiseEvent(contours);
+    this._contours = contours;
+    this._json.contours = contours.toJSON();
+  }
 }
 
 /** Provides access to the settings defined by a [[DisplayStyle3d]] or [[DisplayStyle3dState]], and ensures that
@@ -1075,7 +1096,6 @@ export class DisplayStyleSettings {
  */
 export class DisplayStyle3dSettings extends DisplayStyleSettings {
   private _thematic: ThematicDisplay;
-  private _contours: ContourDisplay;
   private _hline: HiddenLine.Settings;
   private _ao: AmbientOcclusion.Settings;
   private _solarShadows: SolarShadowSettings;
@@ -1092,7 +1112,6 @@ export class DisplayStyle3dSettings extends DisplayStyleSettings {
   public constructor(jsonProperties: { styles?: DisplayStyle3dSettingsProps }, options?: DisplayStyleSettingsOptions) {
     super(jsonProperties, options);
     this._thematic = ThematicDisplay.fromJSON(this._json3d.thematic);
-    this._contours = ContourDisplay.fromJSON(this._json3d.contours);
     this._hline = HiddenLine.Settings.fromJSON(this._json3d.hline);
     this._ao = AmbientOcclusion.Settings.fromJSON(this._json3d.ao);
     this._solarShadows = SolarShadowSettings.fromJSON(this._json3d.solarShadows);
@@ -1170,8 +1189,6 @@ export class DisplayStyle3dSettings extends DisplayStyleSettings {
       }
     }
 
-    props.contours = this.contours.toJSON();
-
     return props;
   }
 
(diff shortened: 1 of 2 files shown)


core/frontend/src/internal/render/webgl/Graphic.ts
```diff
@@ -370,6 +371,13 @@ export class Branch extends Graphic {
     this.inSectionDrawingAttachment = opts.inSectionDrawingAttachment;
     this.disableClipStyle = opts.disableClipStyle;
     this.transformFromExternalIModel = opts.transformFromIModel;
+    const target = IModelApp.viewManager.selectedView?.target;
+    if((target as Target).plan.contours === undefined && opts.contours !== undefined){
+      (target as Target).plan.contours = opts.contours;
+      const viewport = IModelApp.viewManager.selectedView;
+      if (viewport)
+        viewport.displayStyle.settings.contours = opts.contours;
```
pmconne: No, the `Viewport` drives the `Target` and everything it renders, not the other way around.

pmconne:
> Currently contour lines are not able to be drawn at all on SectionDrawings as they are limited to 3D display styles only. This is fixed by removing the is3d() check in createRenderPlanFromViewport, and moving all contour logic from DisplayStyle3dSettings into DisplayStyleSettings.

Contour lines are by definition only applicable to 3d views. A 2d view (whether a section drawing, or a sheet) can have a spatial view (or more than one, for sheets) attached to them. The attached spatial view has its own 3d display style that defines how it gets drawn in the context of the 2d view. It does not alter the 2d view. We "push" the 3d view with its display style, graphics, etc; draw it; then "pop" it, restoring the previous 2d state. The contour line settings should be part of the pushed 3d state.

danielzhong-bentley:
> Contour lines are by definition only applicable to 3d views. A 2d view (whether a section drawing, or a sheet) can have a spatial view (or more than one, for sheets) attached to them. The attached spatial view has its own 3d display style that defines how it gets drawn in the context of the 2d view. It does not alter the 2d view. We "push" the 3d view with its display style, graphics, etc; draw it; then "pop" it, restoring the previous 2d state. The contour line settings should be part of the pushed 3d state.

Could you please take a look at my latest changes and let me know if I’m on the right track?

core/frontend/src/internal/render/webgl/BranchUniforms.ts
```diff
@@ -267,4 +269,11 @@ export class BranchUniforms {
       this.clipStack.intersectionStyle.alpha = disableClipStyle ? 0 : (style.intersectionStyle ? style.intersectionStyle.width : 0);
     }
   }
+
+  private setContourLine(contourLine: ContourDisplay) {
```
pmconne: No, you are still allowing the Target to alter the Viewport, and you are assuming you are drawing to the ViewManager's selected view, which will not always be true.

pmconne:
> Could you please take a look at my latest changes and let me know if I’m on the right track?

Look at how `Branch.edgeSettings` gets propagated or overridden as branches are pushed, and restored when they are popped. Similar behavior should apply here. When binding the contour line uniforms, you should be consulting the top of the branch stack, not `ViewManager.selectedView.view` or whatever.

danielzhong-bentley:
> > Could you please take a look at my latest changes and let me know if I’m on the right track?
> 
> Look at how `Branch.edgeSettings` gets propagated or overridden as branches are pushed, and restored when they are popped. Similar behavior should apply here. When binding the contour line uniforms, you should be consulting the top of the branch stack, not `ViewManager.selectedView.view` or whatever.

My latest change use similar idea that updates the contour information during both push and pop by modifying the render plan.

pmconne:
> My latest change use similar idea that updates the contour information during both push and pop by modifying the render plan.

No. The render plan is an input - do not modify it.

Look again at `BranchState.edgeSettings`. It is initialized *from* the render plan's hidden line settings for the branch at the top of the stack by `Target.changeRenderPlan`. Thereafter, other branches can be pushed with their own edge settings. We never consult the render plan to find out what the current edge settings are - we always consult the branch stack. Contour line settings should work the same way.

danielzhong-bentley:
> > My latest change use similar idea that updates the contour information during both push and pop by modifying the render plan.
> 
> No. The render plan is an input - do not modify it.
> 
> Look again at `BranchState.edgeSettings`. It is initialized _from_ the render plan's hidden line settings for the branch at the top of the stack by `Target.changeRenderPlan`. Thereafter, other branches can be pushed with their own edge settings. We never consult the render plan to find out what the current edge settings are - we always consult the branch stack. Contour line settings should work the same way.

Please review my latest change

common/api/core-frontend.api.md
```diff
@@ -4304,6 +4304,8 @@ export interface GraphicBranchOptions {
     // @internal (undocumented)
     classifierOrDrape?: RenderPlanarClassifier | RenderTextureDrape;
     clipVolume?: RenderClipVolume;
+    // @internal (undocumented)
```
markschlosseratbentley, line 4307: @danielzhong-bentley Can you check out the failing contour line test? https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4134920&view=logs&j=751a5ea6-e147-5a0b-b7c5-e28ecfcb00e2&t=742556c9-a44d-5542-8a03-cf20a9c5ac3e&l=13817

core/frontend/src/internal/render/webgl/BatchUniforms.ts
```diff
@@ -95,7 +95,7 @@ export class BatchUniforms {
   }
 
   public get wantContourLines(): boolean {
-    const contours = this._target.plan.contours;
+    const contours = this._target.is3d? this._target.plan.contours : this._target.currentContours;
```
pmconne:
```suggestion
    const contours = this._target.currentContours;
```

Why the need to branch based on whether the view is 3d? You could have a branch in a 3d view that pushes different contour settings than those (if any) defined on the view.

core/frontend/src/internal/render/webgl/ContourUniforms.ts
```diff
@@ -57,14 +57,15 @@ export class ContourUniforms {
 
   public update(target: Target): void {
     const plan = target.plan;
-
-    if (this.contourDisplay && plan.contours && this.contourDisplay.equals(plan.contours)) {
+    const contourLine = target.is3d? plan.contours : target.currentContours;
```
pmconne: Why branch based on `target.is3d`? `target.currentContours` should have been initialized from the `RenderPlan` regardless.

core/frontend/src/internal/render/webgl/Contours.ts
```diff
@@ -40,14 +40,15 @@ export class Contours implements WebGLDisposable {
   }
 
   private matchesSubCategories(): boolean {
-    if (this._contours === undefined && this.target.plan.contours === undefined)
+    const contours = this.target.is3d? this.target.plan.contours : this.target.currentContours;
```
pmconne: Same question about `target.is3d`.

pmconne (commented):
You are almost there. I don't understand why you're checking `target.is3d` everywhere.

danielzhong-bentley:
> You are almost there. I don't understand why you're checking `target.is3d` everywhere.

Because the original contour line unit tests failed after I changed to use the graphic branch instead of the render plan, even though it looked correct to me when I tested my own example. So I tried to make 3D work like the original using the render plan, and only 2D use the graphic branch information. I will investigate this more.

pmconne:
With your fix reverted the SceneCompositor is reading the contour data just fine, but the shaders are writing zeroes for every pixel instead of the contour line info. That's very confusing if as you claim opening a 3d view and applying contours causes them to display correctly.

danielzhong-bentley:
> With your fix reverted the SceneCompositor is reading the contour data just fine, but the shaders are writing zeroes for every pixel instead of the contour line info. That's very confusing if as you claim opening a 3d view and applying contours causes them to display correctly.

That's odd. I've also included a demo video in the description above.

pmconne:
Awaiting #8272 for audit fix.

markschlosseratbentley (approved):

danielzhong-bentley (approved):

PR #8239, opened 2025-06-16
fix docs links

ben-polinsky:
Several documentation links in 5.0.0 changehistory were not validated. The root cause is our docs building not validating nextVersion.md links.

@DanRod1999 looks like we still need to get 5.0.0 changehistory into master as well

### DIFF
--- a/docs/changehistory/5.0.0.md
+++ b/docs/changehistory/5.0.0.md
@@ -99,10 +99,10 @@ Because the `SelectionSet` now stores additional types of ids, existing code tha
 
 By default a box select with the selection tool will only identify visible elements (i.e. elements that light up a pixel in the current view). Sometimes it is desirable to select all elements that are inside or overlap the box regardless of whether they are currently obscured by other elements. Applications can now change [ToolSettings.enableVolumeSelection]($core-frontend) to enable box selection by volume in spatial views.
 
-The following protected methods on SelectTool had their signature changed to support volume selection:
+The following protected methods on SelectionTool had their signature changed to support volume selection:
 
-- [SelectTool.selectByPointsProcess]($core-frontend)
-- [SelectTool.selectByPointsEnd]($core-frontend)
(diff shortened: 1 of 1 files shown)


DanRod1999:
looks like we forget to kick off the workflow, so will do that now, should fix changelogs

ben-polinsky:
Fixing docs build so we *will* validate nextVersion.md and the changehistory index.md file (most current change history - are in effect the same).

This issue stems from the need to not validate historical changehistories since there could be APIs removed.

aruniverse:
> looks like we forget to kick off the workflow, so will do that now, should fix changelogs

why isnt this automated?

docs/changehistory/5.0.0.md
```diff
@@ -922,20 +923,15 @@ For more information read [Pull merge & conflict resolution](../learning/backend
 
 Existing calls like `context.getSchemaItem<EntityClass>("schema:myName")` have to be adjusted either into
 `context.getSchemaItem("schema", "myName", EntityClass)` or more verbose as a general item followed by a type-guard:
-
-```ts
-const item: SchemaItem = await iModel.schemaContext.getSchemaItem("BisCore", "Element");
-if (item && EntityClass.isEntityClass(item)) {
-}
-```
+s
```
ColinKerr: Replacing code block with 's'?

ben-polinsky: That's a k+s key command to save without formatting gone wrong. Thanks

ben-polinsky: fixed

DanRod1999:
> > looks like we forget to kick off the workflow, so will do that now, should fix changelogs
> 
> why isnt this automated?

For some reason despite numerous attempts to fix this, the change log script does not kick off for major/minor releases. Only reason I can find is it might be due to the size of the commit being too large and it messes with GH's triggers, but that is still only a guess. We'd have to look into a different way of triggering this workflow if we want a permanent solution

PR #8247, opened 2025-06-17
Incremental loading changes

christophermlawson:
Changes for incremental schema loading. In ecschema-metadata, all new code is marked as internal. The IModelIncrementalSchemaLocater in core/backend and the RpcIncrementalSchemaLocater in ecschema-rpc are marked as beta. These are the only two classes that will be used in the public API.


core/ecschema-metadata/src/IncrementalLoading/IncrementalSchemaLocater.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  /**
+   * Loads the schema info objects for the given context.
+   * @param context   The schema context to load the schema infos for.
+   * @returns         A promise that resolves to an iterable of schema infos.
+   */
+  public abstract loadSchemaInfos(context: SchemaContext): Promise<Iterable<SchemaInfo>>;
+
+  /**
+   * Checks if the context contains teh right schemas to support incremental schema loading.
```
ColinKerr:
```suggestion
   * Checks if the context contains the right schemas to support incremental schema loading.
```

core/ecschema-metadata/src/IncrementalLoading/IncrementalSchemaLocater.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    const orderedSchemaInfos = SchemaGraphUtil.buildDependencyOrderedSchemaInfoList(schemaInfos) as Array<SchemaInfo & { props: SchemaProps }>;
+    return orderedSchemaInfos.map((schemaInfo) => schemaInfo.props);
+  }
+}
+
+/**
+ * Helper class to manage schema infos for a schema context.
+ */
+class SchemaInfoCache {
```
ColinKerr, line 259: I don't like this, any cache should be on the schema context not in the individaul locater.  I get why it's here given the current interfaces but it still seems wrong.

I am going to accept it here because it's internal to the implementation

core/backend/src/test/assets/IncrementalSchemaLocater/configs/old.config.json
```diff
@@ -0,0 +1,68 @@
+{
+  "label": "Old Profile iModel",
+  "bimFile": "OldIModel.bim",
```
ColinKerr: Ideally we would generate our test bims at runtime or use one of the existing test bims.  Since this one is specifically supposed to be an elder iModel, maybe the tradeoff is worth it to have the bim checked in.

core/backend/src/test/incrementalSchemaLocater/IncrementalLoading.test.ts
```diff
@@ -0,0 +1,203 @@
...
+      iModelConfiguration.schemas.forEach((testSchemaConfiguration: any) => {
+        describe(`Schema: ${testSchemaConfiguration.name}`, () => {
+          let testSchemaKey: SchemaKey;
+
+          before(async function () {
+            testSchemaKey = await resolveSchemaKey(testSchemaConfiguration.name);
+          });
+
+          beforeEach(function () {
+            // this.currentTest!.iModel = iModelConfiguration.label;
```
hl662: Do you still need this? If not, I would remove

christophermlawson: Removed.

hl662 (approved):
Don't see how the precision could be longer than 1000 characters, so approving for core-quantity

core/backend/src/IModelIncrementalSchemaLocater.ts
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Gets [[SchemaProps]]($ecschema-metadata) for the given [[SchemaKey]]($ecschema-metadata).
+   * This is the full schema json with all elements that are defined in the schema.
+   * @param schemaKey The key of the schema to be resolved.
+   */
+  public async getSchemaProps(schemaKey: SchemaKey): Promise<SchemaProps | undefined> {
+    // To keep the main thread from being blocked in sync loading cases the resolving
+    // is triggered through a timeout. Even if there is no delay, it improves loading
+    // time by ~3x.
+    return new Promise((resolve, reject) => setTimeout(() => {
```
ben-polinsky, line 45: Did you consider `setImmediate` instead of `setTImeout` `0`? Similar performance?

core/backend/src/test/assets/IncrementalSchemaLocater/configs/old.config.json
christophermlawson: I kept the OldIModel.bim because we are validating certain schema information that's expected. The other two .bims have been removed.

core/backend/src/IModelIncrementalSchemaLocater.ts
StefanApfel-Bentley: I looked at it too, however since it was marked deprecated in the MDN docs, I went with the setTimeout.
https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes.

Since that schema locater (or better one of it's implementations is suppose to run in web browsers too, we better not use it, even if it does not seem to be deprecated for node.js.

ben-polinsky: Ah, I missed that it could run on the browser. Indeed.

PR #8248, opened 2025-06-17
TabRuns

### LINKED ISSUE #7678: TextAnnotations - TabRun

I would like to be able to insert tabs into a TextBlock. It's just whitespace, so I'm thinking something similar to `LineBreakRun` that adds whitespace for tabbing.

A new type of run should be created called `TabRun`
* `StyleOverrides` should include a `spaces` property to specify how many spaces make up a tab. Usually 2 or 4 spaces. Default would be 4.
* In the `stringify` method, users should be able to specify if they want their tabs to be computed with spaces or as a tab (`"\t"`) character.

claudiaareneee:
## Description

This PR contains the following changes:
* Creates a new type of run called `TabRun`
* Associates the tab styling within `TextStyleSettings`
* Adds tests for new run type
* Adds a new geometry feature for debugging the ranges of individual text runs
* Adds a `isWhitespace` method

## Link to Issue

closes #7678 

## Screenshots

tabbing


core/common/src/annotation/TextBlock.ts
```diff
@@ -497,7 +559,9 @@ export class TextBlock extends TextBlockComponent {
     return TextBlock.create({ styleName: "" });
   }
 
-  /** Returns true if every paragraph in this text block is empty. */
+  /** Returns true if every paragraph in this text block is empty.
+   * idk, should I get rid of this? I can't see a use for it. using isWhitespace is probably better in most cases.
+  */
```
claudiaareneee: I need to remove this comment, but the question is still valid. Is isEmpty useful? I mean it corresponds to createEmpty, but it shouldn't be used when determining whether or not geometry can be created. That would rely on `isWhitespace`

pmconne: They are two different questions. isEmpty asks, does this text block contain anything at all? A text block containing only whitespace is not empty.

claudiaareneee: Okay, I can leave it then and remove the comment

pmconne:
Your screenshot doesn't have any examples of consecutive tab stops - did you test those?
Did you test lines containing only tab stops? Lines ending in tab stops?
Did you test how tab stops behave with line wrapping?

pmconne:
Did you test how layout behaves when you have a text block containing multiple tab runs where the style is overridden to specify different tab widths for different runs?

core/backend/src/annotations/TextBlockGeometry.ts
```diff
@@ -137,7 +137,8 @@ export function produceTextBlockGeometry(layout: TextBlockLayout, documentTransf
   for (const line of layout.lines) {
     const lineTrans = Transform.createTranslationXYZ(line.offsetFromDocument.x, line.offsetFromDocument.y, 0);
     for (const run of line.runs) {
-      if ("linebreak" === run.source.type) {
+      // Skip runs that are solely whitespace
```
pmconne, line 140: Consider calling a new `isWhitespace` function instead so that your intention is clear without requiring a comment and that this code will continue to work if you later introduce a new type of whitespace.

pmconne (approved):
Just need a bunch more tests especially around edges cases.

claudiaareneee:
@pmconne, I added tests and fixed some bugs. I ended up cleaning up the `TextBlock.test.ts` in the backend so the diff probably looks illegible now. For clarity, everything following [75b15bf](https://github.com/iTwin/itwinjs-core/pull/8248/commits/75b15bf6215f4437552e0d4e319d5517c2a7e179) is just whitespace and rearranging. If you want an easier to read diff since you last reviewed, checkout the commit here: [ed98732](https://github.com/iTwin/itwinjs-core/pull/8248/commits/ed9873235249b26815679068b46e99d4096c47a7#diff-0406ba7598de52017fbad055e36dd1b66a8f703fdddea92d0c9712f7a00a41b7)

For reference, this is what the new organization looks like: 

image

core/backend/src/annotations/TextBlockGeometry.ts
claudiaareneee: done

claudiaareneee:
That should be all the changes you requested. Let me know if I missed anything. I have three new test blocks: 
* `applies tab shifts`
* `applies consecutive tab shifts`
* `wraps tabs`

I did make changes to `TextBlockLayout` since you last viewed.

core/common/src/annotation/TextBlock.ts
```diff
@@ -50,6 +50,10 @@ export interface TextBlockStringifyOptions {
    * Default: "/"
    */
   fractionSeparator?: string;
+  /** Whether of not to use spaces or \t for tabs.
```
MichaelSwigerAtBentley: Maybe this doc should be clarified that tabsAsSpaces is not just a boolean, but a number representing the number of spaces to use in place of a tab.

core/common/src/annotation/TextStyle.ts
```diff
@@ -182,6 +192,7 @@ export class TextStyleSettings {
     this.superScriptOffsetFactor = props.superScriptOffsetFactor ?? defaults.superScriptOffsetFactor;
     this.superScriptScale = props.superScriptScale ?? defaults.superScriptScale;
     this.widthFactor = props.widthFactor ?? defaults.widthFactor;
+    this.tabInterval = undefined === props.tabInterval ? defaults.tabInterval : props.tabInterval;
```
MichaelSwigerAtBentley: nit: Could this use `??` operator like the pervious values for consistency? 

```suggestion
    this.tabInterval = props.tabInterval ?? defaults.tabInterval;
```

core/common/src/annotation/TextBlock.ts
claudiaareneee: oh right, yes. fixed!

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -401,6 +412,12 @@ export class RunLayout {
     });
   }
 
+  public applyTabShift(parent: LineLayout, context: LayoutContext) {
```
pmconne:
```suggestion
  /** @internal */
  public applyTabShift(parent: LineLayout, context: LayoutContext) {
```

`LayoutContext` is not exported, no one outside this source file can use it so it doesn't belong in a public method. (I don't know why api-extractor didn't complain about that).

pmconne: Moved it to file scope.

pmconne:
Thanks for your patience @claudiaareneee - very nice PR!

PR #8263, opened 2025-06-24
Provide a way to mark changes as indirect with less restrictive locking rules

rschili:
imodel-native: https://github.com/iTwin/imodel-native/pull/1140

Allow specifying if changes are indirect. The API area where the flag could be provided is huge, for now we just expose what's needed during the upcoming OpenSite+ sprint. We will most likely iterate on this.

### DIFF
--- a/full-stack-tests/backend/src/integration/IndirectChanges.test.ts
+++ b/full-stack-tests/backend/src/integration/IndirectChanges.test.ts
@@ -134,22 +134,6 @@ describe("Indirect changes flag on elements", () => {
     b1.close();
   });
 
-  it("Indirect changes should not require a lock", async () => {
-    const b1 = await openNewBriefcase(user1AccessToken);
-    await assertThrowsAsync(async () => {
-      makeSubject(b1, "Subject 1", "Description 1", false);
-    }, "shared lock not held on model for insert");
-
-    await assertThrowsAsync(async () => {
-      makeSubject(b1, "Subject 2", "Description 2");
-    }, "shared lock not held on model for insert");
-
-    const b1s3 = makeSubject(b1, "Subject 3", "Description 3", true);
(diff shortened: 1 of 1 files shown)


core/backend/src/Element.ts
```diff
@@ -200,12 +206,16 @@ export class Element extends Entity {
...
-    const { iModel, props } = arg;
+    const { iModel, props, options } = arg;
     const operation = "insert";
-    iModel.channels[_verifyChannel](arg.props.model);
-    iModel.locks.checkSharedLock(props.model, "model", operation); // inserting requires shared lock on model
-    if (props.parent)   // inserting requires shared lock on parent, if present
-      iModel.locks.checkSharedLock(props.parent.id, "parent", operation);
+    iModel.channels[_verifyChannel](props.model);
+    const isIndirectChange = (options && options.indirect === true);
+    if (!isIndirectChange) {
```
khanaffan: Remove this check

full-stack-tests/backend/src/integration/IndirectChanges.test.ts
```diff
@@ -0,0 +1,262 @@
...
+  } catch (e) {
+    if (e instanceof Error && contains) {
+      assert.isTrue(e.message.includes(contains));
+    }
+    return;
+  }
+  throw new Error(`Failed to throw error with message: "${contains}"`);
+}
+
+describe.only("Indirect changes flag on elements", () => {
```
pmconne:
```suggestion
describe("Indirect changes flag on elements", () => {
```

core/backend/src/Element.ts
rschili: Done

rschili:
As discussed with @ColinKerr we will not merge this into master for now, instead we will merge it into our hackathon OS+ branch, and if we choose to keep the feature we will merge to master afterwards.

PR #8272, opened 2025-06-25
Add lint step to version bump script

### LINKED ISSUE #8270: Automatic deprecation date linting on release
There was an [eslint rule created](https://github.com/iTwin/eslint-plugin/pull/102) which, upon execution, adds appropriate deprecation dates to deprecated elements. Those dates inform developers when a specific deprecated element might be removed from code.

The issue that is being solved can be found [in this engineering working group discussion](https://github.com/iTwin/engineering-wg/issues/37)

Main goal would be modifying current [itwin-core pipelines](https://github.com/iTwin/itwinjs-core/tree/master/common/config/azure-pipelines) such, that upon releasing a new itwin-core version the following would happen:

- Before mentioned eslint "deprecation" rule would be run on release branch to add deprecation dates where they don't exist or replace existing, already passed dates with comment implying that code now may be removed.
- The changes on deprecation comments are mirrored on the master branch as well

augustasgri:
**Closes #8270: Automatic deprecation date linting on release**

### DIFF
--- a/common/config/azure-pipelines/jobs/version-bump.yaml
+++ b/common/config/azure-pipelines/jobs/version-bump.yaml
@@ -166,6 +166,9 @@ jobs:
       - script: node common/scripts/install-run-rush.js install
         displayName: rush install
 
+      - ${{ if or(eq(parameters.BumpType, 'minor'), eq('${{ parameters.BumpType }}', 'patch')) }}:
+          - template: "../templates/run-eslint-deprecation.yaml"
+
       - bash: node common/scripts/install-run-rush.js audit
         displayName: rush audit
 
--- a/common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
+++ b/common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
@@ -1,0 +1,78 @@
+steps:
+  - bash: node common/scripts/install-run-rush.js lint-deprecation
+    displayName: rush lint-deprecation
+
+  - bash: |
+      if [ -n "$(git status --porcelain)" ]; then
+        echo "##vso[task.setvariable variable=changesDetected]true"
+      else
+        echo "No changes to @deprecated comments were made. Skipping deprecation comment handling."
+        echo "##vso[task.setvariable variable=changesDetected]false"
+      fi
+    displayName: Check if any changes happened
+    name: checkChanges
+
+  - bash: git add .
+    displayName: git add .
+    condition: eq(variables['changesDetected'], 'true')
+
+  - bash: |
+      git commit -m "add deprecation dates using custom ESLint rule"
+      relBranchName="release/"$(Build.SourceBranchName)
+      git push https://$(GITHUBTOKEN)@github.com/iTwin/itwinjs-core HEAD:$relBranchName
+    displayName: Commit and push deprecation date changes to release branch
+    condition: eq(variables['changesDetected'], 'true')
+
+  - bash: |
+      # Get the SHA-1 hash of deprecation changes commit to apply it later on master
+      deprecationDateCommitHash=$(git log -1 --format=%H)
+      echo "##vso[task.setvariable variable=hash;isOutput=true]$deprecationDateCommitHash"
+    displayName: Save SHA-1 hash of the commit
+    name: deprecationCommit
+    condition: eq(variables['changesDetected'], 'true')
+
+  - bash: |
+      git fetch
+      git checkout origin/master
+    displayName: Checkout to up-to-date master
+    condition: eq(variables['changesDetected'], 'true')
+
+  - bash: |
+      commitHash=$(deprecationCommit.hash)
+      echo "Attempting to cherry-pick commit: $commitHash"
+      if git cherry-pick $commitHash; then
+        echo "Successfully applied deprecation comment updates to master"
+        git push https://$(GITHUBTOKEN)@github.com/iTwin/itwinjs-core HEAD:master
+      else
+        # Merge conflicts prevent us from committing changes to master directly.
+        # We create a branch with conflicting changes and set-up a PR.
+        # It is then expected that developer manually resolves merge conflicts and merges the PR
+        echo "Cherry-pick operation on master failed (most likely merge conflicts). Manual intervention required"
+        git cherry-pick --abort
+
+        # Create and checkout to conflict handling branch
+        conflictBranchName="cherry-pick-conflict-$(date +%y%m%d%H%m%s)"
+        echo "Creating new branch '$conflictBranchName' for manual resolution"
+        git checkout -b $conflictBranchName
+
+        # Cherry-pick changes onto conflict handling branch
+        commitHash=$(deprecationCommit.hash)
+        git cherry-pick $commitHash # we expect this to fail with conflicts
+
+        # Commit and push all changes
+        git add .
+        git commit -m "Add deprecation comment changes with all merge conflicts"
+        git push --set-upstream https://$(GITHUBTOKEN)@github.com/iTwin/itwinjs-core $conflictBranchName
+        echo "Now manually fix conflicts on branch '$conflictBranchName' and merge changes into master"
+        git fetch
+
+        # Create a PR with new conflict branch
+        gh pr create --title "Merge deprecation comment changes into master" --body "This PR contains all changes after running lint-deprecation ESLint rule and merging with master. Manual intervention required: solve conflicts in this branch and merge PR into master. This happened because deprecation comments in master and in release branch could not be merged automatically." --draft --head $conflictBranchName
+      fi
+    displayName: Perform cherry-pick operation for deprecation changes
+    continueOnError: true
+    condition: eq(variables['changesDetected'], 'true')
+
+  - bash: git checkout $(deprecationCommit.hash)
+    displayName: Checkout back to branch that ran the pipeline
+    condition: eq(variables['changesDetected'], 'true')


ben-polinsky (approved):

aruniverse:
@DanRod1999 plz review

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -238,6 +238,9 @@ jobs:
           git commit -m "$(getVersion.version)" --author="imodeljs-admin <imodeljs-admin@users.noreply.github.com>"
         displayName: Commit version bump
 
+      - ${{ if or(eq(parameters.BumpType, 'minor'), eq(parameters.BumpType, 'patch')) }}:
```
paulius-valiunas, line 241: What about major versions? Don't we need to run the rule for them too?

DanRod1999:
in test three what caused the cherry pick to have conflicts. Also who is notified when conflicts happened and a PR is opened? Is someone auto assigned to the PR?

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+  - bash: |
+      commitHash=$(deprecationCommit.hash)
+      echo "Attempting to cherry-pick commit: $commitHash"
+      if git cherry-pick $commitHash; then
+        git push https://$(GITHUBTOKEN)@github.com/iTwin/itwinjs-core HEAD:master
+        echo "Successfully applied deprecation comment updates to master"
+      else
+        # Merge conflicts: Create branch, set up PR for manual resolution.
+        echo "Cherry-pick operation on master failed (most likely due to merge conflicts). Manual intervention required"
```
paulius-valiunas: Add `##[warning]` to the beginning of this message, that way we can clearly see it when looking at the pipeline.

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
```diff
@@ -0,0 +1,64 @@
...
+        # Create and checkout to conflict handling branch
+        conflictBranchName="cherry-pick-conflict-$(date +%y%m%d%H%m%s)"
+        echo "Creating new branch '$conflictBranchName' for manual resolution"
+        git checkout -b $conflictBranchName
+
+        # Cherry-pick changes onto conflict handling branch
+        git cherry-pick $commitHash
+
+        # Commit and push all changes
+        git add .
```
paulius-valiunas, line 61: I assume you're using `git add` to add the conflicting files? The comment doesn't make that clear.

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
```diff
@@ -0,0 +1,64 @@
...
+        git commit -m "Add deprecation rule changes with all merge conflicts"
+        git push --set-upstream https://$(GITHUBTOKEN)@github.com/iTwin/itwinjs-core $conflictBranchName
+        echo "Now manually fix conflicts on branch '$conflictBranchName' and merge changes into master"
```
paulius-valiunas: Same here, add the warning tag and maybe change the wording a bit because people might not read these messages in sequence.

DanRod1999:
Changes look fine, and you're tests look good.
My only concern is that this is a pretty hefty addition to this pipeline. Does it seem rational to make this its own pipeline? And instead of a template that is called by this pipeline, we create a new azure pipeline (or gh workflow, which ever is easier/more appropriate) and then trigger this new pipeline from this version bump script. Might look something like this:
```
          az pipelines run \
            --branch $(Build.SourceBranch) \  // Not sure what branch is appropriate here, but I assume source branch would work
            --id  8467 \                                    // This id or name would be whatever we create it as
            --parameters "BumpType={{ parameter.bumpType }}" ... ect
```
Some advantages of adding this as its own pipeline are:
- should be easier to debug any problems if something goes wrong since we don't need to look into the version bump script
- can run pipeline job independently from original bump pipeline script if something goes wrong (might not be useful for this scripts intent?)

paulius-valiunas:
> My only concern is that this is a pretty hefty addition to this pipeline. Does it seem rational to make this its own pipeline?

@DanRod1999 I think our release process is already hard to follow and there are too many pipelines. It's confusing when one pipeline run triggers another, ideally I would just have one. We can have different pipelines for different use cases, though (like different release types).

On the other hand, while this script is new and until we build some trust around it, maybe we could add a manual validation step before actually pushing this to npm (unless it's already there). Of course, the validation would be skipped if the script made no changes.

common/config/azure-pipelines/jobs/version-bump.yaml
DanRod1999: major version doesn't exist as a bump type. When preforming a release for a new major the releaser runs this under minor I think (this script is weird). @hl662 can you verify i didn't make this up?

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
```diff
@@ -0,0 +1,64 @@
...
+    name: getNewVersion
+
+  - bash: node common/scripts/install-run-rush.js lint-deprecation
+    displayName: rush lint-deprecation
+
+  - bash: git add .
+    displayName: git add .
+
+  - bash: |
+      git commit -m "Apply deprecation date rule for v$(getNewVersion.version)"
```
paulius-valiunas, line 27: Have you tested this with no new deprecations (where the rule didn't make any changes)? Won't the pipeline error out if the staging area is empty?

common/config/azure-pipelines/jobs/version-bump.yaml
hl662: confirmed, our internal ADO pipeline doesn't expose a `major` option , we leave it to the admins of the repo to create the version...?

paulius-valiunas: that's what I suspected, just making sure.

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
DanRod1999: i thought this was his test 2 situation. If not then I misunderstood

DanRod1999:
> maybe we could add a manual validation step before actually pushing this to npm (unless it's already there). Of course, the validation would be skipped if the script made no changes.

this will run before it is published to npm. the workflow is
- bump version (the pipeline edited in this PR)
- build code
- publish to npm
(this are all separate pipelines, which as you mentioned, does make things a bit confusing)

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
paulius-valiunas: Oh, you're right! I just skimmed over the screenshots and saw they all had "Apply deprecation date rule" commits. Apparently it just commits the new version number in the eslint config, so it's not an empty commit.

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
```diff
@@ -0,0 +1,64 @@
...
+        # Create a PR with new conflict branch
+        gh pr create --title "Merge deprecation comment changes into master" --body "This PR contains all changes after running lint-deprecation ESLint rule and merging with master. Manual intervention required: solve conflicts in this branch and merge PR into master. This happened because deprecation comments in master and in release branch could not be merged automatically." --draft --head $conflictBranchName
+      fi
+    displayName: Perform cherry-pick operation for deprecation changes
+    continueOnError: true
+    env:
+      GH_TOKEN: $(GITHUBTOKEN)
+
+  - bash: git checkout $(deprecationCommit.hash)
+    displayName: Checkout back to branch that ran the pipeline
```
paulius-valiunas, line 75: Ignore the fact this is on line 64, I just needed to create a thread so I picked a random line.

Anyway... Looks like finalize release ran on the wrong branch here?
image

paulius-valiunas: ok I think I finally got it - finalize release is triggered by changelog files being pushed, and since its only purpose is to cherry pick those files onto master, we don't really care if that's the first commit or the second one, as long as it has the changelogs.

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -238,6 +238,9 @@ jobs:
           git commit -m "$(getVersion.version)" --author="imodeljs-admin <imodeljs-admin@users.noreply.github.com>"
```
paulius-valiunas, line 238: Another random line comment...

I think the npm publish step (in `templates/publish.yaml`, which is called from `jobs/fast-ci.yaml`) would run on the wrong commit, the one before you push eslint changes. Maybe you can change the check to look up the git tag or something else instead of looking at the commit message.

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
DanRod1999: not sure how this fork is set up. But the finalize release workflow is triggered by the release branch getting bumped. And then it pushes those changes to the next major release, or master branch. Its made to cascade up all active release branches until it hits master. Ex
new patch on 4.5.5 creates change logs
Finalize release triggered on release/4.5.x
push change logs to release/5.0.x
Finalize release is triggered on release/5.0.x
push change logs to master

augustasgri:
> in test three what caused the cherry pick to have conflicts.

The conflicts were introduced manually. I had to modify either the master or release branch before running the script to simulate a cherry-pick failure.
In theory, these conflicts shouldn't occur, since developers shouldn't need to edit existing `deprecated` comments. However, in practice, it's likely that someone might unintentionally or intentionally edit these comments, for example by moving code around. I'm not sure how well cherry-pick handles such cases automatically, so we decided to automate the conflict resolution process as much as possible.

> who is notified when conflicts happened and a PR is opened?

After discussing it with Paulius, we decided to tag Arun in PR description. The PR notification will be sent, but it might end up in spam. At least there is going to be some record of the PR. But we expect the developer who is running this pipeline to notice the conflicts, fix them and merge the PR. I know this expectation is currently not very well documented and mainly spread by word. If you have any ideas of how to inform people of this let me know.

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
augustasgri: Added, now the messages are highlighted

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
augustasgri: Changed comment to make code more clear

common/config/azure-pipelines/templates/run-eslint-deprecation.yaml
augustasgri: Added warning tag and changed the wording

augustasgri:
> On the other hand, while this script is new and until we build some trust around it, maybe we could add a manual validation step before actually pushing this to npm (unless it's already there). Of course, the validation would be skipped if the script made no changes.

Added a manual validation step to `fast-ci.yaml` pipeline. This step triggers only if eslint command made changes during version bump process. The validation step times out after 3 days. Once this step is triggered Arun is notified (however, the developer running the pipeline should notice this job and approve themselves).

common/config/azure-pipelines/jobs/version-bump.yaml
augustasgri: Rewrote some logic regarding which commit does `publish.yaml` runs on. I changed the logic to use [git notes](https://git-scm.com/docs/git-notes) instead of commit messages. The new logic works like this:

1. `version-bump.yaml` pipeline adds git notes to commits. Currently there are two notes that a commit could have:
    - SHOULD_PUBLISH - commits with this note are recognized in `fast-ci.yaml` script as version bumps and they will be published
    - DEPRECATION_COMMENT_CHANGES - commits with this note inform that eslint rule made deprecation comment changes and the "Manual Validation" job will be run.

Note: with the current implementation `DEPRECATION_COMMENT_CHANGES` note could only be added to a commit that is already marked with `SHOULD_PUBLISH`. There won't be a case where a commit has only `DEPRECATION...` note because there is not need validating anything if nothing is being published

2. `fast-ci.yaml` parses git notes and:
    - waits for manual validation if commit has `DEPRECATION_COMMENT_CHANGES` note set
    - after approval, releases git commit if it has `SHOULD_PUBLISH` note set

paulius-valiunas:
> > On the other hand, while this script is new and until we build some trust around it, maybe we could add a manual validation step before actually pushing this to npm (unless it's already there). Of course, the validation would be skipped if the script made no changes.
> 
> Added a manual validation step to `fast-ci.yaml` pipeline. This step triggers only if eslint command made changes during version bump process. The validation step times out after 3 days. Once this step is triggered Arun is notified (however, the developer running the pipeline should notice this job and approve themselves).

@aruniverse ^ (see both comments above)

paulius-valiunas (approved):

aruniverse:
> > > On the other hand, while this script is new and until we build some trust around it, maybe we could add a manual validation step before actually pushing this to npm (unless it's already there). Of course, the validation would be skipped if the script made no changes.
> > 
> > 
> > Added a manual validation step to `fast-ci.yaml` pipeline. This step triggers only if eslint command made changes during version bump process. The validation step times out after 3 days. Once this step is triggered Arun is notified (however, the developer running the pipeline should notice this job and approve themselves).
> 
> @aruniverse ^ (see both comments above)

Im flattered, but I shouldnt the bottleneck here... Can we assign `@iTwin/itwinjs-core-admins` to those PRs?

common/changes/@itwin/presentation-backend/augustas-eslint-deprecation_2025-07-14-07-02.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-backend",
+      "comment": "Added 'lint-deprecation' ESLint command to fix deprecation comments",
```
saskliutas: Maybe `comment` in change files for these changes should be empty? Not sure this message has any meaning to package consumer.

saskliutas (approved):

common/changes/@itwin/presentation-backend/augustas-eslint-deprecation_2025-07-14-07-02.json
augustasgri: Yes, made the `comment` part an empty string

common/config/eslint/eslint.config.base.js
```diff
@@ -10,5 +10,17 @@ module.exports = [
     rules: {
       "@typescript-eslint/no-unsafe-enum-comparison": "off"
     }
-  }
+  },
+  {
+    files: ["**/*.ts"],
+    rules: {
+      "@typescript-eslint/no-non-null-assertion": "warn"
```
pmconne, line 17: Why? I lost count of how many bugs/crashes in production were traced back to somebody doing `thingThatMightBeNull!.doSomething()`. It should be an error.

augustasgri: Yes, but prior to me bumping the `@itwin/eslint-plugin` version the rule was off. Now that I bumped the version it was seen as an error. There were hundreds of these errors and build pipeline was failing because `rush lint` was failing. So I at least reduced the error to a warning. That way these warnings could get fixed with time

pmconne: Dunno how it got turned off, I remember insisting we turn it on. Please file an issue to turn it back on as an error and fix all the violations.

augustasgri: Ok, made the issue: https://github.com/iTwin/itwinjs-core/issues/8339

Regarding the rule, I think it happened like this:
- Rule got turned on in [this commit](https://github.com/iTwin/eslint-plugin/commit/b2ed3d1f1e1356a2bace0add67a83270f32b82ce)
- New dev version of `@itwin/eslint-plugin` [was published](https://github.com/iTwin/eslint-plugin/commit/98ac62fda2bc400e321ae8da24fb0be4016cf6cb)
- But the version of eslint plugin remained the same inside in itwinjs-core so it was not "applied". Since I bumped the version in this PR that rule also activated

pmconne (approved):

wgoehrig (approved):

PR #8273, opened 2025-06-25
Add changeset ID verification before pushing

khanaffan:
Introduce a validation step to verify the changeset ID before pushing, ensuring the consistency and accuracy of the changeset data.

This update addresses a customer-reported issue where the changeset ID sent to the hub was not the actual SHA1 hash of the changeset file. Although the issue could not be reproduced, this safeguard helps prevent similar inconsistencies in the future.

### DIFF
--- a/core/backend/src/BriefcaseManager.ts
+++ b/core/backend/src/BriefcaseManager.ts
@@ -10,7 +10,7 @@
 
 import * as path from "path";
 import {
-  AccessToken, BeDuration, ChangeSetStatus, GuidString, IModelHubStatus, IModelStatus, Logger, OpenMode, Optional, StopWatch,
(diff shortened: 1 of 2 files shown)


core/backend/src/BriefcaseManager.ts
```diff
@@ -609,6 +609,11 @@ export class BriefcaseManager {
       throw new IModelError(IModelStatus.NoContent, "error creating changeset");
 
     changesetProps.size = fileSize;
+    const id = IModelNative.platform.DgnDb.computeChangesetId(changesetProps);
+    if (id !== changesetProps.id) {
+      Logger.logWarning(loggerCategory, `Changeset id ${changesetProps.id} does not match computed id ${id}.`);
```
aruniverse: is warning sufficient here?

khanaffan: I changed it to exception

aruniverse (approved):

Issue #7830, opened 2025-03-12
Text Leaders


Josh-Schifter:
Related to Issue: [TextBox](https://github.com/iTwin/itwinjs-core/issues/7718)

### Summary:
Propose to add leaders as an optional feature of TextAnnotation2d/3d elements.  This proposal aims to avoid the problems we had with MicroStation 'Notes' which were made of multiple related elements.

![Image](https://github.com/user-attachments/assets/f8fe86e6-18df-4966-962e-fba41245e3ee)

### Detailed proposal:
* Add a new property to TextAnnotation2d/3d 'Leaders' to store properties related to array of leaders.
* Each member of the array would store information describing a leader (see below)
* Store leader styling information in Text Style.
* TextAnnotation2d/3d geometry stream should include the leader geometry

### For each leader, need to store:
* Attachment point stored as a string (ex. "MiddleRight").  This will be related to the shape of the Text Frame.
* Target point stored as Point3d.
* Overrides for styling properties such as color, terminator shape and size, curve type (ex. line, curve, etc.)
* ElementDrivesElement relationship to the target element.

The Drawing Production team would like to do the implementation.

@pmconne,  @jffmarker, @claudiaareneee , @SheetalNair , @Alfonso-Martello

Josh-Schifter:
How I see the pros and cons of this proposal vs. the alternative.

Option 1 (this proposal): A single element combining text, frame, and leaders:

<table>
<tr>
<th> Pro </th>
<th> Con </th>
</tr>
<tr>
<td>

* Geometry generation is centralized
* Easier to implement manipulation dynamics
* Only need relationship to ‘driving’ elements, not between leader and text
* Clear relationship between leader and Text Style

</td>
<td>

* One text element ‘driven’ by N other elements
* Text element has to store more information [leader attachment point, association info, additional data for multisegment leaders]
* All code changes for leaders have to be done in core packages.

</td>
</tr>
</table>

Option 2 (alternative to this proposal): A new element for just a single leader:

<table>
<tr>
<th> Pro </th>
<th> Con </th>
</tr>
<tr>
<td>

* Each leader only has two relationships (no more)
* Decoupled

</td>
<td>

pmconne:
> Attachment point stored as a string (ex. "MiddleRight"). This will be related to the shape of the Text Frame.

So users can't choose any point on the frame shape / bounding box - they are restricted to a small set of options?
[...]

PR #8278, opened 2025-06-25
Text Annotation Leaders

SheetalNair:
This PR only consists of code to create leader geometry and store it in Text Annotation, the later part of creating relationships  between leader and target elements is not yet addressed here. Find details in this issue #7830 
I have introduced attachmentMode property for leaders which provides user with a set of options to select attachment point:
- Nearest : leader will attach to the point on frame closest to the target point.
image

- KeyPoint: User can provide the curveIndex and fraction of the frame as an attachment point (See image below: curveIndex 0, fraction 0)
image

- TextPoint: User can attach the leader either to the first or last line of text (See image below: position:"TopLeft" indicating first line of text)
image

Code changes:
- `LeaderGeometry.ts` : contains functions to generate leader geometry and append it to builder
- `TextAnnotation.ts`: Introduced leaderprops inside TextAnnotation class
- `TextAnnotationGeometry.ts`: Call to appendLeadersToBuilder()
- `TextStyle.ts` : Introduced leader style props [Current decision is for leader style to be a part of TextStyle]
- `TextStyle.test.ts`: fixed failing test case after leader style introduction
- `TextDecoration.ts` : Keyins for testing leaders


test-apps/display-test-app/src/frontend/TextDecoration.ts
```diff
@@ -135,6 +138,28 @@ class TextEditor implements Decorator {
...
+    leader.startPoint = point.plus(directionVector);
+  }
+
+  public setLeaderKeyPoint(leader: TextAnnotationLeader, curveIndex: number, fraction: number) {
+    leader.attachmentMode = { mode: "KeyPoint", curveIndex, fraction };
+  }
+  public setLeaderTextPoint(leader: TextAnnotationLeader, arg: LeaderTextPointOptions) {
+    leader.attachmentMode = { mode: "TextPoint", position: arg };
+  }
+  public setLeaderNearest(leader: TextAnnotationLeader,) {
```
claudiaareneee:
```suggestion
  public setLeaderNearest(leader: TextAnnotationLeader) {
```

Accidentally left in a comma :)

test-apps/display-test-app/src/frontend/TextDecoration.ts
```diff
@@ -363,6 +388,33 @@ export class TextDecorationTool extends Tool {
...
+              editor.setLeaderTextPoint(editor.leaders[latestLeaderIndex], position);
+
+            }
+            else throw new Error("Expected start, keypoint, nearest, textpoint");
+          } else {
+            throw new Error("No leaders created. Use dta text leader new.");
+          }
+
+        }
+        break;
```
claudiaareneee:
```suggestion
      case "leader":
        const attachmentType = inArgs[1];
        const value = inArgs[2];
        if (attachmentType === "new") {
          editor.setLeaderProps();
        } else {
          if (editor.leaders && editor.leaders.length > 0) {
            const latestLeaderIndex = editor.leaders.length - 1;
            if (attachmentType === "start") editor.setLeaderStartPoint(editor.leaders[latestLeaderIndex], Number(value));
            else if (attachmentType === "keypoint") {
              const curveIndex = inArgs[2];
              const fraction = inArgs[3]
              editor.setLeaderKeyPoint(editor.leaders[latestLeaderIndex], Number(curveIndex), Number(fraction));
            }
            else if (attachmentType === "nearest") editor.setLeaderNearest(editor.leaders[latestLeaderIndex]);
            else if (attachmentType === "textpoint") {
              const position = inArgs[2] as LeaderTextPointOptions;
              editor.setLeaderTextPoint(editor.leaders[latestLeaderIndex], position);

            }
            else throw new Error("Expected start, keypoint, nearest, textpoint");
          } else {
            throw new Error("No leaders created. Use dta text leader new.");
          }

        }
        break;
```

test-apps/display-test-app/src/frontend/TextDecoration.ts
```diff
@@ -3,11 +3,11 @@
...
+import { BaselineShift, ColorDef, FractionRun, LeaderTextPointOptions, LineBreakRun, Placement2dProps, TextAnnotation, TextAnnotationAnchor, TextAnnotationFrameShape, TextAnnotationLeader, TextAnnotationProps, TextBlock, TextBlockJustification, TextBlockMargins, TextFrameStyleProps, TextRun, TextStyleSettingsProps } from "@itwin/core-common";
 import { DecorateContext, Decorator, GraphicType, IModelApp, IModelConnection, readElementGraphics, RenderGraphicOwner, Tool } from "@itwin/core-frontend";
 import { DtaRpcInterface } from "../common/DtaRpcInterface";
 import { Id64, Id64String } from "@itwin/core-bentley";
-import { Point3d, YawPitchRollAngles } from "@itwin/core-geometry";
+import { Angle, Point3d, Vector3d, YawPitchRollAngles } from "@itwin/core-geometry";
 
 // Ignoring the spelling of the keyins. They're case insensitive, so we check against lowercase.
 // cspell:ignore superscript, subscript, widthfactor, fractionscale, fractiontype
```
claudiaareneee:
```suggestion
// cspell:ignore superscript, subscript, widthfactor, fractionscale, fractiontype, textpoint
```

core/common/src/annotation/TextStyle.ts
```diff
@@ -88,6 +88,29 @@ export interface TextStyleSettingsProps {
...
+  /** Multiplier to compute height of the leader terminator.
+   * The terminator height is computed as terminatorHeight * [[lineHeight]].
+   * Default: 1.0
+   */
+  terminatorHeight?: number;
+  /** Multiplier to compute width of the leader terminator.
+   * The terminator width is computed as terminatorWidth * [[lineHeight]].
+   * Default: 1.0
+   */
+  terminatorWidth?: number;
```
claudiaareneee:
```suggestion
  /** Multiplier used to compute length of the elbow in the leader line.
   * The elbowLength is computed in meters as elbowLength * [[lineHeight]].
   * Default: 1.0
   */
  elbowLength?: number;
  /** Multiplier to compute height of the leader terminator.
   * The terminator height is computed in meters as terminatorHeight * [[lineHeight]].
   * Default: 1.0
   */
  terminatorHeight?: number;
  /** Multiplier to compute width of the leader terminator.
   * The terminator width is computed in meters as terminatorWidth * [[lineHeight]].
   * Default: 1.0
   */
  terminatorWidth?: number;
```

core/common/src/annotation/TextStyle.ts
```diff
@@ -141,6 +164,22 @@ export class TextStyleSettings {
...
+  */
+  public readonly elbowLength: number;
+  /** Multiplier to compute height of the leader terminator.
+   * The terminator height is computed as terminatorHeight * [[lineHeight]].
+  */
+  public readonly terminatorHeight: number;
+  /** Multiplier to compute width of the leader terminator.
+   * The terminator width is computed as terminatorWidth * [[lineHeight]].
+  */
+  public readonly terminatorWidth: number;
```
claudiaareneee:
```suggestion
  /** Multiplier used to compute length of the elbow in the leader line.
   * The elbowLength is computed in meters as elbowLength * [[lineHeight]].
  */
  public readonly elbowLength: number;
  /** Multiplier to compute height of the leader terminator.
   * The terminator height is computed in meters as terminatorHeight * [[lineHeight]].
  */
  public readonly terminatorHeight: number;
  /** Multiplier to compute width of the leader terminator.
   * The terminator width is computed in meters as terminatorWidth * [[lineHeight]].
  */
  public readonly terminatorWidth: number;
```

core/common/src/annotation/TextStyle.ts
claudiaareneee: Also maybe these should be renamed with a `Factor` at the end to maintain the consistency with the other multiplier settings? Right now the sound like the explicit width/length. That might make the attributes a bit long though. They could be nested inside a `terminator` variable? I know @Alfonso-Martello is working on styles right now so maybe I should just wait for those comments, idk.

core/common/src/annotation/TextStyle.ts
```diff
@@ -204,7 +253,9 @@ export class TextStyleSettings {
       && this.isBold === other.isBold && this.isItalic === other.isItalic && this.isUnderlined === other.isUnderlined
       && this.stackedFractionType === other.stackedFractionType && this.stackedFractionScale === other.stackedFractionScale
       && this.subScriptOffsetFactor === other.subScriptOffsetFactor && this.subScriptScale === other.subScriptScale
-      && this.superScriptOffsetFactor === other.superScriptOffsetFactor && this.superScriptScale === other.superScriptScale;
+      && this.superScriptOffsetFactor === other.superScriptOffsetFactor && this.superScriptScale === other.superScriptScale && this.leaderColor === other.leaderColor
+      && this.wantElbow === other.wantElbow && this.elbowLength === other.elbowLength
+      && this.terminatorHeight === other.terminatorHeight && this.terminatorWidth === other.terminatorWidth;
```
claudiaareneee: Maybe could move that one line down? 

```suggestion
      && this.superScriptOffsetFactor === other.superScriptOffsetFactor 
      && this.superScriptScale === other.superScriptScale && this.leaderColor === other.leaderColor
      && this.wantElbow === other.wantElbow && this.elbowLength === other.elbowLength
      && this.terminatorHeight === other.terminatorHeight && this.terminatorWidth === other.terminatorWidth;
```

core/backend/src/annotations/LeaderGeometry.ts
```diff
@@ -0,0 +1,183 @@
...
+ * @returns `true` if at least one leader with a terminator was successfully appended; otherwise, `false`.
+ * @beta
+ */
+export function appendLeadersToBuilder(builder: ElementGeometry.Builder, leaders: TextAnnotationLeader[], layout: TextBlockLayout, transform: Transform, params: GeometryParams, frame?: TextFrameStyleProps,): boolean {
+  // If there is no frame, use a rectangular frame to compute the attachmentPoints for leaders.
+  if (frame === undefined || frame.shape === "none") {
+    frame = { shape: "rectangle" }
+  }
+  if (frame.shape === undefined || frame.shape === "none") return false;
+  const frameCurve = computeFrame({ frame: frame.shape, range: layout.range, transform });
```
claudiaareneee, line 39: I would pass this in to avoid duplicate calls to `computeFrame`. it would also remove a dependency to another file which could be good.

core/backend/src/annotations/LeaderGeometry.ts
```diff
@@ -0,0 +1,183 @@
...
+ * @param leaders - An array of leader properties.
+ * @param layout - The layout information for the text block, including its range.
+ * @param transform - The transform to apply to the frame and leader geometry.
+ * @param params - The geometry parameters, such as color, to use for the leader lines.
+ * @param frame - (Optional) The style properties for the text frame. If not provided or set to "none", a default rectangle is used.
+ * @returns `true` if at least one leader with a terminator was successfully appended; otherwise, `false`.
+ * @beta
+ */
+export function appendLeadersToBuilder(builder: ElementGeometry.Builder, leaders: TextAnnotationLeader[], layout: TextBlockLayout, transform: Transform, params: GeometryParams, frame?: TextFrameStyleProps,): boolean {
+  // If there is no frame, use a rectangular frame to compute the attachmentPoints for leaders.
```
claudiaareneee:
```suggestion
export function appendLeadersToBuilder(builder: ElementGeometry.Builder, leaders: TextAnnotationLeader[], layout: TextBlockLayout, transform: Transform, params: GeometryParams, frame?: TextFrameStyleProps): boolean {
  let result = true;

  // If there is no frame, use a rectangular frame to compute the attachmentPoints for leaders.
```

core/backend/src/annotations/LeaderGeometry.ts
```diff
@@ -0,0 +1,183 @@
...
+    frame = { shape: "rectangle" }
+  }
+  if (frame.shape === undefined || frame.shape === "none") return false;
+  const frameCurve = computeFrame({ frame: frame.shape, range: layout.range, transform });
+
+  leaders.forEach((leader) => {
+    if (leader.styleOverrides?.leaderColor !== "subcategory" && leader.styleOverrides?.color !== "subcategory") {
+      const color = leader.styleOverrides?.leaderColor ?? leader.styleOverrides?.color;
+      params.lineColor = color ? ColorDef.fromJSON(color) : ColorDef.black;
+      builder.appendGeometryParamsChange(params);
```
claudiaareneee:
```suggestion
      result = result && builder.appendGeometryParamsChange(params);
```

core/backend/src/annotations/LeaderGeometry.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (leader.styleOverrides?.wantElbow) {
+      const elbowLength = (leader.styleOverrides.elbowLength ?? 1) * (leader.styleOverrides.lineHeight ?? 1)
+      const elbowDirection = computeElbowDirection(attachmentPoint, frameCurve, elbowLength);
+      if (elbowDirection)
+        leaderLinePoints.push(attachmentPoint.plusScaled(elbowDirection, elbowLength))
+    }
+
+    leaderLinePoints.push(attachmentPoint)
+
+    builder.appendGeometryQuery(LineString3d.create(leaderLinePoints));
```
claudiaareneee:
```suggestion
    result = result && builder.appendGeometryQuery(LineString3d.create(leaderLinePoints));
```

core/backend/src/annotations/LeaderGeometry.ts
```diff
@@ -0,0 +1,183 @@
...
+    ).normalize();
+
+    const termY = terminatorDirection?.unitCrossProduct(Vector3d.unitZ());
+    if (!termY || !terminatorDirection) return true; // Assuming leaders without terminators is a valid case.
+    const terminatorHeight = (leader.styleOverrides?.terminatorHeight ?? 1) * (leader.styleOverrides?.lineHeight ?? 1);
+    const terminatorWidth = (leader.styleOverrides?.terminatorWidth ?? 1) * (leader.styleOverrides?.lineHeight ?? 1);
+    const basePoint = leader.startPoint.plusScaled(terminatorDirection, terminatorWidth);
+    const termPointA = basePoint.plusScaled(termY, terminatorHeight);
+    const termPointB = basePoint.plusScaled(termY.negate(), terminatorHeight);
+    builder.appendGeometryQuery(LineString3d.create([termPointA, leader.startPoint, termPointB]));
```
claudiaareneee:
```suggestion
    result = result && builder.appendGeometryQuery(LineString3d.create([termPointA, leader.startPoint, termPointB]));
```

core/backend/src/annotations/LeaderGeometry.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    return true;
```
claudiaareneee:
```suggestion
    return result;
```

claudiaareneee (commented):
I left some suggestions. Also I have a feeling that tests are going to be asked for

core/backend/src/annotations/LeaderGeometry.ts
claudiaareneee: (Suggestion is to remove the hanging comma in the function declaration and add a variable to track appending to the builder)

test-apps/display-test-app/src/frontend/TextDecoration.ts
claudiaareneee: Can you rename `k` to something more explicit. Maybe `attachmentType` or something?

My suggestion above uses `attachmentType`

core/common/src/annotation/TextStyle.ts
SheetalNair: Appended the variable name with 'Factor', we had a discussion some time back about not nesting properties in style.

core/common/src/annotation/TextStyle.ts
SheetalNair: done

test-apps/display-test-app/src/frontend/TextDecoration.ts
SheetalNair: all keys are not attachment types so I used the word 'command'

common/api/core-common.api.md
```diff
@@ -10105,13 +10147,15 @@ export class TextStyleSettings {
     readonly color: TextStyleColor;
     static defaultProps: Readonly<Required<TextStyleSettingsProps>>;
     static defaults: TextStyleSettings;
+    readonly elbowLength: number;
```
pmconne: I think it would be helpful to define an interface type that contains all of the leader-related settings, and give TextStyleSettings a single property of that type. Mixing them all in as top-level properties makes it very difficult to tell which are related to leaders, difficult to write APIs that operate on the leader-specific properties as a unit, and make conflicts/confusion likely in future (for example, if we add new annotations for which elbows are relevant).

common/api/core-common.api.md
```diff
@@ -9881,10 +9898,35 @@ export type TextAnnotationFillColor = TextStyleColor | "background";
...
+    startPoint: Point3d;
+    // (undocumented)
+    styleOverrides?: TextStyleSettingsProps;
+}
+
+// @beta
+export interface TextAnnotationLeaderProps {
+    // (undocumented)
+    attachmentMode: LeaderAttachmentMode;
+    // (undocumented)
```
pmconne: Please document all undocumented APIs.

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -65,6 +65,45 @@ export interface TextFrameStyleProps {
   borderWeight?: number;
 }
 
+/**
+ * Specifies where to attach a leader line to a [[TextAnnotation]].
+ * The leader line can be attached to a key point, to the first or last line of text, or to the nearest point on the text block frame.
+ * @beta
+ */
+export type LeaderAttachmentMode =
```
pmconne:
```suggestion
export type LeaderAttachment =
```

This type **has** a `mode`, it is not itself a mode.

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -65,6 +65,45 @@ export interface TextFrameStyleProps {
...
+  | { mode: "Nearest" };
+
+/**
+ * Specifies the possible positions for the leader attachment point relative to the annotation when attachmentMode is TextPoint.
+ * @beta
+ */
+export type LeaderTextPointOptions = "TopLeft" | "TopRight" | "BottomLeft" | "BottomRight"
+
+/**
+ * Properties of a leader line attached to a [[TextAnnotation]].
```
pmconne: Please elaborate upon what a leader line actually is, how it attaches to an annotation, etc.

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -65,6 +65,45 @@ export interface TextFrameStyleProps {
...
+ * The leader line can be attached to a key point, to the first or last line of text, or to the nearest point on the text block frame.
+ * @beta
+ */
+export type LeaderAttachmentMode =
+  | { mode: "KeyPoint"; curveIndex: number; fraction: number }
+  | { mode: "TextPoint"; position: LeaderTextPointOptions }
+  | { mode: "Nearest" };
+
+/**
+ * Specifies the possible positions for the leader attachment point relative to the annotation when attachmentMode is TextPoint.
```
pmconne: I assume the leader attaches to the annotation's frame. How does this work when the frame is not rectangular? e.g., what does `TopLeft` mean if the frame is a triangle?

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -65,6 +65,45 @@ export interface TextFrameStyleProps {
   borderWeight?: number;
 }
 
+/**
+ * Specifies where to attach a leader line to a [[TextAnnotation]].
+ * The leader line can be attached to a key point, to the first or last line of text, or to the nearest point on the text block frame.
```
pmconne: > to the first or last line of text

It's unclear to me given the current documentation how this works.

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -235,16 +294,40 @@ export class TextAnnotation {
...
+  }
+
   /** Returns true if this annotation is logically equivalent to `other`. */
   public equals(other: TextAnnotation): boolean {
     const framesMatch = this.frame?.shape === other.frame?.shape
       && this.frame?.fill === other.frame?.fill
       && this.frame?.border === other.frame?.border
       && this.frame?.borderWeight === other.frame?.borderWeight;
 
+    const leadersMatch = this.areLeadersEqual(this.leaders, other.leaders);
```
pmconne: Do the fast and easy comparisons first, returning early if they don't match. Then only compare frames if necessary, and finally only compare leaders if necessary.

pmconne (changes requested):
You don't appear to have any test coverage for the code in LeaderGeometry.ts.

claudiaareneee:
> You don't appear to have any test coverage for the code in LeaderGeometry.ts.

I didn't have any tests for FrameGeometry either. 

I made a PR for that here: #8305 if you wanted to copy anything I did with that @SheetalNair.

common/api/core-common.api.md
SheetalNair: Introduced interface TextLeaderStyleProps to group leader related settings.  Earlier the properties were laid out flat in the settings object for simplicity when copying and comparing, but it makes sense to group it in an interface considering the scenario you mentioned about new annotations for which same name properties are relevant.

common/api/core-common.api.md
SheetalNair: done

core/common/src/annotation/TextAnnotation.ts
SheetalNair: done

core/common/src/annotation/TextAnnotation.ts
SheetalNair: added more detailed description

core/common/src/annotation/TextAnnotation.ts
SheetalNair: TopLeft basically means a point on frame projected from a point on TopLeft i.e Left side of the first line of text in the textBlock. The TextPoint attachmentMode calculates attachment point w.r.t to lines of text in the text block. I know the name seems misleading but TopLeft actually refers to the point corresponding to line of text, not the TopLeft of the frame. I have added some elaboration in the documentation, please let me know if it makes sense.

core/common/src/annotation/TextAnnotation.ts
SheetalNair: updated documentation

core/common/src/annotation/TextAnnotation.ts
SheetalNair: modified the existing code a bit to get through easier comparisons first.

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -69,6 +69,69 @@ export interface TextFrameStyleProps {
   borderWeight?: number;
 }
 
+/**
+ * Describes how to attach a leader to a frame around [[textBlock]].
```
pmconne:
```suggestion
 * Describes how to attach a [[TextAnnotationLeader]] to the frame around a [[TextBlock]].
```

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -69,6 +69,69 @@ export interface TextFrameStyleProps {
...
+ * - Nearest: attach to the point on frame that is nearest to [[TextAnnotationLeader.startPoint]].
+ * @beta
+ */
+export type LeaderAttachment =
+  | { mode: "KeyPoint"; curveIndex: number; fraction: number }
+  | { mode: "TextPoint"; position: LeaderTextPointOptions }
+  | { mode: "Nearest" };
+
+/**
+ * Specifies the possible positions to attach a leader on the frame around [[textBlock]]
```
pmconne:
```suggestion
 * Specifies the possible positions to attach a leader on the frame around a [[TextBlock]]
```

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -69,6 +69,69 @@ export interface TextFrameStyleProps {
...
+ * @beta
+ */
+export type LeaderTextPointOptions = "TopLeft" | "TopRight" | "BottomLeft" | "BottomRight"
+
+/**
+ * Properties of a leader in a [[TextAnnotation]].
+ * A leader is a piece of geometry that points from the [[TextAnnotation]] to a point in space attached to another element in the iModel.
+ * A leader is always attached to the frame around the [[textBlock]].
+ * If there is no frame around the textBlock, we assume a rectangular frame and attach the leader to it.
+ * @beta
```
pmconne:
```suggestion
 * A line that connects a [[TextAnnotation]] to a point in space relative to another element in the iModel.
 * A leader is always attached to the frame around the annotation's [[TextBlock]].
 * If the frame is not visible, the leader attaches to an invisible rectangular frame around the text block.
 * @see [[TextAnnotation.leaders]] for the leaders associated with an annotation.
 * @see [[TextStyleSettings.leader]] and [[styleOverrides]] to customize the appearance of leaders.
 * @beta
```

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -69,6 +69,69 @@ export interface TextFrameStyleProps {
...
+  /** The point where the leader starts.
+   * This is the point on another element where the leader points to */
+  startPoint: Point3d;
+  /** Describes how to attach the leader to the frame around [[textBlock]].*/
+  attachment: LeaderAttachment;
+  /** Optional intermediate points that the leader should pass through.
+   * If not specified, the leader will be a straight line from startPoint to the point on the frame.
+   * For now, intermediate points are a set of points which create additional LineSegments in the leader, but there could be intermediate shapes instead of straight LineSegments in future*/
+  intermediatePoints?: Point3d[];
+  /** Optional style overrides for the leader. If not specified, the leader will use the default style from [[TextStyleSettings]].
```
pmconne: In LeaderGeometry.ts, you seem only to look at `styleOverrides`, not falling back to `TextStyle.leader`.

core/backend/src/annotations/LeaderGeometry.ts
SheetalNair: Pulling the computeFrame out of the method does not seem to reduce no. of calls to it, as frame again calls it in the appendFrame method and using  just one call for both leaders and frames is not feasible as leader uses an imaginary rectangular frame in absence of frame props

pmconne:
You have a bit of a weird situation with text styles.

A TextAnnotation contains a TextBlock and any number of leaders.
The TextBlock has a style name, and style overrides; so do each of its components.
A leader has only style overrides.
What style is the leader overriding? Surely not the one on the TextBlock, because the leader doesn't belong to the text block, it's a sibling of it.

Frames avoid this problem by defining their appearance directly instead of as part of a text style.
Should TextAnnotation have a TextStyle that all of its components (leaders and text blocks and whatever else we introduce in the future) can override?

@Josh-Schifter

core/common/src/annotation/TextAnnotation.ts
```diff
@@ -69,6 +69,70 @@ export interface TextFrameStyleProps {
...
+  /** The point where the leader starts.
+   * This is the point on another element where the leader points to */
+  startPoint: Point3d;
+  /** Describes how to attach the leader to the frame around [[textBlock]].*/
+  attachment: LeaderAttachment;
+  /** Optional intermediate points that the leader should pass through.
+   * If not specified, the leader will be a straight line from startPoint to the point on the frame.
+   * For now, intermediate points are a set of points which create additional LineSegments in the leader, but there could be intermediate shapes instead of straight LineSegments in future*/
+  intermediatePoints?: Point3d[];
+  /** Optional style overrides for the leader. If not specified, the leader will use the style defined by [[TextBlock.styleName]] as it is.
```
pmconne, line 114: Per my previous comment, it doesn't seem logical to me that the leaders use the text block's style. Have you guys discussed this internally already?

core/common/src/annotation/TextAnnotation.ts
SheetalNair: I updated the comment to specify that if overrides are not provided, the style defined by TextBlock.styleName is gonna be used.  As of now only overrides are available to specify style properties,  but soon a PR which utilizes TextBlock.styleName to fetch style properties is gonna be raised.  We did have a discussion internally to avoid having textStyle in the high level TextAnnotation for simplicity. So textBlock, frames and leaders are going to use the style specified in textBlock.

core/common/src/annotation/TextAnnotation.ts
SheetalNair: We did have a discussion internally to avoid having textStyle in the high level TextAnnotation for simplicity. So textBlock, frames and leaders are going to use the style specified in textBlock.

core/common/src/annotation/TextAnnotation.ts
pmconne: > but soon a PR which utilizes TextBlock.styleName to fetch style properties is gonna be raised

Why does that need to happen in a separate PR?

Josh-Schifter:
> You have a bit of a weird situation with text styles.
> 
> A TextAnnotation contains a TextBlock and any number of leaders. The TextBlock has a style name, and style overrides; so do each of its components. A leader has only style overrides. What style is the leader overriding? Surely not the one on the TextBlock, because the leader doesn't belong to the text block, it's a sibling of it.
> 
> Frames avoid this problem by defining their appearance directly instead of as part of a text style. Should TextAnnotation have a TextStyle that all of its components (leaders and text blocks and whatever else we introduce in the future) can override?
> 
> @Josh-Schifter

We went around and around on this.  There's a reasonable description of the though process here: https://github.com/iTwin/itwinjs-core/pull/8313

core/common/src/annotation/TextAnnotation.ts
SheetalNair: This is the PR I was referring to https://github.com/iTwin/itwinjs-core/pull/8313 . It introduces a slightly different way to store and fetch text styles, so it would be better to have those changes there.

Josh-Schifter: Because, until that PR, none of the text properties are coming from the style so we chose to address all the properties together in one PR https://github.com/iTwin/itwinjs-core/pull/8313 which has been developed in parallel to this one.

SheetalNair:
> You don't appear to have any test coverage for the code in LeaderGeometry.ts.

Added tests for LeaderGeometry.  I have introduced a AnnotationTestUtils.ts file, as there was some code from other tests that needed to be reused in my tests. please let me know if that's fine.

core/backend/src/test/annotations/LeaderGeometry.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    it("should append a leader to the builder", () => {
+      const leaders: TextAnnotationLeader[] = [
+        {
+          startPoint: Point3d.create(10, 0, 0),
+          attachment: { mode: "Nearest" }
+        }
+      ]
+      const result = appendLeadersToBuilder(builder, leaders, layout, transform, defaultParams, frame);
+      expect(result).to.be.true;
+      expect(builder.geometries.length).to.be.greaterThan(0);
```
pmconne: Here and elsewhere, be more precise. Exactly how many entries do you expect? What should their contents be? Verify that the leader attaches at the point specified by `attachment.mode`. Do you expect a `GraphicParams` entry? If so, verify the color etc are as expected. etc.

pmconne:
The geometry tests don't exercise any of the `styleOverrides` except for `wantElbow`. Verify the geometry produced with different elbow lengths and terminator width/height factors, different attachment modes, etc.

core/backend/src/test/annotations/LeaderGeometry.test.ts
SheetalNair: Added more details to tests

SheetalNair:
> The geometry tests don't exercise any of the `styleOverrides` except for `wantElbow`. Verify the geometry produced with different elbow lengths and terminator width/height factors, different attachment modes, etc.

Added more tests for different styleOverrides and different attachmentModes

PR #8286, opened 2025-06-27
Added Voronoi diagram API

saeeedtorabi:
Added a class to represent a Voronoi diagram.
* A Voronoi diagram is a partitioning of a plane into regions based on the distance to a specific set of points.
 * We construct the Voronoi diagram using Delaunay triangulation via the circumcircle algorithm.
 * More info can be found here: https://en.wikipedia.org/wiki/Voronoi_diagram and https://en.wikipedia.org/wiki/Delaunay_triangulation
 * Also added an API to take a CurveChain and return combined Voronoi cells for each child of the CurveChain 
 * Finally, a new API is added to ClipUtils to turn Voronoi cells for each child of CurveChain to clippers


pmconne (approved):

pmconne:
Why does this PR touch 75 mostly seemingly-unrelated files?

dassaf4:
> Why does this PR touch 75 mostly seemingly-unrelated files?

A lot of supporting methods....

pmconne (approved):

core/geometry/src/polyface/PolyfaceBuilder.ts
```diff
@@ -2133,3 +2138,50 @@ function resolveToIndexedXYZCollectionOrCarrier(points: Point3d[] | LineString3d
...
+   * @param graph input topology, z-coordinates ignored
+   * @param testPoint xy point(s) to search
+   * @returns edge in the face loop of the containing face, or `undefined` if `testPoint` is in an exterior face,
+   * or an array of the same in 1-1 correspondence with the input points.
+   */
+  export function findContainingFaceXY(graph: HalfEdgeGraph, testPoint: XAndY | XAndY[]): (HalfEdge | undefined)[] | HalfEdge | undefined {
+    const results: (HalfEdge | undefined)[] = [];
+    const idToFaceIndexMap = graph.constructIdToFaceIndexMap();
+    const facetIndexToFaceIndexMap: number[] = [];
+    const acceptFace = (face: HalfEdge): boolean => {
```
saeeedtorabi, line 2161: This is a bit confusing. `idToFaceIndexMap` stores node's (half-edge) id as key and min index as value. Now `acceptFace` takes each node (not each face) and pushes a number into `facetIndexToFaceIndexMap`

First of all I think you should rename `face` to `node`. Also why a number array is called Map? (`facetIndexToFaceIndexMap`) It's not a key/value map. It's just an array.

Finally, I'm confused about difference between face and facet here!

core/geometry/src/topology/Voronoi.ts
```diff
@@ -6,704 +6,736 @@
...
+      this.bbox.union(r1, this.bbox);
+    if (!this.bbox.isNull && !this.bbox.isSinglePoint) {
+      const pad = 0.02473 * (this.bbox.xLength() + this.bbox.yLength()); // ~5% of average side length
+      this.bbox.expandInPlace(pad);
     }
-    return intersections;
+  }
+  /** Checks if a point is contained within the boundary. */
+  public contains(point: XAndY): boolean {
+    return this.contains(point);
```
saeeedtorabi: how is this possible? how it's not creating an infinite recursion?!

core/geometry/src/polyface/PolyfaceBuilder.ts
dassaf4: A `HalfEdge` can represent a vertex, edge, or face depending on context. So calling it a "vertex", "edge", or "face" gives a clue as to this context, and thus I usually prefer that over the generic "node". In this light, we have for `acceptFace`:

* it is called once per graph face
* the mask it checks is a face mask
* the index it stores in the array is a face index

And these three facts suggest the name of its input should be "face".

The array used to be a Map before I realized its keys are sequential and thus it is more efficient as an array. I'll rename.

A "face" is a part of a graph or a solid (brep).  A "facet" is part of a mesh (`Polyface`).

core/geometry/src/topology/Voronoi.ts
dassaf4: Ha! Indeed it would be infinite, if it were ever called! I'll remove.

dassaf4 (approved):
I think all that remains is an entry in NextVersion.md providing an example of how to use the new `ClipUtilities` method.

PR #8290, opened 2025-06-30
Allow Entity.getMetaData() to Support Relationship Classes

MichaelSwigerAtBentley:
Updates Entity.getMetaData() to return Relationship Classes.
Updates Entity._metadata cache to support Relationship Classes.
Adds getMetaData() overrides for Element and Relationships
Adds a test

Follow-up to discussion from PR https://github.com/iTwin/itwinjs-core/pull/8277


MichaelSwigerAtBentley:
@rschili 

Do we need to provide overrides for Element.getMetaData() and Relationship.getMetaData()? I assume it is more efficient to call getSchemaItem with EntityClass/RelationshipClass vs ECClass hence the need to provide overrides?

MichaelSwigerAtBentley:
@aruniverse @ben-polinsky Could I get a review? Other PR is merged and conflicts corrected

aruniverse (approved):

PR #8293, opened 2025-07-03
Draping Bug

danielzhong-bentley:
https://github.com/orgs/iTwin/projects/37/views/26?sliceBy%5Bvalue%5D=danielzhong-bentley&pane=issue&itemId=117082771&issue=iTwin%7Citwin-graphics-backlog%7C698

This PR fixes an issue where the planar texture projection matrix becomes unstable or incorrect (example: for some model, camera angle being nearly parallel to the drape), resulting in rendering artifacts. 

Solution:
Created `isProjetionMatrixUnstable()` to detect whether the computed projection matrix deviates from a valid and expected format. If the matrix is found to be unstable, the system reuses the last known stable projection matrix to drape.

- [x] Image Test


pmconne:
Won't reusing a previous projection matrix cause the results to be inaccurate? In the user's recording, the camera angle is not very close to parallel to the drape when the "smearing" occurs. What specifically is unstable or correct about the matrix, and is fixing it an option?

danielzhong-bentley:
@pmconne After more investigation, I believe the changes in this PR are more of a hack than a proper solution and likely not helpful. However, please check the root cause of the issue in the backlog link below, any feedback would be greatly appreciated. Thanks!

https://github.com/iTwin/itwin-graphics-backlog/issues/698

PR #8294, opened 2025-07-03
Collect/log statistics about changeset that can be used for health check

RohitPtnkr1996:
imodel-native: https://github.com/iTwin/imodel-native/pull/1164

This PR contains code and tests for the changeset health statistics being tracked when applying a changeset through the node addon.

### DIFF
--- a/common/changes/@itwin/core-backend/nam-cve-GHSA-fjxv-7rqg-78g4_2025-07-21-21-58.json
+++ b/common/changes/@itwin/core-backend/nam-cve-GHSA-fjxv-7rqg-78g4_2025-07-21-21-58.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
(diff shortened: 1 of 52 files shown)


ben-polinsky (approved):

core/common/src/ChangesetProps.ts
```diff
@@ -100,3 +102,31 @@ export interface ChangesetRange {
   /** index of last changeset. If undefined, all changesets after first are returned. */
   end?: ChangesetIndex;
 }
+
+/**
+ * Statistics for a single SQL statement executed during changeset application.
+ * @public
```
aruniverse: do you really want these public right as we add them?

theres `@public @preview` and `@beta`

aruniverse (approved):

core/common/src/ChangesetProps.ts
RohitPtnkr1996: Done !
Updated to beta

PR #8297, opened 2025-07-03
Reduce extraneous edges in output of `RegionOps.regionBooleanXY`

dassaf4:
Many methods of `RegionOps` employ algorithms that work by adding "bridge" edges to connect nested loops, but when these edges persist in output, they can result in unexpected/undesirable edges in downstream objects.

For example, a Boolean subtraction of nested circles with `RegionOps.regionBooleanXY` previously produced a "split washer" type `UnionRegion` consisting of a single `Loop` with a bridge edge traversed twice in opposite directions. When this region was swept into a `LinearSweep` solid and then meshed, the bridge edge caused unexpected boundary edges in the solid, and extraneous edges in sweep faces. (This fix is verified in the "LinearSweepWithHoles" test in Solid.test.ts.)

This PR represents a refactor of RegionOps algorithms involving bridge edges.
* `RegionOps.regionBooleanXY` now produces the simplest `AnyRegion` possible, without bridge edges. The split washer of the above example is now returned as a `ParityRegion` without bridge edge.
* `RegionOps.constructAllXYRegionLoops` now adds bridge edges to improve its output. Bridge edges allow loop nesting to be discovered in the graph, and this reduces the number of returned negative area "exterior" loops, the primary outputs of this method. This new behavior can be disabled by passing `addBridges = false`.
* `HalfEdgeGraph`: graph construction algorithms now mark bridge and regularization edges with new `HalfEdgeMask`s for downstream processing.

Also:
* `CurveChain.isPhysicallyClosedCurve`, `CurvePrimitive.isPhysicallyClosedCurve`: (new) fast checks for endpoint equality.
* `RegionOps.consolidateAdjacentPrimitives`: `ConsolidateAdjacentCurvePrimitivesOptions.consolidateLoopSeam` is a new input option, controlling the merging of first and last primitives in a Loop.
* `LineString3d`: reduce overuse of this class's static "work" objects, which can cause hard-to-debug errors due to reentrant lambda invocations.
* `PolyfaceBuilder`: can now mesh swept solids (`RuledSweep`, `LinearSweep`, `RotationalSweep`) created by sweeping a `UnionRegion`.
* `RegionMomentsXY`: correctly account for islands in ParityRegion area computations.
* `CurveCurveCloseApproachXY`: allow close approach to a zero-length line segment.
* `SortablePolygon`: try 4x harder to find interior points when sorting nested Loops.
* `HalfEdge.isSplitWasherFace`, `HalfEdge.announceEdgesInSuperFace`, `HalfEdge.healEdge`: new methods for detection and traversal of faces with bridges, and for cleaning up after removing bridge edges.


pmconne (approved):

core/geometry/src/curve/Query/PlanarSubdivision.ts
```diff
@@ -195,30 +201,100 @@ export class PlanarSubdivision {
...
+    if (info) {
+      if (info.reversed)
+        return info.detail.curve!.clonePartialCurve(info.detail.fraction1!, info.detail.fraction);
+      return info.detail.curve!.clonePartialCurve(info.detail.fraction, info.detail.fraction1!);
+    }
+    return undefined;
+  }
+  /**
+   * Create a [[Loop]] for the given face.
+   * @param face a node in the face loop, or an array of HalfEdges that comprise a face (e.g., super face).
```
saeeedtorabi: what happens if one passes array of HalfEdge nodes that belong to different faces? One node from a face is sufficient to build the loop in the face, so why allow to pass more than one node?

core/geometry/src/test/curve/RegionBoolean.test.ts
```diff
@@ -826,6 +828,87 @@ describe("RegionBoolean", () => {
...
+      let mergedChildrenAreLoops = false;
+      GeometryCoreTestIO.captureCloneGeometry(allGeometry, merged, x0 += delta);
+      if (ck.testType(merged, UnionRegion, "regionBooleanXY union produced a UnionRegion")) {
+        mergedChildrenAreLoops = ck.testArrayType(merged.children, Loop, "regionBooleanXY produced a UnionRegion with all Loop children");
+        ck.testExactNumber(inputs.length, merged.children.length, "regionBooleanXY added no bridge edges that survive");
+      }
+
+      // union of loops split into groups does not change result
+      if (merged) {
+        const merged2 = RegionOps.regionBooleanXY([inputs[0], inputs[1]], [inputs[2], inputs[3]], RegionBinaryOpType.Union);
```
saeeedtorabi: why not `merged1`? I dont see `merged1`

core/geometry/src/test/curve/RegionBoolean.test.ts
```diff
@@ -826,6 +828,87 @@ describe("RegionBoolean", () => {
...
+          ck.testArrayType(mergedThenSorted.children, Loop, "sortOuterAndHoleLoopsXY after regionBooleanXY produced a UnionRegion with all Loop children");
+        }
+      }
+
+      // finding all loops: "hole" loops are indistinguishable from other positive area loops, negative area loop is the exterior loop
+      const signedLoops = RegionOps.constructAllXYRegionLoops(inputs);
+      let exteriorLoop;
+      if (ck.testExactNumber(1, signedLoops.length, "constructAllXYRegionLoops found one connected component")) {
+        GeometryCoreTestIO.captureCloneGeometry(allGeometry, signedLoops[0].positiveAreaLoops, x0 += delta);
+        GeometryCoreTestIO.captureCloneGeometry(allGeometry, signedLoops[0].negativeAreaLoops, x0, 0, 10);
```
saeeedtorabi, line 879: I would do `dy=10` not `dz=10`. I was looking at the output and then realized there 2 shapes on top of each other in the top view. If you use `dy=10` then it's clear and no need to rotate the top view.

core/geometry/src/test/solid/Solid.test.ts
```diff
@@ -522,6 +525,54 @@ describe("Solids", () => {
     GeometryCoreTestIO.saveGeometry(allGeometry, "Solid", "DgnSolids");
     expect(ck.getNumErrors()).toBe(0);
   });
+
+  it("LinearSweepWithHoles", () => {
+    const ck = new Checker();
+    const allGeometry: GeometryQuery[] = [];
+    let x0 = 0;
+    let y0 = 0;
+    const outer = Loop.create(LineString3d.create(Sample.createRectangleXY(0, 0, 4, 3, 0)));
```
saeeedtorabi: optional: remove last 0 because z is optional
same for next 2 lines

core/geometry/src/topology/Graph.ts
```diff
@@ -334,6 +326,37 @@ export class HalfEdge implements HalfEdgeUserData {
...
+   * @param doomed one of two nodes added by [[splitEdge]]. These nodes should form a vertex loop of two nodes.
+   * On successful return this node and its mate are isolated.
+   * @param checkParallel whether to check that the doomed edge and the preceding edge in its face loop are parallel.
+   * When passing `true` the assumption is that edge geometry is linear. If nonlinear edge geometry is attached, the
+   * caller should a) verify that the geometry on either side of the doomed vertex can be merged, and if so, they
+   * should b) call this method passing `false`, and c) adjust the geometry of the returned edge and its edge mate
+   * as appropriate.
+   * @returns the former (surviving) face predecessor of `doomed`, or undefined if the edge can't be healed.
+   */
+  public static healEdge(doomed: HalfEdge, checkParallel: boolean = true): HalfEdge | undefined {
```
saeeedtorabi, line 340: This API is reverse of `splitEdge`.
Opposite of `split` is `merge` or `join`.
I think those 2 words are better than `heal`

core/geometry/src/topology/Graph.ts
```diff
@@ -991,6 +1022,71 @@ export class HalfEdge implements HalfEdgeUserData {
...
   }
+  /**
+   * Search around the instance's face loop for nodes with the specified mask value.
+   * * Returned nodes satisfy `node.isMaskSet(mask) === value`.
+   * @param mask target mask.
+   * @param value target boolean value for mask on half edges (default `true`).
+   * @param result optional array to be cleared, populated with masked nodes, and returned.
+   * @return array of masked half edges
+   */
+  public collectMaskedEdgesAroundFace(mask: HalfEdgeMask, value: boolean = true, result?: HalfEdge[]): HalfEdge[] {
```
saeeedtorabi, line 1033: This goes around a face and finds nodes with a specific mask value and returns them has a half-edge array. Then users may want to do something with that array. What if we also pass an optional `NodeFunction` similar to `HalfEdge.collectAroundFace` so users can do what they what while calling `collectMaskedEdgesAroundFace`?

core/geometry/src/curve/Query/PlanarSubdivision.ts
dassaf4: It is up to the caller to ensure the nodes comprise a loop. I'll make this clearer. In my use case, I collect an array of `HalfEdge`s comprising a superface to send in. They are in different graph faces, but still connected head-to-tail so form a loop.

core/geometry/src/test/curve/RegionBoolean.test.ts
dassaf4: I guess merged1 didn't survive.

core/geometry/src/test/curve/RegionBoolean.test.ts
dassaf4: I prefer to look at them right on top of each other to ensure edges match. One can always rotate the view to distinguish them. But view rotation does not allow such vertical comparison when they are separated in y.

core/geometry/src/test/solid/Solid.test.ts
dassaf4: Sometimes I prefer the optional arg for emphasis. But here it is truly unimportant.

core/geometry/src/topology/Graph.ts
dassaf4: Merge and join are used elsewhere, and only "heal" connotes a reversal.

core/geometry/src/topology/Graph.ts
dassaf4: Users can do what they want with the array after the method returns. The value in providing a callback to `collectAroundFace` is to allow callers to accumulate something else other than `HalfEdge`s. Here I wanted to give the caller the option to use a pre-allocated HalfEdge[].  And that argument doesn't coexist easily with a callback that can return `any`.

core/geometry/src/topology/Graph.ts
saeeedtorabi: how about `combine`?

dassaf4: "Combine" doesn't connote a reversal either. But I forgot the most important reason: "Heal" is also used in native code for the opposite of split.

common/api/core-geometry.api.md
```diff
@@ -1376,7 +1376,8 @@ export class ConsolidateAdjacentCurvePrimitivesOptions {
     colinearPointTolerance: number;
     consolidateCompatibleArcs: boolean;
     consolidateLinearGeometry: boolean;
-    disableLinearCompression?: boolean;
+    consolidateLoopSeam: boolean;
+    disableLinearCompression: boolean;
```
aruniverse, line 1380: isnt this breaking?

dassaf4: `disableLinearCompression` was added [in May](https://github.com/iTwin/itwinjs-core/pull/8095), so I was thinking not?

aruniverse: i think we need to keep them both (prop you modified and newly added one) optional; otherwise existing code will break because these additional options are now required

dassaf4: Agree. I'll make both optional in my [current PR](https://github.com/iTwin/itwinjs-core/pull/8316). Thanks for pointing this out!

But ugh, the default values don't really work here, do they? What's the best way to specify a bundle of options that have default values?

* An interface does not allow the specification of default values.

* Using a class means you can only get a default value by invoking a constructor. Just passing in json ignores the defaults for unspecified data: [playground](https://www.typescriptlang.org/play/?#code/MYGwhgzhAECC0G8BQ1XQA4FcBGICWw0AZgAwBc0EALgE54B2A5tALzQBEAJgKZFiYgq7ANwo0WXAWIBGAPwVqdJqw48+AoaLQYc+QsAD29RZmBUDNABRh5lWg0YAaaNluKHASkRjtqPEWhraABCFjZMejUGbk4PH19tKgALPAgAOlIVMC0E6H9A7BCw6Aio+hi43N9k1IzpFWwc1ABfJFakIgizPCMMJSpYawBJegpYL2RtQ2MDEG40kANGSwADABIEMBGMkmbnDa36OuaVj1F26YhZ+cXl9gBaR6en9jikAHp3jEgIB2gAKyu9DyjHoFm4MDU-EEMCIFhKxnQ3GA-jwMWgBnQVB69DAIGgnDAVDASHQ-UGCFIFHYcIM7GaZyqVU+HFpzlKvGinHYHy+ZIYA0slOk1Np9MZTLQLO4NBoFgomTBVAxWJxeOc3AAbtxgQB3PDJPJUSG8aHKzV4zDcJC86DJbgY+ggACe0F1YFd5mgjG4yrABNNGmgFpAVryMC9DE1BgA1g7-WYLKTyZZyrq4BLJZKWVxA4JnFCNDz+fRBWm4JYaQY6R5M1noDm2QH1IJiyny4MOUQuc52NgwDRXnWc4X8xx+4PkwLBh3K2LexOh8JmV8qwZnIukEA).

aruniverse: I think what ive typically seen is use defaults in ctor or in the fn signature. So I think youre right with the approach of invoking a ctor. Heres [some other examples](https://www.typescriptlang.org/play/?#code/MYGwhgzhAECCCM0DeAoa7oAcCuAjEAlsNAGYAMAXNBAC4BOBAdgObQC80ARACYCmJYbCBqcA3Ggw58RUvAD8VWgxbsufAUJHiMWPIWLAA9oyXZgNQ3QAUYBdXpNmAGmi47SxwEpkEnegIk0DbQAIRsHNiM6ky83J6+fjo0ABYEEAB05Kpg2onQAUG4oeHQkdGMsfF5filpmYgcuLnoAL4obSigkDCwAEw+OlL6pJT2yszNutLEJPKKDiyTQzJGJvRmFta28+MubjteA3m1GVkcYNAAPpdq-ILCYglJqacNrlc3PHeanAltHV0oHAAMxHdDLGajDyLBIQ2TuBYTWF6FbGUzmSw2A4qDhfDQPPbY1i49T3ETeVDVE6ZMjZSbPOqzVRNP7tFAoEiRcwEYxYZQ0WA2ACSjCoCCucH6N1gwIpCVWEEMIF46RAhmYVgABgASJBgEU0louXX6xj1FqazziAFopUqtUazgAWhdrtdnHiKAA9F6sN1HNAAFaKxj5ZiMSy8GCkzQwEiWUomTC8YABAixaCGTA0HmMMAgaDcMA0MAoTD8wVIchUTjxwycFpW6rVH1cOsuMr8GLcX6t8tMAVWKvwGt1htN5sYVu8Oh0SxULIRmiZ7O5-MuXgAN14oYA7gQUvkaNHvsJoJv89heOzWyleJnGCAAJ7QXdgF8WaDMXjLi4xs8XiAV75DAn5MJuhgANb3hcGJ0GWFZWBUu5wPAE6TpOrZ4mSLj-iICEDoKyGoVYtaGPWnjoRh0BYe2hanvh-aMIOxEIFYnYkN2LicLgYB0B6VFYXh3G8fxBHMURvAoWxZH1iJfECaILa+rJewKeyKBAA) of providing defaults in the ctor.

And i think this is fine because the props youve added were for a class anyway

PR #8300, opened 2025-07-04
[ Fix ] : formatTraitsToArray to return camel case

pankhur94:
- Change method formatTraitsToArray to return Camel case instead of Pascal


rschili (approved):

hl662:
@christophermlawson or @rschili can you check the C++ side and how they handle casing?

aruniverse:
Is this also not breaking??

hl662:
Closing for now, will investigate the native side later

PR #8310, opened 2025-07-10
Tile geometry collector reproject option

eringram:


core/frontend/src/tile/RealityTile.ts
```diff
@@ -633,8 +636,16 @@ export class RealityTile extends Tile {
       case "accept":
         if (!this.isReady)
           collector.addMissingTile(this.loadableTile);
-        else if (this.geometry?.polyfaces)
+        else if (this.geometry?.polyfaces) {
+          const xForm = this._reprojectionTransform;
+          if (this.tree.reprojectGeometry && xForm) {
+            // If the tile is reprojected, transform the polyfaces to the reprojection transform.
+            for (const polyface of this.geometry.polyfaces) {
+              polyface.tryTransformInPlace(xForm);
```
markschlosseratbentley: Consider changing the transform as discussed elsewhere so transform applications do not accumulate.  Instead of the transform-in-place loop, maybe something like this: `polyfaces = polyfaces.map((pf) => pf.cloneTransformed(xForm));`

core/frontend/src/tile/RealityTile.ts
```diff
@@ -633,8 +636,16 @@ export class RealityTile extends Tile {
...
+            }
```
markschlosseratbentley: Consider adding / expanding a unit test that verifies the polyface transform does not accumulate when the collect function is called multiple times.

core/frontend/src/tile/RealityTile.ts
eringram: This has been added

core/frontend/src/tile/RealityTile.ts
eringram: Added a test that calls `tree.collectTileGeometry(collector)` twice in a row and expects the results to be identical to if it's only called once.

core/frontend/src/tile/RealityTile.ts
```diff
@@ -633,8 +636,15 @@ export class RealityTile extends Tile {
...
         if (!this.isReady)
           collector.addMissingTile(this.loadableTile);
-        else if (this.geometry?.polyfaces)
-          collector.polyfaces.push(...this.geometry.polyfaces);
+        else if (this.geometry?.polyfaces) {
+          let polyfaces = this.geometry.polyfaces;
+          const xForm = this._reprojectionTransform;
+          if (this.tree.reprojectGeometry && xForm) {
+            // If the tile is reprojected, transform the polyfaces to the reprojection transform.
+            polyfaces = polyfaces.map((pf) => pf.cloneTransformed(xForm));
```
pmconne: More efficient alternative that avoids potentially repeatedly cloning and transforming the same set of polyfaces would be to change `RealityTileTreeId.produceGeometry` from a boolean to `"no" | "yes" | "reproject"`. Then you would simply reproject the geometry once, when loading it, IFF the tree reference's Id says so.

Don't hold up the PR for this though.

core/frontend/src/tile/TileTreeReference.ts
```diff
@@ -288,6 +293,14 @@ export abstract class TileTreeReference /* implements RenderMemory.Consumer */ {
    */
   protected _collectTileGeometry(collector: TileGeometryCollector): void {
     const tree = this.treeOwner.load();
+
+    // The reprojectGeometry property is only relevant for RealityTileTree because only reality tiles have the reprojectionTransform.
+    if (tree instanceof RealityTileTree) {
+      // Make sure tree knows to reproject when collectTileGeometry is called.
+      // this.reprojectGeometry will only be defined if this is a GeometryTileTreeReference.
+      tree.reprojectGeometry = this.reprojectGeometry;
```
pmconne: Eww. There's no alternative to (1) up-casting and (2) modifying the `TileTree`'s state here? I think my suggestion of putting the "reproject" option into the tile tree Id would solve both of these (`tree.reprojectGeometry` would be set in the tree's constructor).

core/frontend/src/tile/TileTreeReference.ts
```diff
@@ -332,3 +347,14 @@ export abstract class TileTreeReference /* implements RenderMemory.Consumer */ {
     return tileTreeReferenceFromRenderGraphic(args);
   }
 }
+
+/** Options for creating a [[GeometryTileTreeReference]].
+ * @public
+ */
+export interface GeometryTileTreeReferenceOptions {
+  /** If set to true, tile geometry will be reprojected using the tile's reprojection transform when geometry is collected from the referenced TileTree.
+   * Currently only applies to reality tiles.
```
pmconne:
```suggestion
   * Currently only applies to point clouds, reality meshes, and terrain.
```

core/frontend/src/tile/TileTreeReference.ts
eringram: Good point, that's a much better way to get the reproject option into the tree. I changed `RealityTreeId.produceGeometry` to a `type ProduceGeometryOption = "yes" | "no" | "reproject"` and was able to remove the cast here.

There is also still a boolean `MapTileTree.produceGeometry` property, however it doesn't seem to interact with this option.

core/frontend/src/tile/RealityTile.ts
eringram: In the latest commit I tried reprojecting in `RealityTileLoader.loadGeometryFromStream`. This isn't covered by the unit tests in this PR yet, but works but when I test with the TerrainDrapeTool. Is that where you meant by when the geometry is loaded? Also this may be not any better because I'm still using `cloneTransformed`.

core/frontend/src/test/tile/RealityTile.test.ts
```diff
@@ -0,0 +1,243 @@
...
+        maximumSize: 42,
+      }, tileTree);
+
+      this._contentSize = contentSize;
+
+      if (contentSize === 0)
+        this.setIsReady();
+
+      // Create indexed polyface for testing by facetting a sphere
+      const sphere = Sphere.createCenterRadius(Point3d.createZero(), 1);
```
pmconne: consider just making a polyface from a single triangle with coordinates like `(0, 0, 0), (1, 0, 0), (1, 2, 0)` so that your test doesn't need to expect values like `9.346718517561811`. Then it will be clear in your test that the transform applied correctly.

eringram: Simplified the tests to just create a triangle and expect more legible values. My tests will still fail due to this change https://github.com/iTwin/itwinjs-core/pull/8310#discussion_r2211728263 - still figuring out how to test when the reproject occurs in `RealityTileLoader.loadGeometryFromStream`

eringram: Finished adding tests that call `RealityTileLoader.loadGeometryFromStream` and show the reprojection option working.

markschlosseratbentley (approved):
Thanks, @eringram, for your work on this!

Issue #8178, opened 2025-06-04
TextBlockComponent children should implicitly inherit style and styleOverrides from parent TextBlockComponents


Alfonso-Martello:
**Describe the bug**
TextBlockComponents (TextBlocks, Paragraphs, and Runs) all specify a style and can optionally specify styleOverrides. However, only the style on the RunLayout is used to create geometry. That style is a combination of the style/overrides on the Run and a subset of block level settings from the TextBlock. This means if you want to apply a style to a TextBlock or a Paragraph, you must apply the styling to each child Run.

This leads to confusion when a TextStyleSetting is set on a TextBlock, but it does not apply to the TextBlock’s Runs. It also leads to forced repetition. If you want to apply a style to an entire TextBlock, you must repeat that same style on each Run.

**To Reproduce**

1. Create an empty TextBlock.
2. Set a fontName in the style overrides of the TextBlock.
3. Create a TextRun without setting a fontName in the Run's styleOverrides. Include some text content like “Hello World”.
4. Append the TextRun to the TextBlock.
5. Create a TextAnnotation using the TextBlock.
6. Produce geometry for the TextAnnotation.
7. Observe that the text content “Hello World” does not render, because the font in the TextBlock’s override never made it into the RunLayout’s style.

**Expected behavior**
Styles set on a TextBlockComponent should apply to its children, unless the child provides its own style/overrides. So TextBlock styles should apply to all child Paragraphs, and Paragraph styles should apply to all child Runs. Ideally, this should happen implicitly, so that you do not need to repeat the style nor overrides in the child classes.

pmconne:
> **Expected behavior** Styles set on a TextBlockComponent should apply to its children, unless the child provides its own style/overrides.

Do you agree with the following more comprehensive definition?
- If a child specifies a style Id, then none of the parent's styling (including overrides) apply to the child, even if the parent and child specify the **same** style Id. It effectively resets all styling (similar to how specifying a subcategory Id in a geometry stream resets all appearance modifiers).
- Otherwise:
  - The child uses the same style Id as the parent.
  - Each style property overridden or inherited by the parent is also applied to the child, unless the child explicitly provides its own override for that property.

Are you asking someone to implement this or planning to implement it yourself?

Alfonso-Martello:
Yes, I agree with the more comprehensive definition. Thank you for putting that together!

[...]

PR #8313, opened 2025-07-10
Text style and annotation changes

Alfonso-Martello:
This PR addresses a few issues related to Text and specifically Text Styles. The changes are as follows:
1.	TextAnnotation elements use new BIS property `textAnnotationData` introduced [here](https://github.com/iTwin/bis-schemas/pull/440).
2.	An `AnnotationTextStyle` class is introduced to TypeScript layer. `TextBlock`s now reference this class by ID.
3.	`TextBlockComponent`s implicitly inherit styling from parents similar as to what was detailed in #8178. However, Runs and Paragraphs no longer have `styleId` as a property.
4.	`TextAnnotation`s now take into account drawing scale.
5.	Removes the `TextStyle` class as `TextStyleSettings` are now stored in the iModel via `AnnotationTextStyles`.

# Note 1: TextBlocks as the source of truth on styling
You will note that TextBlocks are still the authority on the style, and TextAnnotations just use the style of their TextBlock. Likewise, Leader’s will use the style of the TextBlock (if there is one).
When we were thinking about the structure for styling, we ended up with this diagram:
 
 
image

The reasoning behind this is as follows:
1.	This PR changes the way TextBlockComponents inherit styles. There is now a more complex system for determining the final “effective” style of the text. TextBlocks already deal with this in the `layoutTextBlock` method. If we added styles to the TextAnnotation, then we would need to take into account the annotation when laying out the text block (such as through passing in information about the TextAnnotation or having a whole separate `layoutTextAnnotation` method). 
2.	In Drawing Production, we have Dimensions and Tables that work with and persist TextBlocks. A TextBlock is styled text. A TextAnnotation is styled text in space. These classes already know where the text is spatially located, so they do not need to persist this information in a TextAnnotation. The only time the classes use TextAnnotations is when they are dynamically creating them to produce graphics. We want to be able to support these classes to just be able to plugin a TextBlock.
3.	The TextEditor feature in Drawing Production works on TextBlocks, so we can use the TextEditor on Text, Dimensions, and Tables (anything that uses a TextBlock).
5.	TextAnnotations always have exactly 1 TextBlock. It is simple enough to just use the TextBlock’s style as the TextAnnotation’s style. However, there is never really a need to change the TextAnnotation’s style, when you can just call `textAnnotation.textBlock.styleId = …` as opposed to `textAnnotation.styleId = ...`.

# Note 2
There were some changes/areas of the code that felt unsure about while working on this PR. I have left comments on those areas detailing my thoughts.

# Note 3: What happened to `styleId` on Paragraph and Run?
From the perspective of Drawing Production, a text style represents the purpose of the text. In this way, it does not make sense to create a text block for purpose A with style A, but then have a paragraph or run inside of the that text block have a different purpose B with style B. Thus, Drawing Production is not planning on using/setting the `styleId` on Paragraphs and Runs.

Additionally, we theorize that any style differences within a text block can be sufficiently represented as styleOverrides on the Paragraphs and Runs (for example, bolding a snippet of text).

Finally, removing the `styleId` from Paragraphs and Runs is a breaking change. However, later adding back in the `styleId` to Paragraphs and Runs is not a breaking change (text that only has styleOverrides works the same way as text with an empty styleId and styleOverrides). We thought removing it now would simplify the data but still leave us the future option of reintroducing the `styleId` if our assumptions are proved wrong.


core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -30,70 +42,384 @@ function getElementGeometryBuilderParams(iModel: IModelDb, _placementProps: Plac
...
+
+  /**
+   * Updates the geometry of the TextAnnotation2d on insert.
+   * @inheritdoc
+   * @beta
    */
-  public getAnnotation(): TextAnnotation | undefined {
-    const json = this.jsonProperties.annotation;
-    return json ? TextAnnotation.fromJSON(json) : undefined;
+  protected static override onInsert(arg: OnElementPropsArg): void {
```
Alfonso-Martello, line 114: I included `onInsert` and `onUpdate` in `TextAnnotation2d` and `TextAnnotation3d` because these classes manage updating their geometry. Technically a user can get around updating the geometry by creating their props directly (without calling `toJSON`) and then calling `IModelDb.elements.insertElement(directly-created-props)` or `IModelDb.elements.updateElement(directly-created-props)`. This catches that case.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -30,70 +42,384 @@ function getElementGeometryBuilderParams(iModel: IModelDb, _placementProps: Plac
...
+  protected static override readonly _customHandledProps: CustomHandledProperty[] = [
+    { propertyName: "textAnnotationData", source: "Class" },
+  ];
+
+  /**
+   * TextAnnotation2d deserializes 'textAnnotationData'.
+   * @inheritdoc
+   * @beta
+   */
+  public static override deserialize(props: DeserializeEntityArgs): TextAnnotation2dProps {
```
Alfonso-Martello: I saw `serialize` and `deserialize` were added recently. I went ahead and included them in these classes, but I am not sure when they need to be defined or if I should ever be calling them directly

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -30,70 +42,384 @@ function getElementGeometryBuilderParams(iModel: IModelDb, _placementProps: Plac
...
+  }
+
+  /**
+   * Checks that the AnnotationTextStyle is not in use before deleting it.
+   * @note The in use check is done here instead of in `deleteDefinitionElements`.
+   * @throws an error if it is referenced by any [[TextAnnotation2d]] or [[TextAnnotation3d]] elements.
+   * @inheritdoc
+   * @beta
+   */
+  protected static override onDelete(arg: OnElementIdArg): void {
```
Alfonso-Martello: Since `AnnotationTextStyle` is a `DefinitionElement`, it must be deleted via `deleteDefinitionElements`. `deleteDefinitionElements` calls into the native code, and the native code does not know how to look up the usage of `AnnotationTextStyle`s. Thus, I opted to check in the class and throw an error if the element is in use. 

However, this means that calling `deleteDefinitionElements` on an `AnnotationTextStyle` causes an error to be thrown. Additionally, `deleteDefinitionElements` returns a set of in-use IDs. Even if `deleteDefinitionElements` did not throw an error on `AnnotationTextStyle`s, it still would not include their IDs in the returned set.

An alternative implementation I could see would be to do this logic inside of `deleteDefinitionElements`, or perhaps expose the results of the usage query as a callable function

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -149,50 +164,112 @@ function applyBlockSettings(target: TextStyleSettings, source: TextStyleSettings
...
+ * @beta
+ */
+export interface TextStyleResolverArgs {
+  /** The text block whose styles are being resolved. */
+  textBlock: TextBlock;
+  /** The iModel from which to obtain [[AnnotationTextStyle]]s when resolving styles. */
+  iModel: IModelDb;
+  /** The ID of the model containing the text block, used to compute the scale factor. */
+  modelId?: Id64String;
+  /** @internal chiefly for tests, by default looks an [[AnnotationTextStyle]] in the iModel by ID. */
```
Alfonso-Martello:
```suggestion
  /** @internal chiefly for tests, by default looks up an [[AnnotationTextStyle]] in the iModel by ID. */
```

test-apps/display-test-app/src/backend/TextImpl.ts
```diff
@@ -0,0 +1,137 @@
+import { AnnotationTextStyle, BriefcaseDb, Drawing, IModelDb, TextAnnotation2d } from "@itwin/core-backend";
+import { Id64String } from "@itwin/core-bentley";
+import { Placement2d, Placement2dProps, TextAnnotation, TextAnnotationProps, TextStyleSettings, TextStyleSettingsProps } from "@itwin/core-common";
+
+export async function insertTextStyle(iModelKey: string, name: string, settingProps: TextStyleSettingsProps): Promise<Id64String> {
```
claudiaareneee, line 13: Could add comments (I just used gh copilot to generate these):


```suggestion
/**
 * Inserts a new text style into the iModel.
 * @param iModelKey - Key to identify the iModel.
 * @param name - Name of the text style.
 * @param settingProps - Properties for the text style.
 * @returns The Id of the inserted text style.
 * @throws If insertion fails, abandons changes and rethrows the error.
 */
export async function insertTextStyle(iModelKey: string, name: string, settingProps: TextStyleSettingsProps): Promise<Id64String> {
```

test-apps/display-test-app/src/backend/TextImpl.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    iModel.saveChanges(`Inserted text style '${name}'`);
+    return textStyleId;
+  } catch (e) {
+    iModel.abandonChanges();
+    throw e;
+  }
+}
+
+export async function updateTextStyle(iModelKey: string, name: string, newSettingProps: TextStyleSettingsProps): Promise<void> {
```
claudiaareneee, line 42:
```suggestion
/**
 * Updates an existing text style in the iModel.
 * @param iModelKey - Key to identify the iModel.
 * @param name - Name of the text style to update.
 * @param newSettingProps - New properties for the text style.
 * @throws If update fails, abandons changes and rethrows the error.
 */
export async function updateTextStyle(iModelKey: string, name: string, newSettingProps: TextStyleSettingsProps): Promise<void> {
```

test-apps/display-test-app/src/backend/TextImpl.ts
```diff
@@ -0,0 +1,137 @@
...
+    textStyle.update();
+
+    iModel.saveChanges(`Updated text style '${name}'`);
+  } catch (e) {
+    iModel.abandonChanges();
+    throw e;
+  }
+}
+
+export async function deleteTextStyle(iModelKey: string, name: string): Promise<void> {
```
claudiaareneee, line 66:
```suggestion
/**
 * Deletes a text style from the iModel.
 * @param iModelKey - Key to identify the iModel.
 * @param name - Name of the text style to delete.
 * @throws If deletion fails, abandons changes and rethrows the error.
 */
export async function deleteTextStyle(iModelKey: string, name: string): Promise<void> {
```

test-apps/display-test-app/src/backend/TextImpl.ts
```diff
@@ -0,0 +1,137 @@
...
+    iModel.elements.deleteDefinitionElements([textStyle.id]);
+
+    iModel.saveChanges(`Deleted text style '${name}'`);
+  } catch (e) {
+    iModel.abandonChanges();
+    throw e;
+  }
+}
+
+export async function insertText(iModelKey: string, categoryId: Id64String, modelId: Id64String, placement: Placement2dProps, textAnnotationData?: TextAnnotationProps): Promise<Id64String> {
```
claudiaareneee, line 92:
```suggestion
/**
 * Inserts a new text annotation element into the iModel.
 * @param iModelKey - Key to identify the iModel.
 * @param categoryId - Category Id for the annotation.
 * @param modelId - Model Id for the annotation.
 * @param placement - Placement properties for the annotation.
 * @param textAnnotationData - Optional text annotation properties.
 * @returns The Id of the inserted annotation.
 * @throws If insertion fails, abandons changes and rethrows the error.
 */
export async function insertText(iModelKey: string, categoryId: Id64String, modelId: Id64String, placement: Placement2dProps, textAnnotationData?: TextAnnotationProps): Promise<Id64String> {
```

test-apps/display-test-app/src/backend/TextImpl.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    iModel.saveChanges('Inserted annotation');
+    return annotationId;
+  } catch (e) {
+    iModel.abandonChanges();
+    throw e;
+  }
+}
+
+export async function updateText(iModelKey: string, elementId: Id64String, categoryId?: Id64String, placement?: Placement2dProps, textAnnotationProps?: TextAnnotationProps): Promise<void> {
```
claudiaareneee, line 124:
```suggestion
/**
 * Updates an existing text annotation element in the iModel.
 * @param iModelKey - Key to identify the iModel.
 * @param elementId - Id of the annotation element to update.
 * @param categoryId - Optional new category Id.
 * @param placement - Optional new placement properties.
 * @param textAnnotationProps - Optional new text annotation properties.
 * @throws If update fails, abandons changes and rethrows the error.
 */
export async function updateText(iModelKey: string, elementId: Id64String, categoryId?: Id64String, placement?: Placement2dProps, textAnnotationProps?: TextAnnotationProps): Promise<void> {
```

test-apps/display-test-app/src/backend/TextImpl.ts
```diff
@@ -0,0 +1,137 @@
...
+    await iModel.locks.acquireLocks({ shared: [text.model], exclusive: [elementId] });
+    text.update();
+    iModel.saveChanges('Updated annotation');
+  } catch (e) {
+    iModel.abandonChanges();
+    throw e;
+  }
+}
+
+export async function deleteText(iModelKey: string, elementId: Id64String): Promise<void> {
```
claudiaareneee, line 154:
```suggestion
/**
 * Deletes a text annotation element from the iModel.
 * @param iModelKey - Key to identify the iModel.
 * @param elementId - Id of the annotation element to delete.
 * @throws If deletion fails, abandons changes and rethrows the error.
 */
export async function deleteText(iModelKey: string, elementId: Id64String): Promise<void> {
```

test-apps/display-test-app/src/backend/TextImpl.ts
```diff
@@ -0,0 +1,137 @@
...
+    text.delete();
+
+    iModel.saveChanges(`Deleted text annotation`);
+  } catch (e) {
+    iModel.abandonChanges();
+    throw e;
+  }
+}
+
+export async function setScaleFactor(iModelKey: string, modelId: Id64String, scaleFactor: number): Promise<void> {
```
claudiaareneee, line 177:
```suggestion
/**
 * Sets the scale factor for a drawing element in the iModel.
 * @param iModelKey - Key to identify the iModel.
 * @param modelId - Id of the drawing model.
 * @param scaleFactor - New scale factor to set.
 * @throws If update fails, abandons changes and rethrows the error.
 */
export async function setScaleFactor(iModelKey: string, modelId: Id64String, scaleFactor: number): Promise<void> {
```

test-apps/display-test-app/src/backend/TextImpl.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+}
+
+export async function deleteTextStyle(iModelKey: string, name: string): Promise<void> {
+  const iModel = BriefcaseDb.findByKey(iModelKey);
+
+  try {
+    const textStyle = iModel.elements.getElement<AnnotationTextStyle>(AnnotationTextStyle.createCode(iModel, IModelDb.dictionaryId, name));
+
+    iModel.elements.deleteDefinitionElements([textStyle.id]);
```
claudiaareneee, line 73: Does deleting elements inherently pull out locks, or should exclusive locks be pulled out here?

claudiaareneee: (Same thing on line 111)

Alfonso-Martello:
Following the "BWG - Text Fields" meeting today (07/11/2025), it was mentioned that we should avoid storing element IDs in the JSON representation due to discoverability issues (element handler must be available to Transformation service to parse and change IDs).

I assume this should also apply for text styles?

For fields, an idea was thrown out about instead have the JSON point towards the relationship that can be followed to the source element.

This PR is introducing having the JSON refer to AnnotationTextStyles by ID. To avoid this, would we need to introduce a new ECRelationship like fields? I recall during a past BWG discussion (titled "Text Annotation topics" on 2025-02-07) that we discussed relationships like this existing in the past, but we removed them due to poor performance? Specifically, I recall `GeometryPart`s being mentioned

claudiaareneee:
My recent changes were an accident. I'm going to attempt to undo them 😭😭

Edit: 
It looks like there aren't any changes? Maybe it was just weird whitespace shenanigans

test-apps/display-test-app/src/frontend/TextDecoration.ts
```diff
@@ -6,40 +6,44 @@
 import { BaselineShift, ColorDef, FractionRun, LineBreakRun, Placement2dProps, TabRun, TextAnnotation, TextAnnotationAnchor, TextAnnotationFrameShape, TextAnnotationProps, TextBlock, TextBlockJustification, TextBlockMargins, TextFrameStyleProps, TextRun, TextStyleSettingsProps } from "@itwin/core-common";
 import { DecorateContext, Decorator, GraphicType, IModelApp, IModelConnection, readElementGraphics, RenderGraphicOwner, Tool } from "@itwin/core-frontend";
 import { DtaRpcInterface } from "../common/DtaRpcInterface";
-import { Id64, Id64String } from "@itwin/core-bentley";
+import { assert, Id64, Id64String } from "@itwin/core-bentley";
 import { Point3d, YawPitchRollAngles } from "@itwin/core-geometry";
+import { dtaIpc } from "./App";
 
 // Ignoring the spelling of the keyins. They're case insensitive, so we check against lowercase.
 // cspell:ignore superscript, subscript, widthfactor, fractionscale, fractiontype
```
claudiaareneee:
```suggestion
// cspell:ignore superscript, subscript, widthfactor, fractionscale, fractiontype, subscriptscale, superscriptscale, insertstyle, updatestyle, deletestyle, textblock
```

core/common/src/ElementProps.ts
```diff
@@ -169,11 +168,10 @@ export interface GeometricElement3dProps extends GeometricElementProps {
  * @extensions
  */
 export interface TextAnnotation3dProps extends GeometricElement3dProps {
```
claudiaareneee, line 170: Something about storing this as a string feels weird to me, but I'm not confident as to why. So not like a real comment here, but this just feels odd.

Alfonso-Martello: I agree, it does feel a little odd. The surface level reasoning is that the string is the stringified JSON representation as required by the [BIS schema](https://github.com/iTwin/bis-schemas/blob/72795f1cb28656431fd3200ca1a6cdcba26ab399/Domains/0-Core/Released/BisCore.01.00.17.ecschema.xml#L1568). 

My guess on why that is required is that it has something to do with the underlying [SQL-99 implementation used by ECSQL](https://www.itwinjs.org/learning/ecsql/#what-is-ecsql) not having support for native JSON data?

test-apps/display-test-app/src/frontend/TextDecoration.ts
Alfonso-Martello: Done

test-apps/display-test-app/src/backend/TextImpl.ts
Alfonso-Martello: I assume locks are required

Thanks for catching!

core/common/src/ElementProps.ts
pmconne: All properties of extended type "JSON" behave like this, except a couple special ones (e.g., `ElementProps.jsonProperties`) that have custom C++ logic to convert between string and JSON in the native layer. The `Props` types in ElementProps.ts represent the persistence format; people should prefer to use the corresponding `Element` APIs in core-backend where, e.g., `TextAnnotation2d.data` would be a proper object, not a string.

Alfonso-Martello:
Turns out, GitHub will treat `@beta` in a commit message as ating the user with the GitHub handle "beta". I just modified my most recent commit message to no longer accidently @ the GitHub user

Alfonso-Martello:
Re-requesting review from @pmconne now that this PR is out of draft

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: This isn't the right place to do an exhaustive search for usage. `queryDefinitionElementUsage` (or a portion thereof) needs to move out of native code so that subclasses written in JS can disclose their usage. I think this PR is already big enough, so I recommend filing that as a separate issue and deleting this code.

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: I don't love it but it's a consequence of permitting people to bypass the Element APIs to insert/update elements. I think we should stop permitting that but it's a separate conversation.

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: IIUC you only need to implement these methods for custom-handled properties; all other properties should be automatically de/serialized for you. @MichaelSwigerAtBentley confirm?

pmconne (commented):
Lots of undocumented `@preview` APIs. Please consider asking copilot to generate them for you, and then proof-read and adjust them.

core/backend/src/annotations/FrameGeometry.ts
```diff
@@ -13,21 +13,20 @@ import { Angle, AngleSweep, Arc3d, LineString3d, Loop, Path, Point3d, Range2d, T
...
  * @param frame
  * @param range to enclose with the frame
- * @param transform that transform the range to world coordinates
+ * @param transform that transforms the range to world coordinates
  * @returns `true` if any geometry was appended to the builder
 * @beta
  */
 export function appendFrameToBuilder(builder: ElementGeometry.Builder, frame: TextFrameStyleProps, range: Range2d, transform: Transform, geomParams: GeometryParams): boolean {
-  if (undefined === frame.shape || frame.shape === "none") {
+  if (frame.shape === "none" || frame.shape === undefined) {
```
pmconne, line 23: Please consider in a follow-up PR getting rid of options called "none", or requiring them to be defined; having two ways to say "no fill" or "no color" is unnecessarily complicated.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -30,33 +42,123 @@ function getElementGeometryBuilderParams(iModel: IModelDb, _placementProps: Plac
 export class TextAnnotation2d extends AnnotationElement2d {
   /** @internal */
   public static override get className(): string { return "TextAnnotation2d"; }
-  protected constructor(props: TextAnnotation2dProps, iModel: IModelDb) { super(props, iModel); }
+  public textAnnotationData?: string;
```
pmconne: Why provide a public mutable stringified JSON representation of the annotation? Nobody should directly set this - they should call `setAnnotation`. And nobody should be directly reading it - they should call `getAnnotation`. As it is, `getAnnotation` re-parses the JSON string every time it's called.

Why not instead have a `public annotation: TextAnnotation | undefined` that you initialize from the JSON string in the constructor, to replace `get/setAnnotation`?

core/backend/src/annotations/TextAnnotationElement.ts
MichaelSwigerAtBentley: Yes. That is correct. It is for custom properties that need custom logic to interpret what is stored in the database: For example: where the default value might be more complex than simply undefined, or in cases where the value might be named differently then what is stored in the database. In this case, the property is stored as `textAnnotationData` and is expected as `textAnnotationData` so it will be automatically handled.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -30,33 +42,123 @@ function getElementGeometryBuilderParams(iModel: IModelDb, _placementProps: Plac
...
+
+  /** Change the textual content, updating the element's geometry and placement accordingly.
+   * @see [[getAnnotation]] to extract the current annotation.
+   * @param annotation The new annotation
+   */
+  public setAnnotation(annotation: TextAnnotation) {
+    this.textAnnotationData = annotation ? JSON.stringify(annotation.toJSON()) : undefined;
+  }
+
+  protected constructor(props: TextAnnotation2dProps, iModel: IModelDb) {
```
MichaelSwigerAtBentley, line 63: Optionally, the constructor can be omitted in this case as well, since textAnnotationData is being redundantly assigned with no additional logic. Just add `declare` to the `public textAnnotationData` declaration so it is not overwritten by undefined, as noted in this 5.0 changelog: https://www.itwinjs.org/changehistory/#usedefineforclassfields

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: I agree, both to not loving it and stopping the ability to bypass the Element APIs. I will keep at as is for now as it is necessary for completeness

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: `textAnnotationData` shouldn't be a public property, nobody can do anything useful with it without round-tripping it through JSON.

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: Created an issue called [Expose entry point for TypeScript subclasses to disclose usage](https://github.com/iTwin/itwinjs-core/issues/8349).

I will remove the `onDelete` logic from this PR and will react to that issue being resolved by making sure the `AnnotationTextStyle` class discloses its usage.

Unsure if you would like me to attempt to resolve the new issue or if someone from core should do it since it involves the native code.

core/backend/src/annotations/FrameGeometry.ts
Alfonso-Martello: I will create a follow up PR to simplify this.

My thought process that led to the two values was that I like and wanted to preserve the current semantics of the `TextStyleSettings`, where the props are partial definitions and the instances are fully defined definitions. 

How I was maintaining those semantics was to have `undefined` mean "use the default value" on the props whereas “none” is the fully resolved “no” value in the instance.

I think the answer is to get rid of the "none" option and just have undefined mean both "use default value" and if it is the default value mean "no fill"/"no color". But I will do this in a separate PR.

Making `shape`/`fill` always be defined in the props is not possible because the `styleOverrides` are props, and we cannot force the overrides to always define these values because they would always override the referenced `AnnotationTextStyle`.

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: Because of confusion while manually testing. I noticed when calling `iModel.elements.getElement<TextAnnotation2d>(elementId)` that the returned element always had a seemingly extra `textAnnotationData` property. This is with `_textAnnotationData` being the private property and no public property called `textAnnotationData`:
image

I figured that was getting added for a reason so I decided to make it public.

I have now made it private.

> Why not instead have a public annotation: TextAnnotation | undefined that you initialize from the JSON string in the constructor, to replace get/setAnnotation?

In my first iteration, I did this. I was happy with it until I was updating the tests and stumbled across the [“produces a new object each time it is called”]( https://github.com/iTwin/itwinjs-core/blob/f143f8d0821fc63e4825b08b0c566a3b7c0bce71/core/backend/src/test/annotations/TextAnnotation.test.ts#L71) test. Then I realized I was changing the semantics to instead have a reference, which would open the door to someone doing something silly like modifying the `TextAnnotation2d.annotation` without realizing it. So, I wanted to preserve the behavior of getting a new object and my mind then went to getters/setters. We already had the `getAnnotation` and `setAnnotation` methods so I just kept using them

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -141,58 +151,149 @@ function scaleRange(range: Range2d, scale: number): void {
   range.high.scaleInPlace(scale);
 }
 
-function applyBlockSettings(target: TextStyleSettings, source: TextStyleSettings | TextStyleSettingsProps): TextStyleSettings {
+/**
+ * Applies block level settings (lineSpacingFactor, lineHeight, widthFactor, frame, and leader) to a [TextStyleSettings]($common).
+ * These must be set on the block, as they are meaningless on individual paragraphs/runs.
+ * Leaders are a special case. To a paragraph or run, leader settings are meaningless. However, a leader
+ * can override the block's leader settings. Setting `isLeader` to `true` makes the [TextBlock]($common) settings not override the leader's settings.
```
ben-polinsky:
```suggestion
 * Applies block level settings (lineSpacingFactor, lineHeight, widthFactor, frame, and leader) to a [TextStyleSettings]($common).
 * These must be set on the block, as they are meaningless on individual paragraphs/runs.
 * However, leaders are a special case and can override the block's leader settings.
 * Setting `isLeader` to `true` makes the [TextBlock]($common) settings not override the leader's settings.
```

ben-polinsky: Not sure if my suggestion is the best language, but `To a paragraph or run, leader settings are meaningless.` seemed redundant.

core/backend/src/test/AnnotationTestUtils.ts
```diff
@@ -17,10 +17,16 @@ export function doLayout(textBlock: TextBlock, args?: {
   findTextStyle?: FindTextStyle;
   findFontId?: FindFontId;
 }): TextBlockLayout {
+  const textStyleResolver = new TextStyleResolver({
+    textBlock,
+    iModel: {} as any,
```
ben-polinsky, line 22: It seems like iModel shouldn't be defined when iModelId is not defined - is that the case? Wondering if we can avoid the `{} as any`.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -68,32 +161,214 @@ export class TextAnnotation2d extends AnnotationElement2d {
...
+   * @param textAnnotationData Optional [[TextAnnotation]] JSON representation used to create the `TextAnnotation3d`. Essentially an empty element if not provided.
+   * @param code Optional code for the element.
    */
-  public getAnnotation(): TextAnnotation | undefined {
-    const json = this.jsonProperties.annotation;
-    return json ? TextAnnotation.fromJSON(json) : undefined;
+  public static create(iModelDb: IModelDb, category: Id64String, model: Id64String, placement: Placement3dProps, textAnnotationData?: TextAnnotationProps, code?: CodeProps): TextAnnotation3d {
+    const props: TextAnnotation3dProps = {
+      classFullName: this.classFullName,
+      textAnnotationData: JSON.stringify(textAnnotationData),
```
ben-polinsky, line 219: Do you want to stringify `textAnnotationData` if it is `undefined`?

core/backend/src/test/AnnotationTestUtils.ts
ben-polinsky: Oh I thought I checked to see if this was a test or not... if this only used like this in tests, then disregard my comment.

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: I believe that `JSON.stringify(undefined)` returns `undefined`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#return_value

There is a known issue with TypeScript where the function overload for `JSON.stringify` does not show that the return type can be `undefined`.

I could add a more clear check if that would be better? Something like `textAnnotationData: textAnnotationData !== undefined ? JSON.stringify(textAnnotationData) : undefined`?

core/backend/src/test/AnnotationTestUtils.ts
Alfonso-Martello: Almost, the iModel is used in two instances:
1. It is used to look up the `modelId` to see if it is a `Drawing` (and if it is, to get its `scaleFactor`).
2. It is used to look up the `AnnotationTextStyle` specified by the `textBlock`.

In the case of this test utility method, the iModel is not actually needed because we are not scaling (so `modelId` is undefined) AND because we are defining `findTextStyle`, which is an internal callback method specifically for the tests to allow us to get the `TextStyleSettings` without having to insert things into an iModel

Outside of tests the iModel will need to be provided to allow style look ups since normally the `findTextStyle` callback is not available

core/backend/src/annotations/TextAnnotationElement.ts
ben-polinsky: No no, my mistake. Sorry for the noise.

core/backend/src/test/AnnotationTestUtils.ts
ben-polinsky: Got it, thanks for the clarification.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -30,33 +41,115 @@ function getElementGeometryBuilderParams(iModel: IModelDb, _placementProps: Plac
...
 
+  /** Extract the textual content, if present.
+   * @see [[setAnnotation]] to change it.
+   */
+  public getAnnotation(): TextAnnotation | undefined {
+    const textAnnotationProps = parseTextAnnotationData(this._textAnnotationData);
+    return textAnnotationProps ? TextAnnotation.fromJSON(textAnnotationProps) : undefined;
+  }
+
+  /** Change the textual content, updating the element's geometry and placement accordingly.
```
pmconne: Confusing, the geometry and placement aren't actually updated until `toJSON` right?

core/common/src/ElementProps.ts
```diff
@@ -169,11 +168,10 @@ export interface GeometricElement3dProps extends GeometricElementProps {
...
-  jsonProperties?: {
-    [key: string]: any;
-    /** @beta */
-    annotation?: TextAnnotationProps;
-  };
+  /** The stringified JSON representation of the text annotation.
+   * @see [[TextAnnotationProps]]
```
pmconne:
```suggestion
   * @see [[TextAnnotationProps]] for the JSON representation.
   * @note Don't set this property directly - use [TextAnnotation3d.setAnnotation]($backend) to ensure the element's geometry and placement are recalculated.
```

core/common/src/ElementProps.ts
```diff
@@ -251,16 +249,15 @@ export interface GeometricElement2dProps extends GeometricElementProps {
...
  * @extensions
  */
 export interface TextAnnotation2dProps extends GeometricElement2dProps {
-  jsonProperties?: {
-    [key: string]: any;
-    /** @beta */
-    annotation?: TextAnnotationProps;
-  };
+  /** The stringified JSON representation of the text annotation.
+   * @see [[TextAnnotationProps]]
```
pmconne:
```suggestion
   * @see [[TextAnnotationProps]] for the JSON representation.
   * @note Don't set this property directly - use [TextAnnotation2d.setAnnotation]($backend) to ensure the element's geometry and placement are recalculated.
```

core/common/src/annotation/TextStyle.ts
```diff
@@ -273,62 +357,52 @@ export class TextStyleSettings {
...
+   * Returns a list of validation errors for this instance.
+   *
+   * A TextStyleSettings object may contain values that are invalid in all contexts.
+   * If this method returns any error strings, using the settings will likely result in rendering failures or runtime exceptions.
+   *
+   * This method only checks for universally invalid values. Additional domain-specific validation may be required depending on the context in which these settings are used.
+   *
+   * @returns An array of error strings describing the invalid values, or an empty array if the settings are valid.
+   */
+  public getValidationErrors(): string[] {
```
pmconne, line 389: What is the intended usage of this function? Debugging? Or are these errors expected to surface in user interface?

> A TextStyleSettings object may contain values that are invalid in all contexts.

Classes exist in large part to enforce invariants, i.e., prevent invalid states. Shouldn't you reject violations like empty font name or negative height during instantiation?

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: Correct, updated comment

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: No worries, thanks for the review 😄

core/common/src/annotation/TextStyle.ts
Alfonso-Martello: Its usage is in [AnnotationTextStyle.onInsert](https://github.com/iTwin/itwinjs-core/blob/b1577b9ebf9c1bde490b2272ace3b00f6b6e86d7/core/backend/src/annotations/TextAnnotationElement.ts#L358). Which is an insert time check to stop invalid text styles from being inserted into the iModel.

I added the `getValidationErrors` method to the `TextStyleSettings` so callers can at least try to validate the text style earlier.

The problem with doing the check during instantiation is that some of the default text style props are invalid. Specifically, `fontName` defaults to "", because I do not think we can reasonably assume what the default font should be. Thus, the default text style is actually invalid. 

I guess I could still check for errors during instantiation and just let someone relying on the defaults to be correct get an error. This would also mean removing the `TextStyleSettings.defaults` as that static value will always error

pmconne: Ok, if it prevents insert that's fine.
You should probably also check in `onUpdate`.

Alfonso-Martello: Added the check to `onUpdate` as well

core/common/src/ElementProps.ts
```diff
@@ -251,16 +250,16 @@ export interface GeometricElement2dProps extends GeometricElementProps {
   typeDefinition?: RelatedElementProps;
 }
 
-/** JSON representation of a [TextAnnotation2d]($backend).
+/** Properties that define a [TextAnnotation2d]($backend).
  * @public @preview
  * @extensions
  */
 export interface TextAnnotation2dProps extends GeometricElement2dProps {
-  jsonProperties?: {
```
aruniverse, line 259: isnt removing this breaking?

core/common/src/ElementProps.ts
```diff
@@ -630,3 +629,16 @@ export interface SheetReferenceProps extends SheetIndexEntryProps {
   /** The bis:Sheet that this bis:SheetReference is pointing to. */
   sheet?: RelatedElementProps;
 }
+
+/** Properties that define an [AnnotationTextStyle]($backend).
+ * @public @preview
```
aruniverse: i would prefer you start off with beta and then promote to public preview

aruniverse (approved):

core/common/src/ElementProps.ts
pmconne: `jsonProperties` is defined on `ElementProps`; this redeclaration just embellished the type.

Alfonso-Martello: Yes and no. I see that the `TextAnnotationXdProps` classes are marked as `preview`. However, the actual `jsonProperties.annotation` property is marked as `beta`.

I have been operating under the assumption that nobody is really using the text classes besides Drawing Production, and since we do not have users yet now is the time to break things. However, I also want to respect the API support policy.

The breaking change is not that the `TextAnnotationXdProps` no longer specify what their `jsonProperties` look like, since it still inherits this property from its base class (and anyone can set whatever data they like on the `jsonProperties`).

The breaking change is that the [text element classes](https://github.com/iTwin/itwinjs-core/blob/73738b31dd19cb3e915230b45a4fcadb35302b4d/core/backend/src/annotations/TextAnnotationElement.ts#L41) no longer look at `jsonProperties` to figure out the text content.

So essentially the `preview` classes were using a `beta` property that this PR now removes. Which will have a breaking effect on anyone using the `beta` property, but this `beta` property was the only way for the `preview` classes to get the text content.

If we want to take an abundance of caution, I could mark this property as deprecated and make the classes look at both places to determine the text content? And then remove it fully in 3 months?

pmconne: The property was @beta, therefore fine to remove per API support policies. Removing it means there's no reason to redeclare `jsonProperties` here.

core/common/src/ElementProps.ts
Alfonso-Martello: Done, also made the associated class beta

PR #8321, opened 2025-07-11
[SchemaFormatsProvider] Use labels, descriptions from KindOfQuantity when using presentation unit

hl662:
- Use labels, descriptions from KindOfQuantity when using presentation unit retrieved from SchemaFormatsProvider. 
- Strips out all EC Schema related properties from returned `FormatDefintion`.
- Change return type of `SchemaFormatsProvider.getFormat()` to be `FormatDefinition`, still aligning with base return type of `FormatsProvider`
- Improve error message for Ratio FormatType deserialization.


hl662:
This is kind of wonky because right now, `getFormat` in SchemaFormatsProvider returns a `SchemaItemFormatProps`. So if someone decides to deserialize this back into a SchemaItem based it will be very inconsistent, so one solution would be to change the return type to `FormatDefinition` instead...

core/ecschema-metadata/src/SchemaFormatsProvider.ts
```diff
@@ -60,6 +60,15 @@ export class SchemaFormatsProvider implements FormatsProvider {
     this.onFormatsChanged.raiseEvent({ formatsChanged });
   }
 
+  /** When using a presentation unit from a KindOfQuantity, the label and description should come from the KindOfQuantity */
+  private replaceWithKindOfQuantityProps(format: SchemaItemFormatProps, kindOfQuantity: KindOfQuantity): SchemaItemFormatProps {
```
ColinKerr: I think we should be returning the underlying format props rather than the schema item props ... at minimum we should use the base props interface.  stripping out the unnecessary props seems like a good idea ... if we end up serializing the extra pops they will end up in places we don't expect/want

hl662: okay, I turned this method into `convertToFormatDefinition` that strips out all the schema related properties. Also updated the return type of `getFormat` to be `FormatDefinition`, which aligns with the return value in the `FormatsProvider.getFormat()` interface in core-quantity

ColinKerr (approved):

PR #8326, opened 2025-07-15
Text fields

### LINKED ISSUE #6746: Text fields
Users have a need to place text that displays a stringified representation of the property of some other object (usually an element). For example, title blocks on sheets often contain tables that are to be filled in with a last modified date, a sheet number, etc.

In MicroStation this is accomplished using `TextField`s inside of a `TextBlock`. A TextField describes how to obtain and format the property. The displayed text changes automatically when the value of the property changes.

A similar solution to MicroStation. `TextField` would become another kind of `Run` within a `TextBlock`.

n/a

**Additional context**
Need to identify what kinds of properties and formatting people need us to support.

pmconne:
Fixes #6746.

Introduces `FieldRun` as a new type of `Run`, with a property accessor, host element, cached display string, and optional formatting specifications (TBD).

Currently fields are laid out and displayed just like ordinary text. In MicroStation they were displayed (by default, could be disabled via a setting) with a grey background to distinguish them from ordinary text. I'm awaiting feedback on whether that's desirable and - if so - should it be enabled by default and should it be customizable e.g. by the text style.

Introduces `ITextAnnotation` interface to be implemented by any `Element` subclass that can host fields. This is basically `ITextEdit` from MicroStation. Its `getTextBlocks` and `updateTextBlocks` methods are invoked by the new `ElementDrivesTextAnnotation` relationship introduced by https://github.com/iTwin/bis-schemas/pull/540 to update fields. Elements that implement `ITextAnnotation` should call `ElementDrivesTextAnnotation.updateFieldDependencies` from their `onInserted` and `onUpdated` methods.

Awaiting a new BisCore release with https://github.com/iTwin/bis-schemas/pull/540; until then the tests that rely on field dependency relationships will fail.

### TODO (follow-up PRs unless BisCore is delayed further)

- [ ] Formatting
- [ ] Extra metadata for JSON fields to inform formatting
- [x] Fields pointing to properties of aspects
- [ ] display-test-app key-ins
- [ ] NextVersion.md and example-code-snippets
- [ ] Toggleable field background?


core/common/src/annotation/TextBlock.ts
```diff
@@ -402,6 +404,114 @@ export class TabRun extends TextBlockComponent {
...
+
+export interface FieldRunProps extends TextBlockComponentProps {
+  readonly type: "field";
+  target: FieldTarget;
+  accessor: FieldPropertyAccessor;
+  cachedContent?: string;
+}
+
+export class FieldRun extends TextBlockComponent {
+  public static invalidContent = "####"; // maybe this should be specified by the text style?
```
pmconne: @Josh-Schifter do you think the string used when we can't resolve a field's display value should be a property of the text style? Do you think whether or not we display fields with a grey background like MicroStation (or a more customizable appearance) should be part of the text style?

core/common/src/annotation/TextBlock.ts
```diff
@@ -402,6 +406,146 @@ export class TabRun extends TextBlockComponent {
...
+   */
+  arrayIndex?: number;
+}
+
+/** A chain of property accesses that resolves to a primitive value that forms the basis of the displayed content
+ * of a [[FieldRun]].
+ * The chain may traverse through structs, arrays, and JSON objects.
+ * @beta
+ */
+export type FieldPropertyPath = FieldPropertyPathEntry[];
```
pmconne: @khanaffan @grigasp is there an existing/better way to encode an arbitrarily complex access string that may traverse arrays, structs, and JSON properties?

khanaffan: IN ECSql you can access using SQLite [JSON1](https://www.sqlite.org/changes.html) `json_extract()` function that allow nested json access via path

```
SELECT json_extract('{"a":2,"c":[4,5,{"f":7}]}', '$') → '{"a":2,"c":[4,5,{"f":7}]}'
SELECT json_extract('{"a":2,"c":[4,5,{"f":7}]}', '$.c') → '[4,5,{"f":7}]'
SELECT json_extract('{"a":2,"c":[4,5,{"f":7}]}', '$.c[2]') → '{"f":7}'
SELECT json_extract('{"a":2,"c":[4,5,{"f":7}]}', '$.c[2].f') → 7
SELECT json_extract('{"a":2,"c":[4,5],"f":7}','$.c','$.a') → '[[4,5],2]'
SELECT json_extract('{"a":2,"c":[4,5],"f":7}','$.c[#-1]') → 5
SELECT json_extract('{"a":2,"c":[4,5,{"f":7}]}', '$.x') → NULL
SELECT json_extract('{"a":2,"c":[4,5,{"f":7}]}', '$.x', '$.a') → '[null,2]'
SELECT json_extract('{"a":"xyz"}', '$.a') → 'xyz'
SELECT json_extract('{"a":null}', '$.a') → NULL
```

pmconne: But unless I've overlooked something, I cannot use a single ECSql statement (or even a series of them) to evaluate an access string like `Struct.Structs[0].jsonProperties.names[2]`, right? Because ECSql can only select entire arrays, not individual array elements.

pmconne: I should add that while my main goal is to fetch the property value, I'll almost certainly need some metadata about the property, like its kind-of-quantity and extended type. The property path as a series of property names can be used for both purposes.

core/common/src/annotation/TextBlock.ts
Josh-Schifter: My opinion is that neither of those should be controlled by the text style.  Our theory of text styles is that a style represents the 'purpose' or 'role' for the text within the sheet set.  Neither of those would depend on the purpose.

From the user POV, the 'couldn't resolve' text represents an error condition.  It should be something that is obviously not correct.  A string like ERROR or HELP_IM_ON_FIRE would be fine but I'm pretty sure we chose '####' so that we wouldn't have to localize it.

The desirability of the the grey background depends much more on the purpose of the *view* rather than the text style.  In an editing session, you almost always want the background.  In a viewing session, it's less obvious but I think the background can be helpful even then.

pmconne: OK, it sounds like we should produce the field background but provide some way to toggle off its display (view flag? new `GeometryClass`? TBD).

pmconne:
@diegoalexdiaz my unit tests now pass if I overwrite BisCore.ecschema.xml with the upcoming version that adds ITextAnnotation and ElementDrivesTextAnnotation. I am ready to consume a new version whenever it is published.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -38,7 +39,7 @@ function getElementGeometryBuilderParams(iModel: IModelDb, modelId: Id64String,
  * @see [[setAnnotation]] to change the textual content.
  * @public @preview
  */
-export class TextAnnotation2d extends AnnotationElement2d {
+export class TextAnnotation2d extends AnnotationElement2d /* implements ITextAnnotation */ {
```
Alfonso-Martello, line 42: Is `implement ITextAnnotation` commented out for a reason?

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -270,6 +293,48 @@ export class TextAnnotation3d extends GraphicalElement3d {
...
+  const annotation = elem.getAnnotation();
+  return annotation ? [{ textBlock: annotation.textBlock, id: undefined }] : [];
+}
+
+function updateTextBlocks(elem: TextAnnotation2d | TextAnnotation3d, textBlocks: TextBlockAndId[]): void {
+  assert(textBlocks.length === 1);
+  assert(textBlocks[0].id === undefined);
+
+  let annotation = elem.getAnnotation();
+  if (!annotation) {
```
Alfonso-Martello, line 330: Is there a code path where there is no annotation but there is still a text block?

core/backend/src/internal/annotations/fields.ts
```diff
@@ -0,0 +1,302 @@
+/*---------------------------------------------------------------------------------------------
```
Alfonso-Martello, line 1: Noticed that `fields.ts` is lowercase while most of the other files use PascalCase? Should this be renamed to `Fields.ts`?

core/common/src/annotation/TextBlock.ts
```diff
@@ -373,6 +374,203 @@ export class TabRun extends TextBlockComponent {
...
+   * If `propertyName` identifies a struct or array property, then additional [[accessors]] are required to identify the specific value.
+   * If `propertyName` (including any [[accessors]]) resolves to a JSON property, then additional [[jsonAccessors]] are required to identify a specific value within the JSON.
+   * Some examples:
+   * ```
+   * | Access String | propertyName | accessors | jsonAccessors |
+   * | ------------- | ------------ | --------- | ------------- |
+   * | name          | "name"       | undefined | undefined     |
+   * | spouse.name   | "spouse"     | [name]    | undefined     |
+   * | colors[2]     | "colors"     | [2]       | undefined     |
+   * | spouse.favoriteRestaurants[1].address | "spouse" | ["favoriteRestaurants", 1] | undefined |
```
Alfonso-Martello:
```suggestion
   * | spouse.favoriteRestaurants[1].address | "spouse" | ["favoriteRestaurants", 1, "address"] | undefined |
```

core/common/src/annotation/TextBlock.ts
```diff
@@ -373,6 +374,203 @@ export class TabRun extends TextBlockComponent {
...
+   * | jsonProperties.contactInfo.email | "jsonProperties" | undefined | ["contactInfo", "email"] |
+   * | spouse.jsonPropertes.contactInfo.phoneNumbers[0].areaCode | "spouse" | ["jsonProperties"] | ["contactInfo", "phoneNumbers", 0, "areaCode"] |
```
Alfonso-Martello:
```suggestion
   * | spouse.jsonProperties.contactInfo.phoneNumbers[0].areaCode | "spouse" | ["jsonProperties"] | ["contactInfo", "phoneNumbers", 0, "areaCode"] |
```

core/common/src/annotation/TextBlock.ts
```diff
@@ -373,6 +374,203 @@ export class TabRun extends TextBlockComponent {
...
+    return new FieldRun(this.toJSON());
+  }
+
+  /** Convert this FieldRun to a simple string representation. */
+  public override stringify(): string {
+    return this.cachedContent;
+  }
+
+  /** Returns true if `this` is equivalent to `other`. */
+  public override equals(other: TextBlockComponent): boolean {
```
Alfonso-Martello, line 527: Should this call `super.equals(other)` to validate that `styleOverrides` match?

core/common/src/test/annotations/TextBlock.test.ts
```diff
@@ -197,3 +197,177 @@ describe("TextBlock", () => {
...
+      // Modify the original formatter to ensure the FieldRun's copy is unaffected
+      formatter.formatType = "percentage";
+      formatter.precision = 3;
+      formatter.options.locale = "fr-FR";
+      formatter.options.style = "percent";
+
+      expect(fieldRun.formatter).to.deep.equal({ formatType: "currency", precision: 2, options: { locale: "en-US", style: "decimal" } });
+    });
+
+    it("deepyl clones propertyHost", () => {
```
Alfonso-Martello:
```suggestion
    it("deeply clones propertyHost", () => {
```

core/common/src/test/annotations/TextBlock.test.ts
```diff
@@ -197,3 +197,177 @@ describe("TextBlock", () => {
...
+      });
+
+      expect(fieldRun.stringify()).to.equal("cachedValue");
+    });
+  });
+
+  describe("equals", () => {
+    it("compares FieldRuns for equality", () => {
+      const baseProps = {
+        styleName: "fieldStyle",
```
Alfonso-Martello:
```suggestion
```

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: Yes it's @beta and the elements are @public @preview, extract-api doesn't permit that.

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: TextAnnotation2/3d store their TextBlock inside a TextAnnotation.

core/backend/src/internal/annotations/fields.ts
pmconne: It doesn't matter.

core/common/src/annotation/TextBlock.ts
```diff
@@ -373,6 +374,203 @@ export class TabRun extends TextBlockComponent {
...
+    if (!(other instanceof FieldRun)) {
```
pmconne:
```suggestion
    if (!(other instanceof FieldRun) || !super.equals(other)) {
```

core/common/src/annotation/TextBlock.ts
pmconne: good catch - fixed.

pmconne:
@Josh-Schifter @bbastings suggests that we should only draw the field background inside the text editor, in which case we wouldn't put it in the element's geometry stream. [This guy](https://www.youtube.com/watch?v=neJSgxUTRFw) asserts that users complain about having it turned on by default and [apparently](https://bentleysystems.service-now.com/community?id=community_blog&sys_id=b5d2a1431be9ca50f3fc5287624bcb2d) it is disabled by default in new versions of MicroStation. Do you disagree?

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: Yes, my more explicit thoughts were if the TextBlock comes from the TextAnnotation on the TextAnnotationXd, but the TextAnnotationXd has no TextAnnotation, then there is no TextBlock, so this clause would not get hit? Basically trying to understand if there is an expected situation where this would happen

pmconne: I was thinking about like a hypothetical "Edit Text" tool that sets the text on a TextAnnotationXd element that previously had no text. But the current API requires you to obtain the text block (and its Id) from the element so the only way to end up in this situation would be if you removed the annotation in between `getTextBlocks` and `updateTextBlocks`, in which case we should probably throw an error.

pmconne: pushed fix.

pmconne:
Force-merging because flaky tests on linux build:
```
--[ FAILURE: @itwin/analytical-backend ]--------------------[ 37.62 seconds ]--

##[error]Handle leak detected. Node was still running 30 seconds after tests completed.
Try running with the DEBUG_LEAKS env var set to see open handles.
Terminated

--[ FAILURE: @itwin/linear-referencing-backend ]------------[ 36.25 seconds ]--

##[error]Handle leak detected. Node was still running 30 seconds after tests completed.
Try running with the DEBUG_LEAKS env var set to see open handles.
Terminated

--[ FAILURE: @itwin/physical-material-backend ]-------------[ 36.14 seconds ]--

##[error]Handle leak detected. Node was still running 30 seconds after tests completed.
Try running with the DEBUG_LEAKS env var set to see open handles.
Terminated

--[ FAILURE: core-full-stack-tests ]---------------[ 8 minutes 35.1 seconds ]--

ApplicationInsights:Sender [
  'Ingestion endpoint could not be reached. This batch of telemetry items has been lost. Use Disk Retry Caching to enable resending of failed telemetry. Error:',
  [
    "[object Error]{ stack: 'AggregateError [ETIMEDOUT]: \n" +
      '    at internalConnectMultiple (node:net:1134:18)\n' +
      '    at internalConnectMultiple (node:net:1210:5)\n' +
      '    at Timeout.internalConnectMultipleTimeout (node:net:1742:5)\n' +
      '    at listOnTimeout (node:internal/timers:590:11)\n' +
      "    at process.processTimers (node:internal/timers:523:7)', message: '', name: 'AggregateError'"
  ]
  ...12 lines omitted...
  1 failing

  1) White-on-white reversal
       "before all" hook for "should not apply if background is not white":
     Error: Timeout of 240000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

Josh-Schifter:
> @Josh-Schifter @bbastings suggests that we should only draw the field background inside the text editor, in which case we wouldn't put it in the element's geometry stream. [This guy](https://www.youtube.com/watch?v=neJSgxUTRFw) asserts that users complain about having it turned on by default and [apparently](https://bentleysystems.service-now.com/community?id=community_blog&sys_id=b5d2a1431be9ca50f3fc5287624bcb2d) it is disabled by default in new versions of MicroStation. Do you disagree?

This all seems like sufficient evidence that showing the field background in the view is annoying at least some of the time.  To keep things simple, I suggest we start out by never displaying it in the view.  It will be easier to add it in later if we find out we do need it.

PR #8331, opened 2025-07-17
Changed `queryModelRanges` and `queryExtents` to use `ECSqlReader`

LiutaurasSk:
- Changed `queryModelRanges` and `queryModelExtents` to use `ECSqlReader`.
- One query is called to get all extents, if one or more extents for models are not returned by the query, another query is ran so correct status can be returned.


core/frontend/src/IModelConnection.ts
```diff
@@ -887,14 +892,134 @@ export namespace IModelConnection {
...
+              iModel_angles(ge.Yaw, ge.Pitch, ge.Roll),
+              iModel_bbox(
+                ge.BBoxLow.X, ge.BBoxLow.Y, ge.BBoxLow.Z,
+                ge.BBoxHigh.X, ge.BBoxHigh.Y, ge.BBoxHigh.Z
+              )
+            )
+          )
+        ) AS bbox
+      FROM bis.GeometricElement3d AS ge, bis.GeometricModel3d as gm
+      WHERE InVirtualSet(:ids64, ge.Model.Id) AND ge.Origin.X IS NOT NULL AND InVirtualSet(:ids64, gm.ECInstanceId) AND (gm.$->isNotSpatiallyLocated?=true OR gm.$->isNotSpatiallyLocated? IS NULL)
```
GytisCepk: I believe, our assumption here may have been wrong. When `isNotSpatiallyLocated` is `NULL`, it should be treated as if value is `false` and not `true`

core/frontend/src/IModelConnection.ts
```diff
@@ -887,14 +892,134 @@ export namespace IModelConnection {
...
+          false AS isGeometricModel
+      FROM bis.Model
+      WHERE InVirtualSet(:ids64, ECInstanceId)
+        AND ECInstanceId NOT IN (SELECT ECInstanceId FROM GeometricModels)`;
+
       const iModel = this._iModel;
-      if (!iModel.isOpen)
+      if (!iModel.isOpen) {
         return [];
+      }
```
GytisCepk: Please restore previous formatting

core/frontend/src/IModelConnection.ts
LiutaurasSk: Updated query

core/frontend/src/IModelConnection.ts
LiutaurasSk: Reverted formatting

LiutaurasSk:
@khanaffan could you please checkout the `modelExtentsQuery` and `modelExistenceQuery` queries

khanaffan (approved):

khanaffan:
> @khanaffan could you please checkout the `modelExtentsQuery` and `modelExistenceQuery` queries

Is there a reason that the model extends/range is not cached? It's kind of IO intense operation. Specially on hosted backend it going to take longer due to cloud sqlite. 

Any way for readonly conn it should cache for read/write like app it should not cache.

LiutaurasSk:
> Is there a reason that the model extends/range is not cached?

I had overlooked caching, added it now

core/frontend/src/IModelConnection.ts
```diff
@@ -779,6 +779,75 @@ export namespace IModelConnection {
...
+      FROM GeometricModels
+      UNION ALL
+      SELECT
+      ECInstanceId,
+        false AS isGeometricModel
+      FROM bis.Model
+      WHERE InVirtualSet(: ids64, ECInstanceId)
+        AND ECInstanceId NOT IN(SELECT ECInstanceId FROM GeometricModels)`;
+
+    private _loadedExtents: ModelExtentsProps[] = [];
```
GytisCepk, line 849: I don't see any tests to check if values are cached and then retrieved from cache

core/frontend/src/IModelConnection.ts
```diff
@@ -887,14 +961,73 @@ export namespace IModelConnection {
...
 
       if (typeof modelIds === "string")
         modelIds = [modelIds];
 
-      return IModelReadRpcInterface.getClientForRouting(iModel.routingContext.token).queryModelExtents(iModel.getRpcProps(), modelIds);
+      let remainingModelIds = this._getUnloadedModelIds(modelIds);
+      if (remainingModelIds.length > 0) {
+        for (const modelId of remainingModelIds) {
+          if (!Id64.isValidId64(modelId)) {
+            this._loadedExtents.push({ id: modelId, extents: Range3d.createNull(), status: IModelStatus.InvalidId });
```
GytisCepk: We should not cache results for invalid IDs. The reason for caching is to avoid expensive queries and checking for invalid IDs does not require running query.

GytisCepk: The other way you can looks at this is that cache is for extents of the models. If model does not or can't exists, why should there be an entry ?

core/frontend/src/IModelConnection.ts
GytisCepk, line 849: Cache should only be available for read-only connections. In editing application a new geometric element may get added which will invalidate cached extents

pmconne: The backend emits events to inform read-write connections when geometric elements change, which can be used to invalidate caches. This is used, for example, to keep graphics in sync with geometry modifications.

LiutaurasSk: Added tests for this case

core/frontend/src/IModelConnection.ts
LiutaurasSk: Removed caching of extents with invalid id

core/frontend/src/IModelConnection.ts
LiutaurasSk: Added handling for `onModelGeometryChanged` event

core/frontend/src/IModelConnection.ts
```diff
@@ -887,14 +967,86 @@ export namespace IModelConnection {
...
+      for (const modelId of modelIds) {
+        if (!Id64.isValidId64(modelId)) {
+          modelExtents.push({ id: modelId, extents: Range3d.createNull(), status: IModelStatus.InvalidId });
+        }
+      }
+
+      const getUnloadedModelIds = () => modelIds.filter((modelId) => !modelExtents.some((loadedExtent) => loadedExtent.id === modelId));
+
+      let remainingModelIds = getUnloadedModelIds();
+      if (remainingModelIds.length > 0) {
```
ben-polinsky: Is this length check necessary? I believe the `for...of` will not run if `remainingModelIds === 0`

LiutaurasSk: It was unnecessary, removed

ben-polinsky: Thanks.

PR #8333, opened 2025-07-17
Treat missing and empty description fields as equivalent in schema comparison  

rushikesh-bentley:
Issue Link: https://github.com/iTwin/bis-schema-validation/issues/65
 

- Identified that schema comparison was reporting a false diagnostic when one schema had "description": "" and the other had no "description" field.

- Added a helper function to treat missing (undefined) and empty ("") "description" values as equivalent.

- Updated the logic in the comparison function to skip diagnostics for such cases.

- Added a dedicated test case to verify that no diagnostic is reported for missing vs empty "description".


christophermlawson (approved):
Looks good

core/ecschema-editing/src/Validation/SchemaComparer.ts
```diff
@@ -66,6 +66,12 @@ function labelsMatch(label1?: string, label2?: string) {
   return label1 === label2;
 }
 
+function descriptionsAreEqual(desc1?: string, desc2?: string) {
+  desc1 = desc1 === undefined ? "" : desc1;
```
christophermlawson: This is fine, but you could use the shorthand:
desc1 = desc1 ?? "";

christophermlawson (approved):

core/ecschema-editing/src/Validation/SchemaComparer.ts
rushikesh-bentley: Thanks for the suggestion! I’ve updated the code using the ?? shorthand as recommended.

core/ecschema-editing/src/Validation/SchemaComparer.ts
```diff
@@ -66,6 +66,12 @@ function labelsMatch(label1?: string, label2?: string) {
   return label1 === label2;
 }
 
+function descriptionsAreEqual(desc1?: string, desc2?: string) {
+  desc1 = desc1 ?? "";
+  desc2 = desc2 ?? "";
+  return desc1 === desc2;
```
RohitPtnkr1996: Why update the argument values?
This function should only be checking.

function descriptionsAreEqual(desc1?: string, desc2?: string) {
  return (desc1 || "") === (desc2 || "");
}

Also, the function name is misleading, it sounds more of a statement than a check.
You might want to consider some like "areDescriptionsEqual" or similar to make it more readable.

rushikesh-bentley: Thanks for the feedback!
I've removed the argument updates and renamed the function to areStringsEqual to make it reusable for labels, descriptions, etc.

Issue #8350, opened 2025-07-22
Function to create a Transform obj from Helmert2DWithZOffset


DanRod1999:
Recently I created a static function in an [imodel-transformer PR](https://github.com/iTwin/imodel-transformer/pull/265) to convert `Helmert2DWithZOffset` objs to `Transform` objs. Currently in this PR it is part of the ImodelTransformer class.

Not sure if this really makes sense to live in the imodel-transformer repo, as its not a unique part of the transformer. I think it makes more sense to be part of the `Helmert2DWithZOffset`, `Transform`, or maybe some other part of the GeographicCRS class. Any feedback on if this sounds reasonable or not would be appreciated thanks.

This is the function:

```
  public static convertHelmertToTransform(
    helmert: Helmert2DWithZOffset | undefined
  ): Transform {
    if (!helmert) {
      return Transform.createIdentity();
    }

    const rotationXY = Matrix3d.createRotationAroundAxisIndex(
      AxisIndex.Z,
      Angle.createDegrees(helmert?.rotDeg)
    );
    rotationXY.scaleColumnsInPlace(helmert.scale, helmert.scale, 1.0);
    const translation = Vector3d.create(
      helmert.translationX,
      helmert.translationY,
      helmert.translationZ
    );
    const helmertTransform = Transform.createRefs(translation, rotationXY);

    return helmertTransform;
  }
```

saeeedtorabi:
@dassaf4 @saeeedtorabi

PR #8355, opened 2025-07-23
Add convertHelmertToTransform to Helmert2DWithZOffset Class

DanRod1999:
Proposed solution for: https://github.com/iTwin/itwinjs-core/issues/8350
Currently a version of this method exists in my PR linked in the issue, but I don't think it makes sense for this method to exist in the imodel-transformer class.

I think this or a static method in the Transformer class makes the most sense.

Thanks to @dassaf4 helping with the math operations.

### DIFF
--- a/common/api/summary/core-backend.exports.csv
+++ b/common/api/summary/core-backend.exports.csv
@@ -3,6 +3,7 @@ Release Tag;API Item Type;API Item Name
 public;interface;AcquireNewBriefcaseIdArg
 public;class;AnnotationElement2d
 preview;class;AnnotationElement2d
+beta;class;AnnotationTextStyle
 beta;type;AnyDb
 beta;function;appendFrameToBuilder
 beta;function;appendLeadersToBuilder
@@ -183,6 +184,7 @@ public;class;ElementAspect
 preview;class;ElementAspect
 beta;class;ElementDrivesElement
 beta;interface;ElementDrivesElementProps
+beta;class;ElementDrivesTextAnnotation
 public;class;ElementEncapsulatesElements
 public;class;ElementGroupsMembers
 public;interface;ElementGroupsMembersProps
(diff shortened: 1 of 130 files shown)


core/common/src/geometry/AdditionalTransform.ts
```diff
@@ -84,6 +84,25 @@ export class Helmert2DWithZOffset implements Helmert2DWithZOffsetProps {
       Math.abs(this.rotDeg - other.rotDeg) < Geometry.smallAngleDegrees &&
       Math.abs(this.scale - other.scale) < Geometry.smallFraction);
   }
+
+  public convertHelmertToTransform(): Transform {
+    if (!this)
```
ben-polinsky: Can `this` ever be falsey?

DanRod1999: additionalTransform and helmert2dWithZOffset can both be undefined, but not sure if calling this method would ever cause `this` to be undefined.

This was my only reason for thinking a static method in the Transformer class might make more sense, because then you could just pass in a helmert obj (undefined or not) and then return the appropriate transform. However, I'll check my PR to see if returning the identity is necessary. It might be simpler to avoid that in general if possible.

DanRod1999: Removed from this function. Users (like in transformer) will just need to account for the return value possibly being undefined

dassaf4 (approved):
I agree with this location. Helmert transformations are domain-specific, and as such don't really belong in core-geometry.

common/api/summary/core-backend.exports.csv
```diff
@@ -3,7 +3,6 @@ Release Tag;API Item Type;API Item Name
 public;interface;AcquireNewBriefcaseIdArg
 public;class;AnnotationElement2d
 preview;class;AnnotationElement2d
-beta;class;AnnotationTextStyle
```
aruniverse: review your merge, looks like it was bad

aruniverse (approved):

PR #8364, opened 2025-07-27
EXT_mesh_primitive_edge_visibility

pmconne:
Decodes and renders glTF edges encoded using [EXT_mesh_primitive_edge_visibility](https://github.com/KhronosGroup/glTF/pull/2479).


pmconne:
Superseded by #8366 - spec to be updated to encode normals directly instead of requiring client to compute them.

PR #8367, opened 2025-07-29
Export IModelIncrementalSchemaLocater

AureMonke:


core/backend/src/IModelIncrementalSchemaLocater.ts
```diff
@@ -11,7 +11,7 @@ import { IModelDb } from "./IModelDb";
 
 /**
  * A [[ECSqlSchemaLocater]]($ecschema-metadata) implementation that uses the [[IModelDb]] to load schemas incrementally.
- * @beta
+ * @internal
```
pmconne, line 14: Why are you exporting it if it's an internal API? Who's using it? Internal APIs are only supposed to be used by packages within the itwinjs-core monorepo.

AureMonke: First we want to test incremental schema loading in our iModel Quality Analytics Collection Agent (IQACA), see performance, possible drawbacks, etc, before making it available to all.

Issue #8197, opened 2025-06-09
TxnManager.deleteAllTxns not deleting briefcase changes


simihartstein:
**Describe the bug**
Changes still exist in the briefcase after calling `TxnManager.deleteAllTxns`,

**To Reproduce**
1. Open an iModel
2. Make some changes
4. Discard changes by calling deleteAllTxns
5. Restart the application (Not required, but it makes sure it's not just an in-memory caching issue)
6. Open the iModel

You will see all changes that were made in step 2 before discarding.

**Expected behavior**
I would expect that all the changes that were done in step 2 would be reverted in the briefcase. I would not expect the elements to still be present in the viewer.

**Desktop (please complete the applicable information):**

- Electron-Windows
- iTwin.js Version 5.0.0-dev.116

**Additional context**
If you make more changes after the discard and push, only the Txns made after the discard will be pushed. This piece of it is desired behavior.

I haven't investigated, but I think the current implementation could lead to data corruption if you edit an element after the discard that was edited in a conflicting way before the discard.

For example, if I
1. create two manholes
2. call deleteAllTxns
3. create a pipe connecting the manholes

Then I would expect the pipe will have relationships pointing to element ids that doesn't exist in the cloud iModel.

**My uninformed theory**
My best hypothesis is that there is some Txn table in the sqlite db that contains the incremental changesets, but it's separate from the "squashed" source data. And the deleteAllTxns function is deleting the records in the Txn table, but not undoing the changes in the squashed data.

@pmconne @alexdunae

aruniverse:
@simihartstein is this still relevant?

simihartstein:
@aruniverse Yes, we still need the workaround of deleting the entire briefcase.

khanaffan:
@simihartstein @aruniverse will following help. 
```ts
  /**
   * @alpha
   * Permanently discards any local changes made to this briefcase, reverting the briefcase to its last synchronized state.
   * This operation cannot be undone. By default, all locks held by this briefcase will be released unless the `retainLocks` option is specified.
   * @Note This operation can be performed at any point including after failed rebase attempts.
   * @param args - Options for discarding changes.
   * @param args.retainLocks - If `true`, retains all currently held locks after discarding changes. If omitted or `false`, all locks will be released.
   * @returns A promise that resolves when the operation is complete.
   * @throws May throw if discarding changes fails.
   */
  public async discardChanges(args?: { retainLocks?: true }): Promise<void>
```

[...]

PR #8375, opened 2025-07-30
TxnManager.deleteAllTxns not deleting briefcase changes

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-core/issues/8197

imodel-native: https://github.com/iTwin/imodel-native/pull/1180

The function DeleteAllTxns was simply clearing the Dgn_Txns table without reverting/reversing any of the local changes.
This logic has been fixed.

A new function ClearAllTxns does what DeleteAllTxns used to do.

### DIFF
--- a/core/backend/src/TxnManager.ts
+++ b/core/backend/src/TxnManager.ts
@@ -755,13 +755,22 @@ export class TxnManager {
    */
   public get hasLocalChanges(): boolean { return this.hasUnsavedChanges || this.hasPendingTxns; }
 
-  /** Destroy the record of all local changes that have yet to be saved and/or pushed.
+  /** Revert all local changes that have yet to be saved and/or pushed and destroy any record of them.
    * This permanently eradicates your changes - use with caution!
    * Typically, callers will want to subsequently use [[LockControl.releaseAllLocks]].
    * After calling this function, [[hasLocalChanges]], [[hasPendingTxns]], and [[hasUnsavedChanges]] will all be `false`.
    */
   public deleteAllTxns(): void {
     this._nativeDb.deleteAllTxns();
+  }
+
+  /** Destroy the record of all local changes that have yet to be saved and/or pushed.
+   * This permanently eradicates your changes - use with caution!
+   * Typically, callers will want to subsequently use [[LockControl.releaseAllLocks]].
+   * After calling this function, [[hasLocalChanges]], [[hasPendingTxns]], and [[hasUnsavedChanges]] will all be `false`.
+   */
+  public clearAllTxns(): void {
+    this._nativeDb.clearAllTxns();
   }
 
   /** Obtain a list of the EC instances that have been changed locally by the [[BriefcaseDb]] associated with this `TxnManager` and have not yet been pushed to the iModel.
--- a/core/backend/src/test/standalone/TxnManager.test.ts
+++ b/core/backend/src/test/standalone/TxnManager.test.ts
@@ -34,7 +34,7 @@ describe("TxnManager", () => {
       schemaLockHeld: true,
     };
     nativeDb.openIModel(pathname, OpenMode.ReadWrite, upgradeOptions);
-    nativeDb.deleteAllTxns();
+    nativeDb.clearAllTxns();
     nativeDb.closeFile();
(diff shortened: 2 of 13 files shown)


core/backend/src/test/standalone/TxnManager.test.ts
```diff
@@ -956,6 +956,25 @@ describe("TxnManager", () => {
       expect(imodel.txns.hasLocalChanges).to.be.false;
     });
 
+    it("TxnManager.deleteAllTxns not deleting briefcase changes", () => {
```
MichaelSwigerAtBentley: Could we get a new test for verifying that clearAllTxns() only deletes and does not revert?

RohitPtnkr1996: Done !
I have added core-backend and full-stack tests to cover clearAllTxns.

core/backend/src/TxnManager.ts
```diff
@@ -764,6 +764,15 @@ export class TxnManager {
     this._nativeDb.deleteAllTxns();
   }
 
+  /** Destroy the record of all local changes that have yet to be saved and/or pushed.
+   * This permanently eradicates your changes - use with caution!
+   * Typically, callers will want to subsequently use [[LockControl.releaseAllLocks]].
+   * After calling this function, [[hasLocalChanges]], [[hasPendingTxns]], and [[hasUnsavedChanges]] will all be `false`.
+   */
+  public clearAllTxns(): void {
```
khanaffan: This should be marked internal. There is no genuine case other than then checkpoint that ever need to delete txns without reversing them. In fact I would suggest not to even expose it. Caller should call `nativeDb.clearAllTxns()` explicitly.

khanaffan: In fact even in checkpoint case setting briefcase id to 0/unassigned should clear any txns as at that point the briefcase is not a briefcase any more its disconnected from time line and cannot push anything.

So instead of clearAllTxns() call caller instead should setBriefcaseId(Unassigned). Which should clearAllTxns().

RohitPtnkr1996: Done!
Although, resetBriefcases() has a check for pending txns which throws.
So, it is required to clear the txns table prior to calling it.
The function then clears any reversed txns that might still exist.

khanaffan:
So my other PR now includes discard changes function. You may be able to use that instead. May be wait for other PR to merge as your clearTxns() is duplicate for discardChanges()

core/backend/src/TxnManager.ts
```diff
@@ -1014,15 +1014,6 @@ export class TxnManager {
    */
   public get hasLocalChanges(): boolean { return this.hasUnsavedChanges || this.hasPendingTxns; }
 
-  /** Destroy the record of all local changes that have yet to be saved and/or pushed.
-   * This permanently eradicates your changes - use with caution!
-   * Typically, callers will want to subsequently use [[LockControl.releaseAllLocks]].
-   * After calling this function, [[hasLocalChanges]], [[hasPendingTxns]], and [[hasUnsavedChanges]] will all be `false`.
-   */
-  public deleteAllTxns(): void {
```
pmconne, line 1023: You can't delete a @public API. You can deprecate it.

core/backend/src/TxnManager.ts
RohitPtnkr1996: Resolving this comment as the changes have been reverted and tests have been updated to use discardChanges() instead.

core/backend/src/TxnManager.ts
RohitPtnkr1996: Resolving this comment as the changes have been reverted and tests have been updated to use discardChanges() instead.

kabentley:
This function `deleteAllTxns` is working exactly as it was designed and as I would expect it to work. It is used to turn a briefcase into a standalone file, among other things. For those purposes it should not reverse Txns. I don't think you should change its behavior. 

Instead, the *new* function should be called something like `abandonAllTxns` that can reverse and then delete local Txns. The term "abandon" is used to mean reverse and delete elsewhere.

PR #8385, opened 2025-07-31
Support stationBaseFactor when formatting Station values

hl662:
- Adds a new optional property `stationBaseFactor`, defaulted to 1, to the `FormatProps` interface and related Format classes that allows configuring the base factor used in station formatting calculations.

- As of now, this isn't supported on the native layer, and ECFormat hasn't been updated to include this property. Quantity Formatting documentation has been updated to disclose so.

Example:
With a FormatProps containing stationOffsetSize of 2, and a stationBaseFactor of 5, the denominator would be 500. Formatting a value of `1200` with the above format would result in "2+200".

With a FormatProps containing stationOffsetSize of 3, stationBaseFactor of 1, which is how formatting has always behaved, the denominator would be 1000. Formatting `1200` would result in "1+200".


core/quantity/src/Formatter/Format.ts
```diff
@@ -170,6 +178,14 @@ export class BaseFormat {
       if (!Number.isInteger(formatProps.stationOffsetSize) || formatProps.stationOffsetSize < 0) // must be a positive int > 0
         throw new QuantityError(QuantityStatus.InvalidJson, `The Format ${this.name} has an invalid 'stationOffsetSize' attribute. It should be a positive integer.`);
       this._stationOffsetSize = formatProps.stationOffsetSize;
+
+      if (undefined !== formatProps.stationBaseFactor) {
+        // optional
+        // Do we want to limit the range of values accepted for the base factor?
```
hl662: @rschili any thoughts? Right now, fractional base factors, like 0.5 would be accepted. Perhaps we can restrict the number to be positive integers

core/quantity/src/Formatter/Interfaces.ts
```diff
@@ -34,6 +34,9 @@ export interface FormatProps {
   readonly stationOffsetSize?: number;
   readonly stationSeparator?: string;
 
+  /** Optional base factor for station formatting. Defaults to 1. */
```
MichaelSwigerAtBentley: Maybe mention it should be a positive int in these comments?

hl662: I haven't made a call on it being an integer yet, but yes I'll make sure to document, it can currently be a positive fraction

MichaelSwigerAtBentley:
Nice test coverage btw

core/quantity/src/Formatter/Format.ts
rschili: No strong opinion here, I can see arguments for both. We probably don't need fractional, but there's not much benefit from restricting it to integers. What does @ckjackso think? Is it likely we may need fractionals in the future?

docs/learning/quantity/index.md
```diff
@@ -46,6 +50,34 @@ If you're developing a frontend application that takes advantage of the core qua
 
 For a detailed description of all the setting supported by FormatProp see the EC documentation on [Format](../../bis/ec/ec-format.md).
```
rschili, line 51: I think we should mention the new fields on that page, too ( leads to here https://www.itwinjs.org/bis/ec/ec-format/ )

core/quantity/src/Formatter/Formatter.ts
```diff
@@ -350,13 +350,15 @@ export class Formatter {
       // we assume that stopping value is always positive
       posMagnitude = Math.floor(posMagnitude * precisionScale + FPV_ROUNDFACTOR) / precisionScale;
 
-      const denominator = (Math.pow(10, spec.format.stationOffsetSize!));
+      const baseFactor = spec.format.stationBaseFactor ?? 1;
+      const stationOffsetSize = spec.format.stationOffsetSize || 3; // Provide default if undefined
```
ColinKerr: Did you just add this default?  It appears to be new and undocumented.

The native code has a default value of 0 for station offset size.

The default should be consistent and documented.  In an ideal world we would have required a value if the type of format was station.  

I think a default of 0 is reasonable because it will incentivize people to actually explicitly set one.

core/quantity/src/Formatter/Format.ts
ckjackso: Not being a deveroper I am not sure what some of these mean. If we are talking about being able to handle station intervals and how they are formatted.   We need to handle station where the major stations are 100' which is typically imperial.  The format would support major stationing format of 12+00, 13+00, 14+00,....  

If metric the major stations are 1000m meaning the major stations would be 1+000, 2+000. 3+000....

We also need formatting for minor stationing.  In imperial where the minor stationing for station 12+20 would simply be station +20 or 20.  This is common.   Metric for 1+220 would be +220 or 220.

There are cases in the US where the major stationing would be every 500' meaning that the major stations would be 0+00, 5+00, 10+00, .....  Hope this helps.

core/quantity/src/Formatter/Formatter.ts
hl662: We validate that stationOffsetSize is a required field when deserializing from JsonProps, I wanted to get rid of this non null assertion. I changed the stationOffsetSize default to be 0 now, documented it in the doccomment for that field too

core/quantity/src/Formatter/Format.ts
hl662: This PR will cover all the cases you've listed @ckjackso,

I'm limiting the stationBaseFactor to be positive integers for now, we can open it up to be fractional in the future if needed. It's easier to remove limitations than add limitations after the fact, that would be breaking

core/quantity/src/Formatter/Interfaces.ts
hl662: documented this field to be a positive integer now

docs/learning/quantity/index.md
hl662: added a line documenting this new field

ColinKerr (approved):

Issue #8377, opened 2025-07-30
Text fields Property Formatting Reqs


DanRod1999:
Continuation of https://github.com/iTwin/itwinjs-core/issues/6746
From Paul:
For formatting, I think quantity formatting and kinds-of-quantity mostly cover what we need for formatting fields. A few additional requirements for fields:

- [ ] Formatting of date/time values. MicroStation (pre-CONNECT) stored C# format strings. MicroStation CONNECT stores some other format string usable in pure C++ (details unclear, though I probably wrote it...need to look at the implementation).
- [ ] Specify optional prefix and/or suffix, e.g. "Planet Earth is {field value} away". You would think you could just put the prefix and suffix as separate runs from the field but I think they are not displayed if field value can't be resolved.
- [ ] Some people want to specify the casing of strings (unchanged, all upper, all lower, capitalize first word, capitalize each word).
- [ ] Boolean values localized as true/false, yes/no, on/off, enabled/disabled.
- [ ] Localized enum display values.

For fields pointing to JSON properties the user may want to explicitly specify the KoQ since there's no ECProperty from which to obtain it.

pmconne:
Let's put all of the information we need for formatting the field into the field itself. This means:
- For booleans, the true and false display strings.
- For enums, the localized display labels.
- For unitized values, the persistence and presentation units, along with required localized unit labels.

Pros:
- This eradicates the need to synchronously access the schema to obtain this information when we (re)-evaluate the field's display value, and the need to perform localization on the backend.
- It also normalizes the representation for BIS properties vs JSON properties (the latter would need to encode this information in the field anyway since they cannot encode it in the schema).
- It also reduces the amount of work we have to do when computing the field's display string since much of the information will have been precomputed.

Cons:
- It introduces some minor redundancy between schema and field.
- It introduces potential cache invalidation. e.g., if a schema is upgraded to change the persistence unit. However, schema upgrade does not currently do anything to recompute values stored using the old persistence unit anyway, so changing persistence unit would break their data in far worse ways than stale fields.

I think pros outweigh cons. @ColinKerr @khanaffan @DanRod1999 @aruniverse @Josh-Schifter what do you think?

hl662:
> For unitized values, the persistence and presentation units, along with required localized unit labels.

To clarify, both here mean the full JSON representation of each right? So the `UnitProps` and `FormatProps` (`FormatDefinition` is better) objects?

pmconne:
[...]

PR #8386, opened 2025-07-31
WIP Sync Quantity Formatting

DanRod1999:
This provides the needed sync version of create and createFromJSON. This also required sync version of getUnitsByName and UnitsConversion.

Due to modifying the interface there was need to add some sync version in the frontend api's however, they were only async by name i believe and did not use any async features. That being said, since they were frontend related there might be an easy way to avoid creating these async funcs but that would require some change to the Formatter Spec's implementation of create, so this was just simpler for now.

### DIFF
--- a/core/ecschema-metadata/src/Metadata/Unit.ts
+++ b/core/ecschema-metadata/src/Metadata/Unit.ts
@@ -63,19 +63,18 @@ export class Unit extends SchemaItem {
     return true;
   }
 
-  // /**
-  //  * Returns true if a conversion can be calculated between the input units
-  //  * @alpha
-  //  */
-  // public static areCompatible(unitA: Unit, unitB: Unit): boolean {
-  //   const unitAPhenomenon = unitA.phenomenon;
-  //   const unitBPhenomenon = unitB.phenomenon;
+  /**
+   * Returns true if a conversion can be calculated between the input units
+   * @alpha
+   */
+  public static areCompatibleSync(unitA: Unit, unitB: Unit): boolean {
+    const unitAPhenomenon = unitA.phenomenon;
+    const unitBPhenomenon = unitB.phenomenon;
 
-  //   if(SchemaItem.isSchemaItem(unitAPhenomenon) && SchemaItem.isSchemaItem(unitBPhenomenon))
-  //     if (!unitAPhenomenon || !unitBPhenomenon || !unitAPhenomenon.key.matches(unitBPhenomenon.key))
-  //       return false;
-  //   return true;
-  // }
+    if (!unitAPhenomenon || !unitBPhenomenon || !unitAPhenomenon.matches(unitBPhenomenon))
+      return false;
+    return true;
+  }
 
   /**
    * Type guard to check if the SchemaItem is of type Unit.
--- a/core/ecschema-metadata/src/Metadata/Unit.ts
+++ b/core/ecschema-metadata/src/Metadata/Unit.ts
@@ -63,6 +63,20 @@ export class Unit extends SchemaItem {
     return true;
   }
 
+  // /**
+  //  * Returns true if a conversion can be calculated between the input units
+  //  * @alpha
+  //  */
+  // public static areCompatible(unitA: Unit, unitB: Unit): boolean {
+  //   const unitAPhenomenon = unitA.phenomenon;
+  //   const unitBPhenomenon = unitB.phenomenon;
+
+  //   if(SchemaItem.isSchemaItem(unitAPhenomenon) && SchemaItem.isSchemaItem(unitBPhenomenon))
+  //     if (!unitAPhenomenon || !unitBPhenomenon || !unitAPhenomenon.key.matches(unitBPhenomenon.key))
+  //       return false;
+  //   return true;
+  // }
+
   /**
    * Type guard to check if the SchemaItem is of type Unit.
    * @param item The SchemaItem to check.
--- a/core/ecschema-metadata/src/UnitConversion/UnitConverter.ts
+++ b/core/ecschema-metadata/src/UnitConversion/UnitConverter.ts
@@ -58,6 +58,37 @@ export class UnitConverter {
   }
 
   /**
+   * Find conversion between from and to units, formatted {schemaName}.{schemaItemName} or {schemaName}:{schemaItemName}
+   * @param fromUnit SchemaItem full name of source unit
+   * @param toUnit SchemaItem full name of target unit
+   * @returns [[UnitConversion]] converting fromUnit -> toUnit with a factor and an offset
+   * @throws Error if from and to Units' SchemaItem is not found in Schema or Schema prefix is not found in SchemaContext
+   * @throws Error if from and to Units do not belong to the same phenomenon
+   * @throws Error if definitions' SchemaItems cannot be found in its own or referenced Schemas
+   * @throws Error if base units of source and target unit do not match
+   */
+  public calculateConversionSync(fromUnit: string, toUnit: string): UnitConversion {
+    const [fromSchemaName, fromSchemaItemName] = SchemaItem.parseFullName(fromUnit);
+    const [toSchemaName, toSchemaItemName] = SchemaItem.parseFullName(toUnit);
+    const fromSchemaKey = new SchemaKey(fromSchemaName);
+    const toSchemaKey = new SchemaKey(toSchemaName);
+
+    const fromSchema = this._context.getSchemaSync(fromSchemaKey);
+    const toSchema = this._context.getSchemaSync(toSchemaKey);
+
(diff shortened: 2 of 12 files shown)


core/ecschema-metadata/src/UnitProvider/SchemaUnitProvider.ts
```diff
@@ -238,6 +316,37 @@ export class SchemaUnitProvider implements UnitsProvider {
...
+        isValid: true,
+        system: unit.unitSystem === undefined ? "" : unit.unitSystem.fullName,
+      };
+    }
+
+    const invertsUnit = unit.invertsUnit;
+    if (!invertsUnit)
+      return new BadUnit();
+    else {
+      const invertsUnitItem = this._context.getSchemaItemSync(invertsUnit.fullName, Unit);
```
DanRod1999, line 339: Since invertsUnit is lazily loaded, we need a way to get the unit sync, so I'm using the units name and type as Unit to find it.

core/ecschema-metadata/src/UnitConversion/UnitTree.ts
```diff
@@ -138,6 +196,29 @@ export class UnitGraph {
...
+    }
+
+    if (this._graph.hasNode(unit.key.fullName))
+      return;
+
+    this._graph.setNode(unit.key.fullName, unit);
+    if (this.isIdentity(unit))
+      return;
+
+    this.addUnitToGraphSync(unit);
```
hl662, line 219: while everything looks fine on paper, it seems that both sync and async methods of the unitConverter will use the same `_graph` property. Do we have to worry about race conditions...

This is probably very minor, and seems like the graph acts as a cache in this sense, so the consequence might just be a very tiny effect of no cache hit

core/quantity/src/Interfaces.ts
```diff
@@ -106,6 +105,17 @@ export interface UnitsProvider {
   getConversion(fromUnit: UnitProps, toUnit: UnitProps): Promise<UnitConversionProps>;
 }
 
+/** This interface is implemented by the class that is responsible for locating units by name or label and providing conversion values between units.
+ * The methods to be implemented are async allowing the UnitsProvider to query the backend when necessary to look up unit definition and conversion rules.
+ * @beta
+ */
+export interface UnitsProviderSync {
+  // findUnit(unitLabel: string, schemaName?: string, phenomenon?: string, unitSystem?: string): Promise<UnitProps>;
+  // getUnitsByFamily(phenomenon: string): Promise<UnitProps[]>;
```
hl662: I'm fine with leaving these commented out, the use case for this interface is very specific, and we'll only open it up if users really need it (not want)

core/ecschema-metadata/src/test/UnitConversion/CrossSchema.test.ts
```diff
@@ -41,18 +41,37 @@ describe("Cross Schema unit definitions tests", () => {
...
+          `${test.input} ${test.from} in ${test.to} should be ${test.expect}
+          and not ${actual} error = ${Math.abs(test.expect - actual)} > ${tolerance}`,
+        ).to.be.true;
+      });
+    });
+  });
+
+  describe("Sync", () => {
+    testData.forEach((test: TestData) => {
+      it(`(sync) should convert ${test.fromSchema}:${test.from} to ${test.toSchema}:${test.to}`,() => {
```
DanRod1999, line 63: This added about 1.5k tests. Completion time was still fast (3s for 3098 tests). Will remove if they are unnecessary.

core/quantity/src/Interfaces.ts
DanRod1999: I removed since they were a bit confusing having promises in the definitions within the "sync" version.

core/ecschema-metadata/src/Metadata/Unit.ts
```diff
@@ -63,6 +63,19 @@ export class Unit extends SchemaItem {
...
+   */
+  public static areCompatibleSync(unitA: Unit, unitB: Unit): boolean {
+    const unitAPhenomenon = unitA.phenomenon;
+    const unitBPhenomenon = unitB.phenomenon;
+
+    if (!unitAPhenomenon || !unitBPhenomenon || !unitAPhenomenon.matches(unitBPhenomenon))
+      return false;
+    return true;
+  }
+
```
ColinKerr: because this method only needs the keys we can have a single implementation where the async calls the sync method.  The LazyLoadedPhenomenon has all the information needed to see if their key matches, so the async call is actually wrong to await.

core/ecschema-metadata/src/UnitConversion/UnitConverter.ts
```diff
@@ -57,6 +57,37 @@ export class UnitConverter {
...
+   * @throws Error if base units of source and target unit do not match
+   */
+  public calculateConversionSync(fromUnit: string, toUnit: string): UnitConversion {
+    const [fromSchemaName, fromSchemaItemName] = SchemaItem.parseFullName(fromUnit);
+    const [toSchemaName, toSchemaItemName] = SchemaItem.parseFullName(toUnit);
+    const fromSchemaKey = new SchemaKey(fromSchemaName);
+    const toSchemaKey = new SchemaKey(toSchemaName);
+
+    const fromSchema = this._context.getSchemaSync(fromSchemaKey);
+    const toSchema = this._context.getSchemaSync(toSchemaKey);
```
ColinKerr, line 77: These sync calls to the schema context make me uncomfortable.  If we are on the frontend or other place where the schemas are loaded asynchronously via a network call the sync methods behavior is to throw an exception if we are awaiting to load them.

https://github.com/iTwin/itwinjs-core/blob/7bf3122900d0f5f685c280c1ce9c1f574ce64088/core/ecschema-metadata/src/Context.ts#L196

If the schema has not been loaded then any locater that must make a network request will throw an exception if you call the sync method

https://github.com/iTwin/itwinjs-core/blob/7bf3122900d0f5f685c280c1ce9c1f574ce64088/core/ecschema-rpc/common/src/ECSchemaRpcLocater.ts#L64

If we are sure that the schemas are already loaded when we start calling this stuff then it will be OK but this seems fragile and prone to more failures as we start incrementally loading schemas more aggressively.

It is possible to get and cache the conversions in an async context then access them in sync code.

If there is code where we are forced to be sync due to current API then we should document and plan on fixing that in the future.

core/ecschema-metadata/src/Metadata/Unit.ts
DanRod1999: removed my method and updated the current to be sync

core/ecschema-metadata/src/UnitConversion/UnitConverter.ts
DanRod1999: @pmconne, since this is also only intended for use in the backend, would this possibly affect you in any way? Sorry I'm unaware if being in the backend only would change matters or not.

> It is possible to get and cache the conversions in an async context then access them in sync code.

I assume its possible to load these schemas before hand like you said, and then only after that we would call the code needed for text fields.

pmconne: I assume it's possible for the backend to load every single schema into its context at startup so that fields will be able to rely on them having been cached, but is that desirable or even useful? What is the problem with the backend synchronously accessing schemas?

DanRod1999: > I assume it's possible for the backend to load every single schema into its context at startup so that fields will be able to rely on them having been cached, but is that desirable or even useful?

I believe @hl662 said this is what wimse used to do

DanRod1999:
After talking with Paul, we are considering [this](https://github.com/iTwin/itwinjs-core/issues/8377#issuecomment-3197960019) as a new way to preform formatting of text fields. If we proceed, we can close this PR, as we will no longer need sync versions for formatting

DanRod1999:
closing as this should no longer be needed and this pr is succeeded by https://github.com/iTwin/itwinjs-core/pull/8440

core/ecschema-metadata/src/UnitConversion/UnitConverter.ts
ColinKerr: WiMSE loads the schemas asynchronously but the underlying mechanism loads entire schema JSONs as one request.  That said WiMSE does load all schemas upfront because it needs all of them to populate the UI.

This has significant performance implications when there are iModels with a large number of schemas.

There is a current WIP project to load schemas in stages so we have 'stubs' loaded initially then the content is loaded in a second step and maybe in the future as needed.  https://github.com/iTwin/itwinjs-core/tree/master/core/ecschema-metadata/src/IncrementalLoading. This will avoid the need to load them in C++ and allow us to incrementally load them on frontend or backend.

PR #8387, opened 2025-08-01
Improve ClassRegistry documentation and add additional tests

rschili:
Due to poor documentation, a bug was reported which turned out to be a problem with the way the registered class was designed. Added test and documentation to help get it right.


arbrumbaugh:
lgtm!

PR #8391, opened 2025-08-01
Start cleaning up lint warnings

tcobbs-bentley:
Start cleaning up lint warnings. Because there are so many, the cleanup is going to be split into multiple PRs, since this one is already getting unwieldy on GitHub with so many modifications.

1. Where possible, clean up code to not require non-null assertions (`!`).
2. Create new `expectDefined` and `expectNotNull` functions for use when there isn't an obviously correct way to get rid of the `!`. As time passes, code using `expectDefined` and `expectNotNull` will be updated to stop using it.
3. Fix code that triggers the `@typescript-eslint/restrict-template-expressions` warning.


tcobbs-bentley:
@pmconne @aruniverse @bbastings @ben-polinsky I'm trying to clean up lint warnings (most of which are for using `!`), and I'm not sure if this possible solution is an appropriate one. Note that for some places, there is a clear and obvious solution to get rid of the `!`, but for most, there isn't, which is why I created `expectDefined` here. If this seems like a reasonable solution, I will continue, but if not, then I think the individual instances are going to have to all be fixed with people that are very familiar with the code. I will then continue looking for and fixing places that have a fix that is obvious to me, but I will remove the new `Expect.ts` file and remove all the places that `expectDefined` is used.

pmconne:
A non-null assertion means one of two things:
1. I the programmer can prove logically that this value can never be undefined. The compiler fails to infer that. I'm telling the compiler it can trust me.
2. I the programmer am failing to handle the real possibility that this value might be null/undefined.

Situation 2 should be fixed by handling the possibility of null/undefined.
Situation 1 should ideally be fixed by adjusting the code so that the compiler can see that the value can't possibly be null. Failing that, the programmer can make their intention clear by using `assert` instead of an ambiguous and easily-overlooked exclamation mark.

pmconne:
Are you proposing `expectDefined` as a temporary solution so we can quickly get rid of the lint warnings, then come back later and clean up all the `expectDefined` calls to do proper handling of undefined/null values?

ben-polinsky:
A possible issue I see with `expectDefined` is that there may be cases where the undefined "works" (or at least doesn't throw an error and allows execution to continue).

Contrived example:

```ts
const noName: string = person?.name!

function greet(name: string){
 return `Hello, ${name}`
}

return greet(noName) // "quietly" broken, but our users have been working with it, and not expecting an error.
```

tcobbs-bentley:
> Are you proposing `expectDefined` as a temporary solution so we can quickly get rid of the lint warnings, then come back later and clean up all the `expectDefined` calls to do proper handling of undefined/null values?

I hadn't originally created it for that, but I can see using it that way. It's trivial to search for so that the proper people can truly fix the code that they understand.

Side note: we already have an [`assert`](https://github.com/iTwin/itwinjs-core/blob/master/core/bentley/src/Assert.ts) function, but that function just silently ignores failures unless `process.env.NODE_ENV === "development"`. I actually considered using that, but decided that it probably wasn't what we wanted.

pmconne:
> Side note: we already have an [`assert`](https://github.com/iTwin/itwinjs-core/blob/master/core/bentley/src/Assert.ts) function, but that function just silently ignores failures unless `process.env.NODE_ENV === "development"`. I actually considered using that, but decided that it probably wasn't what we wanted.

That's exactly what we want for situation 1 - `assert` means I'm so confident that this statement is true, it would be a waste of time to try to verify it at run-time. If you lack that confidence, we're in situation 2, which calls for `expectDefined`.

I think `expectDefined` is a reasonable way to narrow the tech debt and turn `non-null-assertions` into an error ASAP.

core/common/src/HiddenLine.ts
```diff
@@ -139,7 +139,10 @@ export namespace HiddenLine {
         return true;
       else if (this.ovrColor !== other.ovrColor || this.pattern !== other.pattern || this.width !== other.width)
         return false;
+      else if ((undefined === this.color) !== (undefined === other.color))
+        return false;
```
tcobbs-bentley, line 143: @pmconne This check that they are both undefined or both not undefined was missing before, and as far as I can tell, that was a logic error. I added it based on a bunch of other `equal` functions that do something similar. Am I missing something in the logic of this `equal` function, or was the logic indeed broken before? Note that my eslint-disable below is only appropriate with this check in place.

core/common/src/HiddenLine.ts
```diff
@@ -191,6 +194,8 @@ export namespace HiddenLine {
       else if (json instanceof Settings)
         return json;
       else
+        // Undefined check happened in JsonUtils.isEmptyObjectOrUndefined()
```
pmconne: If you use `isNonEmptyObject` instead, the compiler will know it's not `undefined` and you can get rid of the non-null assertion.

core/common/src/HiddenLine.ts
pmconne: Yep, the linter caught a bug and you fixed it - thanks.

core/common/src/HiddenLine.ts
tcobbs-bentley: Done.

core/geometry/src/Geometry.ts
```diff
@@ -1325,11 +1325,11 @@ export class Geometry {
   public static cloneMembers<T extends Cloneable<T>>(array: T[] | undefined): T[] | undefined {
     if (array === undefined)
       return undefined;
-    const clonedArray: T[] = [];
+    const clonedArray: (T | undefined)[] = [];
```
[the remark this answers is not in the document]
tcobbs-bentley: This is intentional, since this function was already intentionally returning values that don't match the declared return type. The function is deprecated, and changing its behavior would not be acceptable.

common/api/core-common.api.md
```diff
@@ -2906,16 +2906,6 @@ export enum ECSqlValueType {
     StructArray = 15
 }
```
tcobbs-bentley: @aruniverse I don't understand why this file updated, or core-frontent.api.md, core-quantity.api.md, eschema-metadata.api.md, core-common.exports.csv, or core-frontend.exports.csv. As far as I know, I didn't make these changes, although I did merge changes from origin/master. Did some other PR get force merged without extract-api running?

core/frontend/src/tile/map/MapTileTree.ts
```diff
@@ -416,6 +416,7 @@ export class MapTileTree extends RealityTileTree {
         this._mercatorFractionToDb.multiplyPoint3d(gridPoint, scratchCorner);
         if (this.globeMode !== GlobeMode.Ellipsoid || this.cartesianRange.containsPoint(scratchCorner)) {
           if (reprojected !== undefined && reprojected[i] !== undefined)
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             reprojected[i]!.clone(gridPoint);
           else
```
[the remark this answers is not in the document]
tcobbs-bentley: We just verified that `reprojected[i]` is not undefined, and yet the compiler still thinks it might be for some reason.

common/api/core-common.api.md
tcobbs-bentley: This is my fault. I'll fix it.

core/frontend/src/tile/map/MapLayerSources.ts
```diff
@@ -196,16 +196,24 @@ export class MapLayerSources {
...
 
   private static getMapBoxLayerSource(): MapLayerSource[] {
     const mapLayerSources: MapLayerSource[] = [];
+    // We know these hard-coded sources are valid, so ! is okay here.
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     mapLayerSources.push(MapLayerSource.fromBackgroundMapProps({ providerName: "MapBoxProvider", providerData: { mapType: BackgroundMapType.Street } })!);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     mapLayerSources.push(MapLayerSource.fromBackgroundMapProps({ providerName: "MapBoxProvider", providerData: { mapType: BackgroundMapType.Aerial } })!);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     mapLayerSources.push(MapLayerSource.fromBackgroundMapProps({ providerName: "MapBoxProvider", providerData: { mapType: BackgroundMapType.Hybrid } })!);
```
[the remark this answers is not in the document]
tcobbs-bentley: No, we want to remove _inappropriate_ non-null assertions. Appropriate ones will be left with the eslint-disable comment.

core/frontend/src/internal/render/webgl/ScratchDrawParams.ts
```diff
@@ -6,23 +6,28 @@ import { CachedGeometry } from "./CachedGeometry";
 import { DrawParams, ShaderProgramParams } from "./DrawCommand";
 import { Target } from "./Target";
 
-let progParams: ShaderProgramParams | undefined;
-let drawParams: DrawParams | undefined;
+interface ScratchDrawParams {
```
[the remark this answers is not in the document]
tcobbs-bentley: Our eslint rules disallow type aliases, and require us to instead use interfaces.

dassaf4:
core-geometry will be taken care of by a separate PR since it contains 1131 lint errors (well, 1130, thanks @tcobbs-bentley !).

I'll update here once we have a PR started.

core/bentley/src/Expect.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Expects that a value is defined (not undefined).
+ * @note Almost always, places that use this function should in the future be cleaned up to properly
+ * handle undefined values instead of throwing an error.
+ * @param value The value to check.
+ * @param message Optional error message to be used in thrown Error.
+ * @returns The original value if it is defined.
+ * @throws Error if the value is undefined.
+ * @beta
```
pmconne:
```suggestion
 * @internal
```

We don't want anybody outside of itwinjs-core using this, right? We want to use it temporarily and then remove it ASAP after cleaning up all the usages.

core/bentley/src/Expect.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Expects that a value is not null.
+ * @note Almost always, places that use this function should in the future be cleaned up to properly
+ * handle null values instead of throwing an error.
+ * @param value The value to check.
+ * @param message Optional error message to be used in thrown Error.
+ * @returns The original value if it is not null.
+ * @throws Error if the value is null.
+ * @beta
```
pmconne:
```suggestion
 * @internal
```
We don't want anybody outside of itwinjs-core using this, right? We want to use it temporarily and then remove it ASAP after cleaning up all the usages.

tcobbs-bentley:
> A possible issue I see with `expectDefined` is that there may be cases where the undefined "works" (or at least doesn't throw an error and allows execution to continue).
> 
> Contrived example:
> 
> ```ts
> const noName: string = person?.name!
> 
> function greet(name: string){
>  return `Hello, ${name}`
> }
> 
> return greet(noName) // "quietly" broken, but our users have been working with it, and not expecting an error.
> ```

I found at least [one place](https://github.com/iTwin/itwinjs-core/blob/8fab0961d90b7b47c94d2023d531a2482c3e87bf/core/geometry/src/Geometry.ts#L1319-L1325) where having an undefined value go somewhere it doesn't belong is in fact expected behavior. I didn't put `expectDefined` there. Instead, I [updated the code](https://github.com/iTwin/itwinjs-core/blob/8fab0961d90b7b47c94d2023d531a2482c3e87bf/core/geometry/src/Geometry.ts#L1328) to pretend that undefined was allowed, and later [used `as`](https://github.com/iTwin/itwinjs-core/blob/8fab0961d90b7b47c94d2023d531a2482c3e87bf/core/geometry/src/Geometry.ts#L1332) to get back to the expected type.

Having said that, almost all the places I'm making this change, having an undefined would have resulted in an immediate thrown exception due to the undefined value being accessed as an object. If you see any places where you feel my changes are dangerous, please let me know.

core/bentley/src/Expect.ts
tcobbs-bentley: While the functions have potential long-term utility, they weren't created for that, so I changed them both to internal.

tcobbs-bentley:
@pmconne One of my changes triggered a test failure in GraphicsProvider.test.ts. The problem is that the test creates a subclass of `BlankConnection`, which is itself a subclass of `IModelConnection`. The (internal-only) `IModelConnection` constructor makes the name prop optional, but then passes that into `IModel.initialize`, where `name` is required. I put an `expectDefined(iModelProps.name)` here, but since that makes the API itself broken, maybe that should instead be `iModelProps.name ?? "<undefined>"`?

Edit: It seems like changing it to have a default name is the only thing that won't break the API, so I did that [here](https://github.com/iTwin/itwinjs-core/pull/8391/commits/1416cb997fa6840886263273d522b9351284da6d).

core/backend/src/IModelDb.ts
```diff
@@ -2117,8 +2117,10 @@ export namespace IModelDb {
     public getElement<T extends Element>(elementId: Id64String | GuidString | Code | ElementLoadProps, elementClass?: EntityClassType<Element>): T {
       const element = this.tryGetElement<T>(elementId, elementClass);
       if (undefined === element) {
-        if (typeof elementId === "string" || elementId instanceof Code)
+        if (typeof elementId === "string")
           throw new IModelError(IModelStatus.NotFound, `Element=${elementId}`);
+        else if (elementId instanceof Code)
+          throw new IModelError(IModelStatus.NotFound, `Element=${JSON.stringify(elementId)}`);
```
pmconne:
```suggestion
        if (typeof elementId === "string" || elementId instanceof Code)
          throw new IModelError(IModelStatus.NotFound, `Element=${elementId}`);
```

`${thing}` calls `thing.toString()`. `Code` overrides `toString` to provide a string representation.

pmconne: Is linter flagging this as a rule violation? If so, which rule?

tcobbs-bentley: Yes:
```
/Users/travis/src/itwinjs-core/core/backend/src/IModelDb.ts
  2121:67  warning  Invalid type "string | Code" of template literal expression  @typescript-eslint/restrict-template-expressions
```

As far as I can tell, the lint rule requires anything that goes into a `${}` expression in a `` ` `` string to be an actual string, and not just something that is capable of being converted to a string. I can use `${elementId.toString()}` instead of `${JSON.stringify(elementId)}`.

tcobbs-bentley: Note: the lint rule intentionally doesn't treat the existence of `toString` as being ok, because that is defined too many places where useless results are produced (such as `[object Object]`).

tcobbs-bentley: Also, since `"blah".toString()` just returns `"blah"`, I can just add `.toString()` to the original code, without adding the extra `else if`.

tcobbs-bentley: I updated the code to just add `toString()` to the original code.

core/common/src/rpc/web/BentleyCloudRpcProtocol.ts
```diff
@@ -103,7 +103,7 @@ export abstract class BentleyCloudRpcProtocol extends WebAppRpcProtocol {
       }
 
       iTwinId = encodeURIComponent(token.iTwinId || "");
-      iModelId = encodeURIComponent(token.iModelId!);
+      iModelId = encodeURIComponent(token.iModelId || "");
```
tcobbs-bentley, line 106: @pmconne Using `expectDefined` here results in a test failure, which is why I updated it to be the same as the `iTwinId` line above. It seems like this will generate an invalid URL if either `iTwinId` or `iModelId` is undefined. Is that the intended behavior?

pmconne: I can't tell exactly what is intended here. However `token.iModelId` will definitely be defined unless `RpcConfiguration.disableRoutingValidation` is `true` (scroll up a few lines to see the check). It's `false` by default. What was the test failure? Did the test actually care about making valid RPC calls? `encodeURIComponent(undefined)` will return the string "undefined".

If the test doesn't really care about a valid URI then I'd consider just changing `token = { key: "" }` to `token = { key: "", iModelId: "" }`.

tcobbs-bentley: Here is the failure before I removed the `expectDefined`:
```
Expected value to be defined, but it was undefined.
Error: Expected value to be defined, but it was undefined.
    at expectDefined (Volumes/Builds/vsts_work_a/3/s/core/bentley/src/Expect.ts:18:11)
    at protocol.supplyPathForOperation (Volumes/Builds/vsts_work_a/3/s/core/common/src/rpc/web/BentleyCloudRpcProtocol.ts:106:50)
    at new WebAppRpcRequest (Volumes/Builds/vsts_work_a/3/s/core/common/src/rpc/web/WebAppRpcRequest.ts:83:31)
    at TestRpcInterface.forward (Volumes/Builds/vsts_work_a/3/s/core/common/src/RpcInterface.ts:117:23)
    at TestRpcInterface.op9 (Volumes/Builds/vsts_work_a/3/s/full-stack-tests/rpc/src/common/TestRpcInterface.ts:91:17)
    at TestRpcInterface.proxy.<computed> [as op9] (Volumes/Builds/vsts_work_a/3/s/core/common/src/rpc/core/RpcRegistry.ts:216:29)
    at Context.<anonymous> (Volumes/Builds/vsts_work_a/3/s/full-stack-tests/rpc/src/frontend/RpcInterface.test.ts:134:29)
    at callFn (http://localhost:3020/@/Volumes/Builds/vsts_work_a/3/s/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runnable.js:364:21)
    at Test$4.Runnable$3.run (http://localhost:3020/@/Volumes/Builds/vsts_work_a/3/s/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runnable.js:352:5)
    at Runner.runTest (http://localhost:3020/@/Volumes/Builds/vsts_work_a/3/s/common/temp/node_modules/.pnpm/mocha@11.1.0/node_modules/mocha/lib/runner.js:677:10)
```

pmconne: Yeah, those tests explicitly disable routing validation. Wanna try my suggestion? That way we don't have to alter the normal code path to accommodate tests.

tcobbs-bentley: Just to verify, you want me to put back the `expectDefined(token.iModelId)`, and update the test with an empty string as the iModelId?

pmconne: No. I pushed my suggestion.

core/frontend/src/BriefcaseConnection.ts
```diff
@@ -260,10 +260,10 @@ export class BriefcaseConnection extends IModelConnection {
   public override isBriefcaseConnection(): this is BriefcaseConnection { return true; }
 
   /** The Guid that identifies the iTwin that owns this iModel. */
-  public override get iTwinId(): GuidString { return super.iTwinId!; } // GuidString | undefined for IModelConnection, but required for BriefcaseConnection
+  public override get iTwinId(): GuidString { return super.iTwinId ?? "00000000-0000-0000-0000-000000000000"; } // GuidString | undefined for IModelConnection, but required for BriefcaseConnection
```
pmconne:
```suggestion
  public override get iTwinId(): GuidString { return super.iTwinId ?? Guid.empty; } // GuidString | undefined for IModelConnection, but required for BriefcaseConnection
```

core/frontend/src/BriefcaseConnection.ts
```diff
@@ -260,10 +260,10 @@ export class BriefcaseConnection extends IModelConnection {
...
 
   /** The Guid that identifies this iModel. */
-  public override get iModelId(): GuidString { return super.iModelId!; } // GuidString | undefined for IModelConnection, but required for BriefcaseConnection
+  public override get iModelId(): GuidString { return super.iModelId ?? "00000000-0000-0000-0000-000000000000"; } // GuidString | undefined for IModelConnection, but required for BriefcaseConnection
```
pmconne:
```suggestion
  public override get iModelId(): GuidString { return super.iModelId ?? Guid.empty; } // GuidString | undefined for IModelConnection, but required for BriefcaseConnection
```

core/frontend/src/BriefcaseConnection.ts
tcobbs-bentley: I pushed a commit with all 4 empty GUID values updated to use `Guid.empty`.

core/frontend/src/BriefcaseConnection.ts
tcobbs-bentley: I pushed a commit with all 4 empty GUID values updated to use `Guid.empty`.

core/frontend/src/IModelConnection.ts
```diff
@@ -1232,7 +1232,7 @@ export namespace IModelConnection {
       const views: ViewSpec[] = [];
       const viewProps: ViewDefinitionProps[] = await this.queryProps(queryParams);
       viewProps.forEach((viewProp) => {
-        views.push({ id: viewProp.id as string, name: viewProp.code.value!, class: viewProp.classFullName });
+        views.push({ id: viewProp.id as string, name: viewProp.code.value ?? "<Unnamed view>", class: viewProp.classFullName });
```
pmconne:
```suggestion
        views.push({ id: viewProp.id as string, name: viewProp.code.value ?? "", class: viewProp.classFullName });
```

Don't want non-localized string potentially showing up in UI nor do we want to change the name of an unnamed view.

core/frontend/src/common/internal/render/VertexTableBuilder.ts
```diff
@@ -72,15 +72,15 @@ export abstract class VertexTableBuilder {
 
   protected advance(nBytes: number) {
     this._curIndex += nBytes;
-    assert(this._curIndex <= this.data!.length);
+    assert(this._curIndex <= (this.data?.length ?? -1));
```
pmconne:
```suggestion
    assert(undefined !== this.data);
    assert(this._curIndex <= this.data.length);
```

pmconne (approved):

aruniverse (approved):

pmconne:
@tcobbs-bentley are we just awaiting reviews? @bbastings @dassaf4 @saeeedtorabi @calebmshafer @wgoehrig @ben-polinsky @ColinKerr @hl662 @rschili @christophermlawson @StefanApfel-Bentley

core/common/src/geometry/ElementGeometryFB.ts
```diff
@@ -5,6 +5,7 @@
 /** @packageDocumentation
  * @module Serialization
  */
+import { expectNotNull } from "@itwin/core-bentley";
```
bbastings: Since ElementGeometryFB.ts is generated output from the flatbuffers schema files, might be better to disable for this file instead of making lots of modifications?

tcobbs-bentley:
> @tcobbs-bentley are we just awaiting reviews? @bbastings @dassaf4 @saeeedtorabi @calebmshafer @wgoehrig @ben-polinsky @ColinKerr @hl662 @rschili @christophermlawson @StefanApfel-Bentley

Yes. The PR is set to auto-merge, and will do so as soon as the needed review(s) happen.

core/common/src/geometry/ElementGeometryFB.ts
tcobbs-bentley: I wasn't aware that this was a generated file. I can add non-null to the other disable lint rules already in the file and revert my other changes. Are we sure that these non-null assertions are actually safe?

pmconne: Yes, they're safe.

bbastings: @tcobbs-bentley Yes, I think we should trust the flatbuffers library to generate functional code.

There is this comment near the top...
  // automatically generated by the FlatBuffers compiler, do not modify

tcobbs-bentley: Sorry I missed that. Adding the eslint-disable comment is a modification, but obviously a very small one. I did that.

rschili (approved):

pmconne:
Electron backend tests flaked out. Force-merging. Thanks @tcobbs-bentley.
```
  7 passing (1m)
  1 failing

  1) ElectronHost tests.
       Should save main window size, position and maximized flag.:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/vsts/work/1/s/core/electron/lib/cjs/test/backend/RunElectronBackendTests.js)
      at listOnTimeout (node:internal/timers:588:17)
      at process.processTimers (node:internal/timers:523:7)
```

Issue #7764, opened 2025-02-24
When two object are changed by two users rebase does not work as intented.


khanaffan:
It seems when placement is changed conflict hander is never called and incoming change replaces local change.

In this case locally deleted element pop back up during rebase. 

**To Reproduce**
1. With no locks. This is not normal and related to branching.
2. Change location of graphical element on briefcase 1 and push
3. Delete the element on briefcase 2 and pull merge.
4. The deleted element will pop back up and it's also removed from the txn after rebasing.
5. It observes this only happen when placement is changed otherwise it does get deleted.

Here is a test and branch is set that have following test.
```ts
 it.only("delete instance (data conflict) ", async () => {
    const b1 = await ctx.openB1(true /* = noLock */);
    const b2 = await ctx.openB2(true /* = noLock */);

    const e1 = await insertPhysicalObject(b1);
    b1.saveChanges();
    await b1.pushChanges({ description: `inserted physical object [id=${e1}]` });
    await b2.pullChanges();

    const eb1 = b1.elements.getElementProps<PhysicalElementProps>(e1);
    eb1.userLabel = "test1";
    eb1.placement = { origin: { x: 1, y: 1, z: 1 }, angles: { yaw: 1, pitch: 1, roll: 1 } };
    b1.elements.updateElement(eb1);
    b1.saveChanges();
    await b1.pushChanges({ description: `update physical object [id=${e1}]` });

    b2.elements.deleteElement(e1);;
    b2.saveChanges();
    await b2.pullChanges();

    const eb2 = b2.elements.tryGetElementProps<PhysicalElementProps>(e1); // should throw
    assert.isUndefined(eb2);

    b1.close();
    b2.close();
  });
```
**Expected behavior**
The element should stay deleted.

reported by : @wgoehrig

RohitPtnkr1996:
We ask sqlite to ignore all no-ops when rebasing local changes.
A function in sqlite has an additional check in the delete op block for the ignoreNoop flag which ignores all conflicts when set to true (our case).
This causes the conflict handler to never get invoked (trying to delete a row which was updated by another briefcase) and the element ends up never being re-deleted (being treated as a no-op, which it isn't !)
On Affan's suggestion, I've sent an email to the sqlite support team. We will let them fix this and update sqlite once they release the fix.
Putting this work item on hold until them.

RohitPtnkr1996:
The next version of sqlite is scheduled for release by the end of October.
[...]

PR #8396, opened 2025-08-06
Tests rebase when conflicting updates are made by two briefcases

RohitPtnkr1996:
Fixes: https://github.com/iTwin/itwinjs-core/issues/7764


core/backend/src/test/standalone/ChangeMerge.test.ts
```diff
@@ -499,5 +501,34 @@ describe("Change merge method", () => {
     b1.close();
     b2.close();
   });
-});
+  it.only("delete instance (data conflict)", async () => {
```
pmconne:
```suggestion
  it("delete instance (data conflict)", async () => {
```

pmconne (approved):

PR #8408, opened 2025-08-08
Don't use spread operator with potentially large array

eringram:
This came up in PR discussion when adding this same code to implement the EXT_mesh_primitive_restart glTF extension in CesiumJS: https://github.com/CesiumGS/cesium/pull/12792#discussion_r2254550358. There is the possibility of `meshPrimitives` being very large, in which case the `...` spread operator can cause a stack overflow, so these changes just use `Array.concat()` if the array length is over 50000.

This change is based on the same observation in iTwin.js in another location, and the fix in this PR: https://github.com/iTwin/itwinjs-core/pull/7605.


pmconne:
Spread operators are full of gotchas...

> CesiumJS does this in a [function](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/addAllToArray.js) so it can be reused. I'm open to making a new function here as well if there is demand to use it elsewhere, but would like a suggestion for where to put a small utility function like this in core-frontend.

If you decide to add a utility function it almost certainly belongs in core-bentley, not core-frontend.

markschlosseratbentley:
Just a general comment, not meant to hold up this PR, but we should eventually do an audit of the iTwin.js codebase for other places where spread operators might encounter potentially large arrays.

EDIT: that might be a good time to add a utility function.

PR #8409, opened 2025-08-08
DRAFT: Clean up lint warnings in core/ec*

tcobbs-bentley:
__NOTE:__ This was created after a `git stash pop` on master, with a local build that included changes from another PR (notably the creation of `expectDefined` and `expectNotNull`). Until that PR is merged, and then origin/master is merged into this PR, it won't work at all. Once that happens, I will remove this message and promote this PR out of draft.

Clean up the lint warnings in all the core/ec* directories.

1. Where possible, clean up code to not require non-null assertions (`!`).
2. Use the new `expectDefined` and `expectNotNull` functions when there isn't an obviously correct way to get rid of the `!`. As time passes, code using `expectDefined` and `expectNotNull` will be updated to stop using it.
3. Fix code that triggers the `@typescript-eslint/restrict-template-expressions` warning.

### DIFF
--- a/core/ecschema-editing/src/Editing/RelationshipClasses.ts
+++ b/core/ecschema-editing/src/Editing/RelationshipClasses.ts
@@ -6,6 +6,7 @@
  * @module Editing
  */
 
+import { expectDefined } from "@itwin/core-bentley";
 import {
   CustomAttribute, DelayedPromiseWithProps, ECClass, ECClassModifier, EntityClass, LazyLoadedRelationshipConstraintClass, Mixin, NavigationPropertyProps,
   RelationshipClass, RelationshipClassProps, RelationshipConstraint, RelationshipEnd, RelationshipMultiplicity, SchemaItemKey, SchemaItemType,
@@ -102,7 +103,7 @@ export class RelationshipClasses extends ECClasses {
 
       return newClass.key;
     } catch (e: any) {
-      throw new SchemaEditingError(ECEditingStatus.CreateSchemaItemFromProps, new ClassId(this.schemaItemType, relationshipProps.name!, schemaKey), e);
+      throw new SchemaEditingError(ECEditingStatus.CreateSchemaItemFromProps, new ClassId(this.schemaItemType, relationshipProps.name ?? "<unknown>", schemaKey), e);
     }
   }
 
@@ -121,9 +122,9 @@ export class RelationshipClasses extends ECClasses {
     await super.setBaseClass(itemKey, baseClassKey);
 
     try {
-      await this.validate(relClass!);
+      await this.validate(expectDefined(relClass));
     } catch(e: any) {
-      await (relClass! as ECClass as MutableClass).setBaseClass(baseClass);
+      await (expectDefined(relClass) as ECClass as MutableClass).setBaseClass(baseClass);
       throw new SchemaEditingError(ECEditingStatus.SetBaseClass, new ClassId(SchemaItemType.RelationshipClass, itemKey), e);
     }
   }
--- a/core/ecschema-editing/src/Merging/EnumerationMerger.ts
+++ b/core/ecschema-editing/src/Merging/EnumerationMerger.ts
@@ -39,7 +39,7 @@ export async function addEnumeration(context: SchemaMergeContext, change: Enumer
 export async function modifyEnumeration(context: SchemaMergeContext, change: EnumerationDifference, itemKey: SchemaItemKey) {
   const enumeration = await context.targetSchema.lookupItem(itemKey) as MutableEnumeration;
   if(change.difference.type !== undefined) {
-    throw new Error(`The Enumeration ${itemKey.name} has an incompatible type. It must be "${primitiveTypeToString(enumeration.type!)}", not "${change.difference.type}".`);
+    throw new Error(`The Enumeration ${itemKey.name} has an incompatible type. It must be "${enumeration.type ? primitiveTypeToString(enumeration.type) : "<unknown>"}", not "${change.difference.type}".`);
   }
   if(change.difference.label !== undefined) {
     await context.editor.enumerations.setDisplayLabel(itemKey, change.difference.label);
--- a/core/ecschema-editing/src/Validation/DiagnosticReporter.ts
+++ b/core/ecschema-editing/src/Validation/DiagnosticReporter.ts
@@ -70,6 +70,7 @@ export abstract class SuppressionDiagnosticReporter implements IDiagnosticReport
   public report(diagnostic: AnyDiagnostic) {
     if (this._suppressions && this._suppressions.has(diagnostic.schema.fullName)) {
       const suppressedCodes = this._suppressions.get(diagnostic.schema.fullName);
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       if (suppressedCodes!.includes(diagnostic.code))
         return;
     }
--- a/core/ecschema-editing/src/Validation/ECRules.ts
+++ b/core/ecschema-editing/src/Validation/ECRules.ts
@@ -6,6 +6,7 @@
  * @module Validation
  */
 
+import { expectDefined } from "@itwin/core-bentley";
 import {
   AnyClass, AnyProperty, CustomAttribute, CustomAttributeContainerProps, ECClass, ECClassModifier,
   ECStringConstants, EntityClass, Enumeration, Mixin, PrimitiveProperty, PrimitiveType, primitiveTypeToString,
@@ -328,7 +329,7 @@ export async function* incompatibleValueTypePropertyOverride(property: AnyProper
     if (!baseType || primitiveType === baseType)
       return;
 
-    return new Diagnostics.IncompatibleValueTypePropertyOverride(property, [property.class.fullName, property.name, baseClass.fullName, primitiveTypeToString(baseType), primitiveTypeToString(primitiveType!)]);
+    return new Diagnostics.IncompatibleValueTypePropertyOverride(property, [property.class.fullName, property.name, baseClass.fullName, primitiveTypeToString(baseType), primitiveTypeToString(expectDefined(primitiveType))]);
   }
 
   for await (const baseClass of property.class.getAllBaseClasses()) {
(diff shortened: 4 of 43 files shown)


core/ecschema-editing/src/Editing/RelationshipClasses.ts
```diff
@@ -121,9 +122,9 @@ export class RelationshipClasses extends ECClasses {
     await super.setBaseClass(itemKey, baseClassKey);
 
     try {
-      await this.validate(relClass!);
+      await this.validate(expectDefined(relClass));
     } catch(e: any) {
-      await (relClass! as ECClass as MutableClass).setBaseClass(baseClass);
+      await (expectDefined(relClass) as ECClass as MutableClass).setBaseClass(baseClass);
```
ColinKerr: @christophermlawson what is the expected behavior when the item key doesn't resolve to a relationship class?

This should be fixed properly rather than with this hack and the behavior should be documented and tested

core/ecschema-editing/src/Merging/EnumerationMerger.ts
```diff
@@ -39,7 +39,7 @@ export async function addEnumeration(context: SchemaMergeContext, change: Enumer
 export async function modifyEnumeration(context: SchemaMergeContext, change: EnumerationDifference, itemKey: SchemaItemKey) {
   const enumeration = await context.targetSchema.lookupItem(itemKey) as MutableEnumeration;
   if(change.difference.type !== undefined) {
-    throw new Error(`The Enumeration ${itemKey.name} has an incompatible type. It must be "${primitiveTypeToString(enumeration.type!)}", not "${change.difference.type}".`);
+    throw new Error(`The Enumeration ${itemKey.name} has an incompatible type. It must be "${enumeration.type ? primitiveTypeToString(enumeration.type) : "<unknown>"}", not "${change.difference.type}".`);
```
ColinKerr, line 42: Why is the underlying Enumeration._type property possibly undefined?  Any valid enumeration MUST have a type set so we should not force callers to test if the type is null.

core/ecschema-editing/src/Validation/DiagnosticReporter.ts
```diff
@@ -70,6 +70,7 @@ export abstract class SuppressionDiagnosticReporter implements IDiagnosticReport
   public report(diagnostic: AnyDiagnostic) {
     if (this._suppressions && this._suppressions.has(diagnostic.schema.fullName)) {
       const suppressedCodes = this._suppressions.get(diagnostic.schema.fullName);
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       if (suppressedCodes!.includes(diagnostic.code))
```
ColinKerr: @christophermlawson fix this properly

core/ecschema-editing/src/Validation/ECRules.ts
```diff
@@ -328,7 +329,7 @@ export async function* incompatibleValueTypePropertyOverride(property: AnyProper
     if (!baseType || primitiveType === baseType)
       return;
 
-    return new Diagnostics.IncompatibleValueTypePropertyOverride(property, [property.class.fullName, property.name, baseClass.fullName, primitiveTypeToString(baseType), primitiveTypeToString(primitiveType!)]);
+    return new Diagnostics.IncompatibleValueTypePropertyOverride(property, [property.class.fullName, property.name, baseClass.fullName, primitiveTypeToString(baseType), primitiveTypeToString(expectDefined(primitiveType))]);
```
ColinKerr: This code should be rewritten so the call back isn't reaching out of its scope to get primitiveType.  It would be clearer if primitiveType was passed in as the `overrideType` of something like that indicating that this the type of the property that is trying to override a property from the base class.  It would also avoid the need for expectDefined here.

core/ecschema-editing/src/Validation/DiagnosticReporter.ts
pmconne:
```suggestion
    const suppressedCodes = this._suppressions?.get(diagnostic.schema.fullName);
    if (suppressedCodes && suppressedCodes.includes(diagnostic.code))
```

Gets rid of the need to assert non-null and fixes inefficiently looking up the key twice.

core/ecschema-locaters/src/StubSchemaXmlFileLocater.ts
```diff
@@ -84,7 +84,8 @@ export class StubSchemaXmlFileLocater extends SchemaFileLocater implements ISche
       return undefined;
 
     const maxCandidate = candidates.sort(this.compareSchemaKeyByVersion)[candidates.length - 1];
-    const alias = this.getSchemaAlias(maxCandidate.schemaText!);
+    // Note: if maxCandidate.schemaText is undefined, the "" passed to getSchemaAlias will throw an ECSchemaError.
+    const alias = this.getSchemaAlias(maxCandidate.schemaText ?? "");
```
ColinKerr, line 88: This is problematic.  findEligibleSchemaKeys always fills out the schemaText.  Some code uses the schemaText and other code loads it all over again.  It looks like we may load the text from disk more than once and in some cases we may hold onto the text long after we need it (when we construct the schema with a SchemaFileKey)

This should be rationalized and fixed

core/ecschema-metadata/src/Deserialization/Helper.ts
```diff
@@ -142,6 +143,15 @@ export class SchemaReadHelper<T = unknown> {
     return schemaItem !== undefined;
   }
 
+  /**
+   * Same as isSchemaItemLoaded, but also asserts that the schema item is defined using a type predicate.
+   * @note isSchemaItemLoaded already checks if the item is defined, but updating it to be a type predicate
+   * would break existing code that uses it.
+   */
+  protected isSchemaItemLoadedAndDefined(schemaItem: SchemaItem | undefined): schemaItem is SchemaItem {
```
ColinKerr: why is it breaking to make it a type predicate?

core/ecschema-metadata/src/Deserialization/Helper.ts
```diff
@@ -252,13 +262,16 @@ export class SchemaReadHelper<T = unknown> {
   private loadSchemaReferenceSync(ref: SchemaReferenceProps): void {
     const schemaKey = new SchemaKey(ref.name, ECVersion.fromString(ref.version));
     const refSchema = this._context.getSchemaSync(schemaKey, SchemaMatchType.LatestWriteCompatible);
+    if (undefined === this._schema) {
```
ColinKerr: Maybe it would be cleaner to rewrite this method to take the schema in as a parameter.  As called _schema will always be defined but changing the signature makes the method's contract clear.

core/ecschema-editing/src/Validation/DiagnosticReporter.ts
tcobbs-bentley: I'm not sure if it's what you meant, but I cleaned it up to not use `map.has(key)` followed by `map.get(key)!`. Instead it just performs the get and checks that the result is defined.

core/ecschema-metadata/src/Deserialization/Helper.ts
tcobbs-bentley: If `isSchemaItemLoaded` were made a type predicate, `isSchemaItemLoaded(item) && doSomething(item!)` would generate a compile error, because the `!` after `item` is then not allowed. Without it being a predicated, the `!` is required.

core/ecschema-metadata/src/Deserialization/Helper.ts
tcobbs-bentley: I made that simple change. While doing so I noticed that it is being cast to `MutableSchema` without a guard. Should a new guard be added that verifies that it is an instance of `MutableSchema` and throw an exception if not?

core/ecschema-metadata/src/Deserialization/Helper.ts
ColinKerr: That is a lint error not an actual compilation error.  It seems really silly to have a method that ONLY checks if something is not null be wrapped by another method that only checks if the same thing is not null to avoid a possible lint error on a protected method.

The chances that someone uses this protected method are near zero and if they do it won't break them at runtime.

tcobbs-bentley: I'm happy to get rid of the wrapper and update the original to be a type predicate. I did it this way because I thought that wasn't allowed.

tcobbs-bentley: I see that SchemaReadHelper is internal, so changing the API was never a problem. I'll get rid of the new function and update the old one to be a type predicate.

tcobbs-bentley: Done.

core/ecschema-editing/src/Editing/RelationshipClasses.ts
christophermlawson: @ColinKerr  This would be unexpected, so I think an exception would be appropriate here.

core/ecschema-editing/src/Merging/EnumerationMerger.ts
christophermlawson: @ColinKerr The constructor also has this as optional. It's internal so we can change this. Should we require the parameter or have a default (int or string)?

core/ecschema-locaters/src/StubSchemaXmlFileLocater.ts
christophermlawson: @ColinKerr I'm assuming you are referring to the other locaters (json and xml)?  Looks like we may read it twice. The StubSchemaXmlFileLocater looks ok to me.

core/ecschema-editing/src/Validation/ECRules.ts
tcobbs-bentley: I discovered that the requirement for `!` or `expectDefined` was caused by what I feel is a compiler bug. (The compiler for some reason thinks that `primitiveType` might be undefined in `callback`, even though it's a `const` that has already been verified to be defined.) Introducing another `const` removed the need for either `!` or `expectDefined`, so I made that change. Note that this doesn't address your actual comment here, but it does provide what I feel is a better fix to get rid of the `!`, and your comment no longer seems to be in the scope of this PR.

PR #8413, opened 2025-08-11
Improve frame rate during interactive editing

### LINKED ISSUE #4404: TransformGraphicsProvider draws "corrupted" geometry
**Describe the bug**
For some elements displayed dynamics graphics for transform tools look corrupted/inaccurate.

**To Reproduce**

1. Open ui-test-app with write enabled.
2. Open attached bim as standalone local model.
3. Enter *Default - View 2*.
4. Open Visibility Tree and enable *Temporary construction objects* model and category
5. Use *MoveElementsTool* on elements shown in screenshots

**Expected behavior**
Dynamics graphics should look like source element.

**Screenshots**
image
image

- Windows 10
- Electron 11.1.0
- 2.19.44  (also reproduceable in 3.4.0-dev.58)

**Attachments**
[5bim.zip](https://github.com/iTwin/itwinjs-core/files/9684355/5bim.zip)

pmconne:
#4485 made `requestElementGraphics` translate the graphics relative to the center of the element's bounding box. This fixed #4404 wherein decoration graphics were being displayed with absolute coordinates thousands of kilometers from the origin, resulting in floating point precision artifacts.

However, this negatively affects the temporary element graphics drawn by `DynamicIModelTile` to represent elements inserted or modified during a `GraphicalEditingScope`. Because of the translation, each element's graphics had to be placed into its own `GraphicBranch` to be pushed and popped during rendering. Pushing a branch invalidates a ton of uniform state on the GPU that must then be recomputed and re-uploaded.

These element graphics are already translated relative to the center of the model's bounding box, so floating point precision issues are not a concern. So the fix is to specify `useAbsolutePosition: true` when requesting the graphics. Then the element graphics get an identity transform, and don't need to be placed inside their own branches. (They all still end up together inside a single branch that applies the model's transform to them).

Prior to this change, moving 248 elements by half a meter in Z would cause FPS to drop from 43.5 to 30.0. After this change, it drops to 42.5. This should enable applications to keep an editing scope active for much longer without worrying about significant performance impact.

### DIFF
--- a/common/changes/@itwin/core-frontend/pmc-element-graphics_2025-08-11-18-17.json
+++ b/common/changes/@itwin/core-frontend/pmc-element-graphics_2025-08-11-18-17.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Improve frame rate during interactive editing.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/core/frontend/src/internal/tile/DynamicIModelTile.ts
+++ b/core/frontend/src/internal/tile/DynamicIModelTile.ts
@@ -366,6 +366,7 @@ class GraphicsTile extends Tile {
       smoothPolyfaceEdges: this.tree.edgeOptions && this.tree.edgeOptions.smooth,
       clipToProjectExtents: this.tree.is3d,
       sectionCut: this.tree.stringifiedSectionClip,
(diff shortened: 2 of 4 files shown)


aruniverse (approved):

pmconne:
Force-merging because flakery:
- Linux failed with no error message in "Use Node 22.x"
- Windows has been trying to pull the repo for 55 minutes and will shortly time out

MacOS and integration tests succeeded.

aruniverse:

pmconne:

PR #8414, opened 2025-08-11
Add clean method to ECDb for removing unused schema references and tests

BinhL3:


core/backend/src/test/ecdb/ECDb.test.ts
```diff
@@ -2,515 +2,78 @@
...
+import { IModelJsFs } from "../../IModelJsFs";
 import { KnownTestLocations } from "../KnownTestLocations";
 import { ECDbTestHelper } from "./ECDbTestHelper";
-import { QueryOptionsBuilder } from "@itwin/core-common";
-import { EntityClass, SchemaContext, SchemaJsonLocater, SchemaKey } from "@itwin/ecschema-metadata";
 
 describe("ECDb", () => {
   const outDir = KnownTestLocations.outputDir;
 
-  it("should be able to create a new ECDb", () => {
```
ben-polinsky: I see you added a commit to remove these tests to "streamline the test suite" - what's the reasoning behind that decision? Do we have tests elsewhere covering the same code?

BinhL3: I've readded those test cases, I had mistakenly overriden them prior!

ben-polinsky: Phew, thanks.

full-stack-tests/backend/src/integration/DropSchemas.test.ts
```diff
@@ -63,7 +63,7 @@ describe("Drop schemas", function (this: Suite) {
 
     await b1.pushChanges({ description: "pushed TestSchema1" });
 
-    b1[_nativeDb].dropSchema("TestSchema1");
+    b1[_nativeDb].dropSchemas(["TestSchema1"]);
```
ColinKerr: You should call through your new alpha method.  Consider that you did not expose a proper implementation in iModelDb that calls the correct dropSchemas method, you only wrote a method to expose the ECDb one

core/backend/src/IModelDb.ts
```diff
@@ -1686,6 +1686,19 @@ export abstract class IModelDb extends IModel {
     };
   }
 
+  /** Drops schemas from an array of schema
+  * @param schemaNames Array of schema names to drop
+  * @alpha
+  */
+  public dropSchemas(schemaNames: string[]): void {
```
khanaffan: Drop schema is similar to import schema. It need exclusive lock.

khanaffan (changes requested):

core/backend/src/IModelDb.ts
```diff
@@ -853,6 +853,22 @@ export abstract class IModelDb extends IModel {
     return this[_nativeDb].restartTxnSession();
   }
 
+  /** Drops schemas from an array of schema
+* @param schemaNames Array of schema names to drop
+* @alpha
+*/
+  public async dropSchemas(schemaNames: string[]): Promise<void> {
+    if (this[_nativeDb].getITwinId() !== Guid.empty) {
```
khanaffan: Ensure that there is no currently unsaved changes on this connection. If there is throw an exception to user with msg that dropSchema() cannot be called with unsaved changes. User must save or abandon it before calling dropSchema()

core/backend/src/IModelDb.ts
```diff
@@ -853,6 +853,22 @@ export abstract class IModelDb extends IModel {
...
+      await this.acquireSchemaLock();
+    }
+    try {
+      await this[_nativeDb].dropSchemas(schemaNames);
```
khanaffan: call saveChanges() with message e..g `dropped unused schemas`

core/backend/src/IModelDb.ts
```diff
@@ -853,6 +853,22 @@ export abstract class IModelDb extends IModel {
...
+    } catch (error) {
+      Logger.logError(loggerCategory, `Failed to drop schemas: ${error}`);
```
khanaffan, line 877: call abandonChanges() here do not leave connection in a inconsistent state. What if user call saveChanges() after this point.

core/backend/src/IModelDb.ts
khanaffan: Also throw exception if schema sync is on with error that drop schema is not supported with schema sync.

khanaffan (approved):

core/backend/src/IModelDb.ts
```diff
@@ -853,6 +853,31 @@ export abstract class IModelDb extends IModel {
...
+*/
+  public async dropSchemas(schemaNames: string[]): Promise<void> {
+    if (schemaNames.length === 0)
+      return;
+    if (this[_nativeDb].schemaSyncEnabled())
+      throw new IModelError(DbResult.BE_SQLITE_ERROR, "Cannot drop schemas when schema sync is enabled");
+    if (this[_nativeDb].hasUnsavedChanges())
+      throw new IModelError(ChangeSetStatus.HasUncommittedChanges, "Cannot drop schemas with unsaved changes");
+    if (this[_nativeDb].getITwinId() !== Guid.empty)
+      await this.acquireSchemaLock();
```
aruniverse, line 871: shouldnt you release the schema lock upon schema drop?

BinhL3: added release lock

core/backend/src/ECDb.ts
```diff
@@ -166,6 +166,26 @@ export class ECDb implements Disposable {
     }
   }
 
+  /** Drops schemas of array of schema
```
tcobbs-bentley: I don't understand this description.

core/backend/src/IModelDb.ts
```diff
@@ -853,6 +853,32 @@ export abstract class IModelDb extends IModel {
     return this[_nativeDb].restartTxnSession();
   }
 
+  /** Drops schemas from an array of schema
```
tcobbs-bentley: This description seems off also, although I think it's better than the other one.

core/backend/src/IModelDb.ts
```diff
@@ -853,6 +853,32 @@ export abstract class IModelDb extends IModel {
...
+
+    try {
+      this[_nativeDb].dropSchemas(schemaNames);
+      this.saveChanges(`dropped unused schemas`);
+    } catch (error: any) {
+      Logger.logError(loggerCategory, `Failed to drop schemas: ${error}`);
+      this.abandonChanges();
+      throw new IModelError(DbResult.BE_SQLITE_ERROR, `Failed to drop schemas: ${error}`);
+    }
+    await this.locks.releaseAllLocks();
```
tcobbs-bentley: Won't this get missed if an error happens? Is that intentional? Or should this be in a finally block?

PR #8417, opened 2025-08-12
initial fix

### LINKED ISSUE #8404: Desktop iModel crs extent latitude is always `0.0`
When working with an iTwin Viewer on the web versus an iTwin Viewer on desktop(electron in this case), there is a discrepancy on the [`iModelConnection.geographicCoordinateSystem.horizontalCRS.extent`](https://www.itwinjs.org/reference/core-common/geometry/horizontalcrsextentprops/). This prop has `northEast` and `southWest` props. Both are of the type [Carto2DDegrees](https://www.itwinjs.org/reference/core-common/geometry/carto2ddegrees). This type has `latitude` and `longitude` values which are both numbers. This is where the discrepancy exists. 

When opening the same iModel from both the viewers. The web viewer will display non-zero numerical values for `latitude` but on the desktop viewer, the `latitude` value for both northEast and `southWest` props are always `0`. <- this is the bug on desktop side.

Image

In above image, the console log on the left is from a electron iTwin Viewer App and the console on the right is from a web iTwin Viewer app. The log on the left has latitude values as 0 while the one on the right has non-zero values which are correct. 

This is not specific to any one singular iModel and is occurring with various different iModels that have various different CRS being used.

**To Reproduce**

I have created a minimal repo using iTwin Viewer templates to demonstrate and reproduce this issue.
1. git clone https://github.com/anmolshres98/gcs-extents-issue-repro
2. cd into `desktop-viewer` then `pnpm install`
3. cd into `web-viewer` then `pnpm install`
4. `pnpm start` in each directory and observe the console logs to see the difference
5. **_DO NOT_** run both `pnpm start` scripts at the same time since they will conflict

- OS: MacOS 15.5 (24F74)
- Browser: Chrome and Electron
- iTwin.js Version: 5.0.5

**Additional context**
I was able to trace the logic of why latitude is being returned as `0`. When the latitude and longitude values are being set [here](https://github.com/iTwin/itwinjs-core/blob/d7a2a75db7b34a8c0bb3067c1124ed2b7a1050f5/core/common/src/geometry/Projection.ts#L433), the `data` object is shaped as `{
    "_latitude": 37,
    "longitude": -88.5
}`. So when `data.latitude` is access, it is undefined and `latitude` on the object remains at `0.0.`. This is the case with the electron viewer.

However, on the web viewer the `data` object at the exact same code location looks like this `{
    "latitude": 37,
    "longitude": -88.5
}`. In this case `data.latitude` is defined and is accessed and set to the object correctly.

This difference is the entire reason as to why on the electron viewer latitude values are always 0.0 erroneously.

Furthermore, I was able to trace the inception of the `data` object itself up till this call [here](https://github.com/iTwin/itwinjs-core/blob/d7a2a75db7b34a8c0bb3067c1124ed2b7a1050f5/core/frontend/src/BriefcaseConnection.ts#L279). When this ipc call is returned in the electron app, the crs extents value is formatted as `{_latitude:..., longitude:...}` rather than the correct/expected shape `{latitude:..., longitude:...}`

ben-polinsky:
We must convert CRS extents to JSON before sending via electron, using object spread, or any other operations which enumerate properties of the extent object. This is due to how class accessors are set. 

While I would prefer the fix to go on the `Carto2DDegrees` class, I don't see a way to do it in a non-breaking fashion.

Fixes #8404.

### DIFF
--- a/full-stack-tests/core/src/frontend/hub/IModelConnection.test.ts
+++ b/full-stack-tests/core/src/frontend/hub/IModelConnection.test.ts
@@ -23,7 +23,7 @@ async function executeQuery(iModel: IModelConnection, ecsql: string, bindings?: 
   return rows;
 }
 
-describe("IModelConnection (#integration)", () => {
+describe.only("IModelConnection (#integration)", () => {
   let iModel: IModelConnection;
 
   before(async () => {
@@ -214,4 +214,11 @@ describe("IModelConnection (#integration)", () => {
     const formattedValue = spec.applyFormatting(5.0);
     assert.equal(formattedValue, "5.0 m");
   });
+
+  it("properly deserializes gcs latitude", async () => {
+      const iTwinId = await TestUtility.getTestITwinId();
+      const iModelId = await TestUtility.queryIModelIdByName(iTwinId, TestUtility.testIModelNames.smallTex);
+      iModel = await CheckpointConnection.openRemote(iTwinId, iModelId);
+      assert.notEqual(iModel.geographicCoordinateSystem?.horizontalCRS?.extent?.northEast.latitude, 0);
+    })
 });
--- a/core/common/src/IModel.ts
+++ b/core/common/src/IModel.ts
@@ -522,8 +522,8 @@ export abstract class IModel implements IModelProps {
       projectExtents: this.projectExtents.toJSON(),
       globalOrigin: this.globalOrigin.toJSON(),
       ecefLocation: this.ecefLocation,
-      geographicCoordinateSystem: this.geographicCoordinateSystem,
-      ... this._getRpcProps(),
+      geographicCoordinateSystem: this.geographicCoordinateSystem?.toJSON(),
+      ...this._getRpcProps(),
     };
   }
 


full-stack-tests/core/src/frontend/hub/IModelConnection.test.ts
```diff
@@ -23,7 +23,7 @@ async function executeQuery(iModel: IModelConnection, ecsql: string, bindings?:
   return rows;
 }
 
-describe("IModelConnection (#integration)", () => {
+describe.only("IModelConnection (#integration)", () => {
```
hl662:
```suggestion
describe("IModelConnection (#integration)", () => {
```

ben-polinsky: it's on the way, thanks

full-stack-tests/core/src/frontend/hub/IModelConnection.test.ts
```diff
@@ -214,4 +214,11 @@ describe("IModelConnection (#integration)", () => {
     const formattedValue = spec.applyFormatting(5.0);
     assert.equal(formattedValue, "5.0 m");
   });
+
+  it("properly deserializes gcs latitude", async () => {
+      const iTwinId = await TestUtility.getTestITwinId();
+      const iModelId = await TestUtility.queryIModelIdByName(iTwinId, TestUtility.testIModelNames.smallTex);
+      iModel = await CheckpointConnection.openRemote(iTwinId, iModelId);
+      assert.notEqual(iModel.geographicCoordinateSystem?.horizontalCRS?.extent?.northEast.latitude, 0);
```
ben-polinsky, line 222: note: without the change, this failed on electron and passed on chrome.

aruniverse:

PR #8418, opened 2025-08-12
TextAnnotation Ordered and Unordered Lists

### LINKED ISSUE #7679: TextAnnotations - ListRun

I would like to be able to insert lists into a `TextBlock`. These could be either ordered, or unordered. They could also be nested.

A new type of run should be created called `ListRun`
* It should include a property called something like `isOrdered` to specify whether or not to make an ordered list (1., 2., 3., ...) or unordered list (•, •, •, ...)
* It should be similar to a `ParagraphRun` and be able to contain child runs. For nesting purposes, the child runs could include more `ListRuns`
* `StyleOverrides` should include a property to describe either the type of bullet or numbering system
* I'm thinking that the bullets could come from the active font? There might need to be a fallback option if the font doesn't contain any of those special characters.
* `StyleOverrides` should include a property to describe the horizontal whitespace offset of items in the list

claudiaareneee:
## Description

This PR introduces comprehensive support for ordered and unordered lists in text annotations, along with a major refactor of the text model and layout logic. Key changes include:

### Text Model Refactor
- **New Component Structure:** The text model now uses a nested `children` structure for `TextBlock`, `Paragraph`, and `List` components, replacing the previous flat arrays.
- **List Support:** Added `List`, `OrderedListMarker`, and `UnorderedListMarker` types, enabling both ordered and unordered lists with customizable markers.
- **Component Traversal:** Introduced `traverseTextBlockComponent` for depth-first iteration over all text components.

### API and Serialization
- **Updated Methods:** `appendParagraph`, `appendRun`, and new list-related methods now work with the nested component model.
- **JSON Serialization:** All components support updated to/from JSON methods reflecting the new structure and enums.

### TextStyle and Layout
- **New Style Properties:** Added `indentation`, `paragraphSpacingFactor`, and `listMarker` to `TextStyleSettingsProps`.
- **Style Resolution:** Refactored `TextStyleResolver` to handle, indentation and marker text, as well as simplified the resolution logic.
- **Layout Logic:** `TextBlockLayout`, `LineLayout`, and related classes now support nested components, lists, and correct marker/indentation placement.

### Geometry and Visualization
- **List Marker Geometry:** Geometry generation now includes list markers, with correct offsets and appearance.
- **Debug Visualization:** Improved debug rendering for lines, runs, and markers.

### Editor and Keyins
- **TextDecorationTool:** Added keyins for creating lists, list items, setting indentation, and importing/exporting JSON.
- **Editor Logic:** Refactored run/paragraph/list appending logic to support the new model.

### Tests and Documentation
- **Test Refactor:** Updated and expanded tests to cover lists, indentation, paragraph spacing, and stringification.
- **API Docs:** Updated API documentation and type definitions for new structure.

## Link to PR

* Closes #7679

## TODO

- [ ] Cleanup the new keyins in `TextDecoration`
- [ ] Tests
- [ ] Any leftover inline todos
- [ ] Documentation files
- [ ] Change files
- [ ] make sure `paragraphSpacingFactor` only applies to paragraphs within lists and not the lists themselves

### Class Diagram:
image

### Data model:
image

## Example

image

<details>
  <summary>
    See `textBlockProps`:
  </summary>
  
  ```json
    {
        "styleOverrides": {
            "lineHeight": 1,
            "color": 45643
        },
        "type": "textBlock",
        "children": [
            {
                "styleOverrides": {},
                "type": "paragraph",
                "children": [
                    {
                        "styleOverrides": {
                            "fontName": "Arial"
                        },
                        "type": "text",
                        "content": "Example:",
                        "baselineShift": "none"
                    }
                ]
            },
            {
                "styleOverrides": {
                    "fontName": "Arial",
                    "color": 50390,
                    "indentation": 4
                },
                "type": "list",
                "children": [
                    {
                        "styleOverrides": {},
                        "type": "paragraph",
                        "children": [
                            {
                                "styleOverrides": {
                                    "fontName": "Arial"
                                },
                                "type": "text",
                                "content": "Apples",
                                "baselineShift": "none"
                            }
                        ]
                    },
                    {
                        "styleOverrides": {},
                        "type": "paragraph",
                        "children": [
                            {
                                "styleOverrides": {
                                    "fontName": "Arial"
                                },
                                "type": "text",
                                "content": "Bananas",
                                "baselineShift": "none"
                            }
                        ]
                    },
                    {
                        "styleOverrides": {},
                        "type": "paragraph",
                        "children": [
                            {
                                "styleOverrides": {
                                    "fontName": "Arial"
                                },
                                "type": "text",
                                "content": "Oranges",
                                "baselineShift": "none"
                            }
                        ]
                    },
                    {
                        "styleOverrides": {},
                        "type": "paragraph",
                        "children": [
                            {
                                "styleOverrides": {
                                    "fontName": "Arial"
                                },
                                "type": "text",
                                "content": "Grapes",
                                "baselineShift": "none"
                            },
                            {
                                "styleOverrides": {
                                    "fontName": "Arial",
                                    "listMarker": "A.",
                                    "color": 255
                                },
                                "type": "list",
                                "children": [
                                    {
                                        "styleOverrides": {},
                                        "type": "paragraph",
                                        "children": [
                                            {
                                                "styleOverrides": {
                                                    "fontName": "ENGINEERING",
                                                    "isBold": true,
                                                    "color": 255255
                                                },
                                                "type": "text",
                                                "content": "Green",
                                                "baselineShift": "none"
                                            }
                                        ]
                                    },
                                    {
                                        "styleOverrides": {},
                                        "type": "paragraph",
                                        "children": [
                                            {
                                                "styleOverrides": {
                                                    "fontName": "ENGINEERING"
                                                },
                                                "type": "text",
                                                "content": "Red",
                                                "baselineShift": "none"
                                            }
                                        ]
                                    }
                                ]
                            }
                        ]
                    }
                ]
            }
        ],
        "styleId": "",
        "width": 20,
        "justification": "left",
        "margins": {
            "left": 1,
            "right": 1,
            "top": 1,
            "bottom": 1
        }
    }
    ```
</details>


core/backend/src/annotations/ElementDrivesTextAnnotation.ts
```diff
@@ -49,6 +49,31 @@ export function isITextAnnotation(element: Element): element is ITextAnnotation
...
+ *
+ * Traverses the provided component and its children, returning an array of all [FieldRun]($common) instances found.
+ * Useful for extracting all field runs from nested text structures such as [Paragraph]($common)s and [List]($common)s.
+ *
+ * @param component The root [TextBlockComponent]($common) to search for [FieldRun]($common)s.
+ * @param runs (optional) An array to accumulate results; used for recursion.
+ * @returns An array of [FieldRun]($common) instances found within the component tree.
+ * @beta
+ */
+export function collectFieldRuns(component: TextBlockComponent, runs: FieldRun[] = []): FieldRun[] {
```
pmconne: Instead of a dedicated function for this specific purpose requiring fully populating an array, consider adding a generic generator function that iterates over `{ parent, child }` recursively for a given root.

core/common/src/annotation/TextBlock.ts
```diff
@@ -19,6 +19,28 @@ export interface ClearTextStyleOptions {
   preserveChildrenOverrides?: boolean;
 }
 
+/** The different types of [[Run]]s.
+ * @beta
+ */
+export enum RunComponentType {
```
pmconne: What motivated the introduction of this enum vs using string literals?

pmconne:
I don't have any major quibbles with the data model and class diagram. I do majorly disagree with all the new type-switching required to determine what kind of component I'm dealing with and the exposure of all the implementation details of the class hierarchy. IMO the public API should expose simple, flat interfaces described by the data model diagram, and implement those interfaces internally using the class diagram. You can use the [_implementationProhibited](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/guidelines/release-tags-guidelines.md#choosing-appropriate-release-tags) pattern to make those interfaces extensible. Hiding the class hierarchy will prevent people from trying dumb stuff like creating their own subclasses of `TextBlockComponent`.

core/common/src/annotation/TextBlock.ts
claudiaareneee: There was a reason, but I have forgotten it. I can attempt to switch this back to string literals. The reason might reveal itself again 😅

core/backend/src/annotations/ElementDrivesTextAnnotation.ts
claudiaareneee: I like that and will look into it

claudiaareneee:
> I don't have any major quibbles with the data model and class diagram. I do majorly disagree with all the new type-switching required to determine what kind of component I'm dealing with and the exposure of all the implementation details of the class hierarchy. IMO the public API should expose simple, flat interfaces described by the data model diagram, and implement those interfaces internally using the class diagram. You can use the [_implementationProhibited](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/guidelines/release-tags-guidelines.md#choosing-appropriate-release-tags) pattern to make those interfaces extensible. Hiding the class hierarchy will prevent people from trying dumb stuff like creating their own subclasses of `TextBlockComponent`.

On that note, I think I'm going to clean up the append methods like a lot. I don't think they're simple or intuitive enough. 

I'll look at the `_implementationProhibited` release tag. I'd never seen that before. But basically what you're saying is that you're not in favor of the two abstract classes being accessible outside of itwin-core. Is that right?

pmconne:
> basically what you're saying is that you're not in favor of the two abstract classes being accessible outside of itwin-core. Is that right?

Right.

core/backend/src/annotations/ElementDrivesTextAnnotation.ts
claudiaareneee: I haven't worked with generators much so for sanity sake, something like this?:

```ts
/** Depth first iterator of a text block */
export function* getTextBlockGenerator(block: TextBlockComponent, parent?: ContainerComponent): IterableIterator<{parent?: ContainerComponent, current: TextBlockComponent}> {
  yield { parent, current: block };

  if (block instanceof ContainerComponent) {
    for (const child of block.children) {
      yield* getTextBlockGenerator(child, block);
    }
  }
}

// then in the field code

const generator = getTextBlockGenerator(textBlock);
let result = generator.next();
while (!result.done){
  if (result.value.current.type === 'field') {
    // do something with field
  }
  result = generator.next();
}
```

pmconne: You can use ordinary `for..of` syntax. See for example [this usage](https://github.com/iTwin/itwinjs-core/blob/f7d840a9e7f1130b1adf8f779ea9e824ad9e1ef3/core/frontend/src/common/imdl/CompactEdges.ts#L101) of [this generator](https://github.com/iTwin/itwinjs-core/blob/f7d840a9e7f1130b1adf8f779ea9e824ad9e1ef3/core/frontend/src/common/imdl/CompactEdges.ts#L44).

core/common/src/annotation/TextBlock.ts
claudiaareneee: changed these to string literals. I added a type for unions of these for convenience (`RunComponentType` and `ContainerComponentType`)

claudiaareneee:
> > basically what you're saying is that you're not in favor of the two abstract classes being accessible outside of itwin-core. Is that right?
> 
> Right.

@pmconne, I'm going to go ahead and publish this PR. I still need to address this comment, and I have some other small todos left, but nothing structural. Mostly just additive. Would you mind giving this a deeper review now?

aruniverse:
@DanRod1999 this PR will likely merge before you return from vacation, however plz review it after you return (even if it has merged)

core/common/src/annotation/TextBlock.ts
```diff
@@ -816,6 +1096,113 @@ export class TextBlock extends TextBlockComponent {
...
+}
+
+/**
+ * Recursively traverses a tree of TextBlockComponent objects, yielding each component along with its parent container.
+ * This generator enables depth-first iteration over all components in a text block structure, including paragraphs, lists, and runs.
+ *
+ * @param block The root TextBlockComponent to start traversal from.
+ * @param parent (Optional) The parent ContainerComponent of the current block. For the root, this is undefined.
+ * @returns An IterableIterator yielding objects with the current component and its parent container.
+ * @beta
```
pmconne, line 861: You omitted a release tag, so it defaulted to `@public`. It can be at most `@beta` because `TextBlockComponent` is `@beta`. Is that what you want? (I added the tag to make extract-api succeed so I could see a summary of your API changes).

claudiaareneee: Beta is good

claudiaareneee: also yeah, I haven't run extract-api or lint yet

common/api/core-backend.api.md
```diff
@@ -4234,19 +4235,23 @@ export class KnownLocations {
 }
 
 // @beta
-export function layoutTextBlock(args: LayoutTextBlockArgs): TextBlockLayout;
-
-// @beta
-export interface LayoutTextBlockArgs {
+export interface LayoutStyleArgs {
```
pmconne: nit: Unclear what "style" refers to here. This serves as a common set of parameters used for laying out graphemes and text blocks. Consider renaming to `LayoutTextArgs`.

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -583,12 +621,18 @@ export class LineLayout {
...
+  /** Gets the array of RunLayout objects contained in this line. */
   public get runs(): ReadonlyArray<RunLayout> { return this._runs; }
+  /** Indicates whether this line contains any runs. */
   public get isEmpty() { return this._runs.length === 0; }
+  /** Gets the last RunLayout in this line. */
   public get back(): RunLayout {
     assert(!this.isEmpty);
     return this._runs[this._runs.length - 1];
   }
+  /** Gets or sets the marker RunLayout for this line, used for list item markers. */
```
pmconne: Please don't use the term you are defining to define the term. I assume "marker" refers to like, bullet points, numbers, etc preceding each item in a list?

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -566,15 +600,19 @@ export class RunLayout {
  * @beta
  */
 export class LineLayout {
-  public source: Paragraph;
+  public source: TextBlockComponent;
```
pmconne: Not every `TextBlockComponent` can be the source for a `LineLayout`, right? Can you constrain this type to, e.g., `Paragraph | List`?

core/common/src/annotation/TextBlock.ts
```diff
@@ -574,73 +761,124 @@ export class FieldRun extends TextBlockComponent {
...
-      return false;
-    }
+    return (other instanceof List) && super.equals(other);
+  }
 
-    return this.runs.every((run, index) => run.equals(other.runs[index]));
+  public appendToListItem(run: Run, itemIndex?: number): void {
+    if (!this.children || this.children.length === 0) this.appendChild(Paragraph.create());
+    const listItem = this.children[itemIndex ?? this.children.length - 1];
+    listItem.appendChild(run);
```
pmconne: `listItem` may be `undefined` here if `itemIndex` is out of bounds. It's also weird that if someone passes an `itemIndex` > 0 and `this.children` is `undefined` or empty, you add an empty paragraph that they didn't ask for.

What is the motivation for this method, vs just using `appendChild`?

common/api/core-common.api.md
```diff
@@ -1817,6 +1817,28 @@ export class ConflictingLocksError extends IModelError {
     static isError<T extends ConflictingLocks>(error: any): error is T;
 }
 
+// @beta
+export abstract class ContainerComponent<T extends TextBlockComponent = TextBlockComponent> extends TextBlockComponent {
```
pmconne: I still dislike that this tall ugly class hierarchy is being exposed publicly. It leaks your implementation details and makes the APIs harder to understand and more awkward to use. Please strongly consider exposing discriminated unions of mostly flat types instead.

common/api/core-common.api.md
```diff
@@ -1817,6 +1817,28 @@ export class ConflictingLocksError extends IModelError {
...
+    equals(other: TextBlockComponent): boolean;
+    get isEmpty(): boolean;
+    get last(): T | undefined;
+    toJSON(): ContainerComponentProps;
+    abstract readonly type: ContainerComponentType;
+}
+
+// @beta
+export interface ContainerComponentProps extends TextBlockComponentProps {
+    children?: TextBlockComponentProps[];
```
pmconne: This is classic violation of [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle). No sub-type of `ContainerComponentProps` is actually supposed to be permitted to contain any all types of `TextBlockComponentProps`, but the base type declares that all of them can. You are literally permitting people to put a `TextBlockProps` inside of a `ParagraphProps`, for example. Sub-types cannot restrict the capabilities of super-types. This is another undesirable consequence of exposing your implementation details. Now you've got to have logic in all your `fromJSON` methods to deal with people giving you children that you don't allow for a particular parent. Instead, you should define your types to make invalid states like that impossible (from the compiler's perspective).

core/common/src/annotation/TextBlock.ts
```diff
@@ -816,6 +1098,113 @@ export class TextBlock extends TextBlockComponent {
...
+/**
+ * Recursively traverses a tree of TextBlockComponent objects, yielding each component along with its parent container.
+ * This generator enables depth-first iteration over all components in a text block structure, including paragraphs, lists, and runs.
+ *
+ * @param block The root TextBlockComponent to start traversal from.
+ * @param parent (Optional) The parent ContainerComponent of the current block. For the root, this is undefined.
+ * @returns An IterableIterator yielding objects with the current component and its parent container.
+ * @beta
+ */
+export function* getTextBlockGenerator(block: TextBlockComponent, parent?: ContainerComponent): IterableIterator<{ parent?: ContainerComponent, current: TextBlockComponent }> {
```
pmconne: Shouldn't this just take a single, mandatory `ContainerComponent`? It's not useful for non-containers.

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -255,36 +256,69 @@ export class TextStyleResolver {
...
+   * @param component The TextBlockComponent to mend styles for.
+   * @returns A cloned TextBlockComponent with updated style overrides for all children.
+   */
+  public mendSettings(component: TextBlockComponent): TextBlockComponent {
+    const block = component.clone();
+
+    if (component instanceof ContainerComponent) {
+      component.children.forEach(child => {
+        child.styleOverrides = { ...component.styleOverrides, ...child.styleOverrides };
+        if (child instanceof TextBlockComponent) {
```
pmconne: Under what circumstances would it **not** be a `TextBlockComponent`? Did you have to add this to placate the compiler or something?

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -689,71 +749,124 @@ export class TextBlockLayout {
...
           curLine = this.flushLine(context, curLine);
-          continue;
         }
 
-        // If this is a tab, we need to apply the tab shift first, and then we can treat it like a text run.
-        applyTabShift(run, curLine, context);
+        break;
+      }
+      case "list": {
+        const list = component as ContainerComponent;
```
pmconne: Why the need to cast here? If compiler complained, what was its complaint?

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -689,71 +749,124 @@ export class TextBlockLayout {
...
+        const paragraph = component as ContainerComponent;
+        paragraph.children.forEach(child => curLine = this.populateComponent(child, context, docWidth, curLine, paragraph, depth));
+        const nextSibling = parent?.children[component.index + 1];
+        if (curLine && nextSibling) {
+          curLine = this.flushLine(context, curLine, nextSibling, true, depth);
+        }
+        break;
+      }
+      case "text": {
+        const run = component as Run;
```
pmconne: These casts should not be necessary. I suspect the compiler wants them because you defined `type` in an abstract base class. Use interfaces / discriminated unions to solve this. You can still use the messy class hierarchy behind the scenes to implement the public API.

core/backend/src/internal/annotations/fields.ts
```diff
@@ -269,11 +269,9 @@ export function updateField(field: FieldRun, context: UpdateFieldsContext): bool
 // of fields whose display strings changed as a result.
 export function updateFields(textBlock: TextBlock, context: UpdateFieldsContext): number {
   let numUpdated = 0;
-  for (const paragraph of textBlock.paragraphs) {
-    for (const run of paragraph.runs) {
-      if (run.type === "field" && updateField(run, context)) {
-        ++numUpdated;
-      }
+  for (const { current } of getTextBlockGenerator(textBlock)) {
+    if (current.type === "field" && current instanceof FieldRun && updateField(current, context)) {
```
pmconne: Another cast that should not be necessary.

common/api/core-common.api.md
claudiaareneee: is it enough to just not export this? would that satisfy _the mostly flat types_ comment? Or would you rather I did away with this abstract class and created some sort of utility helper instead? What I'm trying to avoid is having a bunch of duplicate code in terms of children setters/getters, appendChild, and other class methods.

pmconne: How are you going to not export the base class for your current public API?

The focus should be on how easy it is for other people to use your API, not how easy it is for you to write it. But if you do it right, you can reuse code inside your implementation without making your API difficult to use.

claudiaareneee: fair enough. so just for clarification, the desired outcome here is that ContainerComponent goes away and the only abstract class that's exported is the original TextBlockComponent?

claudiaareneee:
@pmconne, if you pull this branch, here's the dta key-in that I've been using for ease of testing: 

```
dta text json "{'styleOverrides':{'lineHeight':1,'color':45643},'type':'textBlock','children':[{'styleOverrides':{},'type':'paragraph','children':[{'styleOverrides':{'fontName':'Arial'},'type':'text','content':'Example:','baselineShift':'none'}]},{'styleOverrides':{'fontName':'Arial','color':50390,'indentation':4},'type':'list','children':[{'styleOverrides':{},'type':'paragraph','children':[{'styleOverrides':{'fontName':'Arial'},'type':'text','content':'Apples','baselineShift':'none'}]},{'styleOverrides':{},'type':'paragraph','children':[{'styleOverrides':{'fontName':'Arial'},'type':'text','content':'Bananas','baselineShift':'none'}]},{'styleOverrides':{},'type':'paragraph','children':[{'styleOverrides':{'fontName':'Arial'},'type':'text','content':'Oranges','baselineShift':'none'}]},{'styleOverrides':{},'type':'paragraph','children':[{'styleOverrides':{'fontName':'Arial'},'type':'text','content':'Grapes','baselineShift':'none'},{'styleOverrides':{'fontName':'Arial','listMarker':'A.','color':255},'type':'list','children':[{'styleOverrides':{},'type':'paragraph','children':[{'styleOverrides':{'fontName':'ENGINEERING','isBold':true,'color':255255},'type':'text','content':'Green','baselineShift':'none'}]},{'styleOverrides':{},'type':'paragraph','children':[{'styleOverrides':{'fontName':'ENGINEERING'},'type':'text','content':'Red','baselineShift':'none'}]}]}]}]}],'styleId':'','width':20,'justification':'left','margins':{'left':1,'right':1,'top':1,'bottom':1}}"
```

It looks like this: 

image

core/common/src/annotation/TextBlock.ts
```diff
@@ -49,20 +62,48 @@ export interface TextBlockStringifyOptions {
...
+  /** Discriminator field for the type of [[TextBlockComponent]]. */
+  public readonly abstract type: RunComponentType | ContainerComponentType;
 
   /** @internal */
   protected constructor(props?: TextBlockComponentProps) {
     this._styleOverrides = TextStyleSettings.cloneProps(props?.styleOverrides ?? {});
   }
 
+  /** The index of this component within its parent container's [[ContainerBase.children]] array. */
+  public get index(): number {
```
pmconne: You only use this in two spots in TextBlockLayout. In both spots, you are doing a `children.forEach((child, index))` - so you already have the actual index, and don't need to rely on this error-prone redundant index. Remove this, and now you don't need any of these other children-related methods like `appendChild`, because you don't have to keep `index` in sync. You can delete them all and just expose `children` as an array people can manipulate directly. Then the Container component can be deleted since all it contains is an array of children.

core/common/src/annotation/TextBlock.ts
```diff
@@ -49,20 +62,48 @@ export interface TextBlockStringifyOptions {
...
+ * The [[TextBlock]] can specify an [AnnotationTextStyle]($backend) that formats its contents.
+ * Each component can specify an optional [[styleOverrides]] to customize that formatting.
  * @beta
  */
 export abstract class TextBlockComponent {
   private _styleOverrides: TextStyleSettingsProps;
+  private _index: number = 0;
+
+  /** Discriminator field for the type of [[TextBlockComponent]]. */
+  public readonly abstract type: RunComponentType | ContainerComponentType;
```
pmconne: This is the root of all your casting issues. You have to define `type` only on the leaf types (Paragraph, List, TextRun, etc) so the compiler can use it to narrow types.

core/common/src/annotation/TextBlock.ts
```diff
@@ -675,17 +962,18 @@ export interface TextBlockProps extends TextBlockComponentProps {
...
- * You can change the content of the document by directly modifying the contents of its [[paragraphs]], or via [[appendParagraph]] and [[appendRun]].
+/** Represents a formatted text document consisting of a series of [[Paragraph]]s and [[List]]s, each laid out on a separate line and containing their own content.
+ * [[Paragraph]]s and [[List]]s act as branches and can contain [[Paragraph]]s, [[List]]s, or leaf nodes in the form of [[Run]]s.
+ * To modify the children, you can either directly set the [[TextBlock.children]] property or use the provided methods to append new elements.
  * No word-wrapping is applied to the document unless a [[width]] greater than zero is specified.
  * @see [[TextAnnotation]] to position a text block as an annotation in 2d or 3d space.
  * @beta
  */
-export class TextBlock extends TextBlockComponent {
+export class TextBlock extends ContainerBase<(Paragraph | List)> {
```
pmconne: Why does TextBlock contain both Paragraphs and Lists as top-level children? Why can't it only contain Paragraphs, which may contain Lists?

pmconne:
@claudiaareneee Look at my [most recent commit](https://github.com/iTwin/itwinjs-core/pull/8418/commits/da06377575155d3e6145fc80389f1d6eebb2e69b), specifically at TextBlock.ts. I simplified the public API and also made TextBlock only support Paragraphs as direct children; you can put a List inside a Paragraph. Unfortunately many of your tests are based on Lists being direct children of TextBlock so they'll need reworking.

See what you think and if I overlooked anything obvious. It's a single commit so we can revert it if necessary.

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -255,36 +256,62 @@ export class TextStyleResolver {
...
 
-  /** Resolves the effective style for a [Run]($common). Run should be child of provided Paragraph and TextBlock. */
-  public resolveRunSettings(paragraph: Paragraph, run: Run): TextStyleSettings {
-    let settings = this.resolveParagraphSettingsImpl(paragraph);
+  /**
+   * Resolves and applies style overrides to a TextBlock, ensuring all child components are updated.
+   * @param component The TextBlock to resolve and mend styles for.
+   * @returns A cloned TextBlock with resolved and mended styles.
+   */
+  public resolveAndMendStyle(component: TextBlock): TextBlock {
```
pmconne: Get rid of all this "clone and mend settings" stuff as directed in your TODO in `populateComponent`.

claudiaareneee:
@pmconne, quick comments:
- the only true missing thing is the `clearStyleOverrides` for the structural components, but I can implement that
- instead of `appendChild`, would the user manually edit the `children` property of the different structural components?

more might come up as i'm fixing tests, but those were what i noticed first

pmconne:
> @pmconne, quick comments:
> 
>     * the only true missing thing is the `clearStyleOverrides` for the structural components, but I can implement that

I don't understand. `TextBlockComponent.clearStyleOverrides` still exists.

>     * instead of `appendChild`, would the user manually edit the `children` property of the different structural components?

`thing.children.push(child)` instead of `thing.appendChild(child)`.

claudiaareneee:
> I don't understand. TextBlockComponent.clearStyleOverrides still exists.

Right, but it doesn't iterate through the children and compare against the `preserveChildrenOverrides` flag. Check what's in `Paragraph.clearStyleOverrides` in master. We should still support this, I assume:

https://github.com/iTwin/itwinjs-core/blob/394223f53d56ab519f7ea1d7a56392c0b2fef263/core/common/src/annotation/TextBlock.ts#L616-L624

> `thing.children.push(child)` instead of `thing.appendChild(child)`.

Cool, makes sense

pmconne:
> Right, but it doesn't iterate through the children and compare against the preserveChildrenOverrides flag. Check what's in Paragraph.clearStyleOverrides in master. We should still support this, I assume:

Oh, right. Fixed.

core/common/src/annotation/TextBlock.ts
```diff
@@ -816,6 +897,113 @@ export class TextBlock extends TextBlockComponent {
...
+ *
+ * @param parent The root container whose children should be traversed.
+ * @returns An IterableIterator yielding objects with the current component and its parent container.
+ * @beta
+ */
+export function* getTextBlockGenerator(parent: StructuralTextBlockComponent): IterableIterator<{ parent: StructuralTextBlockComponent, child: List | Paragraph | Run }> {
+  for (const child of parent.children) {
+    yield { parent, child };
+    if (child.type === "list" || child.type === "paragraph") {
+      yield* getTextBlockGenerator(child);
```
pmconne:
```suggestion
export function* traverseTextBlockComponent(parent: StructuralTextBlockComponent): IterableIterator<{ parent: StructuralTextBlockComponent, child: List | Paragraph | Run }> {
  for (const child of parent.children) {
    yield { parent, child };
    if (child.type === "list" || child.type === "paragraph") {
      yield* traverseTextBlockComponent(child);
```

claudiaareneee: fixed

core/backend/src/annotations/TextBlockLayout.ts
claudiaareneee: fixed

common/api/core-backend.api.md
claudiaareneee: fixed

core/backend/src/annotations/TextBlockLayout.ts
claudiaareneee: fixed

aruniverse:
@claudiaareneee you are no longer blocked on this pr right?

core/common/src/annotation/TextStyle.ts
```diff
@@ -172,6 +222,17 @@ export interface TextStyleSettingsProps {
...
+   * In lists, the indentation is added to offset of list items.
+   * [[listMarker]]s will be centered on [[indentation]] + [[tabInterval]]*(depth - 1/2).
+   * [[Paragraph]]s will start at [[indentation]] + [[tabInterval]]*depth.
+   * Default: 0 meters.
+   */
+  indentation?: number;
+  /** The marker used to indicate the start of a list item.
+   * Default: "1.".
+   */
+  listMarker?: ListMarker;
```
DanRod1999, line 221: general question, is there a reason we default to an order list here? I guess in things like markdown or word, there isn't really a "default" maker for order lists (unordered lists I guess use defaults based on whether you're a sublist or not). Just curious if this is a design decision for a reason, cause I probably would've randomly defaulted to an unordered list, or not allow value to be undefined or something. Or have a boolean isOrderedList somewhere that chooses the appropriate default for you (but that sounds annoying too)

claudiaareneee: I think inverse to you, I just default to ordered lists. In most of the drawings I've been referencing, they use ordered lists. The default value here isn't too important because I expect consuming applications to have seperate buttons for ordered and unordered lists, so they would provide whichever value makes sense in their contexts. 

TextStyleSettingsProps are used all over the place so it would be tedious to have to explicitly provide this. Also it's not relevant in individual runs, so it would be confusing at the minimum to have to provide this value.

common/api/core-common.api.md
```diff
@@ -6744,6 +6785,40 @@ export interface OrbitGtBlobProps {
     sasToken: string;
 }
 
+// @beta
+export enum OrderedListMarker {
```
aruniverse: do we really need all these options?

isnt it just 
- lowercaseLetter
- uppercaseLetter
- number
- withParenthesis
- withPeriod

core/common/src/annotation/TextStyle.ts
DanRod1999: cool makes sense thanks

core/backend/src/test/annotations/TextBlock.test.ts
```diff
@@ -174,7 +170,7 @@ describe("layoutTextBlock", () => {
       const textBlock = TextBlock.create({ styleOverrides: { fontName: "grandparent" } });
       const paragraph = Paragraph.create({ styleOverrides: { fontName: "parent" } });
       const run = TextRun.create({ content: "test", styleOverrides: { fontName: "child" } });
-      textBlock.paragraphs.push(paragraph);
+      textBlock.appendParagraph(paragraph);
```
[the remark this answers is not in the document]
claudiaareneee: This seems outdated. I don't think I have the `.paragraphs` attribute anymore. Gonna resolve this one.

common/api/core-common.api.md
claudiaareneee: Fair. I was going for simplest though. This already doesn't handle localization very well, so I expect to have to make edits to this in the future anyways. But that would need to be a seperate PR because it's not a priority right now

claudiaareneee: (I was told to not spend the effort on this now and just to leave it as is until we prioritize the localization)

pmconne: Seems simpler and more extensible if you did something like the following. The various ingredients seem pretty orthogonal.
```
{
  enumerator: "number" | "letter" | "roman-numeral",
  terminator?: "parenthesis" | "period",
  case?: "upper" | "lower",
}
```
This wouldn't require you to do anything with localization.

pmconne (commented):
Did you make re-read the TextBlock API docs to make sure everything is updated? (Asking so you can make adjustments if necessary before I look at them).

claudiaareneee:
I did, but l'm sure I've missed stuff. I'll look over it again real quick

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -224,45 +230,43 @@ export class TextStyleResolver {
...
-
-    return settings;
-  }
-
-  /** Resolves the effective style for a [TextAnnotationLeader]($common). The TextAnnotationLeader should be a sibling of the provided TextBlock. */
-  public resolveTextAnnotationLeaderSettings(leader: TextAnnotationLeader): TextStyleSettings {
+  /**
+   * Resolves the effective text style settings for a given TextBlockComponent, applying block-level overrides.
+   * @param component The TextBlockComponent whose style settings are to be resolved.
+   * @returns The resolved TextStyleSettings for the component.
```
pmconne: Your single `@param` doesn't match either of the actual parameters. (FWIW I usually find the `@param`s and `@returns` stuff superfluous; you can usually document most functions more clearly and succinctly with a single sentence).

core/backend/src/annotations/TextBlockLayout.ts
```diff
@@ -224,45 +230,43 @@ export class TextStyleResolver {
...
-
-    if (run.overridesStyle)
-      settings = settings.clone(run.styleOverrides);
-
-    return applyBlockSettings(settings, this.blockSettings);
+  /**
+   * Computes the indentation for a TextBlockComponent based on its style and nesting depth.
+   * @param component The TextBlockComponent for which to compute indentation.
+   * @param depth The nesting depth of the component.
+   * @returns The computed indentation value.
```
pmconne: Please remove doc tags throughout, or update them to be accurate.

core/common/src/annotation/TextBlock.ts
```diff
@@ -49,10 +49,34 @@ export interface TextBlockStringifyOptions {
    * Default: "undefined" - use "\t".
    */
   tabsAsSpaces?: number;
+  /** A string to insert in between the list marker and the list item.
+   * Default: " " - a single space..
```
pmconne:
```suggestion
   * Default: " " - a single space.
```

pmconne (approved):
Great PR, thank you for persisting with it. I have only nitpickery left for you. Please check for any needed updates to API docs before I give them a read-through.

PR #8425, opened 2025-08-13
Add CRS almostEqual

DanRod1999:
Add ability to compare horizontal CRS without comparing the description. This is useful for merging imodels with CRS data that is functionally equivalent, but would return as false with the current equals functions.
All other data seems necessary, however not sure if these values can also differ but be functionally the same:
- source
- id

I assume datum and ellipsoid id's are also constant and shouldn't differ between imodels, but not 100% sure.

### DIFF
--- a/core/common/src/geometry/CoordinateReferenceSystem.ts
+++ b/core/common/src/geometry/CoordinateReferenceSystem.ts
@@ -281,6 +281,46 @@ export class HorizontalCRS implements HorizontalCRSProps {
 
     return true;
   }
+
+  /** Compares two horizontal CRS, but does not compare descriptions.
+   *  Number compares are applied a minuscule tolerance.
+   *  @public */
+  public almostEquals(other: HorizontalCRS): boolean {
+    if (this.id !== other.id ||
+      this.source !== other.source ||
+      this.deprecated !== other.deprecated ||
+      this.epsg !== other.epsg ||
+      this.datumId !== other.datumId ||
+      this.ellipsoidId !== other.ellipsoidId ||
+      this.unit !== other.unit)
+      return false;
+
+    if ((this.datum === undefined) !== (other.datum === undefined))
+      return false;
+
+    if (this.datum && !this.datum.equals(other.datum!))
+      return false;
+
+    if ((this.ellipsoid === undefined) !== (other.ellipsoid === undefined))
+      return false;
+
+    if (this.ellipsoid && !this.ellipsoid.equals(other.ellipsoid!))
+      return false;
+
+    if ((this.projection === undefined) !== (other.projection === undefined))
+      return false;
+
+    if (this.projection && !this.projection.equals(other.projection!))
+      return false;
+
+    if ((this.extent === undefined) !== (other.extent === undefined))
+      return false;
+
+    if (this.extent && !this.extent.equals(other.extent!))
+      return false;
+
+    return true;
+  }
 }
 
 /** Vertical Geographic Coordinate reference System definition


core/common/src/geometry/CoordinateReferenceSystem.ts
```diff
@@ -281,6 +281,46 @@ export class HorizontalCRS implements HorizontalCRSProps {
...
+    if ((this.datum === undefined) !== (other.datum === undefined))
+      return false;
+
+    if (this.datum && !this.datum.equals(other.datum!))
+      return false;
+
+    if ((this.ellipsoid === undefined) !== (other.ellipsoid === undefined))
+      return false;
+
+    if (this.ellipsoid && !this.ellipsoid.equals(other.ellipsoid!))
```
ben-polinsky, line 313: No non-null assertions, please.

Could consider extracting a helper fn (`compareOptionalProperties` for instance) since all of these comparisons follow the same pattern.

Might want to clean up the other non null assertions in this file while you are at it :) I know @tcobbs-bentley would appreciate it.

core/common/src/geometry/CoordinateReferenceSystem.ts
```diff
@@ -281,6 +281,46 @@ export class HorizontalCRS implements HorizontalCRSProps {
 
     return true;
   }
+
+  /** Compares two horizontal CRS, but does not compare descriptions.
+   *  Number compares are applied a minuscule tolerance.
+   *  @public */
+  public almostEquals(other: HorizontalCRS): boolean {
```
pmconne: `almostEquals` suggests a fuzzy tolerance. Call it something like `isEquivalent` to convey what it actually does.

core/common/src/geometry/CoordinateReferenceSystem.ts
tcobbs-bentley: Actually, this particular sequence is used a lot, and I personally feel that it is fine with an eslint-disable comment. (I added such comments in my lint warning cleanup [PR](https://github.com/iTwin/itwinjs-core/pull/8391).) A whole bunch of equality comparison functions in the code first check that two values are either both undefined or both defined, then check if one is defined, and if it is, use a non-null assertion on the other due to the fact that the linter and compiler aren't smart enough to realize that if one is defined, the other has to also be defined. Side note: my PR already cleans up all the eslint warnings in this file, but hasn't been merged.

ben-polinsky: See that point. I guess the overhead of the disable comment is pretty much the same as checking if the other is defined. Plus could be wrapped up in the helper fn. Either way.

tcobbs-bentley: I believe that something like this would work, but I'm not sure about the effort of updating the code to use it:

```ts
interface HasEquals {
  equals(other: this): boolean;
}

function equals<T extends HasEquals>(left: T | undefined, right: T | undefined): boolean {
  if ((left === undefined) !== (right === undefined))
    return false;
  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  return left === undefined || (left !== undefined && left.equals(right!));
}
```

The code here would then be:

```ts
    if (!equals(this.ellipsoid, other.ellipsoid))
      return false;
```

tcobbs-bentley: Note: I _think_ that the above is equivalent to the code highlighted at the top of this conversation, but I'm not absolutely 100% sure about that.

DanRod1999: would like to leave as is then if we agree thats fine

tcobbs-bentley: I feel you should add the eslint-disable line, but leave the logic alone.

PR #8426, opened 2025-08-13
Use per-model tiles for masking

pmconne:
Producing the model mask from batched tiles requires drawing all of the models and discarding geometry in the vertex shader for models not included in the mask. This results in significant performance degradation.

Just use non-batched tiles instead to draw exactly and only the models required for the mask.


pmconne:
Awaiting ImageTests results.

pmconne:
V1 image tests show no differences, as expected.
V4 shows only one unexpected/notable difference:
image
That view is "DelawareRiverWaterfrontRail_1 Map and Reality Masked".
I'm not sure if I've introduced a bug or fixed one...need to look at the saved view.

pmconne:
I fail to reproduce ImageTests' results for the "Map and Reality Masked" view on my machine.
I did discover that view has a mis-spelling in the reality model URL ("tileRoot.json" should be "TileRoot.json") which explains why it has never displayed the reality model.
I fixed that locally and the reality model correctly displays and masks the map.

The view contains no masking besides reality model masking the map, and no design models are enabled. I will re-run ImageTests to see if it was a fluke.

pmconne:
The difference recurred. It's not masking - you can see the map displayed under new geometry.
It looks like the outline of the design models in the viewport's hilite color.
image

If I enable the "Southbound corridor cant" model in display-test-app it does (correctly) mask:
image

eringram:
> I fail to reproduce ImageTests' results for the "Map and Reality Masked" view on my machine. I did discover that view has a mis-spelling in the reality model URL ("tileRoot.json" should be "TileRoot.json") which explains why it has never displayed the reality model. I fixed that locally and the reality model correctly displays and masks the map.
> 
> The view contains no masking besides reality model masking the map, and no design models are enabled. I will re-run ImageTests to see if it was a fluke.

Does the reality model URL in the image tests saved view need updating? (and I guess probably the reference image, since with the correct URL the reality model should display like your latest screenshot)

And to clarify, you think the difference was not caused by your changes, since with the URL typo there is no masking being done?

pmconne:
> Does the reality model URL in the image tests saved view need updating? (and I guess probably the reference image, since with the correct URL the reality model should display like your latest screenshot)

Currently the view doesn't actually test what it appears to have been intended to test. The reality model is supposed to mask out the map, but the reality model doesn't display because of the typo. So we should probably fix the typo and update the reference image to make the test useful.

> And to clarify, you think the difference was not caused by your changes, since with the URL typo there is no masking being done?

No, the image difference appears to have been caused by my changes, just unclear how. I continue to fail to reproduce it on my machine.

jason-crow:
@pmconne was there any further progress on this or plans to do so? Asking since this was suggested as a fix for https://github.com/iTwin/itwin-graphics-backlog/issues/692 which itself relates to an open service case we have

pmconne:
> @pmconne was there any further progress on this or plans to do so?

We definitely want to finish this up but I haven't had the opportunity to pursue this image difference.
I am currently blocked on other stuff so I'll try to figure this out.

pmconne:
I suspected it might be something left over from previous tests polluting subsequent tests so I ran just the DelawareRiverWaterfrontRail.bim tests, on both main and this branch. They produce identical results - only minor differences from reference in map imagery, no differences in the mask. The "1 Map and Reality Masked" view that previously showed a difference showed no differences at all.

Running the full set of tests against this branch again now.

pmconne:
Image difference did not recur after running full ImageTests against this branch.

pmconne:
@jason-crow do you need this backported to 5.1.x?

jason-crow:
> @jason-crow do you need this backported to 5.1.x?

I apologize @pmconne on the delay, but yes we would benefit from a backport if you are able that would allow us to ship this sooner

pmconne:
@aruniverse backport or no? I know 5.2 is imminent.

aruniverse:
> @aruniverse backport or no? I know 5.2 is imminent.

Was hoping to avoid it, but we can do it. If the pipelines give us trouble and its a lot of work we'll just push it to 5.2

aruniverse:

PR #8430, opened 2025-08-13
Fix flakey ElectronHost window size/position tests

tcobbs-bentley:
This will hopefully fix the intermittent failures of the ElectronHost size/position tests on Linux. The fix runs as before, but if the size/position don't match the expected values after the initial 250ms wait (which before would cause the test to fail), it will repeatedly check for the correct size/position every 50ms up to 20 times.

### DIFF
--- a/common/changes/@itwin/core-electron/tcobbs-fix-electron-host-tests_2025-08-13-22-55.json
+++ b/common/changes/@itwin/core-electron/tcobbs-fix-electron-host-tests_2025-08-13-22-55.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-electron",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-electron"
+}


tcobbs-bentley:
@pmconne FYI

core/electron/src/test/backend/ElectronHost.test.ts
```diff
@@ -217,6 +217,11 @@ async function testWindowSizeSettings() {
   window.setSize(width, height);
   await BeDuration.wait(250); // wait for new size to be saved to settings file
   sizeAndPos = ElectronHost.getWindowSizeAndPositionSetting(storeWindowName);
+  for (let i = 0; i < 20 && (sizeAndPos?.width !== width || sizeAndPos?.height !== height); ++i) {
```
ben-polinsky, line 220: Should this only be run if running on linux?

tcobbs-bentley: Even though we've only seen it fail on Linux, I believe that it could fail anywhere. The 200ms debounce happens on all platforms.

tcobbs-bentley: To clarify: the event handler that is set up to listen for the resize and move events has a 200ms debounce attached, meaning that it is a minimum of 200ms between the time of the resize/move and the time that the stored size/position values get updated.

ben-polinsky: Indeed.

ben-polinsky (approved):

aruniverse:

PR #8435, opened 2025-08-15
Fixed all lint errors

saeeedtorabi:


core/geometry/src/serialization/BGFBAccessors.ts
```diff
@@ -13,6120 +14,6396 @@ import { flatbuffers } from "flatbuffers";
...
-  this.bb_pos = i;
-  this.bb = bb;
-  return this;
-}
-
-/**
- * @returns number
- */
-x(): number {
-  return this.bb!.readFloat64(this.bb_pos);
```
bbastings, line 76: // automatically generated by the FlatBuffers compiler, do not modify
Do not change this file, just disable the warning at the top for the entire file.

dassaf4: See how it is [disabled in the main lint PR for a similar file](https://github.com/iTwin/itwinjs-core/pull/8391/commits/d6bd4351ba9e0002fe31080b4cea86ce1bdb5ea1).

core/geometry/src/Geometry.ts
```diff
@@ -1326,8 +1327,11 @@ export class Geometry {
     if (array === undefined)
       return undefined;
     const clonedArray: T[] = [];
+    let clone: T | undefined;
```
dassaf4: When you resolve conflicts, take the [fix from the main PR](https://github.com/iTwin/itwinjs-core/pull/8391) here.

core/geometry/src/serialization/BGFBAccessors.ts
saeeedtorabi: Done

core/geometry/src/Geometry.ts
saeeedtorabi: Done

core/geometry/src/bspline/BezierCurve3d.ts
```diff
@@ -67,8 +68,13 @@ export class BezierCurve3d extends BezierCurveBase {
...
     const result = LineString3d.create();
-    for (let i = 0; i < this._polygon.order; i++)
-      result.addPoint(this.getPolePoint3d(i)!);
+    let point: Point3d | undefined;
+    for (let i = 0; i < this._polygon.order; i++) {
+      point = this.getPolePoint3d(i);
+      assert(undefined !== point, "BezierCurve3d.copyPointsAsLineString: point should be defined");
+      if (point)
+        result.addPoint(point);
+    }
```
dassaf4: In general, I prefer avoiding `assert` for easily confirmed truths, like loop-bound array indexing. In this case, I'd take the opportunity to improve the loop's efficiency. Note this works because `addPoint` does not capture its input, unlike if we were pushing the point to a `Point3d[]` (why?).

```suggestion
    const point = Point3d.createZero();
    for (let i = 0; i < this._polygon.order; i++) {
      if (this.getPolePoint3d(i, point))
        result.addPoint(point);
    }
```

core/geometry/src/bspline/BezierCurve3d.ts
```diff
@@ -161,7 +167,8 @@ export class BezierCurve3d extends BezierCurveBase {
     const order = this.order;
     if (!transform) {
       this.allocateAndZeroBezierWorkData(order - 1, 0, 0);
-      const bezier = this._workBezier!;
+      const bezier = this._workBezier;
+      assert(undefined !== bezier, "BezierCurve3d.extendRange: bezier should be defined");
```
dassaf4: Optional instance "work" objects are another example of what I'd call an unnecessary assert. Can you avoid `!` and `assert` by turning `allocateAndZeroBezierWorkData` into a ["this-based" type guard](https://www.typescriptlang.org/play/?#code/MYGwhgzhAECC0G8BQ1XQA4FcBGICWw0YA-AFzQQAuATngHYDmA3CmlrgdNmRTfc61Tt8hYDyq1GLNNAD0s6PTyUiROgBMuazRLzpomDQFMAZvSOaT1APYBbaJQAWRvNQcBPdEcEYcIxXTKABQAHuS6jAA00O7hfIwAlOROeDCpiERxkgzR2Fn80AC+iD4y1EaUmNR0BsZmdBbQAIQAvC3QQSkQAHRg0O0hCdAAZMO16qbmmq3tnY6p3dj9MQnSaIVIG0gmhsCUeNY16JKUsEG27rDksEPIMvLQwIcQ1iBG3SDWDEEABmAfRkYTmWABIEBdYL0AUDHIUfqs5Ao8CA3gwwCByBDeooYOhrFA8Lh3ONJg11D48CYOlilJQggAiZwo6z06L09SvEDuJr0hK3UpoJ50F5vD5fX7-N4w0Hgy5QqUMJxwhEPSieIzQBiYMDUTSQCBGaiUGBYvrpCb1CwC1BCkXvT7fH7YaGKxwyrHOhVK+FMREeLya7W6ohQQ3G6AenHQC1Ta1+22ve3in7AF3A9pgrGpr2wn1+5Go9GYuWEdJ4glEkmW8kyDZbY70U5BBoAdzgQT5LCAA)?

This won't work for a static "work" object, but a `!` that accompanies one of these can be removed by employing an instance getter that lazy-initializes the static object if it is not already defined.

core/geometry/src/Geometry.ts
```diff
@@ -7,6 +7,7 @@
  * @module CartesianGeometry
  */
 
+import { assert } from "@itwin/core-bentley";
```
tcobbs-bentley: This doesn't seem to be used anywhere, which I would expect to generate either a lint error or a lint warning.

core/geometry/src/bspline/BezierCurve3d.ts
```diff
@@ -67,8 +68,13 @@ export class BezierCurve3d extends BezierCurveBase {
   /** Return poles as a linestring */
   public copyPointsAsLineString(): LineString3d {
     const result = LineString3d.create();
-    for (let i = 0; i < this._polygon.order; i++)
-      result.addPoint(this.getPolePoint3d(i)!);
+    let point: Point3d | undefined;
+    for (let i = 0; i < this._polygon.order; i++) {
+      point = this.getPolePoint3d(i);
+      assert(undefined !== point, "BezierCurve3d.copyPointsAsLineString: point should be defined");
```
tcobbs-bentley: Note that `assert` is equivalent to `!` on release builds, and only throws an exception if the condition is not met on development builds (NODE_ENV=development). If this is the intended behavior for all of your asserts, that's fine, but they should only be used when you are **extremely** confident that they will always pass. I created `expectDefined` for cases that don't meet the confidence level required for `assert`, although in most cases, some other local logic is preferable to using either `assert` or `expectDefined`.

tcobbs-bentley: (I do see that you check if point is defined below, so it seems like this particular `assert` is intentional.)

tcobbs-bentley:
I made one comment about your `assert` usage, and ironically that particular place seems like a totally appropriate usage of `assert`. But a whole bunch of your other changes here just swap out `!` for `assert`, meaning that while they will now catch problems in _development_ builds, they will continue to behave as before for _production_ builds. I don't know the underlying code, so it's quite possible that all of these changes are in places where you really are extremely confident that undefined will never show up. However, it doesn't seem to me like this is much of an improvement over the old `!` usage. And since it gets rid of the warnings, it means that the code isn't likely to ever be reexamined to verify that the `assert` usage is appropriate.

dassaf4 (commented):
I started reviewing these before getting sidetracked. I agree with @tcobbs-bentley and would like to see bespoke changes to remove each `!`, not generic replacement of `!` with `assert`. Of course this is more work. But this can lead to better code, as in my first two suggestions.

tcobbs-bentley:
> I started reviewing these before getting sidetracked. I agree with @tcobbs-bentley and would like to see bespoke changes to remove each `!`, not generic replacement of `!` with `assert`. Of course this is more work. Usually this means an extra if statement and an early out. But this can also lead to better code, as in my first two suggestions.

I created [`expectDefined`](https://github.com/iTwin/itwinjs-core/blob/3d438797442f932b88c5573f5101583db4d15cbe/core/bentley/src/Expect.ts#L16) and [`expectNotNull`](https://github.com/iTwin/itwinjs-core/blob/3d438797442f932b88c5573f5101583db4d15cbe/core/bentley/src/Expect.ts#L33) to use as a quick replacement for `!` that could then potentially be cleaned up more appropriately at a later date. Furthermore, they behave as a drop-in replacement for `!`, thus not requiring variables to be assigned and then asserted to be defined. I feel that most of the places that have been updated to use `assert` in this PR should instead use `expectDefined` or `expectNotNull`.

dassaf4:
> I feel that most of the places that have been updated to use `assert` in this PR should instead use `expectDefined` or `expectNotNull`.

Right. So, it's your call @saeeedtorabi if you want to use the new functions as an interim solution. But eventually most if not all should be replaced with individual fixes avoiding `!` and `assert`.

@tcobbs-bentley, just to confirm: is core-geometry holding up other groups' PRs due to the linter? If so, the interim solution is more expedient.

tcobbs-bentley:
> @tcobbs-bentley, just to confirm: is core-geometry holding up other groups' PRs due to the linter? If so, the interim solution is more expedient.

No; it's just that all the non-null assertion lint warnings are very annoying, and basically hide any other lint warnings that might be important.

dassaf4:
> No; it's just that all the non-null assertion lint warnings are very annoying, and basically hide any other lint warnings that might be important.

Capiche. 1000+ warnings will do that!

saeeedtorabi:
> Right. So, it's your call @saeeedtorabi if you want to use the new functions as an interim solution. But eventually most if not all should be replaced with individual fixes avoiding `!` and `assert`.

ok I will go over them again and fix the ones I'm confident using individual fixes. I will leave the ones that I dont get the code logic for @dassaf4

core/geometry/src/Geometry.ts
saeeedtorabi: I removed it.

core/geometry/src/bspline/BezierCurve3d.ts
saeeedtorabi: I removed assert.

core/geometry/src/bspline/BezierCurve3d.ts
saeeedtorabi: I removed assert the way you suggested

core/geometry/src/bspline/BezierCurve3d.ts
saeeedtorabi: Done

core/geometry/src/Geometry.ts
tcobbs-bentley: It seems like you haven't pushed your changes for this and the other comments you responded.

saeeedtorabi: No not yet. I'm making changes locally and replying to the comments so I dont forget. Please wait for the next commit. FYI, I changed the PR to draft for now.

core/geometry/src/serialization/BGFBAccessors.ts
```diff
@@ -7,6126 +7,6127 @@ import { flatbuffers } from "flatbuffers";
 /* eslint-disable @typescript-eslint/naming-convention */
 /* eslint-disable @typescript-eslint/explicit-member-accessibility */
 /* eslint-disable @itwin/prefer-get */
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
dassaf4, line 10: This should be the only change in this generated file.

core/geometry/src/bspline/BezierCurve3d.ts
```diff
@@ -159,47 +161,46 @@ export class BezierCurve3d extends BezierCurveBase {
...
-      this.getPolePoint3d(order - 1, this._workPoint0);
-      rangeToExtend.extend(this._workPoint0);
-      for (let axisIndex = 0; axisIndex < 3; axisIndex++) {
-        BezierPolynomialAlgebra.componentDifference(bezier.coffs, this._polygon.packedData, 3, order, axisIndex);
-        const roots = bezier.roots(0.0, true);
-        if (roots) {
-          for (const r of roots) {
-            this.fractionToPoint(r, this._workPoint0);
-            rangeToExtend.extend(this._workPoint0);
+    if (this.allocateAndZeroBezierWorkData(order - 1, 0, 0)) {
```
dassaf4, line 164: 1. In a PR this massive it _really_ helps reviewers if you minimize the delta. Please disable all automatic formatting by your editor for this PR.
2. The only changes should involve the removal of the illegal non-null-assertion operators. This means undoing a lot of existing changes or starting a new PR.
3. For this method, you can avoid the superfluous indentation delta by returning early:

```suggestion
    if (!transform) {
      if (!this.allocateAndZeroBezierWorkData(order - 1, 0, 0))
        return;
      const bezier = ...
```

core/geometry/src/bspline/BezierCurveBase.ts
```diff
@@ -139,57 +153,70 @@ export abstract class BezierCurveBase extends CurvePrimitive {
...
-   * @param orderB length of _workCoffsB (simple array)
+   * @param orderA length of workCoffsA (simple array)
+   * @param orderB length of workCoffsB (simple array)
    */
-  protected allocateAndZeroBezierWorkData(primaryBezierOrder: number, orderA: number, orderB: number) {
+  protected allocateAndZeroBezierWorkData(
+    primaryBezierOrder: number, orderA: number, orderB: number,
+  ): this is { workBezier: UnivariateBezier, workCoffsA: Float64Array, workCoffsB: Float64Array } {
+    if (primaryBezierOrder <= 0 && orderA <= 0 && orderB <= 0)
+      return false;
```
dassaf4, line 200: To reduce chance of a future edit breaking the assumption that your early exit condition is the opposite of the logic that lies beyond it, replace this change with `let allocated = false;`, then inside each top-level if statement, set `allocated = true`. Finally return `allocated`.

core/geometry/src/bspline/BezierCurve3d.ts
```diff
@@ -159,47 +161,46 @@ export class BezierCurve3d extends BezierCurveBase {
...
-        for (let i = 0, k = 0; i < order; i++, k += 3)
-          componentCoffs[i] = transform.multiplyComponentXYZ(axisIndex, data[k], data[k + 1], data[k + 2]);
-        BezierPolynomialAlgebra.univariateDifference(componentCoffs, bezier.coffs);
-        const roots = bezier.roots(0.0, true);
-        if (roots && roots.length > 0) {
-          for (const r of roots) {
-            this.fractionToPoint(r, this._workPoint0);
-            rangeToExtend.extendTransformedPoint(transform, this._workPoint0);
+      } else {
+        this.allocateAndZeroBezierWorkData(order - 1, order, 0);
```
dassaf4, line 182: Similar to previous...

```suggestion
        if (!this.allocateAndZeroBezierWorkData(order - 1, order, 0))
          return;
```

core/geometry/src/clipping/ClipPlane.ts
```diff
@@ -287,16 +287,17 @@ export class ClipPlane extends Plane3d implements Clipper, PolygonClipper {
...
   public getPlane3d(): Plane3dByOriginAndUnitNormal {
     const d = this._distanceFromOrigin;
-    // normal should be normalized, will not return undefined
-    return Plane3dByOriginAndUnitNormal.create(
+    const plane = Plane3dByOriginAndUnitNormal.create(
       Point3d.create(this._inwardNormal.x * d, this._inwardNormal.y * d, this._inwardNormal.z * d),
       this._inwardNormal,
-    )!;
+    );
+    return plane || Plane3dByOriginAndUnitNormal.createXYPlane();
```
dassaf4, line 300: The nullish coalescing operator is more appropriate and readable.

```suggestion
    return plane ?? Plane3dByOriginAndUnitNormal.createXYPlane();
```

core/geometry/src/bspline/BezierCurve3dH.ts
```diff
@@ -219,51 +220,60 @@ export class BezierCurve3dH extends BezierCurveBase {
...
-      this.allocateAndZeroBezierWorkData(productOrder, 0, 0);
-      const bezier = this._workBezier!;
-      // closestPoint condition is:
-      //   (spacePoint - curvePoint) DOT curveTangent = 0;
-      // Each product (x,y,z) of the DOT is the product of two bezier polynomials
-      BezierPolynomialAlgebra.accumulateScaledShiftedComponentTimesComponentDelta(bezier.coffs, this._polygon.packedData, 4, this.order, 1.0, 0, -spacePoint.x, 0);
-      BezierPolynomialAlgebra.accumulateScaledShiftedComponentTimesComponentDelta(bezier.coffs, this._polygon.packedData, 4, this.order, 1.0, 1, -spacePoint.y, 1);
-      BezierPolynomialAlgebra.accumulateScaledShiftedComponentTimesComponentDelta(bezier.coffs, this._polygon.packedData, 4, this.order, 1.0, 2, -spacePoint.z, 2);
-      roots = bezier.roots(0.0, true);
+      if (this.allocateAndZeroBezierWorkData(productOrder, 0, 0)) {
```
dassaf4, line 233: Early out (false) if the typeguard fails.  This preserves the existing indentation to reduce delta.

core/geometry/src/bspline/BezierCurve3dH.ts
```diff
@@ -219,51 +220,60 @@ export class BezierCurve3dH extends BezierCurveBase {
...
-        // x representing loop pass:   (w * spacePoint.x - curve.x(s)) * (curveDelta.x(s) * curve.w(s) - curve.x(s) * curveDelta.w(s))
-        // (and p.w is always 1)
-        for (let k = 0; k < workA.length; k++)workA[k] = 0;
-        for (let k = 0; k < workB.length; k++)workB[k] = 0;
-        BezierPolynomialAlgebra.scaledComponentSum(workA, packedData, 4, orderA, 3, spacePoint.at(i), // w * spacePoint.x
-          i, -1.0); // curve.x(s) * 1.0
-        BezierPolynomialAlgebra.accumulateScaledShiftedComponentTimesComponentDelta(workB, packedData, 4, orderA, 1.0, 3, 0.0, i);
-        BezierPolynomialAlgebra.accumulateScaledShiftedComponentTimesComponentDelta(workB, packedData, 4, orderA, -1.0, i, 0.0, 3);
-        BezierPolynomialAlgebra.accumulateProduct(bezier.coffs, workA, workB);
+      if (this.allocateAndZeroBezierWorkData(productOrder, orderA, orderB)) {
```
dassaf4, line 252: Early out (false) if the typeguard fails. This preserves the existing indentation to reduce delta.

core/geometry/src/bspline/BezierCurveBase.ts
```diff
@@ -139,57 +153,70 @@ export abstract class BezierCurveBase extends CurvePrimitive {
...
-  /** Return the start point.  (first control point) */
+  /**
+   * Return the start point (first control point).
+   * @returns the start point as a Point3d. If insufficiently many control points exist, returns (0,0,0).
+   */
   public override startPoint(): Point3d {
-    const result = this.getPolePoint3d(0)!;   // ASSUME non-trivial pole set -- if null comes back, it bubbles out
-    return result;
+    const result = this.getPolePoint3d(0);
+    return result || Point3d.create();
```
dassaf4, line 162: The nullish coalescing operator is more appropriate and readable.
Using `createZero` has clearer intent.

```suggestion
    return result ?? Point3d.createZero();
```

core/geometry/src/bspline/BezierCurveBase.ts
```diff
@@ -139,57 +153,70 @@ export abstract class BezierCurveBase extends CurvePrimitive {
...
-  /** Return the end point.  (last control point) */
+  /**
+   * Return the end point (last control point).
+   * @returns the end point as a Point3d. If insufficiently many control points exist, returns (0,0,0).
+   */
   public override endPoint(): Point3d {
-    const result = this.getPolePoint3d(this.order - 1)!;    // ASSUME non-trivial pole set
-    return result;
+    const result = this.getPolePoint3d(this.order - 1);
+    return result || Point3d.create();
```
dassaf4, line 170: The nullish coalescing operator is more appropriate and readable.
Using `createZero` has clearer intent.
```suggestion
    return result ?? Point3d.createZero();
```

core/geometry/src/clipping/ClipPrimitive.ts
```diff
@@ -553,15 +558,18 @@ export class ClipShape extends ClipPrimitive {
...
       (ClipMaskXYZRangePlanes.None !== (clipMask & ClipMaskXYZRangePlanes.ZHigh)) ? high.z : undefined,
       transform,
       isMask,
       invisible,
       result,
-    )!;
+    );
+    // ClipShape.createShape only returns undefined if blockPoints.length < 3 which is not the case here
+    // so we shape is always defined and we never return the empty ClipShape
+    return shape || ClipShape.createEmpty();
```
dassaf4, line 572: The nullish coalescing operator is more appropriate and readable.

```suggestion
    return shape ?? ClipShape.createEmpty();
```

core/geometry/src/clipping/ClipPrimitive.ts
```diff
@@ -625,30 +633,56 @@ export class ClipShape extends ClipPrimitive {
     const convexSet = ConvexClipPlaneSet.createEmpty();
     if (cameraFocalLength === undefined) {
       const perpendicular = Vector2d.create(-normal.y, normal.x);
-      convexSet.planes.push(
-        ClipPlane.createNormalAndPoint(Vector3d.create(normal.x, normal.y), Point3d.createFrom(start), this._invisible)!,
+      let clipPlane = ClipPlane.createNormalAndPoint(
```
dassaf4, line 636: At the beginning of this method, the return value of normalize is ignored, allowing tiny nonzero magnitudes that result in normalization failure to slip through. I'd take this opportunity to correct. Otherwise, your changes for this method are good.
````
const normal = start.vectorTo(end);
if (!normal.normalize(normal))
  return false;
````

File: core/geometry/src/clipping/ConvexClipPlaneSet.ts [hunk 18]
```diff
@@ -236,15 +250,17 @@ export class ConvexClipPlaneSet implements Clipper, PolygonClipper {
...
     let y0 = points.getYAtUncheckedPointIndex(i0);
     let x1, y1, nx, ny;
     frustum._planes.length = 0;
     const z0 = points.getZAtUncheckedPointIndex(i0);  // z for planes can stay fixed
-    const planeNormal = points.crossProductIndexIndexIndex(0, 2, 1)!;
+    const planeNormal = points.crossProductIndexIndexIndex(0, 2, 1);
+    if (undefined === planeNormal)
+      return;
     ClipPlane.createNormalAndPointXYZXYZ(planeNormal.x, planeNormal.y, planeNormal.z, x0, y0, z0, false, false, planeOfPolygon);
     if (planeNormal.normalizeInPlace()) {
```
dassaf4, line 265: The `planeOfPolygon` should not be populated if the magnitude of `planeNormal` is too small. Below suggestion fixes this bug and removes the "!" at the same time.

```suggestion
    if (planeNormal?.normalizeInPlace()) {
      ClipPlane.createNormalAndPointXYZXYZ(planeNormal.x, planeNormal.y, planeNormal.z, x0, y0, z0, false, false, planeOfPolygon);
```

dassaf4 (commented):
Here's my review of the third commit...

PR #8438, opened 2025-08-18
Add FormatSetFormatsProvider to ecschema-metadata

hl662:
https://github.com/iTwin/viewer-components-react/pull/1404 adds support for customizing formats in the repo's test viewer. It uses concepts of FormatSet and FormatsProvider to achieve so, and it is increasingly likely this will be useful for other viewers. Thus, adding a `FormatSetFormatsProvider` interface to ecschema-metadata to support backends and frontends usage of this.

- Also, the formats Provider needs to be able to resolve colon separated schema items to period separated (schemaName:schemaItemName -> schemaName.schemaItemName) to address formats requested from native addon.

P.S: In the PR above, a FormatManager is used to coordinate system interaction between the FormatSetFormatProvider and potentially multiple formatSets. TBD if it should be moved into itwinjs-core, we'll see if there's demand for it to be consumed by both backends and frontends.

### DIFF
--- a/common/changes/@itwin/core-backend/2025-08-19-20-58.json
+++ b/common/changes/@itwin/core-backend/2025-08-19-20-58.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -18,8 +18,8 @@ importers:
   ../../core/backend:
     dependencies:
       '@bentley/imodeljs-native':
-        specifier: 5.2.5
-        version: 5.2.5
+        specifier: 5.2.6
+        version: 5.2.6
       '@itwin/cloud-agnostic-core':
         specifier: ^2.2.4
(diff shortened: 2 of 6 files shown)


core/ecschema-metadata/src/Formatting/FormatSetFormatsProvider.ts
```diff
@@ -0,0 +1,46 @@
...
+import { FormatDefinition, FormatsChangedArgs, FormatsProvider, MutableFormatsProvider } from "@itwin/core-quantity";
+import { FormatSet } from "../Deserialization/JsonProps";
+import { SchemaItem } from "../Metadata/SchemaItem";
+
+/**
+ * A mutable format provider that manages format definitions within a format set.
+ * When formats are added or removed, the underlying format set is automatically updated.
+ * @beta
+ */
+export class FormatSetFormatsProvider implements MutableFormatsProvider {
```
christophermlawson, line 11: Can you add docs to each public member please?  Thanks.

christophermlawson (commented):
Looks good. Just the comment about add docs to public members.

core/ecschema-metadata/src/Formatting/FormatSetFormatsProvider.ts
hl662: ok, I've updated the doccomments, some of which are in core-quantity, which this class extends from

christophermlawson (approved):

PR #8440, opened 2025-08-19
Field formatting

### LINKED ISSUE #8377: Text fields Property Formatting Reqs

pmconne:
WIP!

Fixes #8377.

### DIFF
--- a/core/backend/src/internal/annotations/fields.ts
+++ b/core/backend/src/internal/annotations/fields.ts
@@ -9,7 +9,7 @@ import { assert, DbResult, expectDefined, Id64String, Logger } from "@itwin/core
 import { BackendLoggerCategory } from "../../BackendLoggerCategory";
 import { XAndY, XYAndZ } from "@itwin/core-geometry";
 import { isITextAnnotation } from "../../annotations/ElementDrivesTextAnnotation";
-import { AnyClass, EntityClass, Property, StructArrayProperty } from "@itwin/ecschema-metadata";
+import { AnyClass, EntityClass, StructArrayProperty } from "@itwin/ecschema-metadata";
 
 // A FieldPropertyPath must ultimately resolve to one of these primitive types.
 export type FieldPrimitiveValue = boolean | number | string | Date | XAndY | XYAndZ | Uint8Array;
@@ -47,11 +47,11 @@ export interface FieldProperty {
 export interface UpdateFieldsContext {
   readonly hostElementId: Id64String;
 
-  getProperty(field: FieldRun): FieldProperty | undefined
+  getProperty(field: FieldRun): FieldPrimitiveValue | undefined
 }
 
 // Resolve the raw primitive value of the property that a field points to.
-function getFieldProperty(field: FieldRun, iModel: IModelDb): FieldProperty | undefined {
+function getFieldPropertyValue(field: FieldRun, iModel: IModelDb): FieldPrimitiveValue | undefined {
   const host = field.propertyHost;
   const schemaItem = iModel.schemaContext.getSchemaItemSync(host.schemaName, host.className);
   if (!EntityClass.isEntityClass(schemaItem)) {
@@ -216,15 +216,13 @@ function getFieldProperty(field: FieldRun, iModel: IModelDb): FieldProperty | un
     return undefined;
   }
 
-  return {
-    value: curValue.primitive,
-  };
+  return curValue.primitive;
 }
 
 export function createUpdateContext(hostElementId: string, iModel: IModelDb, deleted: boolean): UpdateFieldsContext {
   return {
     hostElementId,
-    getProperty: deleted ? () => undefined : (field) => getFieldProperty(field, iModel),
+    getProperty: deleted ? () => undefined : (field) => getFieldPropertyValue(field, iModel),
   };
 }
 
@@ -236,11 +234,11 @@ export function updateField(field: FieldRun, context: UpdateFieldsContext): bool
 
   let newContent: string | undefined;
   try {
-    const prop = context.getProperty(field);
-    if (undefined !== prop) {
+    const propValue = context.getProperty(field);
+    if (undefined !== propValue) {
       // ###TODO formatting etc.
       // eslint-disable-next-line @typescript-eslint/no-base-to-string
-      newContent = prop.value.toString();
+      newContent = propValue.toString();
     }
   } catch (err) {
     Logger.logException(BackendLoggerCategory.IModelDb, err);
--- a/core/backend/src/internal/annotations/fields.ts
+++ b/core/backend/src/internal/annotations/fields.ts
@@ -39,18 +39,9 @@ type FieldValue = {
   structArray: FieldStructValue[];
 }
 
-// Metadata associated with a FieldProperty, providing info needed for formatting like kind-of-quantity, extended type, etc.
-// That information can be obtained from the EC Property. For JSON fields, we will need to allow the user to specify it explicitly
-// (TBD because formatting is not yet implemented).
-export interface FieldPropertyMetadata {
-  readonly property: Property;
-  // ###TODO probably want to know if it's a JSON property.
-}
-
-// The resolved primitive value of a field with metadata.
+// The resolved primitive value of field.
 export interface FieldProperty {
   value: FieldPrimitiveValue;
-  metadata: FieldPropertyMetadata;
 }
 
 export interface UpdateFieldsContext {
@@ -227,7 +218,6 @@ function getFieldProperty(field: FieldRun, iModel: IModelDb): FieldProperty | un
 
   return {
     value: curValue.primitive,
-    metadata: { property: ecProp },
   };
 }
 
--- a/core/backend/src/internal/annotations/fields.ts
+++ b/core/backend/src/internal/annotations/fields.ts
@@ -3,16 +3,13 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
-import { ECSqlValueType, FieldRun, RelationshipProps, TextBlock } from "@itwin/core-common";
+import { ECSqlValueType, FieldPrimitiveValue, FieldRun, RelationshipProps, TextBlock } from "@itwin/core-common";
 import { IModelDb } from "../../IModelDb";
 import { assert, DbResult, expectDefined, Id64String, Logger } from "@itwin/core-bentley";
 import { BackendLoggerCategory } from "../../BackendLoggerCategory";
 import { XAndY, XYAndZ } from "@itwin/core-geometry";
 import { isITextAnnotation } from "../../annotations/ElementDrivesTextAnnotation";
 import { AnyClass, EntityClass, StructArrayProperty } from "@itwin/ecschema-metadata";
-
-// A FieldPropertyPath must ultimately resolve to one of these primitive types.
-export type FieldPrimitiveValue = boolean | number | string | Date | XAndY | XYAndZ | Uint8Array;
 
 interface FieldStructValue { [key: string]: any }
 
--- a/core/backend/src/internal/annotations/fields.ts
+++ b/core/backend/src/internal/annotations/fields.ts
@@ -3,7 +3,7 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
-import { ECSqlValueType, FieldPrimitiveValue, FieldRun, RelationshipProps, TextBlock } from "@itwin/core-common";
+import { ECSqlValueType, FieldPrimitiveValue, FieldRun, formatFieldValue, RelationshipProps, TextBlock } from "@itwin/core-common";
 import { IModelDb } from "../../IModelDb";
 import { assert, DbResult, expectDefined, Id64String, Logger } from "@itwin/core-bentley";
 import { BackendLoggerCategory } from "../../BackendLoggerCategory";
@@ -235,7 +235,7 @@ export function updateField(field: FieldRun, context: UpdateFieldsContext): bool
     if (undefined !== propValue) {
       // ###TODO formatting etc.
       // eslint-disable-next-line @typescript-eslint/no-base-to-string
-      newContent = propValue.toString();
+      newContent = formatFieldValue(propValue, field.propertyType, field.formatOptions);
     }
   } catch (err) {
     Logger.logException(BackendLoggerCategory.IModelDb, err);
--- a/core/backend/src/test/annotations/Fields.test.ts
+++ b/core/backend/src/test/annotations/Fields.test.ts
@@ -3,10 +3,10 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 import { expect } from "chai";
-import { Code, ElementAspectProps, FieldPropertyHost, FieldPropertyPath, FieldRun, PhysicalElementProps, SubCategoryAppearance, TextAnnotation, TextBlock, TextRun } from "@itwin/core-common";
+import { Code, ElementAspectProps, FieldPrimitiveValue, FieldPropertyHost, FieldPropertyPath, FieldRun, PhysicalElementProps, SubCategoryAppearance, TextAnnotation, TextBlock, TextRun } from "@itwin/core-common";
 import { IModelDb, StandaloneDb } from "../../IModelDb";
 import { IModelTestUtils } from "../IModelTestUtils";
-import { createUpdateContext, FieldPrimitiveValue, updateField, updateFields } from "../../internal/annotations/fields";
+import { createUpdateContext, updateField, updateFields } from "../../internal/annotations/fields";
 import { DbResult, Id64, Id64String } from "@itwin/core-bentley";
 import { SpatialCategory } from "../../Category";
 import { Point3d, XYAndZ, YawPitchRollAngles } from "@itwin/core-geometry";
--- a/core/backend/src/test/annotations/Fields.test.ts
+++ b/core/backend/src/test/annotations/Fields.test.ts
@@ -6,7 +6,7 @@ import { expect } from "chai";
 import { Code, ElementAspectProps, FieldPropertyHost, FieldPropertyPath, FieldRun, PhysicalElementProps, SubCategoryAppearance, TextAnnotation, TextBlock, TextRun } from "@itwin/core-common";
 import { IModelDb, StandaloneDb } from "../../IModelDb";
 import { IModelTestUtils } from "../IModelTestUtils";
-import { createUpdateContext, FieldProperty, updateField, updateFields } from "../../internal/annotations/fields";
+import { createUpdateContext, FieldPrimitiveValue, updateField, updateFields } from "../../internal/annotations/fields";
 import { DbResult, Id64, Id64String } from "@itwin/core-bentley";
 import { SpatialCategory } from "../../Category";
 import { Point3d, XYAndZ, YawPitchRollAngles } from "@itwin/core-geometry";
@@ -27,7 +27,7 @@ describe("updateField", () => {
 
   const createMockContext = (elementId: string, propertyValue?: string) => ({
     hostElementId: elementId,
-    getProperty: (field: FieldRun): FieldProperty | undefined => {
+    getProperty: (field: FieldRun): FieldPrimitiveValue | undefined => {
       const propertyPath = field.propertyPath;
       if (
         propertyPath.propertyName === "mockProperty" &&
@@ -35,7 +35,7 @@ describe("updateField", () => {
         propertyPath.accessors?.[1] === "nestedProperty" &&
         propertyValue !== undefined
       ) {
-        return { value: propertyValue };
+        return propertyValue;
       }
       return undefined;
     },
@@ -291,7 +291,7 @@ describe("Field evaluation", () => {
 
       const context = createUpdateContext(propertyHost.elementId, imodel, deletedDependency);
       const actual = context.getProperty(field);
-      expect(actual?.value).to.deep.equal(expected);
+      expect(actual).to.deep.equal(expected);
     }
 
     it("returns a primitive property value", () => {
--- a/core/backend/src/test/annotations/Fields.test.ts
+++ b/core/backend/src/test/annotations/Fields.test.ts
@@ -35,7 +35,7 @@ describe("updateField", () => {
         propertyPath.accessors?.[1] === "nestedProperty" &&
         propertyValue !== undefined
       ) {
-        return { value: propertyValue, metadata: {} as any };
+        return { value: propertyValue };
       }
       return undefined;
     },
--- a/core/common/src/annotation/TextBlock.ts
+++ b/core/common/src/annotation/TextBlock.ts
@@ -8,7 +8,7 @@
 
 import { Id64String } from "@itwin/core-bentley";
 import { TextStyleSettings, TextStyleSettingsProps } from "./TextStyle";
-import { FieldFormatter, FieldPropertyHost, FieldPropertyPath } from "./TextField";
+import { FieldFormatOptions, FieldPropertyHost, FieldPropertyPath } from "./TextField";
 
 /** Options supplied to [[TextBlockComponent.clearStyleOverrides]] to control how the style overrides are cleared on the component and its child components.
  * @beta
@@ -386,7 +386,7 @@ export interface FieldRunProps extends TextBlockComponentProps {
   /** Describes how to obtain the property value from [[propertyHost]]. */
   propertyPath: FieldPropertyPath;
   /** Specifies how to format the property value obtained from [[propertyPath]] into a string to be stored in [[cachedContent]]. */
-  formatter?: FieldFormatter;
+  formatter?: FieldFormatOptions;
   /** The field's most recently evaluated display string. */
   cachedContent?: string;
 }
@@ -411,7 +411,7 @@ export class FieldRun extends TextBlockComponent {
   /** Describes how to obtain the property value from [[propertyHost]]. */
   public readonly propertyPath: Readonly<FieldPropertyPath>;
   /** Specifies how to format the property value obtained from [[propertyPath]] into a string to be stored in [[cachedContent]]. */
-  public readonly formatter?: FieldFormatter;
+  public readonly formatter?: FieldFormatOptions;
   private _cachedContent: string;
 
   /** The field's most recently evaluated display string. */
--- a/core/common/src/annotation/TextBlock.ts
+++ b/core/common/src/annotation/TextBlock.ts
@@ -386,7 +386,7 @@ export interface FieldRunProps extends TextBlockComponentProps {
   /** Describes how to obtain the property value from [[propertyHost]]. */
   propertyPath: FieldPropertyPath;
   /** Specifies how to format the property value obtained from [[propertyPath]] into a string to be stored in [[cachedContent]]. */
-  formatter?: FieldFormatOptions;
+  formatOptions?: FieldFormatOptions;
   /** The field's most recently evaluated display string. */
   cachedContent?: string;
 }
@@ -411,7 +411,7 @@ export class FieldRun extends TextBlockComponent {
   /** Describes how to obtain the property value from [[propertyHost]]. */
   public readonly propertyPath: Readonly<FieldPropertyPath>;
   /** Specifies how to format the property value obtained from [[propertyPath]] into a string to be stored in [[cachedContent]]. */
-  public readonly formatter?: FieldFormatOptions;
+  public readonly formatOptions?: FieldFormatOptions;
   private _cachedContent: string;
 
   /** The field's most recently evaluated display string. */
@@ -430,7 +430,7 @@ export class FieldRun extends TextBlockComponent {
     this._cachedContent = props.cachedContent ?? FieldRun.invalidContentIndicator;
     this.propertyHost = props.propertyHost
     this.propertyPath = props.propertyPath;
-    this.formatter = props.formatter;
+    this.formatOptions = props.formatOptions;
   }
 
   /** Create a FieldRun from its JSON representation. */
@@ -439,7 +439,7 @@ export class FieldRun extends TextBlockComponent {
       ...props,
       propertyHost: { ...props.propertyHost },
       propertyPath: structuredClone(props.propertyPath),
-      formatter: structuredClone(props.formatter),
+      formatOptions: structuredClone(props.formatOptions),
     });
   }
 
@@ -456,8 +456,8 @@ export class FieldRun extends TextBlockComponent {
       json.cachedContent = this.cachedContent;
     }
 
-    if (this.formatter) {
-      json.formatter = structuredClone(this.formatter);
+    if (this.formatOptions) {
+      json.formatOptions = structuredClone(this.formatOptions);
     }
 
     return json;
@@ -508,12 +508,12 @@ export class FieldRun extends TextBlockComponent {
       return false;
     }
 
-    if (this.formatter && other.formatter) {
-      // ###TODO better comparison of formatter objects.
-      if (JSON.stringify(this.formatter) !== JSON.stringify(other.formatter)) {
+    if (this.formatOptions && other.formatOptions) {
+      // ###TODO better comparison of formatOptions objects.
+      if (JSON.stringify(this.formatOptions) !== JSON.stringify(other.formatOptions)) {
         return false;
       }
-    } else if (this.formatter || other.formatter) {
+    } else if (this.formatOptions || other.formatOptions) {
       return false;
     }
 
--- a/core/common/src/annotation/TextBlock.ts
+++ b/core/common/src/annotation/TextBlock.ts
@@ -8,6 +8,7 @@
 
 import { Id64String } from "@itwin/core-bentley";
 import { TextStyleSettings, TextStyleSettingsProps } from "./TextStyle";
+import { FieldFormatter, FieldPropertyHost, FieldPropertyPath } from "./TextField";
 
 /** Options supplied to [[TextBlockComponent.clearStyleOverrides]] to control how the style overrides are cleared on the component and its child components.
  * @beta
@@ -373,57 +374,6 @@ export class TabRun extends TextBlockComponent {
     return other instanceof TabRun && super.equals(other);
   }
 }
-
-/** A chain of property accesses that resolves to a primitive value that forms the basis of the displayed content
- * of a [[FieldRun]].
-   * The simplest property paths consist of a [[propertyName]] and nothing else, where `propertyName` identifies
-   * a primitive property.
-   * If `propertyName` identifies a struct or array property, then additional [[accessors]] are required to identify the specific value.
-   * If `propertyName` (including any [[accessors]]) resolves to a JSON property, then additional [[jsonAccessors]] are required to identify a specific value within the JSON.
-   * Some examples:
-   * ```
-   * | Access String | propertyName | accessors | jsonAccessors |
-   * | ------------- | ------------ | --------- | ------------- |
-   * | name          | "name"       | undefined | undefined     |
-   * | spouse.name   | "spouse"     | [name]    | undefined     |
-   * | colors[2]     | "colors"     | [2]       | undefined     |
-   * | spouse.favoriteRestaurants[1].address | "spouse" | ["favoriteRestaurants", 1, "address"] | undefined |
-   * | jsonProperties.contactInfo.email | "jsonProperties" | undefined | ["contactInfo", "email"] |
-   * | spouse.jsonProperties.contactInfo.phoneNumbers[0].areaCode | "spouse" | ["jsonProperties"] | ["contactInfo", "phoneNumbers", 0, "areaCode"] |
-   * ```
- * @beta
- */
-export interface FieldPropertyPath {
-  /** The name of the BIS property of the [[FieldPropertyHost]] that serves as the root of the path. */
-  propertyName: string;
-  /** Property names and/or array indices describing the path from [[propertyName]] to the ultimate BIS property. */
-  accessors?: Array<string | number>;
-  /** If [[propertyName]] and [[accessors]] (if defined) resolve to a BIS property of extended type `Json`, property names and/or
-   * array indices for selecting a primitive value within the JSON.
-   */
-  jsonAccessors?: Array<string | number>;
-}
-
-/** Describes the source of the property value against which a [[FieldPropertyPath]] is evaluated.
- * A field property is always hosted by an [Element]($backend). It may be a property of the element's BIS class itself,
- * or that of one of its [ElementAspect]($backend)s.
- * The [[schemaName]] and [[className]] should always identify the exact class that contains [[FieldPropertyPath.propertyName]] - not a subclass thereof.
- * @beta
- */
-export interface FieldPropertyHost {
-  /** The Id of the [Element]($backend) that hosts the property. */
-  elementId: Id64String;
-  /** The name of the schema containing the class identified by [[className]]. */
-  schemaName: string;
-  /** The name of the exact class (not a subclass) containing the property identified by [[FieldPropertyPath.propertyName]]. */
-  className: string;
-}
-
-/** Placeholder type for a description of how to format the raw property value resolved by a [[FieldPropertyPath]] into a [[FieldRun]]'s display string.
- * *** COMING SOON ***
- * @beta
- */
-export interface FieldFormatter { [k: string]: any }
 
 /** JSON representation of a [[FieldRun]].
  * @beta
--- a/core/common/src/annotation/TextBlock.ts
+++ b/core/common/src/annotation/TextBlock.ts
@@ -8,7 +8,7 @@
 
 import { Id64String } from "@itwin/core-bentley";
 import { TextStyleSettings, TextStyleSettingsProps } from "./TextStyle";
-import { FieldFormatOptions, FieldPropertyHost, FieldPropertyPath } from "./TextField";
+import { FieldFormatOptions, FieldPropertyHost, FieldPropertyPath, FieldPropertyType } from "./TextField";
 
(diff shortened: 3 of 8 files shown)


core/common/src/internal/annotations/FieldFormatter.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  return s;
+}
+
+function formatQuantity(v: FieldPrimitiveValue, _o?: QuantityFieldFormatOptions): string | undefined {
+  if (typeof v !== "number") {
+    return undefined;
+  }
+
+  // ###TODO apply quantity formatting...
```
pmconne: @DanRod1999 @hl662 this is where we need to plug in synchronous quantity formatting.

core/common/src/annotation/TextField.ts
```diff
@@ -0,0 +1,100 @@
...
+  falseString?: string;
+}
+
+export interface CoordinateFieldFormatOptions {
+  components?: CoordinateComponentSelector;
+  componentSeparator?: string;
+}
+
+export interface QuantityFieldFormatOptions {
+  // ###TODO source+target units, FormatProps.
```
pmconne: @DanRod1999 @hl662 here's where we need to persist whatever data we need for synchronous quantity formatting.

pmconne:
@grigasp how does presentation format date-time property values?

core/common/src/annotation/TextField.ts
DanRod1999: I don't think FormatProps/FormatDefinition is sufficient here, because in order to create a formatterSpec obj you need a Format. To create a Format you need to use async funcs like fromJson or createFromJson.

grigasp:
> how does presentation format date-time property values?

we're relying on this: https://github.com/iTwin/appui/blob/b5569ab605ce4b4e5c6831713a62ddc7b7e7fd2f/ui/components-react/src/components-react/converters/DateTimeTypeConverter.ts#L28

core/common/src/annotation/TextField.ts
pmconne: The whole point is that we will make it possible to create a `FormatterSpec` synchronously, and persist all the information needed to do that.
I added synchronous creation of `Format`. `FormatterSpec` also needs `UnitConversion`s. Can you add a new function to create a `FormatterSpec`, to which you supply all the required `UnitConversion`s? Then, make `QuantityFieldFormatOptions` persist all that info as JSON.

core/common/src/annotation/TextField.ts
```diff
@@ -0,0 +1,217 @@
...
+  suffix?: string;
+  case?: FieldCase;
+  boolean?: BooleanFieldFormatOptions;
+  coordinate?: CoordinateFieldFormatOptions;
+  quantity?: QuantityFieldFormatOptions;
+  dateTime?: DateTimeFieldFormatOptions;
+  enum?: EnumFieldFormatOptions<number> | EnumFieldFormatOptions<string>;
+}
+
+export function fieldFormatOptionsDeepEquals(opts1: FieldFormatOptions, opts2: FieldFormatOptions): boolean {
```
DanRod1999: got this from copilot. Not sure if this is a major improvement. Copilot also got lazy during QuantityFieldFormatOptions and just did JSON stringify comparisons. If this seems fine, I can create a deep equals for quantity field opts and a some tests. I guess in FieldFormatter.test.ts? Also Not sure if this file is correct place for these functions since its currently just interfaces

core/common/src/annotation/TextField.ts
```diff
@@ -0,0 +1,217 @@
...
+export interface FieldPropertyHost {
+  /** The Id of the [Element]($backend) that hosts the property. */
+  elementId: Id64String;
+  /** The name of the schema containing the class identified by [[className]]. */
+  schemaName: string;
+  /** The name of the exact class (not a subclass) containing the property identified by [[FieldPropertyPath.propertyName]]. */
+  className: string;
+}
+
+export type FieldPropertyType = "quantity" | "coordinate" | "string" | "boolean" | "datetime" | "int-enum" | "string-enum";
```
DanRod1999: does it make more sense to just add ` | string` here instead of in fieldRun.propertyType, or is it more readable to describe it at the end of propertyType

Josh-Schifter:
Checking to make sure I understand the approach here.  It looks like you are storing all the formatting options in the FieldRun.  I can't find the reference to the KOQ which is the proper source of all those options.  Is that implied from the FieldProperty?

I can see how there could be some special cases where a user might want to override the formatting options from the KOQ.  But, in the common case, storing all those options seems redundant.  Is there any provision for updating the formatting options in case of a schema upgrade that changes the KOQ?

pmconne:
We could store a KoQ name. The redundancy is motivated by our no-async-calls constraint. I would love to get in rid of it. Other redundancies are motivated by our need to produce localized strings (including unit labels) in a specific locale on the backend.

Josh-Schifter:
> We could store a KoQ name. The redundancy is motivated by our no-async-calls constraint. I would love to get in rid of it. Other redundancies are motivated by our need to produce localized strings (including unit labels) in a specific locale on the backend.

I think storing the KOQ name would capture important information.  I guess the redundancy is not good, but at least if we have the KOQ name then we haven't lost the source and we can recreate the formatting options when the source changes.

SheetalNair:
> We could store a KoQ name. The redundancy is motivated by our no-async-calls constraint. I would love to get in rid of it. Other redundancies are motivated by our need to produce localized strings (including unit labels) in a specific locale on the backend.

there was a point in the requirements stated here https://github.com/iTwin/itwinjs-core/issues/8377#issue-3277343506, that 
image
So to address the above use case, do you have plans to store koq name somewhere in the field to use it for formatting?
Because for dimensions, I believe we do not have any no-async-calls constraint, so storing redundant information  would not be necessary.

DanRod1999:
>  So to address the above use case, do you have plans to store koq name somewhere in the field to use it for formatting? Because for dimensions, I believe we do not have any no-async-calls constraint, so storing redundant information would not be necessary.

@SheetalNair  I added a field called koqName for now to our quantityFieldFormattingOptions. Its currently unused, but we should keep the same name and type for both of our uses, so let me know what you are doing so we can keep them the same.

test-apps/display-test-app/src/frontend/TextDecoration.ts
```diff
@@ -119,6 +123,102 @@ class TextEditor implements Decorator {
     }));
   }
 
+public async appendFieldRun(elementId: string, propName: string = "string"): Promise<void> {
+    if(this._iModel){
+      const element = (await this._iModel.elements.getProps(elementId))[0];
```
DanRod1999: This key-in only supports getting an elements code value and formatting as a normal string. Its a direct use case the drawing production team will need, but not sure if its an appropriate key-in for dta? To format other data we would need to pass in format options, or hard code preset format opts. @pmconne was the former your initial idea for this?

core/backend/src/internal/annotations/fields.ts
```diff
@@ -302,3 +284,71 @@ export function updateElementFields(props: RelationshipProps, iModel: IModelDb,
   }
 }
 
+// ### TODO: support extended types and custom attributes
+export function computeFieldPropertyType(property: Property): FieldPropertyType | undefined {
```
DanRod1999: Currently I am not handling extend types or custom attributes as I am not sure how to do this. For extended types I'm not entirely sure what the common one's are that might be relevant to use. I was told there is a list somewhere but couldn't find it. I tried this for DateTime and Guid, but haven't tried in tests

core/backend/src/internal/annotations/fields.ts
```diff
@@ -225,16 +207,17 @@ function getFieldProperty(field: FieldRun, iModel: IModelDb): FieldProperty | un
     return undefined;
   }
 
-  return {
-    value: curValue.primitive,
-    metadata: { property: ecProp },
-  };
+  if (field.propertyType !== computeFieldPropertyType(ecProp)){
```
DanRod1999: added this here, so we fail layout if these values differ, but not sure if this was the real intent for this func. In some of my tests I use this func to get the propertyType I want when creating the field. I'm not sure how users would get that value themselves. If we expect them to use the computeFieldPropertyType checking here would be redundant

core/common/src/internal/annotations/FieldFormatter.ts
```diff
@@ -0,0 +1,160 @@
...
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { XAndY, XYAndZ } from "@itwin/core-geometry";
+import { DateTimeFieldFormatOptions, EnumFieldFormatOptions, FieldFormatOptions, FieldPropertyType, QuantityFieldFormatOptions } from "../../annotation/TextField";
+import { Format, FormatterSpec } from "@itwin/core-quantity";
+
+// A FieldPropertyPath must ultimately resolve to one of these primitive types.
+export type FieldPrimitiveValue = boolean | number | string | Date | XAndY | XYAndZ | Uint8Array;
```
DanRod1999, line 12: When getting the date property from an element they are returned only as [strings](https://github.com/iTwin/itwinjs-core/blob/master/core/backend/src/ECSqlStatement.ts#L888). So when it comes time to format we actually do not pass in a Date object. Is it still necessary to include this type in the list of possible primitives?

DanRod1999: Also because of this I only check if a string is passed into the formatter (and the string must be converted to a valid date obj). This value is passed in from getFieldPropertyType(). I don't believe this function can return a date value in a different way but if it can then I would need to update the formatter to also allow date objs. My only guess is maybe extended types or custom attributes could cause this, but unsure?

DanRod1999:
@ColinKerr @aruniverse @Josh-Schifter, there were discussions about removing quantity support in fields last week. Just want to bring the convo here so we are on the same page. from what I understand, since fields currently do not have use for quantities,  not supporting it for now would be best to reduce complexity of fields. Please add any other important context if I missed it.

Josh and I talked a little about enums and how they would be used for displaying scaling, however my understanding is these enums, despite representing quantities, would not need to be formatted as such. Please correct me if I'm wrong Josh.

pmconne:
Formatting of distances etc is pretty fundamental even if @Josh-Schifter doesn't think it's needed for minimum viable product. We already implemented it. Why delete it?

Josh-Schifter:
I won't venture an opinion on removal of quantity formatting.  I can see both sides.

Some commentary on scales...  A scale is a unitless quantity.  For different industries and circumstances, I assume there are different conventions on how to format scales.  As our current focus is on site design in the US, consider that factors of 12 are commonly formatted as a relationship between inches and feet.  For example, the quantity 12 would be written as 1" = 1' and the quantity 48 would be written as 1/4" = 1'.

Matching this sort of convention for scales *is* needed for MVP.

pmconne:
> A scale is a unitless quantity. For different industries and circumstances, I assume there are different conventions on how to format scales.

Is this concept of "scale" as a particular kind of unitless quantity captured in the BIS schema, e.g., via `extendedTypeName`? @diegoalexdiaz?

Do you specifically mean "annotation scale" as opposed to other possible types of scales? MicroStation had 3 built-in "scale" extended types: Scale, NamedScale, and MultipleNamedScale.

diegoalexdiaz:
> Re: @pmconne - Is this concept of "scale" as a particular kind of unitless quantity captured in the BIS schema, e.g., via extendedTypeName? @diegoalexdiaz?

No - unitless concepts have been introduced in the Units schema under a different "Phenomenon" per case. e.g. 
- LENGTH_RATIO (pi as a constant based on it), 
- MASS_RATIO (Kg/Kg, gr/lbm), 
- PROBABILITY (%, fraction), 
- SLOPE (ft/ft, m/m, %)

The semantics behind "scale" might require us to add new Phenomenon with appropriate units. In the meantime, it will help to list the units that those scale-values from Microstation used.

cc: @ColinKerr

pmconne:
> it will help to list the units that those scale-values from Microstation used.

Someone with MicroStation installed, please find its BaseElementSchema and DgnElementSchema XML files and upload them here.
Extended types:
14 = Scale
19 = NamedScale
110 = MultipleNamedScale

https://la-solutions.org/CONNECT/MicroStationAPI/MicroStationAPI-EC-Element-Properties.htm

> Schema XML files are installed with MicroStation. You'll find them in folder ..\MicroStation\ECSchemas and its sub-folders. There are a lot of schemas: some more obscure than others. Those that concert DGN elements, for example, include …

>    Base Element Schema (BaseElementSchema.01.00.ecschema.xml)
>    DGN Element Schema (DgnElementSchema.01.00.ecschema.xml) 

> You will find many more schema files in the ..\MicroStation\ECSchemas folder.

pmconne:
[Here](https://github.com/user-attachments/files/22140838/mstn-schemas.zip) are a bunch of relevant MicroStation ECSchemas, including the two I mentioned.
Also here's [EditorCustomAttributes.01.03.ecschema.xml](https://github.com/user-attachments/files/22140902/EditorCustomAttributes.01.03.ecschema.xml)
All the "annotation scale" properties use NamedScale extended type. They are of type `double` which we try to match up to named scale values created by users or defined by apps. If it can't match up to a predefined scale name, it produces one using Josh's `ScaleDefinition::CustomNameFromFactor` function. @Josh-Schifter is that the kind of formatting you are looking for?

Josh-Schifter:
I'm trying to replace the term 'annotation scale' (from mstn) with 'drawing scale', but yes it's basically the same thing and yes that's what I'm talking about here.  LENGTH_RATIO (at least the name) does seem appropriate for this use.  It's the ratio of a length in the model to a length on the sheet.

I'm not familiar with the way EC tried to present scales in mstn and I've never heard of scale, named scale, or multipleNamedScale.

pmconne:
> I'm trying to replace the term 'annotation scale' (from mstn) with 'drawing scale', but yes it's basically the same thing and yes that's what I'm talking about here. LENGTH_RATIO (at least the name) does seem appropriate for this use. It's the ratio of a length in the model to a length on the sheet.

Ok, if I understand correctly, your drawing scale BIS property needs to specify these units and we need to determine if LENGTH_RATIO formatting is already implemented or if somebody needs to implement it. A quick grep suggests it's not implemented. If we need to implement it, we should do so in a separate PR. Obviously that would mean we do not want to remove quantity formatting from this PR, because you're going to need it after all.

diegoalexdiaz:
Microstation offers the following _annotation scales_ out of the box (the list keeps collapsing right before I can take a screenshot, so I'm forced to type it in):
1:1, 1:2, 1:2.5, 1:5, 1:10, 1:20, 1:25, 1:50, 1:100, 1:200, 1:250, 1:400, 1:500, 1:1000, 1:5000, 1:10000, 1:25000, 1:50000, 1:100000, 1:250000, "Custom".

Maybe this needs its own new Phenomenon with just a couple of Units (decimal and 1:X) that just drives formatting, showing the persisted factor (e.g. 0.5) in the 1:X format (e.g. 1:2). The list above could be introduced as "Constants" - another concept in the Units schema.

@ColinKerr, what are your thoughts?

Josh-Schifter:
In the most recent posts from Paul and Diego, I'm confused by the use of the word 'units'.

Paul wrote:
> your drawing scale BIS property needs to specify these units

Diego wrote:
> Maybe this needs its own new Phenomenon with just a couple of Units (decimal and 1:X) 

The way I see it, a scale is a 'unitless quantity'.  I get the debate about whether this is the same Phenomenon as LENGTH_RATIO.  But I don't get how there can be any units to specify.  Someone please help me with the terminology.

Josh-Schifter:
> 1:1, 1:2, 1:2.5, 1:5, 1:10, 1:20, 1:25, 1:50, 1:100, 1:200, 1:250, 1:400, 1:500, 1:1000, 1:5000, 1:10000, 1:25000, 1:50000, 1:100000, 1:250000, "Custom".

This list has a metric flavor to it.  Imperial scales tend to be factors of 12.  Ex. 1"=1' (12), 1/2"=1' (24), 1/4"=1' (48), 1/8"=1' (96).

diegoalexdiaz:
> I don't get how there can be any units to specify

The concept of Units in "unitless phenomena" is primarily about display-formatting. For instance, the PERCENT phenomenon has two units: decimal (e.g. 0.5 - as persisted) and percent (50%).

I'm proposing to apply the same pattern to annotation-scale: decimal (e.g. 0.5 - as persisted) or X:Y (e.g. 1:2). However, this proposal needs input from the EC team as it seems to require additional context on the formatting side to address SI vs. Imperial formatting.

core/backend/src/internal/annotations/fields.ts
pmconne: Fields are going to get copied from one iModel to another (example: drawing production copying sheet templates from catalog iModels into design iModels), in which case the schemas may differ. Verifying `computeFieldPropertyType` when we evaluate the field allows us to detect incompatible types instead of producing gibberish or throwing run-time exceptions.

core/common/src/internal/annotations/FieldFormatter.ts
pmconne: We should probably be parsing the string into a Date before handing it off to the formatter.

core/common/src/internal/annotations/FieldFormatter.ts
```diff
@@ -0,0 +1,160 @@
...
+  }else
+    return v.toString();
+}
+
+function formatDateTime(v: FieldPrimitiveValue, _o?: DateTimeFieldFormatOptions): string | undefined {
+  if (typeof v === "string") {
+    const date = new Date(v);
+    if (!isNaN(date.getTime())) {
+      if (_o && _o.formatOptions){
+        const formatter = new Intl.DateTimeFormat(_o.locale, _o.formatOptions);
```
pmconne: I considered and rejected using `Intl.DateTimeFormat` because you are [not guaranteed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/supportedLocalesOf) that the locale you want to use will be available for use. We need to produce consistently-formatted dates regardless of what system the code runs on. Did you find evidence that the `Intl.DateTimeFormat` implementation on every platform we support will always support every locale we might need?

DanRod1999: No, but I was told to only focus on supporting USA locale first since that is our target. I guess if that means we needed to change this later, pivoting to a new way to format would cause breaking changes. AppUI supports only a very limited version of date time formats (`toLocaleString()`, `toLocaleDateString()`) which can be supported by [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString#parameters), but I don't think we use it. So I think we just default to the host's locale. But I thought that's what Intl does anyway if you don't provided a supported locale? So my understanding was this should have similar support as AppUI, but just give us more formatting options, but maybe I'm missing something?

[I do think you can create your own locales though](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale)? Wouldn't the app be able to provide any locales they wish to support if it doesn't currently exist?

pmconne: The difference is that unlike AppUI we are persisting the formatted strings and our users will want them formatted using a consistent locale. The strings end up on sheets which are contractual documents. Defaulting to the host's locale won't work and supporting only en-US won't work long term.

A [few quick tests](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBANiYBDOBTCMC8MDaAiAIyTDwBoY8BLAEwFoBXW0WgB0rAE92yLrVaAkgBEeeCACdaAJQDKo1GFoBxAELzFAVTkBdANwAoUJFggWUSuAzYA3vEQpUAWSRRgAC1TiAXBQQgA1vQseDAAvjBIGEbQBtEgaAB0CADmABQCYFBwCUIuqAAqlAC2qABiIOJFLgkQQSwVUKjUADL2aBAA8gBmqQjI7eSm5pYAlCMGAPQTuFR0jMxsnNzkeHyCItr6hpYmZlYw1vowMADuqKj+1EgcPngIYMlkRzAcqEjeFGD0JeKUwE-HIrgKBuW4QNwNAEwK43T7fTx-J7hSIwaJQAz6LoVGCpNF2fqoGAgLozBS0LTqZRqFYSaRyFY0BhMECsdhcEgrNbCPDaEYHZ54rpFWDYMCoE4wDJZHJ5QolcqVFy9NqoQZ7cYCnbQvJYGBiiW5RqpABMAAYAIzG8gAVg1xziiRSqSFUASWMVUFSV0aYwMoS2+iAA) show my Firefox install supports a variety of locales. If we **can** use `Intl` then we should, but I'd like to have more confidence than "it works on my machine". How can we find out?

NOTE: I hesitate to say we should **store** `Intl.DateTimeFormat`-style formatting options because that ties us to a particular technology. We made the same exact mistake in MicroStation by storing .NET-style format strings and had to invent our own parser and formatter when we dropped our dependency on .NET. iModels are supposed to be consumable in a broad variety of environments; persisting ECMAScript-specific format strings impedes that and dictates a specific technology for consumers.

Ideally, we would store something technology-agnostic (a la the proposed "date-time format parts") that would be easy to convert to an input to `Intl.DateTimeFormat` so we don't have to write our own formatting logic.

DanRod1999: I saw that node is supported by intl which is what we'll always be using right? It does have some caveats though in terms of support compared to other browsers, but if I'm reading it correctly thats only pre node 13. I did some tests on my local node env and I had support for a range of locales, which indicates we use full-icu which supports many (100s according to copilot) different locales. If someone is not using full-icu support then that could cause problems, but it seems like its currently the default for node versions above 13

DanRod1999: > How can we find out?

we could also always add a test to make sure we support the correct locales we need, but not sure if that's necessary? It would basically just be checking the version of node we are running in has icu support, which doesn't seem all that useful. I'm having trouble finding a full list of supported locales but its based on [Unicodes CLDR ](https://cldr.unicode.org/)

pmconne: If we move field formatting from core-common to core-backend then yes, we only have to worry about node. That seems acceptable - a front-end UI that wants to show a formatted preview would just have to ask the backend to produce it.

It [looks like](https://nodejs.org/api/intl.html#options-for-building-nodejs) full ICU = full Intl support. So I think we can use Intl for formatting, but store something more generic than an `Intl.DateTimeFormat` options object.

However, we compile node ourselves for the mobile platforms and I don't **think** that has full ICU. We're not doing authoring on mobile devices currently but we may end up with a function that doesn't produce consistent results across platforms.

core/common/src/internal/annotations/FieldFormatter.ts
DanRod1999: done [here](https://github.com/iTwin/itwinjs-core/pull/8440/files#diff-3b82c2d0b4ae7924c5a95f455c0de5e0b2f0ee68483611f9902d86bbcb67193cR78)

core/backend/src/test/annotations/Fields.test.ts
```diff
@@ -358,7 +390,8 @@ describe("Field evaluation", () => {
       expectValue(undefined, { propertyName: "outerStructs", accessors: [0, "innerStruct"] }, sourceElementId);
     });
 
-    it("returns arbitrarily-nested properties of structs and struct arrays", () => {
+    // FIXME: structs should return undefined and fail evaluation?
+    it.skip("returns arbitrarily-nested properties of structs and struct arrays", () => {
```
DanRod1999: This test and one other (check for fixme in this file) now fail because computePropertyType returns undefined for any structs. Given the current compute prop type logic I think these tests are now invalid, but wanted to verify first before deleting

pmconne: Can you provide the actual failure message please?
The test attempts to set up a field pointing to `outerStruct.innerStruct.bool` which is a primitive property. `computePropertyType` should not fail for a boolean property.

DanRod1999: 1) Field evaluation
       getProperty
         returns arbitrarily-nested properties of structs and struct arrays:
     AssertionError: expected undefined to deeply equal false
      at expectValue (src\test\annotations\Fields.test.ts:322:30)
      at Context.<anonymous> (src\test\annotations\Fields.test.ts:395:7)
      at processImmediate (node:internal/timers:478:21)

  2) Field evaluation
       ElementDrivesTextAnnotation
         supports complex property paths:

      AssertionError: expected '####duck' to equal '2duck'

pmconne: I rewrote `computeFieldPropertyType`. You have to evaluate the whole property path, not just the root `Property`.

core/common/src/annotation/TextField.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export interface CoordinateFieldFormatOptions {
+  components?: CoordinateComponentSelector;
+  componentSeparator?: string;
+}
+
+export interface QuantityFieldFormatOptions {
+  koqName?: string;
+  sourceUnit?: UnitProps;
+  targetUnit?: UnitProps;
```
pmconne: @DanRod1999 `QuantityFieldFormatOptions` is never used. Is it supposed to be?

pmconne: AFAICT it's redundant with `formatProps`. Deleting.

pmconne:
@ColinKerr @rschili @hl662 @aruniverse @ben-polinsky @markschlosseratbentley please review.

core/quantity/src/Formatter/Interfaces.ts
```diff
@@ -10,6 +10,19 @@ import { BeEvent } from "@itwin/core-bentley";
 import { UnitProps } from "../Interfaces";
 import { DecimalPrecision, FormatTraits, FormatType, FractionalPrecision } from "./FormatEnums";
 
+/** [[FormatProps.composite]] lacks documentation, please add a description here of what this represents.
```
aruniverse: need to update docs
@hl662

core/common/package.json
```diff
@@ -42,12 +42,16 @@
   },
   "peerDependencies": {
     "@itwin/core-bentley": "workspace:*",
-    "@itwin/core-geometry": "workspace:*"
+    "@itwin/core-geometry": "workspace:*",
+    "@itwin/core-quantity": "workspace:*"
```
aruniverse: this is technically breaking; can we make this optional?

+@hl662

core/common/src/annotation/TextField.ts
```diff
@@ -0,0 +1,179 @@
+/*----------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Annotation
+ */
+
+import { Id64String } from "@itwin/core-bentley";
+import { ResolvedFormatProps, UnitConversionSpec, UnitProps } from "@itwin/core-quantity";
```
aruniverse:
```suggestion
import type { ResolvedFormatProps, UnitConversionSpec, UnitProps } from "@itwin/core-quantity";
```

aruniverse: feel free to ignore this; added this to help trace what is actually adding core-qty as a peer, and its the import in the FieldFormatter

core/common/src/annotation/TextBlock.ts
```diff
@@ -435,16 +385,22 @@ export interface FieldRunProps extends TextBlockComponentProps {
   propertyHost: FieldPropertyHost;
   /** Describes how to obtain the property value from [[propertyHost]]. */
   propertyPath: FieldPropertyPath;
-  /** Specifies how to format the property value obtained from [[propertyPath]] into a string to be stored in [[cachedContent]]. */
-  formatter?: FieldFormatter;
-  /** The field's most recently evaluated display string. */
+  /** Specifies how to format the property value obtained from [[propertyPath]] into a string to be stored in [[cachedContent]].
+   * The specific options used depend upon the field's [[propertyType]].
+   */
+  formatOptions?: FieldFormatOptions;
```
ColinKerr, line 390: It looks like this means the full format object plus the resolved unit conversion factors, etc will be stored in the field run.  I thought we were just going to store the KOQ as a key to look up the format info at runtime when needed to format a value.  I don't see anything wrong with caching this info in memory but it seems wrong to store in the run

pmconne: We cannot look up kinds of quantity synchronously. If/when we ever get that ability, or escape our constraint that field evaluation must occur in a synchronous context, we can stop caching.

core/common/src/annotation/TextField.ts
```diff
@@ -0,0 +1,179 @@
...
+  /** Specifies which component(s) (X, Y, and/or Z) to include in the display string.
+   * Default: "XYZ".
+   * @note For two-dimensional points, Z is always omitted even if specified here.
+   */
+  components?: CoordinateComponentSelector;
+  /** The string to insert between components.
+   * Default: "," (comma).
+   */
+  componentSeparator?: string;
+}
```
ColinKerr: Do you expect that any shared default is defined in project props somewhere or that they will be entered for each field if customized?

pmconne: Don't know. Ultimately wherever the strings came from they get stored here. User could type them in, or select from a localized pick-list, or whatever. I don't want to invent some system where we store the strings elsewhere and refer to them indirectly here.

pmconne:
@Josh-Schifter and @claudiaareneee wanted some key-ins or tools in display-test-app to add fields, in part so they can verify the API meets their requirements. TBD whether we hold up this PR for that or add them in a follow-up PR. I think a follow-up PR makes more sense given they don't involve formatting and this PR is already pretty self-contained. We can revise relevant APIs based on feedback from implementing those key-ins.

core/common/src/annotation/TextField.ts
pmconne: Yeah, `import type` is always better where appropriate (failing to use it often angers vite when building the test apps).

core/common/package.json
pmconne: You tell me.

pmconne:
@aruniverse @hl662 @ColinKerr FYI I'm blocked on #8534 until this merges.

claudiaareneee:
> @aruniverse @hl662 @ColinKerr FYI I'm blocked on #8534 until this merges.

Agreed, I can't validate the correctness until I can test it with the keyins.

pmconne:
Test failed on iOS, presumably due to lack of support for `Intl` being unsupported in the mobile backend (see similar issue in TextBlock.test.ts). I will disable the test on mobile since we don't for the foreseeable future intend to support authoring on mobile.

```
  1) Field evaluation
       Format Validation
         validates formatting options for datetime objects:

      AssertionError: expected '####' to equal 'Aug 28, 2025'
      + expected - actual

      -####
      +Aug 28, 2025
      
      at Context.<anonymous> (main.js:145506:85)
      at callFn (main.js:366815:21)
      at Runnable.run (main.js:366803:5)
      at Runner.runTest (main.js:367613:10)
      at /Users/platform-bld/Library/Developer/CoreSimulator/Devices/2315999E-DB6F-4347-AE99-A7630F97FC37/data/Containers/Bundle/Application/57246C30-DB44-4B56-A587-FA3E7AB5A05D/core-test-runner.app/Assets/main.js:367736:12
      at next (main.js:367528:14)
      at /Users/platform-bld/Library/Developer/CoreSimulator/Devices/2315999E-DB6F-4347-AE99-A7630F97FC37/data/Containers/Bundle/Application/57246C30-DB44-4B56-A587-FA3E7AB5A05D/core-test-runner.app/Assets/main.js:367538:7
      at next (main.js:367421:14)
      at Immediate._onImmediate (main.js:367506:5)
      at process.processImmediate (node:internal/timers:491:21)
```

core/common/package.json
hl662: I don't think it can be optional, and if text fields needs concepts of formatting, including [KoQs](https://github.com/iTwin/itwinjs-core/issues/8377#issuecomment-3201582705), it's time to include core quantity in there

core/common/src/annotation/TextBlock.ts
ColinKerr: I don't have a problem with caching them in memory.  The problem is that, as written they will be persisted in the iModel.  @Alfonso-Martello 

WRT synch/async what we need to do is make it possible to do this async instead of making more things sync.  I know that is a long term project but we need to start.

core/quantity/src/Formatter/Interfaces.ts
hl662: added a commit to clarify docs

core/common/src/annotation/TextField.ts
ColinKerr: @Josh-Schifter  or @Alfonso-Martello  where do you plan to grab this kind of formatting information from?

ColinKerr: Another example is where you would get enum formatting info from.  Display labels are defined in the schema but also they are stored in the FieldFormatOptions

pmconne: Yes, because we must do the formatting on the backend where we have no localization and certainly don't have every possible locale we would need. I am fully aware of the redundancies. We have no current or near-term solution for it.

core/common/src/annotation/TextBlock.ts
aruniverse: Just to note, we _were_ working on a sync qty api: https://github.com/iTwin/itwinjs-core/pull/8386, but it was abandoned because I thought we were going to do this async and not store within the iModel

pmconne: If you're asking me to make changes to this PR, please be specific about what changes you want. #8386 was abandoned because we needed only a subset of quantity formatting to be synchronously accessible and didn't want to propagate sync throughout the public API.

core/common/package.json
pmconne: So what if anything needs to change here?

pmconne:
@dassaf4 FYI this has failed the same way twice on Windows only, surely unrelated to my changes.
```
--[ FAILURE: @itwin/core-geometry ]----------------[ 3 minutes 42.0 seconds ]--

⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/test/polyface/PolyfaceQuery.test.ts > PolyfaceSpeedup > AnnounceBoundaryEdges

- Expected
+ Received

- 0
  ...2 lines omitted...
 ❯ src/test/polyface/PolyfaceQuery.test.ts:1553:31
    1551| 
    1552|     GeometryCoreTestIO.saveGeometry(allGeometry, "PolyfaceSpeedup", "A…
    1553|     expect(ck.getNumErrors()).toBe(0);
    1554|   });
    1555| 
```

core/common/src/annotation/TextField.ts
pmconne: What if anything do you want changed here @ColinKerr?

dassaf4:
> @dassaf4 FYI this has failed the same way twice on Windows only, surely unrelated to my changes.

I've seen this once last week on my box. It verifies a performance improvement, but it seems sensitive to processor load. I'll disable: https://github.com/iTwin/itwinjs-core/pull/8538

core/common/package.json
hl662: Nothing to change here, probably just document in NextVer.md that core quantity is now a peer dependency for core common, and explain that its cause backends are now doing formatting

pmconne: Added.

aruniverse: This is still breaking, we shouldnt be doing this within the 5.x major.

Using the following command, I got the number of pkgs in this monorepo that has a peer on core-common:
```
find . -name "package.json" -exec sh -c 'jq -e ".peerDependencies | has(\"@itwin/core-common\")" "{}" > /dev/null && echo "{}"' \; | wc -l
```

There are 26 pkgs in this monorepo alone that have a peer on core-common that will now be affected by this change. Of the 26, only 5 of them today already have peers on core-quantity, calculcated via:

```
find . -name "package.json" -exec sh -c 'jq -e ".peerDependencies | (has(\"@itwin/core-common\") and has(\"@itwin/core-quantity\"))" "{}" > /dev/null && echo "{}"' \; |  wc -l
```
Therefore 21 pkgs, now need to add a dev dep on core-quantity since its a transitive peer.  We don't know how many packages outside of our repo will be affected, but I'm sure it will be a magnitude larger. Every single one of these packages now needs to add a dev-dep on core-quantity because otherwise you will have duplicate installs of core-common. This is due to how package managers resolve peers.

So while _"text fields needs concepts of formatting, including https://github.com/iTwin/itwinjs-core/issues/8377#issuecomment-3201582705"_ and _"it's time to include core quantity in there"_  this will be massively disrupting.

There is an argument to be made that anyone building a client-side or backend app today probably already pulls in core-quantity, itll cause dopplegangers due to middleware packages not being updated.

pmconne: Field formatting APIs can be moved to core-backend if you are ok with introducing a peer dep on core-quantity there instead.

Otherwise, tell me what you want to happen here.

aruniverse: attempting to make it optional in [45895f5](https://github.com/iTwin/itwinjs-core/pull/8440/commits/45895f5850eb63a51e5aff4f503ee1703d7c786d); fingers crossed this works as this would be the easiest solution

core/common/src/annotation/TextBlock.ts
ColinKerr: In the future we will support async change propagation, but not in a timeframe that lets us merge this quickly.

I suggest we clearly identify the values we are caching in field run options as a cache by naming them appropriately and documenting that they are a cache and should be overridden by the current settings as they are changed.

Discussion of text field formatting requirements is here: https://github.com/iTwin/itwinjs-core/issues/8377#issuecomment-3197960019

The information necessary for unitized numeric formatting and enums clearly fits into the cache category.  Boolean display labels are less clear because there must be some central configuration for the standard options for true/false, yes/no, etc but the current implementation could support arbitrary true/false strings.

I would suggest a clear naming convention for each option which is a cache or to group the options which are caches into a single group which is clearly identified as a cache.

If there are options that are not immediately needed it would be ideal if they were not persisted by default (some kind of feature flag?).

pmconne: deleted all the formatting types we don't immediately need, to buy time to figure out the async field formatting problem and/or the sync quantity formatting problem.

core/common/package.json
pmconne: removed the core-quantity dependency entirely.

docs/changehistory/NextVersion.md
```diff
@@ -16,6 +16,7 @@ publish: false
     - [@itwin/presentation-frontend](#itwinpresentation-frontend)
   - [Display](#display)
     - [Draco decoding](#draco-decoding)
+  - [New core-common peer dependency](#new-core-common-peer-dependency)
```
aruniverse: lets revert this

core/common/package.json
```diff
@@ -48,6 +48,8 @@
     "@itwin/build-tools": "workspace:*",
     "@itwin/core-bentley": "workspace:*",
     "@itwin/core-geometry": "workspace:*",
+    "@itwin/ecschema-metadata": "workspace:*",
```
aruniverse: ecschema metadata no longer looks to be used

core/common/package.json
```diff
@@ -48,6 +48,8 @@
...
+    "@itwin/ecschema-locaters": "workspace:*",
```
aruniverse: neither is ecschema-locaters

core/common/src/annotation/TextField.ts
```diff
@@ -0,0 +1,109 @@
+/*----------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Annotation
+ */
+
+import { Id64String } from "@itwin/core-bentley";
+import type { ResolvedFormatProps, UnitConversionSpec, UnitProps } from "@itwin/core-quantity";
```
aruniverse:
```suggestion
```

aruniverse (commented):
just need to clean up vestigial usage of core-qty

@hl662, plz re-review. while the changes to core-qty in this pr no longer does anything of value, i dont think we need to discard changes to it if we need to ressurect it later

common/config/rush/browser-approved-packages.json
```diff
@@ -220,7 +220,7 @@
     },
     {
       "name": "@itwin/ecschema-locaters",
-      "allowedCategories": [ "backend", "internal", "tools" ]
+      "allowedCategories": [ "backend", "common", "internal", "tools" ]
```
hl662: revert this if ecschema-locaters is no longer used

core/common/src/annotation/TextField.ts
pmconne: deleted.

PR #8445, opened 2025-08-20
Field format quantities

DanRod1999:
wip to be part of https://github.com/iTwin/itwinjs-core/pull/8440


core/common/package.json
```diff
@@ -48,6 +48,9 @@
     "@itwin/build-tools": "workspace:*",
     "@itwin/core-bentley": "workspace:*",
     "@itwin/core-geometry": "workspace:*",
+    "@itwin/ecschema-metadata": "workspace:*",
+    "@itwin/ecschema-locaters": "workspace:*",
+    "@itwin/core-quantity": "workspace:*",
```
pmconne, line 54: core-quantity is a regular dependency, not a dev dependency, right? It's used by FieldFormatter.ts; the others are only used by tests.

core/common/src/annotation/TextField.ts
```diff
@@ -71,6 +72,10 @@ export interface CoordinateFieldFormatOptions {
 
 export interface QuantityFieldFormatOptions {
   // ###TODO source+target units, FormatProps.
+  sourceUnit: UnitProps;
```
pmconne: We want to be able to format unitless quantities too. I *think* that's supported.

core/common/package.json
pmconne: After we merge your changes we should consider moving `formatFieldValue` to core-backend and avoid introducing a core-common dependency on core-quantity. Currently nobody has a use case for formatting field values outside the backend, but you could imagine a field editor UI on frontend that wants to show a preview of the formatted value.

PR #8449, opened 2025-08-21
Avoid CDN for Draco decoding

markschlosseratbentley:
This PR changes Draco decoding on iTwin.js so it does not use a CDN to request decoding source files. Instead, it bundles those into iTwin.js from a new `draco3d` dependency and asks the `loaders.gl` library to locally use those resources.

Possible concerns:
- How much does this increase the iTwin.js package size? @aruniverse @hl662 could you possibly measure this?
- This approach still uses `loaders.gl` - this PR configures `loaders.gl` to use an imported version of `draco3d` with iTwin.js-bundled resources (notably the decoding wasm).  We could in theory just bundle and execute the proper js / wasm functions ourselves directly and forgo `loaders.gl`.  [See this documentation](https://loaders.gl/docs/modules/draco/). Note, however, that in this PR we are passing `draco3d` into `DracoLoader.parse` which is less documented.
- Related to the above, it is unclear if we are bundling the IE11 draco3d fallback properly - I was unsure how to test this. @aruniverse @hl662?

Here is a Draco-encoded GLB file for testing: [CesiumManDraco.zip](https://github.com/user-attachments/files/21917391/CesiumManDraco.zip)

To test this branch:

1. Build `itwinjs-core/test-apps/display-test-app`
2. `cd test-apps/display-test-app`
3. `npm run start`
4. Open a blank connection:
image

6. In the key-in field in the top left, type in `dta gltf` and press enter
7. In the file dialog, open up `CesiumManDraco.glb`
8. You should see the local Draco resources be requested over the network tab in the Electron dev tools.
9. You should be able to find the model loaded in the view somewhere in the following area of this parking lot: 
image
Screenshot 2025-08-20 132246

Below are screenshots showing this local Draco-decoding source retrieval occur successfully using this branch.

Screenshot 2025-08-20 132027

Screenshot 2025-08-20 132139


pmconne:
Why do we care about IE11 decoding? We don't support IE at all.

markschlosseratbentley:
> Why do we care about IE11 decoding? We don't support IE at all.

You're right - We don't care directly. However, `draco3d` / `loaders.gl` cares and might try to do something different if it detects IE11.

pmconne:
Don't our apps literally pop up an error dialog and terminate when run inside IE? Couldn't you throw an error if you detect IE before invoking draco?

markschlosseratbentley:
> Don't our apps literally pop up an error dialog and terminate when run inside IE? Couldn't you throw an error if you detect IE before invoking draco?

Not a bad idea. We'll have to dig up that dialog code. EDIT: https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/public/scripts/checkbrowser.js

markschlosseratbentley:
> > Don't our apps literally pop up an error dialog and terminate when run inside IE? Couldn't you throw an error if you detect IE before invoking draco?
> 
> Not a bad idea. We'll have to dig up that dialog code. EDIT: https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/public/scripts/checkbrowser.js

@pmconne @aruniverse I added a check to throw an error if IE/old Edge is detected, largely based on that checkbrowser.js example.  This should suffice for a number of the concerns outlined in the PR description - I will edit it.

core/frontend/src/tile/GltfReader.ts
```diff
@@ -2123,6 +2123,11 @@ export abstract class GltfReader {
       return;
 
     try {
+      // refuse to continue decoding if using Internet Explorer or old Microsoft Edge (lifted from checkbrowser.js).
+      if (!!(document as any).documentMode || !!(window as any).StyleMedia) {
```
pmconne: Consider moving to a function in ProcessDetector.

markschlosseratbentley: As suggested, I added an IE-detection function in ProcessDetector and am now calling it from GltfReader.ts instead of this inline code.

hl662:
> How much does this increase the iTwin.js package size?

I ran an A/B of the compiled output from core-frontend, and DTA. The difference is 300 kb

markschlosseratbentley:
> > How much does this increase the iTwin.js package size?
> 
> I ran an A/B of the compiled output from core-frontend, and DTA. The difference is 300 kb

This seems reasonable to me. If anyone has a differing opinion, feel free to voice it.

common/changes/@itwin/core-frontend/markschlosser-draco-security-fix_2025-08-20-17-57.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "",
```
aruniverse: i would call out this change in the changelog even if its an internal detail
as well as mention it in nextver.md

core/frontend/src/tile/GltfReader.ts
```diff
@@ -2123,8 +2123,15 @@ export abstract class GltfReader {
       return;
 
     try {
+      // Refuse to continue decoding if using Internet Explorer or old Microsoft Edge. We do not want to trigger any legacy decoding fallbacks within draco3d.
+      if (ProcessDetector.isIEBrowser) {
```
aruniverse, line 2127: seems odd we need to add this at all here since we dont support ie at all anyway.

core/frontend/package.json
```diff
@@ -98,6 +100,7 @@
     "@itwin/webgl-compatibility": "workspace:*",
     "@loaders.gl/core": "^3.1.6",
     "@loaders.gl/draco": "^3.1.6",
```
aruniverse: @markschlosseratbentley we should look at updating our deps on these loaders.gl deps since we are a major ver behind

core/frontend/src/tile/GltfReader.ts
markschlosseratbentley: Can you direct me to code in iTwin.js that guarantees it will not run in IE?  I found checkbrowser.js, but I think it was only consumed by DTA and DPTA.

ben-polinsky:
> > How much does this increase the iTwin.js package size?
> 
> I ran an A/B of the compiled output from core-frontend, and DTA. The difference is 300 kb

What's the package size pre-inclusion?

hl662:
> What's the package size pre-inclusion?

For core frontend:
- 27.7 mb for lib/esm
- lib/public is 1.2 mb (which is where the draco worker is put)

For DTA: 
- the assets folder that is where core-frontend is bundled by vite, is 35516 kb.

After checking out this branch, core-frontend's lib/esm went up by 300 kb, lib/public went up by 300kb, and DTA's assets folder went up to 35816 (so a difference of 300kb)

core/frontend/package.json
markschlosseratbentley: I agree. In reaction to this comment, I just upgraded them to the latest and tested the draco code path locally to verify draco decoding still functions properly.

common/changes/@itwin/core-frontend/markschlosser-draco-security-fix_2025-08-20-17-57.json
markschlosseratbentley: Sure. I added a NextVersion entry and also fleshed out the changelog.

markschlosseratbentley:
Image Tests are ongoing. I will edit this comment with results.

- [x] Image Tests begun.
- [x] Image Tests finished.
- [x] Image Tests look okay.

core/frontend/src/tile/GltfReader.ts
aruniverse: If you're using vite, you need to copy static assets, and we show how to do so in our examples; [see](https://github.com/iTwin/viewer/blob/master/packages/templates/web/vite.config.mts#L15-L24). That should also copy the `checkbrowser.js` script, its up to the dev to add that script to `index.html`. So looks like we don't explicitly disallow IE in  iTwin.js, but if im not mistaken we now only support webgl2 as of `4.0` and IE doesnt even support webgl2. If anything we should add a check in `itwin/webgl-compatibility`?

aruniverse (approved):
approving; feel free to close comment about IE if its too nit picky

core/frontend/package.json
markschlosseratbentley: @aruniverse If you are satisfied with these final changes to package.json, could you resolve this comment?

common/changes/@itwin/core-frontend/markschlosser-draco-security-fix_2025-08-20-17-57.json
markschlosseratbentley: @aruniverse If you are satisfied with the new NextVersion entry, could you resolve this comment?

hl662:
Rush change is flagging core bentley for missing a changefile, can you run that again @markschlosseratbentley ?

core/frontend/src/tile/GltfReader.ts
markschlosseratbentley: Okay, yes that makes sense. I do not disagree - you make a valid point about WebGL2.  I think this warrants a larger investigation eventually outside this PR as to when iTwin.js will keep running in certain situations and make these checks more consistent. For now, I feel most comfortable leaving the check in place "just in case" iTwin.js **_somehow_** gets in a state that this code gets hit in IE11. It seems better to not rely on a side effect of IE11 not supporting WebGL2 as well -- in the future, this code will make more sense than without it.  (At the minimum, this should at least warrant a comment.)

pmconne: The default `RenderSystem` uses webgl-compatibility but apps can override it, so you're not guaranteed here that we've already safeguarded against IE. I think just leave the check here; ultimately we're going to delete all this code anyway and let CesiumJS handle glTF decoding.

markschlosseratbentley: Resolving this comment per this https://github.com/iTwin/itwinjs-core/pull/8449#pullrequestreview-3177037829 by @aruniverse. If we somehow circumvent the WebGL2 rendering system, it could be possible to hit this code, so IMO let's be careful and leave this check in place until we can do a more thorough analysis of circumstances under which iTwin.js should be allowed to run --- independent of WebGL2 checks.

markschlosseratbentley:
@hl662 Could I get a second set of eyes on the pnpm-lock.yaml "rush update" changes on this PR after the recent merge?  I had to hit it with a hammer to make it happy and the process resulted in a large diff for that file. I just want to make sure all is as expected. Extract-api CI is passing, at least.

hl662:
I tried running rush update locally, all seems fine, if there was a problem GH would have alerted a merge conflict in pnpm-lock.yaml

markschlosseratbentley:
> I tried running rush update locally, all seems fine, if there was a problem GH would have alerted a merge conflict in pnpm-lock.yaml

Thank you. 🙏

markschlosseratbentley:
CI `rush cover` failure is [occurring](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4272557&view=logs&j=607f17b7-2b33-5225-6b64-00ff0e2e25e4&t=89ec18ca-7724-51ff-9cd4-fafaa006e1ca&l=12391) (regarding path polyfills for draco).

I had to do something in [the webpack here](https://github.com/iTwin/itwinjs-core/pull/8449/files#diff-b299572f4fa5c3d88f49062c2f5c18daef1fb676a1a574d07023291abd6e3660L23) to resolve a similar error in itwinjs-core when building DTA.  Perhaps we need something similar for tests?

```
--[ FAILURE: @itwin/core-frontend ]-------------------------[ 13.00 seconds ]--

ERROR in ../../common/temp/node_modules/.pnpm/draco3d@1.5.1/node_modules/draco3d/draco_decoder_nodejs.js 32:191-214
Module not found: Error: Can't resolve 'path' in '/Volumes/Builds/vsts_work_a/4/s/common/temp/node_modules/.pnpm/draco3d@1.5.1/node_modules/draco3d'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "path": require.resolve("path-browserify") }'
	- install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
  ...43 lines omitted...
 @ ../../common/temp/node_modules/.pnpm/draco3d@1.5.1/node_modules/draco3d/draco3d.js 5:26-59
 @ ./lib/esm/tile/GltfReader.js 1729:35-52
 @ ./lib/esm/tile/internal.js 49:0-29 49:0-29
 @ ./lib/esm/common/imdl/ParseImdlDocument.js 19:0-56 227:41-58
 @ ./lib/esm/common/internal/render/GraphicDescriptionBuilderImpl.js 17:0-64 164:38-54
 @ ./lib/esm/common/render/GraphicDescriptionBuilder.js 8:0-137 18:15-53 27:19-48
 @ ./lib/esm/common.js 10:0-58 10:0-58
 @ ./lib/esm/test/worker/test-worker.js 6:0-71 39:24-56 40:18-35

webpack compiled with 4 errors

```

pmconne:
> I had to do something in [the webpack here](https://github.com/iTwin/itwinjs-core/pull/8449/files#diff-b299572f4fa5c3d88f49062c2f5c18daef1fb676a1a574d07023291abd6e3660L23) to resolve a similar error in itwinjs-core when building DTA. Perhaps we need something similar for tests?

Probably more important to find out if apps are going to run into similar issues...that would be bad.

markschlosseratbentley:
> > I had to do something in [the webpack here](https://github.com/iTwin/itwinjs-core/pull/8449/files#diff-b299572f4fa5c3d88f49062c2f5c18daef1fb676a1a574d07023291abd6e3660L23) to resolve a similar error in itwinjs-core when building DTA. Perhaps we need something similar for tests?
> 
> Probably more important to find out if apps are going to run into similar issues...that would be bad.

@aruniverse @hl662 Is there a simple way to test this before we publish an npm package of core-frontend? Any tips / help very welcome so we can feel confident pushing this out.

hl662:
> > > I had to do something in [the webpack here](https://github.com/iTwin/itwinjs-core/pull/8449/files#diff-b299572f4fa5c3d88f49062c2f5c18daef1fb676a1a574d07023291abd6e3660L23) to resolve a similar error in itwinjs-core when building DTA. Perhaps we need something similar for tests?
> > 
> > 
> > Probably more important to find out if apps are going to run into similar issues...that would be bad.
> 
> @aruniverse @hl662 Is there a simple way to test this before we publish an npm package of core-frontend? Any tips / help very welcome so we can feel confident pushing this out.

Using [cospaces](https://github.com/aruniverse/itwin-cospace) is one option, you can also use `pnpm patch` in a separate repo, and replace the compiled builds for core bentley and core frontend. The viewer [repo](https://github.com/iTwin/viewer) is a good example, or you can just create a new web app by following [our startup instructions](https://www.itwinjs.org/learning/tutorials/develop-web-viewer/), which uses the same thing in the viewer repo, then run pnpm patch @itwin/core-bentley and etc

markschlosseratbentley:
@hl662 @aruniverse FYI - Yesterday this was all fixed so `rush build` and `rush cover` both succeeded. Then the latest merge to master happened (today).  Now `rush build` is [failing at trying to build core-full-stack-tests](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4278813&view=logs&j=8673a5cb-f235-5320-c755-005ead49efbb&t=f3b75961-d101-5755-bff1-b7fcef1a1191&l=693). Scratching my head trying to figure out what the merge did beyond changing dependency versions.  Based on the error, perhaps a package was introduced that expects to use `path`?  If I revert back to [commit `fix rush cover`](https://github.com/iTwin/itwinjs-core/pull/8449/commits/35e7e32c5d2ef9c4d2cc6a057baf2e4e98d73628) or go off this branch and onto the tip of master, the build succeeds.

markschlosseratbentley:
@aruniverse @hl662 No further progress made on this today. Based on the "Module not found" error in building core-full-stack-tests, I believe a problem is happening with some code (AzureServerStorage.js 23:23-53) trying to import `@azure/storage-blob/dist-esm/storage-blob/src/index.browser.js` but it is seemingly not exported in the storage-blob's package.json.

This seems unrelated to my draco changes, so I'm a bit baffled right now.

Any assistance appreciated.

[Link to the build failure on CI.](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4278813&view=logs&j=8673a5cb-f235-5320-c755-005ead49efbb&t=f3b75961-d101-5755-bff1-b7fcef1a1191&l=693)

```
ERROR in ../../common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/object-storage-azure/lib/server/AzureServerStorage.js 23:23-53
Module not found: Error: Package path ./dist-esm/storage-blob/src/index.browser.js is not exported from package D:\code\itwinjs-core\common\temp\node_modules\.pnpm\@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13\node_modules\@azure\storage-blob (see exports field in D:\code\itwinjs-core\common\temp\node_modules\.pnpm\@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13\node_modules\@azure\storage-blob\package.json)
 @ ../../common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/object-storage-azure/lib/server/index.js 23:13-44
 @ ../../common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/object-storage-azure/lib/index.js 23:13-32
 @ ../../common/temp/node_modules/.pnpm/@itwin+imodels-client-authoring@5.8.1_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/imodels-client-authoring/lib/IModelsClient.js 11:31-69
 @ ../../common/temp/node_modules/.pnpm/@itwin+imodels-client-authoring@5.8.1_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/imodels-client-authoring/lib/index.js 25:13-39
 @ ./lib/frontend/hub/ITwinPlatformEnv.js 5:35-77
 @ ./lib/frontend/TestUtility.js 17:27-60
 @ ./lib/frontend/standalone/WatchForChanges.test.js 15:22-47
```

aruniverse:
> @aruniverse @hl662 No further progress made on this today.

resolved

aruniverse:
> Please do not merge until we've determined whether and how apps will need to react to this change (ideally they won't).

I'm not worried about this breaking apps because of missing node-polyfills, and if so we can find out as soon as this pr merges and a new nightly build is available.

pmconne:
> I'm not worried about this breaking apps because of missing node-polyfills, and if so we can find out as soon as this pr merges and a new nightly build is available.

OK, as long as you're not worried about it.

pmconne (approved):

markschlosseratbentley:
> > @aruniverse @hl662 No further progress made on this today.
> 
> resolved

Thank you, @aruniverse! 🙏

jjbeckman13:
I see Draco loading from the local host, but I cannot get the Reality Data to show.

image

iTwin Experience - [Link](https://qa-itwinexperience.bentley.com/itwin/375ae030-508c-4226-a3a7-684356e68dd6/model?sourceApp=1&realityDataId=9c510f2c-0e5f-485c-b32a-1049cec45bff)
image

LocalHost Pineapple - [Link](http://localhost:3000/context/375ae030-508c-4226-a3a7-684356e68dd6?local=false&realityDataId=9c510f2c-0e5f-485c-b32a-1049cec45bff)

image

The other Cesium Reality Data works
image

cc: @MarcBedard8, @jason-crow, @aruniverse

markschlosseratbentley:
@jjbeckman13 In the case where it is not working - this also did not work before, right?
Any warnings / errors in the console?

jjbeckman13:
> @jjbeckman13 In the case where it is not working - this also did not work before, right? Any warnings / errors in the console?

@markschlosseratbentley, It works in iTwinJs 5.1.5
[Link](https://qa-review-uks.itwin.bentley.com/context/375ae030-508c-4226-a3a7-684356e68dd6?local=false&realityDataId=9c510f2c-0e5f-485c-b32a-1049cec45bff)

image

jjbeckman13:
@markschlosseratbentley and I did a debug session to find that the drao_decoder.wasm isn't being loaded. Its looking in the wrong location on the server. 

@aruniverse, @hl662, @jason-crow, do you recall anything that would have been done to make this work for parse-imdl-worker.js. I couldn't find anything in our bundler config for this. I do see it referencing being loaded from ${IModelApp.publicPath}scripts/parse-imdl-worker.js in ImdlParser.ts.

image

pmconne:
> do you recall anything that would have been done to make this work for parse-imdl-worker.js. I couldn't find anything in our bundler config for this.

You had to copy the worker from core-frontend's public assets to your public path. [Lots of people](https://github.com/iTwin/itwinjs-core/issues?q=is%3Aissue%20parse-imdl-worker) had the same problem with that worker script. @aruniverse this was my concern when I talked about wanting to avoid breaking apps.

aruniverse:
Seems to be a similar issue under the hood to customizability of the loaders like in https://github.com/visgl/loaders.gl/issues/3185

@jjbeckman13 I would try on your end copying the wasm file to the root of your server / webapp. 

@markschlosseratbentley if we cant fix this we should add a note to nextver.md (cc @hl662 )
Also, digging into the docs a bit more, feel like we should avoid passing in the entirety of the draco3d library and just need to provide the `draco_wasm_wrapper.js` and `draco_decoder.wasm` files. Unfortunately as we see in the issue above that doesnt seem to work today because of a bug in the loader itself.

> @aruniverse this was my concern when I talked about wanting to avoid breaking apps.

Yes, this was valid, and luckily we caught this in a pre-release build. Testing this out via pnpm patch or cospace wouldve been tricky. We have workaround we'll investigate before official release. Thank you justin for testing.

markschlosseratbentley:
> feel like we should avoid passing in the entirety of the draco3d library and just need to provide the `draco_wasm_wrapper.js` and `draco_decoder.wasm` files. Unfortunately as we see in the issue above that doesnt seem to work today because of a bug in the loader itself

As you note, this more granular approach does not work for `loaders.gl`. This aligns with my testing and observations. Further, it seems to be a limit of the API we use - the `loaders.gl` API we are using (`DracoLoader.parse()`) only meaningfully accepts the `draco3d` library. Thankfully, based on @hl662's measurements, the package size did not overly grow with this package introduction (the wasm and wrapper js are the bulk already, I believe).

PR #8453, opened 2025-08-21
Add ECDb deleteSchemaItems method and corresponding tests

BinhL3:
imodel-native: https://github.com/iTwin/imodel-native/pull/1199


core/backend/src/ECDb.ts
```diff
@@ -235,6 +235,20 @@ export class ECDb implements Disposable {
...
 
+  /**
+   * Delete items from a schema.
+   * @param schemaName The name of the schema from which to delete items.
+   * @param schemaItemNames The names of the items to delete.
+   * @alpha
+   */
+  public deleteSchemaItems(schemaName: string, schemaItemNames: string[]) {
+    try {
+      this[_nativeDb].deleteSchemaItems(schemaName, schemaItemNames, { schemaLockHeld: true });
```
hl662: shouldn't the schemaLockHeld be checked prior to this? Don't think hard passing in `true` is a good idea

Issue #7701, opened 2025-02-13
Presentation: Deprecate hierarchies-related APIs


grigasp:
The APIs should be deprecated in favor of the ones in `@itwin/presentation-hierarchies` package.

PR #8463, opened 2025-08-28
Presentation: Deprecate hierarchy-related APIs

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/7701


presentation/backend/eslint.config.cjs
```diff
@@ -9,10 +9,11 @@ module.exports = [
   },
   ...eslintBaseConfig,
   {
-    files: ["**/*.{ts,tsx}"],
+    files: ["**/*.ts"],
     rules: {
       ...prettierConfig.rules,
       curly: ["error", "all"],
+      "@typescript-eslint/no-non-null-assertion": "off",
```
tcobbs-bentley, line 16: It seems that this should be set to warn instead of being turned off completely. I was mostly ok with turning this off for the tests, but for the real code, I don't think it's a good idea.

presentation/common/eslint.config.cjs
```diff
@@ -9,10 +9,11 @@ module.exports = [
   },
   ...eslintBaseConfig,
   {
-    files: ["**/*.{ts,tsx}"],
+    files: ["**/*.ts"],
     rules: {
       ...prettierConfig.rules,
       curly: ["error", "all"],
+      "@typescript-eslint/no-non-null-assertion": "off",
```
tcobbs-bentley, line 16: Same as with backend; this seems like a bad idea.

presentation/frontend/eslint.config.cjs
```diff
@@ -9,10 +9,11 @@ module.exports = [
   },
   ...eslintBaseConfig,
   {
-    files: ["**/*.{ts,tsx}"],
+    files: ["**/*.ts"],
     rules: {
       ...prettierConfig.rules,
       curly: ["error", "all"],
+      "@typescript-eslint/no-non-null-assertion": "off",
```
tcobbs-bentley, line 16: Also seems like a bad idea.

presentation/backend/eslint.config.cjs
MichaelSwigerAtBentley: I agree that this can be disabled for test files. Its definitely unnecessary to check if a variable is undefined before asserting that its 'equal to 4' or something.

MichaelSwigerAtBentley: This eslint rule should only not apply to `.test.ts` files correct? We definitely should not disable it for code

MichaelSwigerAtBentley (approved):

presentation/backend/eslint.config.cjs
grigasp: IMO, that rule is a piece of crap. TS already gives you an error if you try to use a nullable value, and developer has to use `!` to override that. Why do we need a lint rule on top of that?.. All this rule gives is a bunch of warnings that pollute lint output and makes real problems hard to see (in my case there were 4 warnings for actual problems hidden between tens of these non-null-assertion ones).

presentation/common/eslint.config.cjs
grigasp: Replied with https://github.com/iTwin/itwinjs-core/pull/8463#discussion_r2309180546

presentation/frontend/eslint.config.cjs
grigasp: Replied with https://github.com/iTwin/itwinjs-core/pull/8463#discussion_r2309180546

presentation/backend/eslint.config.cjs
grigasp: BTW, I disabled this rule only for presentation packages, since our team doesn't add `!` if we're not certain the value can't be null. Other packages will still require the eslint-disable comment.

tcobbs-bentley: Having just gone through a bunch of code that was causing a warning due the recent enabling of this rule, I can state that `!` get used a **lot** in places where it absolutely should not be used. The rule makes it so that you have to actively add the disable comment any time you use `!`, which hopefully means that you think about what you are doing (and ideally find a better way). But I don't work on the presentation packages, so I can't force you to keep the rule enabled.

tcobbs-bentley (approved):

presentation/backend/eslint.config.cjs
grigasp: > I can state that `!` get used a **lot** in places where it absolutely should not be used

I can't disagree with that, definitely saw that happening. But it should not be the norm, and in @iTwin/itwinjs-core-presentation team we're trying not to use it where it shouldn't be used.

> The rule makes it so that you have to actively add the disable comment any time you use `!`, which hopefully means that you think about what you are doing (and ideally find a better way).

I have a couple of comments about this:

- If a developer adds `!` to overcome the TS error without thinking, what makes you think that requiring to add an extra `eslint-disable` comment would change anything and prevent him/her to follow the same pattern - just add the comment and not think about consequences?

  We discussed about this lint rule internally in the team, and agreed that the comment makes it easier for reviewers to notice the red flag. But this only applies if the reviewers actually look through he changes thoroughly.

- I'd argue that the rule, as it's currently set up today, has absolutely no value, and maybe is even harmful. It doesn't cause a lint error, so developers are NOT forced to add the eslint-disable comment, making the rule useless. On the other hand, if you run the `lint` script locally and get a bunch of warnings - they make other, more important, warnings harder to notice. IMO, if you think the rule is valuable in your packages, it should at least be an error rather than a warning.

grigasp: I'm resolving this comment to not block the pr.

PR #8465, opened 2025-08-28
Add Unit System property to FormatSet interface

hl662:
- Added  `unitSystem` property to [FormatSet]($ecschema-metadata) interface, using [UnitSystemKey]($quantity) type. This will help move APIs away from relying on `activeUnitSystem` in `quantityFormatter`, as they move to the new formatting APIs using `IModelApp.formatsProvider`. Looking ahead, tools and components that use formatting APIs can then listen to just the `onFormatsChanged` event from `IModelApp.formatsProvider` instead of `IModelApp.quantityFormatter.onActiveUnitSystemChanged`.
- Change type for formats in `FormatSet` from `SchemaItemFormatProps` to `FormatDefinition`

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -2777,6 +2777,7 @@ export class DynamicsContext extends RenderContext {
     // @internal (undocumented)
     add(graphic: RenderGraphic, isOverlay: boolean): void;
     addGraphic(graphic: RenderGraphic): void;
+    addOverlay(graphic: RenderGraphic): void;
     // @internal (undocumented)
     changeDynamics(): void;
     createGraphic(options: Omit<ViewportGraphicBuilderOptions, "viewport">): GraphicBuilder;
--- a/common/changes/@itwin/core-backend/2025-09-02-15-25.json
+++ b/common/changes/@itwin/core-backend/2025-09-02-15-25.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/changes/@itwin/core-frontend/pmc-dynamic-overlay-api_2025-09-02-13-24.json
+++ b/common/changes/@itwin/core-frontend/pmc-dynamic-overlay-api_2025-09-02-13-24.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Add DynamicsContext.addOverlay",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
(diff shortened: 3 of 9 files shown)


core/ecschema-metadata/src/Deserialization/JsonProps.ts
```diff
@@ -290,11 +290,13 @@ export type SchemaItemFormatProps = SchemaItemProps & FormatProps;
...
  * This interface defines properties necessary to support persistence of a set of formats.
- * @beta
+ * @public @preview
  */
 export interface FormatSet {
   name: string;
   label: string;
+  /** A [UnitSystemKey]($quantity) that determines the unit system for this format set. */
+  unitSystem: UnitSystemKey;
   formats: { [kindOfQuantityId: string]: SchemaItemFormatProps };
```
MichaelSwigerAtBentley: If this is going to public from beta, can we get a comment for formats too?

common/api/ecschema-metadata.api.md
```diff
@@ -795,16 +795,14 @@ export class Format extends SchemaItem {
     get uomSeparator(): string;
 }
 
-// @beta
+// @public @preview
```
ColinKerr: Be careful once this goes public preview you cannot add new required parameters.

I cannot think of another thing we need to add but I would old off until we've gotten some feedback from someone designing and using their own format set.

core/ecschema-metadata/src/Deserialization/JsonProps.ts
```diff
@@ -290,11 +290,16 @@ export type SchemaItemFormatProps = SchemaItemProps & FormatProps;
...
- * @beta
+ * @public @preview
  */
 export interface FormatSet {
+  /** The unique name identifier for this format set. */
   name: string;
+  /** The display label for this format set. */
   label: string;
+  /** A [UnitSystemKey]($quantity) that determines the unit system for this format set. */
+  unitSystem: UnitSystemKey;
```
ColinKerr, line 301: UnitSystemKey is still in beta.  We should review the systems listed in this type then move it to public preview at the same time we move the FormatSet interface

docs/learning/quantity/index.md
```diff
@@ -284,6 +289,46 @@ When retrieving a format from a schema, users might want to ensure the format th
...
+  const formatterSpec = await FormatterSpec.create(
+    formatDef.name,
+    formatDef,
+    new SchemaUnitsProvider(schemaContext), // Use a schema units provider
+    persistenceUnit
+  );
+
+  const formattedValue = formatterSpec.applyFormatting(value);
+  console.log(`Formatted value: ${formattedValue}`);
+}
```
ColinKerr: This code should be harvested from a test so we ensure it continues to build.

hl662: i've moved the code to example-code-snippet, so it's now part of a test

common/api/ecschema-metadata.api.md
hl662: I've changed the interface back to beta

core/ecschema-metadata/src/Deserialization/JsonProps.ts
hl662: I've changed the FormatSet interface back to beta, so we won't be moving UnitSystemKey to public preview

PR #8470, opened 2025-08-29
Move Deprecation Date Commit in Version Bump Pipeline

andremig-bentley:
Currently, the deprecation date commit occurs after the version bump commit in the version bump pipeline, leading to a confusing commit history where the most recent commits after every bump are applying the deprecation date rule (as seen [here](https://github.com/iTwin/itwinjs-core/commits/release/5.1.x)). This PR solves this issue by splitting the deprecation date rule application workflow in to two parts and integrating it into the version bump process.

Now, we begin by applying the deprecation date rule, and if any changes are present, we attempt to merge them in to master to check for merge conflicts. If no problems are present (which is the most likely outcome), then we  move on to do the version bump after the deprecation date, leading to a much clearer commit history where the deprecation date changes are wrapped up within the version bump commit. If we fail to merge the deprecation date changes in to master (most likely due to to merge conflicts), then we do not commit the deprecation date changes. Instead,  we move on to the version bump, and after the version bump, we create a PR for the deprecation date changes to be manually validated. In this case, the deprecation date commit will occur after the version bump, but the version bump would not be blocked by a failed deprecation date commit which would happen if all deprecation date logic was moved to entirely before the version bump.

To test these changes, I've created a [fork](https://github.com/andremig-bentley/itwinjs-core-fork) of itwinjs. This PR is still WIP due to testing on the fork for all possible scenarios. I will be continually updating this PR with more info on what's been tested as it happens until it is ready to publish. 

The method of solving the original issue, as in how exactly we move the deprecation date logic, is still up for discussion. Any alternative ideas or questions are appreciated.


andremig-bentley:
Both patch version bumps and minor version bumps with and without necessary deprecation date changes have been tested and are working as planned. Fork link HERE.

On the fork's 5.2.x [branch](https://github.com/iTwin/itwinjs-core/compare/master...andremig-bentley:itwinjs-core-fork:release/5.2.x), we can see two patches, 5.2.6 and 5.2.7, both containing changes to deprecation date comments. The first screenshot below shows the commit history now shows these commits as just "5.2.6" and "5.2.7", and the next two show that those commits successfully include the deprecation date changes.

patchhistory
5.2.x Commit History


5.2.6 Commit with Deprecation Date Changes


5.2.7 Commit with Deprecation Date Changes

Then, on the fork's 5.3.x [branch](https://github.com/iTwin/itwinjs-core/compare/master...andremig-bentley:itwinjs-core-fork:release/5.3.x), we can see a minor version bump to 5.3.0, again containing changes to deprecation date comments. The first screenshot below shows the commit history now shows this commit as just "5.3.0" and the second shows that this commit successfully includes the deprecation date changes.

minorhistory
5.3.x Commit History


5.3.0 Commit with Deprecation Date Changes

At this point, the final thing to test before this PR is ready to publish is intentionally creating merge conflicts in the deprecation date comments in order to test the manual validation workflow.

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -169,6 +169,67 @@ jobs:
...
+              continueOnError: true
+              env:
+                GH_TOKEN: $(GITHUBTOKEN)
+              name: applyDeprecation
+
+            - bash: git checkout $(deprecationCommit.hash)
+              displayName: Checkout back to branch that ran the pipeline
+
       # Can be run on any branch to do a standard version bump. Which is currently -dev bumps.
       - ? ${{ if or(eq(parameters.BumpType, 'nightly'), eq(parameters.BumpType, 'releaseCandidate')) }}
```
MichaelSwigerAtBentley, line 231: Have you also tested this doesn't mess with nightlys?

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -169,6 +169,67 @@ jobs:
       - bash: node common/scripts/install-run-rush.js audit
         displayName: rush audit
 
+      - ${{ if or(eq(parameters.BumpType, 'minor'), eq(parameters.BumpType, 'patch')) }}:
```
MichaelSwigerAtBentley, line 175: Previously, this was separated into another file. Have you experimented with bundling this into a side script for cleanliness? I'm ok if its not in a separate file, but if its easy, it might be worth it to keep the version-bump clean

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -239,10 +300,32 @@ jobs:
         displayName: Commit version bump
 
       - ${{ if or(eq(parameters.BumpType, 'minor'), eq(parameters.BumpType, 'patch')) }}:
-        - template: "../templates/run-eslint-deprecation.yaml"
+          - ${{ if eq(variables['needsPR'], 'true') }}:
```
MichaelSwigerAtBentley: Definitely check what happens when this logic runs. I don't think any of us have seen what happens when there is a conflict. This conflict would stop the version bump too correct?

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: I've been testing with nightlys on the fork, from what I can tell it's working for the most part, but for some reason the "dev" is not appearing in the name, shown in the screenshot below. This does mean that the check previous commit step doesn't work as expected since it's looking for "dev" in the name of the previous commit. I don't think this is an issue with the actual logic I've added and might be something wrong with how I've set up the fork, but looking in to it to make sure.

nightly

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: The main reason I have not separated this into a side script of some sort was the need to pass along the "needsPR" value until after the version bump commit. This is necessary as long as we want to keep the manual validation fail safe, and there seem to be ways to pass a value like that along through env variables, but it might not be as straight forward. If we did not keep the fail safe (which is discussed in my comment below), then this would be a simple clean up, but for now I'll refrain from making changes until we decide on whether or not we're keeping the manual validation step.

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: I don't think this conflict would stop the version bump, because the check for conflict is separate from the version bump logic and then opens a PR after the version bump commit. However, I haven't been able to actually confirm this, because I can't seem to create a scenario in which a merge conflict is present.

The only thing that would create this merge conflict would be for the release branch and master to have conflicting changes on the same deprecation date comment, something I attempted to recreate between these two commits [here](https://github.com/iTwin/itwinjs-core/commit/137cf51ffe1662097fa70c5d1da024b321fcc604) on master, and [here](https://github.com/iTwin/itwinjs-core/commit/15692df4fd58fa1d1192c6af06dd163476888c02) on the release branch for example. In both commits, `Viewport.ts` on line 1602 had the deprecation date comment altered in the same spot before the version bump was run and the changes to the release branch ensured that the deprecation date step in the version bump would be triggered. However, no conflicts happened when the bump was run because the commit to master was already fully merged in, so the commit made by the deprecation date step then didn't have any issues. 

I can't seem to find a scenario in which a true conflict would happen since the release branch is attempting to be merged in to master where things should be stable - am I missing something? Is this manual validation necessary or should we just remove it? @augustasgri @DanRod1999 Tagging for more eyes, what was the purpose of the check originally? What was the scenario we were afraid might happen?

augustasgri: @andremig-bentley This check was created to catch conflicts that happen when the same line differs in both branches. The scenario you described is why we made the check and I'm not sure why it hasn't produced a conflict for you. The workflow that produces a conflict goes like this:
1. A new `release` branch is created from `master`
2. Someone changes a deprecation comment in `master` (fixes a typo or changes wording)
3. A change to the same deprecation comment happens in `release` (`version-bump` script is run and it changes past deprecation date to text "can be removed in next major version")
4. When `version-bump` script tries to cherry-pick the commit from `release` branch onto `master` it finds the same line changed so it should result in a conflict.

Maybe your changes somehow fixed this edge-case 😄 

As for what happens differently when a conflict is detected:
- deprecation comment changes are not merged into `master` but a PR is created instead (someone needs to fix conflicts manually and then merge)
- there was also a "wait for manual validation" step in the release pipeline, that triggers if deprecation comment changes were made. We made that because this was a big addition to the pipeline and we did not 100% trust the script, so we decided to add the manual validation step until the logic "matures" so we didn't accidentally release some nonsense changes.

andremig-bentley: Ok great, thanks for the clarifications. Since I know how to reproduce the problem we're concerned about, I'll keep testing until I either confirm we need the manual validation or I feel safe that we can remove it.

common/config/azure-pipelines/jobs/version-bump.yaml
MichaelSwigerAtBentley: We will also need to check that the build is still getting flagged as 'nightly' in npm, but we will probably need to wait until this is merged to do so.

common/config/azure-pipelines/jobs/version-bump.yaml
MichaelSwigerAtBentley: It could just be a parameter of the sperate file. Resolving for now

common/config/azure-pipelines/jobs/version-bump.yaml
DanRod1999: I think that happens somewhere in the publish pipeline?

andremig-bentley:
Updates on this PR - 

After clarification on the purpose of the manual validation steps discussed [here](https://github.com/iTwin/itwinjs-core/pull/8470#discussion_r2318298585), I was able to find that merge conflicts were actually happening contrary to what I had previously thought ([here](https://github.com/iTwin/itwinjs-core/pull/8470#discussion_r2317345576)), but the manual validation steps were not ever hit. This was because I was attempting to use variables which were altered at run time like `DeprecationCommentChangesMade` within a template expression in the conditional, and that template expression gets processed at compile time, meaning the variable's value was always set to false no matter what happened within the run. 

After that fix and fixing the subsequent errors I had not found due to steps not ever being triggered, I believe I now have a working version. Now, the workflow can be boiled down to:

1. Run lint deprecation step, commit changes to branch.
2. Attempt to cherry pick changes onto master.
3. If successful, merge and move on to version bump.
4. If not successful (due to merge conflicts), move on to version bump, then after the version bump commit, create a PR with the deprecation date changes to be manually resolved.

Test results on the [fork](https://github.com/andremig-bentley/itwinjs-core-fork) are below. Link to release branch commit history [here](https://github.com/iTwin/itwinjs-core/compare/release/5.1.x...andremig-bentley:itwinjs-core-fork:release/5.1.x). Link to master commit history [here](https://github.com/iTwin/itwinjs-core/compare/release/5.1.x...andremig-bentley:itwinjs-core-fork:master).

First, testing with no merge conflicts for both patches and minors, here we see the commit history including commits altering deprecation date comments. The order of commits is: commit making the change, deprecation date commit, version bump commit, which is what we are looking for. This is present on both the release and master branches, though note that the merge of the version bump commit into master is not included in the commit history of the master branch since that is not done within the version bump pipeline so it was not included in my testing.
noconflict masterChanges

Next, testing with merge conflicts, we see the commit history is the same in the release branch, however in master we see no associated deprecation date commit, which is the expected behavior. A link to the PR created for this run is [here](https://github.com/andremig-bentley/itwinjs-core-fork/pull/3).

releaseConflict

Finally, testing with no changes to any deprecation date comments, we see the commit history is simply: commit making unrelated change, version bump commit.

noChanges

As a final test note, here is a link to a successful [nightly run](https://github.com/iTwin/itwinjs-core/commit/e6f9270ece1eebe988ac2132afa2f1a97fc2b894).

I am now marking this PR ready for review as I believe all questions and ambiguities we still had have been answered.

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: All discussion here has been addressed in latest [comment](https://github.com/iTwin/itwinjs-core/pull/8470#issuecomment-3264970402).

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: Addressed this briefly in latest [comment](https://github.com/iTwin/itwinjs-core/pull/8470#issuecomment-3264970402) with a link to a successful nightly run. I believe the issue with the weird versioning here was with the fork itself like I might have thought. It seems to me that if there was already a dev version created then the nightly runs worked fine, and it was only after i started updating with patches and minors without creating any new dev versions that it started messing up like above. Since this isn't how the real life branch will work and, more importantly, I didn't touch any logic for the nightly runs, this seems like it adequately resolves things. That being said, I can do more thorough testing if that's desired.

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -239,10 +297,33 @@ jobs:
...
+              git cherry-pick $commitHash
+
+              # Add, commit and push all ESLint deprecation changes (including conflicts) to conflict handling branch for manual resolution.
+              git add .
+              git commit -m "Add deprecation rule changes with all merge conflicts"
+              git push --set-upstream https://$(GITHUBTOKEN)@github.com/iTwin/itwinjs-core $conflictBranchName
+              echo "##[warning] Branch '$conflictBranchName' created and pushed. It contains all ESLint deprecation changes and cherry-pick conflicts. Manually fix conflicts inside branch and merge changes into master. Draft PR will be created now"
+
+              # Create a PR with new conflict branch
+              gh pr create --title "Merge deprecation comment changes into master" --body "This PR contains all changes after running lint-deprecation ESLint rule and merging with master. Manual intervention required: solve conflicts in this branch and merge PR into master. This happened because deprecation comments in master and in release branch could not be merged automatically. @iTwin/itwinjs-core-admins" --draft --head $conflictBranchName
```
aruniverse: might be good to include the ver in the pr title

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -169,6 +172,61 @@ jobs:
...
+              if git diff --quiet; then
+                echo "No deprecation comment changes were made"
+              else
+                echo "Deprecation comment changes were made"
+                git add .
+                echo "##vso[task.setvariable variable=deprecationCommentChangesMade]true"
+              fi
+            displayName: Determine whether ESLint rule made any changes
+
+          - bash: |
```
aruniverse, line 189: is it possible to do the deprecation and ver bump in the same git commit?

andremig-bentley: I believe we need two separate commits so that we can save the hash from the deprecation date commit to attempt the cherry pick step further down. If we included the stuff from both commits in the cherry pick, there's a chance that merge conflicts within the deprecation date changes would hold up the version bump. This way, the version bump can continue no matter what and the deprecation date changes can be pulled out into a separate PR if necessary.

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: Done in latest commit

common/config/azure-pipelines/jobs/version-bump.yaml
aruniverse: Well, i guess could continue to do the deprecation commit on the release branch to get the sha to cherry pick to main. but aftewards, when we do the version bump, we do a git commit ammend, to add that onto the deprecation commit so its effectively 1 commit?

andremig-bentley: It took me a little bit to get amend working as expected, but it is now working as of the latest commit to this PR. Testing results below.

We now expect the history on the release branch to look like:

1. Changes either requiring deprecation date comments or not
2. The version bump commit

And that should be it. Here is a screenshot of the release branch's commit history. First is changes that require deprecation date comments, then changes with no needed deprecation date comments, then finally changes that create merge conflicts requiring a PR.

releaseHistory

Within commit 5.2.5, we can see the deprecation date comment changes have been amended to be included in the version bump commit, screenshot below.

ammendedCommit

Lastly, here is the commit history of master. We expect the history to be:

1. Changes, either requiring deprecation date comments or not
2. If necessary, deprecation date application.

For our purposes here, we expect the deprecation date commit to only appear on master for the first round, as the second round contained no changes needing comment changes, and the third round created merge conflicts by not pushing the same changes on master that were present on the release branch, meaning a PR needed to be created. The created PR can be seen here: https://github.com/andremig-bentley/itwinjs-core-fork/pull/6

The screenshot below shows all commits are as expected on the master branch as well.

masterHistory

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -292,8 +292,12 @@ jobs:
 
       - bash: |
           echo Committing version bump $(getVersion.version)...
-
-          git commit -m "$(getVersion.version)" --author="imodeljs-admin <imodeljs-admin@users.noreply.github.com>"
+          if [ "$(deprecationCommentChangesMade)" = "true" ]; then
+            echo "Amending previous deprecation commit with version bump changes"
+            git commit --amend -m "$(getVersion.version)" --author="imodeljs-admin <imodeljs-admin@users.noreply.github.com>"
```
MichaelSwigerAtBentley, line 297: So this will change the previous commit message to $(getVersion.version)?

andremig-bentley: It does, but it also now includes all new changes for a commit at this point in the original commit. So if I have commit A with some changes, then I make more changes and normally that would be included in commit B, using amend instead at the point where I would have made commit B will mean that commit A is changed to include all changes that would have been included in commit B and now have the specified commit message.

MichaelSwigerAtBentley: Awesome, thats perfect

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -169,6 +172,61 @@ jobs:
...
+          - bash: |
+              git commit -m "Apply deprecation date rule"
+            displayName: Commit deprecation comment changes to release branch
+            condition: eq(variables['deprecationCommentChangesMade'], 'true')
+
+          - bash: |
+              deprecationDateCommitHash=$(git log -1 --format=%H)
+              echo "##vso[task.setvariable variable=hash;isOutput=true]$deprecationDateCommitHash"
+            displayName: Save SHA-1 hash of the commit
+            name: deprecationCommit
```
aruniverse, line 198: what is this name supposed to be? the name of the env var?

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -234,15 +292,42 @@ jobs:
...
         displayName: Commit version bump
 
       - ${{ if or(eq(parameters.BumpType, 'minor'), eq(parameters.BumpType, 'patch')) }}:
-        - template: "../templates/run-eslint-deprecation.yaml"
+          - bash: |
+              # Merge conflicts: Create branch, set up PR for manual resolution.
+              echo "##[warning] Cherry-pick operation on master failed (most likely due to merge conflicts). Manual intervention required"
+
+              # Create and checkout to conflict handling branch
+              conflictBranchName="cherry-pick-conflict-$(date +%y%m%d%H%m%s)"
```
aruniverse: i would prefer to not use date in the branch name, preferrably the ver

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -234,15 +292,42 @@ jobs:
...
+
+              # Add, commit and push all ESLint deprecation changes (including conflicts) to conflict handling branch for manual resolution.
+              git add .
+              git commit -m "Add deprecation rule changes with all merge conflicts"
+              git push --set-upstream https://$(GITHUBTOKEN)@github.com/iTwin/itwinjs-core $conflictBranchName
+              echo "##[warning] Branch '$conflictBranchName' created and pushed. It contains all ESLint deprecation changes and cherry-pick conflicts. Manually fix conflicts inside branch and merge changes into master. Draft PR will be created now"
+
+              # Create a PR with new conflict branch
+              gh pr create --title "Merge deprecation comment changes for version $(getVersion.version) into master" --body "This PR contains all changes after running lint-deprecation ESLint rule and merging with master. Manual intervention required: solve conflicts in this branch and merge PR into master. This happened because deprecation comments in master and in release branch could not be merged automatically. @iTwin/itwinjs-core-admins" --draft --head $conflictBranchName
```
aruniverse: no need to mention/tag the admins in the pr body, they will get added as reviewers anyway

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: It's the name of the step so that it can be used to access the variables created within the step (hash, on line 196) later like on line 208:

`commitHash=$(deprecationCommit.hash)`

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: Changed in latest commit.

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: Removed in latest commit.

andremig-bentley:
All recent changes including the copilot suggestions have been tested using the fork. The pipeline is behaving correctly as outlined in comments above.

aruniverse (approved):

PR #8487, opened 2025-09-05
Fix 'ghost' elements when using GraphicalEditingScope with map mask

pmconne:
If the background map is masked by models during a `GraphicalEditingScope`, then the 'ghosts' of any modified elements might show up in the mask and sometimes in the scene. 

`FeatureOverrides.buildLookupTable` calls `target.currentBranch.getFeatureAppearance` which includes a `FeatureAppearanceProvider` that hides modified and deleted elements in the static tiles. But when we collect the graphics for the mask, we push to a separate `BranchStack`, not the one that belongs to the `Target`. So depending on the order in which we produce the mask vs populate the scene, the modified elements may or may not end up hidden in the scene, and are never hidden in the mask.

Fix this by passing in to `FeatureOverrides.buildLookupTable` (through many layers of indirection) the `BranchState` that serves as the current `FeatureAppearanceProvider`.

Relatedly, some masking modes require their own `FeatureSymbology.Overrides` which must be updated whenever the view's own overrides change. They were being recomputed immediately in a event callback invoked when any geometric element is modified. This was error-prone as a different callback for the same event is used to update the `FeatureAppearanceProvider` to hide modified elements. Depending on the ordering of events, the mask's overrides might not hide the newly-modified elements. This has been addressed by making the mask's callback instead set a dirty flag telling it to recompute the overrides on the next update.

### DIFF
--- a/common/changes/@itwin/core-backend/2025-09-05-16-56.json
+++ b/common/changes/@itwin/core-backend/2025-09-05-16-56.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -18,8 +18,8 @@ importers:
   ../../core/backend:
     dependencies:
       '@bentley/imodeljs-native':
-        specifier: 5.2.12
-        version: 5.2.12
+        specifier: 5.2.13
+        version: 5.2.13
       '@itwin/cloud-agnostic-core':
         specifier: ^2.2.4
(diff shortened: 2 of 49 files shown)


pmconne:

pmconne:
Awaiting ImageTests results to confirm no regressions.

pmconne:
ImageTests shows regressions in V1 and V4 tilesets primarily related to masking and classification. Investigating.

pmconne:
Reverted original fix and pushed a better one. Details in PR description.

Now ImageTests reports no differences for V1 tilesets, but still a few for V4.

pmconne:
@khanaffan flaked out again on Linux.

```
--[ FAILURE: @itwin/core-backend ]-----------------[ 4 minutes 34.8 seconds ]--


  1624 passing (4m)
  26 pending
  1 failing

  1) ConcurrentQuery
       restart query:

      AssertionError: expected 3 to equal 2
      + expected - actual

      -3
      +2
      
      at Context.<anonymous> (src/test/ecdb/ConcurrentQuery.test.ts:135:28)
```

PR #8489, opened 2025-09-05
Introduce cesium-renderer package

eringram:
This PR introduces a new `@itwin/cesium-renderer` package which is planned to eventually contain a CesiumJS-based render system for iTwin.js and any related APIs.

So far it will only be used by display-test-app, so does not need to be published yet.

Currently, the new package contains a utility function called `createCesiumCameraProps` that converts a `ViewDefinition3dProps` to a `CesiumCameraProps` interface that can be used to create/modify a CesiumJS [Camera](https://cesium.com/learn/cesiumjs/ref-doc/Camera.html). I also added some unit tests for perspective and orthographic views.

### DIFF
--- a/common/changes/@itwin/core-frontend/arun-draco_2025-09-16-19-28.json
+++ b/common/changes/@itwin/core-frontend/arun-draco_2025-09-16-19-28.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/common/config/rush/browser-approved-packages.json
+++ b/common/config/rush/browser-approved-packages.json
@@ -539,10 +539,6 @@
       "allowedCategories": [ "backend", "integration-testing", "internal" ]
     },
     {
-      "name": "draco3d",
-      "allowedCategories": [ "frontend" ]
-    },
-    {
       "name": "electron",
       "allowedCategories": [ "backend", "internal", "tools" ]
     },
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -859,9 +859,6 @@ importers:
       '@loaders.gl/draco':
         specifier: ^4.3.4
         version: 4.3.4(@loaders.gl/core@4.3.4)
-      draco3d:
-        specifier: ^1.5.1
-        version: 1.5.1
       fuse.js:
         specifier: ^3.3.0
         version: 3.3.0
@@ -7116,9 +7113,6 @@ packages:
 
   dottie@2.0.6:
     resolution: {integrity: sha512-iGCHkfUc5kFekGiqhe8B/mdaurD+lakO9txNnTvKtA6PISrw86LgqHvRzWYPyoE2Ph5aMIrCw9/uko6XHTKCwA==}
-
-  draco3d@1.5.1:
-    resolution: {integrity: sha512-CxorIEl3kFWQvSdCAkGdcZoPKzkjuLKL1C78f509LZEdm8UPMySZjz1CXFOayF+yvJl3UPRpBC1V27uDPJkFSw==}
 
   draco3d@1.5.7:
     resolution: {integrity: sha512-m6WCKt/erDXcw+70IJXnG7M3awwQPAsZvJGX5zY7beBqpELw6RDGkYVU0W43AFxye4pDZ5i2Lbyc/NNGqwjUVQ==}
@@ -13653,8 +13647,6 @@ snapshots:
   dotenv@8.2.0: {}
 
   dottie@2.0.6: {}
-
-  draco3d@1.5.1: {}
 
   draco3d@1.5.7: {}
 
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -15,7 +15,7 @@
     "copy:js:cjs": "cpx \"./src/**/*.js\" ./lib/cjs",
     "copy:js:esm": "cpx \"./src/**/*.js\" ./lib/esm",
     "copy:workers": "cpx \"./lib/workers/webpack/parse-imdl-worker.js\" ./lib/public/scripts",
-    "copy:draco": "cpx \"./node_modules/draco3d/draco_decoder.wasm\" ./lib/public",
+    "copy:draco": "cpx \"./node_modules/@loaders.gl/draco/dist/libs/*\" ./lib/public/scripts",
     "docs": "betools docs --json=../../generated-docs/core/core-frontend/file.json --tsIndexFile=./core-frontend.ts --onlyJson --excludes=webgl/**/*,**/map/*.d.ts,**/tile/*.d.ts,**/*-css.ts",
     "extract-api": "betools extract-api --entry=core-frontend && npm run extract-extension-api",
     "extract-extension-api": "eslint --no-inline-config -c extraction.eslint.config.js  \"./src/**/*.ts\" 1>&2",
@@ -100,7 +100,6 @@
     "@itwin/webgl-compatibility": "workspace:*",
     "@loaders.gl/core": "^4.3.4",
     "@loaders.gl/draco": "^4.3.4",
-    "draco3d": "^1.5.1",
     "fuse.js": "^3.3.0",
     "wms-capabilities": "0.4.0"
   }
--- a/core/frontend/src/tile/GltfReader.ts
+++ b/core/frontend/src/tile/GltfReader.ts
@@ -2128,10 +2128,9 @@ export abstract class GltfReader {
         throw new Error("Unsupported browser for Draco decoding");
       }
 
-      const dracolib = await import("draco3d");
       const dracoLoader = (await import("@loaders.gl/draco")).DracoLoader;
 
-      await Promise.all(dracoMeshes.map(async (x) => this.decodeDracoMesh(x, dracoLoader, dracolib)));
+      await Promise.all(dracoMeshes.map(async (x) => this.decodeDracoMesh(x, dracoLoader)));
     } catch (err) {
       Logger.logWarning(FrontendLoggerCategory.Render, "Failed to decode draco-encoded glTF mesh");
       Logger.logException(FrontendLoggerCategory.Render, err);
@@ -2160,7 +2159,7 @@ export abstract class GltfReader {
     } catch { }
   }
 
-  private async decodeDracoMesh(ext: DracoMeshCompression, loader: typeof DracoLoader, draco3d: any): Promise<void> {
+  private async decodeDracoMesh(ext: DracoMeshCompression, loader: typeof DracoLoader): Promise<void> {
     const bv = this._bufferViews[ext.bufferView];
     if (!bv || !bv.byteLength)
       return;
@@ -2172,7 +2171,17 @@ export abstract class GltfReader {
     const offset = bv.byteOffset ?? 0;
     buf = buf.subarray(offset, offset + bv.byteLength);
 
-    const mesh = await loader.parse(buf, { modules: { draco3d } }); // NB: `options` argument declared optional but will produce exception if not supplied. Regardless, we are specifying our own bundled draco3d module to avoid CDN requests.
+    const mesh = await loader.parse(buf, {
+      draco: {
+        decoderType: "wasm",
+      },
+      modules: {
+        "draco_wasm_wrapper.js": `${IModelApp.publicPath}scripts/draco_wasm_wrapper.js`,
+        "draco_decoder.wasm": `${IModelApp.publicPath}scripts/draco_decoder.wasm`,
+      },
+      worker: false,
+      useLocalLibraries: true,
+    });
     if (mesh)
       this._dracoMeshes.set(ext, mesh);
   }


pmconne:
Let's introduce a new cesium-renderer package that depends on core-frontend where this function can live along with the rest of the renderer code.
See #5170 for an example of how to introduce a new core package. Specifically, look at:
- frontend-tiles/config/
- frontend-tiles/package.json
- .github/CODEOWNERS.md
- config/rush/browser-approved-packages.json
- rush.json

I would put the new package next to frontend-tiles in extensions/.

core/frontend/src/internal/render/CesiumCamera.ts
```diff
@@ -0,0 +1,87 @@
...
+*--------------------------------------------------------------------------------------------*/
+
+import { Cartographic, EcefLocation, ViewDefinition3dProps } from "@itwin/core-common";
+import { Point3d, Range3d, Vector3d, YawPitchRollAngles } from "@itwin/core-geometry";
+
+/** Properties that define a Cesium frustum.
+ * A [perspective frustum](https://cesium.com/learn/cesiumjs/ref-doc/PerspectiveFrustum.html) requires an fov and an [orthographic frustum](https://cesium.com/learn/cesiumjs/ref-doc/OrthographicFrustum.html) requires a width.
+ * @internal
+ */
+interface CesiumFrustum {
```
markschlosseratbentley: These interfaces contain properties that get fed into CesiumJS to create the relevant CesiumJS objects.

It could be an option to put these interfaces, and future interfaces that serve the same "interop" goal, into a separate file than the file that references them.

Then, they can be named simpler, like just Camera and Frustum.

When imported into a file where the system to which they refer needs clarity, they can get imported as something different - like import Frustum as CesiumFrustum.  Maybe we also would need to do such an import-renaming to avoid type naming collisions.

This is just a possibility I thought of while looking at the code --- I'm not necessarily saying this is the way it should be.

Semi-related: It might also be worthwhile to make the naming indicate the "interop" role of these types.  Cesium has a [Camera](https://cesium.com/learn/ion-sdk/ref-doc/Camera.html), for example.  It might be naturally assumed that a "CesiumCamera" or even a "Camera" here refers directly to that camera.  But it actually refers to properties used to make such a camera.

eringram: What do you think of naming these interfaces `CesiumCameraProps` and `CesiumFrustumProps` instead to distinguish them from the actual Camera/Frustum objects?

markschlosseratbentley: @eringram That sounds great - feels clearer to me.

common/api/cesium-renderer.api.md
```diff
@@ -0,0 +1,42 @@
...
+
+import { EcefLocation } from '@itwin/core-common';
+import { Point3d } from '@itwin/core-geometry';
+import { Range3d } from '@itwin/core-geometry';
+import { Vector3d } from '@itwin/core-geometry';
+import { ViewDefinition3dProps } from '@itwin/core-common';
+
+// @alpha
+export interface CesiumCameraProps {
+    // (undocumented)
```
markschlosseratbentley: Should we document these properties?

core/frontend/src/internal/render/CesiumCamera.ts
markschlosseratbentley: I am satisfied that my original comment here has been resolved with the latest form of this PR.  I am resolving this conversation.

extensions/cesium-renderer/src/CesiumCamera.ts
```diff
@@ -0,0 +1,87 @@
...
+  position: Point3d;
+  direction: Vector3d;
+  up: Vector3d;
+  frustum: CesiumFrustumProps;
+}
+
+/** Returns the position, orientation (direction, up), and frustum needed to create/modify a [Cesium camera object](https://cesium.com/learn/cesiumjs/ref-doc/Camera.html).
+ * @alpha
+ */
+export function createCesiumCameraProps(viewDefinition: ViewDefinition3dProps, ecefLoc?: EcefLocation, modelExtents?: Range3d): CesiumCameraProps {
```
markschlosseratbentley: What do you think about, instead of a list of arguments to the API function, having it take an options object?  Something like this:

```ts
export function createCesiumCameraProps(opts: { viewDefinition: ViewDefinition3dProps, ecefLoc?: EcefLocation, modelExtents?: Range3d}) : CesiumCameraProps
```

extensions/cesium-renderer/src/CesiumCamera.ts
```diff
@@ -0,0 +1,87 @@
...
+  const transformedUp = ecefLocation.getTransform().multiplyVector(up);
+  const transformedDirection = ecefLocation.getTransform().multiplyVector(direction);
+
+  const frustum: CesiumFrustumProps = {
+    near: 0.01,
+    far: 1000000,
+    fov,
+    width
+  };
+  const CesiumCameraProps: CesiumCameraProps = {
```
markschlosseratbentley: Is the casing of this constant's name intentional?  I think it might throw a lint error.

extensions/cesium-renderer/src/test/CesiumCamera.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    latitude: 0.6972007432483922,
+    longitude: -1.311456937133241,
+    height: 4.102413240985213
+  }
+};
+const ecefLocation = new EcefLocation(ecefLocProps);
+
+describe("createCesiumCameraProps", () => {
+  it("creates a CesiumCameraProps object from SpatialViewDefinitionProps with the camera on (perspective projection)", () => {
+    const CesiumCameraProps = createCesiumCameraProps(cameraOnView, ecefLocation);
```
markschlosseratbentley: Is the casing on this constant name intentional?

extensions/cesium-renderer/src/test/CesiumCamera.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    expectPointToEqual(CesiumCameraProps.up, expectedCesiumCameraProps.up, false);
+
+    expect(CesiumCameraProps.frustum.fov).toBeCloseTo(expectedCesiumCameraProps.frustum.fov);
+    expect(CesiumCameraProps.frustum.width).to.be.undefined;
+    expect(CesiumCameraProps.frustum.near).to.equal(expectedCesiumCameraProps.frustum.near);
+    expect(CesiumCameraProps.frustum.far).to.equal(expectedCesiumCameraProps.frustum.far);
+  });
+
+  it("creates a CesiumCameraProps object from SpatialViewDefinitionProps with the camera off (orthographic projection)", () => {
+    const CesiumCameraProps = createCesiumCameraProps(cameraOffView, ecefLocation);
```
markschlosseratbentley: Is the casing on this constant name intentional?

common/api/cesium-renderer.api.md
eringram: Added docs for them

extensions/cesium-renderer/src/CesiumCamera.ts
eringram: Nope, did this by accident with find and replace. Fixed it and the other instances

extensions/cesium-renderer/src/CesiumCamera.ts
eringram: Agreed because we try to do this most places for extensibility, just added

eringram:
@aruniverse @ben-polinsky We only plan to use this package in display-test-app for prototyping/experimentation for now. Is setting `"private": true` in the package.json and `"shouldPublish": false` in rush.json correct for this use case? We figured that's fine since DTA uses `workspace:*` for most itwin packages.

aruniverse:
> @aruniverse @ben-polinsky We only plan to use this package in display-test-app for prototyping/experimentation for now. Is setting `"private": true` in the package.json and `"shouldPublish": false` in rush.json correct for this use case? We figured that's fine since DTA uses `workspace:*` for most itwin packages.

Do you intend for external users (1st party or 3rd party) to use this package yet? Or just for internal consumption?

> Is setting `"private": true` in the package.json and `"shouldPublish": false` in rush.json correct for this use case?

Yes this should suffice, and you should be able to use the workspace protocol to consume this pkg in dta for testing

eringram:
> Do you intend for external users (1st party or 3rd party) to use this package yet? Or just for internal consumption?

No, just internal. The description of alpha release tags felt appropriate for the new APIs here ("collaborating with other itwinjs-core developers on highly experimental API"), but open to suggestions for it

markschlosseratbentley:
> and you should be able to use the workspace protocol to consume this pkg in dta for testing

I verified this works.

markschlosseratbentley (approved):

pmconne:
> Do you intend for external users (1st party or 3rd party) to use this package yet? Or just for internal consumption?

This package will eventually provide a self-contained implementation of core-frontend's `RenderSystem` interface, implemented using CesiumJS. `RenderSystem` is the public-facing API. The only public API cesium-renderer would expose (whether to 1st or 3rd parties) is some function to instantiate an instance of `RenderSystem` to be supplied to `IModelApp`.

rush.json
```diff
@@ -328,6 +332,13 @@
       "shouldPublish": true,
       "versionPolicyName": "prerelease-monorepo-lockStep"
     },
+    {
+      "packageName": "@itwin/cesium-renderer",
+      "projectFolder": "extensions/cesium-renderer",
+      "reviewCategory": "frontend",
+      "shouldPublish": false,
+      "versionPolicyName": "prerelease-monorepo-lockStep"
```
aruniverse: i dont think pkgs with should publish need to set the version policy name. take a look at internal-tools

extensions/cesium-renderer/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "lint-fix": "eslint --fix -f visualstudio \"./src/**/*.ts\" 1>&2",
+    "lint-deprecation": "eslint --fix -f visualstudio --no-inline-config -c ../../common/config/eslint/eslint.config.deprecation-policy.js \"./src/**/*.ts\"",
+    "test": "vitest --run"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/iTwin/itwinjs-core.git",
+    "directory": "extensions/cesium-renderer"
+  },
+  "keywords": [
```
aruniverse, line 29: i would add cesium to this list

extensions/cesium-renderer/LICENSE.md
```diff
@@ -0,0 +1,9 @@
+# MIT License
+
+Copyright © 2025 Bentley Systems, Incorporated. All rights reserved.
```
aruniverse: just get rid of the yr

extensions/cesium-renderer/src/CesiumCamera.ts
```diff
@@ -0,0 +1,102 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { Cartographic, EcefLocation, ViewDefinition3dProps } from "@itwin/core-common";
```
aruniverse: a new pkg gives us the opportunity to have a clean start and establish best practices. wuld be nice to start adding the import type qualifier where we are just using types. this will massively help us review dependencies listed in pkg.json

rush.json
pmconne: Does it make a difference that we intend to eventually publish this package?

aruniverse: they should be mutually exclusive; i believe we've run into issues in the past with publishing pkgs accidentally because we had both set

similar to whats being mentioned in https://github.com/microsoft/rushstack/issues/958

extensions/cesium-renderer/package.json
markschlosseratbentley: We should add this once we actually call CesiumJS APIs in here, IMO.  Which we will once we merge stuff from the prototype branch into this package. Though, it couldn't hurt to add the dependency now, I suppose. Here is where DTA prototype adds it, for reference: https://github.com/iTwin/itwinjs-core/blob/c3ab8ec0e2b161f50be4144c589643e66d57fcd9/test-apps/display-test-app/package.json#L81

aruniverse: as long as you remember to add the cesium keyword to the list

rush.json
eringram: Removed `versionPolicyName`

extensions/cesium-renderer/package.json
eringram: I interpreted this as adding Cesium to the keywords, not deps yet. Added the keyword now

extensions/cesium-renderer/LICENSE.md
eringram: Done

extensions/cesium-renderer/src/CesiumCamera.ts
eringram: Done, `ViewDefinition3dProps` is imported only as a type

rush.json
eringram: @aruniverse without `"versionPolicyName": "prerelease-monorepo-lockStep"`, the package is [not built](https://github.com/iTwin/itwinjs-core/actions/runs/17771042100/job/50506898926#step:6:7) during the build step of the extract-api job, so the extract-api command [fails](https://github.com/iTwin/itwinjs-core/actions/runs/17771042100/job/50506898926#step:7:106). Should we just publish this early now so it's treated like any other package with APIs?

extensions/cesium-renderer/src/CesiumCamera.ts
```diff
@@ -0,0 +1,102 @@
...
+  up: Vector3d;
+  /** The region of space in view of the camera. See [Camera.frustum](https://cesium.com/learn/cesiumjs/ref-doc/Camera.html#frustum). */
+  frustum: CesiumFrustumProps;
+}
+
+/** Returns the position, orientation (direction, up), and frustum needed to create/modify a [Cesium camera object](https://cesium.com/learn/cesiumjs/ref-doc/Camera.html).
+ * @alpha
+ */
+export function createCesiumCameraProps(opts: { viewDefinition: ViewDefinition3dProps, ecefLoc?: EcefLocation, modelExtents?: Range3d}): CesiumCameraProps {
+  const defaultOrigin = Cartographic.fromDegrees({ longitude: 0, latitude: 0, height: 0 });
```
ben-polinsky: nit: `Cartographic.createZero` is a bit more expressive and concise.

extensions/cesium-renderer/src/CesiumCamera.ts
```diff
@@ -0,0 +1,102 @@
...
+    position.addScaledInPlace(direction, -viewExtents.z);
+    width = viewExtents.x;
+  }
+
+  const transformedPosition = ecefLocation.getTransform().multiplyPoint3d(position);
+  const transformedUp = ecefLocation.getTransform().multiplyVector(up);
+  const transformedDirection = ecefLocation.getTransform().multiplyVector(direction);
+
+  const frustum = {
+    near: 0.01,
```
ben-polinsky: Perhaps extract the near and far values to constants, with a comment about why these specific values?

rush.json
aruniverse: the build script for the extract-api workflow runs the following: `node common/scripts/install-run-rush.js build -v -p max --to-version-policy prerelease-monorepo-lockStep`, which is why this new pkg isnt being built. you can build the cesium-renderer pkg on its own after the above command

we could, also just publish this pkg if we are certain we will eventually have to, but i would prefer to wait until we are sure in order to not pollute npm with itwin pkgs

markschlosseratbentley: @aruniverse so Erin could just add a new build step to the extract-api workflow to target just this package? Would you be okay with that?

extensions/cesium-renderer/src/CesiumCamera.ts
eringram: Fixed

extensions/cesium-renderer/src/CesiumCamera.ts
eringram: Made them into constants. This also made me realize I didn't know the motivation for these values (they are taken from other code) so I changed them to Cesium's defaults.

ben-polinsky: Indeed :) Thanks

ben-polinsky (approved):

PR #8500, opened 2025-09-08
Update Google Maps attribution

eringram:
Updated the Google Maps attribution to use the logo that says "Google Maps" instead of "Google" in accordance with their [policy](https://developers.google.com/maps/documentation/tile/policies#map-tiles-api).

Also:
- Updated tests to react to this
- Fixed a bug in `NativeGoogleMapsSession.getViewportInfoRequest` that caused the request to form incorrectly, and thus not return the 2D map tiles attribution info

### Screenshots

Attribution when displaying 2D satellite imagery or Photorealistic 3D Tiles

image

image

>**Note:** the attributions list is only shown on-screen currently for the Photorealistic 3D Tiles. For 2D map tiles they are only shown in the pop-up, screenshotted below. This is due to a technical limitation and would just take some more time in a later PR to implement. 

Attribution when displaying 2D roadmap or terrain imagery

image

Attribution in pop-up

image

### TODO
- [x] Run image tests
- [x] Add screenshots here

### DIFF
--- a/common/changes/@itwin/presentation-common/presentation-handle-struct-properties-field_2025-09-10-12-36.json
+++ b/common/changes/@itwin/presentation-common/presentation-handle-struct-properties-field_2025-09-10-12-36.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-common",
+      "comment": "Fix `ContentTraverser` droping metadata available for struct member properties.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/presentation-common"
+}
--- a/common/config/rush/version-policies.json
+++ b/common/config/rush/version-policies.json
@@ -2,7 +2,7 @@
   {
     "policyName": "prerelease-monorepo-lockStep",
     "definitionName": "lockStepVersion",
-    "version": "5.2.0-dev.25",
+    "version": "5.2.0-dev.26",
     "nextBump": "prerelease"
   }
 ]
--- a/core/backend/package.json
+++ b/core/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-backend",
-  "version": "5.2.0-dev.25",
+  "version": "5.2.0-dev.26",
   "description": "iTwin.js backend components",
   "main": "lib/cjs/core-backend.js",
   "module": "lib/esm/core-backend.js",
--- a/core/bentley/package.json
+++ b/core/bentley/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-bentley",
-  "version": "5.2.0-dev.25",
+  "version": "5.2.0-dev.26",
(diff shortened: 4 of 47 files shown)


core/frontend/src/RealityDataSource.ts
```diff
@@ -364,8 +364,13 @@ export class Google3dTilesProvider implements RealityDataSourceProvider {
       copyrightMsg += sortedCopyrights.map(([key]) => `<li>${key}</li>`).join("");
       copyrightMsg += "</ul>";
 
+      const iconSrc = document.createElement("img");
+      iconSrc.src = `${IModelApp.publicPath}images/GoogleMaps_Logo_Gray.svg`;
```
mdastous-bentley, line 368: Looks like we are assuming `IModelApp.publicPath` will end with a `/` character.  Are we sure it will always be the case?  If not, we might have a nasty bug.

markschlosseratbentley (approved):
Approving, but it would be good to try to get image tests to run before merging.

core/frontend/src/RealityDataSource.ts
ben-polinsky: It seems this assumption is made throughout the codebase. A utility function might be nice moving forward.

pmconne: If you're going to fix it, fix it [here](https://github.com/iTwin/itwinjs-core/blob/e93eb205ff55de05bf108374299218bcaebf26b7/core/frontend/src/IModelApp.ts#L429).

ben-polinsky: I'll open a good first issue issue. Let's not hold up this PR.
https://github.com/iTwin/itwinjs-core/issues/8510

LisaBosCesium:
@eringram Looks good from a Google compliance perspective, with one possible note. The Google policy guide now includes a requirement to ensure a minimum contrast between the logo and the background. The guide links to [this doc](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html). Is this something you evaluated?

eringram:
> Looks good from a Google compliance perspective, with one possible note. The Google policy guide now includes a requirement to ensure a minimum contrast between the logo and the background. The guide links to [this doc](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html). Is this something you evaluated?

Thanks, I had not looked at that linked contrast guide specifically until now. I used an online contrast ratio calculator with colors picked from the logos and maps background, and they all have acceptable ratios above 4.5:1.

I also followed this example of using the gray logo on a pure white background, dark logo on a light background, and the white logo on a dark/multicolor background (such as satellite imagery). 

image

eringram:
Follow-up issue for displaying attributions on-screen for 2D as well: https://github.com/iTwin/itwinjs-core/issues/8516

aruniverse:

PR #8504, opened 2025-09-09
Remove spurious assertions

### LINKED ISSUE #8503: FeatureAppearance.equals assert for lineRgb check
We were seeing an exception manifesting in Civil XSection with OS+ setting a feature override, I traced it to the equality check and can reproduce it with just the unit test below:

```
      assert.doesNotThrow(() => {
        const apprA = FeatureAppearance.defaults;
        const apprB = FeatureAppearance.fromJSON({ lineRgb: RgbColor.fromColorDef(ColorDef.white) });
        assert.isFalse(apprA.equals(apprB));
      });
```

The defaults have undefined for `lineRgb`, which causes the execution to fall into the assert path in this function:

```
function equalLineRgb(a: RgbColor | false | undefined, b: RgbColor | false | undefined): boolean {
  if (a === b) {
    return true;
  } else if (a instanceof RgbColor && b instanceof RgbColor) {
    return equalRgb(a, b);
  } else {
    assert(a === undefined || a === false);
    assert(b === undefined || b === false);
    return false;
  }
}
```

pmconne:
Fixes #8503. The assertions were never correct, but would only produce exceptions if assertions were enabled.

### DIFF
--- a/common/changes/@itwin/core-common/pmc-fix-feature-appearance-assertions_2025-09-09-16-04.json
+++ b/common/changes/@itwin/core-common/pmc-fix-feature-appearance-assertions_2025-09-09-16-04.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
+      "comment": "Fix spurious assertions in FeatureAppearance.equals.",
+      "type": "none"
+    }
+  ],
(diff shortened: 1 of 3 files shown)


pmconne:

ben-polinsky (approved):

markschlosseratbentley (approved):

PR #8512, opened 2025-09-10
fix: allow IModelApp.publicPath without trailing slash

### LINKED ISSUE #8510: Allow IModelApp.publicPath without trailing slash
From: https://github.com/iTwin/itwinjs-core/pull/8500#discussion_r2334532946

We assume all `IModelApp.publicPath` end with a trailing slash, but this is inflexible and could cause confusion for users. @pmconne points out a good spot to fix this is here: https://github.com/iTwin/itwinjs-core/blob/e93eb205ff55de05bf108374299218bcaebf26b7/core/frontend/src/IModelApp.ts#L429.

stim19:
Fixes #8510 

This change allows `IModelApp.publicPath` to be passed **without a trailing slash**. 
Both `"static"` and `"static/"` are normalized to `"static/"` internally.

### Details
Previously, `IModelApp.publicPath` assumed the path ended with a trailing slash, which caused errors if the path was provided without one. This fix normalizes the path when setting `_publicPath` in the constructor so that either form works.

### Example

```ts
new IModelApp({ publicPath: "assets" })._publicPath; // "assets/"
new IModelApp({ publicPath: "assets/" })._publicPath; // "assets/"
new IModelApp({})._publicPath; // ""
```


aruniverse:
Hi @stim19 , thank you for opening this PR, this is off to a great start!

Can you review this section on https://github.com/iTwin/itwinjs-core/blob/master/CONTRIBUTING.md#source-code-edit-workflow;? Your PR is missing rush changefiles generated by `rush change`.

Also, would you be able to add a simple unit test to verify your changes work as expected?

Once again, thank you for your contribution!

stim19:
Sure, I'll add a test and build the rust docs. Gimme a moment.

stim19:
Hi, I completed the tests and added the rush changefiles

package.json
```diff
@@ -29,5 +29,8 @@
     "url": "http://www.bentley.com"
   },
   "//dependencies": "NOTE: dependencies will be empty here and instead specified in each package.",
-  "//devDependencies": "NOTE: devDependencies will be empty here and instead specified in each package."
-}
\ No newline at end of file
+  "//devDependencies": "NOTE: devDependencies will be empty here and instead specified in each package.",
+  "dependencies": {
+    "@microsoft/rush": "^5.158.1"
```
MichaelSwigerAtBentley: Hi, thanks for adding that unit test and thanks for the contribution! 

As a monorepo, our dependencies are defined in the individual packages themselves. As for rush itself, its required as a part of our build process, but the version isn't mandated. Could you remove this dependencies block? 

Again, thanks for the contribution, and the other changes look good to me!

MichaelSwigerAtBentley:
Hi, it looks like the test is failing - 

IModelApp.startup() can only be called 1 time per shutdown, additional calls will simply exit as IModelApp is already initialized. 

We can do something like this:
```TS
describe("IModelApp startup tests", () => {
  afterEach(async () => {
    if (IModelApp.initialized)
      await IModelApp.shutdown();
  });

  it("should normalize path correctly", async() => {
    await IModelApp.startup({ publicPath: "assets"});
    expect(IModelApp.publicPath).toBe("assets/");
    await IModelApp.shutdown();
    await IModelApp.startup({ publicPath: "assets/"});
    expect(IModelApp.publicPath).toBe("assets/");
    await IModelApp.shutdown();
    await IModelApp.startup();
    expect(IModelApp.publicPath).toBe("")
    await IModelApp.shutdown();
  });
});
```

We can add a describe block to the end of the IModelApp.test.ts file, so our IModelApp.startup() calls won't conflict with the calls in the first describe block. Then we just need to call shutdown after each expect.

stim19:
Hi thanks for pointing that out. I added `shutdown( )` after each `startup( )` and moved the describe block to the bottom so that it doesn't cause any conflicts with the first block. 

### Test results
```
 ✓  chromium  src/test/IModelApp.test.ts (6 tests) 10671ms
   ✓ IModelApp > TestApp should override correctly
   ✓ IModelApp > Should get localized keyin, flyover, and description for tools
   ✓ IModelApp > Should do localizations
   ✓ IModelApp > Should support WebGL 1586ms
   ✓ IModelApp > Should create mock render system without WebGL 1529ms
   ✓ IModelApp startup tests > Should normalize path correctly 6636ms

 Test Files  1 passed (1)
      Tests  6 passed (6)
```

common/changes/@itwin/core-frontend/fix-imodel-app-public-path_2025-09-11-01-00.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "fix: allow IModelApp.publicPath to be passed without a trailing slash /",
```
aruniverse:
```suggestion
      "comment": "fix: allow IModelApp.publicPath to be passed without a trailing slash `/`",
```

core/frontend/src/test/IModelApp.test.ts
```diff
@@ -2,7 +2,7 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
-import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { afterAll, beforeAll, describe, expect, it, afterEach } from "vitest";
```
andremig-bentley:
```suggestion
import { afterAll, afterEach, beforeAll, describe, expect, it} from "vitest";

```

aruniverse:
Thank you for your contribution @stim19 !

PR #8522, opened 2025-09-12
Improve rebase for multi-user workflow

khanaffan:
imodel-native: https://github.com/iTwin/imodel-native/pull/1213

- **Restore Point Mechanism Added:**  
  - Introduces a system for creating, dropping, checking, and restoring "restore points" (snapshots) in BriefcaseDb via a new StashManager utility.
  - A special restore point named $pull_merge_restore_point is used to safely manage merge/rebase operations.

- **BriefcaseManager Enhancements:**  
  - Adds pullAndApplyChangesets logic to create/restore/destroy restore points during merges.
  - Ensures restore points are dropped after use.
  - Adds methods to create, drop, check, and restore restore points.

- **Transaction/Rebase Management Overhaul:**  
  - Expands ChangeMergeManager to orchestrate rebases, resume/pause/abort merges, and manage transaction (txn) state.
  - Adds support for registering/revoking conflict handlers for resolving merge conflicts.
  - New methods and types for querying transaction details and properties.

- **TxnManager Improvements:**  
  - Adds querying of transaction properties, last saved txn, and iterating all transactions.
  - Adds support for "indirect" transaction mode and checking for pending schema changes.
  - Adds event types and handler interfaces for rebase operations.

- **BriefcaseDb API Additions:**  
  - Adds discardChanges method to allow users to permanently revert all local changes, optionally keeping locks.

- **IModelDb API Changes:**  
  - Updates saveChanges to accept a structured argument (SaveChangesArgs) for richer change descriptions.
  - Prevents schema imports during rebase or indirect transactions.

- **Testing:**  
  - New tests for rebase/merge and restore point logic (Rebase.test.ts).

- **General:**  
  - Improved documentation and comments across affected classes and methods.
  - New file: StashManager.ts (for stash/restore point management).

### DIFF
--- a/core/backend/src/StashManager.ts
+++ b/core/backend/src/StashManager.ts
@@ -1,7 +1,7 @@
 import { DbResult, GuidString, Id64Array, Id64String, Logger, OpenMode } from "@itwin/core-bentley";
 import { ChangesetIdWithIndex, LocalDirName, LockState } from "@itwin/core-common";
 import { existsSync, mkdirSync, readdirSync, statSync, unlinkSync } from "node:fs";
-import * as path from "path";
+import * as path from "node:path";
 import { BriefcaseManager } from "./BriefcaseManager";
 import { BriefcaseDb } from "./IModelDb";
 import { _elementWasCreated, _getHubAccess, _hubAccess, _nativeDb, _resetIModelDb } from "./internal/Symbols";


core/backend/src/BriefcaseManager.ts
```diff
@@ -550,55 +568,155 @@ export class BriefcaseManager {
     if (reverse)
       changesets.reverse();
 
+    const nativeDb = db[_nativeDb];
```
MichaelSwigerAtBentley: If storing this as a const for clarity, can we move this to the top of the function and replace the remaining uses of db[_nativeDb] so its consistent for the whole function

core/backend/src/BriefcaseManager.ts
```diff
@@ -550,55 +568,155 @@ export class BriefcaseManager {
...
+    const briefcaseDb = db instanceof BriefcaseDb ? db : undefined;
 
-    let appliedChangesets = -1;
-    if (db[_nativeDb].hasPendingTxns() && !reverse && !arg.noFastForward) {
```
MichaelSwigerAtBentley, line 555: It looks like noFastForward arg is no longer checked and logic removed, is it no longer needed?

core/backend/src/BriefcaseManager.ts
```diff
@@ -550,55 +568,163 @@ export class BriefcaseManager {
...
+   * It then drops any existing restore point with the same name before creating a new one.
+   * The restore point is created using the `StashManager`, and its identifier is saved locally
+   * in the database under a key derived from the restore point name.
+   *
+   * @param args - An object containing:
+   *   - `db`: The {@link BriefcaseDb} instance for which to create the restore point.
+   *   - `name`: The unique name for the restore point. Must be a non-empty string.
+   * @returns A promise that resolves to the created stash object representing the restore point.
+   */
+  public static async createRestorePoint(args: { db: BriefcaseDb, name: string }) {
```
MichaelSwigerAtBentley: Nit: the rest of the BriefcaseManager functions use `db` + `arg` instead of `args` for their arguments. (I prefer args myself since it makes it clear its more than one)

Also should we add a new interface CreateRestorePointArgs incase we want to expand it?

core/backend/src/BriefcaseManager.ts
```diff
@@ -550,55 +568,163 @@ export class BriefcaseManager {
...
+   *
+   * This method looks up the restore point by its name, retrieves its internal identifier,
+   * and removes the associated stash if it exists. If the provided restore point name is empty,
+   * an error is thrown.
+   *
+   * @param args - An object containing:
+   *   - `db`: The {@link BriefcaseDb} instance from which to drop the restore point.
+   *   - `name`: The name of the restore point to be dropped. Must be a non-empty string.
+   */
+  public static dropRestorePoint(args: { db: BriefcaseDb, name: string }) {
```
MichaelSwigerAtBentley: Same note about `db` + `arg`

core/backend/src/BriefcaseManager.ts
```diff
@@ -550,55 +568,163 @@ export class BriefcaseManager {
...
+    if (args.name.length === 0) {
+      throw new IModelError(IModelStatus.BadArg, "Invalid restore point name");
+    }
+    const restorePointId = args.db[_nativeDb].queryLocalValue(`restore_point/${args.name}`);
```
MichaelSwigerAtBentley: Should we return restorePointId so we can tell when something is deleted vs when one didnt exist in the first place?

tcobbs-bentley: restorePointId wouldn't be useful since it is about to get deleted, but bool might be useful (although apparently not now).

core/backend/src/BriefcaseManager.ts
```diff
@@ -550,55 +568,163 @@ export class BriefcaseManager {
...
+   * @returns A promise that resolves to the created stash object representing the restore point.
+   */
+  public static async createRestorePoint(args: { db: BriefcaseDb, name: string }) {
+    if (args.name.length === 0) {
+      throw new IModelError(IModelStatus.BadArg, "Invalid restore point name");
+    }
+    this.dropRestorePoint(args);
+
+    const stash = await StashManager.stash({ db: args.db, description: `restore_point/${args.name}` });
+    args.db[_nativeDb].saveLocalValue(`restore_point/${args.name}`, stash.id);
```
tcobbs-bentley: It's minor, but I would suggest creating a function to generate the key:
```ts
private static restorePointKey(name: string): string {
  return `restore_point/${name}`;
}
```

tcobbs-bentley: Actually, you can put the `name.length > 0` logic in `restorePointKey` and throw the exception from that function. That would get rid of having to do that in all the places that you use the restore point name.

core/backend/src/BriefcaseManager.ts
```diff
@@ -550,55 +568,163 @@ export class BriefcaseManager {
...
+      throw new IModelError(IModelStatus.BadArg, "Invalid restore point name");
+    }
+    const restorePointId = args.db[_nativeDb].queryLocalValue(`restore_point/${args.name}`);
+    if (!restorePointId) {
+      return false;
+    }
+
+    const stash = StashManager.getStash({ db: args.db, stash: restorePointId });
+    if (!stash) {
+      return false;
```
tcobbs-bentley, line 689: Should it delete the local value if it gets here? It seems like that value is now invalid. Or should it throw?

core/backend/src/IModelDb.ts
```diff
@@ -801,13 +820,18 @@ export abstract class IModelDb extends IModel {
    * @note This will not push changes to the iModelHub.
    * @see [[IModelDb.pushChanges]] to push changes to the iModelHub.
    */
-  public saveChanges(description?: string): void {
+  public saveChanges(description?: string | SaveChangesArgs): void {
```
tcobbs-bentley: I feel like this should be switched to an overridden function:
```ts
public saveChanges(description?: string): void;
public saveChanges(args?: SaveChangesArgs): void;
public saveChanges(descriptionOrArgs?: string | SaveChangesArgs): void {
...
}
```
The top two could then have more clear documentation about their param.

core/backend/src/IModelDb.ts
```diff
@@ -3004,6 +3046,30 @@ export class BriefcaseDb extends IModelDb {
     return db?.isBriefcaseDb() ? db : undefined;
   }
 
+  /**
+   * Permanently discards any local changes made to this briefcase, reverting the briefcase to its last synchronized state.
+   * This operation cannot be undone. By default, all locks held by this briefcase will be released unless the `holdLocks` option is specified.
```
tcobbs-bentley: Should `holdLocks` be `retainLocks`?

core/backend/src/IModelDb.ts
```diff
@@ -3004,6 +3046,30 @@ export class BriefcaseDb extends IModelDb {
...
+   * @Note This operation can be performed at any point including after failed rebase attempts.
+   * @param args - Options for discarding changes.
+   * @param args.holdLocks - If `true`, retains all currently held locks after discarding changes. If omitted or `false`, all locks will be released.
```
tcobbs-bentley: `holdLocks` -> `retainLocks`

core/backend/src/BriefcaseManager.ts
khanaffan: That is correct. It's not used any more. It's declare internal.

Can I remove it from args?

core/backend/src/BriefcaseManager.ts
khanaffan: I have separate them out but I do not not think we need createRestorePointsArgs. This api is internal so we can change it as required.

core/backend/src/BriefcaseManager.ts
khanaffan: I have separate them out but I do not not think we need createRestorePointsArgs. This api is internal so we can change it as required.

core/backend/src/BriefcaseManager.ts
khanaffan: I added delete code. But reason I didn't at first was the fact we cannot call saveChanges() in this function. So I added the code to delete but with comment that saveChanges() should not be called in this function.

core/backend/src/BriefcaseManager.ts
tcobbs-bentley: It looks like PullChangesArgs is public, so I guess it can't be removed. Maybe update the comment to say "unused"?

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+import { IModelHost } from "./IModelHost";
+import { BackendLoggerCategory } from "./BackendLoggerCategory";
+
+const loggerCategory = BackendLoggerCategory.StashManager;
+
+/**
+ * Custom error class for stash-related errors.
+ * @internal
+ */
+export class StashError extends Error {
```
tcobbs-bentley, line 19: Do these all get caught internally and then converted to something like IModelError? I seems iffy to have a new direct Error subclass.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+/**
+ * Properties for creating a stash
+ * @internal
+ */
+export interface CreateStashProps {
+  /** Briefcase database instance */
+  readonly db: BriefcaseDb;
+  /** description of the stash */
+  readonly description: string;
+  /** discard all local changes and unless keepLocks flag is set, all locks will be released */
```
tcobbs-bentley:
```suggestion
  /** discard all local changes and unless retainLocks flag is set, all locks will be released */
```

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+  readonly discardLocalChanges?: true;
+  /** keep all locks after discarding local changes */
```
tcobbs-bentley:
```suggestion
  /** retains all locks after discarding local changes */
```

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+ *
+ * Extends {@link StashArgs} with an additional `method` property to specify the operation.
+ *
+ * @property method - Specifies the stash operation to perform.
+ *   - `"apply"`: Apply the stash without removing it.
+ *   - `"restore"`: Apply the stash and remove it from the stash list.
+ * @internal
+ */
+export interface StashApplyArgs extends StashArgs {
+  readonly method: "apply" | "restore";
```
tcobbs-bentley: I'm trying to understand why you use a union of string literal here and a enum type for `LockOrigin`. Other than having 2 values vs 3 values, I don't really see a difference in the two use cases. If we go with the "string literal unions are preferred over enums" argument, then why have an enum for LockOrigin. If we go with the "enums are preferred over string literal unions" argument, then why does this use a string literal union?

Note: I honestly don't know if one of the two (string literal unions vs enums) is considered to be "better" in iTwin. However, even if we don't care, it seems a bit odd that these two (right next to each other in the code) answer the question with opposite answers.

tcobbs-bentley: Forgot to mention: I do get that `LockOrigin` is a numeric enum, and not a string-based enum, but I don't think that really affects things (although I could be wrong).

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+}
+
+enum LockOrigin {
+  Acquired = 0,
+  NewElement = 1,
+  Discovered = 2,
+}
+
+/**
+ * Stash manager allow stash, drop, apply and merge stashes
```
tcobbs-bentley, line 97: I'm guessing that this is conceptually similar to git stash. Assuming this is correct, it might be good to mention that here.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+ * @internal
+ */
+export class StashManager {
+
+  private static readonly STASH_ROOT_DIR_NAME: string = ".stashes";
```
tcobbs-bentley: I think having a singular constant (`STASH`) with a plural value (`stashes`) is not ideal.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+    if (!db.isOpen || db.isReadonly)
+      throw new StashError("Database is not open or is readonly");
+
+    if (!existsSync(db[_nativeDb].getFilePath())) {
+      throw new StashError("Could not determine briefcase path");
+    }
+
+    const stashDir = path.join(path.dirname(db[_nativeDb].getFilePath()), StashManager.STASH_ROOT_DIR_NAME, `${db.briefcaseId}`);
+    if (ensureExists && !existsSync(stashDir)) {
+      mkdirSync(stashDir, { recursive: true });
```
tcobbs-bentley, line 120: Should this have an exception handler that throws a `StashError`?

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+export class StashManager {
+
+  private static readonly STASH_ROOT_DIR_NAME: string = ".stashes";
+  /**
+   * Retrieves the root folder path for stash files associated with the specified BriefcaseDb.
+   *
+   * @param db - The BriefcaseDb instance for which to determine the stash root folder.
+   * @param ensureExists - If true, the stash root directory will be created if it does not already exist.
+   * @returns The absolute path to the stash root directory.
+   * @throws IModelError if the database is not open, is readonly, has unsaved changes, or if the briefcase path cannot be determined.
```
tcobbs-bentley:
```suggestion
   * @throws StashError if the database is not open, is readonly, has unsaved changes, or if the briefcase path cannot be determined.
```

core/backend/src/StashManager.ts
tcobbs-bentley: If not, the `@throws` part of the comment is probably wrong, since it can also throw whatever error `mkdirSync` throws.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+  private static getStashId(args: StashArgs) {
+    return (typeof args.stash === "string" ? args.stash : args.stash.id).toLowerCase();
+  }
+
+  /**
+   * Retrieves the file path to the stash file associated with the provided arguments.
+   *
+   * @param args - The arguments required to identify the stash, including the database reference.
+   * @returns The absolute path to the stash file.
+   * @throws IModelError with status `IModelStatus.BadArg` if the stash root folder or the stash file does not exist.
```
tcobbs-bentley: This isn't correct, since it throws `StashError`.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+    const stashRoot = this.getStashRootFolder(args.db, false);
+    if (!existsSync(stashRoot)) {
+      throw new StashError("Stash root folder does not exist");
+    }
+
+    const stashFile = path.join(stashRoot, `${this.getStashId(args)}.stash`);
+    if (!existsSync(stashFile)) {
+      throw new StashError("Invalid stash");
+    }
+    return stashFile;
```
tcobbs-bentley: `stashFile` -> `stashFilePath`?

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+  }
+
+  /**
+   * Queries the stash database for lock IDs matching the specified state and origin.
+   *
+   * @param args - The arguments required to access the stash database.
+   * @param state - The lock state to filter by.
+   * @param origin - The lock origin to filter by.
+   * @returns An array of lock IDs (`Id64Array`) that match the given state and origin.
```
tcobbs-bentley, line 163: `@throws` missing: `withStash` can throw `IModelError`. Not sure if anything else throws `StashError`.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+   */
+  private static queryLocks(args: StashArgs, state: LockState, origin: LockOrigin): Id64Array {
+    return this.withStash(args, (stashDb) => {
+      const query = `SELECT JSON_GROUP_ARRAY(FORMAT('0x%x', Id)) FROM [locks] WHERE State = ${state} AND origin = ${origin}`;
+      return stashDb.withPreparedSqliteStatement(query, (stmt) => {
+        while (stmt.step() === DbResult.BE_SQLITE_ROW) {
```
tcobbs-bentley: Does this use `while` instead of `if` because that's the standard for dealing with the results of a SQL fetch? Obviously there will never be more than one iteration of the "loop".

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+          return JSON.parse(stmt.getValueString(0)) as Id64Array;
+        }
+        return [];
+      });
+    });
+  }
+
+  /**
+   * Acquire locks for the specified stash. If this fail then stash should not be applied.
+   * @param args The stash arguments.
```
tcobbs-bentley, line 179: `@throws` missing: this also seems like it can throw.

core/backend/src/BriefcaseManager.ts
khanaffan: done

core/backend/src/BriefcaseManager.ts
MichaelSwigerAtBentley: Shouldn't we tag it deprecated, with a note its no longer needed, then we can remove it in next major

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+  /**
+   * Creates a stash of changes for the specified briefcase.
+   *
+   * This method generates a stash in the stash root directory for the given briefcase, using the provided description and iModelId.
+   * Optionally, it can reset the briefcase by releasing all locks after stashing.
+   *
+   * @param args - The properties required to create a stash, including the briefcase, description, iModelId, and an optional resetBriefcase flag.
+   * @returns A promise that resolves to the properties of the created stash.
+   *
+   * @throws Error if the stash operation fails.
```
tcobbs-bentley: I think if you don't want to specify the actual error type, you should just omit `Error`.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+   */
+  public static async stash(args: CreateStashProps): Promise<StashProps> {
+    const stashRootDir = StashManager.getStashRootFolder(args.db, true);
```
tcobbs-bentley: I think this should happen after the other checks that throw `StashError`s (to just above the current line 237).

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+
+    if (args.db.txns.hasUnsavedChanges) {
+      throw new StashError("Unsaved changes exist");
+    }
+
+    if (args.db.txns.hasPendingSchemaChanges) {
+      throw new StashError("Pending schema changeset. Stashing is not currently supported for schema changes");
+    }
+
+    const stash = args.db[_nativeDb].stashChanges({ stashRootDir, description: args.description, iModelId }) as StashProps;
```
tcobbs-bentley, line 219: Is there a reason that `stashChanges` returns `any` instead of `StashProps`? Is that to hide `StashProps` from end users?

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+
+  /**
+   * Retrieves the stash properties from the database for the given arguments.
+   *
+   * This method attempts to read the stash information from the local database using a prepared SQL statement.
+   * If successful, it returns the stash properties along with the file path of the stash database.
+   * If an error occurs during the process, it logs the error and returns `undefined`.
+   *
+   * @param args - The arguments required to locate and access the stash.
+   * @returns The stash file properties if found; otherwise, `undefined`.
```
tcobbs-bentley, line 232: `@throws` missing.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+   */
+  public static getStash(args: StashArgs): StashProps | undefined {
+    try {
+      return this.withStash(args, (stashDb) => {
+        const stashProps = stashDb.withPreparedSqliteStatement("SELECT [val] FROM [be_Local] WHERE [name]='$stash_info'", (stmt) => {
+          if (stmt.step() !== DbResult.BE_SQLITE_ROW)
+            throw new IModelError(IModelStatus.BadArg, "Invalid stash");
```
tcobbs-bentley: Why do we use `IModelError` here and below and `StashError` in almost everywhere for the rest of the file?

core/backend/src/StashManager.ts
tcobbs-bentley: Never mind; I see that it catches all errors and returns `undefined`.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+          return JSON.parse(stmt.getValueString(0)) as StashProps;
+        });
+        return stashProps;
+      });
+    } catch (error: any) {
+      const stashId = typeof args.stash === "string" ? args.stash : args.stash.id;
```
tcobbs-bentley: Doesn't `getStashId` perform this logic?

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+   *
+   * @param args - The arguments including the target database and stash properties.
+   * @throws IModelError if the restore operation fails.
+   */
+  public static async restore(args: StashArgs): Promise<void> {
+    const { db } = args;
+    Logger.logInfo(loggerCategory, `Restoring stash: ${this.getStashId(args)}`);
+
+    db.clearCaches();
+    db[_nativeDb].clearECDbCache();
```
tcobbs-bentley: Should these two lines get moved down, so they don't happen at all in the event of some error being thrown?

core/backend/src/IModelDb.ts
khanaffan: I tried it and it cause lint error
```
  818:22  error  These overloads can be combined into one signature taking `string | SaveChangesArgs`  @typescript-eslint/unified-signatures
  ```

Comment (tcobbs-bentley) [reply core/backend/src/IModelDb.ts]: I guess update the param comment instead?

Comment (khanaffan) [reply core/backend/src/IModelDb.ts]: it seem one of the overload has to be none optional. Then it work fine.

```ts
  public saveChanges(description: string): void;
  public saveChanges(args?: SaveChangesArgs): void;
  public saveChanges(descriptionOrArgs?: string | SaveChangesArgs): void
  ```

Comment (tcobbs-bentley) [reply core/backend/src/IModelDb.ts]: It would probably be best to make description optional so that its signature exactly matches the original function.

File: core/backend/src/StashManager.ts (hunk 11 above)
Comment (khanaffan) [reply core/backend/src/StashManager.ts:19]: `IModelError` is good. I was previously using one with error number. And i do not have really any error numbers here. So i defined a custom one with log message.

I think IModelError is perfect. Will switch to that.

Comment (khanaffan) [reply core/backend/src/StashManager.ts:19]: Ah wait. IModelError also require error number.

Comment (tcobbs-bentley) [reply core/backend/src/StashManager.ts:19]: Yeah; you use the bad arg status in other places.

File: core/backend/src/StashManager.ts (hunk 23 above)
Comment (khanaffan) [reply core/backend/src/StashManager.ts:179]: i am removing throw from all docs in this file. As native methods in there also can throw and of different types

File: core/backend/src/StashManager.ts (hunk 22 above)
Comment (khanaffan) [reply core/backend/src/StashManager.ts]: if will be fine here as well. It just have single row as result.

File: core/backend/src/StashManager.ts (hunk 30 above)
Comment (khanaffan) [reply core/backend/src/StashManager.ts]: I should have removed it. I moved part of this as IModelDb.discardChanges() which already does all the cache clean up

File: core/backend/src/StashManager.ts (hunk 14 above)
Comment (khanaffan) [reply core/backend/src/StashManager.ts]: So i will remove this as well. Stash can be applied on top of local changes but its not part of this PR. This PR can only do restore usecase where briefcase is reverted to a previous state. So i am remove StashApplyArgs and will add it when we implement Stash apply on top of local change feature.

File: core/backend/src/StashManager.ts [hunk 31]
```diff
@@ -0,0 +1,458 @@
...
+    if (db.changeset.id !== stash.parentChangeset.id) {
+      // Changeset ID mismatch
+      Logger.logWarning(loggerCategory, "Changeset ID mismatch");
+      const stashChangeset = await this.queryChangeset({ db, stash });
+      await BriefcaseManager.pullAndApplyChangesets(db, { toIndex: stashChangeset.index });
+    }
+
+    db[_nativeDb].stashRestore(stashFile);
+    (db as any).loadIModelSettings();
+    (db as any).initializeIModelDb("pullMerge");
```
tcobbs-bentley: Is there perhaps a better way to call protected and private functions? If not, I feel that a comment is needed to explain why db is being cast to any.

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,458 @@
...
+    db.saveChanges();
+    Logger.logInfo(loggerCategory, `Restored stash: ${this.getStashId(args)}`);
+  }
+
+  /**
+   * Applies a stashed change to the specified {@link BriefcaseDb}.
+   *
+   * This method validates the stash, ensures the database is in a valid state, and applies or restores the stash as requested.
+   *
+   * @param args - The arguments for applying the stash, including the method ("restore").
```
tcobbs-bentley: I'm confused. This forces you to set method to `"restore"`, but the function is named `apply`, and it takes `StashApplyArgs`. But passing a `method` of `"apply"` causes it to throw an exception. This seems to be the only function that uses `StashApplyArgs`. What's going on?

core/backend/src/TxnManager.ts
```diff
@@ -295,18 +296,204 @@ interface IConflictHandler {
...
+ * @property prevId - (Optional) The identifier of the previous transaction in the sequence.
+ * @property props - The arguments or properties associated with the save changes operation.
+ * @property type - The type of transaction, which can be "Data", "EcSchema", or "Ddl".
+ * @property reversed - Indicates whether the transaction has been reversed.
+ * @property grouped - Indicates whether the transaction is grouped with others.
+ * @property timestamp - The timestamp when the transaction was created.
+ */
+export interface TxnProps {
+  id: TxnIdString;
+  sessionId: number;
```
tcobbs-bentley, line 326: Not documented above.

core/backend/src/TxnManager.ts
```diff
@@ -295,18 +296,204 @@ interface IConflictHandler {
...
+  type: TxnType;
+  reversed: boolean;
+  grouped: boolean;
+  timestamp: string;
+}
+
+/**
+ * Manages the process of merging and rebasing local changes (transactions) in a [[BriefcaseDb]] or [[StandaloneDb]].
+ *
+ * The `ChangeMergeManager` coordinates the rebase of local transactions when pulling and merging changes from other sources,
```
tcobbs-bentley:
```suggestion
 * The `RebaseManager` coordinates the rebase of local transactions when pulling and merging changes from other sources,
```

core/backend/src/TxnManager.ts
```diff
@@ -295,18 +296,204 @@ interface IConflictHandler {
...
+   * 3. Ends the rebase process and saves changes if the database is not read-only.
+   * 4. Drops any restore point associated with the pull-merge operation.
+   *
+   * If an error occurs during the process, the rebase is aborted and the error is rethrown.
+   *
+   * @throws {Error} If a transaction cannot be found or if any step in the rebase process fails.
+   */
+  public async resume() {
+    const nativeDb = this._iModel[_nativeDb];
+    const txns = this._iModel.txns;
```
tcobbs-bentley, line 379:
```suggestion
    const txns = this._txns;
```

core/backend/src/TxnManager.ts
```diff
@@ -295,18 +296,204 @@ interface IConflictHandler {
...
+    try {
+      nativeDb.pullMergeRebaseBegin();
+      let txnId = nativeDb.pullMergeRebaseNext();
+      while (txnId) {
+        const txnProps = this._txns.getTxnProps(txnId);
```
tcobbs-bentley:
```suggestion
        const txnProps = txns.getTxnProps(txnId);
```

core/backend/src/TxnManager.ts
```diff
@@ -295,18 +296,204 @@ interface IConflictHandler {
...
+        if (!txnProps) {
+          throw new Error(`Transaction ${txnId} not found`);
+        }
+
+        txns.onRebaseTxnBegin.raiseEvent(txnProps);
+        Logger.logInfo(BackendLoggerCategory.IModelDb, `Rebasing local changes for transaction ${txnId}`);
+        const shouldReinstate = this._customHandler ? this._customHandler.shouldReinstate(txnProps) : true;
```
tcobbs-bentley: I _think_ the following does the same thing:
```suggestion
        const shouldReinstate = this._customHandler?.shouldReinstate(txnProps) ?? true;
```

core/backend/src/TxnManager.ts
```diff
@@ -295,18 +296,204 @@ interface IConflictHandler {
...
+  }
+
+  /**
+   * Sets the handler to be invoked for rebase operations.
+   *
+   * @param handler - The {@link RebaseHandler} to handle rebase events.
+   */
+  public setCustomHandler(handler: RebaseHandler) {
+    if (this._customHandler) {
+      throw new IModelError(IModelStatus.BadArg, "Rebase handler already set");
```
tcobbs-bentley: Is there a reason they're not allowed to change it after setting it once?

core/backend/src/TxnManager.ts
```diff
@@ -295,18 +296,204 @@ interface IConflictHandler {
...
+ *
+ * @alpha
+ */
+export class RebaseManager {
   private _conflictHandlers?: IConflictHandler;
+  private _customHandler?: RebaseHandler;
   public constructor(private _iModel: BriefcaseDb | StandaloneDb) { }
-  public resume() {
-    this._iModel[_nativeDb].pullMergeResume();
+  private get _txns() {
```
tcobbs-bentley: Should you also have `private get _nativeDb()` and then use use `this._nativeDb` in places where you currently use `this._iModel[_nativeDb]`?

core/backend/src/TxnManager.ts
```diff
@@ -627,11 +864,11 @@ export class TxnManager {
   public readonly onReplayedExternalTxns = new BeEvent<() => void>();
 
   /** @internal */
-  public readonly onRebaseTxnBegin = new BeEvent<(txn: TxnArgs) => void>();
+  public readonly onRebaseTxnBegin = new BeEvent<(txn: TxnProps) => void>();
```
tcobbs-bentley, line 874: Are these events really ok to be internal?

core/backend/src/TxnManager.ts
tcobbs-bentley: I see later that `_nativeDb` is already implemented as a getter. I feel that all the code that currently uses `this._iModel[_nativeDb]` should be updated to use that getter.

core/backend/src/BriefcaseManager.ts
```diff
@@ -56,7 +57,7 @@ export interface PushChangesArgs extends TokenArg {
   /** The delay to wait between retry attempts on failed pushes. Default is 3 seconds. */
   pushRetryDelay?: BeDuration;
   /**
-   *  For testing purpose
+   *  (unused)
    * @internal
    */
   noFastForward?: true;
```
tcobbs-bentley, line 64: This should be marked as deprecated.

core/backend/src/BriefcaseManager.ts
```diff
@@ -80,7 +81,7 @@ export type PullChangesArgs = ToChangesetArgs & {
    */
   onProgress?: ProgressFunction;
   /**
-   *  For testing purpose
+   *  (unused)
    * @internal
    */
   noFastForward?: true;
```
tcobbs-bentley, line 89: This should be marked as deprecated.

core/backend/src/IModelDb.ts
```diff
@@ -801,13 +820,33 @@ export abstract class IModelDb extends IModel {
...
-  public saveChanges(description?: string): void {
+  public saveChanges(description?: string): void ;
+  /** Commit unsaved changes in memory as a Txn to this iModelDb.
+   * @param args Optional  [[SaveChangesArgs]] of the changes
+   * @throws [[IModelError]] if there is a problem saving changes or if there are pending, un-processed lock or code requests.
+   * @note This will not push changes to the iModelHub.
+   * @see [[IModelDb.pushChanges]] to push changes to the iModelHub.
+   */
+
+    public saveChanges(descriptionOrArgs: SaveChangesArgs): void;
```
tcobbs-bentley:
```suggestion
    public saveChanges(args: SaveChangesArgs): void;
```

core/backend/src/IModelDb.ts
```diff
@@ -801,13 +820,33 @@ export abstract class IModelDb extends IModel {
    * @note This will not push changes to the iModelHub.
    * @see [[IModelDb.pushChanges]] to push changes to the iModelHub.
    */
-  public saveChanges(description?: string): void {
+  public saveChanges(description?: string): void ;
+  /** Commit unsaved changes in memory as a Txn to this iModelDb.
+   * @param args Optional  [[SaveChangesArgs]] of the changes
```
tcobbs-bentley:
```suggestion
   * @param args [[SaveChangesArgs]] of the changes
```

core/backend/src/StashManager.ts
khanaffan: You have to put a type in some other package like bentley or common and then wait for it to publish then reference it back in imodel-native. So its not hidden it just impossible to reference a type that is declared in typescript and used in native side in same PR.

common/api/core-backend.api.md
```diff
@@ -6642,8 +6686,34 @@ export class TxnManager {
...
+    // (undocumented)
+    sessionId: number;
+    // (undocumented)
+    timestamp: string;
+    // (undocumented)
+    type: TxnType;
 }
 
+// @alpha
+export type TxnType = "Data" | "EcSchema" | "Ddl";
```
ColinKerr: Should we use `EC` instead of `Ec` to be consistent with other uses of EC?

core/backend/src/BriefcaseManager.ts
ColinKerr: Remember to deprecate we need to document how a caller can update if they set it to true/false explicitly.

core/backend/src/IModelDb.ts
```diff
@@ -76,6 +76,25 @@ import { ElementLRUCache, InstanceKeyLRUCache } from "./internal/ElementLRUCache
 
 const loggerCategory: string = BackendLoggerCategory.IModelDb;
 
+/**
+ * Arguments for saving changes to the iModel.
+ * @public
```
ColinKerr: Should we be creating this new interface and the associated overloads public now rather than starting them off as alpha or beta so we have a chance to modify them?

core/backend/src/IModelHost.ts
```diff
@@ -203,6 +203,11 @@ export interface IModelHostOptions {
    * this option can be set to true to revert to old workflow.
    */
   disableThinnedNativeInstanceWorkflow?: boolean;
+
+  /**
+   * Configuration controlling whether to disable the creation of restore points during pull/merge operations.
+   */
+  disableRestorePointOnPullMerge?: true;
```
ColinKerr, line 211: This should start as beta just like the implementation in IModelHostConfiguration

core/backend/src/TxnManager.ts
khanaffan: Application must set it once. I would preferred an assert() but I think I can log warning instead.

core/backend/src/TxnManager.ts
khanaffan: no they need to be upgraded to alpha like rest of rebase api

common/api/core-backend.api.md
khanaffan: Following camel case we have been using Ec prefix for identifiers. Though this is string identifier so I changed it to "ECSchema"

core/backend/src/IModelDb.ts
khanaffan: For that we need to add new method for `saveChanges()` that is marked as alpha. saveChanges() is public so any arg it get also need to be public.

core/backend/src/BriefcaseManager.ts
khanaffan: I marked it deprecated

core/backend/src/BriefcaseManager.ts
khanaffan: In my view dropRestorePoint() only throws if it fail to drop a valid stash. A boolean variable will not work as it also throw.

core/backend/src/BriefcaseManager.ts
khanaffan: fixed

core/backend/src/IModelDb.ts
khanaffan: fixed

core/backend/src/IModelDb.ts
khanaffan: fixed

core/backend/src/IModelDb.ts
khanaffan: fixed

core/backend/src/StashManager.ts
khanaffan: Fixed. I have removed all `IModelError` from StashManager.ts

core/backend/src/StashManager.ts
khanaffan: fixed

core/backend/src/StashManager.ts
khanaffan: fixed

core/backend/src/StashManager.ts
khanaffan: I have removed StashApplyArgs and StashManager.apply() for now. Its something we will implement in future to able to apply stash on top of local changes.

core/backend/src/StashManager.ts
khanaffan: Yes it's similar to git stash. Currently it only does restore() but apply/merge is something we will implement in future. This api is internal for that reason. Once we have apply/merge we will make it public. And methods would probably be on briefcase e.g. 
BriefcaseDb.stash()
BriefcaseDb.applyStash()
BriefcaseDb.dropStash()
BriefcaseDb.stashes

core/backend/src/StashManager.ts
khanaffan: fixed

core/backend/src/StashManager.ts
khanaffan: fixed. All throw docs are removed.

core/backend/src/StashManager.ts
khanaffan: fixed

core/backend/src/StashManager.ts
khanaffan: fixed

core/backend/src/StashManager.ts
khanaffan: throw docs removed.

core/backend/src/StashManager.ts
khanaffan: all IModelError removed from stashManager and replaced with StashError.

core/backend/src/StashManager.ts
khanaffan: Good catch. Removed and replaced with getStashId()

core/backend/src/StashManager.ts
khanaffan: fixed

core/backend/src/StashManager.ts
khanaffan: Fixed and use internal symbols to access those functions.

core/backend/src/StashManager.ts
khanaffan: StashApplyArgs removed.

core/backend/src/TxnManager.ts
khanaffan: fixed

core/backend/src/TxnManager.ts
khanaffan: fixed

core/backend/src/TxnManager.ts
khanaffan: removed _txns getter.

core/backend/src/TxnManager.ts
khanaffan: removed _txns getter

core/backend/src/TxnManager.ts
khanaffan: fixed

core/backend/src/TxnManager.ts
khanaffan: fixed

core/backend/src/TxnManager.ts
khanaffan: fixed made them alpha.

core/backend/src/BriefcaseManager.ts
khanaffan: fixed

core/backend/src/BriefcaseManager.ts
khanaffan: fixed

core/backend/src/IModelDb.ts
khanaffan: fixed

core/backend/src/IModelDb.ts
khanaffan: fixed

core/backend/src/IModelHost.ts
khanaffan: fixed

core/backend/src/IModelDb.ts
ColinKerr: Is there no option to make one overload of the a different level than another?  Then make the combined one internal?

core/backend/src/internal/ChannelAdmin.ts
```diff
@@ -69,7 +69,7 @@ class ChannelAdmin implements ChannelControl {
 
   public [_verifyChannel](modelId: Id64String): void {
     // Note: indirect changes are permitted to change any channel
-    if (this._allowedModels.has(modelId) || this._iModel[_nativeDb].isIndirectChanges())
+    if (this._allowedModels.has(modelId) || this._iModel[_nativeDb].getTxnMode() === "indirect")
```
ColinKerr, line 72: Should TxnMode be an enum or type?  You use 'indirect' and 'direct' as strings in several methods both to check and set type.

core/backend/src/test/hubaccess/Rebase.test.ts
```diff
@@ -0,0 +1,660 @@
...
+  private _data = 0;
+  public constructor() { }
+  public async startup() {
+    HubMock.startup("TestIModel", KnownTestLocations.outputDir);
+    this.iModelId = await HubMock.createNewIModel({ iTwinId: HubMock.iTwinId, iModelName: "Test", description: "TestSubject" });
+    const b1 = await HubWrappers.downloadAndOpenBriefcase({ iTwinId: HubMock.iTwinId, iModelId: this.iModelId });
+    b1.channels.addAllowedChannel(ChannelControl.sharedChannelName);
+    b1.saveChanges();
+    const schema1 = `<?xml version="1.0" encoding="UTF-8"?>
+    <ECSchema schemaName="TestDomain" alias="ts" version="01.00" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.1">
```
ColinKerr: Use EC 3.2 spec for new test schemas unless there is a specific reason why we need to use 3.1

core/backend/src/IModelDb.ts
```diff
@@ -796,18 +820,39 @@ export abstract class IModelDb extends IModel {
...
+   * @note This will not push changes to the iModelHub.
+   * @see [[IModelDb.pushChanges]] to push changes to the iModelHub.
+   */
+  public saveChanges(descriptionOrArgs?: string | SaveChangesArgs): void {
     if (this.openMode === OpenMode.Readonly)
       throw new IModelError(IModelStatus.ReadOnly, "IModelDb was opened read-only");
 
-    const stat = this[_nativeDb].saveChanges(description);
+    const args = typeof descriptionOrArgs === "string" ? { description: descriptionOrArgs } : descriptionOrArgs;
+    if (!this[_nativeDb].hasUnsavedChanges()) {
```
MichaelSwigerAtBentley, line 851: Should we just return here if there are no changes, to avoid calling into native?

core/backend/src/IModelDb.ts
```diff
@@ -3004,6 +3067,31 @@ export class BriefcaseDb extends IModelDb {
...
+   * @param args.retainLocks - If `true`, retains all currently held locks after discarding changes. If omitted or `false`, all locks will be released.
+   * @returns A promise that resolves when the operation is complete.
+   * @throws May throw if discarding changes fails.
+   */
+  public async discardChanges(args?: { retainLocks?: true }): Promise<void> {
+    Logger.logInfo(loggerCategory, "Discarding local changes");
+    this.clearCaches();
+    this[_nativeDb].clearECDbCache();
+    this[_nativeDb].discardLocalChanges();
+    this[_resetIModelDb]();
```
MichaelSwigerAtBentley, line 3087: Should we wrap in a trycatch to better parse potential errors?

core/backend/src/IModelHost.ts
```diff
@@ -197,12 +197,18 @@ export interface IModelHostOptions {
   allowSharedChannel?: boolean;
 
   /**
-   * Setting this to true withh revert to the previous behavior of using the native side for all CRUD operations.
+   * Setting this to true with revert to the previous behavior of using the native side for all CRUD operations.
```
MichaelSwigerAtBentley:
```suggestion
   * Setting this to true will revert to the previous behavior of using the native side for all CRUD operations.
```

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,385 @@
...
+   * Retrieves the stash properties from the database for the given arguments.
+   *
+   * This method attempts to read the stash information from the local database using a prepared SQL statement.
+   * If successful, it returns the stash properties along with the file path of the stash database.
+   * If an error occurs during the process, it logs the error and returns `undefined`.
+   *
+   * @param args - The arguments required to locate and access the stash.
+   * @returns The stash file properties if found; otherwise, `undefined`.
+   */
+  public static getStash(args: StashArgs): StashProps | undefined {
```
MichaelSwigerAtBentley: Should we split into tryGetStash (that just calls getStash, swallows potential errors and returns undef) and getStash that throws its errors? It might be useful to throw the errors so we can differentiate between a case where there is no stash found and when there is an error retrieving it during runtime

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,385 @@
...
+   *
+   * @param args - The arguments including the target database and stash properties.
+   */
+  public static async restore(args: StashArgs): Promise<void> {
+    const { db } = args;
+    Logger.logInfo(loggerCategory, `Restoring stash: ${this.getStashId(args)}`);
+
+    const stash = this.getStash(args);
+    if (!stash) {
+      throw new StashError(`stash not found ${this.getStashId(args)}`);
```
MichaelSwigerAtBentley:
```suggestion
      throw new StashError(`Stash not found ${this.getStashId(args)}`);
```
NIT: the other errors are capitalized

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,385 @@
...
+    if (db.iModelId !== stash.iModelId) {
+      throw new StashError(`Stash does not belong to this iModel`);
+    }
+
+    if (db.briefcaseId !== stash.briefcaseId) {
+      throw new StashError(`Stash does not belong to this briefcase`);
+    }
+
+    const stashFile = this.getStashFilePath({ db, stash });
+    await db.discardChanges({ retainLocks: true });
```
MichaelSwigerAtBentley, line 378: Why do we retain the locks on the discarded changes here? Is it because the changes could be saved but not pushed so we want the locks to remain?

Also we have checked to make sure there are no unsaved changes, so we call this to ensure the caches are cleared?

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,385 @@
...
+    await this.acquireLocks(args);
+    if (db.changeset.id !== stash.parentChangeset.id) {
```
MichaelSwigerAtBentley, line 380: Does this revert to the parentChangeset of the stash? Is it possible to apply stash to the latest changeset in cases like this - even if it creates conflicts that will need resolution?

core/backend/src/test/hubaccess/Rebase.test.ts
khanaffan: fixed

core/backend/src/StashManager.ts
khanaffan: Well it could be possible to apply on tip. But it's beyond scope of this PR. It will fall into case where we can apply stash on top of local changes or different tip. I wrote a issue to continue work on that in future

https://github.com/iTwin/itwinjs-backlog/issues/1626

core/backend/src/StashManager.ts
khanaffan: fixed

MichaelSwigerAtBentley (approved):

core/backend/src/StashManager.ts
khanaffan: Good point. There reason I have retainLocks is because restore() is use for short-term checkpoint in an operation like pull/merge. There are no new locks that are taken by twin.js though there could be new locks taken by rebase process. Which should be fine to release. 

We need a way to only release new locks that is not part of stash locks. Some how take intersection of two set and only retain those locks.

May be another item for future to make it a bit smarter.

core/backend/src/StashManager.ts
khanaffan: fixed

common/api/core-backend.api.md
khanaffan: fixed

core/backend/src/internal/ChannelAdmin.ts
khanaffan: fixed made it a type.

core/backend/src/IModelDb.ts
khanaffan: in above I am only aware of discardLocalChanges() throwing IModelError.

core/backend/src/IModelDb.ts
khanaffan: no, savechange() call commit which still commit sqlite transaction. We do not track all table e.g. be_Local is not tracked and changing them will still cause hasUnsavedChanges() to return false. But there is changes and we do want to store it.

core/backend/src/IModelDb.ts
khanaffan: yes that worked

marking SaveChangesArgs as alpha
marking saveChanges(description | SaveChangesArgs) as internal
while saveChanges(description?) stay public

common/api/core-backend.api.md
```diff
@@ -6559,18 +6597,24 @@ export class TxnManager {
     appCustomConflictHandler?: (args: DbRebaseChangesetConflictArgs) => DbConflictResolution | undefined;
     beginMultiTxnOperation(): DbResult;
     cancelTo(txnId: TxnIdString): IModelStatus;
-    // @internal (undocumented)
-    readonly changeMergeManager: ChangeMergeManager;
     deleteAllTxns(): void;
     endMultiTxnOperation(): DbResult;
     getChangeTrackingMemoryUsed(): number;
+    getCurrentSessionId(): number;
```
aruniverse, line 6605: should this be alpha?

common/api/core-backend.api.md
```diff
@@ -6559,18 +6597,24 @@ export class TxnManager {
...
+    getMode(): TxnMode;
     getMultiTxnOperationDepth(): number;
     getRedoString(): string;
     getTxnDescription(txnId: TxnIdString): string;
+    // @alpha
+    getTxnProps(id: TxnIdString): TxnProps | undefined;
     getUndoString(): string;
     get hasFatalError(): boolean;
     get hasLocalChanges(): boolean;
+    get hasPendingSchemaChanges(): boolean;
```
aruniverse, line 6620: should this be alpha?

common/api/core-backend.api.md
```diff
@@ -3974,6 +3975,8 @@ export class IModelHostConfiguration implements IModelHostOptions {
     // (undocumented)
     static defaultTileRequestTimeout: number;
     // @beta
```
aruniverse, line 3978: everything else youve added in this pr is alpha, any reason this is beta?

ColinKerr (approved):

core/backend/src/IModelDb.ts
```diff
@@ -796,18 +820,41 @@ export abstract class IModelDb extends IModel {
...
    * @note This will not push changes to the iModelHub.
    * @see [[IModelDb.pushChanges]] to push changes to the iModelHub.
    */
-  public saveChanges(description?: string): void {
+  public saveChanges(descriptionOrArgs?: string | SaveChangesArgs): void {
     if (this.openMode === OpenMode.Readonly)
       throw new IModelError(IModelStatus.ReadOnly, "IModelDb was opened read-only");
 
-    const stat = this[_nativeDb].saveChanges(description);
+    const args = typeof descriptionOrArgs === "string" ? { description: descriptionOrArgs } : descriptionOrArgs;
```
aruniverse, line 850: why do we even check & log this?

core/backend/src/BriefcaseManager.ts
```diff
@@ -551,54 +572,151 @@ export class BriefcaseManager {
...
+        nativeDb.pullMergeRebaseBegin();
+        let txnId = nativeDb.pullMergeRebaseNext();
+        while (txnId) {
+          nativeDb.pullMergeRebaseReinstateTxn();
+          nativeDb.pullMergeRebaseUpdateTxn();
+          txnId = nativeDb.pullMergeRebaseNext();
+        }
+        nativeDb.pullMergeRebaseEnd();
+        if (!nativeDb.isReadonly) {
+          nativeDb.saveChanges("Merge.");
```
aruniverse, line 617: should the save comment mention rebase at all?

core/backend/src/StashManager.ts
```diff
@@ -0,0 +1,392 @@
+import { DbResult, GuidString, Id64Array, Id64String, Logger, OpenMode } from "@itwin/core-bentley";
+import { ChangesetIdWithIndex, LocalDirName, LockState } from "@itwin/core-common";
+import { existsSync, mkdirSync, readdirSync, statSync, unlinkSync } from "node:fs";
+import * as path from "path";
```
aruniverse:
```suggestion
import * as path from "node:path";
```

aruniverse: does `import path from "node:path"` work? or do you need to use `import *`

core/backend/src/StashManager.ts
aruniverse: Shouldnt we be using ITwinError?

core/backend/src/IModelDb.ts
khanaffan: Because developer might think that he called saveChanges() with edit command and all information and yet it never  stored because at some point before saveChanges() some other code had done saveChanges().

I found this annoying when writing test and found that some internal code was calling saveChanges() and my comments were not getting saved.

khanaffan: e.g. you might want to call saveChanges() after schema import and say what schema you have imported. But that does not work because somewhere inside importschema() we already called saveChanges() and so developer comments is never saved and calling getTxnProps() never return it.

core/backend/src/BriefcaseManager.ts
khanaffan: It does not really matter what we put here. It's StandaloneDb case. which never pullMerge. This is there because of historic reason as BriefcaseManager work on IModelDb instead of BriefcaseDb. Only briefcases can pull/merge. So there are check to see if IModelDb is briefcase or not which is pointless but unless we fix BriefcaseManager to take BriefcaseDb we would have these else cases.

core/backend/src/StashManager.ts
tcobbs-bentley: @khanaffan I would resolve this based on you handling my original comment, but I'm not sure about the last comment in this thread from @aruniverse.

core/backend/src/StashManager.ts
khanaffan: `import path from "node:path"` does not work.

core/backend/src/StashManager.ts
aruniverse: resolving since this is internal, but anything new we publish should be using ITwinError

common/api/core-backend.api.md
khanaffan: The thing this flag control is all internal not alpha. I was not really sure what to mark it? Do you have suggestion here.

common/api/core-backend.api.md
khanaffan: It's just primitive function nothing about it will change. Though I agree we should mark it alpha for now.

common/api/core-backend.api.md
khanaffan: Same logic like hasPendingSchemaChanges(). Just too primitive it return current session id from native side. I do not know what could change about it. Though I will mark it alpha.

common/api/core-backend.api.md
aruniverse: mark it as alpha instead of beta then

Issue #7943, opened 2025-04-09
Improvements for the interaction and the visuals of Accudraw's inputs at cursor.


mathieu-fournier:
1- Accudraw Input fileds at cursor should highlight the text, when appropriate.
Reason : When typing in the input fields, the reflex most users had was to start by removing the text with backspaces. Having the text highlighted would have notified the user that they can start typing right away to overwrite the value, no backspace needed. 

2- Accudraw input fields should have locks icons, when the field(s) are locked.
Reason : Users didn't notice that the icon buttons where a bit darker when locked. Having a darker button is also not standard to mean that something is "locked". So they had trouble knowing if the fields where locked or not. Having a lock icon at the end of the input field would fix that. 

3- Accudraw input fields should allow to delete all characters. 
Reason : Most users naturally backspace the entire text to "reset" / "start anew". Currently, removing the last character automatically resets the fields to it's original value. This felt like a bug, we should allow backspacing the field completely. 

4- Accudraw visual should match ITwinUI standards. 
Reason : The visual should be coherent with the visual scheme of our apps. Ex: It should change when changing from light to dark theme, it should look similar to other input fields, etc... "Kiwi is going to make a mockup for that"

5- Accudraw "icons" should not look like buttons. 
Reason : A lot of users tried to click them and told us explicitely that it was frustrating not being able to click them. 

6- Accudraw "icons" should be replaced with labels OR we need to improve the icons. 
Reason : Users understood the labels "X,Y,Distance,Angle,Bearing". But they had difficulty knowing what the icons meant.

7-  Up and Down arrow should not focus the prev/next field. This should be reserved to the standard keys for that "Tab" / "Shift-Tab".

bbastings:
1- Accudraw Input fileds at cursor should highlight the text, when appropriate.
It looks quite unpleasant and is very flishy-flashy constantly reselecting the string when the value is changing in dynamics. Desire was to try to avoid this as it looks unprofessional...

2- Accudraw input fields should have locks icons, when the field(s) are locked.
They aren't just darker, they are pushed in. I will try to make that clearer, toggle icon buttons are a standard ui element. The desire was for a compact ui element that could be a fixed size as opposed to a lable + lock icon.

3- Accudraw input fields should allow to delete all characters.
[...]

PR #8527, opened 2025-09-15
Attempt to make Accudraw more conformant with ItwinUI color standards

bbastings:
Try to address 4 from #7943

### DIFF
--- a/common/config/rush/version-policies.json
+++ b/common/config/rush/version-policies.json
@@ -2,7 +2,7 @@
   {
     "policyName": "prerelease-monorepo-lockStep",
     "definitionName": "lockStepVersion",
-    "version": "5.3.0-dev.10",
+    "version": "5.3.0-dev.11",
     "nextBump": "prerelease"
   }
 ]
--- a/core/backend/package.json
+++ b/core/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-backend",
-  "version": "5.3.0-dev.10",
+  "version": "5.3.0-dev.11",
   "description": "iTwin.js backend components",
   "main": "lib/cjs/core-backend.js",
   "module": "lib/esm/core-backend.js",
--- a/core/bentley/package.json
+++ b/core/bentley/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-bentley",
-  "version": "5.3.0-dev.10",
+  "version": "5.3.0-dev.11",
   "description": "Bentley JavaScript core components",
   "main": "lib/cjs/core-bentley.js",
   "module": "lib/esm/core-bentley.js",
--- a/core/common/package.json
+++ b/core/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-common",
-  "version": "5.3.0-dev.10",
+  "version": "5.3.0-dev.11",
   "description": "iTwin.js components common to frontend and backend",
   "main": "lib/cjs/core-common.js",
   "module": "lib/esm/core-common.js",
--- a/core/ecschema-editing/package.json
+++ b/core/ecschema-editing/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-editing",
-  "version": "5.3.0-dev.10",
+  "version": "5.3.0-dev.11",
   "description": "ECSchema editing and validation API",
   "license": "MIT",
   "main": "lib/cjs/ecschema-editing.js",
--- a/core/ecschema-locaters/package.json
+++ b/core/ecschema-locaters/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-locaters",
-  "version": "5.3.0-dev.10",
+  "version": "5.3.0-dev.11",
   "description": "EC Schema file locaters",
   "license": "MIT",
   "main": "lib/cjs/ecschema-locaters.js",
--- a/core/ecschema-metadata/package.json
+++ b/core/ecschema-metadata/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-metadata",
-  "version": "5.3.0-dev.10",
+  "version": "5.3.0-dev.11",
   "description": "ECObjects core concepts in typescript",
   "license": "MIT",
   "main": "lib/cjs/ecschema-metadata.js",
--- a/core/ecschema-rpc/common/package.json
+++ b/core/ecschema-rpc/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-rpcinterface-common",
-  "version": "5.3.0-dev.10",
+  "version": "5.3.0-dev.11",
   "description": "Schema RPC Interface common interface",
   "main": "lib/cjs/ecschema-rpc-interface.js",
   "module": "lib/esm/ecschema-rpc-interface.js",
--- a/core/ecschema-rpc/impl/package.json
+++ b/core/ecschema-rpc/impl/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-rpcinterface-impl",
-  "version": "5.3.0-dev.10",
+  "version": "5.3.0-dev.11",
   "description": "Schema RPC Interface backend implementation",
   "main": "lib/cjs/ecschema-rpc-impl.js",
   "module": "lib/esm/ecschema-rpc-impl.js",
--- a/core/ecsql/common/package.json
+++ b/core/ecsql/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecsql-common",
-  "version": "5.3.0-dev.10",
+  "version": "5.3.0-dev.11",
   "description": "ECSql component that can be reference on backend and frontend",
   "main": "lib/cjs/ecsql-common.js",
   "module": "lib/esm/ecsql-common.js",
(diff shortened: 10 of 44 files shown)


core/frontend/src/tools/AccuDrawViewportUI.ts
```diff
@@ -120,15 +145,20 @@ export class AccuDrawViewportUI extends AccuDraw {
...
+          ? true
+          : undefined)
+    )
       return;
     IModelApp.accuSnap.userSettings.toolTip = false;
     IModelApp.notifications.clearToolTip();
   }
 
   private unsuspendToolTips() {
-    if (undefined === this._toolTipsSuspended)
```
bbastings, line 130: Please undo these gratuitous formatting changes. In particular it is helpful when debugging if the return is on a separate  line.

It also makes it much more difficult to review and see the actual changes...

core/frontend/src/tools/AccuDrawViewportUI.ts
```diff
@@ -41,44 +41,71 @@ export class AccuDrawViewportUI extends AccuDraw {
...
+    border: {
+      width: "1px",
+      style: "solid",
+      /** Corner radius of text input fields and locks buttons. */
+      radius: "var(--iui-border-radius-1)",
+    },
     /** Settings specific to text input fields and lock button labels. */
     text: {
       /** Font family to use for text input field values and button labels. */
-      fontFamily: "sans-serif",
```
bbastings, line 63: There's no fallback for these? Did you try this with display-test-app where these css variables aren't defined?

core/frontend/src/tools/AccuDrawViewportUI.ts
carlos-menezes: Unsure how this happened because I have Prettier auto-formatting turned off for Bentley repositories. Will undo.

core/frontend/src/tools/AccuDrawViewportUI.ts
carlos-menezes: Done.

carlos-menezes:
@bbastings here's the look on the `display-test-app`:

[Kapture 2025-09-19 at 17.20.32.webm](https://github.com/user-attachments/assets/7b887efd-7ad0-41a9-99d8-6001752fa9e9)

carlos-menezes:
Typing in the input fields:

[Kapture 2025-09-19 at 17.24.42.webm](https://github.com/user-attachments/assets/f6222a94-c1ae-45ca-af07-aa5f33bdeab7)

bbastings:
When you are satisfied with the changes, the next step is to show UX in a studio app. If they approve, change this PR to ready to review.

bbastings:
You've eliminated the code for the fixed horizontal layout. This is an option for folks that might be bothered by having stuff on the cursor. Take a look at the screenshots from the [documentation](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/frontend/AccuDrawUI.md) it's mentioned right at the top.

Here's what it's doing now, still using the vertical layout fixed at the bottom of the view.
image

Also, I don't think the contrast between the focused border outline and the view background is very good. The "focus outline" that used to be there by default looked pretty good I think?
image

core/frontend/src/tools/AccuDrawViewportUI.ts
```diff
@@ -46,39 +46,91 @@ export class AccuDrawViewportUI extends AccuDraw {
...
-      fontSize: "9pt",
-      /** Font color to use for text input field values and button labels. */
-      color: "white",
-      /** Background color of focused text input field. */
-      focusColor: "rgba(50, 50, 200, 0.75)",
+    /** Settings that apply to both text input fields and lock buttons. */
+    field: {
+      /** Number of visible characters to show in text input fields. */
+      size: 12,
+      height: "var(--iui-component-height-small, 1.75rem)",
```
bbastings: Please document these new public settings.

carlos-menezes: Done.

The other comments have been addressed too. I moved the fixed horizontal layout logic to another place and I messed up the merge when applying my stashed changes. I agree with the point on contrast. I've tried to match it a bit closer to what's in Figma and I think we're pretty close now.

Both changes are visible below:

image

carlos-menezes:
@joshuapage-ux this is how these changes look in Studio:


[Kapture 2025-09-22 at 13.44.51.webm](https://github.com/user-attachments/assets/66d5976d-38bd-41a2-8f17-397b4087526e)

bbastings:
Focus contrast looks good now. I think the vertical gap between the input fields could be smaller though...
image
Also, maybe un-focused field background could be slightly more grey to improved contrast with a black view background (which is pretty common)?
image

carlos-menezes:
I've set the opacity to `--iui-opacity-2`. We can turn it up to the next level, but at that point the opacity would be `1`.

Spacing is `4px` between rows/columns. UX will probably ask for changes on the horizontal layout spacing but we'll see.

image

Screenshot 2025-09-22 at 14 10 52

bbastings:
Vertical spacing and contrast looks good.

> UX will probably ask for changes on the horizontal layout spacing but we'll see.

Yeah, I like to see a bigger gap following the toggle buttons. :)

aruniverse:
shouldnt this be handled at the appui / desktop framework level?

cc @GerardasB @johnnyd710

core/frontend/src/tools/AccuDrawViewportUI.ts
```diff
@@ -46,39 +46,97 @@ export class AccuDrawViewportUI extends AccuDraw {
...
-      /** Font family to use for text input field values and button labels. */
-      fontFamily: "sans-serif",
-      /** Font size to use for text input field values and button labels. */
-      fontSize: "9pt",
-      /** Font color to use for text input field values and button labels. */
-      color: "white",
-      /** Background color of focused text input field. */
-      focusColor: "rgba(50, 50, 200, 0.75)",
+    /** Settings that apply to both text input fields and lock buttons. */
+    field: {
```
GerardasB, line 50: Are we fine breaking the `AccuDrawViewportUI.controlProps` API marked as `@beta`?

GerardasB:
> shouldnt this be handled at the appui / desktop framework level?

We have considered setting up the iTwinUI CSS variables from AppUI layer (i.e. a function to set-up the static `AccuDrawViewportUI.controlProps` OR introducing a similarly shaped structure as a non-static property of `AccuDrawViewportUI` which could be configured via `IModelApp` startup opts), but decided that this works good enough w/o requiring any additional setup. Note, that this will use iTwinUI CSS variables where available, but fallback to default styling outside of iTwinUI.

joshuafoxley:
> I've set the opacity to `--iui-opacity-2`. We can turn it up to the next level, but at that point the opacity would be `1`.
> 
> Spacing is `4px` between rows/columns. UX will probably ask for changes on the horizontal layout spacing but we'll see.
> 
> image Screenshot 2025-09-22 at 14 10 52

Hi Carlos,

These look great, and are almost there!

Regarding padding, I recommend 1x spacing (4px) between the field and lock button, and 3x (12px) between the groupings. See the screenshots below:

Horizontal

Vertical

joshuafoxley:
There is one final tweak requested from UX, following a design review with the Strata lead designer:

Could we please change the lock state from the 'accent' green, to a more neutral white? This will help reduce visual overwhelm with all locks enabled, and distinguishes the 'active' field state, from a 'locked' state. See the annotated screenshot below for exact values, or [review the live figma file](https://www.figma.com/design/0PZ2LIrmwV7ITga7K4LxYF/precisionDrawing?node-id=2151-64687&m=dev).

Annotated

carlos-menezes:
image
image

@joshuapage-ux

joshuafoxley:
Nice!

Could we add a base fill underneath the locked buttons to match the others? So the white translucent fill (bg/glow/strong/pressed) is applied on top of the normal fill?

At the moment is appears more translucent than the others, which is not the intention.

carlos-menezes:
@joshuapage-ux I just copied the color straight from Figma. Still, I tried to make it match a bit better. Does this look better to you? 

image

joshuafoxley:
Yes much better!
Happy with this, thanks Carlos.

core/frontend/src/tools/AccuDrawViewportUI.ts
```diff
@@ -46,39 +46,97 @@ export class AccuDrawViewportUI extends AccuDraw {
...
+      /** Number of visible characters to show in text input fields. */
+      size: 12,
+      /** Height of text input fields and lock buttons. */
+      height: "var(--iui-component-height-small, 1.75em)",
```
aruniverse, line 54: what does all this look like if someone isnt using itwinui?
@GerardasB can you test that out?
instead of using hardcoded itwinui vars, could we use the actual resolved values? is that possible

GerardasB: There are resolved values in the fallback (in this case `1.75em`) https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties#defining_fallbacks_in_the_var_function

bbastings: @aruniverse @GerardasB I tested it with display-test-app, it looks the same because the fallbacks use the css variable value...

aruniverse (approved):

calebmshafer:
@Mergifyio backport release/5.2.x

PR #8533, opened 2025-09-16
try to slim down draco and make it so apps have less to do

hl662:
A Follow up PR to the issues with first party applications [not locating the draco assets ](https://github.com/iTwin/itwinjs-core/pull/8449#issuecomment-3270748577).

Overrode the urls that the loaders.gl library uses to fetch from the CDN, to point to the public path of the `.wasm` assets (dove deep into their [utility code](https://github.com/visgl/loaders.gl/blob/master/modules/worker-utils/src/lib/library-utils/library-utils.ts#L40)).

Exports the draco static assets into a new `public/draco` folder of core-frontend's build output folder.

TODO: Will write the NextVer.md to highlight that applications will need to copy this new public/draco folder's assets into their app's public asset directory.

### DIFF
--- a/common/changes/@itwin/core-frontend/arun-draco_2025-09-16-19-28.json
+++ b/common/changes/@itwin/core-frontend/arun-draco_2025-09-16-19-28.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "address loaders.gl shortcomings",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -15,7 +15,7 @@
     "copy:js:cjs": "cpx \"./src/**/*.js\" ./lib/cjs",
     "copy:js:esm": "cpx \"./src/**/*.js\" ./lib/esm",
     "copy:workers": "cpx \"./lib/workers/webpack/parse-imdl-worker.js\" ./lib/public/scripts",
-    "copy:draco": "cpx \"./node_modules/draco3d/draco_decoder.wasm\" ./lib/public",
+    "copy:draco": "cpx \"./node_modules/@loaders.gl/draco/dist/libs/*\" ./lib/public/draco",
     "docs": "betools docs --json=../../generated-docs/core/core-frontend/file.json --tsIndexFile=./core-frontend.ts --onlyJson --excludes=webgl/**/*,**/map/*.d.ts,**/tile/*.d.ts,**/*-css.ts",
     "extract-api": "betools extract-api --entry=core-frontend && npm run extract-extension-api",
     "extract-extension-api": "eslint --no-inline-config -c extraction.eslint.config.js  \"./src/**/*.ts\" 1>&2",
@@ -100,7 +100,8 @@
     "@itwin/webgl-compatibility": "workspace:*",
     "@loaders.gl/core": "^4.3.4",
     "@loaders.gl/draco": "^4.3.4",
-    "draco3d": "^1.5.1",
+    "@loaders.gl/loader-utils": "^4.3.4",
+    "draco3d": "^1.5.7",
     "fuse.js": "^3.3.0",
     "wms-capabilities": "0.4.0"
   }
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -100,7 +100,6 @@
     "@itwin/webgl-compatibility": "workspace:*",
     "@loaders.gl/core": "^4.3.4",
     "@loaders.gl/draco": "^4.3.4",
-    "@loaders.gl/loader-utils": "^4.3.4",
     "draco3d": "^1.5.7",
     "fuse.js": "^3.3.0",
     "wms-capabilities": "0.4.0"
--- a/core/frontend/src/tile/GltfReader.ts
+++ b/core/frontend/src/tile/GltfReader.ts
@@ -2172,11 +2172,6 @@ export abstract class GltfReader {
     const offset = bv.byteOffset ?? 0;
     buf = buf.subarray(offset, offset + bv.byteLength);
 
-    // const jsWrapper = await import(
-    //   `${IModelApp.publicPath}/draco/draco_wasm_wrapper.js`
-    // );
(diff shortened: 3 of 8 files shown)


core/frontend/package.json
```diff
@@ -15,7 +15,7 @@
     "copy:js:cjs": "cpx \"./src/**/*.js\" ./lib/cjs",
     "copy:js:esm": "cpx \"./src/**/*.js\" ./lib/esm",
     "copy:workers": "cpx \"./lib/workers/webpack/parse-imdl-worker.js\" ./lib/public/scripts",
-    "copy:draco": "cpx \"./node_modules/draco3d/draco_decoder.wasm\" ./lib/public",
+    "copy:draco": "cpx \"./node_modules/@loaders.gl/draco/dist/libs/*\" ./lib/public/draco",
```
markschlosseratbentley: @aruniverse - @hl662 mentioned to me that we could consider copying these to `public/scripts` instead of `public/draco`.  If that is not a heavy lift, I lean toward placing them in `public/scripts` - Overall, I vote for apps not having to change anything rather than an ideal folder organization on our end.

common/changes/@itwin/core-frontend/arun-draco_2025-09-16-19-28.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Fix loaders.gl issue with resolving draco static assets",
```
aruniverse: This was never released to users officially, so there is nothing thats actually being fixed
```suggestion
      "comment": "",
```

core/frontend/package.json
aruniverse: Even if were to copy to scripts, I now think apps would need to make some additional changes to properly fetch/resolve the asset

@hl662 can you confirm what this copies over as its not just the .wasm files, but some interop js file too right? we are copying different assets now than before (from draco3d)

core/frontend/src/tile/GltfReader.ts
```diff
@@ -2172,7 +2172,17 @@ export abstract class GltfReader {
     const offset = bv.byteOffset ?? 0;
     buf = buf.subarray(offset, offset + bv.byteLength);
 
-    const mesh = await loader.parse(buf, { modules: { draco3d } }); // NB: `options` argument declared optional but will produce exception if not supplied. Regardless, we are specifying our own bundled draco3d module to avoid CDN requests.
+    const mesh = await loader.parse(buf, {
+      draco: {
+        decoderType: "wasm",
+      },
+      modules: {
+        "draco_wasm_wrapper.js": `${IModelApp.publicPath}draco/draco_wasm_wrapper.js`,
```
aruniverse: does this actually work? is this not an issue or did i incorrect configure things before? https://github.com/visgl/loaders.gl/issues/3185

we dont need to resolve the module ourselves?

core/frontend/src/tile/GltfReader.ts
```diff
@@ -2172,7 +2172,17 @@ export abstract class GltfReader {
...
+        "draco_decoder.wasm": `${IModelApp.publicPath}draco/draco_decoder.wasm`,
+      },
+      worker: false,
+      useLocalLibraries: true,
+    }); // NB: `options` argument declared optional but will produce exception if not supplied. Regardless, we are specifying our own bundled draco3d module to avoid CDN requests.
```
aruniverse: update the comment here, were not bundling draco3d anymore

core/frontend/package.json
aruniverse: actually nvm, about apps having to make additional changes. we explicitly set the path to where the assets will be hosted so it doesnt matter

markschlosseratbentley: IIRC draco3d previously would have served up the interop js file internally from its package.

If it is indeed possible that apps could avoid having to configure anything new, that seems ideal to me.

core/frontend/src/tile/GltfReader.ts
hl662: the issue is the loaders.gl code doesn't actually want us to resolve the module ourselves. When you passed in the arrayBuffer representing the module, that gets fed into the [libraryUrl](https://github.com/visgl/loaders.gl/blob/d74df445180408fd772567ca6a6bbb8d42aa50be/modules/worker-utils/src/lib/library-utils/library-utils.ts#L40) variable in their loadLibrary() method. Then javascript, expecting a primitive string value, threw a `Cannot convert object to primitive value`. So libraryUrl does, actually, mean a url to the relevant assets, so I changed it to pass in the url to fetch the asset, and laoders.gl uses that url to call `loadLibraryFromFile`, the RHS of the `||` operator in the link above

aruniverse: ok maybe respond to my comment on the issue then lol; nice find

core/frontend/package.json
```diff
@@ -100,7 +100,7 @@
     "@itwin/webgl-compatibility": "workspace:*",
     "@loaders.gl/core": "^4.3.4",
     "@loaders.gl/draco": "^4.3.4",
-    "draco3d": "^1.5.1",
+    "draco3d": "^1.5.7",
```
markschlosseratbentley: If we are not bundling draco3d anymore, do we still need this dep?

core/frontend/package.json
hl662: alright, moved the assets into scripts/, changed the urls GltfReader uses to point there too . Yes, there are some interop .js files we copy as well.

core/frontend/src/tile/GltfReader.ts
hl662: updated

core/frontend/package.json
hl662: removed

hl662:
> just make sure weve tested this with dta & our consumer apps via pnpm patch; or we can wait for dev release again

I used pnpm-patch to test with our consumer app locally, we could also wait for another dev release afterwards

core/frontend/package.json
markschlosseratbentley: Looks good, thanks

core/frontend/package.json
markschlosseratbentley: Looks good, thanks

aruniverse (approved):

Issue #8496, opened 2025-09-08
Expose classes and methods needed to implement RenderSystem and related classes outside of core


markschlosseratbentley:
This item is related to the work of making the necessary iTwin.js changes to enable a CesiumJS render system implementation.

We put together a barebones prototype of a CesiumJS render system in display-test-app.  This necessitated some hacky API changes to allow the implementation.

Looking at [this diff](https://github.com/iTwin/itwinjs-core/compare/master...cesium-renderer), consider every API change that was necessary to implement that barebones prototype in display-test-app and consider whether / how they should be done for real in iTwin.js itself.

Implement the necessary changes in iTwin.js.

markschlosseratbentley:
See https://github.com/iTwin/itwinjs-core/pull/8535#issuecomment-3303359657 for details on how to work around many of these issues.

Issue #8525, opened 2025-09-15
Viewport rendersToScreen behavior should be customizeable or reassessed


markschlosseratbentley:
In the context of implementing a CesiumJS rendering subsystem within iTwin.js, we need to consider a way to customize `core/frontend/src/Viewport.ts`'s getter and setter for `rendersToScreen`. The rendering-to-screen optimization within iTwin.js allows us to render directly to the screen canvas in the case where only one viewport exists.  In the past, this showed a performance benefit in certain browsers.

In the context of CesiumJS, the resulting shuffling around of canvases will cause complexities.

Options:

1. Allow the `rendersToScreen` behavior to be circumvented / customized via API so a customized CesiumJS renderer can disable the canvas-shuffling.
2. Assess whether the optimization makes sense anymore - perhaps it can be removed to make all cases (including CesiumJS simpler).

Issue #8498, opened 2025-09-08
Devise an appropriate abstraction level to enable the cesium-renderer package to render iTwin.js tile trees


markschlosseratbentley:
This item is related to the work of making the necessary iTwin.js changes to enable a CesiumJS render system implementation.

The idea of this work item is to 1) plan and 2) implement an appropriate abstraction level that allows the `cesium-renderer` package used by iTwin.js to use CesiumJS to render reality tile trees (3D Tiles), namely using relevant CesiumJS APIs for tile tree parsing, instead of iTwin.js' own tile tree parsing.

This work item is not necessarily about the actual implementation, which will explore using the CesiumJS Scene API to attach a 3D tileset to the cesium scene. Rather, this is about hooking up the wiring such that reality tile trees created via iTwin.js APIs will be able to be powered by the API in the new `cesium-renderer` package.  That package will handle all the implementation details of calling CesiumJS properly.

PR #8535, opened 2025-09-17
Rework frontend rendering APIs so RenderSystem and related classes can be implemented outside core-frontend package

markschlosseratbentley:
## Description

Fixes https://github.com/iTwin/itwinjs-core/issues/8496

This PR works around some key hurdles encountered when creating the [CesiumJS/iTwin.js Render System Prototype](https://github.com/iTwin/itwinjs-core/compare/master...cesium-renderer) branch. These hurdles had to do with implementing `RenderSystem` and related classes outside of the `core-frontend` package.

Before we officially merge that prototype work into the [upcoming `cesium-renderer` package](https://github.com/iTwin/itwinjs-core/pull/8489), we would like to merge these key `core-frontend` changes that seem to be necessary to implement `RenderSystem` and its related classes outside of the `core-frontend` package.

## Changes

Major changes that were found necessary:
- `RenderPlan`, `RenderAreaPattern`, and `RenderGeometry` need to be publicly exposed via the `core-frontend` package.  This is because importing them into display-test-app directly from the `internal/` directory in `core-frontend` results in them being considered different types than their types referenced by `RenderSystem` which is imported via the `core-frontend` package.  
  - TL;DR: A class and an implementation of that class should pull any types they both reference from the same package / location.
  - This worked fine in the past when these implementations only existed within `core-frontend` itself.
  - `PrimitiveBuilder` was originally also exposed in the prototype but we should be able to implement our own `GraphicBuilder` instead.
- [`_implementationProhibited`](https://github.com/iTwin/itwinjs-core/blob/f68e3bd17e0c54a72d67b2b700e4499356ec078a/core/backend/src/internal/Symbols.ts#L9) needs to be removed from `RenderTarget`. This is because we are now implementing this class outside the `core-frontend` package, which this property prevents.
- For consistency, certain files that were affected were moved from `src/internal/render/` to `src/render/`.

## Questions

- Should the APIs that have been newly exposed be marked `internal`, `public`, or...? `RenderSystem` is already `public`, so there is that precedent.  For this first cut, I conservatively tagged any now-exposed API as `internal`.

## Remaining Work

In followup PRs:
- [Viewport rendersToScreen behavior should be customizeable or reassessed](https://github.com/iTwin/itwinjs-core/issues/8525)
- [Devise an appropriate abstraction level to enable the cesium-renderer package to render iTwin.js tile trees](https://github.com/iTwin/itwinjs-core/issues/8498)


pmconne:
We already have a [pattern](https://www.itwinjs.org/learning/guidelines/release-tags-guidelines/#internal-apis) for internal APIs that are used across packages in the itwinjs-core monorepo. Those APIs remain in <package>/src/internal/ and are selectively exported from <package>/src/internal/cross-package.ts. This can include `_implementationProhibited` for `RenderTarget`.

(Next time let's start an issue to discuss potential solutions before implementing proposed solutions).

markschlosseratbentley:
> We already have a [pattern](https://www.itwinjs.org/learning/guidelines/release-tags-guidelines/#internal-apis) for internal APIs that are used across packages in the itwinjs-core monorepo. Those APIs remain in /src/internal/ and are selectively exported from /src/internal/cross-package.ts. This can include `_implementationProhibited` for `RenderTarget`.
> 
> (Next time let's start an issue to discuss potential solutions before implementing proposed solutions).

I will take a look at that, thanks @pmconne. I agree, we should have discussed this on https://github.com/iTwin/itwinjs-core/issues/8496 - next time I will add more detail about the exact problems.

PR #8540, opened 2025-09-18
BriefcaseManager provides 'deviceName' when acquiring a Briefcase  

MichaelSwigerAtBentley:
Adds an optional parameter 'deviceName' to 'acquireNewBriefcaseId'
BriefcaseManager now automatically provides `os.hostname()` as deviceName to all newly acquired Briefcases

This is useful for debugging purposes.

`BackendHubAccess` implementations will still need to do something with deviceName, such as pass it along to iModleHub, but at least it will be provided. 

--------------------------------------------

Resolves issue: https://github.com/iTwin/itwinjs-backlog/issues/1620


core/backend/src/BriefcaseManager.ts
```diff
@@ -229,7 +230,7 @@ export class BriefcaseManager {
    * @note *It is invalid to edit briefcases on a shared network drive* and that is a sure way to corrupt your briefcase (see https://www.sqlite.org/howtocorrupt.html)
    */
   public static async downloadBriefcase(arg: RequestNewBriefcaseArg): Promise<LocalBriefcaseProps> {
-    const briefcaseId = arg.briefcaseId ?? await this.acquireNewBriefcaseId(arg);
+    const briefcaseId = arg.briefcaseId ?? await this.acquireNewBriefcaseId({...arg, deviceName: os.hostname()});
```
karolis-zukauskas: The way it is written now, applications cannot override `deviceName` even though it is available to be set in `RequestNewBriefcaseArg`. You want: `{ deviceName: os.hostname(), ...arg }` (expand `arg` after).

karolis-zukauskas:
> changes lgtm, but wait on @karolis-zukauskas to approve he might be looking for additional details like:
> 
> * [os.type](https://nodejs.org/api/os.html#ostype)
> * [os.arch](https://nodejs.org/api/os.html#osarch)
> * username from [os.userInfo](https://nodejs.org/api/os.html#osuserinfooptions)

@aruniverse @MichaelSwigerAtBentley 

Lets go for combo: `` `${os.hostname()}:${os.type()}:${os.arch()}` `` - I think hostname should be enough, but it doesn't hurt to add the other two. `username` should NOT be used as that is PII

core/backend/src/BriefcaseManager.ts
ColinKerr: The test should cover the case where the caller provides a deviceName in the args

karolis-zukauskas: [TS Playground](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgIJQOYGdkG8CwAUMicnAPwBcyIArgLYBG0ANEaco1TQ81G8VIJuWMFFAYBAXyJEEAexCiymaumzIAvHjLUAjC07UATIYTUARAFkAmsgBqqADIBVAKIXkUgNyzCCpXkAGwgAOiD5DAAKCwgADwAHOBAAE04IGHkoCAtDXGRQwrhMM0sUjLhaILBPKQBKX39FLGCwiOjYxOS0uBhIKFydc2QLcvgqmsNC0OKMLwaiIA)

MichaelSwigerAtBentley: Corrected the expand, good catch. 

Added custom deviceName to the test as well.

MichaelSwigerAtBentley:
Swapped so `${os.hostname()}:${os.type()}:${os.arch()}` is the defualt value

Issue #8497, opened 2025-09-08
Devise an appropriate abstraction level to enable the cesium-renderer package to render iTwin.js decorations


markschlosseratbentley:
This item is related to the work of making the necessary iTwin.js changes to enable a CesiumJS render system implementation.

The idea of this work item is to 1) plan and 2) implement an appropriate abstraction level that allows the cesium-renderer package used by iTwin.js to use CesiumJS to create and render decorations using relevant CesiumJS APIs.

This work item is not necessarily about the actual implementation, which is exploring the usage of Entity and Primitive. Rather, this is about hooking up the wiring such that decorations created via iTwin.js APIs will be able to be powered by the API in the new `cesium-renderer` package.

PR #8544, opened 2025-09-19
First milestone cut of cesium-renderer display-test-app prototype

markschlosseratbentley:
Fixes https://github.com/iTwin/itwinjs-core/issues/8525
Fixes https://github.com/iTwin/itwinjs-core/issues/8496
Fixes https://github.com/iTwin/itwinjs-core/issues/8497

## Description

This pull request moves and expands upon the original CesiumJS/iTwin.js prototype work which at the time placed all the CesiumJS rendering implementations inside the display-test-app directory.  Now, all of that code is moved to the `cesium-renderer` package. A barebones, WIP implementation of CesiumJS rendering will be viewable in display-test-app when configured a certain way (See section "Running").

## Changes

The `cesium-renderer` package now exposes:
- `createCesiumRenderSystem()` - a function which returns a prototype CesiumJS implementation of `RenderSystem`
- `createCesiumViewManager()` - a function which returns a `ViewManager` implementation which overrides `updateRenderToScreen()` so CesiumJS can avoid being affected by the canvas-shuffling optimizations done internally by iTwin.js in the normal WebGL rendering path.

`display-test-app`, when configured to use this CesiumJS prototype, will instantiate those two types and pass them into the proper iTwin.js startup routines. These classes will then be used instead of the typical iTwin.js implementations of those classes.

Prototype decorations now also render in the viewport using CesiumJS. These can be viewed by running the prototype. These test decorations are created using the iTwin.js decoration APIs, but end up getting rendered via CesiumJS internally.

## Running

From your command line, within a cloned copy of the iTwin.js monorepo:

1. `git checkout cesium-renderer`
2. `rush install`
3. `rush build --to display-test-app`
4. `set IMJS_USE_CESIUM=1`
5. `cd test-apps/display-test-app`
6. `npm run start`
7. Click the star-shaped icon in the top toolbar of the UI.
8. Observe that the Cesium globe renders. Some test decorations are rendered on top of it.
9. To navigate in the Cesium camera, regular iTwin controls do not currently work. Instead:
  a. Press `ctrl` + `w`, `a`, `s`, `d` to move in and out and rotate
  b. Press `ctrl` + `arrow keys` to pan

image

image

image

## TODO

- [x] Add unit tests in `cesium-renderer`
- [x] Run Image Tests


extensions/cesium-renderer/src/System.ts
```diff
@@ -0,0 +1,118 @@
...
+
+import { ElementAlignedBox3d, RenderFeatureTable } from "@itwin/core-common";
+import { Transform } from "@itwin/core-geometry";
+import { BatchOptions, CreateGraphicFromTemplateArgs, CustomGraphicBuilderOptions, GraphicBranch, GraphicBranchOptions, GraphicBuilder, IModelApp, IModelConnection, InstancedGraphicParams, RenderAreaPattern, RenderGeometry, RenderGraphic, RenderSystem, RenderTarget, ViewportGraphicBuilderOptions, ViewRect } from "@itwin/core-frontend";
+import { OffScreenTarget, OnScreenTarget } from "./Target.js";
+import { CesiumGraphic } from "./Graphic.js";
+import { PrimitiveConverterFactory } from "./decorations/PrimitiveConverterFactory.js";
+
+/** @internal */
+export class System extends RenderSystem {
```
eringram: If we use `CesiumViewManager` and `CesiumGraphic`, I feel like it makes sense to also use the Cesium prefix here for something like `CesiumRenderSystem`

Or is this intentionally the same name as the `System` class in core/frontend/src/internal/render/webgl/System.ts?

extensions/cesium-renderer/src/System.ts
```diff
@@ -0,0 +1,118 @@
...
+    // Pass the single geometry to CesiumGraphic
+    return new CesiumGraphic([geometry], geometry.renderGeometryType);
+  }
+
+  public createGraphicList(_primitives: RenderGraphic[]): RenderGraphic {
+    // ##TODO implement logic to create a graphic list from the array of primitives.
+    return new CesiumGraphic();
+  }
+
+  public createGraphicBranch(_lBranch: GraphicBranch, _transform: Transform, _options?: GraphicBranchOptions): RenderGraphic {
```
eringram:
```suggestion
  public createGraphicBranch(_branch: GraphicBranch, _transform: Transform, _options?: GraphicBranchOptions): RenderGraphic {
```
Might be a typo

core/frontend/src/Viewport.ts
```diff
@@ -68,6 +68,7 @@ import { FlashSettings } from "./FlashSettings";
 import { GeometricModelState } from "./ModelState";
 import { GraphicType } from "./common/render/GraphicType";
 import { compareMapLayer } from "./internal/render/webgl/MapLayerParams";
+import { System } from "./internal/render/webgl/System";
```
eringram: Don't see it used so I think this can be removed

extensions/cesium-renderer/package.json
```diff
@@ -38,18 +39,26 @@
...
     "@types/node": "~20.17.0",
+    "cpx2": "^8.0.0",
     "eslint": "^9.31.0",
     "internal-tools": "workspace:*",
     "rimraf": "^6.0.1",
     "typescript": "~5.6.2",
     "vitest": "^3.0.6"
+  },
+  "dependencies": {
+    "cesium": "1.133.0"
```
eringram: We need to ask core reviewers if this should be in dependencies, or peerDependencies/devDependencies. I _think_ dependencies is correct, because unlike the iTwin packages listed in peerDeps, we would not expect an app that starts consuming this package to already provide cesium

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  /**
+   * Create CesiumJS camera from iTwin.js ViewDefinition
+   * Based on your provided example with spatialToEcef integration
+   * @param viewDefinition iTwin.js ViewDefinition3dProps
+   * @param ecefLoc Optional EcefLocation override
+   * @param modelExtents Optional model extents override
+   * @returns CesiumCamera parameters
+   */
+  public createCesiumCamera(
```
eringram: Duplicate, can import from CesiumCamera.ts

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
```diff
@@ -0,0 +1,321 @@
...
+
+import { Cartographic, EcefLocation, ViewDefinition3dProps } from "@itwin/core-common";
+import { Point3d, Range3d, Vector3d, XYAndZ, YawPitchRollAngles } from "@itwin/core-geometry";
+import { IModelConnection } from "@itwin/core-frontend";
+import { Cartesian3, Cartographic as CesiumCartographic } from "cesium";
+
+/**
+ * Interface for CesiumJS frustum parameters
+ */
+interface CesiumFrustum {
```
eringram: This and `CesiumCamera` should be imported from CesiumCamera.ts, and they have been renamed to `CesiumFrustumProps` and `CesiumCameraProps`

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
```diff
@@ -0,0 +1,321 @@
...
+ */
+interface CesiumCamera {
+  position: Point3d;
+  direction: Vector3d;
+  up: Vector3d;
+  frustum: CesiumFrustum;
+}
+
+/**
+ * Professional coordinate conversion utilities between iTwin.js and CesiumJS
```
eringram: Nit but not sure it's accurate to call this professional when it's still WIP 😄

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
```diff
@@ -0,0 +1,321 @@
...
+export class CesiumCoordinateConverter {
+  private _iModel: IModelConnection;
+  private _ecefLocation?: EcefLocation;
+
+  constructor(iModel: IModelConnection) {
+    this._iModel = iModel;
+    this._ecefLocation = iModel.ecefLocation;
+    
+    if (!this._iModel.isGeoLocated) {
+      console.warn('IModel is not geo-located, coordinate conversions may not be accurate');
```
eringram: Do you use null island/lat long 0 if the iModel is not geolocated? We should probably be clarify what happens in that case throughout the methods in this file. Should it simply not be called without a geolocation, if the result is going to be inaccurate?

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
```diff
@@ -0,0 +1,321 @@
...
+  direction: Vector3d;
+  up: Vector3d;
+  frustum: CesiumFrustum;
+}
+
+/**
+ * Professional coordinate conversion utilities between iTwin.js and CesiumJS
+ * Handles spatial coordinate transformations, camera conversions, and ECEF mappings
+ */
+export class CesiumCoordinateConverter {
```
eringram, line 15: This would be a good class to add some unit tests for

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
```diff
@@ -0,0 +1,321 @@
...
+  /**
+   * Convert CesiumJS Cartesian3 (ECEF) to iTwin.js spatial coordinates
+   * Uses iTwin.js built-in ecefToSpatial method
+   * @param cartesian3 Position in CesiumJS ECEF coordinates
+   * @returns Point3d in iTwin.js spatial coordinates
+   */
+  public cesiumCartesian3ToSpatial(cartesian3: Cartesian3): Point3d {
+    if (!this._iModel.isGeoLocated) {
+      console.warn('IModel not geo-located, conversion may not be accurate');
+      return new Point3d(cartesian3.x, cartesian3.y, cartesian3.z);
```
eringram: Should this also use the same `_getFallbackCartesian3` method?

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
```diff
@@ -0,0 +1,321 @@
...
+  private _getFallbackCartesian3(spatial: XYAndZ): Cartesian3 {
+    const center = this._iModel.projectExtents.center;
+    
+    // Calculate relative position from model center
+    const relativeX = spatial.x - center.x;
+    const relativeY = spatial.y - center.y;
+    const relativeZ = spatial.z - center.z;
+    
+    // Convert to approximate geographic coordinates
+    const longitude = relativeX * 0.00001; // Rough meters to degrees
```
eringram: I'm a little confused where in practice this would place iModels on the earth and how this conversion happens. Maybe some tests would help clarify?

extensions/cesium-renderer/src/decorations/CoordinateBuilder.ts
```diff
@@ -0,0 +1,136 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
```
eringram: I believe without a `@docs-package-description`, such as this [core-frontend.ts](https://github.com/iTwin/itwinjs-core/blob/master/core/frontend/src/core-frontend.ts#L192) example, this doesn't do anything. It's also for the purpose mainly of public docs, so I think we should remove these.

markschlosseratbentley:
@pmconne @eringram Please feel free to give this an early review before we mark it fully ready for review, hopefully next week.

extensions/cesium-renderer/src/decorations/LineStringPrimitiveConverter.ts
```diff
@@ -0,0 +1,171 @@
...
+import { IModelConnection } from "@itwin/core-frontend";
+import { Point3d } from "@itwin/core-geometry";
+import { CesiumScene } from "../CesiumScene.js";
+import { PrimitiveConverter, RenderGraphicWithCoordinates } from "./PrimitiveConverter.js";
+import { DecorationPrimitiveEntry } from "./DecorationTypes.js";
+
+/** Converts iTwin.js LineString decorations to Cesium Polylines */
+export class LineStringPrimitiveConverter extends PrimitiveConverter {
+  protected readonly primitiveType = 'linestring';
+
```
eringram, line 27:
```suggestion
```
Nit: remove extra whitespace here and elsewhere in the file

extensions/cesium-renderer/src/decorations/LineStringPrimitiveConverter.ts
```diff
@@ -0,0 +1,171 @@
...
+        const firstLineString = originalLineStrings[0];
+        if (firstLineString && firstLineString.length > 0) {
+          positions = this.convertPointsToCartesian3(firstLineString, iModel);
+        }
+      }
+      
+      if (positions.length === 0 && geometries && geometries.length > 0) {
+        const geometry = geometries[0];
+        interface Coord { x: number; y: number; z: number }
+        const hasCoords = (g: unknown): g is { coordinateData: Coord[] } =>
```
eringram: Is there a way we can make `geometries` an array of `{ coordinateData: Coord[] }`, so it doesn't have to be type `unknown`?

extensions/cesium-renderer/src/decorations/LineStringPrimitiveConverter.ts
```diff
@@ -0,0 +1,171 @@
...
+            id: lineId,
+            positions,
+            width: 2,
+            material: Material.fromType(Material.ColorType, { color }),
+            ...this.getDepthOptions(type || 'world'),
+          });
+      }
+  }
+
+  private extractColorFromGraphic(graphic?: RenderGraphicWithCoordinates): Color | undefined {
```
eringram: There is a lot of overlap between this and the other converters' `extractColorFromGraphic` methods. Maybe we can try to combine some of the functionality into the parent `PrimitiveConverter` class.

eringram (commented):
I'm still reviewing the decorations directory, but adding these comments now so we can start discussing

extensions/cesium-renderer/src/System.ts
markschlosseratbentley: I intentionally named this the same as the class in `webgl/System.ts`, since there seems to be a pattern of that when inheriting from `RenderSystem`.  You can note that in DTA, I actually import it _as_ a `CesiumSystem`.

I would have named `CesiumViewManager` as a `ViewManager` as well, but since it inherits from `ViewManager`, that would not work, hence the Cesium prefix.

That said, I am open to reconsidering the naming.  You, @eringram, have already established a pattern of `Cesium` prefixes for things like `CesiumCameraProps`, especially when they are formatted to be fed into the CesiumJS system.

@danielzhong do you have any thoughts?

extensions/cesium-renderer/src/decorations/ShapePrimitiveConverter.ts
```diff
@@ -0,0 +1,220 @@
...
+/** @packageDocumentation
+ * @module Cesium
+ */
+
+import { Cartesian3, Color, ColorGeometryInstanceAttribute, GeometryInstance, PerInstanceColorAppearance, PolygonGeometry, PolygonHierarchy, Primitive, PrimitiveCollection } from "cesium";
+import { IModelConnection } from "@itwin/core-frontend";
+import { Point3d } from "@itwin/core-geometry";
+import { CesiumScene } from "../CesiumScene.js";
+import { PrimitiveConverter, RenderGraphicWithCoordinates } from "./PrimitiveConverter.js";
+import { ColorDef } from "@itwin/core-common";
```
eringram: Nit: package import statements should go above imports from files for consistency

extensions/cesium-renderer/src/decorations/ArcPrimitiveConverter.ts
eringram, line 1: The copyright header should be added

extensions/cesium-renderer/src/decorations/ArcPrimitiveConverter.ts
```diff
@@ -0,0 +1,269 @@
...
+  protected readonly primitiveType = 'arc';
+  private _currentScene?: CesiumScene;
+
+  protected override getCollection(scene: CesiumScene): PrimitiveCollection {
+    this._currentScene = scene; // Store scene reference for later use
+    return scene.primitivesCollection;
+  }
+
+  protected override extractPrimitiveData(coordinateData: DecorationPrimitiveEntry[] | undefined, primitiveType: string): DecorationPrimitiveEntry[] | undefined {
+    if (!coordinateData || !Array.isArray(coordinateData)) {
```
eringram: If `coordinateData` has to be defined here why make it `| undefined` in the parameters? Could we make it required to be a `DecorationPrimitiveEntry[]` in this parameter list, and check if it's undefined before this method is called, in `PrimitiveConverter.convertDecorationsTemplate()`?

extensions/cesium-renderer/src/System.ts
markschlosseratbentley: Auto-resolved by committing the suggestion.

extensions/cesium-renderer/src/decorations/LoopPrimitiveConverter.ts
```diff
@@ -0,0 +1,278 @@
...
+import { Cartesian3, Color, ColorGeometryInstanceAttribute, GeometryInstance, PerInstanceColorAppearance, PolygonGeometry, PolygonHierarchy, Primitive } from "cesium";
+import { CesiumScene } from "../CesiumScene.js";
+import { PrimitiveConverter, RenderGraphicWithCoordinates } from "./PrimitiveConverter.js";
+import { DecorationPrimitiveEntry } from "./DecorationTypes.js";
+
+export class LoopPrimitiveConverter extends PrimitiveConverter {
+  protected readonly primitiveType = 'loop' as const;
+  private _currentScene?: CesiumScene;
+
+  protected override getCollection(scene: CesiumScene): import('cesium').PrimitiveCollection {
```
eringram: Should import `PrimitiveCollection` from cesium at the top of file, also in other instances in this file like L37. Just in general should remove all inline imports like this

core/frontend/src/Viewport.ts
markschlosseratbentley: Thanks for the catch. Removed.

extensions/cesium-renderer/package.json
markschlosseratbentley: Tagging @aruniverse and @ben-polinsky for early opinions on this particular topic, even though the PR is not fully ready for review.

extensions/cesium-renderer/src/decorations/PolyfacePrimitiveConverter.ts
```diff
@@ -0,0 +1,243 @@
...
+    graphic: RenderGraphicWithCoordinates,
+    primitiveId: string,
+    _index: number,
+    _collection: import('cesium').PrimitiveCollection,
+    iModel?: IModelConnection,
+    originalData?: unknown,
+    _type?: string
+  ): Primitive | null {
+    const data = Array.isArray(originalData) ? (originalData as DecorationPrimitiveEntry[]) : undefined;
+    const isPolyfaceEntry = (e: DecorationPrimitiveEntry): e is import('./DecorationTypes.js').PolyfaceEntry => e.type === 'polyface';
```
eringram: Same issue where import should be at the top of the file. Also, imo `isPolyfaceEntry` is unnecessary as all you need is `return e.type === "polyface"` on the next line.

extensions/cesium-renderer/src/decorations/PolyfacePrimitiveConverter.ts
```diff
@@ -0,0 +1,243 @@
...
+    return coordinateData.filter((entry: DecorationPrimitiveEntry) => entry.type === primitiveType);
+  }
+
+  protected override createPrimitiveFromGraphic(
+    graphic: RenderGraphicWithCoordinates,
+    primitiveId: string,
+    _index: number,
+    _collection: import('cesium').PrimitiveCollection,
+    iModel?: IModelConnection,
+    originalData?: unknown,
```
eringram: Why not make this type `DecorationPrimitiveEntry[] | undefined`? if it's not of that type it becomes undefined, and if I follow the logic, if `data` is undefined then `polyface` becomes undefined on L52, and the function terminates on L56. We could just check for undefined earlier on and simplify this code in general.

extensions/cesium-renderer/src/decorations/LineStringPrimitiveConverter.ts
danielzhong: I think my newest changes handled this issue. I will work on other issues now. Thanks for the feedback!

extensions/cesium-renderer/src/System.ts
markschlosseratbentley: @eringram @danielzhong Today, I went ahead and made sure we consistently use the Cesium prefix for core system, target, etc. classes. I think @eringram is right and it feels a bit jarring that some core classes have the prefix and some do not, even if it was by necessity. @eringram please resolve this comment if you are satisfied.

danielzhong: Looks good to me! Thanks for updating the naming and keep everything consistently prefixed with Cesium!

extensions/cesium-renderer/package.json
ben-polinsky: > we would not expect an app that starts consuming this package to already provide cesium

If this is a correct statement, then yes, regular ol' dependencies is correct. 

Edit: And I agree it seems to be correct :)

extensions/cesium-renderer/src/decorations/CoordinateBuilder.ts
ben-polinsky: Correct, assuming there's not an annoying bug in BeMetalsmith.

extensions/cesium-renderer/package.json
aruniverse: + @jason-crow @matmarchand who have been pulling cesium into some of their pkgs

also + @ggetz 

only concern w/ cesium as direct dependency is what happens if multiple versions of the lib get installed? if it manages state internally, i imagine wed probably want it as a peer. 

we might just have to play around with it a bit and find out what the right answer here is

markschlosseratbentley: If a hard dependency is a concern for right now, we could add cesium as a dependency of display-test-app and then a peerDependency of this package.

Note: eventually cesium will need to be a real dependency of iTwin.js, I think, unless we expect every app to have a cesium version in their own package.json. I think we probably want to avoid that because then we can't easily dictate what cesium version to use and we can't ensure certain versions will work flawlessly in iTwin.js like we can if we put our own version in place.

matmarchand: Do you need the whole thing or just the engine?
In our case, we only need it for styling expressions parser and have a direct dependency on "@cesium/engine".

extensions/cesium-renderer/src/System.ts
eringram: Looks good, thanks

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
```diff
@@ -0,0 +1,279 @@
...
+
+  constructor(iModel: IModelConnection) {
+    this._iModel = iModel;
+    this._ecefLocation = iModel.ecefLocation;
+    
+  }
+
+  /**
+   * Convert iTwin.js spatial coordinates to CesiumJS Cartesian3 (ECEF coordinates)
+   * Uses iTwin.js built-in spatialToEcef method from IModel.ts:636
```
eringram: Nit: this and some of the other docs/comments in `cesiumCartesian3ToSpatial` and `spatialToGeographic` feel unnecessary and add clutter to these functions. For example, there's no need to document what `spatialToEcef` does or its file/line number when someone can just check the import statement.

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
```diff
@@ -0,0 +1,279 @@
...
+      const ecefPoint = this._iModel.spatialToEcef(point);
+      cesiumPositions.push(new Cartesian3(ecefPoint.x, ecefPoint.y, ecefPoint.z));
+    }
+    
+    return cesiumPositions;
+  }
+
+  /**
+   * Create CesiumJS camera from iTwin.js ViewDefinition
+   * Based on your provided example with spatialToEcef integration
```
eringram:
```suggestion
```

extensions/cesium-renderer/src/decorations/LineStringPrimitiveConverter.ts
```diff
@@ -0,0 +1,143 @@
...
+      return {
+        ...baseOptions,
+        clampToGround: false,
+        distanceDisplayCondition: undefined,
+      };
+    }
+    return baseOptions;
+  }
+
+
```
eringram:
```suggestion
```

extensions/cesium-renderer/src/decorations/ArcPrimitiveConverter.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  protected override createPrimitiveFromGraphic(
+    graphic: RenderGraphicWithCoordinates,
+    primitiveId: string,
+    _index: number,
+    _collection: PrimitiveCollection,
+    iModel?: IModelConnection,
+    originalData?: unknown,
+    _type?: string
+  ): Primitive | Polyline | null {
```
eringram: Use [undefined instead of null](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/guidelines/typescript-coding-guidelines.md#do-not-use-null) where possible (surprised linter doesn't flag this)

extensions/cesium-renderer/src/decorations/SolidPrimitivePrimitiveConverter.ts
```diff
@@ -0,0 +1,252 @@
...
+      minimum: new Cartesian3(-width/2, -height/2, -depth/2),
+      maximum: new Cartesian3(width/2, height/2, depth/2),
+      vertexFormat: VertexFormat.POSITION_AND_NORMAL,
+    });
+
+    // Create translation using Cesium coordinates (keep consistent with other converters)
+    const converter = iModel ? new CesiumCoordinateConverter(iModel) : undefined;
+    const cesiumCenter = converter
+      ? converter.spatialToCesiumCartesian3(new Point3d(centerX, centerY, centerZ))
+      : new Cartesian3(centerX, centerY, centerZ);
```
eringram: I'm unclear what the effect of falling back to `new Cartesian3(centerX, centerY, centerZ)` is. If that center point is in iTwin spatial coordinates, wouldn't this just forgo a conversion and treat that center point as though it's in ECEF (what Cesium expects), which would have odd results? 

Let me know if I'm misunderstanding here

extensions/cesium-renderer/src/decorations/SolidPrimitivePrimitiveConverter.ts
```diff
@@ -0,0 +1,252 @@
...
+    return { geometry: boxGeometry, modelMatrix };
+  }
+
+  private convertSphereToGeometry(sphere: Sphere, iModel?: IModelConnection): { geometry: SphereGeometry; modelMatrix: Matrix4 } | null {
+    // Get sphere properties using iTwin.js methods
+    const center = sphere.cloneCenter();
+
+    // Use the radius from the sphere creation (which should be 15000)
+    let radius = 15000; // Use the radius from CesiumDecorator
+
```
eringram, line 165:
```suggestion
```

extensions/cesium-renderer/src/decorations/SolidPrimitivePrimitiveConverter.ts
```diff
@@ -0,0 +1,252 @@
...
+    const transform = sphere.getConstructiveFrame();
+    if (transform) {
+      const matrix = transform.matrix;
+      // Extract scale from transformation matrix (approximate radius)
+      const scaleX = Math.sqrt(matrix.at(0,0)**2 + matrix.at(1,0)**2 + matrix.at(2,0)**2);
+      if (scaleX > 1000) { // Only use if it seems reasonable
+        radius = scaleX;
+      }
+    }
+
```
eringram, line 175:
```suggestion
```

extensions/cesium-renderer/src/decorations/CoordinateStorage.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Cesium
+ */
```
eringram:
```suggestion
```

extensions/cesium-renderer/src/decorations/CoordinateStorage.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/** Generic coordinate storage for captured decoration coordinate data */
+export class CoordinateStorage {
+  private static _coordinateData = new Map<symbol, import('./DecorationTypes.js').DecorationPrimitiveEntry[]>();
```
eringram: Import should be at the top of the file, not inline

extensions/cesium-renderer/src/decorations/PathPrimitiveConverter.ts
```diff
@@ -0,0 +1,94 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Cesium
+ */
```
eringram:
```suggestion
```

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
```diff
@@ -0,0 +1,145 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Cesium
+ */
```
eringram:
```suggestion
```

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
```diff
@@ -0,0 +1,145 @@
...
+
+import { Cartesian3, Color, PointPrimitive, PointPrimitiveCollection } from "cesium";
+import { IModelConnection } from "@itwin/core-frontend";
+import { Point3d } from "@itwin/core-geometry";
+import { CesiumScene } from "../CesiumScene.js";
+import { PrimitiveConverter, RenderGraphicWithCoordinates } from "./PrimitiveConverter.js";
+
```
eringram, line 19:
```suggestion
```

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
```diff
@@ -0,0 +1,145 @@
...
+/** Converts iTwin.js point decorations to Cesium PointPrimitives */
+export class PointPrimitiveConverter extends PrimitiveConverter {
+  protected readonly primitiveType: 'pointstring' | 'pointstring2d';
+
+  public constructor(primitiveType: 'pointstring' | 'pointstring2d' = 'pointstring') {
+    super();
+    this.primitiveType = primitiveType;
+  }
+
```
eringram, line 34:
```suggestion
```

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  protected override getCollection(scene: CesiumScene): PointPrimitiveCollection {
+    return scene.pointCollection;
+  }
+
```
eringram, line 38:
```suggestion
```

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (isOverlay) {
+      return {
+        ...baseOptions,
+        disableDepthTestDistance: Number.POSITIVE_INFINITY,
+      };
+    }
+    return baseOptions;
+  }
+
+
```
eringram:
```suggestion
```

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
```diff
@@ -0,0 +1,145 @@
...
+      entityPosition = positions[0];
+    }
+
+    if (!entityPosition)
+      return null;
+
+    const color = this.extractLineColorFromGraphic(graphic, this.primitiveType);
+    if (!color)
+      return null;
+    switch (geometryType) {
```
eringram: Any particular reason to use a switch statement with only one case here? The `default` logic happens no matter the type of `geometryType`, right?

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
```diff
@@ -0,0 +1,145 @@
...
+      if (firstPointString && firstPointString.length > 0) {
+        const candidate = firstPointString[0];
+        const z = typeof candidate.z === 'number' ? candidate.z : 0;
+        realSpatialPoint = Point3d.create(candidate.x, candidate.y, z);
+      }
+    }
+
+    if (!realSpatialPoint && geometries && geometries.length > 0) {
+      const geometry = geometries[0];
+      const firstCoord = (geometry as { coordinateData?: Array<{ x: number; y: number; z: number }> })?.coordinateData?.[0];
```
eringram: Is there a way to have `geometry` not `unknown` so there isn't a need for all this casting?

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
danielzhong: Removed

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
danielzhong: Forgot to clean up, fixed!

extensions/cesium-renderer/src/decorations/CoordinateStorage.ts
danielzhong: Fixed!

extensions/cesium-renderer/src/decorations/ArcPrimitiveConverter.ts
danielzhong: Fixed!

extensions/cesium-renderer/src/decorations/ArcPrimitiveConverter.ts
danielzhong: Added!

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
danielzhong: Indeed! Most of them just note for myself, forgot to remove. Thanks!

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
danielzhong: Modified!

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
eringram: Thanks. But after the change, what's the point of this function now that it just calls `createCesiumCameraProps` and does nothing else? It also doesn't seem to be used anywhere.

extensions/cesium-renderer/src/decorations/CoordinateBuilder.ts
eringram: @danielzhong Thanks for removing. Based on a search I think it's still left over in PrimitiveConverter.ts, PrimitiveConverterFactory.ts, ArcPrimitiveConverter.ts, LineStringPrimitiveConverter.ts, and ShapePrimitiveConverter.ts.

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
danielzhong: Modified!

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
danielzhong: I’m not using it for now. The current prototype moves the camera with Ctrl + WASD and Ctrl + arrow keys. This is just to prepare the iTwin/Cesium camera conversion later

extensions/cesium-renderer/package.json
jason-crow: yes it definitely needs to be a peer, but ofc pnpm will force it to be a single installed instance of cesium on our side, the idea is since its a package with side effects that could be depended on by other packages an app may use it should communicate those side effects by listing it as a peer

@markschlosseratbentley on the point about synchronizing with itwinjs, its going to have to be a peer always unless its expected to import cesium from itwinjs which re-exports it, which I don't think you want to do. The outcome of quietly allowing multiple instances of cesium in the app at the same time is a bigger issue than forcing consumers to mind the cesium version when they upgrade itwinjs imo

markschlosseratbentley: Thanks for the feedback. Per the suggestions, I moved from using `cesium` to `@cesium/engine`. Furthermore, I made `@cesium/engine` be a peerDependency for the `cesium-renderer` package, not a hard dependency. `display-test-app` now has a hard dependency on `@cesium/engine`. This all runs fine from my local testing.

@eringram please resolve this conversation if you are satisfied.

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
markschlosseratbentley: @danielzhong fyi

extensions/cesium-renderer/src/decorations/CoordinateBuilder.ts
markschlosseratbentley: @eringram @danielzhong is this still relevant?

extensions/cesium-renderer/src/decorations/LineStringPrimitiveConverter.ts
markschlosseratbentley: @danielzhong fyi

extensions/cesium-renderer/src/decorations/LineStringPrimitiveConverter.ts
markschlosseratbentley: @eringram if you are satisfied with Daniel's response, could you resolve this conversation?

extensions/cesium-renderer/src/decorations/ArcPrimitiveConverter.ts
markschlosseratbentley: @danielzhong fyi

extensions/cesium-renderer/src/decorations/PolyfacePrimitiveConverter.ts
markschlosseratbentley: @danielzhong fyi

extensions/cesium-renderer/src/decorations/PolyfacePrimitiveConverter.ts
markschlosseratbentley: @danielzhong fyi

extensions/cesium-renderer/src/decorations/SolidPrimitivePrimitiveConverter.ts
markschlosseratbentley: @danielzhong fyi

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
markschlosseratbentley: @eringram if you are satisfied with Daniel's response, could you resolve this conversation?

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
markschlosseratbentley: @danielzhong fyi

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
markschlosseratbentley: @danielzhong fyi

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
markschlosseratbentley: @danielzhong fyi

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
markschlosseratbentley: @danielzhong @eringram does it make sense to remove this code since @eringram is doing other work that will likely replace it?

extensions/cesium-renderer/src/decorations/CoordinateBuilder.ts
eringram: No, looks like they were removed, thanks

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
eringram: I see a test file was added for this, so I'll resolve this and comment directly there if needed

markschlosseratbentley:
@aruniverse We're running into a very similar problem to [what we ran into in the Draco PR](https://github.com/iTwin/itwinjs-core/pull/8449#issuecomment-3255385932), with `core-full-stack-tests`.

In the case of the Draco PR, you made a commit titled "fix lockfile" which altered `pnpm-lock.yaml` and resolved the issue.

I have tried deleting that file and rerunning `rush update` but still get this error.  What exactly did you do to solve this before?

Thanks for your assistance.

[Link to the build failure on CI.](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4362315&view=logs&j=607f17b7-2b33-5225-6b64-00ff0e2e25e4&t=685759b9-6040-5f29-9202-429429ce5747&l=700)

Error:

```
--[ FAILURE: core-full-stack-tests ]------------------------[ 15.45 seconds ]--

ERROR in ../../common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/object-storage-azure/lib/server/AzureServerStorage.js 23:23-53
Module not found: Error: Package path ./dist-esm/storage-blob/src/index.browser.js is not exported from package /Volumes/Builds/vsts_work_a/4/s/common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@azure/storage-blob (see exports field in /Volumes/Builds/vsts_work_a/4/s/common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@azure/storage-blob/package.json)
 @ ../../common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/object-storage-azure/lib/server/index.js 23:13-44
 @ ../../common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/object-storage-azure/lib/index.js 23:13-32
 @ ../../common/temp/node_modules/.pnpm/@itwin+imodels-client-authoring@5.8.1_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/imodels-client-authoring/lib/IModelsClient.js 11:31-69
 @ ../../common/temp/node_modules/.pnpm/@itwin+imodels-client-authoring@5.8.1_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/imodels-client-authoring/lib/index.js 25:13-39
 @ ./lib/frontend/hub/ITwinPlatformEnv.js 5:35-77
 @ ./lib/frontend/TestUtility.js 17:27-60
 @ ./lib/frontend/standalone/WhiteOnWhiteReversal.test.js 11:22-47
```

aruniverse:
> @aruniverse We're running into a very similar problem to [what we ran into in the Draco PR](https://github.com/iTwin/itwinjs-core/pull/8449#issuecomment-3255385932), with `core-full-stack-tests`.
> 
> In the case of the Draco PR, you made a commit titled "fix lockfile" which altered `pnpm-lock.yaml` and resolved the issue.
> 
> I have tried deleting that file and rerunning `rush update` but still get this error. What exactly did you do to solve this before?
> 
> Thanks for your assistance.
> 
> [Link to the build failure on CI.](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4362315&view=logs&j=607f17b7-2b33-5225-6b64-00ff0e2e25e4&t=685759b9-6040-5f29-9202-429429ce5747&l=700)
> 
> Error:
> 
> ```
> --[ FAILURE: core-full-stack-tests ]------------------------[ 15.45 seconds ]--
> 
> ERROR in ../../common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/object-storage-azure/lib/server/AzureServerStorage.js 23:23-53
> Module not found: Error: Package path ./dist-esm/storage-blob/src/index.browser.js is not exported from package /Volumes/Builds/vsts_work_a/4/s/common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@azure/storage-blob (see exports field in /Volumes/Builds/vsts_work_a/4/s/common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@azure/storage-blob/package.json)
>  @ ../../common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/object-storage-azure/lib/server/index.js 23:13-44
>  @ ../../common/temp/node_modules/.pnpm/@itwin+object-storage-azure@2.3.0_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/object-storage-azure/lib/index.js 23:13-32
>  @ ../../common/temp/node_modules/.pnpm/@itwin+imodels-client-authoring@5.8.1_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/imodels-client-authoring/lib/IModelsClient.js 11:31-69
>  @ ../../common/temp/node_modules/.pnpm/@itwin+imodels-client-authoring@5.8.1_inversify@6.0.1_reflect-metadata@0.1.13/node_modules/@itwin/imodels-client-authoring/lib/index.js 25:13-39
>  @ ./lib/frontend/hub/ITwinPlatformEnv.js 5:35-77
>  @ ./lib/frontend/TestUtility.js 17:27-60
>  @ ./lib/frontend/standalone/WhiteOnWhiteReversal.test.js 11:22-47
> ```

I might recommend you wait for this PR to merge: https://github.com/iTwin/itwinjs-core/pull/8553

Also, you seem to have updated the entirety of the lockfile, id revert the change and run `rush update` afterwards to just update wrt the packages you have modified.

markschlosseratbentley:
> Also, you seem to have updated the entirety of the lockfile, id revert the change and run `rush update` afterwards to just update wrt the packages you have modified.

This fixed it, at least locally (CI still running). Thanks @aruniverse.

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
eringram: Yes, I would suggest just removing it to keep things simple and we'll call `createCesiumCameraProps` where necessary

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
eringram: I misread something here - this method returns a `Point3d` not `Cartesian3`, so using the same fallback method wouldn't make sense

extensions/cesium-renderer/src/decorations/ArcPrimitiveConverter.ts
eringram: This is fixed

extensions/cesium-renderer/src/decorations/ArcPrimitiveConverter.ts
```diff
@@ -0,0 +1,254 @@
...
+    this.primitiveType = primitiveType;
+  }
+
+  protected override getCollection(scene: CesiumScene): PrimitiveCollection {
+    this._currentScene = scene; // Store scene reference for later use
+    return scene.primitivesCollection;
+  }
+
+  protected override extractPrimitiveData(coordinateData: DecorationPrimitiveEntry[], primitiveType: string): DecorationPrimitiveEntry[] | undefined {
+    if (!Array.isArray(coordinateData)) {
```
eringram: This check is unnecessary now that `coordinateData` must be type `DecorationPrimitiveEntry[]`. This can also be removed from all other PrimitiveConverter subclasses that override this method

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
eringram: From what I can tell this hasn't been changed. 

This occurs because the `geometries` parameter in `PointPrimitiveConverter.createPointFromGeometry()` is of type `unknown[]`, which is what needs to be fixed.

Going up to where `createPointFromGeometry()` is called in `PointPrimitiveConverter.createPointPrimitiveFromGraphic()` on L94, the argument for `geometries` is a `RenderGraphicWithCoordinates.geometries` which is of type `RenderGeometry[]` (can't be undefined because that's already checked a few lines earlier on L91).

So imo the original parameter should be changed to `geometries: RenderGeometry[]`, and the extraneous undefined checks and casting can be removed (this issue is also in `LineStringPrimitiveConverter` and `ShapePrimitiveConverter`)

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
markschlosseratbentley: @danielzhong can you remove?

markschlosseratbentley:
I fixed the various build errors.  This branch is now passing CI.

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
danielzhong: Removed!

extensions/cesium-renderer/src/decorations/ArcPrimitiveConverter.ts
danielzhong: Fixed!

extensions/cesium-renderer/src/decorations/PointPrimitiveConverter.ts
danielzhong: Fixed

extensions/cesium-renderer/src/decorations/LineStringPrimitiveConverter.ts
danielzhong: Newest changes updated

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
danielzhong: @markschlosseratbentley Do you have suggestions on how to handle this case? Should we refuse conversion or provide a fallback?

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
danielzhong: @markschlosseratbentley 
Should we use fallback when no geolocated?

Cesium fallback: when an iModel isn’t geolocated, do we want to fail fast or keep returning the current “best effort” position?

markschlosseratbentley: I recommend if something is not geolocated, let's place it at [Null Island (0, 0)](https://en.wikipedia.org/wiki/Null_Island).

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
markschlosseratbentley: I recommend if something is not geolocated, let's place it at [Null Island (0, 0)](https://en.wikipedia.org/wiki/Null_Island).

danielzhong: Updated!

extensions/cesium-renderer/src/decorations/CesiumCoordinateConverter.ts
danielzhong: Updated!

common/api/cesium-renderer.api.md
```diff
@@ -26,6 +44,44 @@ export interface CesiumFrustumProps {
     width?: number;
 }
 
+// @internal (undocumented)
+export class CesiumSystem extends RenderSystem {
```
pmconne: Don't export `CesiumSystem` nor `CesiumViewManager` - these are implementation details. Export functions that return a `RenderSystem` and a `ViewManager`, respectively.

extensions/cesium-renderer/src/Graphic.ts
```diff
@@ -0,0 +1,31 @@
...
+    this.geometryType = geometryType;
+  }
+  // ##TODO implement dispose and statistics, if necessary.
+  public dispose() {}
+  public collectStatistics(_stats: RenderMemory.Statistics) {}
+
+  /** Extend `range` to include the bounding box of this graphic, including any child graphics.
+  * @internal
+  */
+  public unionRange(_range: Range3d) {
```
pmconne, line 28: You should be able to move `RenderGraphic.unionRange` to `Graphic.unionRange` - it's only used in the context of core-frontend's `System` class.

common/api/cesium-renderer.api.md
markschlosseratbentley: Good approach. Changes pushed.

extensions/cesium-renderer/src/Graphic.ts
markschlosseratbentley: I started going through and trying to implement this. [GltfDecoration.ts in display-test-app](https://github.com/iTwin/itwinjs-core/blob/02d43712975a6d4b828913871c4e0c41212bdc76/test-apps/display-test-app/src/frontend/GltfDecoration.ts#L209) seems to be a slight complication because it directly calls `.unionRange()` on a `RenderGraphic`.

We could just force that code to cast to a core-frontend `Graphic` and accept it won't work with this CesiumJS prototype.

Suggestions welcome.

pmconne: Oh, the language server didn't find that one when I searched for references to `unionRange`. Cesium's [Geometry](https://cesium.com/learn/cesiumjs/ref-doc/Geometry.html) provides a bounding sphere. Primitive allows you to [show the bounding volume](https://cesium.com/learn/cesiumjs/ref-doc/Primitive.html#debugShowBoundingVolume) for debugging but doesn't appear to expose the volume itself.

Check with @ggetz?

markschlosseratbentley: @pmconne Regarding your original comment here, we're now thinking we need `unionRange` to remain on `RenderGraphic`, right?  The CesiumJS path will need its own implementation. If so, I think we can resolve this original comment since the CesiumJS one is marked TODO in the prototype.

pmconne: Well it's an @internal method used exactly one place outside of our `System` implementation, in a test app. So I think it's an open question whether or not we need it in `RenderGraphic`.

It was introduced in #6535.

markschlosseratbentley: I tend to agree.  For now, we could possibly just make the test app code detect if we are using a core-frontend System object and, if so, cast directly to a frontend `Graphic` in order to call `unionRange`.  Then we could add some CesiumJS-specific code down the road if we deem it necessary.

pmconne: Yeah, no reason to hold up this PR for it.

markschlosseratbentley: Just to be clear, you think we should remove `unionRange` from `RenderGraphic` for this PR, or defer on that?

pmconne: I think your TODO here is fine for now, we can defer a decision on how to address it to a later PR.

pmconne:
> set IMJS_USE_CESIUM=1

Put this in display-test-app's README.

test-apps/display-test-app/src/frontend/Surface.ts
```diff
@@ -163,6 +164,18 @@ export class Surface {
       },
     }));
 
+    tb.addItem(createToolButton({
+      iconUnicode: "\ue9f4",
+      tooltip: "Empty Example",
```
pmconne:
```suggestion
      tooltip: "Cesium Renderer Example",
```

extensions/cesium-renderer/src/test/PrimitiveConverters.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    expect(collection.items).toHaveLength(0);
+  });
+});
+
+
+
+
+
+
+
```
eringram:
```suggestion
```
5 extra lines

extensions/cesium-renderer/vitest.config.mts
```diff
@@ -5,6 +5,16 @@ export default defineConfig({
   },
   test: {
     dir: "src",
+    include: ["**/CesiumSystem.test.ts"],
```
eringram:
```suggestion
    // include: ["**/<insert-file-name-here>.test.ts"],
```

test-apps/display-test-app/README.md
```diff
@@ -152,7 +152,7 @@ You can use these environment variables to alter the default behavior of various
   * If defined, shader source code will be preserved as internal strings, useful for debugging purposes.
 * IMJS_TILETREE_EXPIRATION_SECONDS
   * If defined, the number of seconds after a TileTree has been most recently drawn before purging it.
-* IMJS_TILE_EXPIRATION_SECONDS
+* IMJS_TILE_EXPIRATION_SECONDSF
```
eringram:
```suggestion
* IMJS_TILE_EXPIRATION_SECONDS
```

test-apps/display-test-app/vite.config.mts
```diff
@@ -129,6 +138,12 @@ export default defineConfig(() => {
         prefix: "IMJS_",
       }),
     ],
+    define: {
+      "process.env": process.env, // injects process.env into the frontend
```
eringram: This was recently [removed](https://github.com/iTwin/itwinjs-core/pull/8553/files#diff-26d45bd4d87a5d2a94dd41fd8770a1dc003513bafb9fa68f95b16a248b653b7b) from DTA in https://github.com/iTwin/itwinjs-core/pull/8553
Is it required again or can we just use `import.meta.env`?

markschlosseratbentley: If it was removed, I think it got put back in during a merge-back-to-master by accident. It's probably safe to remove.

eringram:
We are seeing some issues with PrimitiveConverters.test.ts that we plan to address in a future PR, so skipping them for now
Issue: https://github.com/iTwin/itwinjs-core/issues/8627

PR #8553, opened 2025-09-23
update `imodels-*` and `object-storage-*` pkgs to latest

MichaelSwigerAtBentley:
Updates itwin/object-storage-core to latest
Updates itwin/cloud-agnostic-core to latest
Updates itwin/object-storage-azure to latest
Updates itwin/imodels-access-backend to "^6.0.2"
Updates itwin/imodels-access-frontend to "^6.0.2"
Updates itwin/imodels-access-common to "^6.0.2"
Corrects incoming breaking changes

Updates these packages to resolve CVE https://github.com/advisories/GHSA-4hjh-wcwx-xvwj and address tech dept with outdated packages

### DIFF
--- a/core/backend/package.json
+++ b/core/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-backend",
-  "version": "5.3.0-dev.7",
+  "version": "5.3.0-dev.8",
   "description": "iTwin.js backend components",
   "main": "lib/cjs/core-backend.js",
   "module": "lib/esm/core-backend.js",
@@ -113,7 +113,7 @@
     "webpack": "^5.97.1"
   },
   "dependencies": {
-    "@bentley/imodeljs-native": "5.3.6",
+    "@bentley/imodeljs-native": "5.3.8",
     "@itwin/object-storage-azure": "^3.0.4",
     "@azure/storage-blob": "^12.28.0",
     "form-data": "^4.0.4",
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-frontend",
-  "version": "5.3.0-dev.7",
+  "version": "5.3.0-dev.8",
   "description": "iTwin.js frontend components",
   "main": "lib/cjs/core-frontend.js",
   "module": "lib/esm/core-frontend.js",
--- a/full-stack-tests/ecschema-rpc-interface/package.json
+++ b/full-stack-tests/ecschema-rpc-interface/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-rpcinterface-tests",
-  "version": "5.3.0-dev.7",
+  "version": "5.3.0-dev.8",
   "description": "Integration tests for the Schema RPC Interface",
   "author": {
     "name": "Bentley Systems, Inc.",
--- a/full-stack-tests/rpc-interface/package.json
+++ b/full-stack-tests/rpc-interface/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/rpcinterface-full-stack-tests",
-  "version": "5.3.0-dev.7",
+  "version": "5.3.0-dev.8",
   "description": "Test the full iTwin.js Core stack (frontend and backend) using standard RPC interfaces",
   "license": "MIT",
   "scripts": {
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -282,6 +282,7 @@ import { TypeDefinitionElementProps } from '@itwin/core-common';
 import { UpgradeOptions } from '@itwin/core-common';
 import { UrlLinkProps } from '@itwin/core-common';
 import { Vector3d } from '@itwin/core-geometry';
+import { VersionedJSON } from '@itwin/core-common';
 import { ViewAttachmentLabelProps } from '@itwin/core-common';
 import { ViewAttachmentProps } from '@itwin/core-common';
 import { ViewDefinition2dProps } from '@itwin/core-common';
@@ -327,6 +328,7 @@ export class AnnotationTextStyle extends DefinitionElement {
     static fromJSON(props: AnnotationTextStyleProps, iModel: IModelDb): AnnotationTextStyle;
     protected static onInsert(arg: OnElementPropsArg): void;
     protected static onUpdate(arg: OnElementPropsArg): void;
+    static remapTextStyleId(sourceTextStyleId: Id64String, context: IModelElementCloneContext): Id64String;
     static serialize(props: AnnotationTextStyleProps, iModel: IModelDb): ECSqlRow;
     settings: TextStyleSettings;
     toJSON(): AnnotationTextStyleProps;
@@ -2542,6 +2544,7 @@ export class ElementDrivesTextAnnotation extends ElementDrivesElement {
     static onDeletedDependency(props: RelationshipProps, iModel: IModelDb): void;
     // @internal (undocumented)
     static onRootChanged(props: RelationshipProps, iModel: IModelDb): void;
+    static remapFields(clone: ITextAnnotation, context: IModelElementCloneContext): void;
     static updateFieldDependencies(annotationElementId: Id64String, iModel: IModelDb): void;
 }
 
@@ -4947,6 +4950,9 @@ export class OrthographicViewDefinition extends SpatialViewDefinition {
     setRange(range: Range3d): void;
 }
 
+// @internal
+export function parseTextAnnotationData(json: string | undefined): VersionedJSON<TextAnnotationProps> | undefined;
+
 // @beta
 export class PartialECChangeUnifier implements Disposable {
     [Symbol.dispose](): void;
@@ -6439,6 +6445,8 @@ export class TextAnnotation2d extends AnnotationElement2d {
     protected constructor(props: TextAnnotation2dProps, iModel: IModelDb);
     // @internal (undocumented)
     static get className(): string;
+    // (undocumented)
+    protected collectReferenceIds(referenceIds: EntityReferenceSet): void;
     // @beta
     static create(iModelDb: IModelDb, arg: TextAnnotation2dCreateArgs): TextAnnotation2d;
     // @internal
@@ -6451,6 +6459,8 @@ export class TextAnnotation2d extends AnnotationElement2d {
     getAnnotation(): TextAnnotation | undefined;
     // @internal (undocumented)
     getTextBlocks(): Iterable<TextBlockAndId>;
+    // @internal (undocumented)
+    protected static onCloned(context: IModelElementCloneContext, srcProps: TextAnnotation2dProps, dstProps: TextAnnotation2dProps): void;
     // @beta
     protected static onInsert(arg: OnElementPropsArg): void;
     // @internal (undocumented)
@@ -6484,6 +6494,8 @@ export class TextAnnotation3d extends GraphicalElement3d {
     protected constructor(props: TextAnnotation3dProps, iModel: IModelDb);
     // @internal (undocumented)
     static get className(): string;
+    // (undocumented)
+    protected collectReferenceIds(referenceIds: EntityReferenceSet): void;
     // @beta
     static create(iModelDb: IModelDb, arg: TextAnnotation3dCreateArgs): TextAnnotation3d;
     // @internal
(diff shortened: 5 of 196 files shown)


core/backend/src/IModelHost.ts
```diff
@@ -673,10 +673,9 @@ export class IModelHost {
         baseUrl: credentials.baseUrl ?? `https://${credentials.account}.blob.core.windows.net`,
       },
     };
-    const ioc: Container = new Container();
-    ioc.bind<DependenciesConfig>(ExtensionTypes.dependenciesConfig).toConstantValue(config);
-    new AzureServerStorageBindings().register(ioc, config.ServerSideStorage);
-    IModelHost.tileStorage = new TileStorage(ioc.get(ServerStorage));
+    const ioc = InversifyWrapper.create();
```
MichaelSwigerAtBentley: @aerimagne

It seems like 'Container' is removed in 3.0.2. I see that this 'InversifyWrapper' is used in some of the test examples. Is this the replacement?

Should we now be implementing our own 'container' based off of that DIcontainer interface?

core/backend/src/IModelHost.ts
```diff
@@ -673,10 +673,9 @@ export class IModelHost {
...
+    ioc.registerInstance<DependenciesConfig>(ExtensionTypes.dependenciesConfig, config);
```
MichaelSwigerAtBentley: What is the intended workflow for utilizing DIContainer?

I've been running into a bunch of type conflicts trying to replace the removed 'Container'

core/backend/package.json
```diff
@@ -113,12 +113,12 @@
   },
   "dependencies": {
     "@bentley/imodeljs-native": "5.2.16",
-    "@itwin/cloud-agnostic-core": "^2.2.4",
-    "@itwin/object-storage-azure": "^2.3.0",
-    "@itwin/object-storage-core": "^2.3.0",
+    "@itwin/cloud-agnostic-core": "^3.0.2",
```
anmolshres98: semver probably covers it but still, 3.0.3 is latest released now

core/backend/src/IModelHost.ts
```diff
@@ -673,10 +673,9 @@ export class IModelHost {
...
+    IModelHost.tileStorage = new TileStorage(ioc.resolve<ServerStorage>(ExtensionTypes.serverStorage));
```
GytisCepk: Do we need to use `inversify` ?

Since method is called `setupAzureTileCache`, I assume this part is well aware of what provider (Azure) it's using. We could try creating `AzureServerStorage` directly ([Gist](https://gist.github.com/GytisCepk/bcaf0921a082feb502e640a52f17c585))

core/backend/package.json
MichaelSwigerAtBentley: updated

core/backend/src/IModelHost.ts
MichaelSwigerAtBentley: Swapped it over to use AzureServerStorage directly. This adds a dep on azure/blob-storage but seems to work.

core/backend/src/IModelHost.ts
aerimagne: 3.0.0 no longer requires a direct dependency on any specific DI framework. Previously we were using annotations on classes to have inversify working, but now it's possible to implement your own DI container or write a wrapper for a different one (as long as it supports named dependencies and child containers). The DI containers are not required, but we are going to provide limited support for using the storage wrappers directly (e.g., I would treat those constructors as internal and at more liberty to be broken in a minor version, because intended usage is to automatically construct those classes through the registered factories).

The main purpose of these packages is to provide a framework and wrappers to easily switch between different implementations. We do this by pushing the decision of which implementation (in case of storage "azure" or "google") to use to the config files, but in that case more of the cloud-agnostic-core classes would have to be used. Using DI container directly loses that benefit and just simplifies of not requiring to call multiple constructors directly.

full-stack-tests/rpc-interface/src/test/backend.ts
```diff
@@ -38,7 +39,7 @@ const settings = new Settings(process.env);
   RpcConfiguration.developmentMode = true;
 
   // Start the backend
-  const iModelClient = new IModelsClient({ api: { baseUrl: `https://${process.env.IMJS_URL_PREFIX ?? ""}api.bentley.com/imodels` } });
+  const iModelClient = new IModelsClient({ cloudStorage: new AzureClientStorage(new BlockBlobClientWrapperFactory()), api: { baseUrl: `https://${process.env.IMJS_URL_PREFIX ?? ""}api.bentley.com/imodels` } });
```
aerimagne, line 42: You can just write: 
```
const hubAccess = new BackendIModelsAccess({api: { baseUrl: `https://${process.env.IMJS_URL_PREFIX ?? ""}api.bentley.com/imodels` }});
await IModelHost.startup({ hubAccess });
```

aruniverse: we are explicitly passing in the cloud storage so we dont enter the codepths that pull in the google dependencies that cuase us build issues. there is too much coupling today with the storage providers

full-stack-tests/core/webpack.config.js
```diff
@@ -50,7 +50,6 @@ function createConfig(shouldInstrument) {
         url: false,
       },
       alias: {
-        "@azure/storage-blob$": "@azure/storage-blob/dist-esm/storage-blob/src/index.browser.js",
```
MichaelSwigerAtBentley, line 53: @aruniverse 
When I updated azure/storage-blob this line started to cause the build to fail ("path not found"). I'm assuming this was a hold over from the cjs/esm node changes. Removed it and its building fine, is this needed still?

aruniverse: if it builds fine then no

full-stack-tests/rpc-interface/src/test/backend.ts
aerimagne: You are using this only for full-stack-tests and you're removing the ability to run these tests in the environment that iModels API runs.

test-apps/display-test-app/vite.config.mts
```diff
@@ -129,9 +129,9 @@ export default defineConfig(() => {
         prefix: "IMJS_",
       }),
     ],
-    define: {
-      "process.env": process.env, // injects process.env into the frontend
-    },
+    // define: {
+    //   "process.env": process.env, // injects process.env into the frontend
```
MichaelSwigerAtBentley: Should we leave this around incase we want to enable it again? or remove it as unneeded

core/backend/package.json
```diff
@@ -113,16 +113,15 @@
   },
   "dependencies": {
     "@bentley/imodeljs-native": "5.3.6",
-    "@itwin/cloud-agnostic-core": "^2.2.4",
-    "@itwin/object-storage-azure": "^2.3.0",
-    "@itwin/object-storage-core": "^2.3.0",
+    "@itwin/cloud-agnostic-core": "^3.0.4",
+    "@itwin/object-storage-azure": "^3.0.4",
+    "@itwin/object-storage-core": "^3.0.4",
```
aruniverse: review if we need a direct dep on object-storage-core
look at using `import type` to help you determine if its a real dep or not

core/backend/package.json
```diff
@@ -113,16 +113,15 @@
   },
   "dependencies": {
     "@bentley/imodeljs-native": "5.3.6",
-    "@itwin/cloud-agnostic-core": "^2.2.4",
-    "@itwin/object-storage-azure": "^2.3.0",
-    "@itwin/object-storage-core": "^2.3.0",
+    "@itwin/cloud-agnostic-core": "^3.0.4",
```
aruniverse: review if cloud-agnostic-core is needed

core/frontend/package.json
```diff
@@ -94,8 +94,8 @@
     "NOTE: core-frontend should remain UI technology agnostic, so no react/angular dependencies are allowed"
   ],
   "dependencies": {
-    "@itwin/cloud-agnostic-core": "^2.2.4",
-    "@itwin/object-storage-core": "^2.3.0",
+    "@itwin/cloud-agnostic-core": "^3.0.4",
+    "@itwin/object-storage-core": "^3.0.4",
```
aruniverse: does this need to be a direct dep?

full-stack-tests/rpc-interface/package.json
```diff
@@ -34,10 +34,11 @@
     "url": "http://www.bentley.com"
   },
   "dependencies": {
-    "@itwin/imodels-access-backend": "^5.2.1",
-    "@itwin/imodels-access-frontend": "^5.2.1",
-    "@itwin/imodels-client-authoring": "^5.8.1",
-    "@itwin/imodels-client-management": "^5.8.1",
+    "@itwin/imodels-access-backend": "^6.0.2",
+    "@itwin/imodels-access-frontend": "^6.0.2",
+    "@itwin/imodels-client-management": "^6.0.2",
```
aruniverse, line 39: review if we actually need both imodels-client- management and authoring; you might be able to get by with just authoring

full-stack-tests/ecschema-rpc-interface/package.json
```diff
@@ -37,10 +37,11 @@
...
     "@itwin/ecschema-rpcinterface-common": "workspace:*",
     "@itwin/ecschema-rpcinterface-impl": "workspace:*",
-    "@itwin/imodels-access-backend": "^5.2.1",
-    "@itwin/imodels-access-frontend": "^5.2.1",
-    "@itwin/imodels-client-authoring": "^5.8.1",
-    "@itwin/imodels-client-management": "^5.8.1",
+    "@itwin/imodels-access-backend": "^6.0.2",
+    "@itwin/imodels-access-frontend": "^6.0.2",
+    "@itwin/imodels-client-management": "^6.0.2",
+    "@itwin/imodels-client-authoring": "^6.0.2",
```
aruniverse, line 43: review deps

test-apps/display-test-app/vite.config.mts
aruniverse: i think we should prob nuke it but does need review
cc @hl662

aruniverse (commented):
please review the dependencies!

test-apps/display-test-app/vite.config.mts
hl662: need to verify that a runtime keyin/method that might use env variables doesn't break

full-stack-tests/rpc-interface/package.json
MichaelSwigerAtBentley: We do need both, the iModelAccessFrontend expects the type from management. Removing it and swapping to the iModelClient from authoring causes the build error I ran into earlier.

full-stack-tests/ecschema-rpc-interface/package.json
MichaelSwigerAtBentley: Same as rpc-interface

core/backend/package.json
MichaelSwigerAtBentley: Can use import type everywhere, swapped to dev

core/backend/package.json
MichaelSwigerAtBentley: done, can be removed now

core/frontend/package.json
MichaelSwigerAtBentley: swapped to dev

test-apps/display-test-app/vite.config.mts
hl662: i can pick up import.meta.env being used correctly when testing locally. @MichaelSwigerAtBentley check display performance test app's vite config and also remove this?

@eringram google photorealistic 3d tiles is the other place in the frontend using this too. Do the image tests test those tiles? Regardless of the answer, can you run image tests against this branch to confirm things are fine

eringram: Image tests don't test the Google 3D Tiles yet, but I started a run for this branch. I also just manually tested them in DTA on this branch and it picks up the API key from .env

MichaelSwigerAtBentley: Removed the `process.env` from the perf test app's config as well, and swapped over all `process.env` for `import.meta` in the perf test app frontend.

Deleted the commented out section as well.

hl662 (approved):

test-apps/display-test-app/vite.config.mts
eringram: Image tests results look good too

PR #8569, opened 2025-09-25
Updating TextStyleSettings

Alfonso-Martello:
This PR makes several modifications to the properties in `TextStyleSettings`. It contains the following changes:

1. `lineHeight` is replaced with `textHeight`.  Only TextBlocks applied the `lineHeight` property. `textHeight` is applied by any `TextBlockComponent`. This allows for multi-sized text:

image

2. Slightly changes the names of some frame style properties to be more descriptive. `border` -> `borderColor`, `fill` -> `fillColor`.
3. Replaces `fontName` with `font`, which is now a `FontFamilySelector`. Fonts are uniquely identified by the combination of their name + type. By default, we assume most fonts are TrueType. `FontFamilySelector` allows the font type to be optionally provided.
4. Moves `margins` from `TextBlock` into `TextStyleSettings`. Our assumption is that users will usually want to set up the margins once as part of their text style instead of defining it on a per instance basis. Style overrides still allow per instance behavior.
5. Moves `justification` from `TextBlock` into `TextStyleSettings`. We have the same assumption as for margins (that users want to specify their `justification` once per style). Eventually, we would like to be able to apply justification on more than just the TextBlock (such as on Paragraphs). Style overrides seem a convenient way to be able to do that.


core/common/src/annotation/TextStyle.ts
```diff
@@ -55,6 +56,25 @@ export type TextAnnotationFrameShape = typeof textAnnotationFrameShapes[number];
  */
 export type TextAnnotationFillColor = TextStyleColor | "background" | "none";
 
+/** Describes the margins around the content inside a [[TextBlock]]. It's measured in meters.
```
pmconne:
```suggestion
/** Describes the margins around the content inside a [[TextBlock]], measured in meters.
 * All margins default to zero if `undefined`.
```

core/common/src/annotation/TextStyle.ts
```diff
@@ -406,6 +454,12 @@ export class TextStyleSettings {
     if (props.frame) {
       copy.frame = { ...props.frame };
     }
+    if (props.font) {
```
aruniverse: why not use `structuredClone`?

aruniverse (approved):

aruniverse:

core/common/src/annotation/TextStyle.ts
Alfonso-Martello: Good suggestion, made the change

claudiaareneee (approved):
Looks good to me

core/common/src/annotation/TextStyle.ts
```diff
@@ -242,20 +269,20 @@ function deepFreeze<T>(obj: T) {
...
+  public readonly font: Readonly<Required<FontFamilySelector>>;
+  /** The height of the text, in meters. Many other settings use the text height as the basis for computing their own values.
+   * For example, the height and offset from baseline of a subscript [[TextRun]]  are computed as textHeight * [[subScriptScale]] and
+   * textHeight * [[subScriptOffsetFactor]], respectively.
    */
-  public readonly lineHeight: number;
+  public readonly textHeight: number;
   /** Multiplier used to compute the vertical distance between two lines of text.
-   * The distance is computed in meters as lineSpacingFactor * [[lineHeight]].
+   * The distance is computed in meters as lineSpacingFactor * tallest [[textHeight]] in the line.
```
claudiaareneee: actually one comment, i don't know if this is true. I think it's just multiplied by the block font height: https://github.com/iTwin/itwinjs-core/blob/6be5023d7d988603d82ebcb4929edbf4e7f4c548/core/backend/src/annotations/TextBlockLayout.ts#L893

Alfonso-Martello: I think you are right, updated the comment

Issue #8509, opened 2025-09-10
Misleading error thrown by getElementProps


Josh-Schifter:
**Describe the bug**
[iModel.Elements.getElementProps](https://www.itwinjs.org/reference/core-backend/imodels/imodeldb/imodeldb.elements/getelementprops/) throws 'Element not found' regardless of the cause of the error.

**To Reproduce**

    const regionId = "0x000001" // a valid id for a region element in the iModel
    const badId = {id: regionId} as unknown as Id64String;
    const element = iModel.elements.getElementProps<SectionDrawingLocationProps>({ id: badId, wantGeometry: true });

other variations throw the same error:

    //const spatialRegion = iModel.elements.getElementProps<SectionDrawingLocationProps>({ id: badId , wantGeometry: true });
    //const spatialRegion = iModel.elements.getElementProps<SectionDrawingLocationProps>(badId );
    //const spatialRegion = iModel.elements.tryGetElementProps(badId );
    //const spatialRegion = iModel.elements.tryGetElement(badId );

**Expected behavior**
Throw a more meaningful error.

pmconne:
Where did you get that Id from? "0x000001" is not a valid element Id - leading zeroes are not permitted. "0x1" is a valid element Id, but it's always the root subject, not a region element.

Josh-Schifter:
The id is not the point of the issue.... its just a string.  The point is that the argument is not a valid type for the function.

pmconne:
> The id is not the point of the issue.... its just a string. The point is that the argument is not a valid type for the function.

The issue claims we throw "not found" regardless of the cause of the error.
That it not true. We throw "not found" if you give us an identifier like "0x000001" that doesn't identify any element.
If you just made up "0x000001" for this issue, then please provide an actual case that produces the problem instead.

pmconne:
I think you are taking issue with the fact that `getElementProps` calls `tryGetElementProps` which returns `undefined` if any exception occurs. `getElementProps` throws "not found" if `tryGetElementProps` returned `undefined`. Any other error that occurs inside `tryGetElementProps` is ignored.

If so, I agree, that's kinda dumb. It should work the other way: the guts of `tryGetElementProps` belong inside `getElementProps` with no try-catch; `tryGetElementProps` should call `getElementProps` inside a `try-catch` and return undefined if any exception is thrown.

[...]

PR #8570, opened 2025-09-25
First step towards JSON versioning

Alfonso-Martello:
This PR is the first step towards JSON versioning (specifically for text/drawing production).
### The Problem
Drawing Production needs a way to version our JSON data. We have two goals with this:
1. Make sure we never break iModels that have Drawing Production data written to them.
2. Be able to update and migrate our data model based on user feedback and changing requirements.  
Our vision is to version the stringified JSON data such that we can check the version when deserializing and migrate the data if it is out of date or throw an error if the data is newer (we are assuming an evergreen application like iTwin Studio for now).
### The long-term solution
Long-term, the idea is to extend the EC technology to allow it to describe JSON data in terms of EC constructs.
### This PR
This PR is instead adding a version directly to the instances of TextAnnotation2d/3d. We suspect that even once the long-term strategy is implemented, it will still be helpful for data migration to have instances with their JSON data version.
This PR introduces a new interface:
```
interface VersionedJSON<T> {
  version: string;
  data: T;
}
```
For Drawing Production, we are planning on stringifying that object and storing it in the iModel, so the version is always with the JSON string. I kept it locally to the text file for now as this strategy is still new, but Drawing Production is planning on using the same interface.
The `version` string is a SemVer string in the format `MAJOR.MINOR.PATCH`.
Generally, we want to check the version and possibly migrate data on read, and make sure we include the version on write. There are several workflows for reading/writing in iTwin, I have tried to cover all workflows to ensure a version is present and valid.
### Next steps
Next steps after this PR is to define a [JSON Schema](https://json-schema.org/) for the text JSON. This schema will serve 3 purposes:
1. A thought exercise to make sure we are persisting the data we expect to persist (not accidentally allowing edge cases).
2. Documentation on the JSON data model and how it has evolved over time.
3. Validation tests to detect when breaking schema changes happen.
JSON schemas are usually identified by their `$id`, which is traditionally in the format of a URI (even if that URI does not resolve to anything). We are imagining a situation where the schema ID URI can be built on the fly using a formula similar to: `https://dev.bentley.com/json_schemas/<domain>/<json-property-name>/<version>` (so text would be at something like `https://dev.bentley.com/json_schemas/Core/TextAnnotationData/1.0.0` and Drawing Production schemas would be at something like `https://dev.bentley.com/json_schemas/DrawingProduction/TableData/1.0.0`). This is opposed to storing the full schema URI as the `version` string.


core/common/src/ElementProps.ts
```diff
@@ -644,8 +646,19 @@ export interface SheetReferenceProps extends SheetIndexEntryProps {
 export interface AnnotationTextStyleProps extends DefinitionElementProps {
   /** An optional human-readable description of the text style.*/
   description?: string;
-  /** The stringified JSON representation of the text style.
-   * @see [[TextStyleSettingsProps]]
+  /** The stringified versioned JSON representation of the text style.
+   * @see [[VersionedJSON]] for the JSON representation.
+   * @see [[TextStyleSettingsProps]] for the data model.
    */
   settings?: string;
```
aruniverse, line 653: I don't personally love this "hack" that the settings is a string, and you guys can just jsonify whatever you want and store it.
The versioning helps, but couldnt this be more strongly typed?

core/common/src/ElementProps.ts
```diff
@@ -644,8 +646,19 @@ export interface SheetReferenceProps extends SheetIndexEntryProps {
...
+}
+
+/** Wrapper for versioned JSON data.
+ * @beta
+ */
+export interface VersionedJSON<T> {
+  /** The semver version of the JSON data. */
+  version: string;
```
aruniverse: would prefer a more stronger type here, maybe use a [template literal](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)? 

```
type semverString = `${string}.${string}.${string}`
```

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -6,25 +6,69 @@
...
-
-function parseTextAnnotationData(json: string | undefined): TextAnnotationProps | undefined {
-  if (!json) return undefined;
+import { CustomHandledProperty, DeserializeEntityArgs, ECSqlRow } from "../Entity";
+import * as semver from "semver";
+
+/** The version of the JSON stored in `TextAnnotation2d/3dProps.textAnnotationData` used by the code.
+ * @internal
+*/
+export const TEXT_ANNOTATION_JSON_VERSION = "1.0.0";
```
aruniverse, line 23: are we really confident enough with these schemas to start off with 1 as the major?

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -363,6 +486,33 @@ function updateTextBlocks(elem: TextAnnotation2d | TextAnnotation3d, textBlocks:
   elem.update();
 }
 
+/** The version of the JSON stored in `AnnotationTextStyleProps.settings` used by the code.
+ * @internal
+*/
+export const TEXT_STYLE_SETTINGS_JSON_VERSION = "1.0.0";
```
aruniverse, line 488: same comment re major

core/common/src/ElementProps.ts
aruniverse: but i guess you need this to align with the semver pkg, and any types it has

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -6,25 +6,69 @@
...
+
+/** The version of the JSON stored in `TextAnnotation2d/3dProps.textAnnotationData` used by the code.
+ * @internal
+*/
+export const TEXT_ANNOTATION_JSON_VERSION = "1.0.0";
+
+function validateAndMigrateVersionedJSON<T>(
+  json: string,
+  currentVersion: string,
+  migrate: (old: VersionedJSON<T>) => T
```
aruniverse, line 28: who will eventually provide these migrate fns?
would they live entirely in core or in the app?

core/common/src/ElementProps.ts
pmconne: This is the raw JSON representation (wire format) representing what is read out of or written into the database as specified by the BIS schema. The BIS property is of type `string`, extended type `JSON`.

pmconne:
> The version string is a SemVer string in the format MAJOR.MINOR.PATCH.

I thought you wanted something equivalent to [ECVersion](https://www.itwinjs.org/reference/ecschema-metadata/metadata/ecversion/) which is `read.write.minor`.

core/common/src/ElementProps.ts
ben-polinsky: Does the semver package have a [utility type like this](https://www.typescriptlang.org/play/?#code/PTAEFpK7QCwUwDYAd4CcDOFo4LACgAXAT1VAGEB7AO0IEMBLajAHgGVR4APQ+agEywZCaJgHMANBU48+g0MNHUxAPlABeAqFAduvAVgAGAEgDei8QF8z5a+ZFXDoAPygRAV3igAXKABmdIgY8ADcBAQgoABEgYig-AxiDIQYUaAA7smwoNSUConUwPzwAMYMALaBBCRkAJIYACKJyay6sgYKDspqmvjaVLSMzOxSUQB0UWp6clgeXq4BQfBaPhQ09Eyto+CTMvrycy7+gcErvm37RmbU7uUARuiWTq6HvovBYfjVpF5NSSkAOUoABl4HQEsoAFroSitPYzTpKVQaFb1P4tdhTdoHNCeFbaVwACguCMMAAYzEw-OhQAAlJ5HQm0+EdKJpVwcXzUeAAN3QAEpVmx+fjVty+WhPhEwAAxShoUDINDwcDKxBg4KgBjFWgMPwMdAYXx6xXuNXEHK3dAMEpSBIYWKUdKgdXg8SgABeMIwn0ilEICDQmU1dBKJXgyEIoEJ6S8-BoAHIo8qxNxwCUECUANagR2x-igDN0TBuPJZ+AR0AAFQ4Du5-LG3zIAAVlbV+AB5ADS7BZ8gs3RRfVAaOaKUxfdmuPm0fRgJBYIhYmhaFhE+mHXFNNcW4V50F5ylRB+oAa-t4-Fb8G7vY3-a6yN62hJHRMpipNIAEpYxpTqNSFSrJ5RSJK92xvT8sUuHJeW3GCJVWM9CAvK8byrFQRWHbRfDAzsezYFQj0iHAcELeUvGCcoEIADTGABNMZIWwEjICbLw2HgcoADV0CoZVb2xIQHx6FYX3kN8PwVABBH8-wA0AACFZPff8aVsQwQNPMcMCBUE3Shb0WCkqCEV3I5d1FbCtP+HSF305dDIUkzN1ghUd1cyzEO03TF3EFc13IZz5DM9yJU885RS5VyiLAFioH8dxEDiSiEMyANFWVVUkA1Lw6AEUA7ncBhEALOLWL6Gp2M4ni0AE6CB0fFZIjS7JCuKgtyngeh+DoehRMnUAJNUhVPwAUSkhpLAAajMBrgKwxlPO0SIxom0BKgtWA6D5DKVTVHLQHlHJ-SW0BVoaAahvk8gO1pUbLHAOSaWbO75u0d6PsZDjuN48iWBuu6gqwEL4PQU7PsQ894EvZUbxe0agdBtykfBz791R3xiWq37+POxGQbM9GFu0TDlrAFrdqy11NUKqNcgKoqSrOHRLqehUAfux6VPk+G3veolvpqvj4H+26EYGgmPOJ97fCQlDYZ7eH8dc8ypYhqzhSasBkEQIYyOVZnBZxkWCJi5iWJkOhyh1+AsDK8ACBKGhhEOrMAEZzmx2qojdsYACYxgAZl2dRol9gPg8+J3mCjSgsz9z2fu9skxhTslwECZAtpD6I09TjOUGzqPndjrNA8TmqWCiAO3ZTwPVRKMY3amtqSrGAB2duc+rpu64bpuW8Z-gO67o9o5du5wQ9nQvarslw6D7v5-9xex5LgrwQTmek6rhf6-n7u9-AA+15jjf+HL7fK59-3D9vkIgA)?

Alfonso-Martello:
> > The version string is a SemVer string in the format MAJOR.MINOR.PATCH.
> 
> I thought you wanted something equivalent to [ECVersion](https://www.itwinjs.org/reference/ecschema-metadata/metadata/ecversion/) which is `read.write.minor`.

Yes, we want something exactly like that. Really, I was just thinking of the format `<number>.<number>.<number>`, but the semantics will be like ECVersion (`read.write.minor`)

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: For text, I expect the migration functions will live entirely in core since the text data structure code also lives in core

In terms of who would write them, presumably whoever makes a change to the text data model (most likely someone from Drawing Production or someone from core making text changes)

pmconne:
> the semantics will be like ECVersion (read.write.minor)

Great - make sure to document that (e.g., by linking to the `ECVersion` docs from your API doc).

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: We are as confident as we can be without getting actual feedback from users. We are treating having versioning as a step towards getting that feedback. We expect that there will likely be changes in the future

I am happy to start with a lower version, but it is not completely clear to me what the "confidence" threshold will be for moving to a higher version

Additionally, [the semantics](https://www.itwinjs.org/bis/guide/schema-evolution/schema-versioning-and-generations/#schema-version-numbers) are actually closer to that of ECVersion, which might change which numbers we want to start with

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -6,25 +6,69 @@
...
+  let parsed;
   try {
-    return JSON.parse(json);
+    parsed = JSON.parse(json) as VersionedJSON<T>;
   } catch {
     return undefined;
   }
+
+  const version = parsed.version;
+  if (typeof version !== "string" || !semver.valid(version))
```
Alfonso-Martello, line 38: What should we do with text elements currently being written by the connectors? They are not writing versioned data.

Originally, I was thinking un-versioned json is simply a mistake, but if we already have un-versioned data that exists, should we instead treat un-versioned data as needing to migrate to the first version (`1.0.0` or whatever we decide for this PR)?

core/backend/src/annotations/TextAnnotationElement.ts
ColinKerr: If we plan on supporting elements created with this version in the future then 1.0.0 is appropriate.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -6,25 +6,69 @@
...
+  if (typeof parsed.data !== "object" || parsed.data === null)
+    throw new Error("JSON data is missing or invalid.");
+
+  // Newer
+  if (semver.gt(version, currentVersion))
+    throw new Error(`JSON version ${parsed.version} is newer than supported version ${currentVersion}. Application update required to understand data.`);
+
+  // Older
+  if (semver.lt(version, currentVersion)) {
+    parsed.data = migrate(parsed);
```
Alfonso-Martello, line 50: This method does data migration on parse, which means it does data migration on most reads. Is this potentially problematic for read-only workflows?

Alfonso-Martello (commented):
Doing a self-review to document two questions I have about this implementation

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -6,25 +6,69 @@
...
+  return parsed;
+}
+
+function migrateTextAnnotationData(oldData: VersionedJSON<TextAnnotationProps>): TextAnnotationProps {
+  if (oldData.version === TEXT_ANNOTATION_JSON_VERSION) return oldData.data;
+
+  // Place migration logic here.
+
+  throw new Error(`Migration for textAnnotationData from version ${oldData.version} to ${TEXT_ANNOTATION_JSON_VERSION} failed.`);
 }
```
ColinKerr, line 63: What is the behavior in your app when these errors are thrown?  Will it be a graceful failure or a crash?

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: Is any connector producing elements containing `TextAnnotation`s in their JSON? I thought they all produced static pictures of text, not actual text blocks.

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: When reading data out of the database, you want to upgrade it (in memory!) to match the current version so that all other APIs can operate upon it using that version; otherwise every API would have to check the version and react. So this seems correct to me.

ColinKerr:
> Validation tests to detect when breaking schema changes happen.
> JSON schemas are usually identified by their $id, which is traditionally in the format of a URI (even if that URI does not resolve to anything). We are imagining a situation where the schema ID URI can be built on the fly using a formula similar to: `https://dev.bentley.com/json_schemas/<domain>/<json-property-name>/<version>` (so text would be at something like `https://dev.bentley.com/json_schemas/Core/TextAnnotationData/1.0.0` and Drawing Production schemas would be at something like `https://dev.bentley.com/json_schemas/DrawingProduction/TableData/1.0.0`). This is opposed to storing the full schema URI as the version string.

This URI format doesn't match the standard formatting used by the json spec: https://json-schema.org/understanding-json-schema/reference/schema#schema. They put the dialect before the schema name.  I suggest trying to be consistent with JSON schema terminology and formatting.  I don't think that means you cannot use a three digit version.

I probably would have also followed the JSON spec and put the $schema property in the JSON but I do not have a strong enough opinion on this to say it cannot be this way.

What I'd really like to review is the actual json schema

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: There are a lot of code paths that can reach here, so the behavior depends on the code path.

Generally, we have logic to catch errors and handle them gracefully. For these specific errors, we have no handling because they are new. For newer data, I would say we need to alert the users since it requires their action (upgrading the app). For older data, a failing migration is a worst case scenario and I have not yet thought through the UX of how that should look.

Some code paths obfuscate these errors. For example, calling `iModel.elements.getElement<TextAnnotation2d>(elementId)` eventually causes `tryGetElementProps` to be called. This in turn, calls `TextAnnotation2d.deserialize`, which could cause these errors to throw. However, `tryGetElementProps` will catch and just [return undefined](https://github.com/iTwin/itwinjs-core/blob/5f4c886ebe56620dcbad99e9a19bf2e1e08a77af/core/backend/src/IModelDb.ts#L2202). So the error that is raised is a "Not Found" error, which is misleading. We will have to know to take a look at the JSON version when we get those errors

ColinKerr: We need some documented standards here and guidance for how to handle failures to load elements because the handler detects an error.  That said I don't want to hold up this PR until we have that, I suggest we create an issue to track and then resolve this comment.  

@aruniverse do you agree?

aruniverse: Cross linking, https://github.com/iTwin/itwinjs-core/issues/8509,  for reference and will resolve this to not hold up this pr. Can tackle this in the same linked issue

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: Unsure, there was a back and forth where I think the connectors even briefly stopped writing TextAnnotation2d/3d elements all together. If they are just producing static pictures, that would be perfect

core/common/src/ElementProps.ts
Alfonso-Martello: I went ahead and used the template literal type. It should be compatible with the semver package, since most the methods just take a `string`.

There is a [semver types package](https://www.npmjs.com/package/@types/semver?activeTab=readme), but I did not see any types to use as the version string.

I could take the code @ben-polinsky shared and add that utility type logic. Then we would end up with something like
```
export interface VersionedJSON<T, K extends string> {
  /** The semver version of the JSON data.
   * Uses the same semantics as [ECVersion]($ecschema-metadata).
   */
  version: SemVer<K>;
  /** The JSON data. */
  data: T;
}
```
But that might be overkill.

The version string is the same as the one used in [ECVersion](https://www.itwinjs.org/reference/ecschema-metadata/metadata/ecversion/), but I did not see any special typing with that class that could be reused here.

core/common/src/ElementProps.ts
```diff
@@ -653,14 +653,19 @@ export interface AnnotationTextStyleProps extends DefinitionElementProps {
   settings?: string;
 }
 
+/** A string in the format `read.write.minor` where the semantics match those of [ECVersion]($ecschema-metadata).
+ * @beta
+ */
+export type SemverString = `${string}.${string}.${string}`
+
```
pmconne: I don't mean to nitpick but "semver" means "major.minor.patch" to everyone else in the world. Should this be called "ECVersionString" to make it clear what it represents?

Alfonso-Martello:
> This URI format doesn't match the standard formatting used by the json spec: https://json-schema.org/understanding-json-schema/reference/schema#schema. They put the dialect before the schema name. I suggest trying to be consistent with JSON schema terminology and formatting. I don't think that means you cannot use a three digit version.

You are right. I think that just slightly changes the formula for crafting the URI to `https://dev.bentley.com/json_schemas/<domain>/<version>/<json-property-name>` (or some permutation of this)

> I probably would have also followed the JSON spec and put the $schema property in the JSON but I do not have a strong enough opinion on this to say it cannot be this way.

Shouldn’t the `$schema` property be part of the actual JSON schema and not the JSON data? Since the `$schema` identifies the overall JSON Schema version (like, is the schema written in Draft 7 JSON Schema? Draft 6 JSON schema, etc)?

> What I'd really like to review is the actual json schema

I would like to write the JSON schema as part of the next PR. Having a version moves us towards release in a major way, having actual JSON schemas helps future-proof us. Trying to sneak the version into release 5.2

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: Double-check with @abeesh and/or @jffmarker but that's my understanding. Most of them build off a C++ codebase that doesn't know anything about text blocks stored in JSON. We've always ignored the old, deprecated BisCore.TextAnnotationData aspect, if they're still producing those.

jffmarker: We needed to keep producing TextAnnotation elements because clash detection needs to have rules based on annotations, which was impossible when we switched to generic elements with text graphic streams. Connectors will create text annotation elements with flatbuffer data, and they have no idea about the JSON that you have in itwin.js. When the JSON format was finalized, we wanted to revisit whether it was worth writing a parallel implementation in CPP for the Connector... but generally speaking your tools shouldn't be touching anything the Connectors make in their own channel, so (a) there is no JSON today, and (b) why would your tools be trying to do anything to the Connector text?

CC: @abeesh

pmconne: > your tools shouldn't be touching anything the Connectors make in their own channel

Yes exactly.

Alfonso-Martello:
@pmconne, @ColinKerr, @aruniverse 

Does anything else need to happen for this to be approved/merged?

ColinKerr:
> Shouldn’t the `$schema` property be part of the actual JSON schema and not the JSON data? Since the `$schema` identifies the overall JSON Schema version (like, is the schema written in Draft 7 JSON Schema? Draft 6 JSON schema, etc)?

The json schema spec doesn't define any way to reference the schema within a json object.  This is a pretty reasonable write up why they do not: https://github.com/orgs/json-schema-org/discussions/473. We did choose to do it for ECSchema JSON and I think it's reasonable for us, but since it's not the standard I can agree that we are under no obligation to put it into our json objects

ColinKerr (approved):

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -87,31 +132,33 @@ export class TextAnnotation2d extends AnnotationElement2d /* implements ITextAnn
    * @beta
    */
   public defaultTextStyle?: TextAnnotationUsesTextStyleByDefault;
-  /** Optional string containing the data associated with the text annotation. */
-  private _textAnnotationData?: string;
+  /** The versioned data associated with the text annotation. */
+  private _textAnnotationData?: VersionedJSON<TextAnnotationProps>;
```
pmconne: Seems easier just to store the JSON here without the version. You migrate it to current version on read (or throw an error), so `version` should always match current. You only need to add the current version on write, but instead you're doing it in several places (fromJSON, create). Or did I overlook some scenario in which `_textAnnotationData.version !== TEXT_ANNOTATION_JSON_VERSION`? If not, this is just a nit.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -154,35 +200,73 @@ export class TextAnnotation2d extends AnnotationElement2d /* implements ITextAnn
...
+    if (!props.elementGeometryBuilderParams && textAnnotationData) {
+      props.elementGeometryBuilderParams = getElementGeometryBuilderParams(arg.iModel, props.model, props.category, props.placement ?? Placement2d.fromJSON(), textAnnotationData.data, props.defaultTextStyle?.id);
     }
+  }
 
-    props.elementGeometryBuilderParams = getElementGeometryBuilderParams(iModelDb, props.model, props.placement ?? Placement2d.fromJSON(), props.textAnnotationData, props.category, props.defaultTextStyle ? props.defaultTextStyle.id : undefined);
+  /**
+   * TextAnnotation2d custom HandledProps include 'textAnnotationData'.
+   * @inheritdoc
+   * @beta
```
pmconne:
```suggestion
   * @internal
```

Nobody needs to know you override this. Same for TextAnnotation3d and AnnotationTextStyle.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -275,8 +361,8 @@ export class TextAnnotation3d extends GraphicalElement3d /* implements ITextAnno
    */
   public static create(iModelDb: IModelDb, arg: TextAnnotation3dCreateArgs): TextAnnotation3d {
     const elementProps: TextAnnotation3dProps = {
-      classFullName: this.classFullName,
-      textAnnotationData: JSON.stringify(arg.textAnnotationData),
+      classFullName: this.className,
```
pmconne: Why did you change this?

pmconne (commented):
To be clear, every iModel containing a TextAnnotation will now become unusable, right? (It will throw when we try to read the element).

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: No, you're right - the version should always be up to date after parse

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: Done for `_customHandledProps`

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: Not intentionally

I have been enjoying the AI auto-complete/suggestion feature recently - I suspect that was an AI change I overlooked

Alfonso-Martello:
> To be clear, every iModel containing a TextAnnotation will now become unusable, right? (It will throw when we try to read the element).

Yes, we are planning to scrap our testing iModels once this merges, and there are no real production iModels with this yet

core/backend/src/annotations/TextAnnotationElement.ts
Alfonso-Martello: When you say

> You only need to add the current version on write, but instead you're doing it in several places (fromJSON, create)

Are you suggesting that the version should only be added in one place? The most "on write" place I can think of would be `onInsert` and `onUpdate`. I am currently parsing there a) for the geometry generation and b) to prevent bad JSON from accidently being inserted (for example by going through a props only workflow that bypasses the class)

aruniverse:

PR #8572, opened 2025-09-25
Modified CurveFactory.createFilletsInLineString

saeeedtorabi:
Now `CurveFactory.createFilletsInLineString` allows users to work on polygons by adding fillet at the start/end point for closed line strings


core/geometry/src/curve/CurveFactory.ts
```diff
@@ -191,32 +213,38 @@ export class CurveFactory {
           blendArray[i].arc = undefined;
         }
       }
-      /* The "1-b" logic above prevents this loop from ever doing anything.
-      // on edge with conflict, suppress the arc with larger fraction
-      for (let i = 1; i < n; i++) {
-        const b0 = blendArray[i - 1];
-        const b1 = blendArray[i];
-        if (b0.fraction12 > 1 - b1.fraction10) {
-          const b = b0.fraction12 > b1.fraction12 ? b1 : b0;
```
saeeedtorabi, line 200: I did not understand this block and it was commented out so I deleted it.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -138,48 +138,70 @@ export class CurveFactory {
...
    * @param radius fillet radius or array of radii indexed to correspond to the points.
    * @param allowBackupAlongEdge true to allow edges to be created going "backwards" along edges if needed to create the blend.
+   * @param filletAtTheSeam if true and the line string is closed, create fillet at the seam (first/last point). If false,
+   * no fillet is constructed at the first or last point and therefore, those entries in `radius[]` are never referenced.
    */
   public static createFilletsInLineString(
     points: LineString3d | IndexedXYZCollection | Point3d[],
     radius: number | number[],
     allowBackupAlongEdge: boolean = true,
+    filletAtTheSeam: boolean = false,
```
dassaf4: Please add a new options bundle to simplify the method signature (see e.g., `RegionOps.regionBooleanXY` in [PR](https://github.com/iTwin/itwinjs-core/pull/8585)):
* Create a new interface `CreateFilletsInLineStringOptions` that collects both boolean arguments as optional properties.
* Replace the booleans with `allowCuspOrOptions: boolean | CreateFilletsInLineStringOptions = true`.
* At method top, parse the expanded type and derive a value for both options. If the options bundle is undefined, or an option in the bundle is undefined, use its default value.
* Simplify option name: `CreateFilletsInLineStringOptions.allowCusp`. "Backup" has overloaded meaning. Cusp is more informative. We are unfortunately stuck with the default being true.
* Simplify option name: `CreateFilletsInLineStringOptions.filletClosure`. This reflects the new behavior. While I like the brevity of "seam", this term is not widely used in core-geometry or elsewhere.

core/geometry/src/test/curve/CurveFactory.test.ts
```diff
@@ -183,9 +181,53 @@ describe("CurveFactory", () => {
     const radii = [0, 2, 1, 0.8, 0.6, 0.4];
     const chain2 = CurveFactory.createFilletsInLineString(lineString0, radii, true)!;
     GeometryCoreTestIO.captureCloneGeometry(allGeometry, chain2, x0, 0.0);
+
     GeometryCoreTestIO.saveGeometry(allGeometry, "CurveFactory", "FilletsInLineString");
     expect(ck.getNumErrors()).toBe(0);
   });
+  it("FilletsInPolygon", () => {
```
dassaf4, line 189: Please add:
* result verification to the fillet tests. The existing "FilletsInLineString" test is good for coverage only. We can do better.
* test radius[0] > 0 when `filletClosure = true`.
* test square line string of side length `s` and `radius = s/2` fillets. Confirm that the resulting chain consists entirely of arcs and is circular when `filletClosure = true`.
* test _open_ line string when `filletClosure = true`. The resulting chain should be closed and the seam filleted.
* test edge cases: 2-point and 3-point line strings with `filletClosure = true/false`

dassaf4 (changes requested):
I simplified the code and fixed two issues I found in my review. I also have a couple of requests.

The changes I fixed (need verification):
1. I've changed the behavior for the new option to fillet at the first input point even if the input is not closed. Many times, a polygon is passed around without the duplicate start = end point; callers shouldn't have to add this redundant closure point in order to fillet it.
2. When `filletClosure = true` the output chain was starting at the second input vertex/fillet. This is unexpected. Now it starts at the first vertex/fillet.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -141,109 +149,76 @@ export class CurveFactory {
...
    * @param radius fillet radius or array of radii indexed to correspond to the points.
-   * @param allowBackupAlongEdge true to allow edges to be created going "backwards" along edges if needed to create the blend.
-   * @param filletAtTheSeam if true and the line string is closed, create fillet at the seam (first/last point). If false,
-   * no fillet is constructed at the first or last point and therefore, those entries in `radius[]` are never referenced.
+   * @param allowCuspOrOptions this option has 2 elements: `allowCusp` and `filletClosure`. If `allowCusp` is `true`
+   * (default) it allows creation of retrograde edges to join large-radius fillets (this results in cusps in the output
+   * path). If `false`, such a fillet is disallowed, resulting in a simple corner. Also if `filletClosure` is true, treat
+   * `points` as a polygon (closure point optional but the returned path will be closed) and create a fillet at the start
+   * point. If `false` (default), the first and last points receive no fillet and their respective entries in the radius
+   * array are ignored.
```
dassaf4: It's all here, but please move the documentation for the options bundle to the interface definition. The doc line here should be for `allowCusp` and its default value only; the user can refer to the interface definition for the full documentation of the options bundle. This minimizes doc duplication and potential for divergence. See the new doc for `RegionOps.regionBooleanXY` on main branch for an example.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -106,6 +107,13 @@ interface SmoothCurveData {
   endTangent?: Vector3d;
 };
 
+interface CreateFilletsInLineStringOptions {
```
dassaf4: This interface should be exported, explained by a brief doc block, and `@public`.

core/geometry/src/test/curve/CurveFactory.test.ts
```diff
@@ -188,42 +188,119 @@ describe("CurveFactory", () => {
...
+
+    // square
+    x0 += 20;
+    const s = 5;
+    const square = LineString3d.create([0, 0], [s, 0], [s, s], [0, s]);
+    let radius = s / 2;
+    let chain = CurveFactory.createFilletsInLineString(square, radius, { allowCusp: true, filletClosure: true })!;
+    GeometryCoreTestIO.captureCloneGeometry(allGeometry, chain, x0);
+    for (const c of chain.children)
+      ck.testTrue(c instanceof Arc3d, "expect only arcs in the square with s = radius/2");
```
dassaf4: Another way to verify this particular test case is to call `RegionOps.consolidateAdjacentPrimitives` on the filleted chain and verify you get back a single full-sweep Arc3d of expected radius and center.

core/geometry/src/test/curve/CurveFactory.test.ts
```diff
@@ -188,42 +188,119 @@ describe("CurveFactory", () => {
...
+    chain = CurveFactory.createFilletsInLineString(specialLineString, radius, { allowCusp: false, filletClosure: true })!;
+    GeometryCoreTestIO.captureCloneGeometry(allGeometry, chain, x0);
+    for (const c of chain.children)
+      ck.testTrue(c instanceof LineSegment3d, "expect only line segments in the linestring chain with 2 points");
+    chain = CurveFactory.createFilletsInLineString(specialLineString, radius, { allowCusp: false, filletClosure: false })!;
+    GeometryCoreTestIO.captureCloneGeometry(allGeometry, chain, x0, 8);
+    for (const c of chain.children)
+      ck.testTrue(c instanceof LineSegment3d, "expect only line segments in the linestring chain with 2 points");
+
+    // special case with 3 points
```
dassaf4, line 303: Do we expect any fillet arcs in this 3-point test? Yes:
* `filletClosure: false`: the output should have exactly 1 arc, at index 1
* `filletClosure: true` the output should have exactly 3 arcs, at indices 0, 2, 4, and the first arc should be the closest arc to the first input point.

These are all good verifications to add.

core/geometry/src/test/curve/CurveFactory.test.ts
dassaf4: Good!

core/geometry/src/curve/CurveFactory.ts
dassaf4: Good!

core/geometry/src/curve/CurveFactory.ts
saeeedtorabi: addressed

core/geometry/src/curve/CurveFactory.ts
saeeedtorabi: addressed

core/geometry/src/test/curve/CurveFactory.test.ts
saeeedtorabi: addressed

core/geometry/src/test/curve/CurveFactory.test.ts
saeeedtorabi: addressed

core/geometry/src/curve/CurveFactory.ts
dassaf4: Further minimized doc duplication.

dassaf4:
Added a slight improvement to the test...

Issue #8192, opened 2025-06-09
Dispose of event listener in QuantityTypesFormatProvider


hl662:
Add Symbol.dispose logic to the QuantityTypesFormatProvider, to remove the activeUnitSystem event listener created [on init.](https://github.com/iTwin/itwinjs-core/blob/b2c1278fbb5f2dec31a75c33daedc60ae66a0ff6/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L329)

PR #8576, opened 2025-09-25
Handle formatting cases where sign of value is flipped after conversion

hl662:
Closes https://github.com/iTwin/itwinjs-core/issues/8192

This pull request introduces significant improvements to the quantity formatting logic, especially around unit conversions that may result in negative values (such as temperature conversions), and enhances test coverage for these scenarios. It also updates development dependencies to support these changes.

### Quantity formatting improvements

* Refactored the `Formatter` class to handle sign preservation and formatting correctly during unit conversions, particularly for cases where conversions (e.g., Kelvin to Fahrenheit) can produce negative results and vice versa, where a negative value would turn into a positive result. The logic for formatting single-unit and composite-unit values now consistently applies sign rules and ensures correct output for edge cases. 

### Testing enhancements

* Expanded the test suite in `Formatting.test.ts` to include comprehensive tests for temperature conversions between Kelvin, Celsius, Fahrenheit, and Rankine, covering both negative and positive results, as well as fractional formatting and edge cases like absolute zero.

### Resource management

* Updated `QuantityTypeFormatsProvider` to track and dispose of event listeners using the new `[Symbol.dispose]` method, improving resource management and preventing potential memory leaks.


core/quantity/src/test/Formatting.test.ts
```diff
@@ -1223,3 +1235,190 @@ describe("Formatting tests:", () => {
     });
   });
 });
+
+describe("Formatting tests with more extensive units:", () => {
```
rschili: This is my only nitpick, we should put a better test name. Maybe "Flips sign of value during formatting". Extensive isn't a good description of what this tests.

hl662: Got it, I renamed it to "Formatting tests handling temperature conversions where sign is flipped". I moved the test suite over to ecschema-metadata to avoid circular deps

hl662:
@Mergifyio backport release/5.2.x

ColinKerr:
It is a bad idea to split formatting into two paths that each do unit conversion differently.  I believe this could be written more cleanly with a single path by:
1. Converting the input value (no abs value)
2. Getting the sign after the conversion
3. For composite units you should be able to make the math work with negative numbers so long as the component units are also using the same sign as the main magnitude.  e.g. if it's -0.2794 meter to be displayed in FT/IN you convert to FT (-1.0833) then subtract whole feet (-1.0833 - -1.0 = -0.0833), then get inches (-1) you will get proper 1 FT 1 IN and then you can apply sign using common code

Test that sign formatting options are not changed by your changes.

ColinKerr (approved):

aruniverse (approved):

PR #8583, opened 2025-09-30
Automatically Backport Security Fixes

andremig-bentley:
WIP

Still in testing on fork of core, working on bumping branch versions automatically and creating PR tempates

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -27,3 +27,18 @@ pull_request_rules:
         message: |
           This pull request is now in conflicts. Could you fix it @{{author}}? 🙏
           To fixup this pull request, you can check out it locally. See documentation: https://help.github.com/articles/checking-out-pull-requests-locally/
+
+  - name: Automatic backport for Security fixes
+    conditions:
+      - and:
+          - merged
+          - base=master
+          - label="Security"
+    actions:
+      backport:
+        branches:
+          # Automatically backport to all active release branches
+          # Update these branch names as needed for your project
+          - "release/5.2.x"
+          - "release/5.0.x"
+          - "release/4.9.x"
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -37,8 +37,6 @@ pull_request_rules:
     actions:
       backport:
         branches:
-          # Automatically backport to all active release branches
-          # Update these branch names as needed for your project
           - "release/5.2.x"
           - "release/5.0.x"
           - "release/4.9.x"


.github/PULL_REQUEST_TEMPLATE/security_fix.md
```diff
@@ -0,0 +1,34 @@
+<!--
+Please rename the PR to follow the following convention:
+PR Title: "Security Fix: <Short Description of Changes>"
```
andremig-bentley, line 3: A plain template by itself can't populate a PR's title or add labels, but this can be done with github actions where the workflow runs conditionally off of branch names. Is it worth it to add a branch naming convention and then a github action for security fixes?

.github/mergify.yml
```diff
@@ -1,4 +1,9 @@
 # Configuration file for mergify
+
+# Branch variables for backport targets
+latestMinor: &latestMinor "release/5.2.x"
```
aruniverse: 1. plz link to docs explaining the syntax
2. how can we automate update this latest minor branch name every time we do a release?
3. latest minor is not 4.9, its 4.11

.github/PULL_REQUEST_TEMPLATE/security_fix.md
aruniverse: i think i prob would prefer the label over branch naming convention

.github/PULL_REQUEST_TEMPLATE/security_fix.md
```diff
@@ -0,0 +1,34 @@
...
+
+Please add the "securty" label to the PR.
+-->
+
+## Security Fix
+
```
aruniverse, line 9: would like to see an option to link to the actual cve / ghas notice

@wgoehrig is there anything else you would like to see in this template?

.github/mergify.yml
aruniverse: disregard 2; i see its handled in ver bump pipeline

andremig-bentley: re 1 - which syntax do you mean, the branch naming convention?

aruniverse: no the variable, `&latestMinor`

andremig-bentley: Its using YAML's anchors and aliases:
https://yaml.org/spec/1.2.2/#anchors-and-aliases
https://yaml.org/spec/1.2.2/#alias-nodes
https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/

It seemed like a better solution to me than just having the release branches directly named in step, but it is a little more complicated. It can be changed to just:

```
backport:
    branches:
        - "release/5.2.x"
        - "release/4.11.x"
 ```

File: .github/PULL_REQUEST_TEMPLATE/security_fix.md (hunk 3 above)
Comment (andremig-bentley) [reply .github/PULL_REQUEST_TEMPLATE/security_fix.md:9]: added a notice link section in the latest commit

File: .github/mergify.yml (hunk 2 above)
Comment (aruniverse) [reply .github/mergify.yml]: why wouldnt hardcoding these instead of adding it to another var that is only referenced once and needs to be updated not be easier?

Comment (andremig-bentley) [reply .github/mergify.yml]: it would be easier to just have them hardcoded, my main thinking was just clarity, as in what are these branches, why are we backporting to them specifically, etc. But, that can be taken care of with doc comments. In the latest commit, I've simplified it to hardcode the branch names, a simpler implementation with a doc comment is probably the better way to go.

File: .github/mergify.yml [hunk 4]
```diff
@@ -27,3 +32,15 @@ pull_request_rules:
         message: |
           This pull request is now in conflicts. Could you fix it @{{author}}? 🙏
           To fixup this pull request, you can check out it locally. See documentation: https://help.github.com/articles/checking-out-pull-requests-locally/
+
+  - name: Automatic backport for Security fixes
+    conditions:
+      - and:
+          - merged
+          - base=master
+          - label="security"
```
hl662, line 37: do we want **all** PRs labeled security to fall into this bucket of backport? I know PRs resolving CVEs are the main thing, but is there anything more than that, we would naturally categorize as security? From static code analysis fixes, SCAs, etc... Then again, those should be backported...?

@wgoehrig

aruniverse: > From static code analysis fixes, SCAs, etc...

why wouldnt those be labeled security and not backported?
we can also manually run the backport cmd for those dont fall into the bucket and need a manual review

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -254,8 +258,22 @@ jobs:
...
+
+          # When creating a minor release, the mergify.yml file needs to be updated to change the hardcoded branch names for security fixes
+          - bash: |
+              mergifyPath=".github/mergify.yml"
+              version=$(echo $(getVersion.version) | sed 's/\([0-9]*\.[0-9]*\)\.[0-9]*/\1.x/')
+              releaseBranch="release/$version"
+              echo "Updating mergify.yml to use $releaseBranch as the recent minor branch for security fixes"
+
+              # Update the latest minor backporting release branch in mergify.yml
+              sed -i "0,/\"release\/[0-9]*\.[0-9]*\.x\"/ s|\"release/[0-9]*\.[0-9]*\.x\"|\"$releaseBranch\"|" "$mergifyPath"
```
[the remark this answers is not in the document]
andremig-bentley: After some testing/research, using yq may result in other changes to the file's format other than intended as the file is processed. Sticking with sed to avoid any problems and match how similar steps are done elsewhere in the pipelines.

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -254,8 +258,22 @@ jobs:
...
         name: getVersion
 
-      # When creating a minor release, the NextVersion.md need to be cleared and the contents placed into a {Version Number}.md file
       - ${{ if eq(parameters.BumpType, 'minor') }}:
+
+          # When creating a minor release, the mergify.yml file needs to be updated to change the hardcoded branch names for security fixes
+          - bash: |
+              mergifyPath=".github/mergify.yml"
+              version=$(echo $(getVersion.version) | sed 's/\([0-9]*\.[0-9]*\)\.[0-9]*/\1.x/')
+              releaseBranch="release/$version"
```
DanRod1999, line 271: if we were to update 4.11 to 12, would this still work? It looks fine, but just double checking updating prev major won't cause unintended problems

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -113,6 +113,10 @@ jobs:
           # manually to be the most recent previous major release
           # e.g. if this release is `release/5.0.x`, value in `gather-docs.yaml`
           # should be `release/4.<whatever_last_minor_release_version_was>.x`
+          # additionally if major version bump, the `mergify.yml` also needs to be edited manually
+          # if this release is `release/5.0.x`,change the `latestMinor` variable to `release/5.0.x`
```
aruniverse: comments are out of date

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: I think there's a chance it would break with this current implementation. Currently it basically just takes the version created from the version bump step and sets the first release branch in the list to that value. If we were to run a version bump on release/4.11.x for a new minor, I think it would set 5.whatever.x to 4.12.x and leave 4.11.x there. There wouldn't be any problem with patches. Is there any precedent for us releasing a new minor on a previous major, though?

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: updated in most recent commit

common/config/azure-pipelines/jobs/version-bump.yaml
andremig-bentley: As a fail safe, the most recent commit changed the logic so that the version-bump step looks at the major version number of the new version, and updates the release branch which starts with that same major version number. This means that if we were to release a new minor on a previous major version (4.11 -> 4.12), the bump would update the correct branch.

aruniverse: explain to me the regex here
if you think theres a chance this could break then this is wrong
it is very possible for us to have a new minor on the last major ver

andremig-bentley: I thought there was a chance of breaking before the most recent commit removed that, sorry if that wasn't clear. This line here is getting the name of the release branch through the version number. So it's finding the version number through `getVersion.version`, then creating the name of the release branch by prepending "release" and then replacing the patch number with "x".

Originally, the workflow from this point was basically just looking for the first time the word "release" appeared in the branch lists and updating that branch. That's what had a chance of breaking, but that logic was changed to be effectively:

1. Figure out what release branch we're on
2. Find the release branch in the list that has a matching major version
3. Update that branch

That way, in the case of a new minor on a previous major, the logic wouldn't update the first in the list and remove the branch representing the most recent minor version, but it would instead go find the previous major branch and update that one instead. That logic is all handled on lines 272-276.

aruniverse: i ended up asking copilot 
image

this seems ok to me. resolving...

aruniverse (approved):

PR #8584, opened 2025-09-30
Return error result when unit labels aren't matched during quantity parsing

hl662:
This pull request improves error handling in the `@itwin/core-quantity` package's quantity parser. The parser now returns an explicit error result when a unit label is supplied but cannot be matched to any known unit, rather than silently failing or accepting invalid labels. This is a breaking change that ensures more accurate feedback for consumers of the parsing API.

**Error handling improvements:**

* The `Parser` class now throws a `QuantityError` with `QuantityStatus.UnitLabelSuppliedButNotMatched` when unique unit labels are present but not matched to any known unit.
* The main parsing logic catches this error and returns a result with `ParseError.UnitLabelSuppliedButNotMatched`, ensuring callers receive a clear error when parsing fails due to unknown unit labels.

**Testing and documentation:**

* Added a new test case to `Parsing.test.ts` to verify that parsing with invalid unit labels returns the expected error.
* Updated the change history documentation to highlight this, warning that parsing operations with invalid unit labels will now return errors instead of succeeding silently.

### DIFF
--- a/common/changes/@itwin/core-backend/2025-10-14-12-48.json
+++ b/common/changes/@itwin/core-backend/2025-10-14-12-48.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -22,8 +22,8 @@ importers:
         specifier: ^12.28.0
         version: 12.28.0
       '@bentley/imodeljs-native':
-        specifier: 5.3.14
-        version: 5.3.14
+        specifier: 5.3.15
+        version: 5.3.15
       '@itwin/object-storage-azure':
         specifier: ^3.0.4
         version: 3.0.4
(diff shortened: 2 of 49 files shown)


hl662:
I will highlight that none of our previous tests failed, there are no regressions **so far**, including this relevant test case where when a unitless format is passed in, the parser will use the first unit label that matches any known units for parsing

    const testData = [
      { value: "12,345.345 - 1", magnitude: 12345.345 - 1}, // unitless
      { value: "1m + 1FT + 1", magnitude: 1 + 0.3048 + 1}, // default to meters
      { value: "1FT + 1m + 1", magnitude: 0.3048 + 1 + 0.3048}, // default to feet
      { value: "1 + 1 FT + 1m", magnitude: 0.3048 + 0.3048 + 1 }, // default to feet
    ];

rschili (approved):

docs/changehistory/NextVersion.md
```diff
@@ -3,3 +3,12 @@ publish: false
...
 
+- [NextVersion](#nextversion)
+  - [@itwin/core-quantity](#itwincore-quantity)
+    - [Changes](#changes)
+
+## @itwin/core-quantity
+
+### Changes
+
+- **Breaking Change**: Enhanced [Parser]($quantity) error handling to properly propagate `ParseError.UnitLabelSuppliedButNotMatched` when unit labels are provided but cannot be matched to known units. **This change may cause parsing operations that previously succeeded with invalid unit labels to now return error results instead**, providing more accurate feedback and preventing silent failures when parsing quantities with unrecognized unit labels.
```
hl662: @grigasp do you foresee any problems with property-grid if this change is introduced? I can imagine there could be many typos of unit labels when users are editing a property value, which would lead to this situation

grigasp: Just checked both - our legacy and the new quantity editor, and they both use `ParserSpec.parseToQuantityValue` to parse the value, and they both check if the result is `ok`. I think that means we're good?

Adding @saskliutas just in case, too.

aruniverse: i want to be cautious of using the term "breaking"

plz have issue author chime in on your api changes and make sure it meets their needs
and check with the ecdb team

hl662: i asked the author of the issue, they are good with it. @ColinKerr are you okay with this

ColinKerr: This should be improved in two ways.

1. This change should be described in terms of the error it is fixing rather than as a breaking change, so as a bug fix rather than a breaking change.  It should be described in concrete terms like: `12 im` instead of `12 in` it would have been interpreted as input successfully parsed as 12 meters.  This change treats `12 im` as an error to ensure parsing errors are noticed and handled by the caller.  This is technically a breaking functional change but so is almost every bug fix that isn't fixing a hard crash.  
2. The behavior of the parser needs to be described in some actual documentation besides this note in the next version.

> NOTE: If this was persisted data (e.g. we saved `12 im` in some way) we would have to consider how to handle already persisted data ... since that is not the case here we do not need to consider how this might impact loading persisted data.

hl662: I updated the change documented in NextVer to not mention it's a breaking change. Explicitly mentioned how this error would be raised, and updated the documentation for the Quantity learnings page to explain at a high level how the parser works.

PR #8585, opened 2025-09-30
Improvements to `RegionOps.regionBooleanXY` and `constructAllXYRegionLoops`

dassaf4:
`RegionOps.regionBooleanXY` improvements:
* Some arc-arc and line-arc chains don't join exactly. The solver's tight parametric tolerance misses some of these intersections, resulting in an incorrectly constructed graph. Now we follow up near-roots at endpoints with a distance test using the caller's potentially more relaxed distance tolerance. These problems were more common for `Arc3d`s that do not start/end along an arc axis.
* New option to post-process Boolean union to return exterior `Loop`(s) instead of `UnionRegion`.
* Result region now inherits z-coordinate from the first input region (or the second if `BMinusA`).

`RegionOps.constructAllXYRegionLoops` improvements:
* `PlanarSubdivision.addHalfEdge`: handle large tolerances
* `RegionOps.computeXYAreaTolerance`, `ClusterableArray.clusterIndicesLexical`: handle zero tolerance
* `Arc3d.areaToChordXY`: sign for clockwise sweep was wrong, resulting in incorrect output
* `GraphComponent.buildFaceData`: in some cases, face areas were not computed, resulting in incorrect output. Now fall back on polygon area to ensure every face has an associated signed area.


core/geometry/src/curve/Query/PlanarSubdivision.ts
```diff
@@ -23,6 +24,26 @@ import { RegionGroupMember, RegionGroupOpType } from "../RegionOpsClassification
  * @module Curve
  */
 
+function sortAngle(curve: CurvePrimitive, fraction: number, reverse: boolean): number {
```
saeeedtorabi: I understand that you just moved this function but the name is confusing. Initially I thought it sorts some angles but then realized that `sortAngle` is the angle used for sort-around-vertex. I think a name like `getSortAngle` or `calculateSortAngle` makes more sense (next 2 functions also start with `get`).

saeeedtorabi (approved):

core/geometry/src/curve/Query/PlanarSubdivision.ts
dassaf4: Good point. Fixed.

aruniverse:

PR #8586, opened 2025-10-01
Commit auto-created deprecation dates in Ruleset.schema.json[release/5.2.x]

tcobbs-bentley:
Also revert the 5.2.0 commit.


DanRod1999:
should we revert the 5.2.0 commit in this PR while we're at it?

tcobbs-bentley:
> should we revert the 5.2.0 commit in this PR while we're at it?

I think that should probably be done in a separate PR, given how many files will be restored.

tcobbs-bentley:
This cannot be done anywhere but as part of the release pipeline. The process that adds the deprecation date needs to be fixed to deal with the presentation schema files.

PR #8592, opened 2025-10-01
Update Version Bump to Build Presentation Common after the Deprecation Date Commit

andremig-bentley:


tcobbs-bentley (approved):

presentation/common/scripts/post-process-json-schema.cjs
```diff
@@ -46,7 +46,8 @@ const processedSchema = handleObject(schema);
 fs.writeFileSync(schemaPath, JSON.stringify(processedSchema, undefined, 2));
 
 const isCI = process.env.TF_BUILD;
-if (isCI) {
+const isVersionBump = process.env.VERSION_BUMP === "True";
```
aruniverse: this NEEDS to be documented

andremig-bentley: doc comments added in latest commit

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -176,8 +176,15 @@ jobs:
...
+          # We need to rebuild presentation-common after running the lint-deprecation rule
+          # to ensure that ruleset.schema.json is updated correctly.
+          - bash: node common/scripts/install-run-rush.js build --to @itwin/presentation-common
+            displayName: rush build to presentation-common
+            env:
+              VERSION_BUMP: "true"
+
           - bash: |
-              if git diff --quiet; then
+              if git diff --exit-code; then
```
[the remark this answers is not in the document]
andremig-bentley: We want the diff printed, resolving it.

MichaelSwigerAtBentley (approved):

aruniverse:
Doesnt this need to be cherry picked back to master?

saskliutas:
I was just wondering maybe it would be possible to handle this in a generic way instead of explicitly handling only `presentation-common` by changing order of actions: `lint-deprecations` -> `rush build`?

andremig-bentley:
> I was just wondering maybe it would be possible to handle this in a generic way instead of explicitly handling only `presentation-common` by changing order of actions: `lint-deprecations` -> `rush build`?

This is already the order of actions we're taking, we run `lint-deprecation` and then we run `rush build`. We don't have a build step before `lint-deprecation`, in fact before this change we actually had no build step within the version bump script at all. To clarify, are you asking why we are only building to `presentation-common` instead of using an entire `rush build`?

andremig-bentley:
> Doesnt this need to be cherry picked back to master?

Cherry pick pr [here](https://github.com/iTwin/itwinjs-core/pull/8598).

saskliutas:
> > I was just wondering maybe it would be possible to handle this in a generic way instead of explicitly handling only `presentation-common` by changing order of actions: `lint-deprecations` -> `rush build`?
> 
> This is already the order of actions we're taking, we run `lint-deprecation` and then we run `rush build`. We don't have a build step before `lint-deprecation`, in fact before this change we actually had no build step within the version bump script at all. To clarify, are you asking why we are only building to `presentation-common` instead of using an entire `rush build`?

Ok, look like I misunderstood what is happening in that pipeline. If I understand now correctly we need to rebuild `presentation-common` after lint-deprecation rule to make sure we commit changes that were produced during `presentation-common`. In that case I still think we should not make this explicit to `presentation-common` but just do full build to make sure that if some other package will start to produce changes during build due to deprecation comment update we would not need to come back to this.

Issue #6903, opened 2024-06-27
Prototype: Loading schemas incrementally via ECSql


ColinKerr:
@grigasp has run into performance issues when loading large schemas on the frontend for presentation: https://github.com/iTwin/presentation/issues/601

Recent improvements have helped but for large schemas it's unlikely that an async incremental way to load schemas would improve startup time.

ecschema-metadata has the delayed promise abstraction around schema items but the implementation loads the entire schema at once.  We should investigate an implementation that incrementally loads the schema using ECSql.

1. schema key + reference schema keys - This is enough to return the SchemaInfo for a schema
2. SchemaItem names - This is enough to create the schema item keys and hopefully construct the lazy loaded schema item
3. chunks of schema item bodies - We don't want to make an HTTP call for each schema item, that would be too many calls even for a small schema.  We can investigate a single query that loads all dependencies for a given item.  Loading all base classes for a class is easy to do, but what about all other dependencies?

Goals of this investigation
1. Is the current Lazy Load/DelayedPromise paradigm the correct one to use?
2. Can we efficiently incrementally load schemas?
3. Can we load enough to answer some important questions quickly then use other mechanisms to load the rest of the schema?
4. Suggestion how to proceed

ColinKerr:
One possible implementation idea:

1. Load all schema item infos (a concept similar to SchemaInfo) which have item key plus key of referenced items
2. When a schema item is requested use SchemaItemInfo for that item to walk the tree of what referenced items are needed so a single request can be made to load all those items.  Partially loaded schema knows what is already loaded so tree list of referenced items will only include those not yet loaded.

> Consider adding other items to request if request is small to further reduce requests for very flat schemas
> Consider loading units, koqs, formats, enums, structs or some of those on first load so list of referenced items doesn't have to go down to property level.
> Consider filling requests using ECSql ECDbMeta queries instead of in memory C++ schemas

rschili:
An alternative possible implementation that I have in mind:

1. Flatten the schemas. Like a "compiled" version of the schema. So each class has on its own all of its properties and infos attached.
2. Retain the names of there an info came from (like property.Origin.SchemaItemKey, property.KindOfQuantity.SchemaItemKey)

This should simplify loading single items, but it would bload in scenarios where a full schema is needed.
Advanced version:

1. Support both, the flattened items, and fully loaded schemas
2. After a certain threshold we switch from the first to the latter and in-place replace the cached schema contents
[...]

PR #8594, opened 2025-10-02
Add incremental schema loading to imodel schemacontexts

StefanApfel-Bentley:
Adds the incremental schema locaters to IModelDb and IModelConnection schemacontext as default locater.
Also added a setting incrementalSchemaLoading to the `IModelHost` and `IModelApp` options.

*Frontend*
```ts
await IModelApp.startup({
  incrementalSchemaLoading: "enabled"
});

await IModelApp.startup({
  incrementalSchemaLoading: "disabled"
});
```
*Backend*
```ts
await IModelHost.startup({
  incrementalSchemaLoading: "enabled",
});

await IModelHost.startup({
  incrementalSchemaLoading: "disabled",
});
```

For now the incremental schema loading is disabled by default.

### DIFF
--- a/docs/changehistory/NextVersion.md
+++ b/docs/changehistory/NextVersion.md
@@ -74,7 +74,7 @@ await IModelApp.startup({
   incrementalSchemaLoading: "disabled"
 });
 ```
-*Backend
+*Backend*
 ```
 await IModelHost.startup({
   incrementalSchemaLoading: "enabled",


pmconne:
Can you please explain in the PR or issue description why we are making this change?

> Also introduces an internal api to enable/disable incremental schema loading through a helper method.

Why is it internal? Who is expected to call it? AFAICT, right now it is only used by tests.

StefanApfel-Bentley:
> Can you please explain in the PR or issue description why we are making this change?

We added incremental schema loading as additional way to load schemas incrementally instead of all at once as JSON documents. After some testing now, we want to add the incremental schema locaters as default locaters on the schemacontexts of all IModels and IModelConnections replacing the default schemaProps loader. 

More information: [Loading schemas incrementally via ECSql](https://github.com/iTwin/itwinjs-core/issues/6903)

> 
> > Also introduces an internal api to enable/disable incremental schema loading through a helper method.
> 
> Why is it internal? Who is expected to call it? AFAICT, right now it is only used by tests.

It's internal because we don't want to get it added to the public API, only internal components (within itwinjs) shall enable it if they want to use it now. One candidate for that is presentation (who also requested the change in the first place). I can imagine after some additional test runs of several components, we might switch the default from disabled to enabled.

ColinKerr:
There as got to be some better way to do this.  Outside the scope of this one case we are going to need the ability to have some configuration to flag on/off code.  Should each case be handled individually or should we have some pattern(s) to follow?

@aruniverse

core/ecschema-metadata/src/IncrementalLoading/IncrementalSchemaLocater.ts
```diff
@@ -21,6 +21,20 @@ interface IncrementalSchemaInfo extends SchemaInfo {
 
 type LoadSchemaInfoHandler = (context: SchemaContext) => Promise<Iterable<SchemaInfo>>;
 
+// Global options for incremental schema loading.
+const schemaLoadingOptions = {
+  isEnabled: false,
```
aruniverse: from the name of the top level obj and this property, i would not have guessed this was wrt to incremental schema loading.

i dont think i love this approach of a psuedo global var that is checked in both frontend and backend. is there a reason why both frontend and backend need to respect this, and both of them (frontend and backend) could decide indepently which approach (incremental vs not) to use?

grigasp:
> It's internal because we don't want to get it added to the public API, only internal components (within itwinjs) shall enable it if they want to use it now. One candidate for that is presentation (who also requested the change in the first place)

Our components that would benefit from incremental schema loading are outside of this repo, so any API they need to call should be public.

However, why not just have it always enabled? 

I don't really like the idea of a single component deciding to change global state that affects the whole application. IMO, if for some reason we can't have that enabled by default, then it should be a configuration option at the application level.

StefanApfel-Bentley:
I'm currently trying to switch the incremental schema activation from the global function to the proposed application level setting. This is definitely a better approach as it would not require to rebuild the code if someone needs to change the setting.

I saw there is an class `NativeAppStorage` in `core-backend` which allows to specify settings per application. 
In `core-frontend`, we have a similar concept with a `Storage` on the `NativeApp` class.
Just to make sure we're going the right direction here, this are the classes we're suppose to use for this cases or are there any other apis I haven't seen yet?

Also, we follow Grigas comment and always activate the incremental schema loading by default, if a client runs into issues on their side, it can then use the app config to opt-out to get the previous behavior.

pmconne:
> I saw there is an class `NativeAppStorage` in `core-backend` which allows to specify settings per application. In `core-frontend`, we have a similar concept with a `Storage` on the `NativeApp` class. Just to make sure we're going the right direction here, this are the classes we're suppose to use for this cases or are there any other apis I haven't seen yet?
> 
> Also, we follow Grigas comment and always activate the incremental schema loading by default, if a client runs into issues on their side, it can then use the app config to opt-out to get the previous behavior.

Typically, apps configure the behavior of the core packages through [IModelAppOptions](https://www.itwinjs.org/reference/core-frontend/imodelapp/imodelappoptions/) (frontend) and/or [IModelHostOptions](https://www.itwinjs.org/reference/core-backend/imodelhost/imodelhostoptions/) (backend). The app can decide how to choose the configuration (e.g., using a feature flagging service, hard-coding features they always/never want enabled, whatever). The `NativeAppStorage` APIs you mention aren't really relevant here.

StefanApfel-Bentley:
> Typically, apps configure the behavior of the core packages through [IModelAppOptions](https://www.itwinjs.org/reference/core-frontend/imodelapp/imodelappoptions/) (frontend) and/or [IModelHostOptions](https://www.itwinjs.org/reference/core-backend/imodelhost/imodelhostoptions/) (backend). The app can decide how to choose the configuration (e.g., using a feature flagging service, hard-coding features they always/never want enabled, whatever). The `NativeAppStorage` APIs you mention aren't really relevant here.

But that would again a code only approach, which means, clients have to specify their options explicitly on startup. 
They can (or should) take the setting from an env (for example), but that would mean we'd have to add the settings flag to the public api (the startup options in this case). Which is not ideal, as we originally planned to "hide" that flag. using the stores directly  could help archive that, but it's far less intuitive than through the options.

On the other hand, if I look at the options in the IModelAppOptions or IModelHostOptions, the public settings rather allow to set whole components than having flags that control the behavior of certain features inside the internals IModels have to work as they work, which is why I feel currently hesitant to expose that as a public option.

pmconne:
> f I look at the options in the IModelAppOptions or IModelHostOptions, the public settings rather allow to set whole components than having flags that control the behavior of certain features inside the internals IModels have to work as they work, which is why I feel currently hesitant to expose that as a public option.

I'm afraid I don't follow your misgivings. For example, [TileAdmin](https://www.itwinjs.org/reference/core-frontend/tiles/tileadmin/tileadmin.props/) configuration provided at startup is exactly a set of "flags that control the behavior of certain features inside the internals", provided for seemingly the same purposes as your new feature (either opting in to a new experimental feature, or opting out of such a feature if it causes problems).

core/ecschema-metadata/src/IncrementalLoading/IncrementalSchemaLocater.ts
rschili: I was just looking at this because Colin asked me to put the new schema locking mechanism behind a feature flag as well, and I wonder how that's going to be used by both backend and frontend... I was hoping to copy the approach from here, but this seems to need more thought.

common/changes/@itwin/core-backend/add-IncrementalSchemaLoading-to-imodel-schemacontexts_2025-10-02-13-03.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
```
grigasp, line 6: This is a pretty significant behavior change which also includes API additions - is it really not worth mentioning in the changelogs?

IMO, it's not only worth mentioning here, but also warrants a section in `NextVersion.md`.

StefanApfel-Bentley: I did not update the comment, because it's still beta at the moment and disabled by default because we noticed it slows down the execution of small schemas slightly but with the amount of tests we run in our pipelines, it can happen the pipeline times out (>60min).

I'll have to do deeper investigations here, whether we're able to switch to the props method for imodels if the schemas are under a certain complexity, but that threshold is yet to find.

grigasp: > disabled by default

That completely caught me off guard, since the configuration option is named `disableIncrementalSchemaLoading`, which suggests it's enabled by default (`undefined`, being a falsy value, should generally mean the same thing as `false`). But sounds like I have to pass `disableIncrementalSchemaLoading: false` to enable it? I think it's very counter-intuitive, and would suggest switching to `enableIncrementalSchemaLoading` flag. If you plan to switch the default at some point, then probably a string union is better, with docs saying which value is the default: `incrementalSchemaLoading: "enabled" | "disabled"`.

> it slows down the execution of small schemas slightly but with the amount of tests we run in our pipelines, it can happen the pipeline times out (>60min)

I suppose "slightly" is negligible in real-world apps? Why not have it enabled by default and disable just in tests, then?

StefanApfel-Bentley: Yeah, good point and I like the approach of `incrementalSchemaLoading: "enabled" | "disabled"` which seems to be the best approach for the setting. I'll change it accordingly

> I suppose "slightly" is negligible in real-world apps? Why not have it enabled by default and disable just in tests, then?

We have a few more changes in branches that promise some performance improvements. Since we benefit most in the frontend from the incremental loading, I could imagine to have it enabled for the frontends. A lot of our tests uses the schemas from imodels (backend), if we decide to use incremental schema loading as default way to load schemas from imodels, I think, our tests shall use that too - unless it's a specific test scenario.

grigasp: > Yeah, good point and I like the approach of `incrementalSchemaLoading: "enabled" | "disabled"` which seems to be the best approach for the setting. I'll change it accordingly

I would still add entries to change files and `NextVersion.md`. How else are consumers supposed to find out about this new feature, especially if they have to opt-in?

StefanApfel-Bentley:
I enabled incremental schema loading for the frontends and despite the tests succeeded locally, they now failed in all three machines with Unable to locate the unit Units.M.

We have to double check that, maybe we have a found a race-condition issue here. So I'll disable incremental schema loading for the frontend again and file a bug to get that fixed (as it would be out of scope for this PR)

aruniverse:
@StefanApfel-Bentley please wait for the author of comments to resolve the comment, and don't resolve them yourself.
@grigasp can you review the changes since and make sure they satisfy your concerns

grigasp:
> @grigasp can you review the changes since and make sure they satisfy your concerns

Yes, looks good. Thanks.

docs/changehistory/NextVersion.md
```diff
@@ -57,3 +58,29 @@ For detailed information about the recommended KindOfQuantity to use in your too
   const traverseContent = createContentTraverser(myVisitor);
   traverseContent(content.descriptor, content.contentSet);
   ```
+
+## Incremental Schema Loading
+
+Incremental Schema Loading support has been added to the the backend on `IModelDb` and `IModelApp` on frontends. In this release, the incremental schema loading is disabled by default, but clients can enable it using the incrementalSchemaLoading setting on the `IModelDb` and `IModelApp` options. Incremental schema loading allows to load a schema partially. Clients can recieve their schema and its elemehts while references and lazy loaded schema elements are loaded in the background.
```
aruniverse, line 64: Shouldnt this be `IModelApp`/`IModelHost` and `IModelDb`/`IModelConnection` ?

StefanApfel-Bentley: I fixed the issue and reworked the sentences a bit: #8793

PR #8598, opened 2025-10-02
Update Version Bump to Build Presentation Common after the Deprecation Date Commit

andremig-bentley:
Cherry-Pick of https://github.com/iTwin/itwinjs-core/pull/8592.

- Adds a step to the version bump pipeline that builds itwin/presentation-common after running the Deprecation Date Commit lint rule

- Adds a variable "VERSION_BUMP" to the itwin/presentation-common ruleset-json-schema:post-process script to allow for changes during ADO CI

This is needed because itwin/presentation-common compiles function comments and deprecation notes in a "Ruleset.schema.json" file as a part of its build step. If there are Deprecation Date changes in presentation-common from the lint rule, they wont be added to this ruleset json file until presentation-common is next built. This results in a failure in the build pipeline as the ruleset json file checks for mismatches and will error if the changes are not committed.

By rebuilding presentation-common after running the lint rule, we ensure that the ruleset file is regenerated, and any potential changes to it will be committed along with the other lint rule changes.


tcobbs-bentley (approved):

andremig-bentley:
> I'd recommend taking @saskliutas advice from [#8592 (comment)](https://github.com/iTwin/itwinjs-core/pull/8592#issuecomment-3361441970) and doing a full `rush build` in `version-bump.yaml` rather than just up to `presentation-common`. That'd make the pipeline more future-proof.

The only hesitation I had with a full build instead of just to presentation-common was added time to the pipeline. After doing some testing, it seems that moving to the full build added ~ 5 minutes to the step and therefore to the pipeline as a whole. For the sake of future proofing, I think this added time is worth it, so it's been changed in the latest commit.

aruniverse (approved):

Issue #8534, opened 2025-09-17
Drawing production MVP requirements


pmconne:
**Is your feature request related to a problem? Please describe.**

@Josh-Schifter's drawing production team has minimum requirements for their initial release.

**Describe the solution you'd like**

They need to create a field in a sheet template and see that field be updated when the template is copied from a catalog iModel to create a new `BisCore.Sheet` in the design iModel.

**Tasks**

- [x] Implement `ITextAnnotation.onCloned`
  - [x] Remap element Ids in JSON when copying a field from one iModel to another
  - [x] Copy or create ElementDrivesTextAnnotation relationships between iModels
  - [x] Copy default text style or remap based on name
- [x] #8615 and cloning a text style imports font
- [x] Enhance display-test-app for testing and API validation
  - [x] Key-ins
    - [x] Save `TextDecoration`'s current annotation as a `TextAnnotation2d` element, possibly replacing the annotation on an existing element.
    - [x] Populate `TextDecoration`'s current annotation from an existing element.
    - [x] Add a `FieldRun` to `TextDecoration`
  - [x] Support specifying channels to which DTA is permitted to write.
- [x] Support all of the property types in #8490
  - [x] Formatting
    - [x] DateTime
    - [x] String
    - [ ] ~~Scale~~ not for MVP
  - [x] Field definition+evaluation
    - [x] Element properties
    - [x] [ECView properties](https://github.com/iTwin/itwinjs-core/pull/8624)
    - [x] [Project properties and Sheet properties](https://github.com/iTwin/bis-schemas/pull/614)
    - [x] SheetReference properties (ECView)
issues/8490#issuecomment-3438055100) awaiting a decision on how to persist them.
- [x] [Font cache invalidation](https://github.com/iTwin/itwinjs-core/issues/7833)
- [x] Detect when a drawing may need to be regenerated
- [x] #8101
- [x] #8745

Josh suggested testing with sheet number (`CodeValue` for `SheetReference`) to start since that one's already implemented, but see question above about when sheet references actually get created in the workflow.

pmconne:
@Josh-Schifter I need to know how scales are supposed to be formatted.

DanRod1999:
Scales will not be included as part of MVP. See [here](https://github.com/iTwin/itwinjs-core/issues/8490#issuecomment-3353282156) for original post.

pmconne:
[...]

PR #8603, opened 2025-10-02
Remap text annotations when cloned

pmconne:
Part of #8534.

When cloning a text annotation element from one iModel into another, ensure that element Ids in fields are remapped, and that the default text style is remapped, importing a copy of the style if not already present.

There appear to be some discrepancies between element types as to what cloning is supposed to do.
- Most elements do nothing when cloning within the same iModel, even if the clone context specifies relevant element Ids should be remapped. Materials are the exception - they always remap regardless (I think they just forget to check). My take is that cloning should **only** serve the purpose of creating an equivalent copy, not arbitrarily remapping bits of an element.
- Some definition elements like categories always end up in the dictionary model, others like geometry parts require the source definition model to be remapped to a target definition model. Materials do both: if the definition model is not remapped, they fall back to the dictionary model. My take is that the transformer has to be responsible for explicitly remapping definition models - we shouldn't just copy stuff into arbitrary models - and my assumption based on conversation with @bbastings is that it does.

The drawing production folks will need to test their full sheet template import workflows themselves - it uses `IModelTransformer` which depends on core-backend so can't be tested here.


bbastings:
> * Materials are the exception - they always remap regardless (I think they just forget to check).

@pmconne That does sound like an oversight...

bbastings (approved):
Changes appear to be doing things in the right places. :)

aruniverse:

pmconne:
@Alfonso-Martello `AnnotationTextStyle.create` seems to expect the caller to provide the JSON versioning if they want to supply any settings - that should all be handled by you internally, public APIs should not be aware of versioning.


See this failing test for example.
```
  1) TextAnnotation element
       defaultTextStyle
         TextAnnotation2d
           onCloned
             between iModels
               remaps multiple occurrences of same style to same Id:
     Error: Error inserting element [JSON version is missing or invalid.]
      at validateAndMigrateVersionedJSON (src/annotations/TextAnnotationElement.ts:40:11)
      at parseTextAnnotationData (src/annotations/TextAnnotationElement.ts:70:10)
      at Function.validateVersionAndUpdateGeometry (src/annotations/TextAnnotationElement.ts:228:32)
      at Function.onInsert (src/annotations/TextAnnotationElement.ts:207:10)
      at Elements.insertElement (src/IModelDb.ts:2316:53)
      at getRemappedTextStyleId (src/annotations/TextAnnotationElement.ts:536:42)
      at remapTextStyle (src/annotations/TextAnnotationElement.ts:502:22)
      at Function.onCloned (src/annotations/TextAnnotationElement.ts:306:5)
      at IModelElementCloneContext.cloneElement (src/IModelElementCloneContext.ts:173:24)
      at Context.<anonymous> (src/test/annotations/TextAnnotation.test.ts:744:36)
      at processImmediate (node:internal/timers:483:21)
```

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -457,6 +475,86 @@ export class TextAnnotation3d extends GraphicalElement3d /* implements ITextAnno
...
+  }
+
+  /** @internal */
+  protected static override onCloned(context: IModelElementCloneContext, srcProps: TextAnnotation3dProps, dstProps: TextAnnotation3dProps): void {
+    super.onCloned(context, srcProps, dstProps);
+
+    const srcElem = TextAnnotation3d.fromJSON(srcProps, context.sourceDb);
+    ElementDrivesTextAnnotation.remapFields(srcElem, context);
+    const anno = srcElem.getAnnotation();
+    dstProps.textAnnotationData = anno ? JSON.stringify(anno.toJSON()) : undefined;
```
Alfonso-Martello:
```suggestion
    dstProps.textAnnotationData = anno ? JSON.stringify({ version: TEXT_ANNOTATION_JSON_VERSION, data: anno.toJSON() } : undefined;
```
Shouldn't the stringified data be versioned?

pmconne: I wrote this before merging in all your versioning stuff.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -287,6 +288,23 @@ export class TextAnnotation2d extends AnnotationElement2d /* implements ITextAnn
...
+  }
+
+  /** @internal */
+  protected static override onCloned(context: IModelElementCloneContext, srcProps: TextAnnotation2dProps, dstProps: TextAnnotation2dProps): void {
+    super.onCloned(context, srcProps, dstProps);
+
+    const srcElem = TextAnnotation2d.fromJSON(srcProps, context.sourceDb);
+    ElementDrivesTextAnnotation.remapFields(srcElem, context);
+    const anno = srcElem.getAnnotation();
+    dstProps.textAnnotationData = anno ? JSON.stringify(anno.toJSON()) : undefined;
```
Alfonso-Martello:
```suggestion
dstProps.textAnnotationData = anno ? JSON.stringify({ version: TEXT_ANNOTATION_JSON_VERSION, data: anno.toJSON() }) : undefined;
```
Version change also needed for 2d text

PR #8607, opened 2025-10-03
Remove support for reading contour info from GPU

pmconne:
Reverts #7986.

The people we added this for will not use it, as it won't efficiently solve all their use cases.
It's hogging the last available fragment shader output, complicating the fix for #8101 and limiting future possible uses of that output.

TODO: run ImageTests.


pmconne:
V4 ImageTests had no differences from reference besides the usual (mostly map-related) flakiness.
V1 had several differences like the one below where it kinda looks like patches of dirt polluting the image.
I fail to reproduce any of these images in display-test-app on my machine.

image

Views:
- Apollo_11_height-smooth and hillshade-stepped
- HealthCare_X6, X8, and Y0
- KA27_Heli_M0
- marsiellesURL_Test3 and Test5
- StoneTowers X2 and X5
- TurbineFBX_Legs
- TyrensLargeTextures_Fitted

pmconne:
ImageTests anomalies are known flakery, nothing to do with this PR. We need to fix it so we can rely on our tests.

PR #8609, opened 2025-10-03
Added custom `portal` property for ArcGIS map-layers, to support OAuth2 authentication.

mdastous-bentley:
We previously inferred portal URL from map-layer URL, which can potentially to lead to failures.  Instead, if a ArcGIS enterprise server has a custom portal URL, end-user should should be able to supply it part of its map-layer definition, and the provider will pass the information to access client.

This will also align with the OAuth2 support offered by `@esri/arcgis-rest-request`

### DIFF
--- a/common/changes/@itwin/core-backend/2025-10-07-13-17.json
+++ b/common/changes/@itwin/core-backend/2025-10-07-13-17.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -22,8 +22,8 @@ importers:
         specifier: ^12.28.0
         version: 12.28.0
       '@bentley/imodeljs-native':
-        specifier: 5.3.8
-        version: 5.3.8
+        specifier: 5.3.9
+        version: 5.3.9
       '@itwin/object-storage-azure':
         specifier: ^3.0.4
         version: 3.0.4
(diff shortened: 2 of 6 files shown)


core/frontend/src/tile/map/MapLayerAuthentication.ts
```diff
@@ -27,8 +27,12 @@ export interface MapLayerAccessToken {
 
 /** @beta */
 export interface MapLayerAccessTokenParams {
+  /* @deprecated in 5.2.1 use `portal` instead */
```
aruniverse:
```suggestion
  /* @deprecated in 5.2.x use `portal` instead */
```

core/frontend/src/internal/tile/map/ImageryProviders/ArcGISImageryProvider.ts
```diff
@@ -99,6 +99,7 @@ export abstract class ArcGISImageryProvider extends MapLayerImageryProvider {
     if (this._accessTokenRequired && this._accessClient) {
       this._lastAccessToken = await ArcGisUtilities.appendSecurityToken(urlObj, this._accessClient, {
         mapLayerUrl: new URL(this._settings.url),
+        portal: typeof this._settings.properties?.portal === "string" ? this._settings.properties.portal : undefined,
```
aruniverse, line 102: why is there a type check here?
why isnt just assigning portal to settings.properties.portal ?

mdastous-bentley: `settings.properties` allow any primitive type of properties.  In this case, we will only accept a `portal` library defined as `string`.  

Do you see any issue doing so ?

mdastous-bentley:
@aruniverse Ideally we would like get this backported to latest release branch as we need it in the short term.

aruniverse:

aruniverse (approved):

eringram (approved):

PR #8615, opened 2025-10-07
Make element cloning async

pmconne:
This began as part of #8534.
Cloning a text style requires importing a font, which is an async operation. Cloning a text annotation element requires cloning a text style. I expect some other element types may need to perform asynchronous operations during cloning as well.

`Element.onCloned` was previously synchronous, returning `void`. Now it returns `void | Promise<void>`. This permits existing subclasses that declare their return types as `void` to continue compiling, but they will receive a lint error (assuming they've configured `no-floating-promises`) if they call `super.onCloned` without `await`.

`onCloned` is `protected`, so only subclasses of `Element` can invoke it legally. `IModelElementCloneContext.cloneElement` calls it illegally via `jsClass["onCloned"]`. So does `IModelTransformer` from the imodel-transformer package, which overrides (and does not call -  but almost entirely copy-pastes the contents of - `super.cloneElement`).

`IModelElementCloneContext.cloneElement` is `@internal`. It used to return `ElementProps`. Now it returns `Promise<ElementProps>`.

The only implementations of `onCloned` that actually return a `Promise` are all currently leaves in the BIS class hierarchy: `TextAnnotation2/3d` and `AnnotationTextStyle`. So existing code that currently calls `super.onCloned` without `await` will continue to function the same. However, existing code that calls `IModelElementCloneContext.cloneElement` without `await`ing will no longer compile. Since that's an `@internal` API, theoretically nobody should be calling it therefore nobody should be affected.

We need to talk to the transformer people about how their package and its consumers (including "evergreen" products like drawing production) are supposed to work with the transformer's dependencies on `@internal` APIs.

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -4941,6 +4941,8 @@ export class IModelApp {
     static get hasRenderSystem(): boolean;
     static get hubAccess(): FrontendHubAccess | undefined;
     static get initialized(): boolean;
+    // @internal
+    static get isEventLoopStarted(): boolean;
     static get localization(): Localization;
     // (undocumented)
     static get locateManager(): ElementLocateManager;
--- a/common/changes/@itwin/core-frontend/check-event-loop-started_2025-10-15-15-20.json
+++ b/common/changes/@itwin/core-frontend/check-event-loop-started_2025-10-15-15-20.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/core/frontend/src/IModelApp.ts
+++ b/core/frontend/src/IModelApp.ts
@@ -525,6 +525,11 @@ export class IModelApp {
       }, IModelApp.animationInterval.milliseconds);
   }
 
+  /** Return true if the main event processing loop has been started.
+   * @internal
+   */
+  public static get isEventLoopStarted() { return IModelApp._wantEventLoop; }
+
   /** @internal */
   public static startEventLoop() {
     if (!IModelApp._wantEventLoop) {
--- a/core/frontend/src/tools/ToolAdmin.ts
+++ b/core/frontend/src/tools/ToolAdmin.ts
@@ -790,6 +790,10 @@ export class ToolAdmin {
    * @internal
    */
   public static addEvent(ev: Event, vp?: ScreenViewport): void {
+    // Don't add events to queue if event loop hasn't been started to process them...
+    if (!IModelApp.isEventLoopStarted)
+      return;
+
     if (!ToolAdmin.tryReplace(ev, vp)) // see if this event replaces the last event in the queue
       this._toolEvents.push({ ev, vp }); // otherwise put it at the end of the queue.
 


pmconne:
@khanaffan @ColinKerr @aruniverse problem: `Element.onCloned` is not async. I have to acquire codes, which is async. We've had other cases where people needed callbacks like this to do async stuff. Do we have a plan for addressing that?

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -754,4 +754,23 @@ export class AnnotationTextStyle extends DefinitionElement {
...
+    const font = TextStyleSettings.fromJSON(settingsProps?.data).font;
+
+    const fontsToEmbed = [];
+    for (const file of context.sourceDb.fonts.queryEmbeddedFontFiles()) {
+      if (file.type === font.type && file.faces.some((face) => face.familyName === font.name)) {
+        fontsToEmbed.push(file);
+      }
+    }
+
+    await Promise.all(fontsToEmbed.map((file) => context.targetDb.fonts.embedFontFile({ file })));
```
pmconne: This won't compile, `onCloned` is not `async`.

khanaffan:
> @khanaffan @ColinKerr @aruniverse problem: `Element.onCloned` is not async. I have to acquire codes, which is async. We've had other cases where people needed callbacks like this to do async stuff. Do we have a plan for addressing that?

We’re considering two strategies to enable asynchronous handling:

1. **Move the relevant code into TypeScript from Native**, which would allow us to make the handlers asynchronous directly.
2. **A less invasive approach** is to split the native function into `begin` and `end` phases. This would return control to the TypeScript side between the two phases, enabling it to invoke asynchronous handlers. I used this technique during the rebase work, where I broke down a large native function into steps and triggered async events between them.

I believe the second approach is feasible for these handlers, as it requires minimal code changes. However, this work will need to be scheduled.

pmconne:
> We’re considering two strategies to enable asynchronous handling:
> 
>     1. **Move the relevant code into TypeScript from Native**, which would allow us to make the handlers asynchronous directly.
> 
>     2. **A less invasive approach** is to split the native function into `begin` and `end` phases. This would return control to the TypeScript side between the two phases, enabling it to invoke asynchronous handlers. I used this technique during the rebase work, where I broke down a large native function into steps and triggered async events between them.
> 
> 
> I believe the second approach is feasible for these handlers, as it requires minimal code changes. However, this work will need to be scheduled.

`Element.onCloned` is not invoked by native. It's invoked by `IModelElementCloneContext.cloneElement`, which is synchronous. We could add `cloneElementAsync` and deprecate `cloneElement`. We could also change `Element.onCloned` to be async. All these APIs are `@beta`. However, making such breaking changes would torpedo apps who are definitely using these APIs and are relying on us to be "evergreen".

~~I'll try to come up with a way to add async cloning while preserving backward compatibility.~~ See updated PR description.

pmconne:
@DanRod1999 anyone else in imodel-transformer orbit who should review this PR?

DanRod1999:
FYI @mindaugasdirg

pmconne:
@khanaffan more concurrent query test flakery.
```
--[ FAILURE: @itwin/core-backend ]-----------------[ 4 minutes 32.7 seconds ]--


  1692 passing (4m)
  26 pending
  1 failing

  1) ConcurrentQuery
       restart query:

      AssertionError: expected 3 to equal 2
      + expected - actual

      -3
      +2
      
      at Context.<anonymous> (src/test/ecdb/ConcurrentQuery.test.ts:135:28)
```

khanaffan:
I have seen this failing today. I looking at it.

core/backend/src/annotations/ElementDrivesTextAnnotation.ts
```diff
@@ -148,7 +148,7 @@ export class ElementDrivesTextAnnotation extends ElementDrivesElement {
 
   /** When copying an [[ITextAnnotation]] from one iModel into another, remaps the element Ids in any [FieldPropertyHost]($common) within the cloned element
    * so that they refer to elements in the `context`'s target iModel, and sets any Ids that cannot be remapped to [Id64.invalid]($bentley).
-   * Implementations of `ITextAnnotation` should invoke this function from their implementations of [[Element._onCloned]].
+   * Implementations of `ITextAnnotation` should invoke this function from their implementations of [[Element.onCloned]].
```
pmconne, line 151: @ben-polinsky curious why this didn't break docs build previously (no such method `Element._onCloned`).

ben-polinsky: Looked into it briefly and I'm not sure. @andremig-bentley and I will see what's going on.

pmconne:
@dassaf4 @saeeedtorabi Voronoi test flaked out.
```
stdout | src/test/topology/Voronoi.test.ts > Voronoi > CurveChain
ERROR
[
  'Expect defined',
  undefined,
  [ 'interior point containing face found' ]
]

 ❯ src/test/topology/Voronoi.test.ts (14 tests | 1 failed) 1055ms
```

```
--[ FAILURE: @itwin/core-geometry ]-------------------------[ 59.41 seconds ]--


⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/test/topology/Voronoi.test.ts > Voronoi > CurveChain

- Expected
+ Received

- 0
  ...2 lines omitted...
 ❯ src/test/topology/Voronoi.test.ts:795:31
    793|     }
    794|     GeometryCoreTestIO.saveGeometry(allGeometry, "Voronoi", "CurveChai…
    795|     expect(ck.getNumErrors()).toBe(0);
    796|   });
    797| 

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

```

saeeedtorabi:
> @dassaf4 @saeeedtorabi Voronoi test flaked out.

@pmconne This test generates 100 random points and had never failed until now. Since it's random, we cannot regenerate it. However, we added logs here https://github.com/iTwin/itwinjs-core/pull/8639, so next time it failed we can regenerate it. Please note that I ran it 10 more times locally (i.e., for 10000 random points) and it never failed.

core/backend/src/Element.ts
```diff
@@ -387,7 +387,7 @@ export class Element extends Entity {
    * @note If you override this method, you must call super.
    * @beta
    */
-  protected static onCloned(_context: IModelElementCloneContext, _sourceProps: ElementProps, _targetProps: ElementProps): void { }
+  protected static onCloned(_context: IModelElementCloneContext, _sourceProps: ElementProps, _targetProps: ElementProps): Promise<void> | void { }
```
aruniverse, line 390: should this alway be async? should we drop void as the return?

pmconne: PR description explains why.

> This permits existing subclasses that declare their return types as void to continue compiling, but they will receive a lint error (assuming they've configured no-floating-promises) if they call super.onCloned without await.

PR #8616, opened 2025-10-08
ECSqlReader/ConcurrentQuery unable to see modification to file

khanaffan:
fixes: https://github.com/iTwin/itwinjs-backlog/issues/1657

`ConcurrentQuery` relies on `PRAGMA data_version` to detect if the database file has been modified by another connection. However, this mechanism doesn't work reliably when there are active statements, as these likely lock the database and prevent `PRAGMA data_version` from reflecting any changes until those statements are closed.

In ECDb, `InstanceReader` maintains active statements for performance reasons. For example, executing a query like `SELECT $ FROM Bis.Element` creates an active statement, which in turn blocks `PRAGMA data_version` from updating.

As a workaround, we now check the primary connection to determine if it has changed since the last time a query was executed via `ConcurrentQuery`. If a change is detected, we clear the cache to ensure consistency.

### DIFF
--- a/common/changes/@itwin/core-backend/affank-query-reader-bugfix_2025-10-08-02-26.json
+++ b/common/changes/@itwin/core-backend/affank-query-reader-bugfix_2025-10-08-02-26.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "ECSqlReader/ConcurrentQuery unable to see modification to file",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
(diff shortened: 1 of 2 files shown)


aruniverse:
@khanaffan you messed up linking the prs so the addon was published directly to master instead of this pr

DanRod1999:
Sorry Affan, don't think it was directly related to the PR linking. I believe it linked correctly during PR val. It appears the reason the addon didn't properly merge here was because your commit message didn't contain the PR number like normal PRs do (unsure how this happens). Because of this your addon didn't get merged back to this PR.

image

(#PRNUM) should be on the end of this by default

khanaffan:
> @khanaffan you messed up linking the prs so the addon was published directly to master instead of this pr

What went wrong with the linking?

aruniverse:
> > @khanaffan you messed up linking the prs so the addon was published directly to master instead of this pr
> 
> What went wrong with the linking?

the merge commit missing the pr num that dan mentioned above; not sure how that happened. its nbd; something we need to keep an eye on

khanaffan:
I created PR from VS code and set it to auto squash. May be AI help fix the message.

aruniverse:

PR #8619, opened 2025-10-09
Bulk relationship instance delete API

RohitPtnkr1996:
imodel-native: https://github.com/iTwin/imodel-native/pull/1237

Fixes: https://github.com/iTwin/itwinjs-backlog/issues/1514

### DIFF
--- a/common/changes/@itwin/certa/master_2025-10-28-08-20.json
+++ b/common/changes/@itwin/certa/master_2025-10-28-08-20.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/certa",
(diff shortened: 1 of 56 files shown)


full-stack-tests/backend/src/perftest/DeleteRelationshipInstances.test.ts
```diff
@@ -0,0 +1,404 @@
+/*---------------------------------------------------------------------------------------------
```
ColinKerr, line 1: Behavioral/Unit tests need to be implemented independently from performance tests

RohitPtnkr1996: Done

PR #8623, opened 2025-10-09
Add PI construction to `Arc3d`

dassaf4:
This is a generalization more suitable to `Arc3d`. When the instance is a circular fillet, the default inputs yield the arc's PI.

### DIFF
--- a/common/changes/@itwin/core-backend/2025-10-15-11-14.json
+++ b/common/changes/@itwin/core-backend/2025-10-15-11-14.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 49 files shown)


dassaf4:
@saeeedtorabi Here's an untested implementation. Please generate some tests and verify. One way to verify for circular arc inputs is to call the line string fillet method you recently worked on and try to recover the original line string vertices removed by each fillet arc. Some edge cases to consider:
* f0 === f1
* 0/pi/2pi sweeps (or within `Geometry.smallAngleRadians`)
* non-circular arc
* non-xy arc

saeeedtorabi (approved):

Issue #8631, opened 2025-10-13
Updating Elements Code value does not clear InstanceKeyCache


DanRod1999:
**Describe the bug**
Element's instanceKeyCache is not properly cleared when updating an element's code value. This bug was found in an imodel-transformer test and can be replicated with this test below added to IModel.test.ts. Note a better/simpler test should be written since this was specific to transforming elements. Just updating one element's code should suffice.

**To Reproduce**
```
  it.only("should successfully process changes when Definition Elements' codeValues are switched around", async () => {
    const dbFileName = IModelTestUtils.prepareOutputFile("IModel", "change-codeValues.bim");
    const imodelDb = SnapshotDb.createEmpty(dbFileName, {
      rootSubject: { name: "change-codeValues" },
    });
    let categoryA = SpatialCategory.create(
      imodelDb,
      IModel.dictionaryId,
      "A"
    );
    let categoryB = SpatialCategory.create(
      imodelDb,
      IModel.dictionaryId,
      "B"
    );
    categoryA.userLabel = "A";
    categoryB.userLabel = "B";
    categoryA.insert();
    categoryB.insert();
    imodelDb.saveChanges();

    categoryA = imodelDb.elements.getElement(
      SpatialCategory.createCode(imodelDb, IModel.dictionaryId, "A")
    );
    categoryB = imodelDb.elements.getElement(
      SpatialCategory.createCode(imodelDb, IModel.dictionaryId, "B")
    );
    categoryA.code.value = "temp";
    categoryA.update();
    categoryB.code.value = "A";
    categoryB.update();
    categoryA.code.value = "B";
    categoryA.update();
    imodelDb.saveChanges();

    categoryA = imodelDb.elements.getElement(
      SpatialCategory.createCode(imodelDb, IModel.dictionaryId, "A")
    );
    categoryB = imodelDb.elements.getElement(
      SpatialCategory.createCode(imodelDb, IModel.dictionaryId, "B")
    );
```

**Expected behavior**
userLabel of category should properly update, and element cached results should be cleared during update
-->

DanRod1999:
[WIP PR](https://github.com/iTwin/itwinjs-core/pull/8632)

PR #8632, opened 2025-10-13
Clear Element's InstanceKeyCache during element updates

DanRod1999:
Resolves: https://github.com/iTwin/itwinjs-core/issues/8631

Clear instanceKeyCache during updateElement so when retrieving element we don't return an invalid cached result.


core/backend/src/IModelDb.ts
```diff
@@ -2334,6 +2334,10 @@ export namespace IModelDb {
      */
     public updateElement<T extends ElementProps>(elProps: Partial<T>): void {
       try {
+        if (!elProps.id) {
+          throw new Error("id must not be undefined when clearling instanceKeyCache");
```
MichaelSwigerAtBentley: I don't think we should throw like this-

I see that Id is required for updateElement. Why do we even need fedGuid or code at all then? Is it just so we can directly use an elProps that is returned from a getElement?

Either way, I think we should throw this error in instanceKeyCache.deleteById since its cache specific, and add a more generic error into update element, somthing along the lines of `element id must be provided`

DanRod1999: is it fine to only delete if id is present and then just not delete the instance cache otherwise
```
        if (elProps.id) 
          this[_instanceKeyCache].deleteById(elProps.id)
```

MichaelSwigerAtBentley: According to the comment on updateElement:

`All parts of `elProps` are optional *other than* `id`. If id is missing, an exception is thrown.`
But that's not true as of right now?
So then we should always have an id in update element according to the comment.

That being said, if updateElement works without id, (which is what I would expect, as just FedGuid is enough to locate an element) then we cannot just clear the cache if id is present. It must always be cleared:

What if someone queried an element by code (and just by code). That would cache the code and the resulting Id.
Then someone calls updateElement with just the fedGuid or code, and updates the code. The cache will be broken.

MichaelSwigerAtBentley: Also as a side note, if we do add a error if Id is not defined, it should be outside of the catch block, and should throw an iTwinError, not a generic Error

DanRod1999: `delete(resolveInstanceKeyArgs)` also requires id to be defined. However, it is a little weird because it takes in a "partialKey" which can be undefined, however, the props within the partial key must be defined, so maybe the solution is making those elements potentially undefined

```
  interface ResolveInstanceKeyArgs {
    partialKey?: { id: Id64String, baseClassName: string };
    federationGuid?: GuidString;
    code?: CodeProps;
  }
```

MichaelSwigerAtBentley: Id only needs to be defined if a partialKey is provided. The required Id/baseClassName are not inherited as required when creating a ResolveInstanceKey

MichaelSwigerAtBentley: In other words, the cache can be searched by 3 different things:
partialKey (which is just Id) || fedGuid || code

Or any combination of them, such that the result will be cached to all possible combinations.
If you provide a fedGuid and a Code forinstance, then just search by code next time, you should get a cache hit

MichaelSwigerAtBentley (approved):

core/backend/src/test/ElementLRUCache.test.ts
```diff
@@ -495,6 +495,22 @@ describe('InstanceKeyLRUCache', () => {
...
+  cache.set(testArgs3, testResults[2]);
+  expect(cache.size).to.equal(3);
+
+  expect(() => cache.delete({
+    partialKey: undefined,
+    federationGuid: undefined,
+    code: undefined
+  })).to.throw(Error, "ResolveInstanceKeyArgs must have a partialKey, code, or federationGuid");
+
+  expect(cache.size).to.equal(3);
```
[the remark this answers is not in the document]
DanRod1999: resolved in own commit

khanaffan (approved):

core/backend/src/IModelDb.ts
```diff
@@ -2334,6 +2334,13 @@ export namespace IModelDb {
      */
     public updateElement<T extends ElementProps>(elProps: Partial<T>): void {
       try {
+        if (elProps.id)
```
aruniverse: dont really love conditionals w/o braces, esp for lines like 2340-2343

DanRod1999:

PR #8637, opened 2025-10-15
Improve usability of editing api

RohitPtnkr1996:
**WIP**

### DIFF
--- a/editor/backend/EditingApiRedesign.md
+++ b/editor/backend/EditingApiRedesign.md
@@ -1,335 +1,395 @@
-# Edit Command Design Document
+*** **Work In Progress** ***
 
-## Summary
+# Current Design Overview:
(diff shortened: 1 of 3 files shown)


core/backend/src/UtilityFunctions.ts
MichaelSwigerAtBentley, line 1: I think this file should be renamed into something more specific, 'UtilityFunctions' is very broad. 

Also, are these utils only for use within iModelDb.ts? If so, we should look at if they should be internal and not flagging them for public use.

bbastings:
Please stop working on this. @calebmshafer @khanaffan

RohitPtnkr1996:
This investigation/proposal is put on hold until a better time.
Closing this PR.

PR #8639, opened 2025-10-15
Add log for verifyVoronoiTopology

saeeedtorabi:


pmconne:
This is an attempt to gather more info on [this flaky failure](https://github.com/iTwin/itwinjs-core/pull/8615#issuecomment-3405834382), right?

saeeedtorabi:
> @saeeedtorabi the solution is to never use random data in tests - that makes them inherently flaky. You could of course generate the data randomly ahead of time and copy-paste it into your test, but the test should always execute against the same data.

ok @pmconne I will generate fixed points so the test is not flaky. However, we expect the test to pass for any random point and each failure shows a potential bug in the src code. Therefore, I keep random points but put them behind the flag `enableLongTests` so CI builds, which run with this flag disabled, don't fall over due to flakiness. However, we will need a dedicated CI job running geometry tests with this flag enabled so that the geom team can discover the rare edge cases like this. We would need someone knowledgeable to set this up for us.

DanRod1999:
if this fixes a flaky test is it worth back porting?
I got another [failure](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4432545&view=logs&j=751a5ea6-e147-5a0b-b7c5-e28ecfcb00e2&s=96ac2280-8cb4-5df5-99de-dd2da759617d&t=742556c9-a44d-5542-8a03-cf20a9c5ac3e&l=24246) on a bp pr but not sure if its related to this
```
 FAIL  src/test/topology/Voronoi.test.ts > Voronoi > GraphWith2Triangles

- Expected
+ Received

- 0
  ...2 lines omitted...
 ❯ src/test/topology/Voronoi.test.ts:577:31
    575| 
    576|     GeometryCoreTestIO.saveGeometry(allGeometry, "Voronoi", "GraphWith…
    577|     expect(ck.getNumErrors()).toBe(0);
    578|   });
    579| 
```

saeeedtorabi:
> if this fixes a flaky test is it worth back porting? 

This is strange @dassaf4 
The report says `GraphWith2Triangles` test failed but log message is for `findContainingFaceXY` test
image

by the way, `findContainingFaceXY` is also generating random test points so we may want to guard it by `enableLongTests`

dassaf4:
> However, we will need a dedicated CI job running geometry tests with this flag enabled so that the geom team can discover the rare edge cases like this. We would need someone knowledgeable to set this up for us.

@aruniverse Maybe this would be a good onboarder project....

> The report says `GraphWith2Triangles` test failed but log message is for `findContainingFaceXY`

This also had me confused until I realized the same log message is used in another test in this file. D'oh!

> by the way, `findContainingFaceXY` is also generating random test points so we may want to guard it by `enableLongTests`

@saeeedtorabi Good idea. Wouldn't hurt to make a new PR to go over all core-geometry tests to 1) shunt randomness to the `enableLongTests` scope, and 2) replace the opposite scope with a smaller sample of statically computed points like you did here. Then get with Dan to backport it where he needs it.

> if this fixes a flaky test is it worth back porting?

@DanRod1999 Yes indeed. Make it so! Voronoi.test.ts is a fairly new file, so can't go too far back....

saeeedtorabi:
> go over all core-geometry tests to 1) shunt randomness to the `enableLongTests` scope, and 2) replace the opposite scope with a smaller sample of statically computed points like you did here. Then get with Dan to backport it where he needs it.

@dassaf4  Created an issue for it https://github.com/iTwin/itwinjs-core/issues/8675
I can work on it after spiral intersection issue is done.

DanRod1999:

Issue #8635, opened 2025-10-14
Typing in an Input field triggers ToolAdmin.processShortcutKey after changing frontstages.


mathieu-fournier:
The bug is easier to explain with repro steps. 
Repro steps : 
1. When the app loads, we start on the imodel selection frontstage. 
2. I search for an IModel typing "mf". 
<img width="616" height="163" alt="Image" src="https://github.com/user-attachments/assets/609ce02c-2f56-45cb-8c2a-958957a73924" />

3. I pick an IModel by clicking on a tile. 
4. The Main frontstage loads and `ToolAdmin.processShortcutKey` gets wrongly called for the previousely typed "mf" keys. This activates shortcuts without any intention to do so. 

Process shortcut key should not have been called because those keys where typed in an input field and in another frontstage. 
The desired behaviour would be to not have shortcuts activated at that moment. AKA ToolAdmin.processShortcutKey should not have been called. 

Core version is 5.2.1

Looking at the callstack (see attached file), this is what happens. 
Activating the frontstage loads the viewport, which starts the event loop, which then calls process Event, which process shortcut keys. 
The issue is that, at that moment, there should be no event/keys to process. 

Thanks

[CallStack_UIFramework_setActiveFrontstage_Calling_ToolAdmin_processShortcutKey_01.txt](https://github.com/user-attachments/files/22909820/CallStack_UIFramework_setActiveFrontstage_Calling_ToolAdmin_processShortcutKey_01.txt)

mathieu-fournier:
A temporary fix to this, while the solution is behing implemented, could be to manually clear `ToolAdmin._toolEvents` before changing frontstage. 
<img width="541" height="50" alt="Image" src="https://github.com/user-attachments/assets/1c08ebfd-e420-401d-8ebc-271bb6a12942" />
Can this break other things? Seems fairly safe to me at first glance.

aruniverse:
[...]

PR #8640, opened 2025-10-15
Events should not be added to queue if event loop is not started.

bbastings:
Address #8635


aruniverse:

PR #8658, opened 2025-10-21
Intersection of spiral and different curves

saeeedtorabi:
This approximates the spiral by a line string

### DIFF
--- a/core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
+++ b/core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
@@ -1183,9 +1183,9 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
       if (newtonSearcher.runIterations()) {
         spiralFraction = newtonSearcher.getU();
         otherFraction = newtonSearcher.getV();
+        if (this.acceptFraction(false, spiralFraction, false) && this.acceptFraction(false, otherFraction, false))
+          this.recordPointWithLocalFractions(otherFraction, otherCurve, 0, 1, spiralFraction, spiral, 0, 1, reversed);
       }
-      if (this.acceptFraction(false, spiralFraction, false) && this.acceptFraction(false, otherFraction, false))
-        this.recordPointWithLocalFractions(otherFraction, otherCurve, 0, 1, spiralFraction, spiral, 0, 1, reversed);
     }
   }
   /** Double dispatch handler for strongly typed spiral curve. */
--- a/core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
+++ b/core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
@@ -2397,10 +2397,10 @@ describe("CurveCurveIntersectXY", () => {
           curve0Detail = intersection.detailB;
           curve1Detail = intersection.detailA;
         }
-        let pointOnCurve0 = Point3d.create();
-        let pointOnCurve1 = Point3d.create();
+        const pointOnCurve0 = Point3d.create();
+        const pointOnCurve1 = Point3d.create();
         if (!(curve0 instanceof CurveCollection)) {
-          pointOnCurve0 = curve0Detail.point;
+          pointOnCurve0.setFrom(curve0Detail.point);
           const intersectionPoint0 = curve0.fractionToPoint(curve0Detail.fraction);
           ck.testPoint3d(
             pointOnCurve0,
@@ -2409,7 +2409,7 @@ describe("CurveCurveIntersectXY", () => {
           );
         }
         if (!(curve1 instanceof CurveCollection)) {
-          pointOnCurve1 = curve1Detail.point;
+          pointOnCurve1.setFrom(curve1Detail.point);
           const intersectionPoint1 = curve1.fractionToPoint(curve1Detail.fraction);
           ck.testPoint3d(
             pointOnCurve1,
@@ -2439,13 +2439,18 @@ describe("CurveCurveIntersectXY", () => {
     let dx = 0;
     let dy = 0;
 
-    const rotationTransform = Transform.createFixedPointAndMatrix(
+    const rotationTransform0 = Transform.createFixedPointAndMatrix(
       Point3d.create(70, 0),
       Matrix3d.createRotationAroundVector(Vector3d.create(0, 0, 1), Angle.createDegrees(90))!,
     );
+    const rotationTransform1 = Transform.createFixedPointAndMatrix(
+      Point3d.create(0, 0),
+      Matrix3d.createRotationAroundVector(Vector3d.create(0, 1, 0), Angle.createDegrees(45))!,
+    );
     const moveTransform = Transform.createTranslationXYZ(0, 0, 10);
     const nonPlanarTransform = Transform.createZero();
-    nonPlanarTransform.setMultiplyTransformTransform(rotationTransform, moveTransform);
+    nonPlanarTransform.setMultiplyTransformTransform(rotationTransform0, moveTransform);
+    nonPlanarTransform.setMultiplyTransformTransform(rotationTransform1, nonPlanarTransform);
     // integrated spirals
     const integratedSpirals = [];
     const r0 = 0;
@@ -2454,7 +2459,7 @@ describe("CurveCurveIntersectXY", () => {
     for (const integratedSpiralType of ["clothoid", "bloss", "biquadratic", "sine", "cosine"]) {
       for (const transform of [
         Transform.createIdentity(),
-        rotationTransform, // rotated spirals have indices (n*i)+1
+        rotationTransform0, // rotated spirals have indices (n*i)+1
         nonPlanarTransform, // non-planar spirals have indices (n*i)+2
       ]) {
         integratedSpirals.push(
@@ -2486,7 +2491,7 @@ describe("CurveCurveIntersectXY", () => {
     ]) {
       for (const transform of [
         Transform.createIdentity(),
-        rotationTransform, // rotated spirals have indices (n*i)+1
+        rotationTransform0, // rotated spirals have indices (n*i)+1
         nonPlanarTransform, // non-planar spirals have indices (n*i)+2
       ]) {
         directSpirals.push(
--- a/core/geometry/src/bspline/SurfaceLocationDetail.ts
+++ b/core/geometry/src/bspline/SurfaceLocationDetail.ts
@@ -11,7 +11,7 @@ import { Point2d } from "../geometry3d/Point2dVector2d";
 import { Point3d } from "../geometry3d/Point3dVector3d";
 
 /**
- * CurveLocationDetail carries point and paramter data about a point evaluated on a curve.
+ * CurveLocationDetail carries point and parameter data about a point evaluated on a curve.
  * * These are returned by a variety of queries.
  * * Particular contents can vary among the queries.
  * @public


core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1048,6 +1049,12 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
         segmentA, this._extendA0, segmentA.point0Ref, 0.0, segmentA.point1Ref, 1.0, this._extendA1,
         this._geometryB, this._extendB0, this._extendB1, false,
       );
+    } else if (this._geometryB instanceof TransitionSpiral3d) {
+      const linestring = LineString3d.create();
+      this._geometryB.emitStrokes(linestring);
+      this.computeSegmentLineString(
+        segmentA, this._extendA0, this._extendA1, linestring, this._extendB0, this._extendB1, false,
```
dassaf4: I don't think it makes sense to compute intersections to the _extended_ linestring approximant in all these cases. In the next PR, when you refine the linestring-based intersection seeds with Newton, you can revisit allowing extension of the spiral. It is possible the spiral API accommodates extension, but for some spiral types, extending the spiral beyond its active interval becomes increasingly inaccurate.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1072,6 +1079,10 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
       this.dispatchLineStringBSplineCurve(
         lsA, this._extendA0, this._extendA1, this._geometryB, this._extendB0, this._extendB1, false,
       );
+    } else if (this._geometryB instanceof TransitionSpiral3d) {
+      const linestring = LineString3d.create();
+      this._geometryB.emitStrokes(linestring);
+      this.computeLineStringLineString(lsA, linestring, false);
```
dassaf4: Note that this method doesn't pass in extend flags. In order to disallow extension of the spiral approximant, you can add extend flags to the `computeLineStringLineString` arg list (like `computeSegmentLineString`) and refactor.

core/geometry/src/curve/spiral/IntegratedSpiral3d.ts
```diff
@@ -183,7 +183,7 @@ export class IntegratedSpiral3d extends TransitionSpiral3d {
   /**
    * Create a transition spiral with radius and bearing conditions.
    * @param radius01 radius (inverse curvature) at start and end (radius of zero means straight line).
-   * @param bearing01 bearing angles at start and end.  bearings are measured from the x axis, positive clockwise
+   * @param bearing01 bearing angles at start and end. bearings are measured from the x axis, positive clockwise
```
dassaf4:
```suggestion
   * @param bearing01 bearing angles at start and end. Bearings are measured from the x axis, positive clockwise
```

core/geometry/src/curve/spiral/NormalizedTransition.ts
```diff
@@ -96,13 +99,15 @@ export class NormalizedClothoidTransition extends NormalizedTransition {
  * @internal
  */
 export class NormalizedBlossTransition extends NormalizedTransition {
-  // bloss curve is (3 - 2x) x ^2 = 3 x^2 - 2 x^3
+  // bloss curvature is 3x^2 - 2x^3 or x^2 (3 - 2x)
```
dassaf4:
```suggestion
  // bloss curvature is 3x^2 - 2x^3 = x^2 (3 - 2x)
```

core/geometry/src/curve/spiral/NormalizedTransition.ts
```diff
@@ -96,13 +99,15 @@ export class NormalizedClothoidTransition extends NormalizedTransition {
...
   // derivative 6x (1-x)
-  // 2nd derivative 6 - 12 x
+  // 2nd derivative 6 - 12x
   // derivatives zero at 0,1
   // inflection zero at 0.5
-  // integral is x^3 - x^4 / 2 = x^3 (1-x/2)
-  constructor() { super(); }
+  // integral is x^3 - x^4/2 or x^3 (1-x/2)
```
dassaf4:
```suggestion
  // integral is x^3 - x^4/2 = x^3 (1-x/2)
```

core/geometry/src/curve/spiral/NormalizedTransition.ts
```diff
@@ -174,7 +180,12 @@ export class NormalizedBiQuadraticTransition extends NormalizedTransition {
  * @internal
  */
 export class NormalizedSineTransition extends NormalizedTransition {
-  constructor() { super(); }
+  // sine transition is f(x) = x - sin(2 pi x) / (2 pi)
+  // derivative f'(x) = 1 - cos(2 pi x)
+  // integral If(x) = x^2 / 2 + (cos(2 pi x) - 1) / (4 pi^2)
```
dassaf4: The simplest antiderivative of `f` is below. Please investigate why we aren't using it.
```suggestion
  // f'(x) = 1 - cos(2 pi x)
  // If(x) = x^2 / 2 + cos(2 pi x) / (4 pi^2)
```

core/geometry/src/curve/spiral/NormalizedTransition.ts
```diff
@@ -201,7 +212,12 @@ export class NormalizedSineTransition extends NormalizedTransition {
  * @internal
  */
 export class NormalizedCosineTransition extends NormalizedTransition {
-  constructor() { super(); }
+  // cosine transition is f(x) =  0.5 (1 - cos(pi x))
+  // derivative f'(x) = 0.5 pi sin(pi x)
+  // integral If(x) = 0.5 x - sin(pi x) / pi
```
dassaf4:
```suggestion
  // f'(x) = 0.5 pi sin(pi x)
  // If(x) = 0.5 (x - sin(pi x) / pi)
```

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
```diff
@@ -2356,11 +2360,115 @@ describe("CurveCurveIntersectXY", () => {
...
+    const ck = new Checker();
+    const allGeometry: GeometryQuery[] = [];
+    let dx = 0;
+    let dy = 0;
+
+    const rotationTransform = Transform.createFixedPointAndMatrix(
+      Point3d.create(70, 0),
+      Matrix3d.createRotationAroundVector(Vector3d.create(0, 0, 1), Angle.createDegrees(90))!,
+    );
+    const curvePrimitives = [
```
dassaf4: Ideally `curvePrimitives` should have one of _each_ type of Integrated and Direct spiral so that you are testing all combinations of spiral-spiral intersection. (At the very least, `curvePrimitives` should have an IntegratedSpiral3d so that you test the intersection of two IntegratedSpiral3d.) This will be more important for testing the second PR (Newton refinement).

core/geometry/src/curve/spiral/NormalizedTransition.ts
saeeedtorabi: I used `or` because initially I was confused and thought `3x^2 - 2x^3 = x^2 (3 - 2x)` is an equation for curvature (I mean right hand side and left hand side of an equation) . I change it to `bloss curvature is x^2 (3 - 2x)`

core/geometry/src/curve/spiral/NormalizedTransition.ts
saeeedtorabi: The `NormalizedTransition` class doc says `The symmetry condition ensures that the integral from 0 to 1 is 1/2` so I think `-1` is added to the integral to ensure that.

core/geometry/src/curve/spiral/NormalizedTransition.ts
dassaf4: I like that even better. (Do same for the integral comment a few lines down.)

core/geometry/src/curve/spiral/NormalizedTransition.ts
dassaf4: Both integrals over [0,1] are 1/2. We know this because the two antiderivatives differ only by a constant. Do the tests pass if we use the simpler integral?

saeeedtorabi: The `SnapFunctions` test expects `fractionToArea` to return 0.5 for `u = 1` 
```
ck.testCoordinate(0.5, snap.fractionToArea(1.0), "integrated area from fraction 0 to 1 is expected to be 0.5");
```
and this does not happen for 
```
  public fractionToArea(u: number): number {
    const a = 2.0 * Math.PI;
    return 0.5 * u * u + Math.cos(u * a) / (a * a);
  }
  ```
  
  `AlexGSamples` test fails too.

Comment (dassaf4) [reply core/geometry/src/curve/spiral/NormalizedTransition.ts]: Nice. The tests are doing their job! And I think I understand now: `fractionToArea` is not just any antiderivative `A` of `f` such that integrating `f` over [0,1] is `A(1) - A(0) = 1/2`. `A` is chosen such that `A(0) = 0`. I have no idea why we choose `A` this way. It seems like this fact should be documented, and the reason (once we figure it out) should be documented too. Is `NormalizedSineTransition` the only spiral for which `fractionToArea` is not the simplest antiderivative (i.e., for which the constant of integration is nonzero)?

File: core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts [hunk 9]
```diff
@@ -2388,7 +2388,7 @@ describe("CurveCurveIntersectXY", () => {
   };
 
   it("SpiralCurvePrimitiveIntersection", () => {
-    const ck = new Checker();
+    const ck = new Checker(true, true);
```
dassaf4: Reminder to remove this eventually.

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
```diff
@@ -2397,6 +2397,48 @@ describe("CurveCurveIntersectXY", () => {
       Point3d.create(70, 0),
       Matrix3d.createRotationAroundVector(Vector3d.create(0, 0, 1), Angle.createDegrees(90))!,
     );
+    const integratedSpirals = [];
+    const r0 = 0;
+    const r1 = 50;
+    const activeInterval = Segment1d.create(0, 1);
+    for (const integratedSpiralType of ["clothoid", "bloss", "biquadratic", "sine", "cosine"]) {
+      for (const transform of [Transform.createIdentity(), rotationTransform]) { // rotated spirals have even indices
```
dassaf4:
```suggestion
      for (const transform of [Transform.createIdentity(), rotationTransform]) { // rotated spirals have odd indices
```

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
```diff
@@ -2397,6 +2397,48 @@ describe("CurveCurveIntersectXY", () => {
...
+      "HalfCosine",
+      "AustralianRailCorp",
+      "WesternAustralian",
+      // TODO: enable below lines after https://github.com/iTwin/itwinjs-backlog/issues/1693 is resolved
+      // "Czech",
+      // "MXCubicAlongArc",
+      // "Italian",
+      // "Polish",
+    ]) {
+      for (const transform of [Transform.createIdentity(), rotationTransform]) { // rotated spirals have even indices
```
dassaf4:
```suggestion
      for (const transform of [Transform.createIdentity(), rotationTransform]) { // rotated spirals have odd indices
```

core/geometry/src/curve/spiral/NormalizedTransition.ts
dassaf4: This still needs to be fixed.

core/geometry/src/curve/spiral/NormalizedTransition.ts
saeeedtorabi: `fractionToArea` is only used for integrated spirals and yes, `NormalizedSineTransitio`n the only spiral for which `fractionToArea` is not the simplest antiderivative.

dassaf4: Ok, I'd like to see brief comments in the code explaining:
1. Why A is chosen such that A(0) = 0. I believe this is to support the "normalization" of the snap functions, so that `fractionToArea(1) = 1/2`.
2. The sine snap function is unique in this regard.

core/geometry/src/curve/spiral/DirectSpiral3d.ts
```diff
@@ -34,13 +34,14 @@ import { XYCurveEvaluator } from "./XYCurveEvaluator";
 * * Direct spirals that flow through serialization to native imodel02 are created with these static methods:
```
dassaf4:
```suggestion
* * Direct spirals are created with these static methods:
```

core/geometry/src/curve/spiral/IntegratedSpiral3d.ts
```diff
@@ -85,11 +85,12 @@ export class IntegratedSpiral3d extends TransitionSpiral3d {
   public static readonly defaultSpiralType = "clothoid";
   /** Use the integrated function to return an angle at fractional position. */
   public globalFractionToBearingRadians(fraction: number): number {
+    // calculate area under curvature curve from 0 to fraction and add it to start radian to get angle at the given fraction
```
dassaf4:
```suggestion
    // calculate area under curvature curve from 0 to fraction and add it to start angle to get angle at the given fraction
```

core/geometry/src/curve/spiral/NormalizedTransition.ts
```diff
@@ -183,6 +180,8 @@ export class NormalizedSineTransition extends NormalizedTransition {
   // sine transition curvature is x - sin(2 pi x) / (2 pi)
   // derivative 1 - cos(2 pi x)
   // integral x^2 / 2 + (cos(2 pi x) - 1) / (4 pi^2)
+  // note: integral or If is anti-derivative of curvature function and it's chosen such that If(0) = 0.
+  // this is to support the "normalization" of the snap functions, so that fractionToArea(1) = 1/2.
```
dassaf4: This is better split into two places. General stuff goes up with the description of `fractionToArea` in the class doc:
```
 *   * fractionToArea -- antiderivative A of f, chosen such that A(0) = 0. The value of this function at fraction u is the area under f from 0 to u. In particular, the symmetry condition on f implies that 1/2 = integral of f over [0,1] = A(1) - A(0) = A(1).
```
Then here:
```suggestion
  // note: this is the only snap function whose antiderivative is chosen with nonzero integration constant.
```

core/geometry/src/curve/spiral/TransitionSpiral3d.ts
```diff
@@ -46,16 +46,16 @@ export type IntegratedSpiralTypeName = "clothoid" | "bloss" | "biquadratic" | "c
...
-  | "Arema"              // 2 terms from each of the X,Y clothoid series expansions; identical to ChineseCubic
+  "Arema"                // 2 terms from each of the X,Y clothoid series expansions; identical to ChineseCubic
+  | "JapaneseCubic"      // 1 term from each of the X,Y clothoid series expansions; y = x^3 / (6RL)
   | "ChineseCubic"       // Identical to Arema
+  | "WesternAustralian"  // simple cubic; 2 terms of x series, 1 term of y series
   | "HalfCosine"         // high continuity cosine variation from quadratic
   | "AustralianRailCorp" // cubic with high accuracy distance series
-  | "WesternAustralian"  // simple cubic -- 2 terms of x series, 1 term of y series
   | "Czech"              // simple cubic with two term distance approximation
+  | "Italian"
```
dassaf4: Can we say anything about Italian v. Czech? They look really similar.

dassaf4 (approved):

core/geometry/src/curve/spiral/NormalizedTransition.ts
```diff
@@ -16,10 +16,13 @@
  * * Each implementation provides:
  *   * fractionToCurvature -- the f(u) function described above
  *   * fractionToCurvatureDerivative -- df(u)/du
- *   * fractionToArea -- integral of the area under f(u) from 0 to u.
+ *   * fractionToArea -- anti-derivative A of f, chosen such that A(0) = 0. The value of this function at fraction u
+ *     is the area under f from 0 to u. In particular, the symmetry condition on f implies that 1/2 = integral of f
+ *     over [0,1] = A(1) - A(0) = A(1).
  * * The symmetry condition ensures that the integral from 0 to 1 is 1/2.
```
dassaf4: This line is now redundant.
```suggestion
```

core/geometry/src/curve/spiral/NormalizedTransition.ts
```diff
@@ -16,10 +16,13 @@
  * * Each implementation provides:
  *   * fractionToCurvature -- the f(u) function described above
  *   * fractionToCurvatureDerivative -- df(u)/du
- *   * fractionToArea -- integral of the area under f(u) from 0 to u.
+ *   * fractionToArea -- anti-derivative A of f, chosen such that A(0) = 0. The value of this function at fraction u
```
dassaf4: The word "antiderivative" is not hyphenated. If cspell complains, add it to .vscode/cSpell.json.  Also fix 2nd instance below.

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,132 @@
+# Spirals
+
+## General Math
+
+A spiral is a curve that winds around a point (usually the origin), moving progressively closer or farther away from it.
```
dassaf4, line 3: Please cover only clothoid spiral math. Core-geometry does not support Archimedean or logarithmic spirals, and this info does not facilitate understanding the API.

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,132 @@
...
+
+$$\boxed{
+\begin{aligned}
+  x(s)= L \int_{0}^{s} \Bigl(\int_{0}^{Lu} 𝜅(v)dv\Bigl) du \\ y(s)= L \int_{0}^{s} \Bigl(\int_{0}^{Lu} 𝜅(v)dv\Bigl) du
+\end{aligned}
+}$$
+
+**Example:**
+
+*Euler Spiral (Clothoid)*
```
dassaf4: Use the term "clothoid" primarily, as this is how we refer to this curve in core-geometry.
```suggestion
*Clothoid (Euler) Spiral*
```

core/geometry/internaldocs/Spiral.md
saeeedtorabi: Although our code does not support Archimedean or logarithmic spirals, this helped me to better understand spiral math (specially when I looked at the image of 3 spirals I added to the doc) so I prefer to keep it. That's why I divided the doc into "Math" and "Code" sections. I can add a note that those 2 are not supported yet. Also this is an internal doc so I dont see any harm in keeping that.

.vscode/cSpell.json
```diff
@@ -467,4 +468,4 @@
     "hte",
     "minium"
   ]
-}
+}
```
saeeedtorabi: This empty line was removed as the result of auto-save. Added it back in the next commit.

core/geometry/internaldocs/Spiral.md
dassaf4: Just trying to keep you focused here. :) An upfront warning that you are discussing things that do not apply to core-geometry would be helpful for future devs that are skimming the document. Core-geometry won't be supporting these spiral types, as they aren't relevant to our users.

core/geometry/src/curve/spiral/NormalizedTransition.ts
```diff
@@ -16,10 +16,9 @@
  * * Each implementation provides:
  *   * fractionToCurvature -- the f(u) function described above
  *   * fractionToCurvatureDerivative -- df(u)/du
- *   * fractionToArea -- anti-derivative A of f, chosen such that A(0) = 0. The value of this function at fraction u
- *     is the area under f from 0 to u. In particular, the symmetry condition on f implies that 1/2 = integral of f
- *     over [0,1] = A(1) - A(0) = A(1).
- * * The symmetry condition ensures that the integral from 0 to 1 is 1/2.
+ *   * fractionToArea -- antiderivative A of f, chosen such that A(0) = 0. The value of this function at fraction u
+ *     is the area under f from 0 to u. In particular, the symmetry condition on f implies that
+ *     "integral of f over [0,1]" = A(1) - A(0) = A(1) = 1/2.
```
dassaf4: The order is important. We are _proving_ that A(1) = 1/2.

```suggestion
 *     1/2 = "integral of f over [0,1]" = A(1) - A(0) = A(1).
```

core/geometry/internaldocs/Spiral.md
saeeedtorabi: I see your point but for any math topic in iTwin, I prefer to understand the math first and then link it to the code. Those 2 spirals are more natural than the spirals we have in iTwin so I think it's a good first step to understand the spiral math. I dont find it distracting at all for future devs.

dassaf4: Pretty sure "natural" is in the eye of the civil engineer. :)  But I won't die on this hill. I would revise your warning that there are no plans to support these spirals in core-geometry.

dassaf4 (approved):

dassaf4 (approved):

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+$$r=a+bθ$$
+Parametric (Cartesian) form:
+$$x(θ)=(a+bθ)cosθ \\ y(θ)=(a+bθ)sinθ​$$
+
+*Logarithmic Spiral:*
+$$r=ae^{bθ}$$
+Parametric (Cartesian) form:
+$$x(θ)=ae^{bθ}cosθ \\ y(θ)=ae^{bθ}sinθ​$$
+
+**Note:** core-geometry does not currently support Archimedean or logarithmic spirals, and there are no plans to implement them. Their inclusion here is solely to illustrate the mathematical concepts behind spirals.
```
dassaf4, line 31: Github's rendering blows this emphasis out of proportion! Maybe use only one asterisk?

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+
+$$\frac{dθ}{d\bar{s}} = κ(\bar{s})$$
+
+where where $κ(\bar{s})$ is curvature. If $p$ is the position vector we have
+
+$$p(s) = (x'(\bar{s}),y'(\bar{s})) = (cosθ(\bar{s}), sinθ(\bar{s}))$$
+
+so integrating gives positions:
+
+$$x(\bar{s})= \int_{0}^{\bar{s}} cosθ(u)du \\ y(\bar{s})= \int_{0}^{\bar{s}} sinθ(u)du $$
```
dassaf4: Consider putting the equations for x and y on different lines here and below.  They run together when rendered (at least in github's markup viewer).

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+
+*Logarithmic Spiral:*
+$$r=ae^{bθ}$$
+Parametric (Cartesian) form:
+$$x(θ)=ae^{bθ}cosθ \\ y(θ)=ae^{bθ}sinθ​$$
+
+**Note:** core-geometry does not currently support Archimedean or logarithmic spirals, and there are no plans to implement them. Their inclusion here is solely to illustrate the mathematical concepts behind spirals.
+-----------
+
+Another way to represent a spiral is via its curvature. In general, for a curve parameterized by curve length $\bar{s}$ we have:
```
dassaf4: The bar is hard to see and also distracting. I would use `s` for arclength (the convention in math literature) and `f` for fractional arclength (the convention in core-geometry for a fractional parameter).

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+
+$$\frac{dθ}{d\bar{s}} = κ(\bar{s})$$
```
dassaf4: This is not the same theta as the polar coordinate! What is missing is the definition of theta. Here, theta is a function of arclength: it is the angle that the tangent vector of the curve at arclength `s` makes with the x-axis. That the derivative of theta is just the curvature function is a consequence of the Frenet-Serre formulas, which apply to all 3D curves with nonzero curvature.

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+
+where where $κ(\bar{s})$ is curvature. If $p$ is the position vector we have
+
+$$p(s) = (x'(\bar{s}),y'(\bar{s})) = (cosθ(\bar{s}), sinθ(\bar{s}))$$
```
dassaf4: You are not defining the position vector here. This is the definition of the unit tangent vector. Again, this is true for any curve covered by Frenet-Serre. This is why when you integrate below, you get position.

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+
+so integrating gives positions:
+
+$$x(\bar{s})= \int_{0}^{\bar{s}} cosθ(u)du \\ y(\bar{s})= \int_{0}^{\bar{s}} sinθ(u)du $$
+
+For a spiral defined by curvature varying smoothly with length, we have fraction $s = \frac{\bar{s}}{L}$ or $\bar{s} = Ls$ where $L$ is the total spiral length. This gives:
```
dassaf4: Note that the full clothoid (of which your picture only shows a portion of the half in the 1st quadrant) has infinite length. L is referring to the length of some piece of the spiral. Your statement above has nothing to do with spirals. It is how to define a change of variables from s in [a,b] to f in [0,1] where b-a=L, namely f(s) = s/L, so s(f) = fL and ds/df = L. This is the u-substitution that explains the L factor after the substitution.

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+
+$$θ(s)= \int_{0}^{Ls} κ(u)du$$
+
+Replacing this into previous formula and switching variable to fraction form gives:
+
+$$\boxed{
+\begin{aligned}
+  x(s)= L \int_{0}^{s} \Bigl(\int_{0}^{Lu} 𝜅(v)dv\Bigl) du \\ y(s)= L \int_{0}^{s} \Bigl(\int_{0}^{Lu} 𝜅(v)dv\Bigl) du
```
dassaf4: Forgot the trig functions (and in duplicate integrals below).

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+**TransitionSpiral3d**
+
+Parent class for all iTwin spirals is `TransitionSpiral3d` which is extended by `IntegratedSpiral3d` and `DirectSpiral3d`.
+
+`TransitionSpiral3d` class has an member called `TransitionConditionalProperties` which encapsulates 5 elements: `radius0`, `radius1`, `bearing0`, `bearing1`, and `curveLength`.
+
+We already know:
+$$θ(s)= \int_{0}^{Ls} 𝜅(u)du$$
+This can be rewritten as
+$$θ_1 - θ_0 = L \bar{𝜅} = L \frac{𝜅_0 + 𝜅_1}{2}$$
```
dassaf4, line 140: This only applies to clothoids, not all TransitionSpiral3ds (albeit most are clothoids). Consider the spiral fragment over arclength interval [a,b]. It has length L = b-a. From calculus, we know that 1/(b-a) * integral of f over [a,b] = the average value of f over [a,b] (for suitably smooth f). So for the clothoid, we have 1/L * integral of curvature over [a,b] = average value of curvature over [a,b] = (curvature(a) + curvature(b))/2. The last equality follows from the fact that the curvature function for a clothoid is linear.

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+    + fraction * arcLength * this._curvature01.x0 // pink area
+    + areaFraction * arcLength * this._curvature01.signedDelta(); // green area
+}
+```
+
+**DirectSpiral3d**
+
+Direct spirals do not use integration to calculate $x$ and $y$ coordinates. Instead they use direct formulas to find coordinates.
+
+The 4 spirals `JapaneseCubic`, `Arema`, `ChineseCubic`, and `WesternAustralian` have a formula for $θ$ (which is $θ(s) = c*s^2$).
```
dassaf4: These formulae use truncated Taylor series to approximate the clothoid integrands by polynomials, which we can integrate directly (without quadrature).

core/geometry/internaldocs/Spiral.md
```diff
@@ -0,0 +1,133 @@
...
+\begin{aligned}
+  x(s)= L \int_{0}^{s} \Bigl(\int_{0}^{Lu} 𝜅(v)dv\Bigl) du \\ y(s)= L \int_{0}^{s} \Bigl(\int_{0}^{Lu} 𝜅(v)dv\Bigl) du
+\end{aligned}
+}$$
+
+**Example:**
+
+*Clothoid (Euler Spiral)*
+
+Curvature $𝜅$ increases linearly with arc length $s$:
```
dassaf4: This is what makes a curve a clothoid: it has linear curvature (parameterized by arclength).  I think some of the conclusions come before the suppositions in this document. Something's not quite jibing. In particular, some of this stuff is true for any sufficiently smooth curve, not just clothoids. I'll write something up in an attempt to make it flow better (to me) and share next week.

dassaf4 (approved):
I reviewed Spiral.md today.

core/geometry/internaldocs/Spiral.md
saeeedtorabi: Indeed average function is not just average of start and end values. However, `TransitionSpiral3d` has a member `_designProperties` of type `TransitionConditionalProperties` and if you look at the doc for that class, you see this equation. This class does not mention clothoids. 

For example, look at `IntegratedSpiral3d.createFrom4OutOf5` which instantiate `TransitionConditionalProperties` and takes any `spiralType`.

dassaf4: Ah. I think I understand now. TransitionConditionalProperties is not instantiated for DirectSpiral3ds, but only for IntegratedSpiral3ds, which is a class that generalizes the clothoid. An IntegratedSpiral3d allows the curve's curvature function to be nonlinear while still satisfying the same avg value formula enjoyed by the clothoid with its linear curvature. The NormalizedTransition class defines the 3 constraints on the normalized form of a curvature function to guarantee this avg value formula holds. So my original comment still stands: the average value formula does not hold for all TransitionSpiral3ds.

core/geometry/internaldocs/Spiral.md
dassaf4: What do you use to see the rendered markup? I see this in github's rich diff view: 

image

saeeedtorabi: I render locally in VS Code (fixed the issue with Note paragraph being too lage)

image

core/geometry/internaldocs/Spiral.md
saeeedtorabi: ok I will modify the md doc to reflect this.

core/geometry/internaldocs/Spiral.md
dassaf4: Ok, guess that's what internal devs would be using to view this. Good.

core/geometry/internaldocs/Spiral.md
dassaf4: This ordering is what made sense to me. Feel free to use! (Let me know if you find any errors!) Left page has no spiral-specific definitions. Right page is about how left page applies to clothoid and variants.
PXL_20251104_151425206

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1155,11 +1163,17 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
+   * @param curvePrimitive The other curve primitive.
+   * @param numberOfNewResults The number of results to be refined.
+   * @param reversed Whether the spiral is geometryA (false) or geometryB (true).
    */
-  private refineSpiralResultsByNewton(spiral: TransitionSpiral3d, curvePrimitive: CurvePrimitive, reversed: boolean): void {
-    const oldResults = this._results;
-    this._results = []; // erase old results to be replaced by refined results
-    for (const detail of oldResults) {
+  private refineSpiralResultsByNewton(
+    spiral: TransitionSpiral3d, curvePrimitive: CurvePrimitive, numberOfNewResults: number, reversed = true,
```
dassaf4: It's not typical for a default value to be nonzero.  Either make it default to false, or make it non-optional, like elsewhere in the file.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1155,11 +1163,17 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
   /**
    * Use the intersection found between the spiral approximation and the curve primitive as the initial guess for
    * the Newton iteration to locate the intersection between the actual spiral and the curve primitive.
+   * @param spiral The transition spiral.
+   * @param curvePrimitive The other curve primitive.
+   * @param numberOfNewResults The number of results to be refined.
```
dassaf4:
```suggestion
   * @param numberOfNewResults The number of results in the tail of `this._results` to be refined.
```

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1155,11 +1163,17 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
    */
-  private refineSpiralResultsByNewton(spiral: TransitionSpiral3d, curvePrimitive: CurvePrimitive, reversed: boolean): void {
-    const oldResults = this._results;
-    this._results = []; // erase old results to be replaced by refined results
-    for (const detail of oldResults) {
+  private refineSpiralResultsByNewton(
+    spiral: TransitionSpiral3d, curvePrimitive: CurvePrimitive, numberOfNewResults: number, reversed = true,
+  ): void {
+    const resultsToBeRefined = this._results.slice(this._results.length - numberOfNewResults);
+    this._results = this._results.slice(0, this._results.length - numberOfNewResults); // keep already refined results
```
dassaf4: Don't replace the array, just trim its length.
```suggestion
    this._results.length -= numberOfNewResults; // keep already refined results
```

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1173,19 +1187,28 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
         spiralFraction = newtonSearcher.getU();
         geometryBFraction = newtonSearcher.getV();
       }
-      if (this.acceptFraction(false, spiralFraction, false) && this.acceptFraction(false, geometryBFraction, false))
-        this.recordPointWithLocalFractions(spiralFraction, spiral, 0, 1, geometryBFraction, curvePrimitive, 0, 1, false);
+      if (this.acceptFraction(false, spiralFraction, false) && this.acceptFraction(false, geometryBFraction, false)) {
+        if (reversed)
+          this.recordPointWithLocalFractions(geometryBFraction, curvePrimitive, 0, 1, spiralFraction, spiral, 0, 1, false);
+        else
+          this.recordPointWithLocalFractions(spiralFraction, spiral, 0, 1, geometryBFraction, curvePrimitive, 0, 1, false);
```
dassaf4: Can you take advantage of the reversed argument?
```suggestion
        this.recordPointWithLocalFractions(spiralFraction, spiral, 0, 1, geometryBFraction, curvePrimitive, 0, 1, reversed);
```

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1173,19 +1187,28 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
       spiral.emitStrokes(spiralApproximation);
+      const numPreviousResults = this._results.length;
       this.handleLineString3d(spiralApproximation);
-      this.refineSpiralResultsByNewton(spiral, this._geometryB, false);
+      const numberOfNewResults = this._results.length - numPreviousResults;
+      this.refineSpiralResultsByNewton(spiral, this._geometryB, numberOfNewResults, false);
+    } else if (this._geometryB instanceof CurveChainWithDistanceIndex) {
+      this.dispatchCurveChainWithDistanceIndex(spiral, this.handleBSplineCurve3d.bind(this));
+    } else if (this._geometryB instanceof CurveCollection) {
+      this.dispatchCurveCollection(spiral, this.handleBSplineCurve3d.bind(this));
```
dassaf4: Why are we calling handleBSplineCurve3d here?  Can you add a test that would have caught this?

dassaf4 (approved):

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
saeeedtorabi: copy/paste mistake. I have not tested curve collections yet. That's first in my todo list!

saeeedtorabi: but it should have failed for `CurveChainWithDistanceIndex`. I will investigate

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1165,20 +1165,20 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
+    spiral: TransitionSpiral3d, curvePrimitive: CurvePrimitive, numberOfNewResults: number, reversed = false,
   ): void {
     const resultsToBeRefined = this._results.slice(this._results.length - numberOfNewResults);
-    this._results = this._results.slice(0, this._results.length - numberOfNewResults); // keep already refined results
+    this._results.length -= numberOfNewResults; // keep already refined results
     for (const detail of resultsToBeRefined) {
-      let spiralFraction = detail.detailA.fraction;
-      let geometryBFraction = detail.detailB.fraction;
+      let geometryBFraction = detail.detailA.fraction;
+      let spiralFraction = detail.detailB.fraction;
```
dassaf4: Doc says spiral data is in `detailB` when `reversed` is true. But the code implies the opposite. I think the doc is fine (it aligns with the usage of `reversed` elsewhere in this file, which is based on the order of the method arguments: first is A and second is B iff reversed is false), and the code should be swapped.

Also, the identifier `geometryBFraction` is confusing. I'd change `curvePrimitive` to `otherCurve` and `geometryBFraction` to `otherFraction`.

Doing the assignment all on one line is simpler. Here I've corrected the code to align with the doc:

```suggestion
      const spiralFraction = reversed ? detail.detailB.fraction : detail.detailA.fraction;
      const otherFraction = reversed ? detail.detailA.fraction : detail.detailB.fraction;
```

dassaf4 (approved):

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
saeeedtorabi: It turns out that the code fine and I have to fix the doc. I will also reverse the arg order to align with other code is this file.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
saeeedtorabi: This can be resolved

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1166,29 +1166,25 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
    * @param spiral The transition spiral.
    * @param curvePrimitive The other curve primitive.
    * @param numberOfNewResults The number of results in the tail of `this._results` to be refined.
-   * @param reversed Whether the spiral is geometryA (false) or geometryB (true).
+   * @param reversed Whether the spiral is geometryA (true) or geometryB (false).
    */
   private refineSpiralResultsByNewton(
-    spiral: TransitionSpiral3d, curvePrimitive: CurvePrimitive, numberOfNewResults: number, reversed = false,
+    curvePrimitive: CurvePrimitive, spiral: TransitionSpiral3d, numberOfNewResults: number, reversed = false,
```
dassaf4: To match `otherFraction`:

```suggestion
    otherCurve: CurvePrimitive, spiral: TransitionSpiral3d, numberOfNewResults: number, reversed = false,
```

dassaf4 (approved):

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
dassaf4: What happens when you try to extend the spiral? Is it too much to hope this "just works"? :smile:

saeeedtorabi: 😄Sorry! not working. We can extend line string that initially approximates the spiral but that's not same extension as spiral so for now we are stuck with no extension for spirals.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1112,20 +1138,69 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
+  ): void {
+    const resultsToBeRefined = this._results.slice(this._results.length - numberOfNewResults);
+    this._results.length -= numberOfNewResults; // keep already refined results
+    for (const detail of resultsToBeRefined) {
+      let spiralFraction = reversed ? detail.detailA.fraction : detail.detailB.fraction;
+      let otherFraction = reversed ? detail.detailB.fraction : detail.detailA.fraction;
+      const xyMatchingFunction = new CurveCurveIntersectionXYRRToRRD(spiral, otherCurve);
+      const newtonSearcher = new Newton2dUnboundedWithDerivative(xyMatchingFunction);
+      newtonSearcher.setUV(spiralFraction, otherFraction);
+      if (newtonSearcher.runIterations()) {
```
dassaf4, line 1183: If Newton doesn't converge then I think we should not accept the seed.

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
```diff
@@ -2356,11 +2361,221 @@ describe("CurveCurveIntersectXY", () => {
...
+          curve1Detail = intersection.detailA;
+        }
+        if (!(curve0 instanceof CurveCollection)) {
+          const pointOnCurve0 = curve0Detail.point;
+          const intersectionPoint0 = curve0.fractionToPoint(curve0Detail.fraction);
+          ck.testPoint3dXY(
+            pointOnCurve0,
+            intersectionPoint0,
+            `intersection point on ${curveName0} should match ${curveName0}.fractionToPoint`,
+          );
```
dassaf4: This test confirms that for a given detail, the stored fraction yields the same xy-coords as the stored point. We certainly expect that a CurveLocationDetail should be self-consistent---that its fraction should yield the same xyz point (z included) on the curve.

However IIUC, this test does _not_ confirm that each computed input pair represents an _intersection_, which should be the primary goal of this test. To confirm this, you need to test that the point in detailA is sufficiently near the point in detailB in xy-coordinates only.

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
```diff
@@ -2356,11 +2361,221 @@ describe("CurveCurveIntersectXY", () => {
...
+  it("SpiralIntersection", () => {
+    const ck = new Checker();
+    const allGeometry: GeometryQuery[] = [];
+    let dx = 0;
+    let dy = 0;
+
+    const rotationTransform = Transform.createFixedPointAndMatrix(
+      Point3d.create(70, 0),
+      Matrix3d.createRotationAroundVector(Vector3d.create(0, 0, 1), Angle.createDegrees(90))!,
+    );
```
dassaf4: Can you also test intersections between curves that are _not_ coplanar?  The true superpower of `CurveCurve.intersectionXYPairs` is that you can throw any two fully 3D curves into it, and it will return their virtual intersections as seen from the top view (non-camera). This is useful e.g., for snapping to space curves at virtual intersections in any view.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1112,20 +1138,69 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
     } else if (this._geometryB instanceof CurveCollection) {
       this.dispatchCurveCollection(curve, this.handleBSplineCurve3d.bind(this));
     } else if (this._geometryB instanceof CurveChainWithDistanceIndex) {
       this.dispatchCurveChainWithDistanceIndex(curve, this.handleBSplineCurve3d.bind(this));
     }
     return undefined;
   }
+  /**
+   * Use the intersection found between the spiral approximation and the curve primitive as the initial guess for
+   * the Newton iteration to locate the intersection between the actual spiral and the curve primitive.
```
dassaf4:
```suggestion
   * Assuming the tail of `this._results` holds the given number of approximate solutions to the curve-spiral xy-intersection problem, replace each with its Newton refinement, unless it doesn't converge, in which case it is removed.
```

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
dassaf4: Ok. This makes sense, I suppose, as the length of each spiral CurvePrimitive is baked into its definition.

core/geometry/internaldocs/Spiral.md
dassaf4: The generality of this curvature representation is so beautiful it warms my heart. Makes up for the unseasonably frigid weather outside!

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
saeeedtorabi: I think if the Newton does not converge then we need to enter the intersection found for the line string and that is what's happening at the current code.

dassaf4: I disagree; those seeds that fail to refine may be too coarse to return as intersections. Are there any seeds that fail to converge in your tests? If 15 iterations are not enough, then you can instantiate the relevant Newton object with more.

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
```diff
@@ -2356,11 +2361,258 @@ describe("CurveCurveIntersectXY", () => {
...
+        const oneChainOneCollection = (curve0 instanceof CurveChainWithDistanceIndex && curve1 instanceof CurveCollection)
+          || (curve1 instanceof CurveChainWithDistanceIndex && curve0 instanceof CurveCollection);
+        if (reversed && (oneCurveIsSpiral || bothChain || oneChainOneCollection)) {
+          curve0Detail = intersection.detailB;
+          curve1Detail = intersection.detailA;
+        }
+        let pointOnCurve0 = Point3d.create();
+        let pointOnCurve1 = Point3d.create();
+        if (!(curve0 instanceof CurveCollection)) {
+          pointOnCurve0 = curve0Detail.point;
```
dassaf4: You already have allocated a point, so you can just set it.  Similarly below for `pointOnCurve1`.
```suggestion
          pointOnCurve0.setFrom(curve0Detail.point);
```

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
saeeedtorabi: All my tests converge.

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
saeeedtorabi: done

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
saeeedtorabi: done

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
saeeedtorabi: done

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
```diff
@@ -2434,13 +2443,20 @@ describe("CurveCurveIntersectXY", () => {
       Point3d.create(70, 0),
       Matrix3d.createRotationAroundVector(Vector3d.create(0, 0, 1), Angle.createDegrees(90))!,
     );
+    const moveTransform = Transform.createTranslationXYZ(0, 0, 10);
+    const nonPlanarTransform = Transform.createZero();
+    nonPlanarTransform.setMultiplyTransformTransform(rotationTransform, moveTransform);
```
dassaf4: Ok, but the new transform just changes the z compared to the other transform.  Z-coordinate is ignored by the xy-intersector. Can you test with a curve that isn't horizontal, or isn't planar?

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
dassaf4: Tangent intersections can take longer to converge. Maybe try one of those?

In fact, what should we do for a tangent intersection where the stroked approximation doesn't intersect the other curve? We won't get a seed in this case.

At any rate, a seed that doesn't converge should not be returned.

saeeedtorabi: Okay I will test a tangent case

So you think if it's not converging we should not return the inaccurate 
intersection and just drop it?

dassaf4: Yes. Assert on lack of convergence if you want. This will inform us to try more max iters.

dassaf4: > In fact, what should we do for a tangent intersection where the stroked approximation doesn't intersect the other curve? We won't get a seed in this case.

One idea: try seeding with linestring xy close approach instead of linestring xy intersect. (Maybe throw out close approaches above a certain toleranced distance...)

dassaf4: (Postponing this refactor to another PR.)

PR #8661, opened 2025-10-22
Updated tests for newer biscore schema release

naveedkhan8067:
Biscore schema version 01.00.24 is about to be released. The tests have adjusted to work with the newer version.

### DIFF
--- a/common/changes/@itwin/core-backend/2025-10-23-15-03.json
+++ b/common/changes/@itwin/core-backend/2025-10-23-15-03.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -22,8 +22,8 @@ importers:
         specifier: ^12.28.0
         version: 12.28.0
       '@bentley/imodeljs-native':
-        specifier: 5.3.20
-        version: 5.3.20
+        specifier: 5.3.23
+        version: 5.3.23
       '@itwin/object-storage-azure':
         specifier: ^3.0.4
         version: 3.0.4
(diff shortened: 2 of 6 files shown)


core/backend/src/test/ecsql/queries/AllOrAnyExp.ecsql.md
```diff
@@ -908,15 +908,15 @@ WHERE
 LIMIT 3
```
rschili: Bah, yeah these two tests are nasty. we should move them to something that is schema independent. However, there is no 2 Minute drop in replacement for this that I could think of. We probably need to craft a test class with some elements and change the query to use it. Alternatively we could switch to statically typed values, something like:

```sql
SELECT
  *
FROM
  (VALUES (1), (2), (3)) AS myId(ECInstanceID)
WHERE
  ECInstanceID < ALL (
    SELECT
      ECClassId
    FROM
      (VALUES (3, 5), (2, 4)) AS t(ECClassId, RelECClassId)
  );
```

Expected rows: 1 with value 1
This is plain sqlite though, not sure if it's fully supported in ECSql.

core/backend/src/test/ecsql/queries/DerivedProperty.ecsql.md
```diff
@@ -37,11 +37,11 @@ LIMIT
 
 | ECInstanceId | Model                                  | dt                      | b    | bin                                | DirectStr | DirectLong | DirectDouble | p2d                         | p3d                                      |
 | ------------ | -------------------------------------- | ----------------------- | ---- | ---------------------------------- | --------- | ---------- | ------------ | --------------------------- | ---------------------------------------- |
-| 0x14         | {"Id": "0x11", "RelECClassId": "0x61"} | 2017-01-01T00:00:00.000 | true | BIN(1,2,3)                         | str0      | 1000       | 0.1          | {"X": 1.034,"Y": 2.034}     | {"X": -1,"Y": 2.3,"Z": 3.0001}           |
```
rschili, line 40: Hm this is a navigation property, I don't think there is a nice way to make this schema agnostic. I suggest we remove the model column from this test

core/backend/src/test/standalone/ChangesetReader.test.ts
```diff
@@ -1240,16 +1240,16 @@ describe("Changeset Reader API", async () => {
     const [secondBriefcaseChangeset1, secondBriefcaseChangeset2] = secondBriefcaseChangesets;
 
     expect(secondBriefcaseChangeset1.uncompressedSizeBytes).to.be.greaterThan(40000);
-    expect(secondBriefcaseChangeset1.insertedRows).to.be.greaterThanOrEqual(52);
+    expect(secondBriefcaseChangeset1.insertedRows).to.be.eql(0);
```
rschili, line 1239: This looks odd, why did the two changesets swap their contents?

naveedkhan8067:
https://github.com/iTwin/bis-schemas/pull/604

core/backend/src/test/ecsql/queries/DerivedProperty.ecsql.md
naveedkhan8067: I have removed it

core/backend/src/test/ecsql/queries/AllOrAnyExp.ecsql.md
naveedkhan8067: I think we can just check the result count with an outer query rather than matching the names everytime

core/backend/src/test/standalone/ChangesetReader.test.ts
naveedkhan8067: It is one of the newer tests that was added several week back and I remember it was problematic with the previous release as well. I have create a item for @RohitPtnkr1996 : https://github.com/iTwin/itwinjs-core/issues/8665

PR #8673, opened 2025-10-23
Add error handling and add method withIndirectTxnModeAsync()

khanaffan:
Introduce error handling for indirect changes in `BriefcaseDb` and `IModelDb`, and implement `withIndirectTxnModeAsync` in `TxnManager` for editing apps.

### DIFF
--- a/common/api/core-common.api.md
+++ b/common/api/core-common.api.md
@@ -4435,6 +4435,21 @@ export interface GltfChunk {
 }
 
 // @internal (undocumented)
+export class GltfHeader extends TileHeader {
+    constructor(stream: ByteStream);
+    // (undocumented)
+    readonly binaryPosition: number;
+    // (undocumented)
+    readonly gltfLength: number;
+    // (undocumented)
+    get isValid(): boolean;
+    // (undocumented)
+    readonly scenePosition: number;
+    // (undocumented)
+    readonly sceneStrLength: number;
+}
+
+// @internal (undocumented)
 export enum GltfV2ChunkTypes {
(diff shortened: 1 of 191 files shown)


khanaffan:
These are **alpha APIs** designed to support editing workflows. I suggest we allow the team to start using them and provide feedback. We'll continue to evolve the APIs based on that input until we better understand what’s truly needed and what isn’t.

The **error cases** only occur when using the new APIs. Developers who haven’t adopted them won’t encounter these issues. With the **Async API**, we need to ensure developers don’t unintentionally introduce logic bugs that could later result in runtime errors for other users. I’ve added safeguards to highlight when the API is being misused.

Most of these errors are based on **common-sense constraints**. For example:
- You shouldn’t import a schema in an indirect or recompute scope.
- Avoid performing pull merges inside callbacks for indirect or recompute scopes.
- Developers shouldn’t store file properties in an indirect scope, as they’ll be marked indirect and that affects how they’re merged.

In above error cases they need to fix their code. 

I have add note on docs of each function which tell developer not to call these from specified function callbacks. We can add detail docs with example once we get more feedback from OpenSite+ or when they are close to release with these api and we have to make them public.

Generally speaking these new API cover following, I can create a issue for that documenting following 

1. Marking changes indirect (one that can be recomputed and does not need to be locked)
2. Allow saving extended information in `saveChanges()` which can be access during rebase in `recompute()` handler
3. Allow async callback during rebase to allow app to recompute indirect changes or propagate changes as needed.

ColinKerr (approved):

Issue #8441, opened 2025-08-20
Presentation: Need ability to easily re-format `Content`


grigasp:
At the moment `PresentationManager` uses internal `ContentFormatter` to apply formatting to produced `Content`. However, consumers need ability to re-format the content when formatting settings change, ideally without the need to request it again from `PresentationManager` (that causes an unnecessary trip to the backend).

PR #8683, opened 2025-10-27
Presentation: Added ability to (re-)format `Content`

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/8441

### DIFF
--- a/common/changes/@itwin/certa/bump-playwright-to-1.56.1_2025-10-27-09-53.json
+++ b/common/changes/@itwin/certa/bump-playwright-to-1.56.1_2025-10-27-09-53.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/certa",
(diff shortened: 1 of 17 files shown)


common/api/presentation-common.api.md
```diff
@@ -368,6 +368,9 @@ export enum ContentSpecificationTypes {
 // @public
 export type ContentUpdateInfo = typeof UPDATE_FULL;
 
+// @public
```
aruniverse, line 371: should new apis be public from the get go or should it be public preview or beta?

grigasp: I'm pretty confident we won't need to change it, because:
- it's pretty thin in what it takes and what it does,
- the underlying functionality has been used already for a while.

If you think it's better to use one of the mentioned release tags, I can change it, but I think `@public` is fine for this one, too.

Issue #8532, opened 2025-09-16
Use cesium-renderer API in integration prototype to modify Cesium camera


eringram:
Use new cesium-renderer package to modify the camera in the integration prototype.

markschlosseratbentley:
@eringram Can you make issue(s) with any remaining work once the initial PR for this merges in?

eringram:
Draft PR: https://github.com/iTwin/itwinjs-core/pull/8697

PR #8697, opened 2025-10-28
Add camera controls to CesiumJS RenderSystem prototype

eringram:
Fixes https://github.com/iTwin/itwinjs-core/issues/8532

This PR improves the CesiumJS RenderSystem prototype:
- Adds the ability to control the CesiumJS camera via iTwin.js default view tool
- To enable this, document required tool code changes to use pointer events instead of mouse events
- Fixes decoration converter to not create some unintentional decorations
- Package docs on how to run the example

It can be tested by running DTA with the environment variable `IMJS_USE_CESIUM=1` (instructions in the previous PR https://github.com/iTwin/itwinjs-core/pull/8544) and using the default zoom, pan, and rotate tools.

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -4016,6 +4016,8 @@ export class IModelHostConfiguration implements IModelHostOptions {
     disableThinnedNativeInstanceWorkflow?: boolean;
     // (undocumented)
     hubAccess?: BackendHubAccess;
+    // @beta
+    incrementalSchemaLoading: "enabled" | "disabled";
     // @internal (undocumented)
     logTileLoadTimeThreshold: number;
     // @internal (undocumented)
@@ -4046,6 +4048,8 @@ export interface IModelHostOptions {
     disableThinnedNativeInstanceWorkflow?: boolean;
     enableOpenTelemetry?: boolean;
     hubAccess?: BackendHubAccess;
+    // @beta
+    incrementalSchemaLoading?: "enabled" | "disabled";
     // @internal
     logTileLoadTimeThreshold?: number;
     // @internal
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -4943,6 +4943,8 @@ export class IModelApp {
     static get initialized(): boolean;
     // @internal
     static get isEventLoopStarted(): boolean;
+    // @beta
+    static get isIncrementalSchemaLoadingEnabled(): boolean;
     static get localization(): Localization;
     // (undocumented)
     static get locateManager(): ElementLocateManager;
@@ -5014,6 +5016,8 @@ export interface IModelAppOptions {
     authorizationClient?: AuthorizationClient;
     formatsProvider?: FormatsProvider;
     hubAccess?: FrontendHubAccess;
+    // @beta
+    incrementalSchemaLoading?: "enabled" | "disabled";
     localization?: Localization;
     // @internal (undocumented)
     locateManager?: ElementLocateManager;
--- a/common/changes/@itwin/core-backend/add-IncrementalSchemaLoading-to-imodel-schemacontexts_2025-10-02-13-03.json
+++ b/common/changes/@itwin/core-backend/add-IncrementalSchemaLoading-to-imodel-schemacontexts_2025-10-02-13-03.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/changes/@itwin/core-backend/add-IncrementalSchemaLoading-to-imodel-schemacontexts_2025-11-20-14-58.json
+++ b/common/changes/@itwin/core-backend/add-IncrementalSchemaLoading-to-imodel-schemacontexts_2025-11-20-14-58.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Added incrementalSchemaLoading setting on IModelHostOptions and IModelHostConfiguration to enable or disable incremental schema loading. For now, incremental schema loading is disabled by default.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/changes/@itwin/core-frontend/add-IncrementalSchemaLoading-to-imodel-schemacontexts_2025-10-02-13-03.json
+++ b/common/changes/@itwin/core-frontend/add-IncrementalSchemaLoading-to-imodel-schemacontexts_2025-10-02-13-03.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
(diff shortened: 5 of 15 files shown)


eringram:
@bbastings We are seeking feedback on the changes to `EventController` that allow iTwin.js to use pointer events instead of mouse events, which is necessary to use iTwin tools with the new Cesium renderer. It seems like pointer is supported in all browsers but want to make sure there will be no side effects you are aware of.

extensions/cesium-renderer/src/decorations/ShapePrimitiveConverter.ts
```diff
@@ -179,7 +180,7 @@ export class ShapePrimitiveConverter extends PrimitiveConverter<ShapeCoordinates
           })
         });
 
-        primitivesCollection.add(primitive);
+        (primitive as any).id = shapeId;
```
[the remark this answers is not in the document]
eringram: @danielzhong Could you take a look at this if you get the chance? Or if the assertion is necessary, feel free to resolve

extensions/cesium-renderer/src/decorations/PrimitiveConverter.ts
```diff
@@ -190,37 +196,56 @@ export abstract class PrimitiveConverter<TPrimitiveData = DecorationPrimitiveEnt
...
+    const matchesType = (id: any) => typeof id === "string" && id.startsWith(`${type}_`);
+
+    const removedPoints = this.removeMatching(scene.pointCollection, matchesType);
+    const removedLines = this.removeMatching(scene.polylineCollection, matchesType);
+    const removedPrimitives = this.removeMatching(scene.primitivesCollection, matchesType);
+
+    if (removedPoints || removedLines || removedPrimitives) {
+      // eslint-disable-next-line no-console
+      console.debug("[PrimitiveConverter] clearDecorationsForType", { type, removedPoints, removedLines, removedPrimitives });
     }
```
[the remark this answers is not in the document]
eringram: @danielzhong Would you like to keep these in while the prototype is WIP or can I remove?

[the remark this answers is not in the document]
danielzhong: Please just remove

extensions/cesium-renderer/src/decorations/PrimitiveConverter.ts
```diff
@@ -190,37 +196,56 @@ export abstract class PrimitiveConverter<TPrimitiveData = DecorationPrimitiveEnt
...
+    const removedLines = scene.polylineCollection?.length ?? 0;
+    scene.polylineCollection?.removeAll();
+
+    const removedPrimitives = scene.primitivesCollection?.length ?? 0;
+    scene.primitivesCollection?.removeAll();
+
+    if (removedPoints || removedLines || removedPrimitives) {
+      // eslint-disable-next-line no-console
+      console.debug("[PrimitiveConverter] clearAllDecorations", { removedPoints, removedLines, removedPrimitives });
     }
```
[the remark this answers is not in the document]
danielzhong: Could remove as well

core/frontend/src/tools/EventController.ts
```diff
@@ -23,8 +23,16 @@ export class EventController {
     if (element === undefined)
       return;
 
-    // Put events  on the parentDiv to allows us to stopPropagation of events to the view canvas when they are meant for a sibling of view canvas (markup canvas, for example).
-    this.addDomListeners(["mousedown", "mouseup", "mousemove", "mouseover", "mouseout", "wheel", "touchstart", "touchend", "touchcancel", "touchmove"], element);
+    const pointerSupported = window.PointerEvent !== undefined;
+
+    // Put events on the parentDiv to allow us to stopPropagation of events to the view canvas when they are meant for a sibling of view canvas (markup canvas, for example).
+    if (pointerSupported) {
```
bbastings: Did you test this with touch interaction? My understanding of pointermove/up/down is that it also applies to touch input...and we don't want that automatically being treated the same as moving the mouse or clicking the mouse buttons.

If both the pointer and touch events are being called, or if the pointer events prevent the touch events it will certainly break our current tools.

We have methods like ToolAdmin.convertTouchMoveStartToButtonDownAndMotion that tools can call when/if they want to treat a touch like a normal mouse motion.

See SelectTool.ts implementation of onTouchMoveStart and onTouchMove that it uses for drag box and crossing line selection based on the current selection as opposed to letting the touch events be handled by the idleTool to pan the view.

Another tool to look at/test is the LookAndMoveTool that uses the touch events/multi-touch/changed touches for virtual control sticks, etc.

We still need to keep calling our separate BeTouchEvent and BeButtonEvent tool methods and I think this breaks that.

eringram: Thanks for the feedback. We have not tested it with touch input aside from any unit tests, but it makes sense this could break existing tools relying on touch events.

I'll remove the EventController and ToolAdmin changes and document in the package README that the pointer events are currently needed to manipulate the Cesium camera in our prototype. In the future we might revisit and try to make this work with mouse events or do more testing to support pointer events, but that is not in near term plans.

bbastings: For the future...

If it turns out both the pointer and touch events are currently being called there's a pretty simple potential solution. Just check the source of the pointer event and ignore it if it's touch to avoid having it also process the event.

Pointer events are definitely not intended as a replacement for touch events w/o having the "changed touches" information (I can't imagine they'd expect everyone to create this themselves using the pointer events). Pointer events are just for treating all input sources like a mouse...

eringram: Checking the source and ignoring touch sounds like a good solution if needed in the future.

I'm not sure why, but when we use a Cesium viewer in this prototype, mouse events don't seem to be raised at all no matter what element they are added to, but pointer events fire as expected. A tentative theory is that Cesium is blocking mouse events entirely somewhere, so a fix for that may require changes to CesiumJS. Adding pointer event support to iTwin.js was the easiest workaround.

But either way, I removed the tool changes and updated the package README to document the issue

extensions/cesium-renderer/src/decorations/PrimitiveConverter.ts
```diff
@@ -147,27 +147,33 @@ export abstract class PrimitiveConverter<TPrimitiveData = DecorationPrimitiveEnt
   }
 
 
-  /** Convert all decoration types using switch-based auto dispatch */
-  public convertAllDecorationTypes(decorations: Decorations, scene: CesiumScene, iModel?: IModelConnection): void {
-    if (decorations.world) {
-      this.autoDispatchGraphics(decorations.world, 'world', scene, iModel);
-    }
+  /** Convert decorations of a specific type, optionally clearing existing Cesium primitives for that type first. */
```
markschlosseratbentley, line 150: Are the substantial changes to a number of decoration converter classes intentional?

common/changes/@itwin/core-frontend/eringram-modify-camera_2025-10-30-19-31.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Use pointer events instead of mouse events if supported.",
```
pmconne:
```suggestion
      "comment": "",
```

You reverted those changes.

extensions/cesium-renderer/src/decorations/PrimitiveConverter.ts
eringram: Yes, those are changes made by @danielzhong to stop some decorations from being unintentionally added, like the indicators that show up due to iTwin.js view tools. For example this green circle under the cursor while using the rotate tool-- they were appearing persistently in the Cesium renderer and looked weird so I think he just removed them entirely (screenshot is iTwin renderer just to show an example)

Screenshot 2025-11-20 134206

markschlosseratbentley: Understood, thanks!

core/frontend/src/tools/EventController.ts
eringram: @bbastings Is this ok to resolve?

PR #8698, opened 2025-10-29
Added perf tests for incremental schema loading

AureMonke:


full-stack-tests/backend/src/perftest/IncrementalLoading.test.ts
```diff
@@ -0,0 +1,727 @@
+/*---------------------------------------------------------------------------------------------
```
StefanApfel-Bentley, line 1: There are a lot of exports, since these types aren't used anywhere outside this file, we should not export that.

full-stack-tests/backend/src/perftest/IncrementalLoading.test.ts
```diff
@@ -0,0 +1,727 @@
...
+  referenceSchemas?: { schemaName: string; version: string; alias?: string }[];
+  entityClasses?: ClassGenerationOptions;
+  mixins?: ClassGenerationOptions;
+  relationshipClasses?: ClassGenerationOptions;
+  structClasses?: ClassGenerationOptions;
+  customAttributeClasses?: ClassGenerationOptions;
+  schemaItemCount?: number;
+}
+
+describe("IncrementalLoadingPerformance", async () => {
```
StefanApfel-Bentley: We should not use async functions in describe because I noticed Mocha is executing them directly and if you use only for example, it still be executed. better wrap the async stuff in a before block.

StefanApfel-Bentley (approved):

ben-polinsky (approved):

PR #8699, opened 2025-10-29
Removed schema references from multiple/single queries

AureMonke:
1) Removed repetitive withSchemaReferences from queries
2) Fix for labels of some classes are different when loading the same schema using JSON and Incremental locater


core/ecschema-metadata/src/IncrementalLoading/IncrementalSchemaLocater.ts
```diff
@@ -14,9 +14,11 @@ import { SchemaKey } from "../SchemaKey";
 import { SchemaLoadingController } from "../utils/SchemaLoadingController";
 import { IncrementalSchemaReader } from "./IncrementalSchemaReader";
 
-interface IncrementalSchemaInfo extends SchemaInfo {
+export interface IncrementalSchemaInfo extends SchemaInfo {
```
StefanApfel-Bentley, line 20: I'm thinking, shouldn't that rather be internal? We don't want to use it outside of itwinjs.

StefanApfel-Bentley:
Don't forget to do an extract-api, we changed a few interfaces.

core/ecschema-metadata/src/IncrementalLoading/ClassParsers.ts
```diff
@@ -42,11 +47,39 @@ export class ClassParser extends SchemaItemParser {
...
   private parseProperties(propertyProps: AnyPropertyProps[]) {
     for (const props of propertyProps as MutablePropertyProps[]) {
+      this.decodeNameToDisplayLabel(props);
       props.customAttributes = props.customAttributes && props.customAttributes.length > 0 ? props.customAttributes.map((attr: any) => { return parseCustomAttribute(attr); }) : undefined;
       if (!props.customAttributes)
         delete props.customAttributes;
     }
   }
+  
+  private decodeNameToDisplayLabel(props: MutableClassProps | MutablePropertyProps) {
```
rschili, line 57: This can be improved. The "ecNameRegex" is already defined in ECName, it's actually a "replaced character as hex"-regex.
You could just always call ECName.decode, and it would do nothing if the regex doesn't match. No need to duplicate the regex into this method, I think the logic remains the same without the regex.

rschili: Edit: Ah now I see why this. You only want to set the label if there are any encoded chars in the name.
In that case we should add a method for it to ECName, so we don't have to copy the regex. Something like ECName.hasEncodedCharacters() : bool

PR #8700, opened 2025-10-29
Add support for Node.js 24

GytisCepk:
### Added Node.js 24 as supported version
  - No code changes were made as I couldn't find any usage of deprecated or removed APIs in our code ([Node.js 24 deprecations and removals](https://nodejs.org/en/blog/release/v24.0.0#deprecations-and-removals)).
  - Left `@types/node` version unchanged as some time ago it was decided to use types of lowest support Node.js version to prevent developers accidentally consuming something which is not available in older supported versions.

### Updated pipelines to use Node.js 24

All pipeline scripts were updated to use Node.js 24 with exception of mobile CI pipeline as Node.js 24 support for mobile will come later.

### Upgraded to latest Rush.js version
Even newest Rush.js version is still untested with Node.js 24, so expect to see following warning in console:
```
Your version of Node.js (24.11.0) has not been tested with this release of Rush. Please consider installing a newer version of the "@microsoft/rush" package, or downgrading Node.js.
```

### DIFF
--- a/docs/changehistory/NextVersion.md
+++ b/docs/changehistory/NextVersion.md
@@ -9,4 +9,4 @@ publish: false
 
 ## Node.js 24 support
 
-iTwin.js now officially supports Node.js 24 starting with LTS version of 24.11.0. Node.js 24 support is in addition to Node.js 20 and 22, not a replacement.
+In addition to [already supported Node.js versions](../learning/SupportedPlatforms.md#supported-nodejs-versions), iTwin.js now supports [Node.js 24](https://nodejs.org/en/blog/release/v24.11.0).
--- a/test-apps/display-performance-test-app/package.json
+++ b/test-apps/display-performance-test-app/package.json
@@ -7,9 +7,6 @@
     "url": "http://www.bentley.com"
   },
   "version": "0.0.0",
-  "engines": {
-    "node": "^20.0.0 || ^22.0.0 || ^24.0.0"
-  },
   "private": true,
   "scripts": {
     "build": "npm run -s build:frontend && npm run -s build:backend",
--- a/core/electron/package.json
+++ b/core/electron/package.json
@@ -3,9 +3,6 @@
   "version": "5.4.0-dev.2",
   "description": "iTwin.js ElectronHost and ElectronApp",
   "license": "MIT",
-  "engines": {
-    "node": "^20.0.0 || ^22.0.0 || ^24.0.0"
-  },
   "scripts": {
     "build": "npm run -s build:cjs && npm run -s webpack:test && npm run -s copy:test-assets",
     "build:cjs": "tsc 1>&2 --outDir lib/cjs",
--- a/test-apps/display-test-app/package.json
+++ b/test-apps/display-test-app/package.json
@@ -7,9 +7,6 @@
     "url": "http://www.bentley.com"
   },
   "version": "1.8.0",
-  "engines": {
-    "node": "^20.0.0 || ^22.0.0 || ^24.0.0"
-  },
   "private": true,
   "scripts": {
     "build": "npm run -s build:frontend && npm run -s build:backend",


docs/learning/SupportedPlatforms.md
```diff
@@ -17,8 +17,9 @@ The following Node.js versions are officially supported by the iTwin.js backend
 
 | iTwin.js - Node Support | iTwin.js 3.x | iTwin.js 4.x | iTwin.js 5.x |
 | ----------------------- | ------------ | ------------ | ------------ |
+| Node 22 (>=22.11)       | ❌           | ❌           | ✅ (>= 5.4) |
```
GytisCepk, line 20: To my knowledge 3.x is no longer maintained, so should we remove "iTwin.js 3.x" column ?

README.md
```diff
@@ -43,7 +43,7 @@ Each package has its own **node_modules** directory that contains symbolic links
 ## Prerequisites
 
 - [Git](https://git-scm.com/)
-- [Node](https://nodejs.org/en/): an installation of the latest security patch of Node 20. The Node installation also includes the **npm** package manager.
+- [Node](https://nodejs.org/en/): an installation of the latest security patch of Node 24. The Node installation also includes the **npm** package manager.
```
aruniverse, line 46: do we need to update the node ver badge at the top on L13?

test-apps/display-performance-test-app/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "version": "0.0.0",
   "engines": {
-    "node": "^20.0.0 || ^22.0.0"
+    "node": "^20.0.0 || ^22.0.0 || ^24.0.0"
```
aruniverse: do we need the engine field in apps?
should we bump the `@types/node` dep to 24?

docs/changehistory/NextVersion.md
```diff
@@ -1,5 +1,12 @@
...
 ---
+
 # NextVersion
 
+- [NextVersion](#nextversion)
+  - [Node.js 24 support](#nodejs-24-support)
+
+## Node.js 24 support
+
+iTwin.js now officially supports Node.js 24 starting with LTS version of 24.11.0. Node.js 24 support is in addition to Node.js 20 and 22, not a replacement.
```
aruniverse: feels like this can be reworded to match what you say for electron updates

docs/learning/SupportedPlatforms.md
aruniverse: i thought it might be nice to leave it in here for historical purposes and easy look up of what ver is compatible with what

common/config/azure-pipelines/integration-pr-validation.yaml
```diff
@@ -29,7 +29,7 @@ variables:
   - group: iTwin.js Integration Test Users
 
 jobs:
-  - job: Node_22_x
+  - job: Node_24_x
     condition: succeeded()
     pool:
       vmImage: ubuntu-22.04 # ubuntu-latest is 24.04, and breaks tests using sandbox
```
aruniverse, line 35: is this still valid?

README.md
GytisCepk: No, it's will automatically update once changes are in master.

test-apps/display-performance-test-app/package.json
GytisCepk: > do we need the engine field in apps?

Probably not. Now that I think about this, it probably does not make sense to have one in _core-electron_ package as well. I'll remove them.

> should we bump the @types/node dep to 24?

From PR description: 

_Left `@types/node` version unchanged as some time ago it was decided to use types of lowest support Node.js version to prevent developers accidentally consuming something which is not available in older supported versions._

docs/changehistory/NextVersion.md
GytisCepk: I just copied message it from Node 22 upgrade, so it's possible. However, I would appreciate some guidance on how to write this.

common/config/azure-pipelines/integration-pr-validation.yaml
GytisCepk: I'll take a look at this, but would prefer to do this as separate PR.

test-apps/display-performance-test-app/package.json
aruniverse: Yea i think the pr description comment makes sense for pkgs we publish but not for test apps

docs/changehistory/NextVersion.md
aruniverse: copying this from your electron update pr
image

In addition to [already supported Node.js versions](https://www.itwinjs.org/learning/supportedplatforms/#supported-nodejs-versions), iTwin.js now supports [Node 24](https://nodejs.org/en/blog/release/v24.11.0).

GytisCepk: Updated

test-apps/display-performance-test-app/package.json
GytisCepk: This would imply that test-apps will be only run on Node.js 24, which is not true. display-test-app can be run on Electron which uses Node.js 22 and, in my opinion, display-test-app backend for web should be able to run all supported Node.js versions

PR #8702, opened 2025-10-29
Temporarily skipping the health check stats test

naveedkhan8067:
We can skip it for now to unblock biscore version 01.00.25 release: https://github.com/iTwin/bis-schemas/pull/618. 

@RohitPtnkr1996 already have a fix for this test: https://github.com/iTwin/itwinjs-core/pull/8685

### DIFF
--- a/common/changes/@itwin/core-backend/io-temp-fix-health-stats-test_2025-10-29-16-26.json
+++ b/common/changes/@itwin/core-backend/io-temp-fix-health-stats-test_2025-10-29-16-26.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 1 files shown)


RohitPtnkr1996 (approved):
The test has no bearing on the BisCore version 25.
The real issue is that the changesets were being stored by their changeset ids.
And as hashes might change for each run, the order of the changesets returned might change.
This test expects the changesets to be in a particular order.
It has been fixed in the PR mentioned in the description.
But the pipelines are being blocked by getting flaky tests.
Hence this skip for now, just for the release.

ColinKerr:
Squash pushing suppression of this test because azure is unhappy.

PR #8703, opened 2025-10-29
Fixed flaky test

RohitPtnkr1996:
The expression in the test `Math.random() * (relationships.length + 1)` ends up with an index that might be greater than the size of the array, leading to an undefined being pushed into the array and a subsequent read of the undefined, leading to a flaky behavior of the test.

### DIFF
--- a/core/backend/src/test/imodel/IModel.test.ts
+++ b/core/backend/src/test/imodel/IModel.test.ts
@@ -3481,7 +3481,7 @@ describe("iModel", () => {
       // Create a subset of random relationship entries to delete
       const relationshipsToDelete: RelationshipProps[] = [];
       for (let i = 0; i < 250; ++i) {
-        relationshipsToDelete.push(relationships[Math.floor(Math.random() * (relationships.length + 1))]);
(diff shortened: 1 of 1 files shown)


RohitPtnkr1996:
Already fixed in https://github.com/iTwin/itwinjs-core/pull/8685

common/changes/@itwin/core-backend/io-temp-fix-health-stats-test_2025-10-29-16-26.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Temporarily skipping the health check stats test",
```
MichaelSwigerAtBentley, line 5:
```suggestion
      "comment": "",
```

We don't need to add test upkeep to the changelog

MichaelSwigerAtBentley: (just as a note for the future 😄)

PR #8705, opened 2025-10-29
Add new terminatorShape property in TextAnnotationLeader styles

SheetalNair:
Related issue #7830 
Adding different types of terminatorShape options for TextAnnotationLeaders
As this is an addition to textStyleSettings, the json version assigned to textStyleData has been updated and also migration code has been added to support old style data.
image
image
image
image
image
image


core/backend/src/annotations/LeaderGeometry.ts
```diff
@@ -78,26 +78,116 @@ export function appendLeadersToBuilder(builder: ElementGeometry.Builder, leaders
...
+    // Truncate the first segment of the leader lines to account for the arrowhead size when closedArrow (hollow triangle) terminatorShape is used.
+    if (leaderStyle.leader.terminatorShape === "closedArrow") {
+      if (terminatorDirection)
+        leaderLinePoints[0] = leaderLinePoints[0].plusScaled(terminatorDirection, terminatorWidth);
+    }
+
+    result = result && builder.appendGeometryQuery(LineString3d.create(leaderLinePoints));
+
+    // Terminator geometry
+    if (!terminatorDirection) continue; // Assuming leaders without terminators is a valid case.
```
pmconne: From what I can tell, you *don't* permit leaders without terminators. `TerminatorShape` has no "none" option and when you migrate pre-terminator text style settings you force them to use the default terminator shape. Did I overlook something? If not, is it true that we do not want to support leaders without terminators?

core/backend/src/annotations/LeaderGeometry.ts
```diff
@@ -78,26 +78,116 @@ export function appendLeadersToBuilder(builder: ElementGeometry.Builder, leaders
...
+  const terminatorHeight = textStyleSettings.leader.terminatorHeightFactor * textHeight;
+  const terminatorWidth = textStyleSettings.leader.terminatorWidthFactor * textHeight;
+  const basePoint = point.plusScaled(dir, terminatorWidth);
+  const point1 = basePoint.plusScaled(termY, terminatorHeight / 2);
+  const point2 = basePoint.plusScaled(termY.negate(), terminatorHeight / 2);
+  const radius = terminatorHeight / 2;
+
+  // Helper function to add fill parameters
+  const addFillParams = () => {
+    params.fillColor = ColorDef.black;
```
bbastings: Why is fill color hard-coded to black? By default, if the fillColor isn't specified it defaults to lineColor. This should be the desired behavior for a filled arrow head.

core/backend/src/annotations/LeaderGeometry.ts
SheetalNair: This code is a recovery from an incorrect data problem, We do not want to avoid adding a leader just because there is an issue in creating terminator geometry. So we just avoid adding the terminators in that case.
And about "None" TerminatorShape, I am planning to quickly add an option for that in this PR to be used in future, but the migration code will always assign openArrow, as the old version of data always had openArrow terminators

aruniverse:

core/backend/src/annotations/LeaderGeometry.ts
SheetalNair: fixed

core/backend/src/annotations/LeaderGeometry.ts
```diff
@@ -78,26 +78,117 @@ export function appendLeadersToBuilder(builder: ElementGeometry.Builder, leaders
...
+ * @param builder The geometry builder to append the terminator geometry to.
+ * @param point The starting point of the leader line.
+ * @param dir The direction vector of the leader line.
+ * @param params The geometry parameters to use for the terminator.
+ * @param textStyleSettings The text style settings to use for the terminator.
+ * @param textHeight The height of the text block.
+ * @returns True if the geometry was successfully created, false otherwise.
+ * @beta
+ */
+export function createTerminatorGeometry(builder: ElementGeometry.Builder, point: Point3d, dir: Vector3d, params: GeometryParams, textStyleSettings: TextStyleSettings, textHeight: number): boolean {
```
aruniverse, line 115: instead of a fn w/ multiple args, would a singular object that encapsulates all these be better and more future proof?

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -549,14 +549,19 @@ function updateTextBlocks(elem: TextAnnotation2d | TextAnnotation3d, textBlocks:
...
 */
-export const TEXT_STYLE_SETTINGS_JSON_VERSION = "1.0.0";
+// 1.0.1 - Added terminatorShapes for leaders
+export const TEXT_STYLE_SETTINGS_JSON_VERSION = "1.0.1";
 
 function migrateTextStyleSettings(oldData: VersionedJSON<TextStyleSettingsProps>): TextStyleSettingsProps {
   if (oldData.version === TEXT_STYLE_SETTINGS_JSON_VERSION) return oldData.data;
 
-  // Place migration logic here.
+  if (oldData.data.leader) {
```
aruniverse, line 558: does this if stmt make sense? shouldnt it be if the leader.terminatorShape is missing set it to the default val?

core/backend/src/annotations/LeaderGeometry.ts
aruniverse: im going to resolve the comment since other helper fns in this file does the same, but something to consider before making these public

core/backend/src/annotations/TextAnnotationElement.ts
SheetalNair: I did not notice this while merging, As I have merged this PR, Can this change be done in a subsequent PR?

aruniverse: yes

core/backend/src/annotations/LeaderGeometry.ts
SheetalNair: That would be good to have. Do you suggest this be done before this week's release?

core/backend/src/annotations/TextAnnotationElement.ts
pmconne: Please patch it (and add the unit test that would have caught it!) quickly so we can get the fix backported to 5.3.0 so that your team can use this.

aruniverse: plz push a change to the 5.3. bp pr as well
https://github.com/iTwin/itwinjs-core/pull/8711

core/backend/src/annotations/LeaderGeometry.ts
aruniverse: no thats fine, i think s larger refactor later on. i wont hold you guys up for next week

core/backend/src/annotations/TextAnnotationElement.ts
SheetalNair: https://github.com/iTwin/itwinjs-core/pull/8715

PR #8710, opened 2025-10-30
fix unsafe-inline csp violation in core-frontend

aruniverse:
https://content-security-policy.com/unsafe-inline/

### DIFF
--- a/docs/changehistory/5.1.0.md
+++ b/docs/changehistory/5.1.0.md
@@ -7,7 +7,7 @@ version: '5.1.0'
 
 Table of contents:
 
-- [5.1.0 Change Notes](#5.1.0 Change Notes)
+- [5.1.0 Change Notes](#510-change-notes)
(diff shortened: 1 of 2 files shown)


ben-polinsky:
I sort of question the validity of only setting `font-weight: normal` on the span. What is the point? We're not guarding against any other change? Font-size? Color? All other styles could be added/modded, but the element **has** to have `font-weight: normal`? Presumably spans start with a default font-weight of normal...

edit: But understand it's not that big of a deal.

aruniverse:
> I sort of question the validity of only setting `font-weight: normal` on the span. What is the point? We're not guarding against any other change? Font-size? Color? All other styles could be added/modded, but the element **has** to have `font-weight: normal`? Presumably spans start with a default font-weight of normal...
> 
> edit: But understand it's not that big of a deal.

I thought about removing the class on the span altogether and use the default, but headers are by default bold, and i guess we want to emphasize iTwin.js
image

aruniverse:

PR #8716, opened 2025-10-30
Add test cases for non-breaking whitespace characters for Codes

MichaelSwigerAtBentley:
Expands Code tests to include non-breaking whitespace character (`\xa0`) edge cases, for transformer

### DIFF
--- a/common/changes/@itwin/core-backend/affanK-rebase-bug_2025-08-07-10-26.json
+++ b/common/changes/@itwin/core-backend/affanK-rebase-bug_2025-08-07-10-26.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/changes/@itwin/core-backend/mike-abandon-changes-cache_2025-11-14-17-21.json
+++ b/common/changes/@itwin/core-backend/mike-abandon-changes-cache_2025-11-14-17-21.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Changes iModelDb.abandonChanges() to only clear element, model, and instanceKey caches instead of all backend caches",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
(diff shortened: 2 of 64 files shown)


core/backend/src/test/imodel/Code.test.ts
```diff
@@ -23,292 +23,409 @@ describe("Code insertion tests", () => {
...
+    it("should get undefined when querying an element with an empty code", () => {
+      assert.exists(imodel.elements);
+      const emptyCode = Code.createEmpty();
+      const elementId = imodel.elements.queryElementIdByCode(emptyCode);
+      assert.isUndefined(elementId);
+    });
+
+    it("should get undefined when querying an element with a non-breaking space value", () => {
+      assert.exists(imodel.elements);
+      const nonBreakingSpaceCode = new Code({ scope: "0x11", spec: "0x10", value: "\xa0" });
```
DanRod1999, line 64: does an element in the imodel exist with this code value? Since we don't accept this space values I assume the code in reality would be {... value: ""}. Not sure if that is a legal value in general though.

Is this just checking that it returns undefined and not an error?

core/backend/src/test/imodel/Code.test.ts
```diff
@@ -23,292 +23,409 @@ describe("Code insertion tests", () => {
...
+      assert.equal(updatedElement3.code.value, newCode.value);
+      assert.equal(updatedElement3.code.spec, newCode.spec);
+      assert.equal(updatedElement3.code.scope, newCode.scope);
+
+      newCode = new Code({ scope: "0x1", spec: "0x11", value: "\xa0" }); // non-breaking space value
+      element.code = newCode;
+      element.update();
+
+      const updatedElement4 = imodel.elements.getElement(elementId);
+      assert.equal(updatedElement4.code.value, "");
```
DanRod1999, line 417: do we have any docs or comments to explain why we expect this behavior. For example is this just for trailing spaces, or would something like "hello\xa0world" turn into "helloworld"?

core/backend/src/test/imodel/Code.test.ts
MichaelSwigerAtBentley: An element with "" is inserted, but not \xa0 (which just inserts as "" anyway)

I could add an insert though to these tests to make it explicit

core/backend/src/test/imodel/Code.test.ts
MichaelSwigerAtBentley: I don't think so, https://www.itwinjs.org/bis/guide/fundamentals/codes/ has some notes about value uniqueness, but does not go into specifics. I think it should be outlined there

It only trims trailing spaces, for instance, code values with multiple words like "this is my code value" are fine

core/backend/src/test/imodel/Code.test.ts
MichaelSwigerAtBentley: Added

PR #8718, opened 2025-10-31
Downgrade to Vite 6

GytisCepk:
Recently, test-apps were upgraded to use Vite 7, which only supports Node 20.19 and higher. This is a problem for us since this causes regression pipeline to fail `rush build` as our lowest supported Node version is 20.11.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -898,7 +898,7 @@ importers:
         version: 17.0.2
       '@vitest/browser':
         specifier: ^3.0.6
-        version: 3.0.6(@types/node@22.13.10)(playwright@1.56.1)(typescript@5.6.2)(vite@7.1.7(@types/node@22.13.10)(terser@5.39.0)(yaml@2.7.0))(vitest@3.0.6)
+        version: 3.0.6(@types/node@22.13.10)(playwright@1.56.1)(typescript@5.6.2)(vite@6.4.1(@types/node@22.13.10)(terser@5.39.0)(yaml@2.7.0))(vitest@3.0.6)
       '@vitest/coverage-v8':
         specifier: ^3.0.6
         version: 3.0.6(@vitest/browser@3.0.6)(vitest@3.0.6)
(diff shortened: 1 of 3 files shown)


aruniverse:
Should our lowest supported ver be 20.11 or latest 20.x? Not like nodejs is continuously patching anything but the last minor

hl662:
> this causes regression pipeline to fail rush build as our lowest supported Node version is 20.11.

What was the exact error? Was it 'exports is not defined'? Or vite complaining about the minimum supported version?

GytisCepk:
> > this causes regression pipeline to fail rush build as our lowest supported Node version is 20.11.
> 
> What was the exact error? Was it 'exports is not defined'? Or vite complaining about the minimum supported version?

```
--[ FAILURE: display-test-app ]------------------------------[ 9.34 seconds ]--

You are using Node.js 20.11.0. Vite requires Node.js version 20.19+ or 22.12+. Please upgrade your Node.js version.
✗ Build failed in 614ms
error during build:
[vite:build-html] crypto.hash is not a function
...
```

GytisCepk:
> Should our lowest supported ver be 20.11 or latest 20.x? Not like nodejs is continuously patching anything but the last minor

I think, it should be 20.11. The same way we don't bump peer dependencies in non-major releases

hl662:
> I think, it should be 20.11. 

I feel the same way - 20.19 enables a previously experimental feature [on by default](https://nodejs.org/en/blog/release/v20.19.0#requireesm-is-now-enabled-by-default), and if we want to update our lowest supported ver we would want to dedicate a bit of work on making sure all our dual and non dual exported packages are fine.

Previously we ran into a regression during iTwin 5.1 where our first party product ran into issues with [ESM](https://github.com/iTwin/itwinjs-core/pull/8406). I asked an app dev to try bumping their node ver to 20.19, it didn't fix their issue, stating that our linear-referencing-backend pkg does not have an 'exports' field (note this was when the pkg was ESM only).

aruniverse:

PR #8719, opened 2025-10-31
TextField fixes and display-test-app editing enhancements

pmconne:
Part of #8534.

- [x] Add `dta text field` keyin to add a field to a TextBlock.
- [x] Fix `layoutTextBlock` ignoring fields.
- [x] Permit display-test-app to write to channels specified by the `IMJS_ALLOWED_CHANNELS` environment variable.
- [x] Fix "Invalid JSON" response from `generateElementGraphics` when the geometry stream was produced from a TextBlock containing a field.
- [x] Throw on an attempt to insert a field into an iModel whose version of BisCore does not include `ElementDrivesTextAnnotation`, instead of silently creating fields that never evaluate their display strings.
- [x] Ensure fields are evaluated on insert/update.
- [x] Add extremely basic coordinate formatting to prevent fields from evaluating to `[Object object]`.
- [x] Add API to evaluate fields for a `TextBlock`, and use it for display-test-app's text decoration.

The bugs were discovered while implementing the `dta text field` keyin. Their fixes should be backported to 5.3.x.

### DIFF
--- a/common/changes/@itwin/core-backend/pmc-dta-field-keyin_2025-11-03-10-51.json
+++ b/common/changes/@itwin/core-backend/pmc-dta-field-keyin_2025-11-03-10-51.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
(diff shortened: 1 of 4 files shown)


pmconne:
@aruniverse fyi this PR contains some fixes needed in 5.3.x.

pmconne:
@Mergifyio backport release/5.3.x

PR #8721, opened 2025-11-03
Added test for pulling schema upgrades after data changes

soham-bentley:
imodel-native: https://github.com/iTwin/imodel-native/pull/1283


core/backend/src/IModelDb.ts
```diff
@@ -776,6 +776,7 @@ export abstract class IModelDb extends IModel {
     this.models[_cache].clear();
     this.elements[_instanceKeyCache].clear();
     this.models[_instanceKeyCache].clear();
+    this[_nativeDb].clearECDbCache();
```
soham-bentley: I have added clearECDbCache() call inside the clearCaches method of iModelDb
Does this look like a good place?
I saw in some places we were calling both clearCaches() of iModelDb and also clearECDbCache() on native db
And in some places we were calling only clearCaches() of iModelDb
Does this sound like a good idea to clear native ecdb cache as well inside clearCaches() method of iModelDb?
Or it can cause serious performance implications?
@khanaffan  @rschili @MichaelSwigerAtBentley

rschili: We do not have performance tests that would sufficiently cover this. And covering it well seems quite difficult. We call clearCache from many spots, and doing this additional cache clear could have unpredictable performance implications.

I don't have a good answer to this. Leaving the bug in is certainly the worst of our options. This just shows that we need well maintained reliable performance tests...

It's impossible to predict what side effects this can have without real testing I'm afraid...

soham-bentley: There is one (not very good) workaround to this
Instead of adding clearECDbCache() inside clearCaches() we can just call clearECDbCache() inside applySingleChangeset() of BriefcaseManager()....if the changeset type is a schema one
This also solves the bug.
Does that sound more reasonable?

rschili: Quickly checking the code, doing it in applySingleChangeset seems solid to me. I see no other code paths that do the similar thing which we'd be missing (though I may be wrong).

The only caller being `pullAndApplyChangesets` I wonder if we need to clear after every single apply or if clearing after applying all changesets would also fix the bug?

soham-bentley: We need to clear before applying changeset if it is a schema changeset

soham-bentley: > Quickly checking the code, doing it in applySingleChangeset seems solid to me. I see no other code paths that do the similar thing which we'd be missing (though I may be wrong).
> 
> The only caller being `pullAndApplyChangesets` I wonder if we need to clear after every single apply or if clearing after applying all changesets would also fix the bug?

Modified the fix and instead of calling `clearECDbCache()` in `clearCaches()`, I am just calling it inside `applySingleChangeSet()` when the changeset type is `Schema` or `SchemaSync`

khanaffan: I don’t agree with the performance argument. From what I know, no application explicitly calls `clearCaches()`. It’s invoked only by our own code and tests. Below is a list of functions that call `clearCaches()`. If you review them, these are significant events where the schema may have changed—not something you’d expect to run in a loop. Also, note that most of these already call `ClearECDbCaches()` on the native side. In my opinion, handling this on the native side makes sense. 

But what I really want to move to is that itwin.js `clearCaches()` include `clearECDbCache()` so we know exactly what it does. Currently clearCaches() take care itwin.js caches but the only reason we every clear them is same as when we have to call clearECDbCaches() which is when schema has changed. So why have two level of functionality?

1. `ChangeSummary.detachChangeCache()`
2. `IModelDb.detachDb()`
3. `IModelDb.beforeClose()`
4. `IModleDb.abandonChanges()`
5. `IModelDb.performCheckpoint()`
6. `IModelDb.dropSchemas()`
7. `IModelDb.importSchemas()`
8. `IModelDb.importSchemaStrings()`
9. `IModelDb.discardChanges()`
10. `IModelDb.closeAndReopen()`
11. `IModelDb.revertAndPushChanges()`
12. `SchemaSync.pull()`
13. `TxnManager._onChangesApplied()`
14. `BriefcaseManager.applySingleChangeset()`

khanaffan: `IModleDb.abandonChanges` should not call clear cache. It just want to clear instance cache. @MichaelSwigerAtBentley

MichaelSwigerAtBentley: [Created a separate issue for tracking here](https://github.com/iTwin/itwinjs-core/issues/8768)
I will address in a separate PR

soham-bentley: > I don’t agree with the performance argument. From what I know, no application explicitly calls `clearCaches()`. It’s invoked only by our own code and tests. Below is a list of functions that call `clearCaches()`. If you review them, these are significant events where the schema may have changed—not something you’d expect to run in a loop. Also, note that most of these already call `ClearECDbCaches()` on the native side. In my opinion, handling this on the native side makes sense.
> 
> But what I really want to move to is that itwin.js `clearCaches()` include `clearECDbCache()` so we know exactly what it does. Currently clearCaches() take care itwin.js caches but the only reason we every clear them is same as when we have to call clearECDbCaches() which is when schema has changed. So why have two level of functionality?
> 
> 1. `ChangeSummary.detachChangeCache()`
> 2. `IModelDb.detachDb()`
> 3. `IModelDb.beforeClose()`
> 4. `IModleDb.abandonChanges()`
> 5. `IModelDb.performCheckpoint()`
> 6. `IModelDb.dropSchemas()`
> 7. `IModelDb.importSchemas()`
> 8. `IModelDb.importSchemaStrings()`
> 9. `IModelDb.discardChanges()`
> 10. `IModelDb.closeAndReopen()`
> 11. `IModelDb.revertAndPushChanges()`
> 12. `SchemaSync.pull()`
> 13. `TxnManager._onChangesApplied()`
> 14. `BriefcaseManager.applySingleChangeset()`

As most of these methods internally call `clearECDbCaches()` on the native side, if we club `clearECDbCaches()` with `clearCaches()` on the itwinjs-core side, we must remove the calls from the native side to avoid duplicate calls. But does that not break our native api behavior?
Because our native apis are also public and can be used independently.
I personally think ecdb cache is a native cache mechanism so clearing it should also be the responsibility of native.

khanaffan: We do not have to remove calls to `clearECDbCaches()` from native side. calling clearECDbCaches() twice or multiple time doesn't cost anything. Its only first time it actually release any cached object. Calling it again after that with caches still empty does not cost anything.

soham-bentley: Fixed this on the native side....will be consolidating clearECDbCache inside ClearCaches() in a separate PR for separation of concerns

rschili: Closing comment.

soham-bentley: > Fixed this on the native side....will be consolidating clearECDbCache inside ClearCaches() in a separate PR for separation of concerns

https://github.com/iTwin/itwinjs-core/issues/8776 here is the ticket to track this

PR #8724, opened 2025-11-03
Schema-level locking for multiuser scenarios

rschili:
As part of the effort for making schema merges less disruptive to OS+, described here: https://github.com/iTwin/itwinjs-backlog/issues/1703

We want to make importing schemas not require a full exclusive lock over the entire imodel. We will introduce a new schema-level lock instead.
Using this lock, we prevent multiple people from importing schemas, but if a schema import is not affecting mappings (does not move data around), we want to allow it while people keep making data changes to their imodels.

The goal is to first attempt the import with the schema-level lock, and if moving data is required, the attempt will fail, itwin-js will then obtain the full exclusive lock as it did before and re-attempt the import.
This isn't ideal, but we currently have no good way to tell if a schema import will require data changes as it depends on many different factors (type of schema edits, current mappings, existing data, table size limits, available columns...)

This is a short term solution timeframed around end of december 2025.

Closes: https://github.com/iTwin/itwinjs-backlog/issues/1737

### DIFF
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -338,7 +338,7 @@ export abstract class IModelDb extends IModel {
       description: "Reserved element for schema table locking",
     };
     // need a full exclusive lock to insert the schema table subject element otherwise others could be importing schemas at the same time.
-    await this.locks.acquireLocks({ exclusive: [IModel.dictionaryId, IModel.rootSubjectId] });
+    await this.acquireSchemaLock();
     try {
       const id = this.elements.insertElement(schemaSubjectProps, { forceUseId: true });
       if (id !== IModel.schemaElementId)


core/common/src/IModel.ts
```diff
@@ -377,6 +377,8 @@ export abstract class IModel implements IModelProps {
   public static readonly rootSubjectId: Id64String = "0x1";
   /** The Id of the dictionary model. */
   public static readonly dictionaryId: Id64String = "0x10";
+  /** The Id used for schema locks. */
+  public static readonly schemaLockId: Id64String = "0x11";
```
rschili: @khanaffan is this what you had in mind? I'm locking an imaginary element here... When you first described the approach to me I thought we had a better mechanism for this, but seems like all locks are element-based.

core/backend/src/IModelDb.ts
```diff
@@ -302,11 +302,11 @@ export abstract class IModelDb extends IModel {
    * will be able to acquire *any* locks while the schema lock is held.
    */
   public async acquireSchemaLock(): Promise<void> {
```
khanaffan, line 313: This appears to be a breaking change in behavior. Instead, we should aim to **extend the current behavior** without disrupting the existing API functionality.

I recommend **not modifying the current API**, so both older and newer applications can maintain a consistent understanding of what a schema lock represents.

For limited schema locking, we should modify the entry point in `importSchemas()` / `importSchemasXml()` to acquire a less restrictive lock — let’s call it `additiveSchemaChangeOnlyLock`.

The process should be as follows:

1. Acquire an **exclusive `additiveSchemaChangeOnlyLock`**.
2. Attempt to import the schema:
   - If it fails with `REQUIRE_DATA_TRANSFORM`, then:
     - Acquire an **exclusive `schemaLock`**.
     - Retry the schema import with `schemaLockHeld: true`.

**Important considerations:**

The new and old software versions will only be compatible if:

1. The `additiveSchemaChangeOnlyLock` has an ID that is a **true child of `rootSubjectId`**, **or**
2. The schema import process **always acquires a shared lock on `rootSubjectId`** before taking an exclusive lock on `additiveSchemaChangeOnlyLock`.

Without one of these safeguards, two versions of the software could import schemas in parallel, potentially resulting in:
- A changeset that cannot be applied, or
- A corrupt schema mapping if applied.

core/common/src/IModel.ts
khanaffan: There are no reserved element IDs apart from the ones created when an `IModel` is initialized. The two known reserved IDs are:

1. `RepositoryModelId`: `0x1`  
2. `DictionaryId`: `0x10`

I wish we had reserved the first 100 element IDs, but unfortunately, we didn’t.

The problem with choosing `0x11` is that this ID might already exist in some iModels. Here's why:

- If an application initializes a standalone database with seed data and later uploads it as a briefcase, the next element ID generated will be `0x11`, since IDs are assigned using `MAX(ElementId) + 1`.

While a briefcase can never push `0x11` (because its element IDs include the briefcase ID), a non-briefcase seed file **can** contain this ID, which could lead to conflicts.

### Options to consider:

1. **Choose an ID less than `0x10`**, such as `0x9`.
2. **Create a child of `0x1`** as a placeholder, using a briefcase-based ID. You can then look it up by `UserLabel` or `CodeValue` before locking it.

My recommendation is to go with the simpler option: use `0x9`, and if possible, ensure there’s a backing element for it in any newly created or upgraded iModel to serve as a placeholder.

core/backend/src/IModelDb.ts
rschili: I just realized, this is either a bug or very bad naming. We used to lock "repositoryModelId" here, but when we pass this ID on to hub, it's actually treated as an Element ID. They use Element ID 1 to represent a "SchemaLock", but that of course locks the whole file.
It's NOT a repository model ID at all.
New Hub API only ever locks elements, and nothing else.

IModel.repositoryModelId is defined as 1.

core/backend/src/IModelDb.ts
```diff
@@ -991,23 +997,58 @@ export abstract class IModelDb extends IModel {
...
+   * @param schemaFileName  array of Full paths to ECSchema.xml files to be imported.
+   * @param {SchemaImportOptions} options - options during schema import.
+   * @throws [[IModelError]] if the schema lock cannot be obtained or there is a problem importing the schema.
+   * @note Changes are saved if importSchemas is successful and abandoned if not successful.
+   * - You can use NativeLoggerCategory to turn on the native logs. You can also control [what exactly is logged by the loggers](https://www.itwinjs.org/learning/common/logging/#controlling-what-is-logged).
+   * - See [Schema Versioning]($docs/bis/guide/schema-evolution/schema-versioning-and-generations.md) for more information on acceptable changes to schemas.
+   * @note This method should not be called from {TxnManager.withIndirectTxnModeAsync} or {RebaseHandler.recompute}.
+  * @see querySchemaVersion
+   */
+  public async importSchemas(schemaFileNames: LocalFileName[], options?: SchemaImportOptions): Promise<void> {
```
rschili, line 1109: @ColinKerr @StefanApfel-Bentley @pmconne @aruniverse @wgoehrig 
We need to roll this out very soon (this week) so it makes its way into the next OS+ build and they can start consuming it.
I do not want to use the new locking mechanism for everybody right away, instead, we need some sort of feature flag or mechanism to turn this on and off.
I have not given this much thought yet, just assumed we had some existing feature flag solution that I could replicate. Apparently we don't.
Using a global flag in backend seems silly as that would not play nice with multiple frontends. Providing a local option with the method call also seems wrong, as importSchemas gets implicitly called by other calls, so we'd have to trace all callers and pass this through all the layers. And once we make it default, that'd be another break.
I think we need some sort of feature flag mechansim that is based on some caller context, so the same frontend/user retain their options.

rschili: What I'm asking is.... should this be put into IModelHost options, into a global property, into some frontend context thingy or someplace completely else? I think I'll just go ahead and stuff it into IModelHost options for now and mark it as alpha, so we can remove it later once we make this behavior the new default.

rschili: Decided to use IModelHostOptions for now

core/backend/src/IModelDb.ts
```diff
@@ -169,6 +169,13 @@ const withBriefcaseDb = async (briefcase: OpenBriefcaseArgs, fn: (_db: Briefcase
   }
 };
 
+function isDataTransformRequiredError(error: unknown): error is ITwinError {
```
rschili, line 174: @khanaffan @ColinKerr 
This feels messy. I cannot use ITwinError typesafe as shown below. We rethrow a new error sometimes which only preserves the "errorNumber" field which was accessed on :any.
Using ITwinError.isError with a string seems like a better solution, or maybe having an overload that takes an error number...
I would prefer some standard way for doing this.

kabentley: we should stop using ErrorNumber everywhere. Throw an iTwinError from native code.

docs/learning/backend/IModelFileFormat.md
```diff
@@ -0,0 +1,131 @@
+# iModel File Format
```
rschili, line 1: @ColinKerr @khanaffan Please review this thoroughly.
I did not write all of this by hand, instead I had copilot help with it, though, it produced a ton of nonsense which I then rewrote/removed.
It's supposed to be a start, not a complete documentation. We didn't have anything existing to begin with so we can add to this as we go.

kabentley:
I don't see how you can ever perform a schema upgrade that requires data changes without first acquiring the schema lock. How will that work?

rschili:
> I don't see how you can ever perform a schema upgrade that requires data changes without first acquiring the schema lock. How will that work?

@kabentley the current plan for 2026 is to introduce a high level version of changesets which is not tied to the binary state of an imodel but instead contains a higher level representation of the changes, so when applying, each briefcase can replay the changes and make necessary data changes.

We'll also explore ways to store data transformations happening during schema updates, and then replay them during rebase, which may turn out not feasible. At first glance this seems quite messy, but it's much less work.

kabentley:
> to introduce a high level version of changesets which is not tied to the binary state of an imodel but instead contains a higher level representation of the changes, so when applying, each briefcase can replay the changes and make necessary data changes.

I'd have to have this explained in detail, but it sounds extremely complicated and unnecessary. Schema changes that require data changes should be rare and happen only by admins. I strongly suggest you make them only while the project is "offline". Anything else will be a risky mess.

For example, any schema change that remaps/removes/changes properties will undoubtedly require new versions of all apps that can use them. It will be impossible for a user to "pull changes" in an active session that require new versions of the very apps that made their local changes they need to rebase.

That is why major database upgrades happen with the system offline. People expect that, it's not an undue burden, and it can't possibly rise to a high enough priority to work on given all the other real problems we need to address.

core/backend/src/IModelDb.ts
```diff
@@ -302,11 +309,25 @@ export abstract class IModelDb extends IModel {
    * will be able to acquire *any* locks while the schema lock is held.
    */
   public async acquireSchemaLock(): Promise<void> {
-    return this.locks.acquireLocks({ exclusive: IModel.repositoryModelId });
+    return this.locks.acquireLocks({ exclusive: IModel.rootSubjectId }); // TODO: ModelId treated like Element ID for the lock? This matches rootSubjectId.
```
ColinKerr: repository model id and root subject id will always be the same.  What is the TODO for here?

core/bentley/src/BentleyError.ts
```diff
@@ -572,7 +572,7 @@ export class BentleyError extends Error { // note: this class implements LegacyI
       case DbResult.BE_SQLITE_IOERR_SEEK: return "BE_SQLITE_IOERR_SEEK";
       case DbResult.BE_SQLITE_IOERR_DELETE_NOENT: return "BE_SQLITE_IOERR_DELETE_NOENT";
 
-      case DbResult.BE_SQLITE_ERROR_DataTransformRequired: return "Schema update require to transform data";
+      case DbResult.BE_SQLITE_ERROR_DataTransformRequired: return "Schema update required to transform data";
```
ColinKerr: Is this right?  A schema update is not required to transform data, the schema update requires that the data be transformed.

docs/changehistory/NextVersion.md
```diff
@@ -6,7 +6,27 @@ publish: false
 
 - [NextVersion](#nextversion)
   - [Node.js 24 support](#nodejs-24-support)
+  - [Schema table locks for concurrent schema modifications](#schema-table-locks-for-concurrent-schema-modifications)
 
 ## Node.js 24 support
 
 In addition to [already supported Node.js versions](../learning/SupportedPlatforms.md#supported-nodejs-versions), iTwin.js now supports [Node.js 24](https://nodejs.org/en/blog/release/v24.11.0).
+
+## Schema table locks for concurrent schema modifications
```
ColinKerr: please make more clear that this is an experimental feature

core/bentley/src/BentleyError.ts
rschili: Oh, you are correct. I just noticed and fixed the wrong grammar without noticing the message itself is wrong...

rschili: fixed

docs/changehistory/NextVersion.md
rschili: updated

core/backend/src/IModelDb.ts
rschili: right, I removed the TODO. sticking with using rootSubjectId here as it seems more consistent, logic remains identical

ColinKerr:
> > to introduce a high level version of changesets which is not tied to the binary state of an imodel but instead contains a higher level representation of the changes, so when applying, each briefcase can replay the changes and make necessary data changes.
> 
> I'd have to have this explained in detail, but it sounds extremely complicated and unnecessary. Schema changes that require data changes should be rare and happen only by admins. I strongly suggest you make them only while the project is "offline". Anything else will be a risky mess.
> 
> For example, any schema change that remaps/removes/changes properties will undoubtedly require new versions of all apps that can use them. It will be impossible for a user to "pull changes" in an active session that require new versions of the very apps that made their local changes they need to rebase.
> 
> That is why major database upgrades happen with the system offline. People expect that, it's not an undue burden, and it can't possibly rise to a high enough priority to work on given all the other real problems we need to address.

The current change allows us to make minor schema changes without taking a full db lock while also ensuring those schema changes are done in tandem with data upgrades to the data in a channel.  This simplifies channel upgrades that include both application schema and related data changes.  These changes happen regularly as the application is updated, they may be deferred but are necessary to use new application features.

In 2026 we hope to investigate supporting minor schema changes that result in table mapping changes without a full db lock.  This changes are minor and compatible changes from a schema point of view but result in changes to the internal table mapping used.  Major schema changes would still require a full db lock.

core/common/src/IModel.ts
```diff
@@ -377,6 +377,8 @@ export abstract class IModel implements IModelProps {
   public static readonly rootSubjectId: Id64String = "0x1";
   /** The Id of the dictionary model. */
   public static readonly dictionaryId: Id64String = "0x10";
+  /** The Id representing the schema element for locking. */
+  public static readonly schemaElementId: Id64String = "0x2";
```
pmconne, line 383: If I understand correctly, this refers to a non-existent element. Your PR description claims this is a short-term solution. Will the long-term solution also involve the use of pretend elements?

If no, why make this element Id part of the public API?

If yes, does everybody who wants to interact with locks now need to properly handle locked "elements" that don't refer to real elements? (simple example: a UI displaying the name/guid/code/whatever of all of the elements the briefcase has locked).

rschili: Good point.
I'll flag this as internal for now.
We don't know about the long-term solution yet, so I can't say if it will involve further use of that mechanism.
Whether to show this in a UI... I'm not sure, that is a good question. I'll bring it up in the next multi-user meeting.
Edit: flagging done

pmconne: > Whether to show this in a UI

I'm not specifically asking about UI. I'm saying that this new "element" breaks all kinds of assumptions that all pre-existing code has been written against. e.g., that Ids refer to real elements; that all element Ids except the root subject have a parent element Id; etc. If this fake element Id starts showing up in such contexts it may break things.

Did the BIS folks discuss this, and whether or not there may be a case for using an actual element instead of a pretend one?

kabentley: Please listen to Paul's advice. The "lock" you need isn't an element lock at all - it's a lock on the ClouldSqlite schemaDb, right? Why use element based "locks: at all? Presuming you can use a new static elementId is confusing, risky and unnecessary.

core/backend/src/IModelDb.ts
ColinKerr: We have inconsistencies across the stack, we need to document best practices and code so we are consistent.  Created a new issue to track larger issue and linked in one other existing issue on same topic.  #8774

@rschili  please add this inconsistency to the larger issue and link it here before you resolve this.

rschili:
Abandoning change, we will not pursue this further. There is now semantic rebase which has some overlap with this, for others there is schema sync.

PR #8725, opened 2025-11-03
Exposed `MapLayerImageryProvider.status` as preview api

mdastous-bentley:
When the `DisplayStyleState.attachMapLayer()` method  is used to attach a map-layer, the consuming application as no way to determine if the  initialization / tile tree creation failed.

A typical case of this is when a  map-layer requires  authentication, which remain un-displayable until a authentication token is made available to the imagery provider.  The consuming application needs a way to to get the status of map-layer and inform the end-user.

### DIFF
--- a/common/changes/@itwin/core-frontend/MichelD-providerStatus_2025-11-03-13-31.json
+++ b/common/changes/@itwin/core-frontend/MichelD-providerStatus_2025-11-03-13-31.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/core-frontend",
-      "comment": "Exposed `MapLayerImageryProvider.status` as beta api.",
+      "comment": "Added `MapLayerImageryProvider.status` API.",
       "type": "none"
     }
   ],


core/frontend/src/tile/map/MapLayerImageryProvider.ts
```diff
@@ -65,7 +65,10 @@ export abstract class MapLayerImageryProvider {
   /** @internal */
   protected _firstRequestPromise: Promise<void>|undefined;
 
-  /** @internal */
+  /**
+   * The status of the map layer imagery provider.
+   * @beta
```
pmconne:
```suggestion
   * @public @preview
```

Unless you have some reason to think we will make breaking changes to this API before iTwin.js 6.0?

common/changes/@itwin/core-frontend/MichelD-providerStatus_2025-11-03-13-31.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Exposed `MapLayerImageryProvider.status` as beta api.",
```
pmconne:
```suggestion
      "comment": "Added `MapLayerImageryProvider.status` API.",
```

common/api/core-frontend.api.md
```diff
@@ -6063,7 +6063,7 @@ export abstract class MapLayerImageryProvider {
     setStatus(status: MapLayerImageryProviderStatus): void;
     // (undocumented)
     protected readonly _settings: ImageMapLayerSettings;
-    // @internal (undocumented)
+    // @public @preview
```
aruniverse, line 6066: why did we promote from internal to public preview
we should leave this as beta as the pr originally had

aruniverse:

common/api/core-frontend.api.md
aruniverse: disregard, based on pauls comments above

mdastous-bentley: its already in used by map-layers widget, but I don't mind, one or the other.

@pmconne  ?

PR #815, opened 2021-02-19
Support for editing standalone files in Electron ui-test-app

DanEastBentley:

PR #8730, opened 2025-11-05
Avoid Intel Ultra 7 graphical glitches during vertex discard

markschlosseratbentley:
Certain Intel Ultra 7 chipsets reported in https://github.com/iTwin/itwinjs-core/issues/815, https://github.com/IGCIT/Intel-GPU-Community-Issue-Tracker-IGCIT/issues/1165, and elsewhere exhibit glitches when discarding triangles using the vertex shader (setting all vertices to the same value) if gl_Position was not initialized to a valid position beforehand.

Known to affect much of the Intel Ultra 7 family of chipsets when using Intel driver from some point after driver version 32.0.101.6078 (9/13/2024).

The workaround for this bug involves ensuring that gl_Position is initialized to a valid position before attempting a discard using a degenerate triangle.

This PR makes that happen for all hardware and drivers, even hardware and drivers that do not exhibit this problem. We could consider reverting this workaround once a driver fix occurs.

TODO:
- [x] Run Image Tests
- [x] Finish adding vertex discard unit test - removed because no longer relevant with latest version of workaround
- [x] Verify the workaround always works in the cases where vertex discard is utilized
- [x] Add unit test in Compatibility.test.ts which verifies regex matching works
- [x] Add a suitable coverage of chipset regexes for primary devices that are affected - Resolved: We will unconditionally enable workaround because performance impact is minimal.
- [x] Test performance impact of workaround

### DIFF
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -9855,7 +9855,7 @@ export class SheetViewState extends ViewState2d {
     // (undocumented)
     get areAllTileTreesLoaded(): boolean;
     // (undocumented)
-    get attachmentIds(): Id64Array;
+    get attachmentIds(): readonly string[];
     // @internal
     get attachments(): object[] | undefined;
     attachToViewport(args: AttachToViewportArgs): void;
@@ -9894,6 +9894,7 @@ export class SheetViewState extends ViewState2d {
     isDrawingView(): this is DrawingViewState;
     // (undocumented)
     isSheetView(): this is SheetViewState;
+    readonly onViewAttachmentsReloaded: BeEvent<() => void>;
     // @internal (undocumented)
     protected postload(hydrateResponse: HydrateViewStateResponseProps): Promise<void>;
     // @internal (undocumented)
@@ -9904,7 +9905,7 @@ export class SheetViewState extends ViewState2d {
     // (undocumented)
     toProps(): ViewStateProps;
     // @internal
-    get viewAttachmentInfos(): Id64Array | Array<{
+    get viewAttachmentInfos(): readonly Id64String[] | Array<{
         attachedView: ViewState;
     }>;
     // @internal
--- a/common/changes/@itwin/core-frontend/pmc-react-to-view-attachment-changes_2025-11-10-16-06.json
+++ b/common/changes/@itwin/core-frontend/pmc-react-to-view-attachment-changes_2025-11-10-16-06.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Update view when view attachment elements are modified.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/core/frontend/src/SheetViewState.ts
+++ b/core/frontend/src/SheetViewState.ts
@@ -6,35 +6,26 @@
  * @module Views
  */
 
-import { assert, CompressedId64Set, dispose, expectDefined, Id64Array, Id64String } from "@itwin/core-bentley";
-import { Angle, ClipShape, ClipVector, Constant, Matrix3d, Point2d, Point3d, PolyfaceBuilder, Range2d, Range3d, StrokeOptions, Transform } from "@itwin/core-geometry";
+import { BeEvent, dispose, expectDefined, Id64Array, Id64String } from "@itwin/core-bentley";
+import { Angle, Constant, Point2d, Point3d, Range2d, Range3d, Transform } from "@itwin/core-geometry";
 import {
-  AxisAlignedBox3d, ColorDef, Feature, FeatureTable, Frustum, Gradient, GraphicParams, HiddenLine, HydrateViewStateRequestProps, HydrateViewStateResponseProps, PackedFeatureTable, Placement2d, SheetProps,
-  TextureTransparency, ViewAttachmentProps, ViewDefinition2dProps, ViewFlagOverrides, ViewStateProps,
+  AxisAlignedBox3d, ColorDef, Gradient, GraphicParams, HydrateViewStateRequestProps, HydrateViewStateResponseProps, SheetProps,
+  ViewAttachmentProps, ViewDefinition2dProps, ViewStateProps,
 } from "@itwin/core-common";
 import { CategorySelectorState } from "./CategorySelectorState";
 import { DisplayStyle2dState } from "./DisplayStyleState";
 import { IModelConnection } from "./IModelConnection";
 import { GraphicBuilder } from "./render/GraphicBuilder";
 import { RenderGraphic } from "./render/RenderGraphic";
-import { GraphicBranch } from "./render/GraphicBranch";
 import { Frustum2d } from "./Frustum2d";
-import { Scene } from "./render/Scene";
-import { Decorations } from "./render/Decorations";
-import { MockRender } from "./internal/render/MockRender";
-import { RenderClipVolume } from "./render/RenderClipVolume";
 import { RenderMemory } from "./render/RenderMemory";
-import { FeatureSymbology } from "./render/FeatureSymbology";
 import { DecorateContext, SceneContext } from "./ViewContext";
-import { IModelApp } from "./IModelApp";
-import { CoordSystem } from "./CoordSystem";
-import { OffScreenViewport, Viewport } from "./Viewport";
+import { Viewport } from "./Viewport";
 import { AttachToViewportArgs, ComputeDisplayTransformArgs, GetAttachmentViewportArgs, ViewState, ViewState2d } from "./ViewState";
 import { DrawingViewState } from "./DrawingViewState";
-import { createDefaultViewFlagOverrides, DisclosedTileTreeSet, TileGraphicType } from "./tile/internal";
-import { imageBufferToPngDataUrl, openImageDataUrlInNewWindow } from "./common/ImageUtil";
-import { ViewRect } from "./common/ViewRect";
+import { DisclosedTileTreeSet } from "./tile/internal";
 import { GraphicType } from "./common/render/GraphicType";
+import { SheetViewAttachments } from "./internal/SheetViewAttachments";
 
 // cSpell:ignore ovrs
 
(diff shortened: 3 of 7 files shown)


core/frontend/src/internal/render/webgl/glsl/Vertex.ts
```diff
@@ -344,9 +344,38 @@ const discardVertex = ` {
...
 `;
 
+const earlyVertexDiscard = `  if (checkForEarlyDiscard(rawPosition))${discardVertex}`;
+const vertexDiscard = `  if (checkForDiscard())${discardVertex}`;
+const lateVertexDiscard = `  if (checkForLateDiscard())${discardVertex}`;
+
+// Alternate approach to discard the vertex in the case of a driver that does not handle degenerate
+// triangles properly.
+const discardVertexInFrag = ` {
+    v_vertexDiscard = 1.0;
```
markschlosseratbentley: @pmconne In the workaround case, we are setting this varying to 1, which signifies to the fragment shader, "discard everything from this vertex / triangle."  However, we are not returning early from the main vertex shader, as is done in the non-workaround case.

The reason for this is because gl_Position is possibly not set at this time, and being undefined will still trigger a glitch.  Is returning important, in your view?  We could set gl_Position earlier in the shader in order to be able to return here.

In the case of early, mid, and late vertex discard, the return does avoid certain later shader code from executing.

core/frontend/src/test/render/webgl/VertexDiscard.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        builder.addPointString([pt]);
+        context.addDecorationFromBuilder(builder);
+      }
+    }
+  }
+
+  class TestOverrideProvider implements FeatureOverrideProvider {
+    constructor(private decorator: Decorator) { }
+    public addFeatureOverrides(overrides: FeatureSymbology.Overrides, _vp: Viewport): void {
+      overrides.setNeverDrawn(this.decorator.featureIds[0]);
```
markschlosseratbentley: @pmconne This unit test does not behave as I expect.

The unit test will report 3 unique pixels when nothing is overridden here as `setNeverDrawn`. When I override one of the feature IDs as `setNeverDrawn` as shown in the current code, the unit test will report 2 unique pixels, which indicates that the override is working.

However, when I alter the vertex discard shader such that it would "never discard", I still get two pixels reporting, not three.  Can you please advise if I am attempting to trigger the vertex discard in the way we discussed earlier?

Also if there is a way to verify this shader in display-test-app that would be very helpful to know about too.  There, I tried overriding transparency to 255 for certain elements in an iModel but they always disappear, even when I mess around with the vertex discard shader.

pmconne:
@markschlosseratbentley can  you take #8731 for a spin?

markschlosseratbentley:
> @markschlosseratbentley can you take #8731 for a spin?

It still glitches. 😞

markschlosseratbentley:
@pmconne Your experiment got me thinking more.

Based on your experiment, I began poking some other places. Now I theorize that the Intel glitch does **not** have to do with the degenerate triangles. Instead, I think it has to do with gl_Position not being initialized, and in some cases, the new Intel driver optimizes out a gl_Position assignment it thinks "does nothing." The driver must be optimizing out the degenerate triangle assignment, which means gl_Position is not assigned when we do a vertex discard and return early from the shader. It also must be optimizing your non-degenerate, outside-NDC triangle, which means gl_Position is also not assigned then.

I did another experiment. If I re-order the following in the vertex shader to occur **before** any vertex discard attempt (degenerate triangle or non-degenerate+outside NDC triangle), the glitches go away:

```ts
main.addline("  gl_Position = computePosition(rawPosition);");
// ... do vertex discard stuff.
```

However, I think this may mean that the triangles do not actually get discarded. Which brings us back to using a fragment shader discard and figuring out a way to test these shader changes are working as in https://github.com/iTwin/itwinjs-core/pull/8730#discussion_r2494371646.

pmconne:
> However, I think this may mean that the triangles do not actually get discarded.

Can you find out for certain?

markschlosseratbentley:
> > However, I think this may mean that the triangles do not actually get discarded.
> 
> Can you find out for certain?

I would like to. Do you have any feedback on how to go about that, as in https://github.com/iTwin/itwinjs-core/pull/8730#discussion_r2494371646?

pmconne:
> I would like to. Do you have any feedback on how to go about that, as in [#8730 (comment)](https://github.com/iTwin/itwinjs-core/pull/8730#discussion_r2494371646)?

Open 04_Plant. Select one of the cube-ish red "feet" at the bottom of the equipment. In Feature Overrides panel, set Transparency to 255 and click Apply to hide just that element.

markschlosseratbentley:
> > I would like to. Do you have any feedback on how to go about that, as in [#8730 (comment)](https://github.com/iTwin/itwinjs-core/pull/8730#discussion_r2494371646)?
> 
> Open 04_Plant. Select one of the cube-ish red "feet" at the bottom of the equipment. In Feature Overrides panel, set Transparency to 255 and click Apply to hide just that element.

The foot still vanishes even if I completely remove the code from `discardVertex`, like this: 

```ts
const discardVertex = ` {
    // gl_Position = vec4(2.0, 2.0, 2.0, 1.0); // outside NDC
    // return;
  }
`;
```

pmconne:
Hmm ok, open Revit.bim (or another iModel containing multiple categories but only a single spatial model) and try turning off one category.

markschlosseratbentley:
> Hmm ok, open Revit.bim (or another iModel containing multiple categories but only a single spatial model) and try turning off one category.

Thanks. Revit.bim allows me to turn off a single category and see the effect of the `discardVertex` shader chunk. Based on this one file at least, it appears the degenerate triangle is indeed not rendered, so long as gl_Position was initialized appropriately before setting it to a degenerate value. 🧐 

Based on this, the workaround may only need to involve ensuring gl_Position has a real value before setting it degenerate.

markschlosseratbentley:
> Thanks. Revit.bim allows me to turn off a single category and see the effect of the `discardVertex` shader chunk. 

Does this mean the unit test approach needs to change, since the transparency approach did not seem to actually exercise the shader?

pmconne:
> the workaround may only need to involve ensuring gl_Position has a real value before setting it degenerate.

If so we can probably just do that unconditionally and not have to check against known buggy GPU.

markschlosseratbentley:
> > the workaround may only need to involve ensuring gl_Position has a real value before setting it degenerate.
> 
> If so we can probably just do that unconditionally and not have to check against known buggy GPU.

To accomplish that, we need to call computePosition() earlier in the vertex shader, probably at the very top of it, which means the computePosition() calculation will not conditionally be circumvented in the case of these early discards.

Just initializing gl_Position earlier in the shader to create a triangle within NDC does not seem to be enough.

Are we okay with that potential extra calculation in all cases?

pmconne:
> To accomplish that, we need to call computePosition() earlier in the vertex shader, probably at the very top of it, which means it will not conditionally be circumvented in the case of these early discards.
> 
> Just setting gl_Position to create a triangle within NDC does not seem to be enough.
> 
> Are we okay with that potential extra calculation in all cases?

Experiment with, like passing a uniform initialized to the degenerate triangle and assigning that to gl_Position, to try and bypass the optimization you suspect of the buggy driver. (The driver may be smart enough to see through that, but worth a try). Or, try just setting it to (1,1,1,1) in the shader before proceeding normally.

markschlosseratbentley:
> Experiment with, like passing a uniform initialized to the degenerate triangle and assigning that to gl_Position, to try and bypass the optimization you suspect of the buggy driver. (The driver may be smart enough to see through that, but worth a try). Or, try just setting it to (1,1,1,1) in the shader before proceeding normally.

The suspected driver optimization sees through all of these attempts. The glitch persists unless we actually call `computePosition()` and assign it to `gl_Position` before then possibly assigning it to a degenerate triangle.

Attempted:
1. At top of vertex shader, initialize gl_Position to vec4(1.0).  Assign degenerate triangle as normal.
2. At top of vertex shader, initialize gl_Position to what I think is a legit triangle within NDC space. Assign degenerate triangle as normal.
3. At top of vertex shader, initialize gl_Position to the value of some vec4 uniform - vec4(1.0) or a vec4 indicating degenerate triangle.
4. In the `discardVertex` code, set gl_Position to degenerate triangle passed in via a uniform - vec4(2.0, 2.0, 2.0, 1.0);

core/frontend/src/internal/render/webgl/glsl/Vertex.ts
markschlosseratbentley: This concern does not apply to the latest iteration of this workaround. Resolving.

pmconne:
Did you report your observations about gl_Position to Intel? Ask them to suggest a workaround.

markschlosseratbentley:
> Did you report your observations about gl_Position to Intel? Ask them to suggest a workaround.

I will indeed engage asynchronously on the issue with them about this.  Not sure how soon we will get any suggestions.

Due to the urgency of this issue, for now, I modified this PR to conditionally move the gl_Position computePosition() assignment to the beginning of the shader if a bad chipset is detected.

core/frontend/src/test/render/webgl/VertexDiscard.test.ts
markschlosseratbentley: @pmconne, based on this thread (https://github.com/iTwin/itwinjs-core/pull/8730#issuecomment-3491540470), it seems overriding certain features as fully transparent will not necessarily trigger vertex discard. We had to find a .bim with a single model and turn off a single category to test vertex discard manually in DTA.

Does this mean that the decoration approach to this unit test is not workable?  My testing indicates that this unit test suffers from the same problem as what we initially tried in DTA.  Is there another unit test approach we could consider that you can think of?

I am okay removing this unit test if there is not a good way to exercise vertex discard.  This unit test will not run on Intel hardware at any rate, unless run locally.  Conversely, it still could be good to have some tests running through that bit of code.

Otherwise, the workaround at least has a unit test in `Compatibility.test.ts`.

markschlosseratbentley: Actually, now that the workaround does not even modify the vertex discard code directly, I think this unit test probably does not serve a useful purpose. I vote for removing it. Resolving my own conversation. Please feel free to reopen the conversation or start a new one on this topic.

alarkbentley:
@markschlosseratbentley @pmconne Nice work tracking this one down! 
Once we have this fix ready and tested, do we plan to backport to iTwin.js 4.x or will iTwin.js 5.x be a requirement?

markschlosseratbentley:
> @markschlosseratbentley @pmconne Nice work tracking this one down! Once we have this fix ready and tested, do we plan to backport to iTwin.js 4.x or will iTwin.js 5.x be a requirement?

@aruniverse?

aruniverse:
> @markschlosseratbentley @pmconne Nice work tracking this one down! Once we have this fix ready and tested, do we plan to backport to iTwin.js 4.x or will iTwin.js 5.x be a requirement?

We will bp to 5.x for sure. 4.x as needed. 
Please reach out offline to discuss the product and client this is for

markschlosseratbentley:
FYI @eringram is running some performance testing on master and [workaround-driver-intel-driver-ultra-7-no-conditions](https://github.com/iTwin/itwinjs-core/tree/workaround-intel-driver-ultra-7-no-conditions) branch to see if we need to conditionally do the workaround to avoid a performance hit, or if we can just do the workaround all the time with minimal hit.

markschlosseratbentley:
Running Image Tests.
- [x] Image Tests results look okay.

markschlosseratbentley:
> Running Image Tests.

Image Tests are showing differences with the workaround.

Seems to be centered around line width not being applied.

Analyzing the `Polyline-Opaque-Overrides_VS` shader, I believe computeFeatureOverrides() needs to be called before computePosition().  computeFeatureOverrides() sets `linear_feature_overrides` in computeFeatureOverrides() and then uses that in computeLineWidth().

markschlosseratbentley:
> Analyzing the `Polyline-Opaque-Overrides_VS` shader, I believe computeFeatureOverrides() needs to be called before computePosition(). computeFeatureOverrides() sets `linear_feature_overrides` in computeFeatureOverrides() and then uses that in computeLineWidth().

I moved computePosition() to occur before the second possible discard and after feature overrides are computed.  This still works around the glitch and allows wide lines to still function properly.

markschlosseratbentley:
Image Tests results look good.

aruniverse:

aruniverse:

aruniverse:
@mergifyio backport release/5.2.x release/5.3.x

PR #8747, opened 2025-11-06
Listen on localhost only instead of all interfaces for LocalhostIpcHost

fjch1997:
This prevents the Windows Firewall dialog and fixes security vulnerability of exposing IPC server to the network.

### DIFF
--- a/common/changes/@itwin/core-backend/2025-11-12-22-47.json
+++ b/common/changes/@itwin/core-backend/2025-11-12-22-47.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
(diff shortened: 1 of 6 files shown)


aruniverse:
fyi @kckst8 @johnnyd710 @ImVeryLost

markschlosseratbentley:
> Seems fine to me. We only use `LocalhostIpcHost` internally (DTA), correct?

Just to be safe, have we verified local display-test-app works okay with this change?

fjch1997:
> > Seems fine to me. We only use `LocalhostIpcHost` internally (DTA), correct?
> 
> Just to be safe, have we verified local display-test-app works okay with this change?

Yes, display-test-app works.

ben-polinsky:

PR #8748, opened 2025-11-10
SchemaJsonLocater error handling

StefanApfel-Bentley:
#### Description
SchemaJsonLocater uses a callback passed in the constructor to fetch the SchemaProps for a given schema name. In case this callback throws an error (which it is allowed to do by the callbacks docs) the locater fails on getSchema/getSchemaSync as well. This shouldn't be the case though the definition of the ISchemaLocater interface where those two methods shall return undefined if the requested schema can't be loaded by the implementing locater.

Fix is swallowing failing getSchemaProps operations, but log an error, to be able to track those.

### DIFF
--- a/common/api/ecschema-metadata.api.md
+++ b/common/api/ecschema-metadata.api.md
@@ -2117,7 +2117,7 @@ export interface SchemaItemUnitProps extends SchemaItemProps {
 
 // @public @preview
 export class SchemaJsonLocater implements ISchemaLocater {
-    constructor(schemaPropschemaGetter: SchemaPropsGetter);
+    constructor(getSchemaProps: SchemaPropsGetter);
     getSchema(schemaKey: SchemaKey, matchType: SchemaMatchType, context: SchemaContext): Promise<Schema | undefined>;
     getSchemaInfo(schemaKey: SchemaKey, matchType: SchemaMatchType, context: SchemaContext): Promise<SchemaInfo | undefined>;
     getSchemaSync(schemaKey: SchemaKey, _matchType: SchemaMatchType, context: SchemaContext): Schema | undefined;


core/ecschema-metadata/src/SchemaJsonLocater.ts
```diff
@@ -23,7 +24,21 @@ export type SchemaPropsGetter = (schemaName: string) => SchemaProps | undefined;
...
+
+  public constructor(schemaPropschemaGetter: SchemaPropsGetter) {
+    // Since the schemaPropschemaGetter may throw an error, but the locater contract defines that
+    // getSchema should return undefined if the schema could not be located, we wrap the provided
+    // lookup function in a safe block and log any errors that occured.
+    this._getSchema = (schemaName) => {
+      try {
+        return schemaPropschemaGetter(schemaName);
+      } catch (error: any) {
+        Logger.logError("SchemaJsonLocater", `Failed to get schema JSON for schema ${schemaName}: ${error.message}`);
```
rschili: I vote for not logging an error here. This can and will happen in normal program flow. Writing errors to the logs will probably just confuse consumers.
Ideally, the schemaPropSchemaGetter would only throw errors on actual errors, and return undefined otherwise, then we can remove this block. Today it could be both, so either we check which case it is and only log/rethrow real errors, or not log at all.

core/ecschema-metadata/src/SchemaJsonLocater.ts
```diff
@@ -23,7 +24,21 @@ export type SchemaPropsGetter = (schemaName: string) => SchemaProps | undefined;
  * @public @preview
  */
 export class SchemaJsonLocater implements ISchemaLocater {
-  public constructor(private _getSchema: SchemaPropsGetter) { }
+  private _getSchema: SchemaPropsGetter;
+
+  public constructor(schemaPropschemaGetter: SchemaPropsGetter) {
```
rschili: Something seems off with that name

rschili (approved):

core/ecschema-metadata/src/SchemaJsonLocater.ts
StefanApfel-Bentley: yeah, a bit too much schema in it, I changed the argument name to getSchemaProps

core/ecschema-metadata/src/SchemaJsonLocater.ts
StefanApfel-Bentley: I removed the logging here. As I started in the linked issue, we might need to revisit our locater design anyway.

Issue #8751, opened 2025-11-10
Flakey Test: handles schema import during content request


DanRod1999:
It looks like `handles schema import during content request` has been timing out rush cover in windows and linux. Build does not show logs clearly that test is failing, but appears to stop right after the previous test, so I believe it is hanging during this test. So far I've only seen it on windows and linux.
[Example](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4481776&view=logs&j=f11376b5-c0e7-5865-284f-2bac38e275ef&t=b55417c9-5862-590f-4bb2-5070cd265aab&s=27eb50c5-b53b-5455-6fbd-7f2aeb0e607bl)

Introduced in [PR](https://github.com/iTwin/itwinjs-core/pull/8733)

grigasp:
https://github.com/iTwin/itwinjs-core/pull/8745#issuecomment-3512996399

grigasp:
I managed to reproduce locally, and it looks like the deadlock is in ECDb code:
<img width="504" height="827" alt="Image" src="https://github.com/user-attachments/assets/31ebe120-4636-41e1-ae80-1dfb5d54baf0" />

- **Thread 46420** holds ECDb's mutex in `MainSchemaManager::ImportSchemas` and is waiting for workers to complete.
- **Thread 44876** is one of the workers that **Thread 46420** is waiting for and is hung trying to acquire `ecdb.GetImpl().GetMutex()`.

Repro steps on Windows:
1. Checkout `master` branch of `itwinjs-core`
2. Add `.only` on [this test](https://github.com/iTwin/itwinjs-core/blob/1ab8a3555dde8d66c1e2540384dda9d58e564db8/full-stack-tests/presentation/src/backend/ReadWrite.test.ts#L86)
3. `rush install`
4. `rush build -t presentation-full-stack-tests`
5. `cd full-stack-tests\presentation`
6. `.\run-forever.bat`

It's helpful to execute the `run-forever.bat` in multiple terminal windows at the same time (I ran it in parallel in 4 terminals). For me it takes ~20-30 runs until one of the terminals hangs.

@khanaffan, can you have a look? I also have a dump locally if you can't reproduce it - ping me if you need it.

PR #8758, opened 2025-11-12
Presentation: Fix flaky full stack test

grigasp:
~Fixes https://github.com/iTwin/itwinjs-core/issues/8751.~

Couldn't reproduce locally, but my assumption is that when the first of two tests finishes and "deletes" the test imodel, the second test immediately tries to create a new imodel at the same file path, and FS has that file path locked, causing a hang. The change ensures that we use unique file names for each of the tests.

Update: The above assumption was wrong.

### DIFF
--- a/full-stack-tests/presentation/src/backend/ReadWrite.test.ts
+++ b/full-stack-tests/presentation/src/backend/ReadWrite.test.ts
@@ -10,14 +10,14 @@ import { PresentationManager } from "@itwin/presentation-backend";
 import { ChildNodeSpecificationTypes, Ruleset, RuleTypes } from "@itwin/presentation-common";
 import { initialize, terminate } from "../IntegrationTests.js";
 import { collect, prepareOutputFilePath } from "../Utils.js";
+import { createValidIModelFileName } from "../IModelSetupUtils.js";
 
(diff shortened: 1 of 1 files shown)


grigasp:
I was wrong...

```
  ReadWrite
[2025-11-12T15:38:17.005Z] Tests initialized
    Handling read-write operations
Info | BeSQLite | Opening DgnDb out\8008\readwrite-before-each-hook-for-handles-schema-import-during-nodes-request.bim 
Info | BeSQLite | Created file 'C:\agent\_work\1\s\full-stack-tests\presentation\out\8008\readwrite-before-each-hook-for-handles-schema-import-during-nodes-request.bim-hierarchies' 
Warning | core-backend.IModelDb | there are no unsaved changes 
      √ handles schema import during nodes request (897ms)
Warning | core-backend.IModelDb | there are no unsaved changes 
Info | BeSQLite | Opening DgnDb out\8008\readwrite-before-each-hook-for-handles-schema-import-during-content-request.bim 
##[error]The Operation will be canceled. The next steps may not contain expected logs.
```

will continue looking into this

grigasp:
Closing as the changes in this PR were based under a false assumption and don't solve the problem.

PR #8769, opened 2025-11-13
Fix carry over logic for fractional formatting

hl662:
Addresses a bug in the fractional formatting logic for composite units in the `@itwin/core-quantity` package. This ensures that rounding of fractional components correctly carries over to the parent unit when appropriate, preventing formatting errors in edge cases. Comprehensive tests have been added to verify the new behavior for both fractional and decimal formats.

**Bug fix in fractional formatting logic:**

* Fixed the carry-over logic in `Formatter.formatQuantity` so that when the fractional part of a unit rounds up to a full unit in the next smaller component (e.g., inches rounding up to feet), the parent unit is incremented appropriately. This prevents cases like "11.9999999 ft" from incorrectly displaying as "11'-12\"" instead of "12'-0\"".

**Testing and validation:**

* Added extensive test cases to `CompositeFormats.test.ts` to verify correct carry-over behavior for fractional formats, including positive, negative, and edge values for both feet/inches and degrees/minutes/seconds.
* Added tests to ensure that decimal formats do not use the fractional carry-over logic, maintaining their expected behavior.

### DIFF
--- a/common/changes/@itwin/analytical-backend/master_2025-11-13-20-52.json
+++ b/common/changes/@itwin/analytical-backend/master_2025-11-13-20-52.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "",
+      "type": "none",
+      "packageName": "@itwin/analytical-backend"
+    }
+  ],
+  "packageName": "@itwin/analytical-backend",
+  "email": "imodeljs-admin@users.noreply.github.com"
+}
--- a/common/changes/@itwin/analytical-backend/master_2025-11-13-21-16.json
+++ b/common/changes/@itwin/analytical-backend/master_2025-11-13-21-16.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "",
+      "type": "none",
+      "packageName": "@itwin/analytical-backend"
+    }
+  ],
+  "packageName": "@itwin/analytical-backend",
+  "email": "imodeljs-admin@users.noreply.github.com"
+}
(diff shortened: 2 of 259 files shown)


core/quantity/src/Formatter/Formatter.ts
```diff
@@ -244,7 +244,30 @@ export class Formatter {
...
+        let wholePart = Math.trunc(unitValue);
+
+        // Check if the remaining fractional part will round up to a full unit in the next (smaller) component
+        if (spec.format.type === FormatType.Fractional && i === spec.unitConversions.length - 2) {
+          // For the second-to-last unit with fractional formatting, check if rounding causes carry-over
+          const fractionalPart = unitValue - wholePart;
+          const nextUnitValue = applyConversion(fractionalPart, spec.unitConversions[i + 1].conversion);
+
+          // Create a FractionalNumeric to determine what the rounded value would be
+          const fn = new FractionalNumeric(Math.abs(nextUnitValue), spec.format.precision as FractionalPrecision, true);
```
ben-polinsky, line 256: This cast is jumping out. Why are you sure the precision is FractionalPrecision at this point?

hl662: 2 things : 
1. By this point in `Formatter`, the instantiated Format has already validated the precision value through [parsePrecision](https://github.com/iTwin/itwinjs-core/blob/nam/qty-rounding-bug/core/quantity/src/Formatter/Format.ts#L150-L151), which validates whether the format type is fractional or non fractional.
2. We check above if the spec format type is `Fractional` on top of the first point as well. This follows the same logic Bill added years ago [in another place he used casting](https://github.com/iTwin/itwinjs-core/blob/nam/qty-rounding-bug/core/quantity/src/Formatter/Formatter.ts#L367)

ben-polinsky (approved):

rbreukelaar (approved):
Yes, this looks good, and thorough.

PR #8775, opened 2025-11-14
Improved CurveCurveIntersectXY

saeeedtorabi:
This PR is motivated by a deficiency in the initial [PR](https://github.com/iTwin/itwinjs-core/pull/8658) whereby a tangency intersection is not found because the spiral's linestring approximation does not intersect the other curve, resulting in a failure to compute a Newton seed that refines to the tangency.


core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1057,7 +1058,8 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
       const spiralApproximation = LineString3d.create();
       this._geometryB.emitStrokes(spiralApproximation);
       const numPreviousResults = this._results.length;
-      this.computeSegmentLineString(segmentA, this._extendA0, this._extendA1, spiralApproximation, false, false, false);
+      // this.computeSegmentLineString(segmentA, this._extendA0, this._extendA1, spiralApproximation, false, false, false);
```
saeeedtorabi: I will remove the commented lines

dassaf4 (approved):
Made some changes. Please take a look.

There still seems to be an issue with visualizing DirectSpiral3d curves transformed out of xy-plane (in our test environment). I'm not sure this is due to how we serialize the data in iTwin, or how we deserialize it in native code, or how we process the deserialized object to view it.

Forgot to mention in commit messages: aside from removing dead code in closeApproachXY, I also made sure its output details store the projection distance in the `a` field. This is so we can sort on that in the intersectXY class.

Aside: the new logic for computing spiral intersectionXY via its strokes is applicable to any CurvePrimitive for which we can compute a maximum stroke error. For spirals, this computation is simple, but for a general curve, it could be expensive.

saeeedtorabi:
> There still seems to be an issue with visualizing DirectSpiral3d curves transformed out of xy-plane (in our test environment). I'm not sure this is due to how we serialize the data in iTwin, or how we deserialize it in native code, or how we process the deserialized object to view it.

Created an issue for it: https://github.com/iTwin/itwinjs-backlog/issues/1774

> Aside from removing dead code in closeApproachXY, I also made sure its output details store the projection distance in the `a` field. This is so we can sort on that in the intersectXY class.

What is the benefit of sorting close approaches? If we are going to pick first few ones and drop the rest, sorting makes sense but the code is using all of the close approaches so I don't understand why we have to pay for sorting.

FYI, I reviewed the rest of the code and all made sense (except sorting). I will merge this and modify my close approach PR based on your changes.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1099,104 +1116,228 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
+  private appendDiscreteCloseApproachResults(
+    curveA: CurvePrimitive, lsB: LineString3d, maxDistance: number, reversed: boolean,
+  ): number {
+    const i0 = this._results.length;
+    const closeApproachPairs = CurveCurve.closeApproachProjectedXYPairs(
+      reversed ? lsB : curveA, reversed ? curveA : lsB, maxDistance,
+    );
+    closeApproachPairs.sort((p0: CurveLocationDetailPair, p1: CurveLocationDetailPair) => p0.detailA.a - p1.detailA.a);
+    this._results.push(...closeApproachPairs);
+    return this._results.length - i0;
```
saeeedtorabi, line 1309: why do you return this number? The return value is never used.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1099,104 +1116,228 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
+    this.resetGeometryA(false, false); // lsB is never extended
+    this.resetGeometryB(curveA, extendA0, extendA1);
+    this.handleLineString3d(lsB); // this puts lsB data in detailA, as expected when reversed is true
+    if (!reversed) { // swap lsB data to detailB
+      for (let i = i0; i < this._results.length; i++)
+        this._results[i].swapDetails();
+    }
+    this.resetGeometryA(extendA0, extendA1);
+    this.resetGeometryB(geomB, extendB0, extendB1);
+    return this._results.length - i0;
```
saeeedtorabi, line 1293: why do you return this number? The return value is never used.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
dassaf4: It seemed useful to potential callers and is more informative than a boolean. Doesn't hurt to keep.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
dassaf4: ditto previous

dassaf4:
> Created an issue for it: [iTwin/itwinjs-backlog#1774](https://github.com/iTwin/itwinjs-backlog/issues/1774)

Thanks!

> > Aside from removing dead code in closeApproachXY, I also made sure its output details store the projection distance in the `a` field. This is so we can sort on that in the intersectXY class.
> 
> What is the benefit of sorting close approaches? If we are going to pick first few ones and drop the rest, sorting makes sense but the code is using all of the close approaches so I don't understand why we have to pay for sorting.

Sorting closeApproachXY output is left to the caller, so nothing changed in that class save the data we store (we now store projection distance). One caller who cares about sorting this output is intersectXY, who wants to ensure that the closer seeds it computes via the discrete closeApproachXY problem are processed first. Why is this important? Because intersectXY tries to return the _most accurate_ solution for the i_th intersection. (This is a different constraint than closeApproachXY, which returns all solutions within maxDistance without filtering.)

We don't know a priori which set of seeds will converge to points nearby each other, much less which of those is most accurate. Denote by B_i the set of seeds converging close to the i_th intersection (this is the Newton basin of attraction for that intersection point). Generally, I found that the seeds computed from the discrete intersectXY problem yield the most accurate solutions via Newton (these are the quadratic-convergence cases). Hence we process these seeds first.

But what about the other seeds in B_i, those computed from the discrete closeApproachXY problem? My investigations showed that especially in the sub-quadratic tangency cases where we don't have seeds generated from the discrete intersectXY problem, the (discrete closeApproachXY) seed with smallest projection distance converges to the most accurate solution.

Since the SortedArray we use for filtering stores only the first seed of B_i we process, it's important to process 1) all discrete intersectXY seeds first, then 2) discrete closeApproachXY seeds ordered by increasing projection distance. This gives the best chances that the most accurate solution for the i_th intersection is returned.

PR #8779, opened 2025-11-18
Remove extra MSAA filter

### LINKED ISSUE #8777: Viewport does not display sky background
**Describe the bug**
Render plan does not seem to account for displaySky when setting background color.

**To Reproduce**

1. Open any 3d imodel in pineapple
2. In status bar, click on Presentation widget and toggle enable the Sky background.
3. Note that the viewport shows the color set for Solid background.

**Expected behavior**
The Sky color should be displayed when Sky is toggled on.

**Screenshots**

Image

- Browser: any
- iTwin.js Version: 5.3.0

**Additional context**
This worked in iTwin.js 5.2.0

With the following patch in RenderPlan.ts, this issue is resolved (but maybe not the correct solution):

export function createRenderPlanFromViewport(vp) {
...
    // Determine background color based on display settings
    let bgColor = view.backgroundColor;
    if (style.is3d()) {
      const env = view.getDisplayStyle3d().environment;

      // If sky is displayed, use the sky's nadir color as the background
      if (env.displaySky) {
        bgColor = env.sky.gradient.zenithColor;
      } else if (backgroundMapOn) {
        // If background map is on and base is a solid color, use that color
        const mapBase = style.settings.mapImagery.backgroundBase;
        if (mapBase instanceof ColorDef)
          bgColor = mapBase;
      }
    }

pmconne:
Fixes #8777. Not really testable because the graphical issue appears to be hardware/driver-dependent.


aruniverse:

PR #8780, opened 2025-11-18
Address GHSA-5j98-mcp5-4vw2

hl662:
Address GHSA-5j98-mcp5-4vw2

Bumps direct deps to glob up to `^10.5.0`, transitive dep in cpx that requires a scoped override in `pnpmConfig.json`

Also removed @types/glob from a couple pkgs, [glob comes with types now](https://www.npmjs.com/package/@types/glob)

### DIFF
--- a/common/config/rush/pnpm-config.json
+++ b/common/config/rush/pnpm-config.json
@@ -14,7 +14,7 @@
     "elliptic": "^6.6.1", // https://github.com/advisories/GHSA-vjh7-7g9h-fjfh crypto-browserify>browserify-sign>elliptic
     "form-data": "^4.0.4", // https://github.com/advisories/GHSA-fjxv-7rqg-78g4 azurite>@azure/ms-rest-js>form-data
     "browserslist": "latest", // https://github.com/browserslist/update-db#readme,
-    "glob@>=11.0.0 <11.1.0": "^11.1.0" // https://github.com/advisories/GHSA-5j98-mcp5-4vw2
+    "glob@>=11.0.0 <11.1.0": "^11.1.0" // https://github.com/advisories/GHSA-5j98-mcp5-4vw2 cpx2>glob
   },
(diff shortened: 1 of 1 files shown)


hl662:

common/config/rush/pnpm-config.json
```diff
@@ -13,6 +13,7 @@
     "rollup-plugin-copy>globby": "^11.0.0", // https://github.com/vladshcherbin/rollup-plugin-copy/issues/77
     "elliptic": "^6.6.1", // https://github.com/advisories/GHSA-vjh7-7g9h-fjfh crypto-browserify>browserify-sign>elliptic
     "form-data": "^4.0.4", // https://github.com/advisories/GHSA-fjxv-7rqg-78g4 azurite>@azure/ms-rest-js>form-data
+    "glob": "^10.5.0", // https://github.com/advisories/GHSA-5j98-mcp5-4vw2 cpx2>glob, vitest/coverage-v8>glob
```
GytisCepk: How many packages require override ? If it's only a few, I would suggest adding "scoped" overrides, since our dependencies seem to be using a variety of `glob` versions, include a newer 11

hl662: true, digging deeper, I found that cpx2 uses glob v11, so this would effectively downgrade things. For the transitive dependencies I just needed to run `rush update --full`, so I'm removing the override from pnpmConfig.

I can't update all our dependencies to glob v11 for this though, because glob v11 drops support for node <20, so that won't be nice to our 4.x backport (since we still support node 18 for iTwin 4) . So, keeping glob to v10.5.0 for our direct deps to glob

hl662: nevermind I learned the hard way that rush update full bumps eslint-parser up several minors, introducing new lint rules and fixes old rules (that we skitted by) and that threw everything up. So, adding scoped override for cpx2 and for v11 globs.

PR #8785, opened 2025-11-18
Take RealityTile.transformToRoot into account in loadGeometryFromStream

eringram:
There is a discrepancy between the collected geometry result of `GltfReader.readGltfAndCreateGeometry()`, and the rendered graphics from `GltfReader.readGltfAndCreateGraphics()`.

The cause was is that both methods take a `transformToRoot` argument, which is a transform that is applied to the result. The transform provided to `readGltfAndCreateGraphics` is a `RealityTile.transformToRoot`. On the other hand, the transform provided to `readGltfAndCreateGeometry` is a `RealityTile.tree.iModelTransform`. This means that if a tile had a `transformToRoot` defined, that was likely not taken into account when `readGltfAndCreateGeometry` was called.

My solution in this PR is to multiply the `RealityTile.tree.iModelTransform` and `RealityTile.transformToRoot` to provide one transform to `readGltfAndCreateGeometry`. With this fix, the TerrainDrapeTool in DTA collects geometry at the same height as the rendered geometry.

TODO
- [x] Manually test in display-test-app
- [x] Those who reported the bug also manually tested this fix
- [x] Run image tests

Unit tests will be added in a follow up-PR: https://github.com/iTwin/itwinjs-core/pull/8809

### DIFF
--- a/common/changes/@itwin/frontend-devtools/nelrod-viewdefdec_2025-11-24-15-00.json
+++ b/common/changes/@itwin/frontend-devtools/nelrod-viewdefdec_2025-11-24-15-00.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/frontend-devtools",
+      "comment": "Added Decorator to show bounds of the view's original Frustum",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/frontend-devtools"
+}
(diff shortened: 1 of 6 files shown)


eringram:
@aruniverse is there still time for this to go into the 5.4 release?

aruniverse:
> @aruniverse is there still time for this to go into the 5.4 release?

No, it will either need to go in as a patch or 5.5

aruniverse:
@mergifyio backport release/5.3.x release/5.4.x

PR #8787, opened 2025-11-19
Extend ratio formatting to support scale factors

hl662:
This pull request adds comprehensive support for ratio formatting with scale factors to the `@itwin/core-quantity` and `@itwin/ecschema-metadata` packages. The changes introduce new properties and types for ratio formats, enhance format parsing and validation, and update schema deserialization and format selection logic to handle ratio units.

**Ratio Formatting Enhancements:**

* Added new properties to `Format`, `BaseFormat`, and related interfaces/classes to support `ratioFormatType`, `ratioSeparator`, `ratioType`, enabling flexible definition and parsing of ratio formats. 
* Updated composite unit handling, such that when the formatType is `Ratio`, and two composite units are provided, a conversion factor is calculated based on those two units.
* Introduced the `RatioFormatType` enum and `FormatUnitSpec`/`ResolvedFormatUnitSpec` interfaces to avoid inline types.
* Added the `parseRatioFormatType` function for parsing ratio format types from strings. 

**Schema and Format Parsing/Validation:**

* Updated JSON and XML deserialization logic to validate and load new ratio format properties for EC3.3 schemas and later.
* Enhanced the `Format` class in `ecschema-metadata` to provide accessors for the new ratio properties
* Updated `FormatterSpec` and `ParserSpec` in `quantity` to support on the fly unit conversion calculations when two unit composite format is provided.

### DIFF
--- a/common/config/rush/version-policies.json
+++ b/common/config/rush/version-policies.json
@@ -2,7 +2,7 @@
   {
     "policyName": "prerelease-monorepo-lockStep",
     "definitionName": "lockStepVersion",
-    "version": "5.6.0-dev.12",
+    "version": "5.6.0-dev.13",
     "nextBump": "prerelease"
   }
 ]
--- a/core/backend/package.json
+++ b/core/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-backend",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js backend components",
   "main": "lib/cjs/core-backend.js",
   "module": "lib/esm/core-backend.js",
--- a/core/bentley/package.json
+++ b/core/bentley/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-bentley",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "Bentley JavaScript core components",
   "main": "lib/cjs/core-bentley.js",
   "module": "lib/esm/core-bentley.js",
--- a/core/common/package.json
+++ b/core/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-common",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js components common to frontend and backend",
   "main": "lib/cjs/core-common.js",
   "module": "lib/esm/core-common.js",
--- a/core/ecschema-editing/package.json
+++ b/core/ecschema-editing/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-editing",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "ECSchema editing and validation API",
   "license": "MIT",
   "main": "lib/cjs/ecschema-editing.js",
--- a/core/ecschema-locaters/package.json
+++ b/core/ecschema-locaters/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-locaters",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "EC Schema file locaters",
   "license": "MIT",
   "main": "lib/cjs/ecschema-locaters.js",
--- a/core/ecschema-metadata/package.json
+++ b/core/ecschema-metadata/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-metadata",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "ECObjects core concepts in typescript",
   "license": "MIT",
   "main": "lib/cjs/ecschema-metadata.js",
--- a/core/ecschema-rpc/common/package.json
+++ b/core/ecschema-rpc/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-rpcinterface-common",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "Schema RPC Interface common interface",
   "main": "lib/cjs/ecschema-rpc-interface.js",
   "module": "lib/esm/ecschema-rpc-interface.js",
--- a/core/ecschema-rpc/impl/package.json
+++ b/core/ecschema-rpc/impl/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-rpcinterface-impl",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "Schema RPC Interface backend implementation",
   "main": "lib/cjs/ecschema-rpc-impl.js",
   "module": "lib/esm/ecschema-rpc-impl.js",
--- a/core/ecsql/common/package.json
+++ b/core/ecsql/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecsql-common",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "ECSql component that can be reference on backend and frontend",
   "main": "lib/cjs/ecsql-common.js",
   "module": "lib/esm/ecsql-common.js",
--- a/core/electron/package.json
+++ b/core/electron/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-electron",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js ElectronHost and ElectronApp",
   "license": "MIT",
   "scripts": {
--- a/core/express-server/package.json
+++ b/core/express-server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/express-server",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js express utilities",
   "main": "lib/cjs/express-server.js",
   "module": "lib/esm/express-server.js",
--- a/core/extension/package.json
+++ b/core/extension/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-extension",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js Extension API",
   "type": "module",
   "typings": "index.d.ts",
--- a/core/frontend-devtools/package.json
+++ b/core/frontend-devtools/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/frontend-devtools",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "Debug menu and supporting UI widgets",
   "main": "lib/cjs/frontend-devtools.js",
   "module": "lib/esm/frontend-devtools.js",
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-frontend",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js frontend components",
   "main": "lib/cjs/core-frontend.js",
   "module": "lib/esm/core-frontend.js",
--- a/core/geometry/package.json
+++ b/core/geometry/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-geometry",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js Core Geometry library",
   "main": "lib/cjs/core-geometry.js",
   "module": "lib/esm/core-geometry.js",
--- a/core/hypermodeling/package.json
+++ b/core/hypermodeling/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/hypermodeling-frontend",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js hypermodeling package",
   "main": "lib/cjs/hypermodeling-frontend.js",
   "module": "lib/esm/hypermodeling-frontend.js",
--- a/core/i18n/package.json
+++ b/core/i18n/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-i18n",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js localization code",
   "main": "lib/cjs/core-i18n.js",
   "module": "lib/esm/core-i18n.js",
--- a/core/markup/package.json
+++ b/core/markup/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-markup",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js markup package",
   "main": "lib/cjs/core-markup.js",
   "module": "lib/esm/core-markup.js",
--- a/core/mobile/package.json
+++ b/core/mobile/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-mobile",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js MobileHost and MobileApp",
   "license": "MIT",
   "engines": {
--- a/core/orbitgt/package.json
+++ b/core/orbitgt/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-orbitgt",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "",
   "main": "lib/cjs/core-orbitgt.js",
   "module": "lib/esm/core-orbitgt.js",
--- a/core/quantity/package.json
+++ b/core/quantity/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-quantity",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "Quantity parsing, formatting and conversions for iModel.js",
   "main": "lib/cjs/core-quantity.js",
   "module": "lib/esm/core-quantity.js",
--- a/core/webgl-compatibility/package.json
+++ b/core/webgl-compatibility/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/webgl-compatibility",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "APIs for determining the level of compatibility of a browser+device with the iTwin.js rendering system.",
   "license": "MIT",
   "main": "lib/cjs/webgl-compatibility.js",
--- a/domains/analytical/backend/package.json
+++ b/domains/analytical/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/analytical-backend",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "main": "lib/cjs/analytical-backend.js",
   "module": "lib/esm/analytical-backend.js",
   "typings": "lib/cjs/analytical-backend",
--- a/domains/linear-referencing/backend/package.json
+++ b/domains/linear-referencing/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/linear-referencing-backend",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "main": "lib/cjs/linear-referencing-backend.js",
   "module": "lib/esm/linear-referencing-backend.js",
   "typings": "lib/cjs/linear-referencing-backend",
--- a/domains/linear-referencing/common/package.json
+++ b/domains/linear-referencing/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/linear-referencing-common",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "main": "lib/cjs/linear-referencing-common.js",
   "module": "lib/esm/linear-referencing-common.js",
   "typings": "lib/cjs/linear-referencing-common",
--- a/domains/physical-material/backend/package.json
+++ b/domains/physical-material/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/physical-material-backend",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "main": "lib/cjs/physical-material-backend.js",
   "module": "lib/esm/physical-material-backend.js",
   "typings": "lib/cjs/physical-material-backend",
--- a/editor/backend/package.json
+++ b/editor/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/editor-backend",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
   "description": "iTwin.js editor backend",
   "main": "lib/cjs/editor-backend.js",
   "module": "lib/esm/editor-backend.js",
--- a/editor/common/package.json
+++ b/editor/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/editor-common",
-  "version": "5.6.0-dev.12",
+  "version": "5.6.0-dev.13",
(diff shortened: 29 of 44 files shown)


core/quantity/src/Formatter/Format.ts
```diff
@@ -41,6 +42,8 @@ export class BaseFormat {
   protected _stationOffsetSize?: number; // required when type is station; positive integer > 0
   protected _stationBaseFactor?: number; // optional positive integer base factor for station formatting; default is 1
   protected _ratioType?: RatioType; // required if type is ratio; options: oneToN, NToOne, ValueBased, useGreatestCommonDivisor
+  protected _ratioFormatType?: RatioFormatType; // optional, defaults to Decimal if not specified
```
ben-polinsky: Nit: Don't see a reason to specify "optional" in the comments: it's easy to tell from the `?` and at times you have properties which are optional but do not have the comment that they are optional.

core/quantity/src/Formatter/Format.ts
```diff
@@ -161,6 +170,22 @@ export class BaseFormat {
         throw new QuantityError(QuantityStatus.InvalidJson, `The Format ${this.name} is 'Ratio' type therefore the attribute 'ratioType' is required.`);
 
       this._ratioType = parseRatioType(formatProps.ratioType, this.name);
+
+      if (undefined !== formatProps.ratioSeparator) {
```
ben-polinsky, line 174: nit: yoda conditions here, but below non-yoda/traditional conditions - I would choose one and stick with it.

core/quantity/src/Formatter/Format.ts
```diff
@@ -390,6 +426,16 @@ export class Format extends BaseFormat {
         throw new QuantityError(QuantityStatus.InvalidJson, `The Format ${this.name} has a Composite with no valid 'units'`);
     }
 
+    // Process ratioUnits if provided (for scale factor conversion and/or unit labels)
+    if (jsonObj.ratioUnits && jsonObj.ratioUnits.length === 2) {
```
ben-polinsky:
```suggestion
    if (jsonObj.ratioUnits?.length === 2) {
```

core/ecschema-metadata/src/Metadata/Format.ts
```diff
@@ -107,43 +130,80 @@ export class Format extends SchemaItem {
...
+
+        if (Unit.isUnit(newUnit))
+          this.addUnit(new DelayedPromiseWithProps(newUnit.key, async () => newUnit), unit.label);
+        else if (InvertedUnit.isInvertedUnit(newUnit))
+          this.addUnit(new DelayedPromiseWithProps(newUnit.key, async () => newUnit), unit.label);
+      }
+    }
+
+    // Process ratioUnits
+    if (undefined !== formatProps.ratioUnits && formatProps.ratioUnits.length === 2) {
```
ben-polinsky: This looks like a very similar, if not exactly the same block as where you process ratioUnits below lines 196-206... please extract?

core/ecschema-metadata/src/Metadata/Format.ts
```diff
@@ -107,43 +130,80 @@ export class Format extends SchemaItem {
...
+      for (const unitSpec of formatProps.ratioUnits) {
+        const newUnit = this.schema.lookupItemSync(unitSpec.name);
+        if (undefined === newUnit || (!Unit.isUnit(newUnit) && !InvertedUnit.isInvertedUnit(newUnit)))
+          throw new ECSchemaError(ECSchemaStatus.InvalidECJson, `The Format ${this.fullName} has an invalid ratioUnit '${unitSpec.name}'.`);
+
+        if (Unit.isUnit(newUnit))
+          this.addRatioUnit(new DelayedPromiseWithProps(newUnit.key, async () => newUnit), unitSpec.label);
+        else if (InvertedUnit.isInvertedUnit(newUnit))
+          this.addRatioUnit(new DelayedPromiseWithProps(newUnit.key, async () => newUnit), unitSpec.label);
```
ben-polinsky:
```suggestion
        if (Unit.isUnit(newUnit) || InvertedUnit.isInvertedUnit(newUnit))
          this.addRatioUnit(new DelayedPromiseWithProps(newUnit.key, async () => newUnit), unitSpec.label);
```

core/ecschema-metadata/src/Metadata/Format.ts
```diff
@@ -187,6 +247,21 @@ export class Format extends SchemaItem {
         schemaJson.stationSeparator = this.stationSeparator;
     }
 
+    if (FormatType.Ratio === this.type) {
+      if (undefined !== this.ratioType)
```
ben-polinsky, line 262: And again, we have the same/similar block of code repeated around 303? Extract?

core/ecschema-metadata/src/Metadata/Format.ts
```diff
@@ -225,13 +300,34 @@ export class Format extends SchemaItem {
...
+      itemElement.setAttribute("ratioFormatType", this.ratioFormatType);
     if (undefined !== this.stationOffsetSize)
       itemElement.setAttribute("stationOffsetSize", this.stationOffsetSize.toString());
 
     const formatTraits = formatTraitsToArray(this.formatTraits);
     if (formatTraits.length > 0)
       itemElement.setAttribute("formatTraits", formatTraits.join("|"));
 
+    // Serialize ratioUnits
+    if (undefined !== this.ratioUnits && this.ratioUnits.length === 2) {
```
ben-polinsky: This condition of `this.ratioUnits?.length === 2` is used in several places. Could be worth extracting to a function to give some semantic meaning and reduce errors?

core/ecschema-metadata/src/Formatting/SchemaFormatsProvider.ts
```diff
@@ -92,15 +92,25 @@ export class SchemaFormatsProvider implements FormatsProvider {
...
+    const persistenceUnitSystem = await persistenceUnit?.unitSystem;
+
     for (const matcher of unitSystemMatchers) {
       for (const lazyFormat of presentationFormats) {
         const format = await lazyFormat;
-        const unit = await (format.units && format.units[0][0]);
-        if (!unit) {
+        // Get the first unit from either units (composite) or ratioUnits (ratio format)
+        let unit: Unit | InvertedUnit | undefined;
+        if (format.units && format.units.length > 0 && format.units[0].length > 0) {
```
ben-polinsky:
```suggestion
        if (format.units?.[0].length > 0) {
```

ben-polinsky: Same with below condition.

core/quantity/src/Formatter/Format.ts
```diff
@@ -452,6 +501,16 @@ export class Format extends BaseFormat {
     const azimuthBaseUnit = this.azimuthBaseUnit;
     const revolutionUnit = this.revolutionUnit;
 
+    let ratioUnits: ResolvedFormatUnitSpec[] | undefined;
+    if (this._ratioUnits && this._ratioUnits.length === 2) {
```
ben-polinsky: You have a helper method above `hasRatioUnits` that is this exact same statement.

core/quantity/src/Formatter/FormatterSpec.ts
```diff
@@ -60,6 +61,60 @@ export class FormatterSpec {
   public get azimuthBaseConversion(): UnitConversionProps | undefined { return this._azimuthBaseConversion; }
   public get revolutionConversion(): UnitConversionProps | undefined { return this._revolutionConversion; }
 
+  /** Build conversion specs for ratio format with explicit numerator/denominator units. */
+  private static async getRatioUnitConversions(format: Format, unitsProvider: UnitsProvider, persistenceUnit: UnitProps): Promise<UnitConversionSpec[]> {
+    const conversions: UnitConversionSpec[] = [];
+
+    // Already validated by caller that hasRatioUnits is true
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const [numeratorUnit, numeratorLabel] = format.ratioUnits![0];
```
ben-polinsky: Since `format` isn't used elsewhere in this method, and you're sure that format.ratioUnits[0] and ...[1] are present, just pass in a tuple with the two ratioUnits? Or each as a separate arg? Then no non-null assertion would be required.

core/quantity/src/Formatter/FormatterSpec.ts
```diff
@@ -60,6 +61,60 @@ export class FormatterSpec {
...
+      label: "",
+      system: numeratorUnit.system,
+      conversion: { factor: ratioScaleFactor, offset: 0.0 },
+    };
+    conversions.push(ratioConversionSpec);
+
+    // Numerator unit for label lookup
+    const numeratorSpec: UnitConversionSpec = {
+      name: numeratorUnit.name,
+      label: (numeratorLabel && numeratorLabel.length > 0) ? numeratorLabel : numeratorUnit.label,
```
ben-polinsky:
```suggestion
      label: (numeratorLabel?.length > 0) ? numeratorLabel : numeratorUnit.label,
```

core/quantity/src/Formatter/Interfaces.ts
```diff
@@ -70,6 +90,14 @@ export interface FormatProps {
...
   readonly allowMathematicOperations?: boolean;
   readonly composite?: FormatCompositeProps;
+
+  /** Optional array of 2 unit specifications for ratio scale factors.
+   * The first unit is the numerator unit, the second is the denominator unit.
+   * Example: `[{name: "Units.IN", label: '"'}, {name: "Units.FT", label: "'"}]` for inches per foot.
+   * When provided, a conversion factor is computed from the two units (numerator/denominator).
+   * @beta
+   */
+  readonly ratioUnits?: FormatUnitSpec[];
```
ben-polinsky: Is this always an array of length between 0 and 2? Could be a tuple.

core/quantity/src/Parser.ts
```diff
@@ -903,38 +904,124 @@ export class Parser {
...
+        } else if (token.isOperator && i === 0) {
+          // Handle negative sign at start
+          if (token.value === "-" && i + 1 < tokens.length && tokens[i + 1].isNumber) {
+            value = -(tokens[i + 1].value as number);
+            foundNumber = true;
+            i++; // Skip the next token since we've consumed it
+          }
+        }
+      }
+
```
ben-polinsky: I think some cleanup here would help comprehension.

core/quantity/src/Parser.ts
```diff
@@ -903,38 +904,124 @@ export class Parser {
...
+      return { value, unitLabel };
+    };
+
+    const numeratorPart = parseRatioPart(parts[0]);
+    const denominatorPart = parts.length === 1 ? { value: 1.0 } : parseRatioPart(parts[1]);
+
+    if (isNaN(numeratorPart.value) || isNaN(denominatorPart.value)) return { ok: false, error: ParseError.NoValueOrUnitFoundInString };
+
+    // Handle ratioUnits case - simpler conversion using the pre-computed scale factor
+    if (spec.format.ratioUnits && spec.format.ratioUnits.length === 2 && spec.unitConversions.length >= 1) {
```
ben-polinsky: Same check for ratioUnits and ensuring it's of length 2 again and again.

core/quantity/src/ParserSpec.ts
```diff
@@ -38,6 +38,62 @@ export class ParserSpec {
   public get azimuthBaseConversion(): UnitConversionProps | undefined { return this._azimuthBaseConversion; }
   public get revolutionConversion(): UnitConversionProps | undefined { return this._revolutionConversion; }
 
+  /** Build conversion specs for ratio format with explicit numerator/denominator units. */
+  private static async getRatioUnitConversions(format: Format, unitsProvider: UnitsProvider, outUnit: UnitProps, altUnitLabelsProvider?: AlternateUnitLabelsProvider): Promise<UnitConversionSpec[]> {
+    const conversions: UnitConversionSpec[] = [];
+
+    // Already validated by caller that hasRatioUnits is true
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const [numeratorUnit, numeratorLabel] = format.ratioUnits![0];
```
ben-polinsky: Consider tuples.

core/ecschema-metadata/src/Metadata/Format.ts
hl662: I can't do it this way, when using DelayedPromiseWithProps the type system expects separate types when constructing the delayed promise - One with DelayedPromiseWithProps<LazyLoadedUnit> and the other is DelayedPromiseWithProps<LazyLoadedInvertedUnit>

Edit; I think I have something

core/ecschema-metadata/src/Metadata/Format.ts
hl662: One block is used to serialize to json, the other is used to serialize to XML, each using its own helper functions to help do that. Not sure if we can

core/quantity/src/Formatter/Format.ts
hl662: I'm going to need to do this across the file in a separate PR - this file has a mix of everything in other places

core/quantity/src/Formatter/Format.ts
hl662: cool, I've removed the redundant messaging in the comments

core/quantity/src/Formatter/Interfaces.ts
hl662: eh... javascript doesn't have a native tuple data structure, so while we can use [FormatUnitSpec, FormatUnitSpec], the deserializing and serializing method will still need to validate the array length and use array functions. Plus, I'm following how `composite` units are done for this new `ratioUnits` property, in the (likely) event we'll have to support this new property in ecdb

core/quantity/src/Parser.ts
hl662: thanks, I cleaned up the code, reusing the existing parseQuantitySpecification to handle parsing fractions, and moved parseRatioPart to its own private method instead of an inline method

core/quantity/src/Parser.ts
hl662: reused format.hasRatioUnits to avoid this

core/quantity/src/Formatter/FormatterSpec.ts
hl662: alright, changed function definition to pass in a tuple with 2 ratio units, I still had to use no non-null assertion at function call though, TS linter isn't smart enough to interpret hasRatioUnits() has already checked ratioUnits

core/ecschema-metadata/src/Metadata/Format.ts
hl662: this is only used in 2 places, I know it can seem a lot larger because we also do this similar check against formatProps when deserializing but they are different interfaces so they can't all share a single function. And if we do define a single function, we run into no non-null assertion rules again (since the linter sucks) :(

core/quantity/src/Formatter/Format.ts
hl662: thx for the catch, reused hasRatioUnits

core/quantity/src/ParserSpec.ts
hl662: Thanks, reused this impl https://github.com/iTwin/itwinjs-core/pull/8787#discussion_r2623897876

core/quantity/src/Formatter/Format.ts
ben-polinsky: It's not the biggest concern, thanks.

core/quantity/src/Parser.ts
ben-polinsky: Thanks

core/backend/src/test/imodel/SchemaXmlImport.test.ts
```diff
@@ -9,77 +9,122 @@ import { PhysicalElement, SnapshotDb } from "../../core-backend";
...
+      for (const testCase of [`3.2`, `3.5`]) {
+        assert(await helperFunction(imodel, [`<ECSchema schemaName="TestSchema" alias="ts" version="1.0.0" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.${testCase}"/>`], true), `Schema ${testCase} import should have succeeded.`);
+        assert(await helperFunction(imodel, [`TestSchema`], false), `Schema ${testCase} test should have succeeded.`);
+      }
+    } finally {
+      imodel.close();
     }
+  });
+
+  it("should roundtrip ratio format properties", async () => {
```
rschili: @hl662 sorry, I'm pushing a failing test with your PR :-)
This is a missing piece. If we import a format with the new properties into IModelDb, and load it back, it will not contain the new fields anymore.
We need to adjust the native serializers to support the field as well, and ECDb as well.
I did that during the bearing/azimuth story. But it's a bit dirty. We're exploiting the fact that ecdb stores format props as json instead of columns on the ec_format table. That way we can roundtrip new things without breaking the file format.

@ColinKerr @diegoalexdiaz : I'm getting a bit worried. This is the second time we will be shoveling additional fields into ECSchema without actually incrementing the EC version. To me it seems like we should have incremented to EC 3.3 by now. It's also pure coincidence that we're able to store these new fields in ECDb as I described above (because it persists json). What are your thoughts?

core/backend/src/test/imodel/SchemaXmlImport.test.ts
```diff
@@ -9,77 +9,122 @@ import { PhysicalElement, SnapshotDb } from "../../core-backend";
...
-  after(() => {
-    if (imodel)
-      imodel.close();
   });
 
   it("should import schema XML", async () => {
-    const schemaFilePath = path.join(KnownTestLocations.assetsDir, "Test3.ecschema.xml");
-    const schemaString = fs.readFileSync(schemaFilePath, "utf8");
+    const testFileName = IModelTestUtils.prepareOutputFile("SchemaXMLImport", "SchemaXMLImport.bim");
+    const imodel = SnapshotDb.createEmpty(testFileName, { rootSubject: { name: "SchemaXMLImportTest" } });
```
rschili, line 25: please don't be confused by all the other edits in the file. It's just rearranging things because every test was using the same imodel file instead of an isolated local one. This comment can be marked resolved once seen.

rschili (changes requested):
commented in place. The current implementation won't survive roundtrips through an imodel

core/backend/src/test/imodel/SchemaXmlImport.test.ts
ColinKerr: You should not be able to add one of these new formats to an EC 3.2 schema ... Our current typescript code might not have existing logic to handle detecting and failing on this but we need it.  It should be an obvious and explicit failure rather than a loss of data.

Further more you should not be able to serialize an EC 3.2 schema with these new formats ... that should only be possible with the as of yet not implemented EC 3.3 spec.

hl662: To implement the error handling, we'll have to update the native code, as that's where schema validation is housed. The importSchemaStrings in iModelDb.ts calls this[_nativeDb].importXmlSchemas. We don't want to add the logic to properly handle these new ratio properties yet right? Just add the error raising?

ColinKerr: Currently the C++ code is the only code that does validation based on EC spec version.  But itwin.js does have schema validation and will need to become spec version aware at some point.  ecschema-metadata has avoided because it was implemented after EC 3.2 existed and is not able to load EC 3.1.  We don't need full-blown multi-spec version support but we do need some guard rails to keep us from reading/writing EC 3.2 schemas with unsupported formats.

hl662: In my most recent commit, I added some guard rails to the serializing/deserializing code for Formats on our TS side, to make sure the ratio properties don't appear on EC3.2 schemas and formats. That won't fix this round trip test however, because we need to extend the [importSchemaStrings()](https://github.com/iTwin/itwinjs-core/blob/master/core/backend/src/IModelDb.ts#L1060) in iModelDb.ts to have some extra validation, before calling the native side's [importXmlSchema](https://github.com/iTwin/imodel-native/blob/0d96eebc8a41f18453daaedc73a2f9410cbf4f72/iModelJsNodeAddon/IModelJsNative.cpp#L2218). 

I would prefer to create a separate GH issue to tackle this, and skip this round trip test for now (keeping the test but just skipping) - preferably we could look to expose a validation method from ecschema-metadata that iModelDb would call to validate. The [SchemaReadHelper](https://github.com/iTwin/itwinjs-core/blob/5b5447cf96924dc802f240a3a74628c2a4f4d3a9/core/ecschema-metadata/src/Deserialization/Helper.ts#L47) could be an option as that's what is used on the TS side for validation, but it's internal right now. Personally am not familiar with the codebase enough to decide this.
@rschili @ColinKerr

rschili: We just had a call and decided that this is fine for now and we will disable the test until we decide otherwise. It is a limitation that we may solve later but for now we will just roll with it because it's not simple to support and not needed for the immediate scenarios.

core/backend/src/test/imodel/SchemaXmlImport.test.ts
```diff
@@ -9,77 +9,122 @@ import { PhysicalElement, SnapshotDb } from "../../core-backend";
...
+      for (const testCase of [`3.2`, `3.5`]) {
+        assert(await helperFunction(imodel, [`<ECSchema schemaName="TestSchema" alias="ts" version="1.0.0" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.${testCase}"/>`], true), `Schema ${testCase} import should have succeeded.`);
+        assert(await helperFunction(imodel, [`TestSchema`], false), `Schema ${testCase} test should have succeeded.`);
+      }
+    } finally {
+      imodel.close();
     }
+  });
+
+  it.skip("should roundtrip ratio format properties", async () => {
```
ben-polinsky, line 90: intentionally skipped?

core/ecschema-metadata/src/Metadata/Format.ts
```diff
@@ -107,43 +113,106 @@ export class Format extends SchemaItem {
...
-      else if(InvertedUnit.isInvertedUnit(newUnit))
-        this.addUnit(new DelayedPromiseWithProps(newUnit.key, async () => newUnit), unit.label);
+
+    // Process composite units
+    if (undefined !== formatProps.composite) {
+      for (const unit of formatProps.composite.units) {
+        const newUnit = this.schema.lookupItemSync(unit.name);
+        if (undefined === newUnit || (!Unit.isUnit(newUnit) && !InvertedUnit.isInvertedUnit(newUnit)))
+          throw new ECSchemaError(ECSchemaStatus.InvalidECJson, ``);
+        const lazyUnit = Unit.isUnit(newUnit)
```
ben-polinsky, line 137: Both ternary conditions return the same result?

core/ecschema-metadata/src/Metadata/Format.ts
```diff
@@ -107,43 +113,106 @@ export class Format extends SchemaItem {
...
-      else if(InvertedUnit.isInvertedUnit(newUnit))
-        this.addUnit(new DelayedPromiseWithProps(newUnit.key, async () => newUnit), unit.label);
+
+    // Process composite units
+    if (undefined !== formatProps.composite) {
+      for (const unit of formatProps.composite.units) {
+        const newUnit = await this.schema.lookupItem(unit.name);
+        if (undefined === newUnit || (!Unit.isUnit(newUnit) && !InvertedUnit.isInvertedUnit(newUnit)))
+          throw new ECSchemaError(ECSchemaStatus.InvalidECJson, ``);
+        const lazyUnit = Unit.isUnit(newUnit)
```
ben-polinsky, line 185: see above comment... are the ternary results supposed to be identical? Are my eyes deceiving me?

ben-polinsky: Lines 184 + below.. poor snippet gh

core/ecschema-metadata/src/Metadata/Format.ts
ben-polinsky: lines 136 and below.

hl662: Yeah, it had to be done, typescript couldn't infer the type properly without the ternary operation. Using Unit.isUnit() as a ternary check helps TS narrow the type down correctly. It's not too clear, so I added type casting to show....

core/ecschema-metadata/src/Metadata/Format.ts
hl662: See comment above

core/backend/src/test/imodel/SchemaXmlImport.test.ts
hl662: Yes, see https://github.com/iTwin/itwinjs-core/pull/8787#discussion_r2717304123

Plans and priorities are being discussed internally within the team that may result in the round trip test not being needed in the future, but keeping the test here for now

PR #8792, opened 2025-11-20
Add copilot instructions for iTwin.js development and breaking changes policy

khanaffan:
This is generated with help of Claude Sonnet 4.5. I customize on top of it asking it to emphize on breaking changes. Also expand on core packages.

PR #8794, opened 2025-11-21
Improve copilot instructions regarding documentation

hl662:
Improving on https://github.com/iTwin/itwinjs-core/pull/8792

I tested the copilot instructions with a moderately open ended prompt to add a new feature, intentionally limiting the context to a test file and one source code file, to see if the agent can grab other files across the repo that are relevant. 

image

The result was it was able to grab more, as expected, and it was able to interpret my added instructions on code examples, inserting the code example into NextVersion.md. I hoped it was able to grab the index.md found in docs/quantity, but not bad. I compared the output to my current draft [PR](https://github.com/iTwin/itwinjs-core/pull/8787) for ratio, it was closely similar
image
image

The agent asked for permission to execute rushx build and then rushx test, which was a key difference from when we didn't have copilot instructions (it would use npm build and npm test). We still need more changes to the instructions, haven't figured out how to restructure it yet, but the agent should be able to prompt (or remind) the user to run rush change **after** staging changes to the git branch.

I also tweaked the instructions for vitest - vitest doesn't need to run rushx build, as long as you have the vitest extension installed - copilot will be very effective, selectively running tests, and able to read test results, and fix source code (near instantaneous too) without needing to look at the terminal.

I find GHCP sometimes hangs when trying to read test results from mocha based tests. Another incentive to move to vitest.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -246,6 +246,25 @@ All public APIs must include TSDoc comments with:
 - `@throws` tag for exceptions
 - `@deprecated` tag with migration guidance for deprecated APIs
 
+### Deprecation Format
+
+When adding `@deprecated` tags, follow these formats - **do NOT manually add dates** (the pipeline automatically adds them):
+
+```typescript
+// Basic deprecation with description only:
+/** @deprecated Use NewClass instead. */
+
+// With version (pipeline will auto-add date):
+/** @deprecated in 4.5.0. Use NewClass instead. */
+```
+
+Valid formats recognized by the ESLint rule:
+- `@deprecated {description}` - Description only (5+ characters required)
+- `@deprecated in {major}.{minor}.{patch}. {description}` - With version number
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -211,14 +221,40 @@ iTwin.js uses **EC (Entity-Class)** schemas based on BIS:
...
+
+### TSDoc Comments
+
+All public APIs must include TSDoc comments with:
+- Release tags (`@public`, `@beta`, `@alpha`, `@internal`)
+- Brief description of purpose
+- `@param` tags for all parameters
+- `@returns` tag for return values
+- `@throws` tag for exceptions
+- `@deprecated` tag with migration guidance for deprecated APIs
```
MichaelSwigerAtBentley, line 247: Can we add a clause to tell it to not try to add dates?

hl662: sure, added an explicit section on how to format doccomments that are deprecated

PR #8795, opened 2025-11-21
docs: clarify getDerivedClasses behavior regarding unloaded referencing schemas (#8047)

### LINKED ISSUE #8047: `ECClass.getDerivedClasses()` doesn't include classes from referenced schemas
**Describe the bug**

When an iModel / ECDb contains schemas, whose classes are derived from by classes in referencing schemas, calling `ECClass.getDerivedClasses()` on a base class form referenced schema doesn't return any classes.

**To Reproduce**

To reproduce the behavior, run this test:

```ts
it.only("`ECClass.getDerivedClasses()` doesn't include classes from referenced schemas", async function () {
  await IModelHost.startup();

  // create a new ECDb
  const ecdbPath = "/home/grigas/test.ecdb";
  IModelJsFs.unlinkSync(ecdbPath);
  const ecdb = new ECDb();
  ecdb.createDb(ecdbPath);

  // import referenced schema
  const schemaXml1 = `
    <?xml version="1.0" encoding="UTF-8"?>
    <ECSchema schemaName="TestSchema1" alias="Test1" version="01.00.00" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
      <ECEntityClass typeName="X" />
    </ECSchema>
  `;
  const schemaFilePath1 = "/home/grigas/test_schema1.xml";
  fs.writeFileSync(schemaFilePath1, schemaXml1);
  ecdb.importSchema(schemaFilePath1);
  ecdb.saveChanges();

  // import referencing schema
  const schemaXml2 = `
    <?xml version="1.0" encoding="UTF-8"?>
    <ECSchema schemaName="TestSchema2" alias="Test2" version="01.00.00" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
      <ECSchemaReference name="TestSchema1" version="01.00.00" alias="test1" />
      <ECEntityClass typeName="Y">
        <BaseClass>test1:X</BaseClass>
      </ECEntityClass>
    </ECSchema>
  `;
  const schemaFilePath2 = "/home/grigas/test_schema1.xml";
  fs.writeFileSync(schemaFilePath2, schemaXml2);
  ecdb.importSchema(schemaFilePath2);
  ecdb.saveChanges();

  // set up schema locater
  const locater = new SchemaJsonLocater((name) => ecdb.getSchemaProps(name));
  context.addLocater(locater);
  const testSchema1 = await context.getSchema(new SchemaKey("TestSchema1"), SchemaMatchType.Latest);
  const classX = await testSchema1!.getItem("X", EntityClass);
  const derivedClasses = await classX!.getDerivedClasses();
  expect(derivedClasses!.length).to.equal(1);

  await IModelHost.shutdown();
});
```

**Expected behavior**

- Calling `ECClass.getDerivedClasses()` on a base class, returns all its derived classes, including ones from referencing schemas.
- The test passes.

- OS: Ubuntu
- Node.js version `20.18.2`
- iTwin.js version `5.0.0-dev.92`

**Additional context**

I understand it might be problematic to get derived classes from referencing schemas, if those schemas haven't been loaded. If fixing this is not possible, the situation should be clearly explained on affected methods (I suspect that might be more than just `getDerivedClasses()`), and, ideally, a recommendation on how to get those derived classes should be provided.

ashmitKmishra:
Fixes #8047.

**Analysis:**
`getDerivedClasses` relies on the current `SchemaContext` to identify subclasses. Since `SchemaContext` does not maintain a reverse-index of all possible external schemas that *might* reference a base class, it is architecturally impossible for this method to return classes from unloaded schemas without a global pre-scan.

**Changes:**
1. Added a reproduction test case in `core/backend` demonstrating that `getDerivedClasses` returns empty until the referencing schema is explicitly loaded.
2. Updated the JSDoc for `getDerivedClasses` in `core/ecschema-metadata` to explicitly warn users that referencing schemas must be loaded into the context to be discovered.

This resolves the ambiguity noted by the reporter.

### DIFF
--- a/common/changes/@itwin/core-backend/Ash-8047-clarify-derived-classes_2026-02-02-17-52.json
+++ b/common/changes/@itwin/core-backend/Ash-8047-clarify-derived-classes_2026-02-02-17-52.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/changes/@itwin/ecschema-metadata/Ash-8047-clarify-derived-classes_2026-02-02-17-52.json
+++ b/common/changes/@itwin/ecschema-metadata/Ash-8047-clarify-derived-classes_2026-02-02-17-52.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-metadata",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/ecschema-metadata"
+}


core/backend/src/test/ecdb/Issue8047.test.ts
```diff
@@ -0,0 +1,90 @@
+import { assert } from "chai";
```
rschili: Thanks for contributing!
This test seems alright, but we usually don't follow the pattern of putting tests for one issue into a separate file.
Putting this into the ecdb tests is also not the best fit for this.
We do have a very similar existing SchemaContext embedded in IModelDb.
So either this should be a high level test using IModelDb, in which case `core/backend/src/test/schema/IModelSchemaContext.test.ts` is a good location.
Or if we just want to confirm the ecschema-metadata package behavior, which is more of a unit test, `core/ecschema-metadata/src/test/Context/SchemaContext.test.ts` is the place to put it.

ashmitKmishra: I got it. It makes sense to avoid making tiny test files. I'll update the test script!

ashmitKmishra: Hey, @rschili 
I have added the test case to IModelSchemaContext.test.ts. Please let me know if anything else has to be done in this PR.

rschili: Something went wrong. The ECSqlStatement.test.ts file was deleted and the Issue8047.test.ts file still exists. Please check again

ashmitKmishra: My apologies! I accidentally deleted the wrong file during the cleanup. I have restored ECSqlStatement.test.ts and removed Issue8047.test.ts now. Thanks for catching that.

core/backend/src/test/schema/IModelSchemaContext.test.ts
```diff
@@ -113,3 +117,83 @@ describe("IModel Schema Context", () => {
     assert.isDefined(expectedString);
   });
 });
+
+// --- Reproduction Test for Issue #8047 ---
+describe("Issue #8047: Derived Classes in Reference Schemas", () => {
```
MichaelSwigerAtBentley: Thanks for taking this on, looks good to me

You also don't need to include the issue number in the test case, This test block could just be "Derived Classes in Reference Schemas".

rschili: Agreed, suggested new name "getDerivedClasses returns only loaded schemas"

ColinKerr:
```suggestion
describe("getDerivedClasses returns only loaded schemas", () => {
```

ColinKerr:
I turned the suggestion from @rschili and @MichaelSwigerAtBentley  into a suggested code change.  I have approved pending the resolution of that suggestion.

We may leave the issue opened to track the problem for a future fix.  But in the meantime this update will help many people avoid introducing a tricky bug into the code.  So it's a valuable update to the codebase.

Thanks @ashmitKmishra!

ashmitKmishra:
> I turned the suggestion from @rschili and @MichaelSwigerAtBentley into a suggested code change. I have approved pending the resolution of that suggestion.
> 
> We may leave the issue opened to track the problem for a future fix. But in the meantime this update will help many people avoid introducing a tricky bug into the code. So it's a valuable update to the codebase.
> 
> Thanks @ashmitKmishra!

MichaelSwigerAtBentley:
[Passing Integration Tests](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4755985&view=results)
[Passing Build](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4755982&view=results)

aruniverse (approved):

PR #8797, opened 2025-11-24
ChannelAdmin.makeChannelRoot should use ElementOwnsChannelRootAspect relationship

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/1602

### DIFF
--- a/common/config/rush/version-policies.json
+++ b/common/config/rush/version-policies.json
@@ -2,7 +2,7 @@
   {
     "policyName": "prerelease-monorepo-lockStep",
     "definitionName": "lockStepVersion",
-    "version": "5.4.0-dev.16",
+    "version": "5.4.0-dev.17",
     "nextBump": "prerelease"
   }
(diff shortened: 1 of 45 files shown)


core/backend/src/NavigationRelationship.ts
```diff
@@ -287,3 +287,15 @@ export class SheetOwnsSheetInformationAspect extends ElementOwnsUniqueAspect {
     super(sheetId, relClassName);
   }
 }
+
+/** Relates an [[Element]] to the [[ChannelRootAspect]] that identifies it as a channel root.
+ * @see [[ChannelControl.makeChannelRoot]] to create an instance of this relationship.
+ * @beta
```
pmconne:
```suggestion
 * @public
```
The BIS class this represents has been in BisCore since 01.00.10 and your entire API consists solely of a constructor.

PR #8806, opened 2025-11-28
Incremental schema loading test cleanup

StefanApfel-Bentley:
#### Description

- Removed some unnecessary tests (fetching schemas through _getSchemaProps_)
- Refactored the apis 
    - switched from a static class _IncrementalTestHelper_ to disposable class _TestContext_.
    - removed _IModelSchemaLocater_, using _SchemaJsonLocater_ directly now.
    - Replaced _TestSqlSchemaLocater_ with abstract _TestSchemaLocater_ class to make protected methods accessible.

Also fixed SimpleSchema which was refering to non-existing definitions. That didn't cause any errors, but produced some warnings when importing the schema into the IModel.


StefanApfel-Bentley:
It seems the macOS build doesn't like when I start and close the _IModelHost_ for each test. I changed that actually, to set the _incrementalSchemaLoading_ flag on the host, but if that causes problems, I switch back to manipulate the host configuration instead.

PR #8807, opened 2025-12-01
Add 5.3.0.md

GytisCepk:
Due to an oversight during release, we failed to run pipeline which would have updated this automatically.


GytisCepk:
@Mergifyio  backport release/5.4.x

PR #8809, opened 2025-12-01
Add RealityTileLoader tests

eringram:
Adds some simple unit tests for `RealityTileLoader.loadGeometryFromStream` for this bug fix: https://github.com/iTwin/itwinjs-core/pull/8785. They check that the transform provided to `B3dmReader.readGltfAndCreateGeometry` uses both the tile tree's iModelTransform and the tile's transformToRoot.

There are only two new tests and they share the test classes and set-up from the `RealityTile` reprojection tests, so I added them to the same file in a different `describe` block. This required the test classes to be moved outside the `describe`.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -6551,8 +6551,8 @@ packages:
   base64-js@1.5.1:
     resolution: {integrity: sha512-AKpaYlHn8t4SVbOHCy+b5+KKgvR4vrsD8vbvrbiQJps7fKDTkjkDry6ji0rUJjC0kzbNePLwzxq8iypo41qeWA==}
 
-  baseline-browser-mapping@2.8.26:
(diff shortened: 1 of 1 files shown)


markschlosseratbentley (approved):
LGTM, thanks for adding these.

eringram:
`rush audit` is failing due to this CVE https://github.com/advisories/GHSA-869p-cjfg-cm3x. I believe itwinjs-core doesn't directly rely on the vulnerable package, but some dependencies of dependencies do, so it will probably fail until some external packages are updated

markschlosseratbentley:
> `rush audit` is failing due to this CVE [GHSA-869p-cjfg-cm3x](https://github.com/advisories/GHSA-869p-cjfg-cm3x). I believe itwinjs-core doesn't directly rely on the vulnerable package, but some dependencies of dependencies do, so it will probably fail until some external packages are updated

fyi @aruniverse

aruniverse:
> > `rush audit` is failing due to this CVE [GHSA-869p-cjfg-cm3x](https://github.com/advisories/GHSA-869p-cjfg-cm3x). I believe itwinjs-core doesn't directly rely on the vulnerable package, but some dependencies of dependencies do, so it will probably fail until some external packages are updated
> 
> fyi @aruniverse

https://github.com/iTwin/itwinjs-core/pull/8828

PR #8811, opened 2025-12-01
Improved B-Spline close approach

saeeedtorabi:


core/geometry/src/numerics/Newton.ts
```diff
@@ -79,7 +79,11 @@ export abstract class AbstractNewtonIterator {
...
   /** Smallest iterate size difference in later iterations. */
   private _leastDelta: number = Number.MAX_VALUE;
-  /** The current late iterate has the least delta encountered. Remember it. Implement along with [[restoreCandidate]].*/
+  /**
+   * The current late iterate has the least delta encountered. Remember it. Implement along with [[restoreCandidate]].
+   * The cache ensures that even when Newton doesn't formally converge, the best approximation is returned, not the
+   * last (potentially diverged) iterate.
+   */
   protected cacheCandidate?(): void;
   /** Set Newton result to the cached candidate. Implement along with [[cacheCandidate]]. */
```
dassaf4:
```suggestion
   * The current late iterate has the least delta encountered. Remember it.
   * * Implement this method along with [[restoreCandidate]].
   */
protected cacheCandidate?(): void;
/**
 * Set Newton result to the cached candidate.
 * * This is the best approximation to the root found during Newton iteration and may be useful to the caller even when Newton fails to converge.
 * * Implement this method along with [[cacheCandidate]].
 */
```

core/geometry/src/test/bspline/BsplineCurve.test.ts
```diff
@@ -1112,7 +1112,7 @@ describe("BsplineCurve", () => {
     expect(ck.getNumErrors()).toBe(0);
   });
   it("intersectionXYPairs", () => {
-    const ck = new Checker();
+    const ck = new Checker(true, true);
```
dassaf4:
```suggestion
    const ck = new Checker();
```

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -715,46 +716,34 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
-                errors++;
-            }
             if (this.acceptFraction(false, bcurveAFraction, false) && this.acceptFraction(false, bcurveBFraction, false)) {
-              this.recordPointWithLocalFractions(
-                bcurveAFraction, bcurveA, 0, 1, bcurveBFraction, bcurveB, 0, 1, reversed,
-              );
+              const pointA = bezierA.fractionToPoint(bezierAFraction, CurveCurveIntersectXY._workPointA0);
+              const pointB = bezierB.fractionToPoint(bezierBFraction, CurveCurveIntersectXY._workPointB0);
+              if (!converged) { // Newton may have found close points even if it didn't converge parametrically
+                const strictTolerance = this._coincidentGeometryContext.tolerance * 0.0001;
```
dassaf4: Move the initialization of `strictTolerance` outside of the `roots` loop.

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -670,10 +670,8 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
       this.computeSegmentLineString(segmentA, this._geometryB, false);
     } else if (this._geometryB instanceof Arc3d) {
       this.computeSegmentArc(segmentA, this._geometryB, false);
-    } else if (this._geometryB instanceof BSplineCurve3d) {
-      this.dispatchSegmentBsplineCurve(segmentA, this._geometryB, false);
-    } else if (this._geometryB instanceof TransitionSpiral3d) {
-      this.dispatchCurveSpiral(segmentA, this._geometryB, false);
+    } else if (this._geometryB instanceof BSplineCurve3d || this._geometryB instanceof TransitionSpiral3d) {
```
dassaf4: 1. Make a private helper method `needsStroking(curve: CurvePrimitive): boolean` to encapsulate this type check.
2. Instead of checking for `BSplineCurve3d`, check for `BSplineCurve3dBase`. This catches the homogeneous subclass as well.
3. Call this helper from the "handleXXX" methods (like here) and also in `dispatchCurveStrokedCurve`, where it is used to decide whether to stroke `curveA`.

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -796,37 +782,37 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
...
-   * @param curveA curve to find its close approach with spiralB. May also be a transition spiral.
-   * @param spiralB transition spiral to find its close approach with curveA.
-   * @param reversed whether `spiralB` data will be recorded in `detailA` of each result, and `curveA` data in `detailB`.
+   * Compute the XY close approach of a curve and a spiral or B-spline.
+   * @param curveA curve to find its close approach with spiralOrBSplineB. May also be a transition spiral.
+   * @param spiralOrBSplineB transition spiral or B-spline to find its close approach with curveA.
+   * @param reversed whether `spiralOrBSplineB` data will be recorded in `detailA` of each result, and `curveA` data in `detailB`.
    */
-  private dispatchCurveSpiral(curveA: CurvePrimitive, spiralB: TransitionSpiral3d, reversed: boolean): void {
+  private dispatchCurveSpiralOrBSpline(curveA: CurvePrimitive, spiralOrBSplineB: TransitionSpiral3d | BSplineCurve3dBase, reversed: boolean): void {
```
dassaf4: There is nothing special about `curveB` here that requires it to be a spiral or a B-spline curve. Rename and refactor to remove this unnecessary restriction: `dispatchCurveStrokedCurve(curveA: CurvePrimitive, curveB: CurvePrimitive, reversed: boolean): void`.  In the doc, note that `curveB` is to be stroked.

```suggestion
  private dispatchCurveStrokedCurve(curveA: CurvePrimitive, curveB: CurvePrimitive, reversed: boolean): void {
```

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -844,19 +830,7 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
   }
   /** Double dispatch handler for strongly typed homogeneous bspline curve .. */
   public override handleBSplineCurve3dH(_curve: BSplineCurve3dH): any {
-    /*
     //NEEDS WORK -- make "dispatch" methods tolerant of both 3d and 3dH.
```
dassaf4: 1. You can now implement this method similar to `handleBSplineCurve3d`!
2. Write some tests to flex it.
3. Or create an issue and do this separately.

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -670,10 +670,8 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
       this.computeSegmentLineString(segmentA, this._geometryB, false);
```
saeeedtorabi, line 646: I would like to move all handle functions to the end of `CurveCurveIntersectXY.ts` and `CurveCurveCloseApproachXY.ts` so all handle functions are at the end. Now we have some of handles at the middle, then some helper function, and then some other handles at the end. I did not make this change so you can see my changes. Once the PR is approved, I will make the move.

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -796,37 +782,37 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
...
+  private dispatchCurveSpiralOrBSpline(curveA: CurvePrimitive, spiralOrBSplineB: TransitionSpiral3d | BSplineCurve3dBase, reversed: boolean): void {
     // explicit search for intersections (Newton converges too slowly on DirectSpiral3d tangent intersections)
-    const intersections = CurveCurve.intersectionXYPairs(curveA, false, spiralB, false, this._xyTolerance);
+    const intersections = CurveCurve.intersectionXYPairs(curveA, false, spiralOrBSplineB, false, this._xyTolerance);
     for (const intersection of intersections)
       this.testAndRecordPair(intersection, reversed);
     // append seeds computed by solving the discretized spiral close approach problem, then refine the seeds via Newton
     let cpA = curveA;
-    if (curveA instanceof TransitionSpiral3d)
+    if (curveA instanceof TransitionSpiral3d || curveA instanceof BSplineCurve3dBase)
```
dassaf4: Use the new helper method `needsStroking` here.

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -721,46 +715,38 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
...
-    const xyMatchingFunction = new CurveCurveCloseApproachXYRRtoRRD(curveA, spiralB);
+    const xyMatchingFunction = new CurveCurveCloseApproachXYRRtoRRD(curveA, spiralOrBSplineB);
     const newtonSearcher = new Newton2dUnboundedWithDerivative(xyMatchingFunction, 50, this._newtonTolerance); // seen: 47
     for (const seed of seeds) {
       const detailA = reversed ? seed.detailB : seed.detailA;
       const detailB = reversed ? seed.detailA : seed.detailB;
-      assert(detailB.curve instanceof LineString3d, "Caller has discretized the spiral");
-      newtonSearcher.setUV(detailA.fraction, detailB.fraction); // use linestring fraction as spiral param; it generally yields a closer point than fractional length!
+      assert(detailB.curve instanceof LineString3d, "Caller has discretized the spiral or B-spline");
+      newtonSearcher.setUV(detailA.fraction, detailB.fraction); // use linestring fraction as spiral/bspline param; it generally yields a closer point than fractional length!
```
dassaf4: We really don't know if the seed's fraction on the linestring is closer to the B-spline parameter of the closest point on the B-spline curve to the seed point, compared to say using the seed's fractional length.... This was only apparently a reasonable assumption for typical spirals, which have mild curvature.

```suggestion
      newtonSearcher.setUV(detailA.fraction, detailB.fraction); // use the linestring fraction as initial curveB fraction (ASSUME it's close enough)
```

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -721,46 +715,38 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
...
-   * @param spiralB The transition spiral.
-   * @param reversed whether `spiralB` data is in `detailA` of each recorded pair, and `curveA` data in `detailB`.
+   * @param spiralOrBSplineB The transition spiral or B-spline curve.
+   * @param reversed whether `spiralOrBSplineB` data is in `detailA` of each recorded pair, and `curveA` data in `detailB`.
    */
-  private refineSpiralResultsByNewton(
-    seeds: CurveLocationDetailPair[], curveA: CurvePrimitive, spiralB: TransitionSpiral3d, reversed = false
+  private refineSpiralOrBSplineResultsByNewton(
+    seeds: CurveLocationDetailPair[], curveA: CurvePrimitive, spiralOrBSplineB: TransitionSpiral3d | BSplineCurve3dBase, reversed = false
   ): void {
```
dassaf4: There is nothing special about `curveB` here that requires it to be a spiral or a B-spline curve. Rename and refactor to remove this unnecessary restriction.

```suggestion
  private refineStrokedResultsByNewton(seeds: CurveLocationDetailPair[], curveA: CurvePrimitive, curveB: CurvePrimitive, reversed = false): void {
```

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -670,10 +670,8 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
       this.computeSegmentLineString(segmentA, this._geometryB, false);
     } else if (this._geometryB instanceof Arc3d) {
       this.computeSegmentArc(segmentA, this._geometryB, false);
-    } else if (this._geometryB instanceof BSplineCurve3d) {
-      this.dispatchSegmentBsplineCurve(segmentA, this._geometryB, false);
```
dassaf4, line 674: If the `dispatchXXXBSplineCurve` methods are no longer used, they can be removed, as they are inferior.

core/geometry/src/test/curve/CurveCurveCloseApproachXY.test.ts
```diff
@@ -1807,4 +1817,233 @@ describe("CurveCurveCloseApproachXY", () => {
     GeometryCoreTestIO.saveGeometry(allGeometry, "CurveCurveCloseApproachXY", "SpiralKnownCloseApproach");
     expect(ck.getNumErrors()).toBe(0);
   });
+  it("BsplineCloseApproach", () => {
```
dassaf4, line 1820: I would prefer not to propagate the unwieldy structure of the monstrous spiral test. It is too difficult to debug, and testing returned counts is not as robust as testing known results.

Here are some general testing tips to aid your refactor of this test:

* Make multiple `it`s in a `describe` block; these are easier to debug. You don't have to see the outputs of all tests in the same DGN for eyeballing (and you shouldn't be relying on eyeballing!)---but if you really do, just import each test's IMJS output into the same DGN.

* You don't have to test a B-spline against every spiral type known to humankind. Variation should be on the B-spline curve (see next paragraph), not the other curve in the pair. Just pick one of each "other" curve to test against your B-spline curve---e.g., only one `TransitionSpiral3d` will suffice.

* Use several B-spline examples, not just one curve and a few of its rigid transforms. Try a variety of B-spline curves here: try a periodic one, try one without uniform knots, try one that isn't planar, create an `InterpolationCurve3d` and test its proxy B-spline curve, use MicroStation to create a more complicated B-spline curve, export it to IMJS, then import the IMJS in your test, etc. This also helps flex the API. Put your B-spline curve creation code in a helper method to call from multiple tests if you like.

**Aside:** Speaking of which, if we don't already have one, please add an issue to implement `ProxyCurve`s in `CurveCurveIntersect/CloseApproachXY`. These are curves that carry an exact proxy curve to be used internally for computations. Examples are `InterpolationCurve3d` and `AkimaCurve3d`, both of which store B-spline curve proxies. These types' `dispatchToGeometryHandler` implementations currently just delegate to their proxies; however, this is less than ideal for `CurveCurve` dispatch because the returned `CurveLocationDetail.curve` fields will refer to the B-spline curve proxy, _not_ the original curve, which may be unexpected by the caller. It shouldn't be too hard to create thin implementations in `CurveCurveIntersect/CloseApproachXY` for `handleInterpolation/AkimaCurve3d` that delegate to the proxy, then reset the computed `detail.curve` properties to the actual `Interpolation/AkimaCurve3d` curve.

dassaf4 (changes requested):

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
dassaf4: Appreciate you separating these cosmetic changes that are hard to diff! Though I guess Copilot could confirm nothing changed in the moved code if we asked nicely!

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
saeeedtorabi: ok but why? in is only used on line 743 so why isn't line 742 the best place for it?

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
saeeedtorabi: I will create an issue and refer it inside this method. Tests take time to write as it also should be implemented in `CurveCurveIntersectXY`

core/geometry/src/test/curve/CurveCurveCloseApproachXY.test.ts
saeeedtorabi: ok I will do all. I will rewrite the test and create an issue for proxy curves. Please note that I had written this bspline monstrous test a couple months ago (along with spiral monstrous test) and just copied it here. I did not write it from scratch as we agreed not to write these monstrous tests anymore (E has written bunch of them and that's why I followed).

Since I have a separate [issue](https://github.com/iTwin/itwinjs-backlog/issues/2012) for tangency case, can I create a separate issue to rewrite the tests? Rewriting would take some time and this PR has been up for months so I would prefer to merge it sooner than later. In that issue I can also break the spiral monsters test if you want.

dassaf4: I was thinking leave the spiral test alone. It works after I refactored it. But I don't want to use it as a template for new tests because it is not the best way going forward.

Separate issue to rewrite the B-spline test is fine. The main thing there is to expand the variety of B-spline curves tested. Right now, it's basically testing spirals more than anything else, and we already have The Monster for that.

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
dassaf4: Adding homogeneous B-spline curve support for intersectXY should be a different issue. It will require a different approach (maybe more work): in that class, we use bespoke Bezier intersection code for each primitive type.

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
dassaf4: Outside the loop is (slightly) more efficient.

saeeedtorabi: ok I see as it's only created once

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -1328,6 +1328,7 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
   public override handleBSplineCurve3dH(_curve: BSplineCurve3dH): any {
     // NEEDS WORK -- make "dispatch" methods tolerant of both 3d and 3dH .
     // "easy" if both present BezierCurve3dH span loaders
+    // https://github.com/iTwin/itwinjs-backlog/issues/2021
```
saeeedtorabi, line 1331: I will update this to 2024

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
saeeedtorabi: ok created 2 separate issues for them

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -657,6 +629,10 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
     });
     saveResults.forEach((pair: CurveLocationDetailPair) => this._results.insert(pair));
   }
+  /** Specifies whether the curve needs to be stroked for close approach computation. */
+  private needsStroking(curve: AnyCurve | undefined): curve is BSplineCurve3dBase | TransitionSpiral3d {
```
dassaf4: I think `undefined` is unnecessary (and would crash if curve were undefined?)

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -657,6 +629,10 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
...
+    return curve instanceof BSplineCurve3dBase || curve instanceof TransitionSpiral3d;
+  }
```
dassaf4, line 635:
```suggestion
  private needsStroking(curve?: AnyCurve): curve is BSplineCurve3dBase | TransitionSpiral3d {
    return curve && (curve instanceof BSplineCurve3dBase || curve instanceof TransitionSpiral3d);
  }
```

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
```diff
@@ -844,19 +806,8 @@ export class CurveCurveCloseApproachXY extends RecurseToCurvesGeometryHandler {
...
-        this._geometryB, this._extendB, this._geometryB.point0Ref, 0.0, this._geometryB.point1Ref, 1.0, this._extendB,
-        curve, this._extendA, true);
-    } else if (this._geometryB instanceof LineString3d) {
-      this.dispatchLineStringBSplineCurve(this._geometryB, this._extendB, curve, this._extendA, true);
-    } else if (this._geometryB instanceof Arc3d) {
-      this.dispatchArcBsplineCurve3d(this._geometryB, this._extendB, curve, this._extendA, true);
-    }
-    */
+    // NEEDS WORK -- make "dispatch" methods tolerant of both 3d and 3dH.
+    // https://github.com/iTwin/itwinjs-backlog/issues/2021
```
dassaf4, line 810: Prefer not to add URLs of private issues.

There are two URLs to another issue in CurveCurveIntersectXY.test.ts that can also be removed.

In both issues, indicate where to make the code changes.

```suggestion
```

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
```diff
@@ -715,46 +716,34 @@ export class CurveCurveIntersectXY extends RecurseToCurvesGeometryHandler {
...
             // We have a near intersection at fractions on the two beziers
             // Iterate on the curves for a true intersection
             const xyMatchingFunction = new CurveCurveIntersectionXYRRToRRD(bezierA, bezierB);
             const newtonSearcher = new Newton2dUnboundedWithDerivative(xyMatchingFunction);
             newtonSearcher.setUV(bezierAFraction, bezierBFraction);
-            if (newtonSearcher.runIterations()) {
-              bezierAFraction = newtonSearcher.getU();
-              bezierBFraction = newtonSearcher.getV();
-            }
+
```
dassaf4, line 733: Remove the blank line.

```suggestion
```

core/geometry/src/curve/internalContexts/CurveCurveIntersectXY.ts
dassaf4: Remove this link too (and update issue if necessary).

dassaf4 (approved):

core/geometry/src/curve/internalContexts/CurveCurveCloseApproachXY.ts
saeeedtorabi: I cannot do `return curve && (curve instanceof BSplineCurve3dBase || curve instanceof TransitionSpiral3d);` as this function cannot return undefined (only boolean). 

I can do `return (curve !== undefined) && (curve instanceof BSplineCurve3dBase || curve instanceof TransitionSpiral3d);` to return false if curve is undefined but first part is unnecessary.

My original code `return curve instanceof BSplineCurve3dBase || curve instanceof TransitionSpiral3d;` is fine as it returns false if curve is undefined.

dassaf4: My issue was with `instanceof` operating on `undefined`. But [apparently](https://www.typescriptlang.org/play/?#code/MYGwhgzhAECC0G9oF8BQpIwEKJagLgJ4AOAptAMLQC8c0APtFgNyqoBmArgHbD4CWAe27R2gwQAoAHgH4AXJQCUCqdH4x4jHAlTRoAJ1L5O+kdOgBCWjwAmpdv26kbi6ADI30c44j4wvUkF2OnpGVR8-AKCmRVY0Dh4+IREAIzB9aXklFTUNBiZEXTVgiQspV2BhCEEQUgA6EEEAcwkAIlVbe0dnC1bYosNjU2hw7l9-YEDgzTC1McjJ6JZUePQqmvrGpq8xSUV+yrGNhuad8QknAHc4fdZD6tqT7Yldi9JrrFu2e+OtrzSMl8fo8-hIAW9rrAgesQacwekITF+kA), this does not crash.

dassaf4:
Just needs a rushchange.bat I think.

Issue #8776, opened 2025-11-17
IModelDb.clearCaches() should call clearECDbCache()


soham-bentley:
Currently, IModelDb.clearCaches does not clear the native ECDb cache.

[Noted by @khanaffan here](https://github.com/iTwin/itwinjs-core/pull/8721#discussion_r2523716648)

Also as per discussion with @khanaffan  TxnManager._onChangesApplied should only clear instance caches, not the entire backend cache.

Fix:
IModelDb.clearCaches() should call clearECDbCache()
TxnManager._onChangesApplied should only clear instance caches
Fix tests accordingly

PR #8820, opened 2025-12-02
Clear ECDbCache in IModelDb.clearCaches()

MichaelSwigerAtBentley:
Moves the calls to `_nativeDb.clearECDbCache()` inside `iModelDb.ClearCaches()`
Removes unneeded calls to `_nativeDb.clearECDbCache()`
Updates `TxnManager._onChangesApplied()` to only clear instance Caches, instead of all backend caches

Fixes: https://github.com/iTwin/itwinjs-core/issues/8776

### DIFF
--- a/common/changes/@itwin/core-backend/soham-bentley-debugging-ecdb-hang_2025-12-03-07-23.json
+++ b/common/changes/@itwin/core-backend/soham-bentley-debugging-ecdb-hang_2025-12-03-07-23.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -23,8 +23,8 @@ importers:
         specifier: ^12.28.0
         version: 12.28.0
       '@bentley/imodeljs-native':
-        specifier: 5.5.6
-        version: 5.5.6
+        specifier: 5.5.7
+        version: 5.5.7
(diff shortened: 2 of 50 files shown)


core/backend/src/TxnManager.ts
```diff
@@ -699,7 +699,11 @@ export class TxnManager {
 
   /** @internal */
   protected _onChangesApplied() {
-    this._iModel.clearCaches();
+    // Should only clear instance caches, not all caches
+    this._iModel.elements[_instanceKeyCache].clear();
```
MichaelSwigerAtBentley: @khanaffan These 4 caches are the ones we need to clear during onChangesApplied, correct?

khanaffan: yes

MichaelSwigerAtBentley:
@khanaffan It seems like this change might be causing the core-full-stack-tests to become flaky and time out?

I've been looking to see if this is happening on other runs of the pipeline and it does not seem to be.
Something in "Per-model category visibility overrides" or "Viewstate attached to Viewport" seems to be going wrong and hanging.

I've seen it happen on Windows, Linux, and Mac, but it has also passed all three as well. Could the change to onChangesApplied be responsible?

Trying to reproduce now

MichaelSwigerAtBentley:
> @khanaffan It seems like this change might be causing the core-full-stack-tests to become flaky and time out?
> 
> I've been looking to see if this is happening on other runs of the pipeline and it does not seem to be. Something in "Per-model category visibility overrides" or "Viewstate attached to Viewport" seems to be going wrong and hanging.
> 
> I've seen it happen on Windows, Linux, and Mac, but it has also passed all three as well. Could the change to onChangesApplied be responsible?
> 
> Trying to reproduce now

Fixed by https://github.com/iTwin/itwinjs-core/pull/8823

PR #8823, opened 2025-12-03
Added test cases for ConcurrentQuery deadlock

soham-bentley:
imodel-native: https://github.com/iTwin/imodel-native/pull/1289

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -23,8 +23,8 @@ importers:
         specifier: ^12.28.0
         version: 12.28.0
       '@bentley/imodeljs-native':
-        specifier: 5.5.6
-        version: 5.5.6
+        specifier: 5.5.7
+        version: 5.5.7
       '@itwin/object-storage-azure':
         specifier: ^3.0.4
         version: 3.0.4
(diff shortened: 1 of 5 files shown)


full-stack-tests/presentation/src/backend/ReadWrite.test.ts
```diff
@@ -12,7 +12,7 @@ import { initialize, terminate } from "../IntegrationTests.js";
 import { collect, prepareOutputFilePath } from "../Utils.js";
 
 // Skipped until https://github.com/iTwin/itwinjs-core/issues/8751 is fixed
-describe.skip("ReadWrite", () => {
+describe("ReadWrite", () => {
```
grigasp:
```suggestion
describe("ReadWrite", () => {
```

core/backend/src/test/ecdb/ConcurrentQuery.test.ts
```diff
@@ -176,4 +177,60 @@ describe("ConcurrentQuery", () => {
...
+    // Start spamming simple queries to increase contention
+    promises.push(spam());
+    // Let queries start and establish contention
+    await new Promise(resolve => setTimeout(resolve, 1));
+
+    // eslint-disable-next-line no-console
+    console.log("Shutdown starting...");
+    ConcurrentQuery.shutdown(iModelDb[_nativeDb]);
+    // eslint-disable-next-line no-console
+    console.log("Shutdown complete...");
```
pmconne:
```suggestion
    ConcurrentQuery.shutdown(iModelDb[_nativeDb]);
```
You shouldn't need the console output any more, if the problem is fixed - right?

soham-bentley: I think keeping it would be beneficial for future. That's why I kept it
Should I remove it?

soham-bentley: And just to learn 
1) the config which I used to force the deadlock more, do they look okay?
 2) And I reset the  concurrentQuery config to default config after I shutdown the concurrentQuery, but that's fine, right?
 @pmconne  @grigasp  @khanaffan

pmconne: Why would it be beneficial? Nobody cares about the output unless the test starts failing, at which point they can add whatever debugging output is useful to them at that time.

soham-bentley: Removed

soham-bentley:
Parking this PR for a day because need to have a detailed discussion with @grigasp regarding the actual test which showed this deadlock
Already discussed with him regarding this halt

PR #8833, opened 2025-12-05
API to allow an Element to be moved across models

RohitPtnkr1996:
imodel-native: https://github.com/iTwin/imodel-native/pull/1300

Fixes: https://github.com/iTwin/itwinjs-backlog/issues/1481


core/backend/src/IModelDb.ts
```diff
@@ -2406,6 +2406,25 @@ export namespace IModelDb {
       }
     }
 
+    /** Move an element to a specific model.
+     * @param elementId The Id of the root element to move. Must not have a parent element.
+     * @param modelId The Id of the model to which to move the element.
+     * @note All child elements in the hierarchy are moved along with the root element.
+     * @throws [[ITwinError]] if unable to move the element to the target model.
+     */
+    public moveElementToModel(elementId: Id64String, modelId: Id64String): void {
```
GytisCepk: Why was a new method created instead of updating `updateElement` method to support this ?

RohitPtnkr1996: The updateElement has a specific check in place to block model id updates.
Also the effects of this might be too far reaching for a simple update element.
Every child of an element must belong to the same model as the parent.

I thought it better to have a dedicated call for a move operation as other child elements might also get affected, and also to avoid any possible accidents with updateElement.

kabentley:
I'd like to understand the use case for this. Who is asking and why? Moving an element between models is disallowed for good reasons. It also doesn't seem right to have this magically apply to assemblies.

RohitPtnkr1996:
> I'd like to understand the use case for this. Who is asking and why? Moving an element between models is disallowed for good reasons. It also doesn't seem right to have this magically apply to assemblies.

Diego has mentioned it on the work item for this PR https://github.com/iTwin/itwinjs-backlog/issues/1481.

kabentley:
> Diego has mentioned it on the work item for this PR https://github.com/iTwin/itwinjs-backlog/issues/1481.

I don't have access to the issues. Can you please explain it? Which Diego?

RohitPtnkr1996:
> > Diego has mentioned it on the work item for this PR [iTwin/itwinjs-backlog#1481](https://github.com/iTwin/itwinjs-backlog/issues/1481).
> 
> I don't have access to the issues. Can you please explain it? Which Diego?

Right sorry, I should've been more specific.
@diegoalexdiaz has mentioned on the work item that this change is for domain-specific authoring apps which know how to handle data in their own channels safely.

kabentley:
I spoke to @diegoalexdiaz about this. Please talk to him about how to proceed.

core/backend/src/Model.ts
```diff
@@ -254,6 +254,23 @@ export class Model extends Entity {
     arg.iModel.models[_cache].delete(arg.id);
   }
 
+  /** Called when an Element is about to be moved from an instance of a Model of this class to another Model.
+   * @note throw an exception to disallow the move
+   * @note If you override this method, you must call super.
+   * @note `this` is the class of the Model holding the element
+   * @beta
+   */
+  protected static onMoveElement(_arg: OnElementInModelIdArg, _targetModelId: Id64String): void { }
```
kabentley: i think this should be called something like onChangeModel or something. "MoveElement" sounds like it is moving geometrically.

kabentley: this is the place you should verify the channels of both models, that the exclusive lock is held on the element, and that the shared lock is held on the target model. You're going to need more arguments to this function.

core/backend/src/Model.ts
```diff
@@ -254,6 +254,23 @@ export class Model extends Entity {
...
+
+  /** Called after an Element has been moved from an instance of a Model of this class to another Model.
+   * @note If you override this method, you must call super.
+   * @note `this` is the class of the Model that held the element
+   * @beta
+   */
+  protected static onMovedElement(arg: OnElementInModelIdArg, _targetModelId: Id64String): void {
+    arg.iModel.models[_cache].delete(arg.id);
```
kabentley: I confess that I don't understand what the "models cache" is or what it's used for, but why does adding a new element cause it to be invalidated?

pmconne: It doesn't. The information in that cache does not directly change when an element is inserted.
If the element has geometry, then subsequently the cache will be automatically invalidated as a result of updating the geometry GUID.

core/backend/src/IModelDb.ts
```diff
@@ -2415,6 +2415,30 @@ export namespace IModelDb {
...
+     */
+    public changeElementModel(elementId: Id64String, modelId: Id64String): void {
+      try {
+        const elementProps = this.getElementProps<ElementProps>(elementId);
+        this[_cache].delete({
+          id: elementId,
+          federationGuid: elementProps.federationGuid,
+          code: elementProps.code
+        });
+        this._iModel[_nativeDb].changeElementModel(elementId, modelId);
```
kabentley:
```suggestion
        this._iModel[_nativeDb].changeElementModel(elementId, modelId);
        this.[_cache].delete({id: elementId});
```

core/backend/src/IModelDb.ts
```diff
@@ -2415,6 +2415,30 @@ export namespace IModelDb {
       }
     }
 
+    /** Move an element to a specific model.
+     * @param elementId The Id of the root element to move. Must not have a parent or child elements.
+     * @param modelId The Id of the model to which to move the element.
+     * @note This method will only move the element itself. It will not move any of its children or parent elements to the target model.
+     * @note The caller is responsible for moving the assembly and acquiring the necessary locks before the move.
+     * @throws [[ITwinError]] if unable to move the element to the target model.
```
kabentley:
```suggestion
    /** Change the model of an element.
     * @param elementId The Id of the element to change.
     * @param modelId The Id of the new model for the element.
     * @note This method will only change the element supplied, not its children.
     * @note An exclusive lock must be held on the element, and a shared lock must be held on the target model.
     * @throws [[ITwinError]] if unable to change the element to the target model.
```

core/backend/src/Model.ts
RohitPtnkr1996: Done !

core/backend/src/IModelDb.ts
RohitPtnkr1996: Done !

core/backend/src/IModelDb.ts
RohitPtnkr1996: Done !

RohitPtnkr1996:
As per the recent discussion on the BWG call, the API needs be a bulk move which changes the model of the child elements as well to avoid relaxing the parent-child must be in the same model rule.
Also, an idea of exploring this as a possible use case for the requested re-design of the editing api was floated. This is to ensure the change is atomic and can be rolled back on any failures that occur midway.
However, this PR contains the alternate approach which checks if any unsaved changes exist beforehand to then allow a rollback if the API throws an error.
This PR contains extensive tests as well which are covering all the requirements mentioned.
However, since there are still discussions to be had, and the requirements might change, I am closing this PR for now.
Will reopen and update once the requirements and approach are concrete.

PR #8837, opened 2025-12-05
Added ECClassHierarchy to manage a classes baseclasses and derived classes

StefanApfel-Bentley:
#### Description

This change moves the management of derived classes of an _ECClass_ and their baseclasses from _ECClass_ instances to the newly added _ECClassHierarchy_ class which lives on the schema context.

This is because the current implementation will not work with lazy loaded or incrementally loaded classes as it needs on deserialization time already the baseclass resolved which is contradicting the idea of a lazy loaded baseclass.

- The new class works entirely with schema keys instead of actual schema items.
- The ECClass methods _is_, _getDerivedClasses_ and _getAllBaseClasses_ now call the hierarchy class to get their information.
- Additionally incremental schema loading will benefit from the _addBaseClass_ method to add the known baseclasses to a class, even if the actual base class items may be defined in a referenced schema that is still loading.

**Note**: Some tests had to be adjusted as they did not add their test schemas to the current schema context or items to the schema. This was previously working, as the tested methods used the instance reference directly. Since the hierarchy now stores keys, the base/derived classes need to be resolved first before they can be returned, this would fail if the testschema/items has not been added to the context/schema.

### DIFF
--- a/common/changes/@itwin/core-backend/nam-rush-update-full_2025-12-12-18-30.json
+++ b/common/changes/@itwin/core-backend/nam-rush-update-full_2025-12-12-18-30.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/changes/@itwin/core-common/nam-rush-update-full_2025-12-12-18-30.json
+++ b/common/changes/@itwin/core-common/nam-rush-update-full_2025-12-12-18-30.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-common"
+}
--- a/common/changes/@itwin/core-frontend/nam-rush-update-full_2025-12-12-18-30.json
+++ b/common/changes/@itwin/core-frontend/nam-rush-update-full_2025-12-12-18-30.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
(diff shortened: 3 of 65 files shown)


core/ecschema-metadata/src/Context.ts
```diff
@@ -243,17 +244,24 @@ export class SchemaContext {
 
   private _knownSchemas: SchemaCache;
   private _fallbackLocaterDefined: boolean;
+  private _classHierarchy: ECClassHierarchy;
```
rschili, line 247: I'm not super familiar with how incremental works yet, Aurelija seems a bit unsure if this here is a good spot to hold the cache.
Suggestions:

- Is it a cache? In that case we should reflect that in the name.
- The empty field on context will not really hurt us when not using incremental loading, but it also seems a bit unclean. Maybe create a subclass of context which is being used for incremental loading?
- Context seems like a good place for this to me. Aurelija seemed to favor putting it on schema, which makes me wonder if putting this on class would be an option as well? Given my limited overview, context does seem fine to me

StefanApfel-Bentley: It was on the class before, but that doesn't work if we lazy load the classes properly (not like now were we actually load the class before and only return it in the lazy loaded promise (other pending PR). We originally decided to not subclass anything for the incremental schema loading.

I can give i a try on the schema which would be a little less global than the context. I can add a cache suffix too, but it's not really caching the classes, it just manages the derived and baseclasses for a class by schema item key for faster lookup.

The field is not only used for incremental schema loading, classes already have the ability to lookup base and derived classes now, but with the current implementation, we cannot make use of it unless the schema and it's references are loaded entirely.

StefanApfel-Bentley: I checked moving the ClassHierarchy to the schema and we run into an issue when we add baseclasses of a different schema. In this case, the hierarchy entry would need to add the entry on the classhierachy on the referenced schema, but since that schema might be still loading (incrementally), we might end up waiting here unnecessarily to get the schema instance from the context.

So I tend to keep it on the context for now.

rschili: Sounds good. Alright, resolving the comment

PR #8838, opened 2025-12-05
SchemaItems lazy loaded lookup logic

StefanApfel-Bentley:
#### Description

Improve LazyLoaded lookup logic. In the current logic, the items that shall be lazy loaded, were resolved before the delayed promise was created which then returned the resolved value. This contradicts actually with the idea of lazy loaded items.

Now the schema items will be resolved in the lazy loaded promise instead of outside of it.

Note: KindOfQuantity presentation formats aren't truly lazy loaded as OverrideFormats can't be lazy loaded. That should be changed in itwinjs 6.

### DIFF
--- a/common/changes/@itwin/ecschema-metadata/stefan-LazyLoadedLookupLogic_2025-12-16-10-20.json
+++ b/common/changes/@itwin/ecschema-metadata/stefan-LazyLoadedLookupLogic_2025-12-16-10-20.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-metadata",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/ecschema-metadata"
+}


core/ecschema-metadata/src/Metadata/EntityClass.ts
```diff
@@ -235,13 +236,13 @@ protected override async buildPropertyCache(): Promise<Map<string, Property>> {
           throw new ECSchemaError(ECSchemaStatus.InvalidECJson, `The ECEntityClass ${this.name} has a mixin ("${name}") that cannot be found.`);
```
AureMonke: Check is not needed.

core/ecschema-metadata/src/Metadata/RelationshipClass.ts
```diff
@@ -340,17 +340,14 @@ export class RelationshipConstraint implements CustomAttributeContainerProps {
...
-        async () => {
-          const tempAbstractConstraint = await relClassSchema.lookupItem(relationshipConstraintProps.abstractConstraint!);
-          if (undefined === tempAbstractConstraint ||
-            (!EntityClass.isEntityClass(tempAbstractConstraint) && !Mixin.isMixin(tempAbstractConstraint) && !RelationshipClass.isRelationshipClass(tempAbstractConstraint)))
-            throw new ECSchemaError(ECSchemaStatus.InvalidECJson, `Unable to locate the abstractConstraint ${relationshipConstraintProps.abstractConstraint}.`);
-
-          return tempAbstractConstraint;
-        }));
+      this.setAbstractConstraint(new DelayedPromiseWithProps<SchemaItemKey, AnyConstraintClass>(abstractConstraintSchemaItemKey, async () => {
+        const tempAbstractConstraint = await relClassSchema.lookupItem(relationshipConstraintProps.abstractConstraint!);
```
AureMonke: No need for not nullable assertion for relationshipConstraintProps.abstractConstraint

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -536,30 +536,15 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
-              throw new ECSchemaError(ECSchemaStatus.InvalidECJson, `Unable to locate the baseClass ${classProps.baseClass}.`);
-            return baseItem;
-          });
-      } else {
-        lazyBase = new DelayedPromiseWithProps<SchemaItemKey, ECClass>(ecClassSchemaItemKey,
-          async () => {
-            return baseClass as ECClass;
-          });
-      }
+      const baseClassItemKey = this.schema.getSchemaItemKey(classProps.baseClass);
```
rschili, line 539: In case of normal loading, the base class will already exist in memory, but we defer the lookup until it is accessed. That means if the base class does not exist, we will now successfully load the schema and fail once we need to use it, for example when iterating this classes properties.

That is quite a tradeoff on the normal loading scenario. I'd say the previous behavior was better in case of normal schema loading. For incremental schema loading I can see how this is a big improvement.

Let's discuss and think about it a bit more. Maybe the answer here is to create a different code path for the incremental loading.

StefanApfel-Bentley: good point. We thought about something like that before too, but we didn't wanted to introduce another loading method on the items or have a doublicate of the reader which does almost everything the same except for creating the delayed promise in the old or new way.

Let me check again though if there is another way, without have lots of code duplicates.

core/ecschema-metadata/src/Metadata/RelationshipClass.ts
```diff
@@ -362,12 +359,12 @@ export class RelationshipConstraint implements CustomAttributeContainerProps {
     };
```
AureMonke, line 359: Add exception message. We should await when we looking for constraintClass

core/ecschema-metadata/src/Metadata/KindOfQuantity.ts
```diff
@@ -70,17 +71,37 @@ export class KindOfQuantity extends SchemaItem {
...
+    for (const formatString of presUnitsArr) {
+      const presFormatOverride = OverrideFormat.parseFormatString(formatString);
       const format = await this.schema.lookupItem(presFormatOverride.name, Format);
       if (undefined === format || format.schemaItemType !== SchemaItemType.Format)
         throw new ECSchemaError(ECSchemaStatus.InvalidECJson, `Unable to locate Format '${presFormatOverride.name}' for the presentation unit on KindOfQuantity ${this.fullName}.`);
 
       if (undefined === presFormatOverride.precision && undefined === presFormatOverride.unitAndLabels) {
-        this.addPresentationFormat(new DelayedPromiseWithProps(format.key, async () => format));
+        this.addPresentationFormat(new DelayedPromiseWithProps(format.key, async () => {
+          const formatItem = await this.schema.lookupItem(format.key, Format);
```
AureMonke, line 99: In line 93 we already found format. Should be enough to return format.id.

PR #8840, opened 2025-12-08
Add Previous Minor Branch to Security Fix Backports

andremig-bentley:
This PR adds the ability to accommodate automatically backporting to a previous minor branch as well as the current minor and previous major. This is accomplished by adding the previous minor branch to the backport rule in mergify.yml as well as updating the logic in version-bump.yaml to update the branch names on version bumps. 

The branch names have been moved to being held in variables at the top of the mergify.yaml file to accommodate major version bumps. In that case, the previous minor variable can be updated to `null` as stated in the instructions for a major version bump in version-bump.yaml. Note that as a pure yaml file, `mergify.yml` must use anchors and aliases to define variables as simpler variable definitions are not supported in pure yaml (as they are in azure devops pipeline definition yaml files for example). 

The branch name update logic has also been moved within `version-bump.yaml` to be completed as soon as a new release branch is created, and it will immediately be pushed to master. Note that there is also `continueOnError: true` on this step so that a failure in updating the mergify branch names does not interfere with completing the more important version bump.

### DIFF
--- a/common/config/rush/pnpm-config.json
+++ b/common/config/rush/pnpm-config.json
@@ -17,7 +17,8 @@
     "browserslist": "latest", // https://github.com/browserslist/update-db#readme
     "fast-xml-parser": "^5.3.6", // https://github.com/advisories/GHSA-jmr7-xgp7-cmfj @google-cloud/storage > fast-xml-parser
     "glob@>=11.0.0 <11.1.0": "^11.1.0", // https://github.com/advisories/GHSA-5j98-mcp5-4vw2 cpx2>glob
-    "axios@<1.0.0": "^1.13.5" // https://github.com/advisories/GHSA-43fc-jf86-j433
+    "axios@<1.0.0": "^1.13.5", // https://github.com/advisories/GHSA-43fc-jf86-j433
+    "serialize-javascript": "^7.0.3" // https://github.com/advisories/GHSA-5c6j-r48x-rmvq mocha>serialize-javascript (related to CVE-2020-7660)
   },
   // A list of temporary advisories excluded from the High and Critical list.
   // Warning this should only be used as a temporary measure to avoid build failures
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -13,6 +13,7 @@ overrides:
   fast-xml-parser: ^5.3.6
   glob@>=11.0.0 <11.1.0: ^11.1.0
   axios@<1.0.0: ^1.13.5
+  serialize-javascript: ^7.0.3
 
 pnpmfileChecksum: fm2bgttvmfglngapbjspr7zf2u
 
@@ -881,7 +882,7 @@ importers:
         version: 17.0.2
       '@vitest/browser':
         specifier: ^3.0.6
-        version: 3.0.6(@types/node@24.10.15)(playwright@1.56.1)(typescript@5.6.2)(vite@6.4.1(@types/node@24.10.15)(terser@5.46.0)(yaml@2.8.2))(vitest@3.0.6)
+        version: 3.0.6(@types/node@24.11.0)(playwright@1.56.1)(typescript@5.6.2)(vite@6.4.1(@types/node@24.11.0)(terser@5.46.0)(yaml@2.8.2))(vitest@3.0.6)
       '@vitest/coverage-v8':
         specifier: ^3.0.6
         version: 3.0.6(@vitest/browser@3.0.6)(vitest@3.0.6)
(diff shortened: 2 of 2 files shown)


common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -143,12 +142,36 @@ jobs:
...
+        displayName: Update mergify.yml variables on master
+        continueOnError: true
+
   - job: Bump
     displayName: Bump Version
     dependsOn:
     - CreateBranch
     - CheckPrevCommit
-    condition: and(in(dependencies.CreateBranch.result, 'Succeeded', 'Skipped'), in(dependencies.CheckPrevCommit.result, 'Succeeded', 'Skipped'))
+    condition: and(in(dependencies.CreateBranch.result, 'Succeeded', 'Skipped', 'SucceededWithIssues'), in(dependencies.CheckPrevCommit.result, 'Succeeded', 'Skipped'))
```
DanRod1999, line 186: whats the reason for the SucceededWithIssues?

andremig-bentley: It's to make sure the version bump doesn't get held up by a failed bump to the branch names. I added continue on error to the branch name update steps, but if that fails and everything else succeeds, the create branch job will finish with the succeeded with issues status. It seemed to me that it's not worth holding up the entire version bump just because of the failed branch name bump, so this way it can continue.

DanRod1999: is there any reason that this would fail, and we would need to do something about it manually, or does failure mean no change is needed?

andremig-bentley: The main scenario where this would fail (and the one I was thinking of when I added continueOnError) is if something isn't done correctly in the manual update step for a major version bump (talked about in the added comments of version-bump.yaml). If a typo was introduced and we can't update a variable correctly, then that might cause a failure. In that case something would need to be done about it manually. It's not a huge risk imo but it might potentially be a problem.

tcobbs-bentley: This is now documented.

common/config/azure-pipelines/jobs/version-bump.yaml
```diff
@@ -114,12 +114,11 @@ jobs:
...
           # additionally if major version bump, the `mergify.yml` also needs to be edited manually
-          # if this release is `release/5.0.x`, the branches under actions/backport should be updated to look like:
-          #     branches:
-          #       # current minor release branch
-          #       - "release/5.0.x"
-          #       # previous major release branch
-          #       - `release/4.<whatever_last_minor_release_version_was>.x`
+          # if this release is `release/5.0.x`, the branch variables in `mergify.yml` should be updated to look like:
+          # current_minor: &current_minor "release/5.0.x"
+          # previous_minor: &previous_minor null
```
MichaelSwigerAtBentley, line 119: So the mergify.yml will need to be manually updated on major releases? Other than that, the logic below will copy the current minor to previous minor when it is ran?

DanRod1999: based on line 162 it might error out on major release? We continue on error if that happens, not sure I love that behavior but not the worst thing

tcobbs-bentley: I'm open to suggestions, but I think making it automatically handle major version updates is more trouble than it's worth.

MichaelSwigerAtBentley: Yeah, I don't see a big issue with needing to manually update it for majors, just need to make sure that its documented in the release process. Just wanted to make sure I understood the logic correctly

.github/mergify.yml
```diff
@@ -1,5 +1,11 @@
 # Configuration file for mergify
 
+# Branch version variables
+variables:
+  current_minor: &current_minor "release/5.5.x"
```
hl662:
```suggestion
  current_minor: &current_minor "release/5.6.x"
```

.github/mergify.yml
```diff
@@ -1,5 +1,11 @@
...
+  previous_minor: &previous_minor "release/5.4.x"
```
hl662:
```suggestion
  previous_minor: &previous_minor "release/5.5.x"
```

.github/mergify.yml
```diff
@@ -1,5 +1,11 @@
 # Configuration file for mergify
 
+# Branch version variables
+variables:
+  current_minor: &current_minor "release/5.7.x"
+  previous_minor: &previous_minor "release/5.5.x"
```
aruniverse:
```suggestion
  previous_minor: &previous_minor "release/5.6.x"
```

.github/mergify.yml
```diff
@@ -1,5 +1,11 @@
 # Configuration file for mergify
 
+# Branch version variables
+variables:
+  current_minor: &current_minor "release/5.7.x"
+  previous_minor: &previous_minor "release/5.6.x"
```
anmolshres98, line 6: idk how the master merge made this change but this is not correct. current minor is still 5.6, 5.7 is only rc official 5.7 will be released later today

```suggestion
  current_minor: &current_minor "release/5.6.x"
  previous_minor: &previous_minor "release/5.5.x"
```

anmolshres98: Oh wait, I guess the variable is named somewhat misleadingly. it refers current and previous branches and not the official released version itself. if that is the case, you can resolve my comment and move forward

DanRod1999:
My only concern is this is a lot of updates to a yml file (thats already a little complicated) where files are directly modified (which I'm generally cautious of), and we need to add an additional step to our major releases. The only benefit is - we don't need to call the the mergify command directly anymore? However, someone still needs to remember to add the security tag to their PRs. 

I guess what I'm saying is, we are adding potentially more margins of error to this pipeline, and I don't know if I see the value add is that significant, security tag vs mergify cmd. If this code has been tested and we're fairly confident it won't trip us up in the future, then seems like a fine feature. I've just been burned by my own yml enough to be cautious about changes to release pipelines.

ben-polinsky:
> My only concern is this is a lot of updates to a yml file (thats already a little complicated) where files are directly modified (which I'm generally cautious of), and we need to add an additional step to our major releases. The only benefit is - we don't need to call the the mergify command directly anymore? However, someone still needs to remember to add the security tag to their PRs.
> 
> I guess what I'm saying is, we are adding potentially more margins of error to this pipeline, and I don't know if I see the value add is that significant, security tag vs mergify cmd. If this code has been tested and we're fairly confident it won't trip us up in the future, then seems like a fine feature. I've just been burned by my own yml enough to be cautious about changes to release pipelines.

Good thoughts. Thanks. I tend to agree this isn't a huge time saver, but we're already doing some of this, right? We already do this for the previous minor and the previous major. I think it's fine to see how we do with this and revert/look for other solutions if not.

anmolshres98:
I am in favor of getting this in as initial implementation and reiterating on this solution as we do more releases/backports. I would rather start with an imperfect implementation and reiterate frequently in small increments.

PR #8841, opened 2025-12-08
Add previous-minor Release Tag

andremig-bentley:
Currently when patching a previous minor branch (for example, release/5.2.x when the latest is release/5.3.x), no release tag is added when publishing packages to npm. This then means that npm automatically adds the "latest" tag to the package even though it is not actually the latest version. To fix this, the "previous-minor" tag was created for patching previous minors. Open to discussion on other names for this tag, however "previous" is already applied to previous major versions so that can't be used. "lts" (long term support) is used often and could be applied to either the previous minor or previous major if desired.

### DIFF
--- a/common/api/presentation-common.api.md
+++ b/common/api/presentation-common.api.md
@@ -24,7 +24,7 @@ import { UnitSystemKey } from '@itwin/core-quantity';
 export function addFieldHierarchy(rootHierarchies: FieldHierarchy[], hierarchy: FieldHierarchy): void;
 
 // @public
-export class ArrayPropertiesField extends PropertiesField {
+export class ArrayPropertiesField extends PropertiesField implements ArrayPropertiesFieldProps, IFieldsSource {
     // @deprecated
     constructor(category: CategoryDescription, name: string, label: string, type: TypeDescription, itemsField: PropertiesField, isReadonly: boolean, priority: number, properties: Property[], editor?: EditorDescription, renderer?: RendererDescription);
     constructor(props: ArrayPropertiesFieldProps);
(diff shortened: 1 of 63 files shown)


DanRod1999:
logic looks good. Is the tag shown the one that will be on npm? Also I assume we don't need to change any other pipelines to respond to this new tag? This tag is passed all the way down to publish step?

andremig-bentley:
> logic looks good. Is the tag shown the one that will be on npm? Also I assume we don't need to change any other pipelines to respond to this new tag? This tag is passed all the way down to publish step?

Right, this tag is the one that will be shown on npm, and it should be passed down automatically to the publish step.

MichaelSwigerAtBentley:
Any objections to "previous-minor" as the name of the tag?

Seems good to me since we already use "previous"

GytisCepk:
> Any objections to "previous-minor" as the name of the tag?
> 
> Seems good to me since we already use "previous"

Should we just use "previous" tag instead of "previous-minor" ? I don't think, "previous" tag serves any real purpose now.

My problem with "previous-minor" is more of a nitpick, but what if we don't need to backport anything to previous minor for a few minor versions ? "previous-minor" would be stuck on something like 5.5.x while 5.8.x is out, which means actual previous minor would be 5.7.x. Again, tags are close to meaningless in my opinion (with exception of "latest"), so this may not matter.

aruniverse:
> > Any objections to "previous-minor" as the name of the tag?
> > Seems good to me since we already use "previous"
> 
> Should we just use "previous" tag instead of "previous-minor" ? I don't think, "previous" tag serves any real purpose now.
> 
> My problem with "previous-minor" is more of a nitpick, but what if we don't need to backport anything to previous minor for a few minor versions ? "previous-minor" would be stuck on something like 5.5.x while 5.8.x is out, which means actual previous minor would be 5.7.x. Again, tags are close to meaningless in my opinion (with exception of "latest"), so this may not matter.

Valid edge case, however with our evergreen strategy, we will likely always be updating the prev minor

PR #8848, opened 2025-12-10
Make verticalCRS.id optional

AlainRobertAtBentley:
Make verticalCRS id optional to eventually make crsName the primary identification.

### DIFF
--- a/common/api/presentation-common.api.md
+++ b/common/api/presentation-common.api.md
@@ -24,7 +24,7 @@ import { UnitSystemKey } from '@itwin/core-quantity';
 export function addFieldHierarchy(rootHierarchies: FieldHierarchy[], hierarchy: FieldHierarchy): void;
 
 // @public
-export class ArrayPropertiesField extends PropertiesField {
+export class ArrayPropertiesField extends PropertiesField implements ArrayPropertiesFieldProps, IFieldsSource {
     // @deprecated
     constructor(category: CategoryDescription, name: string, label: string, type: TypeDescription, itemsField: PropertiesField, isReadonly: boolean, priority: number, properties: Property[], editor?: EditorDescription, renderer?: RendererDescription);
     constructor(props: ArrayPropertiesFieldProps);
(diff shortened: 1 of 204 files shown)


AlainRobertAtBentley:
> This is a breaking change to a public API and it doesn't appear to have an actual effect. What is the ultimate intended change?

The id property that used to contain the verticalCRS identifier will now contain the "Legacy identifier". A new identifier will be introduced later though we still want to support the id property.

Since this is an API breaking change, please tell me what should be the target branch for the pull request. Other changes will follow though the next changes will require modifications to imodel-native.

pmconne:
Per our [API support policies](https://www.itwinjs.org/learning/api-support-policies/) you'd have to deprecate it now and keep it functional for at least a year or until the next major version release, whichever comes later.

Can you link to an issue that your proposed changes are intended to resolve?

AlainRobertAtBentley:
> Per our [API support policies](https://www.itwinjs.org/learning/api-support-policies/) you'd have to deprecate it now and keep it functional for at least a year or until the next major version release, whichever comes later.
> 
> Can you link to an issue that your proposed changes are intended to resolve?

I expect the support for the optional id property will be retained for years to come.

I have introduced the issue to guide current and following pull requests concerning Vertical CRS support (https://github.com/iTwin/itwinjs-core/issues/8851)

aruniverse:
> This is a breaking change to a public API

While "breaking", this does widen the scope and makes the api more flexible, so should it really be considered breaking?

pmconne:
> While "breaking", this does widen the scope and makes the api more flexible, so should it really be considered breaking?

According to our API support policies, yes. Changing it can cause code that used to compile to stop compiling, or worse, throw errors at run-time. Those are exactly the problems our policies are designed to prevent, and preventing them is the only way to make our "evergreen" strategy actually successful.

ben-polinsky:
> > While "breaking", this does widen the scope and makes the api more flexible, so should it really be considered breaking?
> 
> According to our API support policies, yes. Changing it can cause code that used to compile to stop compiling, or worse, throw errors at run-time. Those are exactly the problems our policies are designed to prevent, and preventing them is the only way to make our "evergreen" strategy actually successful.

A new type could be added to the string union to indicate that the new identifier should be utilized. While technically breaking, [we note that this could be acceptable](https://www.itwinjs.org/learning/api-support-policies/#breaking-changes) if we don't think anyone is enumerating the values. Which is a bit perilous.

Could also introduce a new boolean to indicate the new identifier should be used.

AlainRobertAtBentley:
@pmconne Is there something else to be done concerning this pull request?

pmconne:
> @pmconne Is there something else to be done concerning this pull request?

Yes. Make it not break the public API.

AlainRobertAtBentley:
> > @pmconne Is there something else to be done concerning this pull request?
> 
> Yes. Make it not break the public API.

The purpose of the pull request was to ease an eventual deprecation or the 'id' property but I see now that will not be possible by changing the API. I will simply abandon the pull request. 
By the way, the documentation (https://www.itwinjs.org/learning/api-support-policies/) does not explicitly indicate how to process the deprecation of an individual property. It only shows examples of deprecation of methods or interfaces. I assume adding a deprecation comment before the property is sufficient. Anyway this will be in another future pull request.

PR #8849, opened 2025-12-10
Remove no longer used @loaders.gl/core npm package

tcobbs-bentley:


tcobbs-bentley:
@markschlosseratbentley I'm pretty sure that the `@loaders.gl/core` npm package is no longer needed, but I could be wrong. I did run the tests and integration tests locally, and they all passed. If this is not something that we should be removing, let me know, and I will close this PR.

markschlosseratbentley:
> @markschlosseratbentley I'm pretty sure that the `@loaders.gl/core` npm package is no longer needed, but I could be wrong.

Draco-compressed glTF model still loads in DTA for me.  Download that model here: [CesiumManDraco.zip](https://github.com/user-attachments/files/24103057/CesiumManDraco.zip)

Screenshot:

image

markschlosseratbentley (approved):
LGTM - see my comment about my testing.

@pmconne please chime in if anything seems amiss that we're not considering.

PR #8850, opened 2025-12-11
Improve mechanism for enabling txns on standaloneDb

rschili:
imodel-native: https://github.com/iTwin/imodel-native/pull/1302

See comment on the imodel-native PR.
The old allowEdit flag is a string and needed to be set to a magic json string that was undocumented, to turn on txns for standaloneDb.
Since the flag is public, I am not deprecating it immediately, instead I have changed the logic as Paul suggested: Setting it to ANY value, will make it use the magic string.
Also I introduced a second safe flag alongside which will eventually replace allowEdit in the next major release. Both flags will work for now.

### DIFF
--- a/core/backend/src/IModelDb.ts
+++ b/core/backend/src/IModelDb.ts
@@ -3941,6 +3941,7 @@ export class StandaloneDb extends BriefcaseDb {
     nativeDb.createIModel(filePath, args);
     // Handle both the legacy allowEdit string and new enableTransactions boolean
     // If either is truthy, set the magic JSON string required by the native layer
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
     const shouldEnableTransactions = args.enableTransactions || args.allowEdit;
     if (shouldEnableTransactions)
       nativeDb.saveLocalValue(BriefcaseLocalValue.StandaloneEdit, `{ "txns": true }`);


core/common/src/IModel.ts
```diff
@@ -195,8 +195,18 @@ export type CreateEmptySnapshotIModelProps = CreateIModelProps & CreateSnapshotI
...
+   * internal value `{ "txns": true }` regardless of the string content provided. This behavior ensures backwards
+   * compatibility while allowing users to enable transaction support.
+   * @note Consider using [[enableTransactions]] instead, which provides a cleaner boolean interface.
+   */
   readonly allowEdit?: string;
+  /** If true, enables transactions in the standalone iModel, allowing local editing. The iModel cannot be used to create changesets.
+   * This is the preferred way to enable editing. If both this and [[allowEdit]] are specified, either being truthy will enable transactions.
+   * @beta
+   */
+  readonly enableTransactions?: boolean;
```
pmconne, line 209:
```suggestion
  /** If present, file will allow local editing, but cannot be used to create changesets.
   * @note This flag is a string for historical reasons. When specified, it will always be set to the required
   * internal value `{ "txns": true }` regardless of the string content provided. This behavior ensures backwards
   * compatibility while allowing users to enable transaction support.
   * @note Consider using [[enableTransactions]] instead, which provides a cleaner boolean interface.
   * @deprecated in 5.5. Use [[enableTransactions]] instead.
   */
  readonly allowEdit?: string;
  /** If true, enables transactions in the standalone iModel, allowing local editing. The iModel cannot be used to create changesets.
   * This is the preferred way to enable editing. If both this and [[allowEdit]] are specified, either being truthy will enable transactions.
   */
  readonly enableTransactions?: boolean;
```

> Since the flag is public, I am not deprecating it immediately.

Can you elaborate? Deprecation is only meaningful for public APIs and you've provided a superior alternative. Why delay?

rschili: Deprecated as suggested.
I was under the impression we don't want to do that unless we release a major version, but if you consider it okay that's more than enough.

DanRod1999:
This PR merged, however I see there is a linked imodel native PR but no new addon was added to this PR. This is mostly due to the addon you created was not published due to errors in the addon publish pipeline (working on fixing that). I'm confused how this was able to merge without a new addon, but my guess is these PRs weren't really dependent on each other?

Also because of this I think its best practice to not link PRs unless they truly are dependent on each other. @aruniverse, I know you've spoken against this before, but wondering on your thoughts with this example. Technically speaking we have a itwinjs changes that's connected to an imodel-native change, but due to issues with the addon publishing it was able to merge without a new addon. If these PRs weren't linked the new addon would just go to master branch, but in the case the new version is not pushed to any itwin branches.

rschili:
> This PR merged, however I see there is a linked imodel native PR but no new addon was added to this PR. This is mostly due to the addon you created was not published due to errors in the addon publish pipeline (working on fixing that). I'm confused how this was able to merge without a new addon, but my guess is these PRs weren't really dependent on each other?

Correct, they did not depend on each other, I just linked them because they were related.
I got an email with a failed addon publishing job, and I was wondering if that would become a problem further down the chain :-)

DanRod1999:
Gotcha thanks for clarifying. Not exactly a problem, but I did have concerns with where that addon should be once published and didn't want to break any branches when the addon is eventually brought it.

I'm wondering if related, but not dependent, PRs should just reference each other but not "linked". This way reviewers can get context but the addon won't necessarily be tethered to a specific branch/PR. I'll create an issue about this so others can weigh in, thanks again.

PR #8881, opened 2025-12-29
Store margins as factors of textheight

SheetalNair:
TextBlockMargin values are now considered as factors that get multiplied to textHeight.
- updated geometry code to multiply margin values by textHeight 
- Added textStyle migration code to convert any absolute margin values in previous versions to factors of textheight
- updated tests to accomodate this change

### DIFF
--- a/common/config/rush/version-policies.json
+++ b/common/config/rush/version-policies.json
@@ -2,7 +2,7 @@
   {
     "policyName": "prerelease-monorepo-lockStep",
     "definitionName": "lockStepVersion",
-    "version": "5.6.0-dev.10",
+    "version": "5.6.0-dev.11",
     "nextBump": "prerelease"
   }
 ]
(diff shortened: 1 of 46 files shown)


SheetalNair:
@copilot

pmconne (commented):
I expected to see a new test that confirms that data stored using v1.0.1 would be correctly upgraded to v1.0.2 when extracted from the db. Did I overlook it?

SheetalNair:
> I expected to see a new test that confirms that data stored using v1.0.1 would be correctly upgraded to v1.0.2 when extracted from the db. Did I overlook it?

Added test https://github.com/iTwin/itwinjs-core/pull/8881/changes#diff-01f5bf6c76722886ddbe84ef801367b477fb5a6c4f5c6d9ca47044c7dbe691efR1000

pmconne (approved):

Issue #7523, opened 2025-01-07
Allow callback on schema import to let upgrade data for channel when schema evolves


khanaffan:
Channel might like to reorganize it data when schema is evolved. 

The actual mechanism need to be sorted out. But it may require two briefcases so app can do lazy conversion.

Another alternative would be to read before data into memory and apply changes to briefcase after schema is imported. But that will work if transformation require involve small set of data.

But generally, if it major changes to channel we might want to take schema lock or if its minor we might stick with locking.  

1. Before schema event that notify app
   - In this event we can choose strategy
     - Clone briefcase so full read access to before state of the briefcase.
     - Do not clone instead just read before data into memory if needed.   
3. Schema is imported.
4. After schema import changes are applied to channel which would include moving data around.
5. Finalize changes and push changeset out & release locks.

Rock2000:
Do we need a 2.5 step, for a callback to upgrade schema based on schema versions? i.e. a new field was added to schema, run manual code to set default values for them.

This feels different than running a channel update, which seems likely to be something like a data move operation. i.e. Definitions in root definition model should be moved to the correct definition model.

Or is there coupling between the two when updating. i.e. App was at schema 1.0.4, channel 1.0.1, then change was made to fix the definition data location which is made at channel 1.0.2, then a later app moved to schema 1.0.5 which has a custom schema upgrade needed. Presumably that code, if it refers to definitions, probably assumes the channel structure is 1.0.2. But if updating a model from the first state (1.0.4 and 1.0.1) would we have done the updates in the right order? So the channel 1.0.2 update was done before the schema 1.0.5 custom update? Or am I completely lost.

PR #8887, opened 2026-01-06
Allow callback on schema import to let upgrade data for channel when schema evolves

RohitPtnkr1996:
## Summary
Adds optional callbacks to the schema import API to allow data transformation and channel reorganization before and after schema import.
Fixes: https://github.com/iTwin/itwinjs-core/issues/7523

### Schema Import Callbacks
- **Pre-import callback**: Execute custom logic before schema import
  - Can cache data or create an iModel snapshot for use in post-import lazy processing.
  - User must acquire necessary locks
  
- **Post-import callback**: Execute custom logic after schema import
  - Access cached data or snapshot from pre-import callback for lazy conversions
  - No explicit lock acquisition needed (schema import already holds full schema lock)

### Channel Upgrade Callback

- Optional callback to re-organize channel before the schema import.
- Allows channel reorganization to match the expectations from the schema import and the pre- and post-import callbacks.
- Note: While semantic versioning for channels has been released in BisCore 1.0.17, the specific API for it is not yet available.

### Execution Order
Channel Upgrade Callback -> Pre-Schema Import Callback -> Schema Import -> Post- Schema Import Callback.

## Pending work:
- Performance testing needed for snapshot strategy with large iModels to establish baseline metrics.

### DIFF
--- a/common/changes/@itwin/core-backend/master_2026-01-09-17-02.json
+++ b/common/changes/@itwin/core-backend/master_2026-01-09-17-02.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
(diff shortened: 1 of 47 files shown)


khanaffan:
It look good but let other review this PR as well.

MichaelSwigerAtBentley:
I've been reading this over as well. It looks good to me so far. 

Could you add a brief PR description

PR #8892, opened 2026-01-09
Improvement to how schema changeset synchronize sqlite table definition

khanaffan:
imodel-native: https://github.com/iTwin/imodel-native/pull/1314

## How Does This Work?

1.  After importing the EC schema and completing the mapping, we update the SQLite schema by creating any missing tables and columns.
2.  We capture the DDL generated during this process and include it in the changeset when it is pushed.
3.  When applying the changeset, the following steps occur:
    *   Apply table and column changes to the SQLite database. This ensures that applying data changes does not fail due to missing columns.
    *   Separate out `ec_*` changes and apply them first. After this, we call `AfterSchemaChangeSetApplied()`.
        *   If we are in schema sync mode, this step creates any missing tables and columns using `ec_*` table information.
        *   If in schema sync mode, DDLs may or may not be available.
    *   Finally, apply data changes, which should succeed since all required tables and columns are now present.

## What Has Changed?

1.  By default, we now sync the SQLite schema using `ec_*` tables, similar to what was previously done only in schema sync mode.
2.  We no longer fail on `ApplyDDLChanges()`. Instead, we attempt to continue. This method remains for backward compatibility, primarily for non-`ec_*` tables.
3.  Added an option to purge unused columns (it only attempt to delete columns and may fail to drop some due to sqlite limitation). Though failure to drop does not fail the operation. 
4. Another change I have is switching from ` sqlite3changeset_apply_v2_strm()` to `sqlite3changeset_apply_v3_strm()` which is use to more narrowly filter schema changes from data changes as it allow filter by each change instead of just table name.

### DIFF
--- a/common/changes/@itwin/core-backend/2026-01-13-21-20.json
+++ b/common/changes/@itwin/core-backend/2026-01-13-21-20.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/version-policies.json
+++ b/common/config/rush/version-policies.json
@@ -2,7 +2,7 @@
   {
     "policyName": "prerelease-monorepo-lockStep",
     "definitionName": "lockStepVersion",
-    "version": "5.6.0-dev.7",
+    "version": "5.6.0-dev.8",
     "nextBump": "prerelease"
   }
 ]
--- a/core/backend/package.json
+++ b/core/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-backend",
-  "version": "5.6.0-dev.7",
+  "version": "5.6.0-dev.8",
   "description": "iTwin.js backend components",
   "main": "lib/cjs/core-backend.js",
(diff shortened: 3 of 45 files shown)


rschili (approved):

core/backend/src/SqliteChangesetReader.ts
```diff
@@ -428,4 +428,11 @@ export class SqliteChangesetReader implements Disposable {
   public [Symbol.dispose](): void {
     this.close();
   }
+  /**
+   * Get DDL changes
```
MichaelSwigerAtBentley:
```suggestion
   * Get Data Definition Language changes
```

common/changes/@itwin/core-backend/affank-table-sync_2026-01-09-16-09.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Add test for how sqlite schema is updated",
```
MichaelSwigerAtBentley:
```suggestion
      "comment": "Add function for retrieving changes to Data Definition Language from native reader",
```

No need to mention tests in changelog

ben-polinsky (approved):

ColinKerr (approved):

Issue #8893, opened 2026-01-09
Documentation Issue: Code snippet does not show up correctly


khanaffan:
This is likely due docs gen unable to handle back quote "`"

Found on page https://www.itwinjs.org/reference/core-backend/imodels/imodeldb/attachdb - "undefined"

ben-polinsky:
PR is up #8894

PR #8894, opened 2026-01-09
fix iModelDb documentation include examples

ben-polinsky:
When using `[[include]]` syntax to embed snippets of example code in code comments, we should prefer the `@example` tag. 

In at least one case (`@param`), the final tsdoc tag before an `[[include]]` slurps up the example code. This can result in [poorly rendered examples](https://www.itwinjs.org/reference/core-backend/imodels/imodeldb/attachdb/).

I've fixed two examples found thanks to https://github.com/iTwin/itwinjs-core/issues/8893 h/t @khanaffan, and updated our guidance to always use `@example` tags.

### DIFF
--- a/common/changes/@itwin/core-backend/master_2026-01-09-17-02.json
+++ b/common/changes/@itwin/core-backend/master_2026-01-09-17-02.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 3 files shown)


ben-polinsky:
@mergifyio backport release/5.5.x

Issue #8897, opened 2026-01-13
core-frontend 5.5.0 expects peer of @loaders.gl/core


gedkek:
after core-frontend update to 5.5.0, getting peer dep error for @loaders.gl/draco requiring @loaders.gl/core@^4.3.0
i see dependency was removed from core-frontend as unused, and new requirement not mentioned as required peer dep anywhere from what i see.

Is this change intended?

markschlosseratbentley:
PR for reference: https://github.com/iTwin/itwinjs-core/pull/8849
@tcobbs-bentley @aruniverse 
AFAIK testing did not reveal this, so I am curious of both your thoughts.

tcobbs-bentley:
I think I need to add `@loaders.gl/core` as a peer dependency.

tcobbs-bentley:
PR created: https://github.com/iTwin/itwinjs-core/pull/8900

I will backport to 5.5.x once it merges.

tcobbs-bentley:
@gedkek The change was intentional, but wasn't expected to cause problems for people.

Do you have automatic peer dependency installation disabled? Because `@loaders.gl/core` is a peer dependency of `@loaders.gl/draco`, which we still depend on. In our test apps, everything worked fine due to that peer dependency, but we don't have automatic peer dependency installation disabled. We are considering adding `@loaders.gl/core` back as a direct dependency of `@itwin/core-frontend`, even though we do not directly depend on it.

gedkek:
yea we have it disabled to ensure same peer versions across many packages and check compatibility when updating anything

tcobbs-bentley:
This will be fixed in the next iTwin release (probably 5.5.1).

hl662:
iTwin.js 5.5.1 packages have been released, thanks @tcobbs-bentley for moving quickly

PR #8900, opened 2026-01-13
Add loaders.gl/core back as a dependency of frontend

tcobbs-bentley:
Fix for https://github.com/iTwin/itwinjs-core/issues/8897

I'm pretty sure that problems only occur if automatic peer dependency installation is disabled. (It is enabled by default.)

### DIFF
--- a/core/frontend/package.json
+++ b/core/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-frontend",
-  "version": "5.6.0-dev.7",
+  "version": "5.6.0-dev.8",
   "description": "iTwin.js frontend components",
   "main": "lib/cjs/core-frontend.js",
   "module": "lib/esm/core-frontend.js",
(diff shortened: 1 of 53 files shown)


tcobbs-bentley:
@Mergifyio backport release/5.5.x

core/frontend/package.json
```diff
@@ -97,8 +97,9 @@
   "dependencies": {
     "@itwin/core-i18n": "workspace:*",
     "@itwin/webgl-compatibility": "workspace:*",
+    "@loaders.gl/core": "^4.3.4",
```
tcobbs-bentley, line 100: This seems to only be needed if users have disabled automatic peer dependency installation in their npm-compatible client. I'm not sure if we want to put it back.

tcobbs-bentley: It looks like we do want to put it back.

PR #8909, opened 2026-01-19
Add support for Electron 40

GytisCepk:
[Electron 40 breaking changes](https://github.com/electron/electron/blob/main/docs/breaking-changes.md#planned-breaking-api-changes-400)


GytisCepk:
> Man it feels like they are releasing new majors of electron too frequently

_"Electron frequently releases major versions alongside every other Chromium release."_
(source: [Electron Releases doc](https://www.electronjs.org/docs/latest/tutorial/electron-timelines))

I assume, they want Electron to have latest Chromium version. Does it require them do a major version bump every time ? I don't know, I'm not familiar with Chromium's versioning and support policy.

PR #8910, opened 2026-01-19
Accept Rendering priority for TextAnnotation elements

SheetalNair:
- Creation of TextAnnotation geometry now accepts RenderPriority.
- RenderPriority is used to assign geomParams.elmPriority for different parts of the TextAnnotation.
- geomParams.elmPriority is used to decide display priority of elements on views.

https://github.com/iTwin/drawing-production/issues/2678

### DIFF
--- a/docs/changehistory/NextVersion.md
+++ b/docs/changehistory/NextVersion.md
@@ -11,6 +11,7 @@ publish: false
   - [vacuum API](#vacuum)
   - [analyze API](#analyze)
   - [optimize API](#optimize)
+  - [TextAnnotation render priorities](#textannotation-render-priorities)
 
 ## @itwin/presentation-common
 
@@ -72,3 +73,26 @@ briefcaseDb.saveChanges();
 briefcaseDb.close();
 ```
 
+### TextAnnotation render priorities
+
+TextAnnotation elements now support custom render priorities for better control over z-ordering in 2D views. The new `renderPriority` option in [appendTextAnnotationGeometry]($backend) allows you to specify different priorities for annotation labels versus other annotation geometry (frame, leaders, etc.).
+
+```typescript
+import { appendTextAnnotationGeometry } from "@itwin/core-backend";
+
+// Set different priorities for text labels and annotation geometry
+appendTextAnnotationGeometry({
+  annotationProps,
+  layout,
+  textStyleResolver,
+  scaleFactor,
+  builder,
+  categoryId,
+  renderPriority: {
+    annotationLabels: 100,  // Priority for text block and fill
+    annotation: 50          // Priority for frame, leaders, and other geometry
+  }
+});
+```
+
+The render priority values are added to [SubCategoryAppearance.priority]($common) to determine the final display priority. This allows text annotations to render correctly relative to other 2D graphics. Note that render priorities have no effect in 3D views.


core/backend/src/annotations/TextAnnotationGeometry.ts
```diff
@@ -51,6 +63,8 @@ export function appendTextAnnotationGeometry(props: AppendTextAnnotationGeometry
 
   // Construct the TextBlockGeometry
   const params = new GeometryParams(props.categoryId, props.subCategoryId);
+  // Set the element priority for the text labels if provided. TextBlock with its fill has a different priority than the rest of the annotation.
+  props.renderPriority?.annotationLabels && (params.elmPriority = props.renderPriority.annotationLabels);
```
pmconne: Here and below, you ignore priority of zero because it's falsy. But zero is a valid priority value.

pmconne:
```suggestion
  if (props.renderPriority?.annotationLabels !== undefined) {
    params.elmPriority = props.renderPriority.annotationLabels;
  }
```

Arguably a nit, but conditional branches should be obvious at a glance.

SheetalNair: fixed

SheetalNair: done

SheetalNair:
> Documentation needs to make crystal clear that these priorities have zero effect in 3d models.

Updated docs

core/backend/src/annotations/TextAnnotationGeometry.ts
```diff
@@ -59,6 +78,11 @@ export function appendTextAnnotationGeometry(props: AppendTextAnnotationGeometry
     result = result && appendFrameToBuilder(props.builder, props.textStyleResolver.blockSettings.frame, props.layout.range, transform.clone(), params);
   }
 
+  // Set the element priority for the entire annotation element except textBlock and its fill.
+  if (props.renderPriority?.annotation !== undefined) {
+    params.elmPriority = props.renderPriority.annotation;
+    result && props.builder.appendGeometryParamsChange(params);
```
pmconne:
```suggestion
    result = result && props.builder.appendGeometryParamsChange(params);
```

I think you intended to assign to `result` here?
Relatedly, why does this function not just `return false` immediately after any given step fails, rather than accumulating multiple potential failures into this `result` variable?

claudiaareneee: Yes, that's what happens. That way it doesn't catastrophically fail if one of the pieces of geometry is failed to be created. The returned value denotes that at least one piece of geometry failed to be generated.

I was following what happens with other append methods. Most of them are simple enough they just fail out, but appendTextBlock compounds the result.

See https://github.com/iTwin/itwinjs-core/blob/e0881cad2a8590df56efc7bd958ebbdc8dcab127/core/common/src/geometry/ElementGeometry.ts#L433  

@SheetalNair, I agree @pmconne's code suggestion. It should be `result = result && ...`

pmconne: What would anyone want to do with a half-broken geometry stream, other than discard it? Especially if the caller has no way of knowing which pieces were successfully created vs which failed?

SheetalNair: Thanks @pmconne for pushing the change for early returns.

core/backend/src/annotations/TextAnnotationGeometry.ts
```diff
@@ -14,6 +14,19 @@ import { produceTextBlockGeometry } from "./TextBlockGeometry";
 import { appendFrameToBuilder, computeIntervalPoints } from "./FrameGeometry";
 import { appendLeadersToBuilder } from "./LeaderGeometry";
 
+
+/**
+ * Render priorities for different parts of a text annotation.
+ * These values are provided to geometryParams.elmPriority.
```
pmconne:
```suggestion
 * These values are used to initialize [GeometryParams.elmPriority]($common).
```
Documentation link.

pmconne (approved):

pmconne:
Please add a note to NextVersion.md.

SheetalNair:
> Please add a note to NextVersion.md.

done.

Issue #8904, opened 2026-01-15
getAllBaseClasses function did not traverse upto the base class that is present in BIS hierarchy


naveedkhan8067:
After the recent changes: https://github.com/iTwin/itwinjs-core/pull/8837, the getAllBaseClasses in a case did not traverse upto the base class that is present in BIS hierarchy that is causing bis rule 600 to fail which verifies that an entity class must derive from BIS hierarchy.

[BuildingSpatial.01.00.01](https://github.com/iTwin/bis-schemas/blob/master/Domains/2-DisciplinePhysical/Building/BuildingSpatial/Released/BuildingSpatial.01.00.01.ecschema.xml) schema can be used as an example where "Building" is an ECEntityClass.

```
      ICompositeVolume
    /
Building 
     \
       Facility               ISpatialOrganizer
               \                 /
                SpatialStructureElement 
                                   \
                                    CompositeElement
                                                   \
                                                     SpatialLocationElement
```

In this case it is not reached to "SpatialLocationElement" which is coming from BisCore and is on top in the hierarchy. For ease of understanding, tree is shown like ablove.

naveedkhan8067:
https://github.com/iTwin/bis-schema-validation/pull/69

anmolshres98:
I am seeing the same issue on our app

I have attempted to create a minimal~ish repo to help reproduce and triangulate the issue here: [anmolshres98/transformer-issue-repro](https://github.com/anmolshres98/transformer-issue-repro?rgh-link-date=2026-01-16T16%3A06%3A01.000Z). You will need to fill in the .env values and you can observe it only fails for a rvt synced iModel but not for a dgn synced iModel. Additionally, the error is thrown during a call to transformer.processModel() call here: [anmolshres98/transformer-issue-repro@7956eca/src/index.ts#L107](https://github.com/anmolshres98/transformer-issue-repro/blob/7956eca5632caad6ceaaba8372173151dd70bbd0/src/index.ts#L107)

rschili:
This is probably a regression introduced in this PR: https://github.com/iTwin/itwinjs-core/pull/8838

I'm considering reverting it

rschili:
Sorry I linked the wrong PR, this is the one we suspect: https://github.com/iTwin/itwinjs-core/pull/8837

grigasp:
[...]

PR #8912, opened 2026-01-20
Fixed the issue of getAllBaseClasses() function not returning all the base classes

soham-bentley:
The issue was while loading the class using the schemaItemKey we were just looking for the class in the current schema and its referenced schemas
But we might have class definition later up the chain like in  `Current Schema -> Referenced Schema -> Referenced Schema`(Containing class definition)

Fixed it with this PR
Fixes https://github.com/iTwin/itwinjs-core/issues/8904

This issue was introduced in PR https://github.com/iTwin/itwinjs-core/pull/8837

### DIFF
--- a/common/changes/@itwin/core-backend/2026-01-22-11-45.json
+++ b/common/changes/@itwin/core-backend/2026-01-22-11-45.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -23,8 +23,8 @@ importers:
         specifier: ^12.28.0
         version: 12.28.0
       '@bentley/imodeljs-native':
-        specifier: 5.6.7
-        version: 5.6.7
+        specifier: 5.6.8
+        version: 5.6.8
       '@itwin/object-storage-azure':
         specifier: ^3.0.4
         version: 3.0.4
@@ -4750,8 +4750,8 @@ packages:
   '@bentley/icons-generic@1.0.34':
     resolution: {integrity: sha512-IIs1wDcY2oZ8tJ3EZRw0U51M+0ZL3MvwoDYYmhUXaa9/UZqpFoOyLBGaxjirQteWXqTIMm3mFvmC+Nbn1ok4Iw==}
 
-  '@bentley/imodeljs-native@5.6.7':
-    resolution: {integrity: sha512-MjOFLnFRZrELcbA8D5phj9bVRgsDtaK19t4khY6Mhoi9on1o/Z8M6n7clabKXNb5WRSTBpSQAWZUW6dbl+R51A==}
+  '@bentley/imodeljs-native@5.6.8':
+    resolution: {integrity: sha512-mEcFiMqObI4QztrCWs3ElMEjf7Cx5nxKg+Dl6ERM/vEEcs4w6Foa5ZTAjOG1nLSTTDj38aJBB27eaWuIf3zr1A==}
 
   '@bentley/linear-referencing-schema@2.0.3':
     resolution: {integrity: sha512-2pFIEN4BS7alIDhGous6N2icKAS8ZhVBfoWB8WvSFaYnneGv5YwbbXl46qATDdPO5jUFezkW6uVxdpp1eOgUHQ==}
@@ -11351,7 +11351,7 @@ snapshots:
 
   '@bentley/icons-generic@1.0.34': {}
 
-  '@bentley/imodeljs-native@5.6.7': {}
+  '@bentley/imodeljs-native@5.6.8': {}
 
   '@bentley/linear-referencing-schema@2.0.3': {}
 
--- a/core/backend/package.json
+++ b/core/backend/package.json
@@ -113,7 +113,7 @@
     "webpack": "^5.97.1"
   },
   "dependencies": {
-    "@bentley/imodeljs-native": "5.6.7",
+    "@bentley/imodeljs-native": "5.6.8",
     "@itwin/object-storage-azure": "^3.0.4",
     "@azure/storage-blob": "^12.28.0",
     "form-data": "^4.0.4",
--- a/test-apps/display-test-app/android/imodeljs-test-app/app/build.gradle
+++ b/test-apps/display-test-app/android/imodeljs-test-app/app/build.gradle
@@ -42,7 +42,7 @@ dependencies {
     implementation 'com.google.android.material:material:1.7.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
     implementation 'androidx.navigation:navigation-ui:2.5.3'
-    implementation 'com.github.itwin:mobile-native-android:5.6.7'
+    implementation 'com.github.itwin:mobile-native-android:5.6.8'
     implementation 'androidx.webkit:webkit:1.5.0'
 }
 
--- a/test-apps/display-test-app/ios/imodeljs-test-app/imodeljs-test-app.xcodeproj/project.pbxproj
+++ b/test-apps/display-test-app/ios/imodeljs-test-app/imodeljs-test-app.xcodeproj/project.pbxproj
@@ -455,7 +455,7 @@
 			repositoryURL = "https://github.com/iTwin/mobile-native-ios";
 			requirement = {
 				kind = exactVersion;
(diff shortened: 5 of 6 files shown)


pmconne (approved):

rschili (approved):

anmolshres98:
I tested this fix on my minimal reproduction test repo by way of [pnpm patch](https://pnpm.io/cli/patch) and this fix seems to have solved the issue for that test scenario: https://github.com/anmolshres98/transformer-issue-repro/commit/bc94d1291e5b07c1d6140b6cd0ecaedd2a11c26e

johnnyd710:
I was also able to test and confirm this fixes our issue https://github.com/iTwin/itwinjs-backlog/issues/1832

rschili:
I discussed the new changes with Soham and approve of these as well. But they are different from what you guys confirmed to be a working fix. Maybe you can confirm again once he pushes in 2-3 minutes. However, pretty sure we will be fine anyways.

johnnyd710:
> I discussed the new changes with Soham and approve of these as well. But they are different from what you guys confirmed to be a working fix. Maybe you can confirm again once he pushes in 2-3 minutes. However, pretty sure we will be fine anyways.

Can confirm the latest changes as of my comment also fix the issue I was having in https://github.com/iTwin/itwinjs-backlog/issues/1832

soham-bentley:
I just pushed an implementation update
 instead of searching for base classes through the context cache I will search it through the reference tree of the schema all the way to the top...that will remove the redundant searches and make it performant
 
 Although I am 95% sure that would fix the issue...still I would request  @anmolshres98  and @johnnyd710  to quickly check once if it fixes your issues or not
 
 Sorry for the inconvenience

DanRod1999:
These seemed to pass on my side

johnnyd710:
Latest impl also passes for me

anmolshres98:
latest changes look good on my test repo: https://github.com/anmolshres98/transformer-issue-repro/compare/main...as/possible-patch-fix

soham-bentley:
> I just pushed an implementation update instead of searching for base classes through the context cache I will search it through the reference tree of the schema all the way to the top...that will remove the redundant searches and make it performant
> 
> Although I am 95% sure that would fix the issue...still I would request @anmolshres98 and @johnnyd710 to quickly check once if it fixes your issues or not
> 
> Sorry for the inconvenience

I just added some tests in `Class.test.ts` under the describe block named `ECClassHierarchy tests with manual schema and class creation`
There in I just created two separate schemas under the same context. Created one entity classes in each of the two schemas and set one class as the base of another. But these two schemas are not referenced in each other. I was able to do so without any validation.
With this test, it shows that we can have base classes in schemas which will not be in the reference tree at all. So that contradicts my logic which I had pushed last. 
So the current implementation (which I just pushed) in getAllBaseClasses() is we will first try to search and get the schema of the base class (using the schema item key) from the reference tree recursively traversing it all the way to the top. If we can't find the schema of the base class in the entire reference tree then we will try to find the schema in the context using the locators of the context

This gives us both, a performance improvement in case the schema of the base class is in our reference tree, and also the stability in case the schema isn't in our recursive reference tree
Sounds good @rschili and @naveedkhan8067 ?

As for @anmolshres98 and @johnnyd710 and @DanRod1999  I think this implementation will also work fine for you given that the previous ones worked just fine for you
Anyways we will be able to know as we merge it

soham-bentley:
Although that raises a concern shouldn't there be validations like these that schema of base class should be referenced directly?
I am not really aware of the expected behavior.

core/ecschema-metadata/src/Metadata/Class.ts
```diff
@@ -586,20 +594,72 @@ export abstract class ECClass extends SchemaItem implements CustomAttributeConta
...
-    for(const baseClassKey of this.schema.context.classHierarchy.getBaseClassKeys(this.key)) {
-      const baseClass = this.schema.lookupItemSync(baseClassKey, ECClass);
+    for (const baseClassKey of this.schema.context.classHierarchy.getBaseClassKeys(this.key)) {
+      let baseClass = this.getClassFromReferencesRecursivelySync(baseClassKey); // First search in schema ref tree all the way to the top and then if not found then in schema context
+      if (baseClass) {
+        yield baseClass;
+        continue;
+      }
+      Logger.logWarning(loggingCategory, `Base class ${baseClassKey.name} not found in entire schema reference tree, looking in schema context.`);
+      baseClass = this.schema.context.getSchemaItemSync(baseClassKey, ECClass);
```
rschili: This is not needed for anything, no?
If we can hit this in a test I'm fine leaving it in, but it does not seem necessary

soham-bentley: We are hitting this part of the code in a test

soham-bentley: I just kept it so that in future it will help us in debugging I guess, can remove if you want, although it seems a logInfo is more appropriate than a warning

rschili: Is it because the test does not call context.addSchema? In that case the test should do that. We don't have to make the code support manual broken test setups.

Going to the context here should never be necessary now that we recurse into references. If we do it, that makes the code unpredictable because you'll never know where the information came from.

soham-bentley: All the test under `ECClassHierarchy tests with manual schema and class creation` in `Class.test.ts` are supposed to hit this part of the code because in those tests the schema containing the base class is not referenced from the main schema

rschili: Which ones? This one?

```ts
    it("class with base class in reference schema", async () => {
      const schemaJson = {
        $schema: ECSchemaNamespaceUris.SCHEMAURL3_2_JSON,
        name: "TestSchema",
        version: "1.2.3",
        alias: "ts",
        references: [
          {
            name: "RefSchema",
            version: "1.0.5",
          },
        ],
        items: {
          testClass: {
            schemaItemType: "EntityClass",
            baseClass: "RefSchema.BaseClassInRef",
          },
        },
      };

      const refSchema = new Schema(context, "RefSchema", "ref", 1, 0, 5);
      const refBaseClass = await (refSchema as MutableSchema).createEntityClass("BaseClassInRef");
      await context.addSchema(refSchema);

      schema = await Schema.fromJson(schemaJson, context);
```

This schema does reference RefSchema, so the class lookup does not have to go through the context. If it does, that's wrong. I do not see any test in that file that should find its baseclass through the context. Which line are you referring to?

rschili: I pulled and ran the tests. The only test that I see which hits this is `class having base class in a schema which is not referenced but know to schema context locators` and that is a test you are creating in this same PR.

Like we discussed, the schema of the base class MUST be directly or indirectly referenced, so this test is wrong. You need to add the RefSchema into the list of referenced schemas, then this code here becomes unnecessary

soham-bentley: Yes that's what I mentioned in comments https://github.com/iTwin/itwinjs-core/pull/8912#issuecomment-3776317679 and https://github.com/iTwin/itwinjs-core/pull/8912#issuecomment-3776320058
Are we absolutely 100% sure that the schema of the base class will always be in the reference tree of the current schema directly or indirectly? If that is the case then shouldn't we have validations which will not allow us to set base class if the schema of the base class is not in the reference tree...Currently there are no validations of that sort. That is what I wanted to document in the test case. 
What I did in the test case is I added a base class to a class but I did not reference the schema of the base class and I was allowed to do that. Any user of the code can do the same and would expect that when they call `getAllBaseClasses()`, they should get all the base classes, even if the schema is not in the reference tree, right? 

When we did not have this `ECClassHierarchy` class and in `getAllBaseClasses` method we used to recursively travel through the `_baseClass` member of `ECClass`, we used to get all the base classes of a particular class irrespective of whether the schema of the base class is referenced or not. Should we change the behavior of the method just because the way for getting base classes changed?

rschili: Yes, we are 100% sure that nobody will do this. It's consistent. ecschema-metadata is not an editing API, people are not supposed to use it to construct schemas. That's just us in our own internal tests. Because of that, it does not validate, and you can shoot yourself in the foot. The validations are in the deserializers.

It's different from our C++ implementation in that respect

soham-bentley: Cool removing it then

soham-bentley: Removed searching for schema of base class using schema context locators, we will just search for schema of base class in the reference tree recursively all the way to the top

aruniverse:
@mergifyio backport release/5.5.x

PR #8914, opened 2026-01-20
Fix not being able to run tests on Electron in Ubuntu 24.04 pipelines

GytisCepk:
Due to [changes in Ubuntu 24.04](https://discourse.ubuntu.com/t/ubuntu-24-04-lts-noble-numbat-release-notes/39890#p-99950-unprivileged-user-namespace-restrictions-15), sandboxed Electron does not work.

Alternative solution would be to make sure that correct profile is assign to necessary files, but it may complicate our pipelines even more.

It would be great, if someone more familiar with Linux permissions model, would take a look at this.

### DIFF
--- a/.github/instructions/pr-review.instructions.md
+++ b/.github/instructions/pr-review.instructions.md
@@ -1,0 +1,35 @@
+---
+applyTo: "**/*"
+---
+
+# iTwin.js pull request review heuristics
+
+These heuristics **supplement** — not replace — GitHub Copilot's default code review behavior. Continue applying standard review practices (correctness, security, performance, readability). The guidance below adds iTwin.js-specific concerns that are easy to miss without domain context.
+
+These heuristics are primarily intended to improve GitHub Copilot pull request reviews, but GitHub's current repository instruction model also lets them influence other Copilot interactions for matching files.
+
+In addition to your standard review checks, pay special attention to issues that are likely to cause runtime regressions, integration churn, consumer confusion, or repeated reviewer rework. Skip pure style nits unless they hide a real issue or contradict an established repo pattern.
+
+## Consumer impact and compatibility
+
+- For dependency, peer dependency, or version bumps, question whether each bump is necessary, whether the peer range can stay broader, and whether downstream apps need follow-up changes.
+- If a PR claims a fallback keeps the change safe, verify that the fallback is real in code or tests. Ask for the exact helper, call path, or test when the proof is missing.
+- For parser, schema, persistence, or serialization changes, check whether old data still works and whether the behavior survives roundtrips.
+
+## APIs, invariants, and reuse
+
+- For public or beta API changes, verify release tags, `rush extract-api` output, and required `rush change` files.
+- Only ask for `docs/changehistory/NextVersion.md` and migration guidance when the PR introduces a breaking change or a significant new user-visible feature.
+- Prefer comments about weak invariants over local style nits: push undefined or loading handling upward, remove redundant state, require props through types, and reuse shared helpers instead of duplicating logic.
+- Call out duplicated conditionals, repeated heuristics, or ad hoc string matching when a shared helper or stronger type would make the behavior safer.
+
+## UI, docs, and examples
+
+- For UI or UX changes, check for a GIF or screenshot in the PR description and flag confusing CTA placement, layout shift, copy, localization, accessibility, or cases that should get UX review.
+- Check that README files, example snippets, comments, screenshots, and localized strings still match the shipped behavior. Prefer compiled or extracted snippets over hand-written README code.
+- For changelog text, prefer concrete user-facing behavior descriptions over vague labels like "breaking" unless the PR truly changes a supported contract.
+
+## Review style
+
+- Prefer high-signal comments that explain downstream impact, missing validation, or future maintenance risk.
+- Do not spend review budget on formatting-only comments unless they create ambiguity or churn that reviewers in this repo have repeatedly called out.
--- a/.github/skills/merge-conflict-resolving/SKILL.md
+++ b/.github/skills/merge-conflict-resolving/SKILL.md
@@ -138,9 +138,44 @@ rush change --bulk --message "" --bump-type none -b origin/release/X.X.x
 
 ## Documentation Conflicts (NextVersion.md)
 
-**Merge both versions intelligently.** Preserve unique content from both branches, organize by category, and maintain the table of contents.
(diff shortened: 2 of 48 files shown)


aruniverse:
@markschlosseratbentley iirc youre a linux user, can you take a pass?

markschlosseratbentley:
> @markschlosseratbentley iirc youre a linux user, can you take a pass?

Unfortunately, I no longer have a Linux partition. IIRC, @pmconne does.

pmconne:
@aruniverse how do you want it tested - just try to run the core-full-stack integration tests? Or presumably I need to run the actual pipeline?

I presume this is the same or similar issue that I experience whenever trying to run display-test-app?
```
j2 display-test-app: pmc/anchor-line-patterns 09:31:26 ~/c/j2/s/itwinjs/test-apps/display-test-app 0> npm run start

> display-test-app@1.8.0 start
> cross-env NODE_ENV=development run-p start:webserver start:electron


> display-test-app@1.8.0 start:webserver
> cross-env NODE_ENV=development vite


> display-test-app@1.8.0 start:electron
> electron ./lib/backend/DtaElectronMain.js --inspect --remote-debugging-port=9223

[26868:0120/093133.752231:FATAL:sandbox/linux/suid/client/setuid_sandbox_host.cc:166] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /home/paul/c/j2/s/itwinjs/common/temp/node_modules/.pnpm/electron@39.0.0/node_modules/electron/dist/chrome-sandbox is owned by root and has mode 4755.
/home/paul/c/j2/s/itwinjs/common/temp/node_modules/.pnpm/electron@39.0.0/node_modules/electron/dist/electron exited with signal SIGTRAP
ERROR: "start:electron" exited with 1.
```

EDIT: yes it's the same error (can we fix the test-apps too or at least document it for other people trying to run them?)

GytisCepk:
> @aruniverse how do you want it tested - just try to run the core-full-stack integration tests? Or presumably I need to run the actual pipeline?

Pipeline is passing. The question is more whether solution is sound.

> EDIT: yes it's the same error (can we fix the test-apps too or at least document it for other people trying to run them?)

From what I gather, there is multiple ways to solve this and it's really depends on user's Linux distribution which solution is the best. For that reason, I think, it's up to the user to make sure his environment does not block running sandboxed Electron.

That being said, Ubuntu 24.04 might be common enough case to warrant a note in display-test-app's README.md

common/config/azure-pipelines/templates/integration-test-steps.yaml
```diff
@@ -34,6 +34,18 @@ steps:
       node ./common/scripts/install-run-rush webpack:test -v
     displayName: "Rush build"
 
+  - script: |
+      SANDBOX_PATH=$(find $PWD/common/temp/node_modules/.pnpm -path "*/electron@*/node_modules/electron/dist/chrome-sandbox" 2>/dev/null | head -n 1)
+      if [ -f "$SANDBOX_PATH" ]; then
+        sudo chown root:root "$SANDBOX_PATH"
+        sudo chmod 4755 "$SANDBOX_PATH"
```
tcobbs-bentley: I think running `chrome-sandbox` as root is really bad idea. I would say that if this is the recommended solution, then the people doing the recommending are not very trustworthy.

MichaelSwigerAtBentley: Can you add a comment about what this chmod enables? This is setting execute (5 - read+execute) for all users, correct?

common/config/azure-pipelines/integration-validation.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         Linux_node_20_x:
-          imageName: ubuntu-22.04 # ubuntu-latest is 24.04, and breaks tests using sandbox
+          imageName: ubuntu-latest
```
MichaelSwigerAtBentley, line 25: Are we sure we want to swap to latest and not a fixed 24.x? What if latest breaks us again in the future

common/config/azure-pipelines/templates/integration-test-steps.yaml
tcobbs-bentley: (But I guess it's possible that I'm missing something, and `chrome-sandbox` is somehow super safe to run as root. I'm dubious about that, though.)

tcobbs-bentley: If it is decided to leave this, then a comment explaining what it does and why it is acceptable is definitely required. The 4 at the beginning means that `chrome-sandbox` will execute using the user ID of the file's owner. The file was just changed to be owned by root, so what this is doing is having `chrome-sandbox` execute as root, with full root privileges. That seems to be the intended fix, but I find it to be very dangerous.

tcobbs-bentley: This may be a safer fix:

https://gist.github.com/0xdeadbad/5ef3a38c8cf6efc264d9a6f11f50e5b8

tcobbs-bentley: Note: the fix fix above is for a different program, but I think it can be adapted to work with Electron/`chrome-sandbox`.

tcobbs-bentley: Also, the following command probably fixes the problem without giving root privileges to the `chrome-sandbox` process:

```
sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
```

I believe that this puts the security back to the way it was before 24.04. That's definitely not ideal, but **may** be better than the current fix. Changing the `0` at the end to `1` restores the original settings.

GytisCepk: Added comment above the step to explain the reason for the new solution

common/config/azure-pipelines/integration-validation.yaml
GytisCepk: If it breaks, it's better we catch it earlier rather than later, since unlike this, it could be something which breaks our code in a way that affects users.

common/config/azure-pipelines/templates/integration-test-steps.yaml
GytisCepk: Likelihood that this could be exploited is close to 0, in my opinion.

Regardless, you're right, there is a safer way to do this. I updated script to instead add a new AppArmor profile which allows Electron to access user namespaces.

PR #8919, opened 2026-01-22
Splitting up, restructure, rewrite quantity documentation

hl662:
This pull request primarily improves the documentation and clarity of the `FormatProps` and related interfaces in the `@itwin/core-quantity` package, making the API easier to understand and use. It also adds new documentation site configuration for the quantity package and updates dependencies in the monorepo lockfile.

**API Documentation Improvements:**

* Enhanced documentation for `FormatProps`, `FormatCompositeProps`, and related interfaces in `core/quantity/src/Formatter/Interfaces.ts` to provide detailed descriptions, clarify requirements, and specify constraints for each property. This includes indicating when fields are required, their valid ranges or types, and the behavior they control. Subsequently, stripped out redundant, in-line code comments.

**Documentation Site Configuration and Improvements:**

* Added a new documentation site entry for "quantity" in `docs/config/docSites.json` and included it in the default doc sites list. 
image

* Restructures Quantity documentation from a single comprehensive page into organized sections (Definitions, Usage), including an intentionally simple diagram of the APIs involved in formatting: 
image

* Created more tested code examples and integrated via extraction blocks.

### DIFF
--- a/common/api/ecschema-editing.api.md
+++ b/common/api/ecschema-editing.api.md
@@ -2271,6 +2271,8 @@ export class SchemaItemMissing extends SchemaItemChange {
 export class SchemaMerger {
     constructor(editingContext: SchemaContext);
     // @alpha
+    getMergeReport(): SchemaMergeReport | undefined;
+    // @alpha
     merge(differenceResult: SchemaDifferenceResult, edits?: SchemaEdits): Promise<Schema>;
     // @alpha
     mergeSchemas(targetSchema: Schema, sourceSchema: Schema, edits?: SchemaEdits): Promise<Schema>;
--- a/common/changes/@itwin/core-frontend/presentation-deprecated-deps_2026-02-02-14-57.json
+++ b/common/changes/@itwin/core-frontend/presentation-deprecated-deps_2026-02-02-14-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/common/changes/@itwin/ecschema-editing/rohitptnkr-enhance-schema-merger-reporting_2026-02-03-10-16.json
+++ b/common/changes/@itwin/ecschema-editing/rohitptnkr-enhance-schema-merger-reporting_2026-02-03-10-16.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/ecschema-editing",
+      "comment": "Updated SchemaMerger to provide a merge report to track which operations merged and which failed the merge operation.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/ecschema-editing"
+}
--- a/common/changes/@itwin/map-layers-formats/presentation-deprecated-deps_2026-02-02-14-57.json
+++ b/common/changes/@itwin/map-layers-formats/presentation-deprecated-deps_2026-02-02-14-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/map-layers-formats",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/map-layers-formats"
+}
--- a/common/changes/@itwin/presentation-backend/presentation-deprecated-deps_2026-02-02-14-57.json
+++ b/common/changes/@itwin/presentation-backend/presentation-deprecated-deps_2026-02-02-14-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/presentation-backend"
+}
--- a/common/changes/@itwin/presentation-common/presentation-deprecated-deps_2026-02-02-14-57.json
+++ b/common/changes/@itwin/presentation-common/presentation-deprecated-deps_2026-02-02-14-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-common",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/presentation-common"
+}
--- a/common/changes/@itwin/presentation-frontend/presentation-deprecated-deps_2026-02-02-14-57.json
+++ b/common/changes/@itwin/presentation-frontend/presentation-deprecated-deps_2026-02-02-14-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/presentation-frontend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/presentation-frontend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -901,9 +901,6 @@ importers:
       source-map-loader:
         specifier: ^5.0.0
         version: 5.0.0(webpack@5.97.1)
-      typemoq:
-        specifier: ^2.1.0
-        version: 2.1.0
       typescript:
         specifier: ~5.6.2
         version: 5.6.2
@@ -2173,9 +2170,6 @@ importers:
       source-map-support:
         specifier: ^0.5.6
         version: 0.5.6
-      typemoq:
-        specifier: ^2.1.0
-        version: 2.1.0
       typescript:
         specifier: ~5.6.2
         version: 5.6.2
@@ -2879,9 +2873,6 @@ importers:
       sinon-chai:
         specifier: ^3.7.0
         version: 3.7.0(chai@4.3.10)(sinon@17.0.2)
-      typemoq:
-        specifier: ^2.1.0
-        version: 2.1.0
       typescript:
         specifier: ~5.6.2
         version: 5.6.2
@@ -3269,9 +3260,6 @@ importers:
       sinon-chai:
         specifier: ^3.7.0
         version: 3.7.0(chai@4.3.10)(sinon@17.0.2)
-      typemoq:
-        specifier: ^2.1.0
-        version: 2.1.0
       typescript:
         specifier: ~5.6.2
         version: 5.6.2
@@ -3375,15 +3363,12 @@ importers:
       sinon-chai:
         specifier: ^3.7.0
         version: 3.7.0(chai@4.3.10)(sinon@17.0.2)
-      typemoq:
-        specifier: ^2.1.0
-        version: 2.1.0
       typescript:
         specifier: ~5.6.2
         version: 5.6.2
       typescript-json-schema:
-        specifier: ^0.55.0
-        version: 0.55.0
+        specifier: ^0.67.1
+        version: 0.67.1
       yargs:
         specifier: ^17.4.0
         version: 17.4.0
@@ -3502,9 +3487,6 @@ importers:
       sinon-chai:
         specifier: ^3.7.0
         version: 3.7.0(chai@4.3.10)(sinon@17.0.2)
-      typemoq:
-        specifier: ^2.1.0
-        version: 2.1.0
       typescript:
         specifier: ~5.6.2
         version: 5.6.2
@@ -5914,8 +5896,8 @@ packages:
   '@types/multiparty@0.0.31':
     resolution: {integrity: sha512-cAFkeGKH55zJiYUjvXznQ3IdShi15VPcaDYxTm9a/lvLhD5dZgSrS+FmQvdS1/vFVIEolFGM1eZCB1avVpiN3w==}
 
-  '@types/node@16.18.126':
-    resolution: {integrity: sha512-OTcgaiwfGFBKacvfwuHzzn1KLxH/er8mluiy8/uM3sGXHaRe73RrSIj01jow9t4kJEW633Ov+cOexXeiApTyAw==}
+  '@types/node@18.19.130':
+    resolution: {integrity: sha512-GRaXQx6jGfL8sKfaIDD6OupbIHBr9jv7Jnaml9tB7l4v068PAOXqfcujMMo5PhbIs6ggR1XODELqahT2R8v0fg==}
 
   '@types/node@20.17.0':
     resolution: {integrity: sha512-a7zRo0f0eLo9K5X9Wp5cAqTUNGzuFLDG2R7C4HY2BhcMAsxgSPuRvAC1ZB6QkuUQXf0YZAgfOX2ZyrBa2n4nHQ==}
@@ -6786,10 +6768,6 @@ packages:
   chrome-trace-event@1.0.4:
     resolution: {integrity: sha512-rNjApaLzuwaOTjCiT8lSDdGN1APCiqkChLMJxJPWLunPAt5fy8xgU9/jNOchV84wfIxrA0lRQB7oCT8jrn/wrQ==}
     engines: {node: '>=6.0'}
-
-  circular-json@0.3.3:
-    resolution: {integrity: sha512-UZK3NBx2Mca+b5LsG7bY183pHWt5Y1xts4P3Pz7ENTwGVnJOUWbRb3ocjvX7hx9tq/yTAdclXm9sZ38gNuem4A==}
-    deprecated: CircularJSON is in maintenance only, flatted is its successor.
 
   cjs-module-lexer@1.4.3:
     resolution: {integrity: sha512-9z8TZaGM1pfswYeXrUpzPrkx8UnWYdhJclsiYMm6x/w5+nN+8Tf/LnAgfLGQCm59qAOxU8WwHEq2vNwF6i4j+Q==}
@@ -9360,11 +9338,6 @@ packages:
     resolution: {integrity: sha512-3Ybi1tAuwAP9s0r1UQ2J4n5Y0G05bJkpUIO0/bI9MhwmD70S5aTWbXGBwxHrelT+XM1k6dM0pk+SwNkpTRN7Pg==}
     engines: {node: ^10 || ^12 || >=14}
 
-  postinstall-build@5.0.3:
-    resolution: {integrity: sha512-vPvPe8TKgp4FLgY3+DfxCE5PIfoXBK2lyLfNCxsRbDsV6vS4oU5RG/IWxrblMn6heagbnMED3MemUQllQ2bQUg==}
-    deprecated: postinstall-build's behavior is now built into npm! You should migrate off of postinstall-build and use the new `prepare` lifecycle script with npm 5.0.0 or greater.
-    hasBin: true
-
   prelude-ls@1.2.1:
     resolution: {integrity: sha512-vkcDPrRZo1QZLbn5RLGPpg/WmIQ65qoWWhcGKf/b5eplkkarX0m9z8ppCat4mlOqUsWpyNuYgO3VRyrYHSzX5g==}
     engines: {node: '>= 0.8.0'}
@@ -10385,17 +10358,13 @@ packages:
     peerDependencies:
       typescript: 4.6.x || 4.7.x || 4.8.x || 4.9.x || 5.0.x || 5.1.x || 5.2.x || 5.3.x || 5.4.x || 5.5.x || 5.6.x
 
-  typemoq@2.1.0:
-    resolution: {integrity: sha512-DtRNLb7x8yCTv/KHlwes+NI+aGb4Vl1iPC63Hhtcvk1DpxSAZzKWQv0RQFY0jX2Uqj0SDBNl8Na4e6MV6TNDgw==}
-    engines: {node: '>=6.0.0'}
-
-  typescript-json-schema@0.55.0:
-    resolution: {integrity: sha512-BXaivYecUdiXWWNiUqXgY6A9cMWerwmhtO+lQE7tDZGs7Mf38sORDeQZugfYOZOHPZ9ulsD+w0LWjFDOQoXcwg==}
+  typescript-json-schema@0.67.1:
+    resolution: {integrity: sha512-vKTZB/RoYTIBdVP7E7vrgHMCssBuhja91wQy498QIVhvfRimaOgjc98uwAXmZ7mbLUytJmOSbF11wPz+ByQeXg==}
     hasBin: true
 
-  typescript@4.8.4:
-    resolution: {integrity: sha512-QCh+85mCy+h0IGff8r5XWzOVSbBO+KfeYrMQh7NJ58QujwcE22u+NUSmUxqF+un70P9GXKxa2HCNiTTMJknyjQ==}
-    engines: {node: '>=4.2.0'}
+  typescript@5.5.4:
+    resolution: {integrity: sha512-Mtq29sKDAEYP7aljRgtPOpTvOfbwRWlS6dPRzwjdE+C0R4brX/GUyhHSecbHMFLNBLcJIPt9nl9yG5TZ1weH+Q==}
+    engines: {node: '>=14.17'}
     hasBin: true
 
   typescript@5.6.2:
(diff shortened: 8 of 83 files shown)


docs/quantity-formatting/index.md
ben-polinsky, line 1: What do you think about naming the folder "Quantity-formatting-and-parsing.md"? Thinking about the link `https://dev.itwinjs.org/quantity/` and that quantity on its own is a bit ambigiuous.

docs/quantity/definitions/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Overview
+
+Format definitions describe _what_ formats are and how they're structured, independent of their runtime usage. These specifications can be:
+
+- Stored in EC schemas and retrievable through [KindOfQuantity](../../bis/ec/kindofquantity.md), defined as presentation units
+- Defined as runtime configuration (via [FormatSet]($ecschema-metadata)) to override EC schema-defined formats
+- Defined programmatically (via [FormatProps]($quantity))
+
+## When to Read This Section
```
tcobbs-bentley: I don't like this wording. To me, the "when" implies some chronological ordering, but it's really just trying to say, "read this if you are doing any of the following". "Why to Read This Section" would I feel be a bit better, but still not great. I think that the wording below would then also need to change. For example:

- **Understand format specifications** ...
- **Work with schemas** ...
- **Define custom formats** ...

Maybe make the header "Read This Section to...", and also make the wording changes.

docs/quantity/definitions/Units.md
```diff
@@ -0,0 +1,56 @@
+# Units
+
+Units are the foundational building blocks for quantity formatting and parsing. Understanding how units work, how they're organized, and how conversions are performed is essential for working with quantity formatting in iTwin.js.
+
+## Common Terms
+
+- **Unit** - A named unit of measure that can be located by its name or label. The definition of any unit is represented through its [UnitProps]($quantity).
+
+- **Unit Family/[Phenomenon]($ecschema-metadata)** - A physical quantity that can be measured (e.g., length, temperature, pressure). Only units in the same phenomenon can be converted between each other.
```
tcobbs-bentley: First of all, the linked `Phenomenon` class desperately needs its own documentation.

Secondly, a question: do we support compound units (like, for example, speed)? If so, it should probably be indicated if those have their own phenomena or not.

docs/quantity-formatting/index.md
```diff
@@ -0,0 +1,36 @@
+# Quantity Formatting and Parsing
+
+Quantity formatting ensures consistent, user-friendly display of measurements across your application, respecting unit systems and format preferences.
+
+iTwin.js provides quantity formatting through two packages: __@itwin/core-quantity__ for formatting and parsing operations, and __@itwin/ecschema-metadata__ for schema-based format and unit providers. Together, they convert numeric values into formatted display strings with appropriate units, and parse user input strings back into numeric values.
```
ben-polinsky: Would it be useful to add an image with caption demonstrating a use case for this? 
Maybe a diagram? Something to reinforce why users want/need this.

docs/quantity-formatting/usage/Providers.md
```diff
@@ -0,0 +1,286 @@
...
+**Characteristics:**
+
+- Loads formats from KindOfQuantity definitions in schemas
+- Filters formats by unit system (metric, imperial, usCustomary, usSurvey)
+- Throws error for invalid [EC full names](https://www.itwinjs.org/bis/ec/ec-name/#full-name)
+- Read-only format provider
+
+**Example: Simple Formatting**
+
+<details>
```
ben-polinsky, line 75: Is there a reasoning behind why some of these code blocks are in summaries and others not? There is a mix on this page, and a mix throughout. 

I do like seeing code not only because it's handy for users, but because it breaks up the documentation flow from being too wordy.

docs/quantity/definitions/index.md
hl662: that makes sense, I changed the header to 'Read this section to' and the bullet points below to use infinitive verbs.

docs/quantity-formatting/index.md
hl662: a little long, but quantity-formatting would be good enough. Users seem to put everything related to quantities under the blanket term 'quantity formatting' so ill do that

ben-polinsky: Nice

docs/quantity/usage/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+### [Parsing and Formatting](./ParsingAndFormatting.md)
+
+**For Library/Tool Developers** - Core concepts for building reusable components
+
+Deep dive into FormatterSpec and ParserSpec usage, parser behavior and error handling, and how iTwin tools and components use these specs for formatting values for display and parsing user input. Includes integration patterns with [QuantityFormatter]($frontend) and examples from measure tools.
+
+### [Unit Conversion](./UnitConversion.md)
+
+**For All Developers** - Understanding the conversion layer is not mandatory, but informative and helps developers who want to retrieve unit conversions values
```
tcobbs-bentley: I find the wording here to be awkward. The following may or may not be better:

```suggestion
**For All Developers** - Understanding the conversion layer is not mandatory, but it is informative, and it helps developers who want to retrieve unit conversions values.
```

docs/quantity-formatting/usage/Providers.md
```diff
@@ -0,0 +1,286 @@
...
+- Locate units by name or label
+- Retrieve [UnitProps]($quantity) for a given unit
+- Generate [UnitConversionSpec]($quantity) objects for converting between units
+
+#### Units Provider Concept
+
+A units provider acts as a registry and converter for units. When you need to format or parse a quantity value, the provider:
+
+1. **Locates the source unit** (e.g., meters for persistence)
+2. **Locates the target unit(s)** (e.g., feet and inches for display)
```
tcobbs-bentley: I don't understand how there can be multiple target units but only one source unit.

docs/quantity/usage/Providers.md
```diff
@@ -0,0 +1,286 @@
...
+- Fallback when schema loading fails
+
+#### SchemaUnitProvider
+
+[SchemaUnitProvider]($ecschema-metadata) loads unit definitions from EC schemas using a [SchemaContext]($ecschema-metadata). It provides access to the extensive Units schema as well as custom units defined in domain schemas.
+
+**Characteristics:**
+
+- Requires access to ECSchemas via [SchemaContext]($ecschema-metadata), commonly through iModels
+- Accesses units through `SchemaContext`
```
tcobbs-bentley: While I personally like using reverse ticks around class names and such, it seems like a conscious decision was made not to do that in these documents, so doing it here stands out as different from almost everywhere else.

docs/quantity/usage/Providers.md
```diff
@@ -0,0 +1,286 @@
...
+- Supports custom domain-specific units
+- More comprehensive than BasicUnitsProvider
+
+**When to use:**
+
+- Applications working with iModels
+- Need for domain-specific units (civil, structural, etc.)
```
tcobbs-bentley: Based on the previous and following bullet points, I think the following wording would be better:
```suggestion
- When domain-specific units (civil, structural, etc.) are needed
```

docs/quantity/usage/Providers.md
```diff
@@ -0,0 +1,286 @@
...
+- When unit definitions must match schema specifications
+
+### FormatsProvider
+
+A [FormatsProvider]($quantity) supplies format definitions for KindOfQuantities. The [FormatDefinition]($quantity) interface extends [FormatProps](../definitions/Formats.md#formatprops) to help identify formats.
```
tcobbs-bentley: Since this is the first mention of KindOfQuantities in this page, perhaps it should be a link.

docs/quantity-formatting/usage/Providers.md
```diff
@@ -0,0 +1,286 @@
...
+- Throws error for invalid [EC full names](https://www.itwinjs.org/bis/ec/ec-name/#full-name)
+- Read-only format provider
+
+**Example: Simple Formatting**
+
+<details>
+<summary>Formatting with SchemaFormatsProvider</summary>
+
+```ts
+[[include:Quantity_Formatting.Schema_Formats_Provider_Simple_Formatting]]
```
tcobbs-bentley, line 79: I can't find the sample code, but one comment says that it will default to km, while a later comment says that the value will be in m. The sample code doesn't allow me to understand why the km to m conversion happens.

docs/quantity/usage/ParsingAndFormatting.md
```diff
@@ -0,0 +1,416 @@
...
+   - **Format with Units** (units explicitly defined): If an unrecognized unit label is provided (e.g., "12 ABCDEF"), the parser falls back to the format's default unit for backward compatibility. For example, with a feet format, "12 ABCDEF" would parse as "12 feet".
+
+4. **Default Unit Behavior**: If no unit label is provided in the input (e.g., just "12"), the parser uses the default unit specified in the format. For unitless formats, if the input contains multiple unit labels, the first successfully matched unit becomes the default for subsequent unitless values in the same expression.
+
+5. **Unit Conversion**: Once units are matched, the parser applies the appropriate unit conversions to produce a value in the persistence unit specified by the `ParserSpec`.
+
+This error handling ensures that parsing errors are caught in unitless format contexts, preventing data corruption from unrecognized or mistyped unit labels, while maintaining backward compatibility for formats with explicitly defined units.
```
tcobbs-bentley: This whole paragraph seems both misplaced and redundant. Point 3 is about the error handling.

docs/quantity/usage/ParsingAndFormatting.md
```diff
@@ -0,0 +1,416 @@
...
+
+Only plus (`+`) and minus (`-`) operators are currently supported. Other operators will return a parsing error or invalid input result.
+
+#### Whitespace Handling
+
+If a format uses a spacer that conflicts with the operators above, additional restrictions apply:
+
+1. Mathematical operations only apply when the operator is in front of whitespace. For example:
+   - `-2FT 6IN + 6IN` is equal to `-2FT-6IN + 6IN`
+   - `-2FT-6IN - 6IN` is NOT equal to `-2FT-6IN- 6IN`
```
tcobbs-bentley: The first example is confusing: adding the actual final value (-2FT?) would probably help there.

The second example contradicts the descriptive text. In `-2FT-6IN- 6IN`, the third `-` **is** in front of whitespace. So according to the rule above, the two expressions should be equal. If they are not, then the descriptive text is wrong.

docs/quantity-formatting/usage/ParsingAndFormatting.md
```diff
@@ -0,0 +1,416 @@
...
+
+1. Mathematical operations only apply when the operator is in front of whitespace. For example:
+   - `-2FT 6IN + 6IN` is equal to `-2FT-6IN + 6IN`
+   - `-2FT-6IN - 6IN` is NOT equal to `-2FT-6IN- 6IN`
+
+<details>
+<summary>Whitespace limitation example</summary>
+
+```ts
+[[include:Quantity_Formatting.Math_Whitespace_Limitation]]
```
tcobbs-bentley, line 146: I know these are in this PR, but the samples have comments that end with `in meters`. Given the use of feet and inches in the code sample lines, I feel that removing the `in` would improve the readability. So, instead of `// -0.6096 in meters`, just say `// -0.6096 meters`. I understand why the `in` is there, but I still feel that dropping it would be better. Note: this shows up in multiple samples.

docs/quantity/usage/ParsingAndFormatting.md
```diff
@@ -0,0 +1,416 @@
...
+
+```ts
+[[include:Quantity_Formatting.Math_Whitespace_Limitation]]
+```
+
+</details>
+
+#### Composite Unit Handling
+
+1. For a value like `2FT 6IN-0.5`, the `-` sign will be treated as a spacer and not subtraction. However, the `0.5` value will use the default unit conversion provided to the parser, because it's not a part of the composite unit when that composite is made up of only 2 units - `FT` and `IN`.
```
tcobbs-bentley: I feel that this behavior is wrong-headed. Having said that, if this really is the desired behavior, then the text should explicitly state that when multiple values are present, they are added together by default. Unless I am misreading, this is saying that `2FT 6IN-0.5` is the same as `2FT 6IN 0.5`, which is also the same as `2FT 6IN + 0.5FT`. The automatic addition behavior should be explicitly described.

docs/quantity-formatting/usage/ParsingAndFormatting.md
```diff
@@ -0,0 +1,416 @@
...
+
+## Usage in iTwin Tools and Components
+
+This section explains how iTwin tools and components integrate FormatterSpec and ParserSpec for consistent quantity formatting and parsing.
+
+### QuantityFormatter Integration
+
+The [QuantityFormatter]($frontend) class in `@itwin/core-frontend` provides a convenient interface for formatting and parsing quantities. It manages:
+
+- Default formats for each [QuantityType]($frontend), but there's an effort to move away from using `QuantityType`. See [Migrating from QuantityType to KindOfQuantity](#migrating-from-quantitytype-to-kindofquantity)
```
tcobbs-bentley, line 172: Is there a reason that QuantityType isn't deprecated?

docs/quantity-formatting/usage/ParsingAndFormatting.md
```diff
@@ -0,0 +1,416 @@
...
+- [AecUnits](../../bis/domains/AecUnits.ecschema.md) - Common layer for AEC applications
+
+More information on schemas and their different layers can be found in [BIS Organization](../../bis/guide/intro/bis-organization.md).
+
+#### Handling Missing Schemas
+
+iModels might not include CivilUnits, DefaultToolsUnits, or AecUnits schemas. In such cases:
+
+1. Integrate your tools/components to use a `FormatsProvider`
+2. Add the missing KindOfQuantity and associated [FormatProps]($quantity) through that FormatsProvider
```
tcobbs-bentley, line 316: I don't know what this is saying. Does it mean that tools must recreate all the default KindOfQuantities, or do those come from the FormatsProvider automatically?

docs/quantity-formatting/usage/ParsingAndFormatting.md
```diff
@@ -0,0 +1,416 @@
...
+3. This works independently from schemas in iModels
+
+#### Default Support
+
+To support migration, `IModelApp` uses an internal [QuantityTypeFormatsProvider]($frontend) by default, which provides formatProps for each KindOfQuantity in the table above. We still strongly encourage developers to either implement their own `FormatsProvider` or use [SchemaFormatsProvider]($ecschema-metadata) if possible.
```
tcobbs-bentley, line 323: Wouldn't it be better to suggest that they use `QuantityTypeFormatProvider` by default if the schema doesn't contain the pertinent information? Saying they should implement their own in this case seems very wrong-headed.

docs/quantity/usage/ParsingAndFormatting.md
```diff
@@ -0,0 +1,416 @@
...
+
+```ts
+[[include:Quantity_Formatting.Metric_Scale]]
+```
+
+</details>
+
+### Imperial Scale Ratio Formatting
+
+The example below demonstrates formatting imperial architectural scales with fractional notation. The format uses `NToOne` ratio type with fractional formatting to display scales like "1/4"=1'" (quarter-inch scale) or "3/4"=1'" (three-quarter-inch scale).
```
tcobbs-bentley: The double quotes mixed with the inch (`"`) and foot (`'`) symbols makes this very hard to read. I think that using code formatting instead would be better:
`1/4"=1'`

docs/quantity/usage/ParsingAndFormatting.md
```diff
@@ -0,0 +1,416 @@
...
+
+```ts
+[[include:Quantity_Formatting.Metric_Scale_Parsing]]
+```
+
+</details>
+
+### Imperial Scale Ratio Parsing
+
+The example below demonstrates parsing imperial architectural scales with fractional notation. The parser can handle fractional values like "1/4"=1'", mixed fractions like "1 1/2"=1'", and decimal values, converting them to decimal length ratio values.
```
tcobbs-bentley: See previous comment.

Side note: to be consistent, you might want to make the same changes to the sample metric ratios.

docs/quantity-formatting/usage/Providers.md
hl662: [composite formats](https://dev.itwinjs.org/quantity/definitions/formats/#composite-formats) are made up of multiple target units, the units provider is used to perform unit lookup for those common cases

tcobbs-bentley: I personally don't feel that that qualifies to make the target unit be potentially plural. It may be a "composite unit", but it's not a "composite units". And if we **_do_** refer to these as being plural, then the same should apply to "source unit" on the line above. (e.g., it should be "source unit(s)"). The example of meters isn't composite, but source units can obviously be composite.

hl662: valid, I'll state it as just 'target unit', but keep the example to use feet and inches...

docs/quantity/usage/Providers.md
hl662: eh, many of these were moved from the years old quantity documentation in learnings/frontend. This wasn't intentional, and it doesn't take a lot of effort to convert backticks to md links, so I've done just that

docs/quantity/usage/Providers.md
hl662: sounds good, changed it

docs/quantity-formatting/usage/Providers.md
hl662: I forgot I must have used the KindOfQuantity associated to longer lengths in the past for the example code, but not updating the comment after switching to a normal length. Changed the comment to accurately state it defaults to 'm', not 'km'.

docs/quantity/usage/ParsingAndFormatting.md
hl662: removed

docs/quantity/usage/ParsingAndFormatting.md
hl662: I added the final values to both examples, and yeah, example on line 167, both would be equal to each other. I've fixed that, also added a test case in core/quantity explicitly testing these

docs/quantity-formatting/usage/ParsingAndFormatting.md
hl662: okay, i changed it up in several of the samples

docs/quantity/usage/ParsingAndFormatting.md
hl662: Makes sense, I added an explicit description of the addition behavior

docs/quantity-formatting/usage/ParsingAndFormatting.md
hl662: We're not ready to deprecate it, civil tools are still in the process of migrating to use KoQ over QuantityType, they're a major user of formatting.

docs/quantity/usage/ParsingAndFormatting.md
hl662: sure, switched to code formatting

docs/quantity/usage/ParsingAndFormatting.md
hl662: yep done

docs/quantity-formatting/usage/ParsingAndFormatting.md
hl662: I edited point 1 to state tools should use a FormatsProvider separate from `IModelApp.formatsProvider`. Also added a statement regarding fallbacks and having a formatProps available in cases of missing formats. 

Tools don't have to recreate all default KoQs, if they use their own formatsProvider they can add just the KoQ they need (e.g a tool measures distance just needs to define the KoQ for length, not area). But the better option I think is just having a default, hard coded formatProps the tool can fallback to if a formatsProvider doesn't have the format they need.

docs/quantity-formatting/usage/ParsingAndFormatting.md
hl662: QuantityTypeFormatProvider is internal, users can't rely on it.
the interface of formatsProvider doesn't guarantee a format is returned when a user calls `formatsProvider.getFormat`. This API contract is the same in previous iterations of quantity formatting - a user tries to get a formatProps associated with QuantityType, the method used can return formatProps or undefined.
In these cases, tools would always have something to fallback on - in my experience I've seen different tools in iTwin apps use `Number.toString()`, define a hard coded formatProps, return a blank string (worst option, although I believe the tool was just a test one)

docs/quantity-formatting/index.md
hl662: done

docs/quantity-formatting/usage/Providers.md
hl662: some of the code blocks came from the old documentation from frontend that did not use extracted code blocks. I've converted them to use extracted code blocks now, the examples that use TS code. I left json blocks alone (for now)

docs/quantity/definitions/Units.md
hl662: I've expanded the Unit Families and Phenomena section within this same md, adding a reference to that section in this line 9. 

The compound units you're mentioning refers to the Velocity phenomena - it does seem pertinent to use that as a good example, so I've added an example explaining Velocity, and how you can use WIMSE to browse the different phenomenon we support.

docs/quantity-formatting/usage/QuantityFormatting-old.md
```diff
@@ -0,0 +1,388 @@
+# Quantity Formatting
+
```
rschili, line 2: Should we mention in text that this has been updated?

```suggestion
A restructured updated version of this documentation is available at [Parsing and Formatting](../ParsingAndFormatting.md)
```

docs/quantity-formatting/usage/index.md
```diff
@@ -0,0 +1,44 @@
...
+This section covers how to apply format definitions at runtime in iTwin.js applications. These topics focus on the practical implementation of quantity formatting and parsing using providers, formatters, and parsers.
+
+## Overview
+
+Runtime usage involves:
+
+- **Providing formats**: Using format providers to supply format definitions
+- **Converting units**: Applying unit conversions between persistence and display units
+- **Formatting values**: Converting numeric values to formatted strings
+- **Parsing strings**: Converting user input strings back to numeric values
```
rschili: This list mismatches the structure further down in the file.

```suggestion
- **Providing formats**: Using format providers to supply format definitions
- **Formatting and parsing**: Converting numeric values to formatted strings and parsing user input back to numeric values
- **Converting units**: Applying unit conversions between persistence and display units
```

rschili (approved):

docs/quantity-formatting/index.md
hl662: I added a diagram to the page, my approach was to minimally show a common workflow, and the APIs needed to do so, without going into details 
image

ben-polinsky (approved):

core/quantity/src/Formatter/Format.ts
```diff
@@ -153,13 +153,13 @@ export class BaseFormat {
     this._type = parseFormatType(formatProps.type, this.name);
 
     if (formatProps.precision !== undefined) {
-      if (!Number.isInteger(formatProps.precision)) // mut be an integer
```
tcobbs-bentley, line 156: Is there a reason you removed so many comments from this file? I will admit that this comment does seem completely redundant to anyone reading the code, but some of them seem like they were appropriate.

core/quantity/src/Formatter/Format.ts
```diff
@@ -188,87 +188,86 @@ export class BaseFormat {
       }
     }
 
-    if (undefined !== formatProps.roundFactor) { // optional; default is 0.0
```
tcobbs-bentley, line 191: Unless the comment is wrong, it seems to convey useful information (default value).

core/quantity/src/Formatter/Interfaces.ts
```diff
@@ -88,7 +108,10 @@ export interface FormatProps {
   /** The name of the unit that represents a revolution/perigon. Required for bearing or azimuth types. */
   readonly revolutionUnit?: string;
 
+  /** Enables calculating mathematic operations during parsing, only addition and subtraction are supported. */
```
tcobbs-bentley:
```suggestion
  /** Enables calculating mathematic operations during parsing; only addition and subtraction are supported. */
```

core/quantity/src/Formatter/Format.ts
hl662: Yeah I removed them because of redundancy, and then updated the FormatProps interface with proper doccomments for all the properties - some of the properties didn't have a doccomment either, so this'll be much more valuable. And it'll be valuable for users, as they don't have to look at parsing code here, instead relying on doccomment.

There were a small few that I did leave, i think one being the composite units, just ones I deemed might warrant still have a comment

tcobbs-bentley: While I agree that we don't want people having to read the source code to understand how to use things, some of the comments you removed seem to be useful to anyone editing the source code in the future, and I don't feel that these should be removed. I commented on one of them, but there are others. I can comment on each one if you want, but I feel it would be better to just put back the ones that convey useful information to people editing the code. (Note: I'm pretty sure at least one of the deleted comments was incorrect, so deleting that was good.)

hl662: got it, I added back some of the comments, just the ones that cover optional properties and that mention what the default would be otherwise.

core/quantity/src/Formatter/Format.ts
hl662: added back thanks

PR #8921, opened 2026-01-23
Update to use service-authorization v2

tcobbs-bentley:
`@itwin/service-authorization` v1 has a peer dependency on iTwin 4.11. Apparently it works with iTwin 5.x, but this updates to use a version that has a peer dependency on iTwin versions with `^5.0.0`.

### DIFF
--- a/full-stack-tests/rpc-interface/package.json
+++ b/full-stack-tests/rpc-interface/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/rpcinterface-full-stack-tests",
-  "version": "5.6.0-dev.13",
+  "version": "5.6.0-dev.14",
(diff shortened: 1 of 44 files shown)


full-stack-tests/rpc-interface/package.json
```diff
@@ -50,7 +50,7 @@
     "@itwin/oidc-signin-tool": "^5.0.2",
     "@itwin/presentation-common": "workspace:*",
     "@itwin/presentation-frontend": "workspace:*",
-    "@itwin/service-authorization": "^1.0.0",
+    "@itwin/service-authorization": "^2.0.0",
```
tcobbs-bentley, line 53: I'm waiting for feedback from the team before allowing this PR to merge.

tcobbs-bentley: Seems good.

PR #8856, opened 2025-12-11
Implementation for BENTLEY_materials_line_style

danielzhong:
Fixes #8836

This PR adds support for the proposed `BENTLEY_materials_line_style` glTF extension. The proposed specification can be found [here](https://github.com/CesiumGS/glTF/pull/89).

This PR introduces support for the `BENTLEY_materials_line_style extension` in glTF. iTwin.js can now interpret and render line primitives with their specified `width` and `pattern` properties when loading glTF files. glTF materials can now specify a line width plus a 16‑bit dash bitfield, and we’ll map it either to one of the legacy `LinePixels` codes or dynamically append a fresh pattern to the shared line-code texture (which now auto-expands beyond the original 10 entries and reloads on demand)

To test this branch:
1) Clone itwinjs-core
2) Build test-apps/display-test-app
3) Run display-test-app
4) Open blank connection

image

5) In the keyin field in the top left, enter the keyin: `dta gltf`

image

6) Within the file open dialog, open test file [line_style.zip](https://github.com/user-attachments/files/24111488/line_style.zip)
7) 
Find the following section in the material definition:

```
"extensions": {
        "BENTLEY_materials_line_style": {
          "width": 3,
          "pattern": 61694
        }
      }
```

Then, modify the `width` and `pattern` values as you like.

iTwin default line pattern values:
| Line Style (LinePixels) | 16-bit Pattern | JSON / Decimal | Description |
|--------------------------|----------------|----------------|--------------|
| **Solid** | `0xFFFF` | `65535` | Solid line |
| **Code1** | `0x8080` | `32896` | 1 on, 7 off (dotted) |
| **Code2** | `0xF8F8` | `63608` | 5 on, 3 off |
| **Code3** | `0xFFE0` | `65504` | 11 on, 5 off |
| **Code4** | `0xFE10` | `65040` | 7 on, 4 off, 1 on, 1 off |
| **Code5** | `0xE0E0` | `57568` | 3 on, 5 off |
| **Code6** | `0xF888` | `63624` | 5 on, 3 off, 1 on, 3 off, 1 on, 3 off |
| **Code7** | `0xFF18` | `65304` | 8 on, 3 off, 2 on, 3 off |
| **HiddenLine** | `0xCCCC` | `52428` | Hidden line (2 on, 2 off) |
| **Invisible** | `0x0001` | `1` | Invisible line |

More customize example:
| Custom Pattern | Decimal (JSON) | Binary (LSB → MSB) | Description |
[...]

PR #8922, opened 2026-01-25
Enhance polyline patterns

danielzhong:
Trying to fix following issue:
https://github.com/iTwin/itwinjs-backlog/issues/1812

This PR is containing https://github.com/iTwin/itwinjs-core/pull/8856 changes for testing

Before:
image

After:
image

Root cause:
The “creepy-crawly” pattern drift is caused by screen-space pattern alignment.

Fix:
* For `glTF/glb`, we still have the ordered linestring topology at decode time, so cumulative distance can be computed and produces MicroStation-like alignment (pattern starts at linestring start).
* For `iMdl tiles`, polylines are already tesselated and the original linestring topology is not present, so cumulative distance can’t be derived decoding/renderer side. Those tiles keep using the anchor fallback until the tile generator bakes cumDist (or topology) in (backend?).

Summary:
* Compute per-vertex cumulative distance along each linestring in `PolylineTesselator` and plumb it through polyline params/buffers as `a_cumDist`.
* Use the cumulative distance in the polyline shader (scaled by `u_pixelsPerWorld`) to stabilize line pattern alignment across zoom/pan.
* Keep the existing per segment anchor logic as a fallback when cumulative distance is unavailable.

### DIFF
--- a/core/frontend/src/tile/GltfReader.ts
+++ b/core/frontend/src/tile/GltfReader.ts
@@ -1365,7 +1365,6 @@ export abstract class GltfReader {
       return undefined;
 
     const hasBakedLighting = undefined === primitive.attributes.NORMAL || undefined !== material.extensions?.KHR_materials_unlit;
-    let lineStyle = this.getMaterialLineStyle(material);
 
     let primitiveType: number = -1;
     switch (meshMode) {
@@ -1386,9 +1385,9 @@ export abstract class GltfReader {
         return undefined;
     }
 
-    if (primitiveType !== MeshPrimitiveType.Polyline && lineStyle) {
-      lineStyle = undefined;
-    }
+    // Only extract line style for line primitives or meshes with edges
+    const hasEdges = primitive.extensions?.EXT_mesh_primitive_edge_visibility || primitive.extensions?.CESIUM_primitive_outline;
+    const lineStyle = (primitiveType === MeshPrimitiveType.Polyline || hasEdges) ? this.getMaterialLineStyle(material) : undefined;
 
     const displayParams = this.createDisplayParams(material, hasBakedLighting, primitiveType === MeshPrimitiveType.Point, lineStyle);
     if (!displayParams)


core/frontend/src/internal/render/webgl/CachedGeometry.ts
```diff
@@ -1124,7 +1124,9 @@ export class PolylineBuffers implements WebGLDisposable {
   public indices: BufferHandle;
   public prevIndices: BufferHandle;
   public nextIndicesAndParams: BufferHandle;
-  private constructor(indices: BufferHandle, prevIndices: BufferHandle, nextIndicesAndParams: BufferHandle) {
+  public cumulativeDistances?: BufferHandle;
```
pmconne: All other vertex attributes are accessed via `VertexTable` to reduce duplication. The cumulative distances should be too.

pmconne:
> * For `glTF/glb`, we still have the ordered linestring topology at decode time, so cumulative distance can be computed and produces MicroStation-like alignment (pattern starts at linestring start).

Computing cumulative distance at decoding time won't work in general, because a line string may span multiple tiles and thus get clipped at the tile boundaries. This will cause discontinuities in the dash pattern at tile boundaries. You'll need to encode the cumulative distance in the tiler prior to clipping, and also compute the distance for any new vertices created by clipping.

You can still keep the code that computes cumulative distance during polyline tesselation, for polylines that are created as decorations on the front end.

> * For `iMdl tiles`, polylines are already tesselated and the original linestring topology is not present, so cumulative distance can’t be derived decoding/renderer side. Those tiles keep using the anchor fallback until the tile generator bakes cumDist (or topology) in (backend?).

The iMdl tiler should encode the cumulative distance too. You could simply enhance the tiler to add the distance to its tesselated output, or change it more significantly to work similarly to glTF (just encode the line string vertices - including cumulative distance - and tesselate them during decoding).

You should get rid of the anchor "fallback".

pmconne:
You have introduced a new vertex attribute which is required for the dash pattern to display correctly. The extension specification needs to specify this. Currently the extension is applied to a material, but the vertex attribute should be defined on the primitives that use that material. You could change the spec to put the dash pattern and width on the primitives too - maybe that's okay for a vendor extension, but it seems wrong. Alternatively, you could specify that the extension is applied to both a material (width and pattern) and to line-type primitives (cumulative distance attribute). Ask @weegeekps on your spec PR if he has any strong preference.

danielzhong:
> > * For `glTF/glb`, we still have the ordered linestring topology at decode time, so cumulative distance can be computed and produces MicroStation-like alignment (pattern starts at linestring start).
> 
> Computing cumulative distance at decoding time won't work in general, because a line string may span multiple tiles and thus get clipped at the tile boundaries. This will cause discontinuities in the dash pattern at tile boundaries. You'll need to encode the cumulative distance in the tiler prior to clipping, and also compute the distance for any new vertices created by clipping.
> 
> You can still keep the code that computes cumulative distance during polyline tesselation, for polylines that are created as decorations on the front end.
> 
> > * For `iMdl tiles`, polylines are already tesselated and the original linestring topology is not present, so cumulative distance can’t be derived decoding/renderer side. Those tiles keep using the anchor fallback until the tile generator bakes cumDist (or topology) in (backend?).
> 
> The iMdl tiler should encode the cumulative distance too. You could simply enhance the tiler to add the distance to its tesselated output, or change it more significantly to work similarly to glTF (just encode the line string vertices - including cumulative distance - and tesselate them during decoding).
> 
> You should get rid of the anchor "fallback".

Thanks! What about the `.bim` file / V1 file? How is the V1 tile encoded, through the backend?

pmconne:
> Thanks! What about the `.bim` file / V1 file? How is the V1 tile encoded, through the backend?

V1 tiles are generated on demand by the iTwin.js backend, using most of the same code as V4 and OGC tiles.

markschlosseratbentley:
@danielzhong can you manually resolve the merge conflicts when you next have some time?

danielzhong:
> @danielzhong can you manually resolve the merge conflicts when you next have some time?

The changes in this PR have already merged into the main line style PR: [BENTLEY_materials_line_style](https://github.com/iTwin/itwinjs-core/pull/8856#top). All merge conflicts have been resolved in that PR.

This PR remains open primarily for reviewers to examine the detailed updates related to the new cumulative distance features and its background/issue.

Update: I just closed this PR

markschlosseratbentley:
> > @danielzhong can you manually resolve the merge conflicts when you next have some time?
> 
> The changes in this PR have already merged into the main line style PR: [BENTLEY_materials_line_style](https://github.com/iTwin/itwinjs-core/pull/8856#top). All merge conflicts have been resolved in that PR.
> 
> This PR remains open primarily for reviewers to examine the detailed updates related to the new cumulative distance features and its background/issue.

If that is the case, I would recommend at least moving it back to "Draft."

PR #8927, opened 2026-01-27
[WIP] Fix broken documentation links

tcobbs-bentley:
The `[[<something>]]` style of documentation links don't work correctly with interfaces and namespaces. This updates the broken links that were identified elsewhere to instead use `{@link <something>}`.


tcobbs-bentley:
@ben-polinsky Due to problems in our tooling, I don't think this PR can progress. The `extract-api` process apparently doesn't pay any attention to the current scope when trying to parse `@link` references, so a whole bunch of my fixes cause errors in `extract-api`. While these links _could_ be updated to include the global scope, that's not the right thing to do. So, for now this PR is on hold.

PR #8930, opened 2026-01-28
node support docs language change + use node_20.x for regression tests

anmolshres98:
Instead of pinning node major versions support to a single minor version or higher, changing language in docs to signal only the latest minor release of any major node version is supported.

Fixes, [long running regression tests failures](https://dev.azure.com/bentleycs/iModelTechnologies/_build?definitionId=4320&_a=summary) on Mac_Node_20 and Linux_Node_20

Passing regression test with this change: https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=4742037&view=results

### DIFF
--- a/docs/learning/SupportedPlatforms.md
+++ b/docs/learning/SupportedPlatforms.md
@@ -15,16 +15,16 @@ They should run on most Windows, Linux and MacOS versions with [Tier 1 support](
 
 The following Node.js versions are officially supported by the iTwin.js backend code.
 
-| iTwin.js - Node Support | iTwin.js 3.x | iTwin.js 4.x | iTwin.js 5.x |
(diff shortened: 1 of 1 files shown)


MichaelSwigerAtBentley:
I agree with this change, but isn't this technically breaking? Is there anywhere else in the documentation we reference node version support that needs to be updated?

MichaelSwigerAtBentley:
Also, great to have the regression tests working again 🎉

ben-polinsky:
> but isn't this technically breaking

I also had that thought. But @aruniverse said "pining node minors seems tough"

aruniverse (approved):

anmolshres98:
@Mergifyio backport release/5.6.x

Issue #7659, opened 2025-02-06
Presentation: `typescript-json-schema` is in maintenance mode


grigasp:
Source: https://github.com/iTwin/itwinjs-core/pull/7590.

`typescript-json-schema`[^3] is a dependency of `@itwin/presentation-common` and is now in ["maintenance mode"](https://github.com/YousefED/typescript-json-schema/commit/43e557d9c50dc1695dfba1ff83fc50f40f9376ca).  We should consider switching to an alternative package or opening a PR to help them update.

PR #8939, opened 2026-02-02
Presentation: Cleanup deprecated deps

grigasp:
Closes https://github.com/iTwin/itwinjs-core/issues/7659.

I also went ahead and removed all uses of `typemoq`, which hasn't been updated for 8 years, and was bringing some deprecated dependencies.

Cleans up the following deprecated dependencies:
```
postinstall-build@5.0.3, 
circular-json@0.3.3,
```

### DIFF
--- a/presentation/common/src/test/RpcRequestsHandler.test.ts
+++ b/presentation/common/src/test/RpcRequestsHandler.test.ts
@@ -3,7 +3,6 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 /* eslint-disable @typescript-eslint/no-deprecated */
-/* eslint-disable @typescript-eslint/unbound-method */
 
 import { expect } from "chai";
 import * as sinon from "sinon";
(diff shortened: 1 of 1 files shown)


ben-polinsky:
Curious if you looked at using vitest/[vitest-mock-extended](https://github.com/eratio08/vitest-mock-extended)? Suppose it might save you some casts?

I think your approach is fine, though.

presentation/common/Ruleset.schema.json
JonasDov, line 1: Why are changes in this file needed?

presentation/common/src/test/RpcRequestsHandler.test.ts
```diff
@@ -295,24 +283,16 @@ describe("RpcRequestsHandler", () => {
...
-        rpcInterfaceMock
-
-          .setup(async (x) => x.getNodesDescriptor(token, rpcOptions))
-          .returns(async () => successResponse(result))
-          .verifiable();
+        rpcInterfaceStub.getNodesDescriptor = sinon.stub().returns(successResponse(result));
         expect(await handler.getNodesDescriptor(handlerOptions)).to.deep.eq(result);
-        rpcInterfaceMock.verifyAll();
+        expect(rpcInterfaceStub.getNodesDescriptor).to.be.calledOnceWithExactly(token, rpcOptions);
       });
```
JonasDov, line 318: Maybe the comment: `/* eslint-disable @typescript-eslint/unbound-method */` could be removed by changing to:
```ts
const stub = sinon.stub().returns(successResponse(result));
rpcInterfaceStub.getNodesDescriptor = stub;
expect(await handler.getNodesDescriptor(handlerOptions)).to.deep.eq(result);
expect(stub).to.be.calledOnceWithExactly(token, rpcOptions);
```
And the same thing for other places in this file

presentation/common/Ruleset.schema.json
grigasp: The new version of `typescript-json-schema` changed single-value enums to const (https://github.com/YousefED/typescript-json-schema/issues/534). It's used to generate `Ruleset.schema.json` from our TS rule declarations.

presentation/backend/src/test/RulesetEmbedder.test.ts
```diff
@@ -4,168 +4,156 @@
...
-    onEntityInsert.onBeforeInsert.resetHistory();
-    onEntityInsert.onAfterInsert.resetHistory();
-    onEntityUpdate.onBeforeUpdate.resetHistory();
-    onEntityUpdate.onAfterUpdate.resetHistory();
+  afterEach(() => {
+    sandbox.reset();
+    // onEntityInsert.onBeforeInsert.resetHistory();
+    // onEntityInsert.onAfterInsert.resetHistory();
+    // onEntityUpdate.onBeforeUpdate.resetHistory();
+    // onEntityUpdate.onAfterUpdate.resetHistory();
```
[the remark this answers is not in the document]
grigasp: removed

presentation/backend/src/test/Presentation.test.ts
```diff
@@ -57,7 +55,10 @@ describe("Presentation", () => {
 
     it("creates a manager instance", () => {
       expect(() => Presentation.getManager()).to.throw(PresentationError);
-      Presentation.initialize();
+      Presentation.initialize({
+        // @ts-expect-error this is carried over to `PresentationManagerDetail` to avoid creating an actual `IModelNative.platform.ECPresentationManager`
```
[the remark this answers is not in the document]
grigasp: updated

extensions/map-layers-formats/src/test/ArcGisFeature/ArcGisSymbologyRenderer.test.ts
```diff
@@ -67,7 +88,6 @@ describe("ArcGisSymbologyRenderer", () => {
   });
 
   it("should construct unique value renderer without metadata default symbol (", async () => {
```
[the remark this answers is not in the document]
grigasp: updated

presentation/common/src/test/RpcRequestsHandler.test.ts
grigasp: refactored to use the same approach as elsewhere - an object with stub functions + a counterpart that's cast to the target type

grigasp:
> Curious if you looked at using vitest/[vitest-mock-extended](https://github.com/eratio08/vitest-mock-extended)? Suppose it might save you some casts?
> 
> I think your approach is fine, though.

I haven't.
- Presentation and Map Layers tests are still on mocha + chai + sinon, so that package doesn't really apply.
- Core Frontend is on vitest, but it also already has sinon dependency and I only needed to change 1 test file, so I figured I'd do that change with as little collateral damage as possible, without introducing new deps.

JonasDov (approved):

PR #8943, opened 2026-02-02
Implemented B-Spline and spiral clipping

saeeedtorabi:
Implemented `announceClipIntervals` for spirals and B-Splines. This will allow us to be able to do clipping on spirals and B-Splines


ben-polinsky:
Could you provide a PR description? Please do so for all PRs.

saeeedtorabi:
Done

dassaf4 (approved):
I simplified things. Please take a look! Tests pass, but double check they work, especially if they require visual confirmation. I didn't look at them closely, but I confirmed the changes are covered.

saeeedtorabi:
> I simplified things. Please take a look! Tests pass, but double check they work, especially if they require visual confirmation. I didn't look at them closely, but I confirmed the changes are covered.

@dassaf4 Thanks for working on this. Indeed it's much cleaner now with much less duplicates. I reviewed your changes and ran the tests. All looked good. Unfortunately, the tests are visual and I'm not sure how we can automatically verify them:

image

If you have an idea to automatically verify them please let me know. Otherwise, the PR is good to go.

P.S. What if I manually run the tests and extract the fractions at the clipped points and then add them to the test?

dassaf4:
> P.S. What if I manually run the tests and extract the fractions at the clipped points and then add them to the test?

That would do the trick! This approach ensures we don't regress (change) in the future.

If you furthermore check that the current results are _correct_ (by visual inspection) then you are also _verifying_ the results as well.

Note that the order of the results may change, so don't bake result order into your explicit comparisons. (One way to do this is to explicitly sort the results in your test code, then compare to sorted known results.)

saeeedtorabi:
> That would do the trick! This approach ensures we don't regress (change) in the future.

@dassaf4 Actually I do have automatic verification in all of the new tests. Something like this:
image

So I think this can be merged.

dassaf4:
> So I think this can be merged.

This does not verify that the clip locations occur at the expected fractions along the curve. You don't need to confirm this for all tests, but it is good to have a few explicit confirmations of this type to guard against regression. Otherwise, a future version of the code could be missing some clips, or computing different clips that still pass your in/on tests.

core/geometry/src/test/clipping/ClipPlanes.test.ts
```diff
@@ -740,11 +740,24 @@ describe("CurveClips", () => {
     GeometryCoreTestIO.captureCloneGeometry(allGeometry, bspline);
     GeometryCoreTestIO.captureCloneGeometry(allGeometry, ls);
 
+    const expectedFractions0 = [0.07075513626850614, 0.7896464616645638];
+    const expectedFractions1 = [0.5023836037788738, 0.955441322252151];
+    let intervalIndex = 0;
+
     plane.announceClippedCurveIntervals(bspline,
       (fraction0: number, fraction1: number, _cp: CurvePrimitive) => {
+        ck.testNearNumber(
```
dassaf4: Use `testFraction` here and elsewhere when comparing fractions.

This employs `Geometry.smallFraction` as tolerance. A micrometer is too large when comparing fractions.

dassaf4 (approved):

core/geometry/src/test/clipping/ClipPlanes.test.ts
saeeedtorabi: done

core/geometry/src/test/clipping/ClipPlanes.test.ts
```diff
@@ -711,8 +714,127 @@ describe("CurveClips", () => {
...
+          "fraction0 of interval " + intervalIndex + " should match expected"
+        );
+        ck.testNearNumber(
+          expectedFractions1[intervalIndex], fraction1, Geometry.smallMetricDistance,
+          "fraction1 of interval " + intervalIndex + " should match expected"
+        );
+        intervalIndex++;
+        const point0 = bspline.fractionToPoint(fraction0);
+        const point1 = bspline.fractionToPoint(Geometry.interpolate(fraction0, 0.5, fraction1));
+        const point2 = bspline.fractionToPoint(fraction1)
```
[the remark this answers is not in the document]
saeeedtorabi: done

Issue #6855, opened 2024-06-18
Return Merging Report instead of Schema


StefanApfel-Bentley:
**What**
At the moment, when SchemaMerger.merge is called, the merging either succeeded and returns the merged schema or it fails by throwing an error. If the error occurs after some differences has been merged already, there is no way for the caller to figure out how much has been merged. 

**Proposal**
It would be nice to get a kind of report, to track which merge operation has been executed successfully.

PR #8944, opened 2026-02-03
Added a merge reporter to the SchemaMerger

RohitPtnkr1996:
Updated SchemaMerger with a reporter that will allow us to track the merge operations and provide stats on the operations performed, which ones failed, etc.

Fixes https://github.com/iTwin/itwinjs-core/issues/6855


rschili (approved):
Looks good. I was thinking maybe including some performance number on the report would be helpful. But it's not strictly necessary, approved anyways.

PR #8957, opened 2026-02-04
Update default engineering length quantities in QuantityFormatter

### LINKED ISSUE #5968: QuantityFormatters:  LengthEngineering should not have 4 decimal precision
**Expected behavior**
There is no reason to have that level of precision, and there is no UI exposed to customize the formatters.
In metric, the default should be 3 decimals, mm
In imperial, it should probably be 2 decimals

**Additional context**
The formatters are set and retrieved by the iModelApp. Today, Pineapple and other apps don't customize them because there is no UI. Civil packages that create analysis and decorations don't change the formatter, they use whatever is set by the app

hl662:
Closes #5968 
Although format sets are now available, fix default engineering length formats used through QuantityType API for applications and tools. Sets precision for metric engineering length to 3 decimal places, and use millimeters. For imperial, use 2 decimal places, keep using feet.

Update tests accordingly.

### DIFF
--- a/docs/changehistory/NextVersion.md
+++ b/docs/changehistory/NextVersion.md
@@ -3,10 +3,19 @@ publish: false
 ---
 # NextVersion
 
-- [Electron 40 support](#electron-40-support)
+- [NextVersion](#nextversion)
+  - [Electron 40 support](#electron-40-support)
+  - [Quantity Formatting](#quantity-formatting)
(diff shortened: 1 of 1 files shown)


ben-polinsky (approved):
Looks good. Anywhere we should note these defaults?

hl662:
> Looks good. Anywhere we should note these defaults?


Wasn't sure if I should, but yeah its for the best. Intend to have this available in the next minor version too, so added section in NextVer.md

PR #8958, opened 2026-02-04
Add integrityCheck() function

MichaelSwigerAtBentley:
imodel-native: https://github.com/iTwin/imodel-native/pull/1324

Adds a new function to IModelDb that allows the user to call [PRAGMA integrity_check](https://www.itwinjs.org/learning/ecsqlreference/pragmas/#pragma-integrity_check-experimental), and all of its specific sub queries. This exposes the ability to triage iModel state from TS. Also parses its output as Typescript would expect.

### Default Check:
   - Quick Check: Runs all integrity checks below and returns whether each check passed or failed, without detailed results.
   
   ### Specific Checks:
   - Data Columns Check: Checks if all the required columns exist in data tables. Issues are returned as a list of those tables/columns.
   - EC Profile Check: Checks if the profile table, indexes, and triggers are present. Does not check be_* tables. Issues are returned as a list of tables/indexes/triggers which were not found or have different DDL.
   - Navigation Class Ids Check: Checks if RelClassId of a Navigation property is a valid ECClassId. It does not check the value to match the relationship class.
   - Navigation Ids Check: Checks if Id of a Navigation property matches a valid row primary class.
   - Linktable Foreign Key Class Ids Check: Checks if SourceECClassId or TargetECClassId of a link table matches a valid ECClassId.
   - Linktable Foreign Key Ids Check: Checks if SourceECInstanceId or TargetECInstanceId of a link table matches a valid row in primary class.
   - Class Ids Check: Checks persisted ECClassId in all data tables and makes sure they are valid.
     Data Schema Check: Checks if all the required data tables and indexes exist for mapped classes. Issues are returned as a list of tables/columns which were not found or have different DDL.
   - Schema Load Check: Checks if all schemas can be loaded into memory.
   - Missing Child Rows Check: Checks if all child rows have a corresponding parent row.
    

Implements: https://github.com/iTwin/itwinjs-backlog/issues/1811?issue=iTwin%7Citwinjs-backlog%7C1848

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -25,8 +25,8 @@ importers:
         specifier: ^12.28.0
         version: 12.28.0
       '@bentley/imodeljs-native':
-        specifier: 5.7.3
-        version: 5.7.3
+        specifier: 5.7.5
+        version: 5.7.5
       '@itwin/object-storage-azure':
         specifier: ^3.0.4
         version: 3.0.4
@@ -4746,8 +4746,8 @@ packages:
   '@bentley/icons-generic@1.0.34':
     resolution: {integrity: sha512-IIs1wDcY2oZ8tJ3EZRw0U51M+0ZL3MvwoDYYmhUXaa9/UZqpFoOyLBGaxjirQteWXqTIMm3mFvmC+Nbn1ok4Iw==}
 
-  '@bentley/imodeljs-native@5.7.3':
-    resolution: {integrity: sha512-WqFGKfCa8p4pjZ+g42JdGcpttg368rKDEqbGGCQaVSy5FHVv/iYMLjJ3UhDBzG1YOPKsUO3VXniAM25e6wGNpg==}
+  '@bentley/imodeljs-native@5.7.5':
+    resolution: {integrity: sha512-qkjIxcxHnQ7WZJv4SWX29bKqIqzgieXQHMXRjbEB5L8eavCzhSx92dSBYCJyeznto+De1AD2jC89X4ftj04eYQ==}
 
   '@bentley/linear-referencing-schema@2.0.3':
     resolution: {integrity: sha512-2pFIEN4BS7alIDhGous6N2icKAS8ZhVBfoWB8WvSFaYnneGv5YwbbXl46qATDdPO5jUFezkW6uVxdpp1eOgUHQ==}
@@ -11315,7 +11315,7 @@ snapshots:
 
   '@bentley/icons-generic@1.0.34': {}
 
-  '@bentley/imodeljs-native@5.7.3': {}
+  '@bentley/imodeljs-native@5.7.5': {}
 
   '@bentley/linear-referencing-schema@2.0.3': {}
 


core/backend/src/IntegrityCheck.ts
```diff
@@ -0,0 +1,259 @@
+import { IModelStatus } from "@itwin/core-bentley";
+import { IModelError } from "@itwin/core-common/lib/cjs/IModelError";
+import { IModelDb } from "./IModelDb";
+
+/** Integrity check types with their display names
```
khanaffan: Keep that internal on in test. The issue with pragma integrity check is it's not lazy loading rows. If you full result set then get ready to get like millions of rows as well. Which some db do have. 

We need to improve this by instead providing virtual table or something that can step as you step through result.

core/backend/src/internal/IntegrityCheck.ts
```diff
@@ -0,0 +1,361 @@
...
+}
+
+/**
+ * Performs a quick integrity check on the given iModel.
+ * @param iModel The IModelDb instance to perform the integrity check on
+ * @returns An array of results for each check performed, including the check name, whether it passed, and the elapsed time in seconds
+ * @internal
+ */
+export async function performQuickIntegrityCheck(iModel: IModelDb): Promise<QuickIntegrityCheckResultRow[]> {
+  const integrityCheckQuery = "PRAGMA integrity_check options enable_experimental_features";
```
ben-polinsky, line 259: So this PRAGMA returns an sql statement or a check key?

ben-polinsky: Oh my mistake - sqlCommand, I get it.

core/backend/src/IntegrityCheck.ts
MichaelSwigerAtBentley: I have moved the integryCheck.ts file into internal completely.

Should I look at capping the results I return in TS to 1000?

common/changes/@itwin/core-backend/mike-fix-pragma_2026-02-06-16-41.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Add iModelDb.integrityCheck function that checks imodel database for corruption and reports results",
+      "type": "none"
```
anmolshres98, line 6: should change type be minor ?

MichaelSwigerAtBentley: Do we utilize "type" from these changes? I thought we just used the comments. Type would normally be used to decide when a major/minor/patch is needed, but we do that on a cycle instead.

MichaelSwigerAtBentley: Plus, we plan to backport this as a patch

anmolshres98: Hmm, I guess not. I was thinking of it more from a traditional change notes perspective with, major, minor, patch changes listed separately. But it seems that we only have one "updates" field where we have all the changes listed together. Resolving, thanks.

aruniverse:
@mergifyio backport release/5.6.x

PR #8963, opened 2026-02-05
Fix `Quantity.convertTo()` return type

hl662:
This pull request updates the `Quantity.convertTo()` method in `@itwin/core-quantity` to always return a valid `Quantity` object, correcting its TypeScript signature and removing unnecessary checks for `undefined` throughout the codebase. The simplifies ease of usage for performing unit conversions, and ensures consistency between the actual implementation and type definitions.

**API and Documentation Updates:**

* The TypeScript signature of `Quantity.convertTo()` was changed from returning `Quantity | undefined` to always returning `Quantity`, reflecting its actual behavior.

**Codebase Simplification:**

* All checks for `undefined` return values from `convertTo()` were removed in production code, including in `Formatter.ts` and `Parser.ts`

**Test Cleanup:**

* Test files were updated to remove unnecessary non-null assertions and checks for `undefined` when using `convertTo()`, simplifying assertions and improving readability.

### DIFF
--- a/docs/changehistory/NextVersion.md
+++ b/docs/changehistory/NextVersion.md
@@ -23,6 +23,6 @@ For applications and tools using [QuantityFormatter]($frontend) and [QuantityTyp
 
 ### Fix `Quantity.convertTo()` return type to reflect actual behavior
 
-The [Quantity.convertTo()]($quantity) method has always returned a valid `Quantity` object since its initial implementation. However, its TypeScript signature incorrectly indicated it could return `undefined` with the type `Quantity | undefined`. This has been corrected to return `Quantity`.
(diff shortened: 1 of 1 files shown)


common/api/core-quantity.api.md
```diff
@@ -539,7 +539,7 @@ export interface PotentialParseUnit {
 // @beta
 export class Quantity implements QuantityProps {
     constructor(unit?: UnitProps, magnitude?: number);
-    convertTo(toUnit: UnitProps, conversion: UnitConversionProps): Quantity | undefined;
+    convertTo(toUnit: UnitProps, conversion: UnitConversionProps): Quantity;
```
aruniverse, line 542: we are narrowing the type, is this breaking?

rschili (approved):

common/api/core-quantity.api.md
rschili: looks reasonable to me. nam says this never returned undefined before, so we're just narrowing the signature, extending it would be breaking, this should be fine. I don't see how this could break anybody. Maybe the compiler will start warning about unnecessary undefined checks for some consumers.

PR #8967, opened 2026-02-09
Add "cover" script to ecschema-rpcinterface-common package to match the existing "test" script

kring:
On my system, `rush test` fails like this (in addition to the other failures I fixed in #8965):

> --[ FAILURE: @itwin/ecschema-rpcinterface-common ]-----------[ 7.46 seconds ]--
>
>
>  4 passing (76ms)
  1 failing
>
>  1) RpcIncrementalSchemaLocater Tests
       get schema, incrementally:
     SyntaxError: "undefined" is not valid JSON
      at JSON.parse (<anonymous>)
      at D:\bentley\itwinjs-core\core\ecschema-metadata\src\IncrementalLoading\ECSqlSchemaLocater.ts:429:19
      at Array.map (<anonymous>)
      at RpcIncrementalSchemaLocater.getSchemaPartials (D:\bentley\itwinjs-core\core\ecschema-metadata\src\IncrementalLoading\ECSqlSchemaLocater.ts:428:32)
      at RpcIncrementalSchemaLocater.loadSchema (D:\bentley\itwinjs-core\core\ecschema-metadata\src\IncrementalLoading\IncrementalSchemaLocater.ts:156:28)
      at SchemaContext.getSchema (D:\bentley\itwinjs-core\core\ecschema-metadata\src\Context.ts:353:22)
      at Context.<anonymous> (src\test\RpcIncrementalSchemaLocater.test.ts:93:20)

This does not happen on CI because CI doesn't run this test. CI runs `rush cover` instead of `rush test`. But in this package, the scripts are defined like this:

```json
"scripts": {
    // ...
    "cover": "",
    // ...
    "test": "mocha"
    // ...
}
```

So the `cover` script doesn't run the tests at all. I believe this is a mistake, because it's the only package that has a test script but no cover script. So this PR adds a cover script.

### DIFF
--- a/core/ecschema-rpc/common/src/test/RpcIncrementalSchemaLocater.test.ts
+++ b/core/ecschema-rpc/common/src/test/RpcIncrementalSchemaLocater.test.ts
@@ -69,13 +69,12 @@ describe("RpcIncrementalSchemaLocater Tests", () => {
         { name: schemaKey.name, version: schemaKey.version.toString(), references: "[]", alias: "ts" },
         { name: "ECDbMeta", version: "4.0.3", references: "[]", alias: "ecdb" },
       ]))
-      .onCall(1).returns(toQueryResult(["name", "version", "alias", "references", "items"], [
+      .onCall(1).returns(toQueryResult(["item"], [
         {
-          name: schemaKey.name,
-          version: schemaKey.version.toString(),
-          alias: "ts",
-          references: "[]",
-          items: "[]",
+          item: JSON.stringify({
+            name: schemaKey.name,
+            schemaItemType: "EntityClass",
+          }),
         }
       ]))
       .returns(toQueryResult(["schema", "items"], [


kring:
No surprise: the test fails on CI, too, now that we're running it. I've looked (very briefly so far) at how to fix it, and it's not obvious to me. Any tips are appreciated, otherwise I'll keep digging.

tcobbs-bentley:
The code that is failing is expecting to have a object with an `item` field, and it expects that item field to contain a JSON string which when parsed will produce an object conforming to this interface:

```ts
interface SchemaItemStubRow {
  readonly name: string;
  readonly schemaItemType: string;
  readonly modifier?: number;
  readonly baseClasses?: Array<BaseClassStubRow>;
  readonly mixins?: Array<{
    readonly schema: string,
    readonly name: string;
    readonly schemaItemType: string;
    readonly modifier: number;
    readonly baseClasses?: Array<BaseClassStubRow>
  }>;
}
```

However, instead of that, the object looks like this:

```json
{
  "name": "TestSchema",
  "version": "01.00.00",
  "alias": "ts",
  "references": "[]",
  "items": "[]"
}
```

Unfortunately, I don't know enough about the schema code to understand what is wrong. I'll try to track down someone who can help.

hl662:
@rschili is this something Stefan was looking to fix as part of the incremental schema loading epic?

anmolshres98:
Not an expert either but looked around a bit to figure it out, for now at least. Basically, it is failing on the second step of loading schemas incrementally where this is called https://github.com/iTwin/itwinjs-core/blob/8f2c9edda2d376b161209d48242669405e6d3074/core/ecschema-metadata/src/IncrementalLoading/ECSqlSchemaLocater.ts#L374

This calls `executeQuery` which eventually calls `queryRows` which we have stubbed to return in the failing test block. In the same method here https://github.com/iTwin/itwinjs-core/blob/8f2c9edda2d376b161209d48242669405e6d3074/core/ecschema-metadata/src/IncrementalLoading/ECSqlSchemaLocater.ts#L429 is where test throws because it is expecting the query response to have `item` prop but it doesn't. The original stub is incorrect there and is returning an object that is incorrect, hence the throw. I have pushed a fix to account for that. 

The `item` needs to conform to this interface, like @tcobbs-bentley mentioned

```js
interface SchemaItemStubRow {
  readonly name: string;
  readonly schemaItemType: string;
  readonly modifier?: number;
  readonly baseClasses?: Array<BaseClassStubRow>;
  readonly mixins?: Array<{
    readonly schema: string,
    readonly name: string;
    readonly schemaItemType: string;
    readonly modifier: number;
    readonly baseClasses?: Array<BaseClassStubRow>
  }>;
}
```

kring:
Thanks everyone!

Issue #8942, opened 2026-02-02
Allow true row by row stepping over query results on a writable connection using ECSqlReader


khanaffan:
## Is your feature request related to a problem? Please describe.
Currently, when working with writable connections in iTwin.js, there is no direct way to perform true row-by-row stepping over query results that remain consistent with the state of the database as mutations (such as deletions) occur during iteration. This limitation is highlighted in the parent issue #7989, where `createQueryReader()` and `withPreparedStatement()` yield different result sets if items are deleted while iterating.

## Describe the solution you'd like
Introduce or extend support for true row-by-row (stepping) execution over query results on a writable connection. The new or improved functionality should make it possible to iterate results such that rows deleted during iteration are not surfaced by subsequent steps (matching the behavior of `withPreparedStatement()`), ensuring accurate, real-time views of the current state.

## Describe alternatives you've considered
- Using `withPreparedStatement()` currently provides the expected row-by-row behavior, but its ergonomics or API may differ from `createQueryReader()`. A unified, ergonomic, and predictable stepping capability supporting mutating operations is desired.
- Manual filtering can work, but this is inefficient and unreliable.

## Additional context
See parent issue #7989 for detailed code samples and reproduction steps. This proposal aims to improve result consistency and developer experience when mutating the iModel during data iteration on writable connections.

khanaffan:
To make this optimal we need to have strategy where row-by-row case keep the cached statement open in concurrent query and hit same connection and just step() and return result. Currently it step() until it get all result constraint by quota.. 

I think for `usePrimaryConn` we should carved our seperate code path (from imodel-native to twinjs-core) optimized for editing app backend. 

1. Get result form primary connection so it has all unsaved changes.
2. Allow row-by-row stepping on a query.

Have performance test to ensure we do not degrade in performance relative to current ECSqlReader performance and also w.r.t. ECSqlStatement.

soham-bentley:
# Initial solution
Writing an `ECSqlStatement` backed executor which will be hooked to `ECSqlReader` giving true row by row results as we step through it.
# Limitations of the solution
The performance of this executor is not that great as now for every row we need a request and response cycle which is ultimately a promise. So essentially for a 1000 ops execution cycle we would be resolving 1000 request response cycle and 1000 promises instead of just one request response cycle of `ECSqlReader`
# Improved solution
[...]

Issue #7756, opened 2025-02-21
The `ECSqlReader` API is asynchronous, but there may be cases where a synchronous version is needed.


khanaffan:
The `ECSqlReader` API is asynchronous, so the caller must also be asynchronous to use it.

The `ECSqlStatement` API, which has been deprecated in version 4.11, is synchronous, allowing the caller to be either asynchronous or synchronous.

To transition from `ECSqlStatement` to `ECSqlReader`, you need to address scenarios where the caller is synchronous and cannot be made asynchronous.

Therefore, synchronous queries with `ECSqlReader` are necessary.

This can be achieved by adding `ECSqlReader.stepSync()` or by implementing a backend-specific `IModelDb.createReaderSync()` method that includes a synchronous step function.

PR #8970, opened 2026-02-10
Adding live row by row behavior to ECSqlReader

soham-bentley:
Fixes: https://github.com/iTwin/itwinjs-core/issues/8942
Fixes: https://github.com/iTwin/itwinjs-core/issues/7756
imodel-native: https://github.com/iTwin/imodel-native/pull/1325
This pull request introduces a new true row by row behavior for ECSqlReader, improves cache management in ECDb.ts, and updates tests to use the new API. The changes primarily add support for row-by-row streaming queries, centralize cache clearing logic in ECDb.ts, and update existing tests to reflect these improvements.

**Streaming ECSQL Query API:**

* Added the `withQueryReader` method to both `ECDb` and `IModelDb`, enabling true row by row behavior with `ECSqlSyncReader` api through the new `ECSqlRowExecutor` and supporting new configuration options via `SynchronousQueryOptions`.
* Introduced new internal methods in `ECSqlStatement` (`toRow`, `getMetadata`, `bindParams`) to support row streaming and metadata extraction for the new executor.

**Cache Management Improvements:**

* Centralized cache clearing logic in `ECDb` and `IModelDb` by adding a `clearCaches` method that clears statement caches and notifies ECSqlRowExecutors to reset, improving consistency and reliability when detaching databases, closing, importing, or dropping schemas.

**Testing**
* Also hooked up the ECSqlTestRunner Test suite to use the `withQueryReader` api as well

**Performance Testing Results**
[CreateQueryReaderVsWithQueryReaderVsWithPreparedStatementPerformanceResults.csv](https://github.com/user-attachments/files/25424128/CreateQueryReaderVsWithQueryReaderVsWithPreparedStatementPerformanceResults.csv)

### DIFF
--- a/docs/changehistory/NextVersion.md
+++ b/docs/changehistory/NextVersion.md
@@ -11,7 +11,7 @@ publish: false
 
 ### WithQueryReader API
 
-A new [withQueryReader](../learning/backend/WithQueryReaderCodeExamples.md) method has been added to both [ECDb]($backend) and [IModelDb]($backend), providing true row-by-row behavior for ECSQL queries with synchronous execution. This API introduces a new [ECSqlSyncReader]($backend) through the [ECSqlRowExecutor]($backend) and supports configuration via [SynchronousQueryOptions]($backend).
+A new [withQueryReader]($docs/learning/backend/WithQueryReaderCodeExamples.md) method has been added to both [ECDb]($backend) and [IModelDb]($backend), providing true row-by-row behavior for ECSQL queries with synchronous execution. This API introduces a new [ECSqlSyncReader]($backend) through the [ECSqlRowExecutor]($backend) and supports configuration via [SynchronousQueryOptions]($backend).
 
 **Key Features:**
 


core/backend/src/ECSqlRowExecutor.ts
```diff
@@ -0,0 +1,540 @@
...
+  private _toBind: boolean = true;
+  private _rowCnt: number;
+  private _stats: ECSqlExecutionStats;
+
+  public constructor(private readonly db: IModelDb | ECDb) {
+    this._stmt = new ECSqlStatement();
+    this._toBind = true;
+    this._rowCnt = 0;
+    this._stats = new ECSqlExecutionStats();
+    this.db.notifyECSQlRowExecutorToBeReset.addListener(() => this.cleanup());
```
soham-bentley: Here I have used this event to notify all the `ECSQlRowExecutors` to dispose off their statements. But a problem with this approach is we are adding this event listener for each `ECSQlRowExecutors` but we are not able to remove them because `ECSqlReader` is not IDisposable so we can't make `ECSqlRowExecutors` as IDisposable

This will cause `ECSQlRowExecutors` objects to stay in memory until the imodelDb or ECDb object is disposed which is not a major object if we have very less row executors but can cause problems if there are many.

Any suggestions related to this?

Instead I can do one thing I can make the signature of `createQueryRowReader` to something like
```
createQueryRowReader(ecsql: string, callback:(reader: ECSqlReader) => void, params?: QueryBinder, config?: QueryOptionsForRowByRowReader)
```
If we make the signature like this then in the body of the `createQueryRowReader` I can use 
```
using rowExecutor = new ECSqlRowExecutor(this) // ECSqlRowExecutor can be made disposable now
```
@khanaffan  @rschili

core/backend/src/ECDb.ts
```diff
@@ -118,9 +123,18 @@ export class ECDb implements Disposable {
...
   public closeDb(): void {
+    this.clearCaches();
+    this[_nativeDb].closeDb();
+  }
+
+  /** Clear all in-memory caches held in this ECDb.
+   * @beta
+  */
+  public clearCaches(): void {
+    this.notifyECSQlRowExecutorToBeReset.raiseEvent();
```
soham-bentley: I have consolidated the clearCaches behavior in ECDb in a single method. As ECDb is public, commenting this to grab the attention of the reviewer.

core/backend/src/IModelDb.ts
```diff
@@ -852,6 +857,26 @@ export abstract class IModelDb extends IModel {
...
+   * - [ECSQL Overview]($docs/learning/backend/ExecutingECSQL)
+   * - [Code Examples]($docs/learning/backend/ECSQLCodeExamples)
+   * - [ECSQL Row Format]($docs/learning/ECSQLRowFormat)
+   *
+   * @param params The values to bind to the parameters (if the ECSQL has any).
+   * @param config Allow to specify certain flags which control how query is executed.
+   * @returns Returns an [ECSqlReader]($common) which helps iterate over the result set and also give access to metadata.
+   * @beta
+   * */
+  public createQueryRowReader(ecsql: string, params?: QueryBinder, config?: QueryOptionsForRowByRowReader): ECSqlReader {
```
khanaffan: I think we need to look into naming these method. RowReader does not sound right. We need it to look at it from caching perspective. 

The new method does not cache rows instead stream them as you step. While current one do paging and caching, it apply quota etc.

core/common/src/ConcurrentQuery.ts
```diff
@@ -149,6 +155,10 @@ export interface QueryOptions extends BaseReaderOptions {
    */
   rowFormat?: QueryRowFormat;
 }
+
+/** @beta */
+export type QueryOptionsForRowByRowReader = Omit<QueryOptions, "suppressLogErrors" | "limit" | "priority" | "restartToken" | "delay" | "quota">;
```
khanaffan: Can use `RowReaderOptions` instead of what you have there. But I think we need better name for RowReader as well.

core/common/src/ECSqlReader.ts
```diff
@@ -239,13 +239,18 @@ export class ECSqlReader implements AsyncIterableIterator<QueryRowProxy> {
...
+   */
   public setParams(param: QueryBinder) {
     if (this._lockArgs) {
       throw new Error("call resetBindings() before setting or changing parameters");
     }
     this._param = param.serialize();
   }
-
+  /**
+   * @deprecated Should not be used. Will be made private in a future release.
```
khanaffan: It should be something like following and add 1 years to current.
```
@deprecated in 4.11 - will not be removed until after 2026-06-13
```

khanaffan:
Can you past perf results in you description?

core/backend/src/IModelDb.ts
khanaffan: According to Claude base on execution model we can name them like
1.`CreateSynchronousQueryReader` & `SynchronousQueryOptions`
2. `CreateDirectQueryReader` & `DirectQueryOptions`

core/common/src/ECSqlReader.ts
soham-bentley: done

core/backend/src/IModelDb.ts
soham-bentley: donw. went ahead with `CreateSynchronousQueryReader` & `SynchronousQueryOptions`

core/common/src/ConcurrentQuery.ts
soham-bentley: donw. went ahead with `CreateSynchronousQueryReader` & `SynchronousQueryOptions`

common/api/core-backend.api.md
```diff
@@ -3667,6 +3681,8 @@ export abstract class IModelDb extends IModel {
     // @alpha
     createBRepGeometry(createProps: BRepGeometryCreate): IModelStatus;
     createQueryReader(ecsql: string, params?: QueryBinder, config?: QueryOptions): ECSqlReader;
+    // @beta
+    createQueryRowReader(ecsql: string, params?: QueryBinder, config?: QueryOptionsForRowByRowReader): ECSqlReader;
```
rschili: We're using lots of different naming for the same thing across this PR.
 
- ECSqlRowReader vs ECSqlReader
- QueryRowReader
- SynchronousQueryReader
- RowByRowReader

This needs to be consistent.
Personally I don't like the "row" naming at all. The other reader reads rows, too. The difference is that one uses caching ahead while the other does not.

We should find a name that perfectly reflects that, and use it all across the board.

Since we already have "Reader" which we cannot change, how about calling the other "DirectReader", or "OnDemandReader" or "StreamingReader". I lean towards "Direct", what do you think, @khanaffan ?

khanaffan:
So did you made sure performance test are comparable. e.g. BLOB is truncated ECSqlStatement/SyncReader case as it does in ECSqlReader case by default?

Following does not sound correct at all. createQueryReader is just more work on ECSqlStatement 

```
ECSqlStatement - SELECT * | Total time (ms) | 707
createQueryReader - SELECT * | Total time (ms) | 447
createSynchronousQueryReader  - SELECT * | Total time (ms) | 1358
```

soham-bentley:
> So did you made sure performance test are comparable. e.g. BLOB is truncated ECSqlStatement/SyncReader case as it does in ECSqlReader case by default?
> 
> Following does not sound correct at all. createQueryReader is just more work on ECSqlStatement
> 
> ```
> ECSqlStatement - SELECT * | Total time (ms) | 707
> createQueryReader - SELECT * | Total time (ms) | 447
> createSynchronousQueryReader  - SELECT * | Total time (ms) | 1358
> ```

Yes

core/backend/src/test/ecsql/queries/CteTests.ecsql.md
```diff
@@ -425,10 +425,10 @@ select y from (with tmp(x) as (SELECT e.i FROM aps.TestElement e order by e.i LI
 | --- |
 | 100 |
 
-# Testing table aliasing in CTE for ConcurrentQuery
+# Testing table aliasing in CTE for QueryReaders
 
 - dataset: AllProperties.bim
-- mode: ConcurrentQuery
+- mode: queryreader
```
MichaelSwigerAtBentley: Should these modes be in CamelCase

core/backend/src/ECDb.ts
```diff
@@ -445,4 +462,38 @@ export class ECDb implements Disposable {
...
+   *
+   * See also:
+   * - [ECSQL Overview]($docs/learning/backend/ExecutingECSQL)
+   * - [Code Examples]($docs/learning/backend/ECSQLCodeExamples)
+   * - [ECSQL Row Format]($docs/learning/ECSQLRowFormat)
+   * @param ecsql The ECSQL query to execute.
+   * @param callback the callback to invoke on the prepared ECSqlSyncReader
+   * @param params The values to bind to the parameters (if the ECSQL has any).
+   * @param config Allow to specify certain flags which control how query is executed.
+   * @returns the value returned by `callback`.
```
MichaelSwigerAtBentley, line 476:
```suggestion
   * @returns the value returned by `callback`.
   * @throws IModelError if db is not open or if error occurs during statement execution
```

core/backend/src/ECDb.ts
```diff
@@ -445,4 +462,38 @@ export class ECDb implements Disposable {
...
+  /** Allow to execute query and read results along with meta data. The result are stepped one by one.
+   *
+   * See also:
+   * - [ECSQL Overview]($docs/learning/backend/ExecutingECSQL)
+   * - [Code Examples]($docs/learning/backend/ECSQLCodeExamples)
+   * - [ECSQL Row Format]($docs/learning/ECSQLRowFormat)
+   * @param ecsql The ECSQL query to execute.
+   * @param callback the callback to invoke on the prepared ECSqlSyncReader
+   * @param params The values to bind to the parameters (if the ECSQL has any).
+   * @param config Allow to specify certain flags which control how query is executed.
```
MichaelSwigerAtBentley:
```suggestion
   * @param config Optional flags which control how query is executed.
```

core/backend/src/ECSqlRowExecutor.ts
```diff
@@ -0,0 +1,158 @@
...
+ * @internal
+ */
+export class ECSqlRowExecutor implements Disposable {
+  // eslint-disable-next-line @typescript-eslint/no-deprecated
+  private _stmt: ECSqlStatement;
+  private _removeListener: () => void;
+
+  public constructor(private readonly _db: IModelDb | ECDb) {
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
+    this._stmt = new ECSqlStatement();
```
MichaelSwigerAtBentley, line 42: Should this be IModelJsNative.ECSqlStatement? ECSqlStatement is slated for removal after June 2026 - Does it have functionality we need to port over?

core/backend/src/ECSqlRowExecutor.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  // --------------------------------------------------------------------------------------------
+  // Core execution
+  // --------------------------------------------------------------------------------------------
+
+  /** Prepare the statement and bind parameters in one step.
+   * Call once during reader initialization — avoids the per-row `ensureStatementReady` check.
+   * @param query - The ECSql text to prepare.
+   * @param args - Optional bind parameters.
+   * @throws Error on preparation or binding failure.
```
MichaelSwigerAtBentley:
```suggestion
   * @throws IModelError on preparation or binding failure.
```

core/backend/src/ECSqlStatement.ts
MichaelSwigerAtBentley, line 1: Same comment about deprecation -  is this something we want to keep using moving forward?

core/backend/src/ECSqlSyncReader.ts
```diff
@@ -0,0 +1,207 @@
...
+import { ECSqlRowExecutor } from "./ECSqlRowExecutor";
+import { DbResult } from "@itwin/core-bentley";
+
+/** @beta */
+export type SynchronousQueryOptions = Omit<QueryOptions, "suppressLogErrors" | "includeMetaData" | "limit" | "priority" | "restartToken" | "delay" | "usePrimaryConn" | "quota">;
+
+/**
+ * Execute ECSQL statements synchronously and read the results one row at a time.
+ *
+ * This is the synchronous counterpart of [[ECSqlReader]] from `@itwin/core-common`.
```
MichaelSwigerAtBentley, line 23: So ECSqlReader is async, ECSqlSyncReader is sync. Should they both live side by side in core-common?

core/backend/src/IModelDb.ts
```diff
@@ -852,6 +854,40 @@ export abstract class IModelDb extends IModel {
...
+   *
+   * See also:
+   * - [ECSQL Overview]($docs/learning/backend/ExecutingECSQL)
+   * - [Code Examples]($docs/learning/backend/ECSQLCodeExamples)
+   * - [ECSQL Row Format]($docs/learning/ECSQLRowFormat)
+   * @param ecsql The ECSQL query to execute.
+   * @param callback the callback to invoke on the prepared ECSqlReader
+   * @param params The values to bind to the parameters (if the ECSQL has any).
+   * @param config Allow to specify certain flags which control how query is executed.
+   * @returns the value returned by `callback`.
```
MichaelSwigerAtBentley, line 868:
```suggestion
   * @returns the value returned by `callback`.
   * @throws IModelError if db is not open
```

core/backend/src/IModelDb.ts
MichaelSwigerAtBentley, line 833: Can we update these comments to clarify when this should be used vs withQueryReader()

MichaelSwigerAtBentley:
Can we add a section in NextVersion.md for this

MichaelSwigerAtBentley:
Also, have we done performance metrics on the new readers compared to the old ones? Interested in if this new Sync reader will fix the performance issue @DanRod1999 is working on

core/backend/src/test/ecsql/queries/CteTests.ecsql.md
soham-bentley: In the markdown testing framework the case doesnot really matter we can use any case

core/backend/src/ECSqlRowExecutor.ts
soham-bentley: I am just using the typescript ECSqlStatement class which wraps IModelJsNative.ECSqlStatement. We can directly use IModeljsNative.ECSqlStatement but I saw other pirces of code we always generally have wrappers which wraps IModelJsNative typescript objects
Even after deprecation we can make them internal
And I just need two or three wrapper methods rest we can just remove

core/backend/src/ECSqlSyncReader.ts
soham-bentley: Because the sync reader should only be used in core-backend not in core-common frontend apps should not use this api

core/backend/src/test/ecsql/queries/CteTests.ecsql.md
soham-bentley: Done

core/backend/src/ECDb.ts
soham-bentley: done

core/backend/src/ECDb.ts
soham-bentley: done

core/backend/src/ECSqlRowExecutor.ts
soham-bentley: done

core/backend/src/IModelDb.ts
soham-bentley: done

core/backend/src/IModelDb.ts
soham-bentley: done
Also there is very detailed documentation added related to this as part of the PR

soham-bentley:
> Also, have we done performance metrics on the new readers compared to the old ones? Interested in if this new Sync reader will fix the performance issue @DanRod1999 is working on

I have added detailed performance csv as part of the PR description

core/backend/src/ECSqlRowExecutor.ts
soham-bentley: Although I can actually use IModelJsNative.ECSqlStatement directly here as well...just let me know which seems more favorable as per standards.....I just followed the standard approach where I see we donot really use IModelJsNative package classes directly we do have a wrapper over them.....after deprecation we can just make ECSqlStatement class internal ( and keep just the methods we use internally)  I guess

core/backend/src/IModelDb.ts
MichaelSwigerAtBentley: Yup, I saw - that documentation looks great btw

DanRod1999:
> > Also, have we done performance metrics on the new readers compared to the old ones? Interested in if this new Sync reader will fix the performance issue @DanRod1999 is working on
> 
> I have added detailed performance csv as part of the PR description

It seems like `createQueryReader()` was the fastest, and this new reader is the slowest? Am I understanding the data correctly? Also is `createQueryReader()` in your tests using `primaryConn: true`?

soham-bentley:
> > > Also, have we done performance metrics on the new readers compared to the old ones? Interested in if this new Sync reader will fix the performance issue @DanRod1999 is working on
> > 
> > 
> > I have added detailed performance csv as part of the PR description
> 
> It seems like `createQueryReader()` was the fastest, and this new reader is the slowest? Am I understanding the data correctly? Also is `createQueryReader()` in your tests using `primaryConn: true`?

Yes I am using usePrimaryConn: true. The new api is not the slowest it is comparable to ECSqlStatement api for most cases sometimes a bit faster than ECSqlStatement api and sometimes a bit slower than ECSqlStatement api

MichaelSwigerAtBentley:
> Yes I am using usePrimaryConn: true. The new api is not the slowest it is comparable to ECSqlStatement api for most cases sometimes a bit faster than ECSqlStatement api and sometimes a bit slower than ECSqlStatement api

So these tests do not capture the scenario that @DanRod1999 is seeing in transformer - And given that the async version should be faster, using the sync option shouldn't help either, no?

rschili:
I may sound like a broken record at this point, but I just don't see the point of deprecating and replacing an API with one that's slower, just to have two consistent APIs.

Afaik still nobody requested that they want this.
Our time would probably be better invested exploring options like this https://github.com/iTwin/itwinjs-backlog/issues/1889 to optimize for performance.

This change would add 2300 lines of code that we have to maintain, with possibly zero consumers. Again please correct me if I'm missing somebody who wants this.

DanRod1999:
> but I just don't see the point of deprecating and replacing an API with one that's slower, just to have two consistent APIs.

I reviewed the data and I don't see it as that much slower than the deprecated api,but the older api does seem generally a bit faster:
image

I'm also going to try to create a test that runs the same query multiple times and compare perf, and will add those findings here

soham-bentley:
> > but I just don't see the point of deprecating and replacing an API with one that's slower, just to have two consistent APIs.
> 
> I reviewed the data and I don't see it as that much slower than the deprecated api,but the older api does seem generally a bit faster: image
> 
> I'm also going to try to create a test that runs the same query multiple times and compare perf, and will add those findings here

@DanRod1999 
I have this https://github.com/iTwin/itwinjs-backlog/issues/1897 to track this whole performance stuff
Can we use this issue for all the performance related discussions?
I actually have mentioned the reason there why it is a bit slower generally

khanaffan (approved):
All new api should be at minimum marked as beta .

rschili:
> This change would add 2300 lines of code that we have to maintain, with possibly zero consumers. Again please correct me if I'm missing somebody who wants this.

There was some confusion because I approved the PR despite my comment :-)

We discussed this in a call yesterday, and there are usecases that I didn't know about and didn't see in the linked issues. Mostly, we deprecated statement because it supported insert/update/delete and apparently our plan always has been to eventually replace it with a new API. This is it.

So this is not primarily about performance or resolving confusing behavior, instead it's about having a sync API at all going forwards. I'm good with that.

MichaelSwigerAtBentley:
>Mostly, we deprecated statement because it supported insert/update/delete and apparently our plan always has been to eventually replace it with a new API. This is it.

By statement you are referring to ECSqlStatement? If this is the replacement, then we should not be using ECSqlStatement in the new solution, no? See my unresolved comment above 

If we are referring to a different statement, then I am good to resolve my issue and approve

khanaffan:
> > Mostly, we deprecated statement because it supported insert/update/delete and apparently our plan always has been to eventually replace it with a new API. This is it.
> 
> By statement you are referring to ECSqlStatement? If this is the replacement, then we should not be using ECSqlStatement in the new solution, no? See my unresolved comment above
> 
> If we are referring to a different statement, then I am good to resolve my issue and approve

* `ECSqlStatement` will be replaced with `ECSqlWriteStatement` which internal to test and use with `ECDb` when doing insert, update & delete.
* `ECSqlStatement` for `ECDb` and `iModelDb` will be replace with `SynchronousReader`. 
*`SynchronousReader` use `ECSqlStatement` which is fine. We will just make ECSqlStatement internal and remove unnecessary/unused api  e.g. `ColumnInfo` stuff from it.

core/backend/src/ECSqlRowExecutor.ts
soham-bentley: Not applicable anymore

core/backend/src/IModelDb.ts
soham-bentley: Not applicable anymore
Went ahead with `WithQueryReader` and `ECSqlSyncReader` and `SynchronousQueryOptions`

core/common/src/ConcurrentQuery.ts
soham-bentley: Not applicable anymore
Went ahead with `WithQueryReader` and `ECSqlSyncReader` and `SynchronousQueryOptions`

core/common/src/ECSqlReaderBase.ts
```diff
@@ -0,0 +1,250 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
```
ben-polinsky: You'll need to add a group for the docs (otherwise it takes the file name as default)

```
/** @packageDocumentation
 * @module iModels
 */
```

soham-bentley: done @ben-polinsky

soham-bentley:
> Can we add a section in NextVersion.md for this

This is also added @MichaelSwigerAtBentley

docs/changehistory/NextVersion.md
```diff
@@ -3,3 +3,64 @@ publish: false
...
 
+- [NextVersion](#nextversion)
+  - [@itwin/core-backend](#itwincore-backend)
+    - [WithQueryReader API](#withqueryreader-api)
+
+## @itwin/core-backend
+
+### WithQueryReader API
+
+A new [withQueryReader](../../docs/learning/backend/WithQueryReaderCodeExamples.md) method has been added to both [ECDb]($backend) and [IModelDb]($backend), providing true row-by-row behavior for ECSQL queries with synchronous execution. This API introduces a new [ECSqlSyncReader]($backend) through the [ECSqlRowExecutor]($backend) and supports configuration via [SynchronousQueryOptions]($backend).
```
ben-polinsky: Arg, now this is throwing. You can use the dollar shortcut to reference the base url.
`[withQueryReader]($docs/learning/backend/WithQueryReaderCodeExamples.md)`

soham-bentley: done

PR #8976, opened 2026-02-10
Extend extraction code into ecschema-metadata and core-frontend

hl662:
This pull request introduces `betools extract` into ecschema-metadata and core-frontend, incorporating them into `rush docs` workflow.

The goal is to have example code snippets meant for users to always be a part of actively maintained, tested code. This was difficult as previously the majority of example code snippets were housed in `example-code` folder, which mixes frontend and backend tests. However, many of these snippets aren't actually tested, only instead part of declared exported functions, as valid frontend tests would need a proper browser environment. 

One solution - Since each package has it's own test infrastructure set up, we can split the tests off into their respective packages, allowing verifiable tests that will always be run as part of our CI. Establish a standard that the extraction script only runs against files found within `test/example-code`

Documentation workflow improvements:

* Added a section in `CONTRIBUTING.md` describing how to extract documentation code snippets from test files, including instructions for marking extraction regions, updating scripts, and referencing snippets in markdown documentation.
* Updated `package.json` scripts in `core/ecschema-metadata` and `core/frontend` to include an `extract` script for snippet extraction and to chain extraction with documentation generation.

Test file and methodology updates:
* Moved and refactored example test files (`Ratio.test.ts`, `UnitConversion.test.ts`) to `src/test/example-code`, updated them to use `vitest` assertions instead of `chai`
* Moved `RatioUnits.ecschema.xml` for ratio formatting examples over to ecschema-metadata

Was able to verify by testing out a docs release off our testing environment, the extracted code showed up.

### DIFF
--- a/core/frontend/vitest.config.mts
+++ b/core/frontend/vitest.config.mts
@@ -11,7 +11,7 @@ const testSchemaFiles = [
   '@bentley/units-schema/Units.ecschema.json',
   '@bentley/formats-schema/Formats.ecschema.json',
   '@bentley/aec-units-schema/AecUnits.ecschema.json',
-].map((specifier) => require.resolve(specifier));
+].map((specifier) => require.resolve(specifier).replace(/\\/g, "/"));
 
 const includePackages: string[] = [
   ...Object.entries(packageJson.peerDependencies)


ben-polinsky (approved):
Thanks!

CONTRIBUTING.md
```diff
@@ -9,8 +9,7 @@ The goal of this document is to provide a high-level overview of how you can get
   - [Repo Setup](#repo-setup)
   - [Source Code Edit Workflow](#source-code-edit-workflow)
     - [Other NPM Scripts](#other-npm-scripts)
-    - [Debugging](#debugging)
```
eringram, line 12: These might have been accidentally deleted when the Documentation Code Snippets section was added

core/frontend/src/test/example-code/quantity/FormatsProvider.test.ts
```diff
@@ -32,26 +32,23 @@ class ExampleFormatProvider implements MutableFormatsProvider {
...
-    const locFormats = new SchemaXmlFileLocater();
-    locFormats.addSchemaSearchPath(schemaFile)
-    schemaContext.addLocater(locFormats);
-
-    const aecSchemaFile = path.join(__dirname, "..", "..", "node_modules", "@bentley", "aec-units-schema");
-    const locAec = new SchemaXmlFileLocater();
-    locAec.addSchemaSearchPath(aecSchemaFile)
-    schemaContext.addLocater(locAec);
+
+    // Load schemas via fetch (browser-compatible) in dependency order
```
aruniverse, line 40: i never understood the old approach of trying to manually resolve the schema files from node modules like this. why arent we using something like `require.resolve`?

this new approach of copying assets from `node_modules` into some test assets dir seems more fragile and more bloated. we should reconsider this imo

hl662: Sure, I've changed it up so the vitest config file will copy the static assets into the testing environment using `require.resolve`. This means i can remove the custom script in package.json and the change in gitignore.

Originally, I wanted to do 
```ts
import unitsSchemaJson from "@bentley/units-schema/Units.ecschema.json";
// ...other schemas imports
```

at the top of the test file, vite should be able to resolve that, but our schema npm packages don't have an exports field defined **just yet**

It's something I asked for and it seems we are close to getting it: https://github.com/iTwin/bis-schemas/pull/654

aruniverse: why even bother to shift this into vite config?
we should push that get pr you linked merged and released and do it the right way instead of accruing more temporary tech debt

naveedkhan8067: > Sure, I've changed it up so the vitest config file will copy the static assets into the testing environment using `require.resolve`. This means i can remove the custom script in package.json and the change in gitignore.
> 
> Originally, I wanted to do
> 
> ```ts
> import unitsSchemaJson from "@bentley/units-schema/Units.ecschema.json";
> // ...other schemas imports
> ```
> 
> at the top of the test file, vite should be able to resolve that, but our schema npm packages don't have an exports field defined **just yet**
> 
> It's something I asked for and it seems we are close to getting it: [iTwin/bis-schemas#654](https://github.com/iTwin/bis-schemas/pull/654)

The [PR](https://github.com/iTwin/bis-schemas/pull/654) has merged and this feature will be available for schema versions released after this change. For older schema versions its not available.

Issue #8774, opened 2025-11-14
iTwinJs Core using errors for basic code flow


ColinKerr:
I see three main code flows where we are throwing errors from some methods and then branching on those errors inside catch block.

1) The first place is here https://github.com/iTwin/itwinjs-core/blob/soham-bentley/debugging-exceptions/core/backend/src/IModelDb.ts#L1286-L1294

<img width="1342" height="656" alt="Image" src="https://github.com/user-attachments/assets/3e38b848-5a4d-446e-afbd-fe4fa14db7aa" />

Here as we can see in the catch block if the error has the error number of `IModelStatus.NotFound` we branch on it and load meta data for this class first and then try to get the class again. 
Upon inspection I found that 
https://github.com/iTwin/itwinjs-core/blob/soham-bentley/debugging-exceptions/core/backend/src/ClassRegistry.ts#L309-L312

<img width="2050" height="274" alt="Image" src="https://github.com/user-attachments/assets/a0c17348-bb24-4a85-88c1-b820e9068e76" />

`getClass` can internally throw error from `generateClass`

https://github.com/iTwin/itwinjs-core/blob/soham-bentley/debugging-exceptions/core/backend/src/ClassRegistry.ts#L281-L292

<img width="2408" height="810" alt="Image" src="https://github.com/user-attachments/assets/9573ea2f-7e01-4876-a07f-ec703829ff2d" />

`generateClass` can throw errors from the following three lines 
https://github.com/iTwin/itwinjs-core/blob/c760c18fe826df828561d964cff75afdcca4cc6d/core/backend/src/ClassRegistry.ts#L285
https://github.com/iTwin/itwinjs-core/blob/c760c18fe826df828561d964cff75afdcca4cc6d/core/backend/src/ClassRegistry.ts#L289
https://github.com/iTwin/itwinjs-core/blob/c760c18fe826df828561d964cff75afdcca4cc6d/core/backend/src/ClassRegistry.ts#L292

Among these three errors , when the metadata for that particular `className` comes as undefined we try to handle it manually inside the catch block (https://github.com/iTwin/itwinjs-core/blob/soham-bentley/debugging-exceptions/core/backend/src/IModelDb.ts#L1286-L1294) by loading the metadata and trying again...the second time the metadata will surely not be undefined.
[...]

PR #8978, opened 2026-02-11
Better exceptions from getElementProps and similar

### LINKED ISSUE #8509: Misleading error thrown by getElementProps

kring:
Fixes #8509 

This PR modifies the following methods of `IModelDb`:

- `getModelProps`
- `getModel`
- `getSubModel`
- `getElementProps`
- `getElement`

Previously, these methods would throw an `IModelError(NotFound)` when pretty much anything went wrong. We even had a test that was checking that we get a NotFound error when the real problem was that the database had been closed.

Now, they aim to provide accurate exception information. An exception thrown by these methods will always be an `IModelError` (because this is the documented public API), but:

1) The `message` will reflect the original exception.
2) The `errorNumber` will reflect the original exception, if the original exception has one (i.e., if it's a `BentleyError`). Otherwise it will be `BadRequest`.
3) The `cause` of the `IModelError` will be a reference to the original exception, providing access to the original call stack.

In addition, the `try` versions of all of these functions have been changed. They now consistently return `undefined` when anything goes wrong that prevents returning the requested object. This was _mostly_ true before, but now they're guaranteed not to throw any exceptions. See [here](https://github.com/iTwin/itwinjs-core/issues/8509#issuecomment-3876169060) for more background on this choice.

To avoid a lot of redundant code, I also introduced a new type, called `Expected`. This is closely modeled on C++'s `std::expected` or Rust's `Result`. The idea is it can encapsulate either a successful result value or an error (IModelError), and has some nice (monadic!) functions for chaining operations together. This helps us implement both the `try` and the throwing versions of all these functions easily, without a lot of unnecessary throwing and catching of exceptions.

### DIFF
--- a/docs/learning/api-support-policies.md
+++ b/docs/learning/api-support-policies.md
@@ -26,13 +26,26 @@ TypeScript's type system and the JavaScript environment in which it executes bot
   - Adding a new property or method with a name that is likely to conflict with existing subclasses.
   - Changing the visibility of a property or method.
   - Changing the type of a property.
-- Functions and methods
-  - Changing the signature (parameters and/or return type).
-    - Exception: appending new arguments with default values.
-    - Exception: changing the return type from `void` to something else.
+    - Exception: changing from a super-type to a sub-type on a property that is not intended to be overridden.
+- Methods intended to be overridden
+  - Changing the signature (parameters and/or return type) at all.
+- Functions and other methods not intended to be overridden
+  - Changing the return type to something that is not a sub-type of the previous. A more specific return is acceptable; a more general one is not.
+  - Changing the non-void return type to void.
+  - Changing a parameter type to something that is not a super-type of the previous. A more general parameter is acceptable; a more specific one is not.
 - Enums and tagged unions
   - Removing values from an enum or types from a union.
   - Adding new values to an enum or types to a tagged union, unless it is unlikely that existing code tests exhaustively for all possible values.
+- Thrown exceptions
+  - Changing the type of the exception thrown in a particular error scenario.
+  - Throwing an exception in a case that previously did not throw.
+
+The following are _not_ considered breaking changes, though we will make a best-effort attempt to document them when they happen:
+
+- Thrown exceptions
+  - Changing the message associated with an exception.
+  - Changing the errorNumber, iTwinErrorId, or other code associated with an exception to make it more accurate or specific.
+  - Producing a result in a scenario that previously threw an exception.
 
 The iTwin.js team strives to avoid breaking changes whenever possible. We also endeavor to design our APIs with flexibility to evolve without the need for breaking changes - for example, by defining functions to receive their parameters as a single object type instead of multiple individual arguments.
 


core/backend/src/IModelDb.ts
```diff
@@ -2540,41 +2553,42 @@ export namespace IModelDb {
...
-     * @throws [[IModelError]] if the element exists, but cannot be loaded.
+     * @returns The element or `undefined` if the element is not found, cannot be loaded, or fails validation when `elementClass` is specified.
      * @note Useful for cases when an element may or may not exist and throwing an `Error` would be overkill.
      * @see getElement
      */
     public tryGetElement<T extends Element>(elementId: Id64String | GuidString | Code | ElementLoadProps, elementClass?: EntityClassType<Element>): T | undefined {
+      return this.tryGetElementImpl<T>(elementId, elementClass).valueOrDefault(undefined);
+    }
+
+    public tryGetElementImpl<T extends Element>(
```
GytisCepk: Does this have to be public ?

Note: since `IModelDb` is marked as `@public`, this method will inherit it's status as public and this doesn't look like something we want to be part of public API.

kring: Oops, that's an oversight - now fixed. Good eye!

But I also made another similar method public, `tryGetElementPropsImpl`, that one purposely. I did this because it's used from `tryGetSubModelImpl`, which is in a different class and therefore can't access a private method. Sadly TS doesn't have `friend`, which might otherwise be an ok alternative.

We definitely don't want it in the public API, though. I was hoping that leaving it undocumented would do that. Is there another way to exclude it, like an `@internal` tag in a doc comment or similar?

GytisCepk: `@internal` tag should be enough, even though ideally we would be using that as little as possible.

While personally I'm not a fan, you can also look into how we use symbols to prevent users from using internal APIs ([Symbols.ts](https://github.com/iTwin/itwinjs-core/blob/master/core/backend/src/internal/Symbols.ts))

anmolshres98:
> This is not quite ready for review yet, but I'm opening a draft PR to let CI run on it.

The CI build is configured to not run on draft PRs. I tried using the `/azp` command and it seems that also will not overwrite the draft pr configuration. I opened the build pipeline on ADO and manually ran the build on your branch. You can manually do this by going to the pipeline on ADO and selecting "Run pipeline" button on top right then selecting your branch name.

If you are not ready to have eyes on your changes yet but want to run the build on it, you can always run the build directly on your branch by going to the ADO webpage like above. You just need to have the branch published, that is all 😄. draft PR should not be needed.

kring:
Thanks for the info @anmolshres98! It looks like I don't have the necessary access though?
image

kring:
Copilot has some good points, I'll address most of it before merging this.

core/backend/src/Expected.ts
```diff
@@ -0,0 +1,142 @@
...
+class ExpectedError<T> extends ExpectedBase<T> {
+  public constructor(public readonly error: IModelError) { super(); }
+}
+
+/**
+ * Represents a value that may either be a successful result (a value of type `T`) or an error (an `IModelError`).
+ */
+type Expected<T> = ExpectedValue<T> | ExpectedError<T>;
+
+namespace Expected { // eslint-disable-line @typescript-eslint/no-redeclare
```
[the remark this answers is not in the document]
kring: It's reasonable for `Expected` to be part of the public API, but I'll mark it internal until we have a clear reason to do so.

ben-polinsky:
These make sense, but I am worried they are breaking changes. Not by the letter of our [api policy](https://www.itwinjs.org/learning/api-support-policies/#breaking-changes), but certainly by its spirit. We can't pull the rug out from users who are relying on this admittedly bad behavior. I know @aruniverse disagreed in the above thread.

Also [relevant discussion](https://github.com/iTwin/itwinjs-core/issues/8774) surrounding a similar `Expected` type idea.

tcobbs-bentley:
I'm not an expert on iTwin errors, but I do know that we added [`ITwinError` functionality](https://www.itwinjs.org/reference/core-bentley/errors/itwinerror/) a while back, and later [tweaked it](https://github.com/iTwin/itwinjs-core/pull/7876#issue-2939412318). I want to make sure that this isn't going against any of the `ITwinError` paradigms. Looking at the code, it seems that `IModelError` implements the `ITwinError`, although it doesn't appear to declare that it does so. So everything regarding `ITwinError` might be fine.

anmolshres98:
I agree with @ben-polinsky that this could be considered a breaking change to the extent that if we know consumers are depending on that bad behavior. For example in this diff 

image

If we have consumers with code relying on error to be thrown with the old message and making conditional checks based on that behavior (which I don't doubt is happening), this would be breaking change for them and we should absolutely wait for a major release.

To @tcobbs-bentley 's point, we are moving to using `iTwinError` over `iModelError` but I think you are correct on leaving the thrown error to `iModelError`, changing that to `iTwinError` would also be breaking. Given my and ben's points above though, it might be better to change iModelError to throw iTwinError and wait till a major version bump to merge all these breaking changes. However, if the consensus is that changing thrown error message is not a breaking change then we should keep the iModelError thrown as you have now and change them to iTwinErrors at a later time during a major version bump.

kring:
Thanks for taking a look!

Ultimately, people with more experience and context on this project than myself will need to make the judgement call about breaking changes. API evolution is always a judgement call. Even in the case of a clear-cut bug, it's always possible to imagine some poor soul relying on the broken behavior such that the fix will break them.

But here's my two cents:

1. I don't think exception _messages_ should ever be considered part of the public API. Sure, sometimes developers do branch on messages (mostly when they have no other choice) but I think it's a clear case of relying on implementation details. Users will appreciate us documenting such changes in the changelog, but I don't think we need to shy away from making them.
2. Exception types and error codes are part of the public API. So changing them is a breaking change and we should strive to avoid it.
3. In this particular case, I think users are _extremely unlikely_ to have written any conditional logic on exception types, error codes, or messages. Why? Because `getElementProps` only ever threw one exception type (`IModelError`), with one `errorNumber` (`NotFound`), with one message. Even when this was a lie. That's what this PR is aiming to fix. So users would gain nothing by attempting to branch on any of this information, and thus we're very unlikely to impact them if we change it. Even if it's technically a breaking change.
4. I've said `getElementProps` above because I know what I've written above to be true for it. I think it's also true for the other functions affected by this PR, but I'm not as certain of it. I'll double check that and report back.
5. If we think the risk of these changes outweighs their benefit, and we want to close this PR, I won't be offended!

kring:
Here's a summary of the old and new behavior of each function:

| *Function* | *master* | *This branch* | *Breaking?* |
|-------------|-----------|----------------|--------------|
| `tryGetElementProps` | Returns `undefined` for any error. | Unchanged | No |
| `getElementProps` | Throws `IModelError(NotFound)` for any error. | Throws `IModelError` for any error, but not necessarily `NotFound`. | Very unlikely |
| `tryGetModelProps` | Returns `undefined` for any error. | Unchanged | No |
| `getModelProps` | Throws `IModelError(NotFound)` for any error. | Throws `IModelError` for any error, but not necessarily `NotFound`. | Very unlikely |
| `tryGetModel` | Returns `undefined` if `tryGetModelProps` or `constructEntity` returns undefined. Throws an exception if `constructEntity` throws an exception. | Returns `undefined` for any error. | Maybe |
| `getModel` | Throws `IModelError(NotFound)` when `tryGetModel` returns undefined. Any exception thrown by `constructEntity` is passed through. |  Throws `IModelError` for any error, but not necessarily `NotFound`. | Maybe |
| `tryGetElement` | Returns `undefined` if `tryGetElementProps` or `constructEntity` returns undefined. Throws an exception if `constructEntity` throws an exception. | Returns `undefined` for any error. | Maybe |
| `getElement` | Throws `IModelError(NotFound)` when `tryGetElement` returns undefined. Any exception thrown by `constructEntity` is passed through. |  Throws `IModelError` for any error, but not necessarily `NotFound`. | Maybe |
| `tryGetSubModel` | Returns undefined if `tryGetElementProps` or `tryGetModel` return undefined. Throws an exception if `tryGetModel` throws an exception. | Returns `undefined` for any error. | Maybe |
| `getSubModel` | Throws `IModelError(NotFound)`, or any other exception thrown by `getModel`. | Throws `IModelError` for any error, but not necessarily `NotFound`. | Maybe |

To summarize the summary, the only potentially-concerning breaking change here (IMO) is this:

All of the "try" and "non-try" functions that call `constructEntity`  - i.e., all the ones with "Maybe" in the last column - would previously allow any exception it throws to escape them. In this PR, the "try" functions will swallow that exception and return undefined, while the "non-try" functions will wrap it in an `IModelError` (if it isn't one already).

ben-polinsky:
@kring Great summary (and summary of summary) - much appreciated. I am definitely being overly-cautious. The history of iTwin.js is that we've broken too many things in the past and made updates frustrating for our users. (Your poor-soul comment did make me laugh.) This has led to the very strict and elongated deprecation period.

With that said, I do see a bit of evidence in some of our internal repos code branching off a caught not found error, though not on the methods you're changing. It's a tough call to make. I'd like to err on the side of caution and defer these to a major change, or at least the "maybe"s you've identified. Happy to here other thoughts and be convinced otherwise.

aruniverse:
tagging @johnnyd710 @karolis-zukauskas for their perspective as consumers

> I don't think exception messages should ever be considered part of the public API. Sure, sometimes developers do branch on messages (mostly when they have no other choice) but I think it's a clear case of relying on implementation details. Users will appreciate us documenting such changes in the changelog, but I don't think we need to shy away from making them.

personally, I am leaning towards agreeing to this

> In this particular case, I think users are extremely unlikely to have written any conditional logic on exception types, error codes, or messages. 

@ben-polinsky , @anmolshres98  I appreciate the cautiousness. We should be able to quickly verify existing usage by grepping on gh & ado

ben-polinsky:
> @ben-polinsky , @anmolshres98 I appreciate the cautiousness. We should be able to quickly verify existing usage by grepping on gh & ado

I did, and we switch off `NotFound` all the time when catching, just not in the methods above I saw on the apps I searched.

core/backend/src/IModelDb.ts
```diff
@@ -2274,21 +2275,23 @@ export namespace IModelDb {
...
+     * @returns The ModelProps or `undefined` if the model is not found or cannot be loaded.
      * @note Useful for cases when a model may or may not exist and throwing an `Error` would be overkill.
      * @see getModelProps
      */
     public tryGetModelProps<T extends ModelProps>(id: Id64String): T | undefined {
-      try {
+      return this.tryGetModelPropsImpl<T>(id).valueOrDefault(undefined);
+    }
+
+    private tryGetModelPropsImpl<T extends ModelProps>(
```
MichaelSwigerAtBentley, line 2291: This is really cool - just want to make sure I'm understanding this correctly:

These implmentations return an 'Expected<T>' which is either an Expected value or an Expected Error of type T. That gives you the option of calling one of the "isValue()", "isError()", "valueOrThrow()" ect checks that will return the actual underlying value, type, or error (depending on the check performed) and can be used as such. So then we can call the same implementation from both the 'try' and 'non-try' versions of the getFunction and then use the Expected checks to filter out the results.

MichaelSwigerAtBentley:
> Why? Because getElementProps only ever threw one exception type (IModelError), with one errorNumber (NotFound), with one message. Even when this was a lie.

Exactly, I do not think that the change to getElementProps should be considered breaking. Previously, we just swallowed up any meaningful error and replaced it with NotFound. If anything, this is adding functionality to allow for conditional error handing, as it gives the option of NotFound being returned correctly, or actual errors bubbling up.

core/backend/src/IModelDb.ts
MichaelSwigerAtBentley: With this change, this comment should be removed - and the returns should be updated.

johnnyd710:
Potentially problematic example I found?

```ts
async function tryGetElementProps(iModel: IModelConnection, arg: Id64Arg) {
  try {
    return await iModel.elements.getProps(arg);
  } catch (error) {
    if (error instanceof BentleyError && error.errorNumber === IModelStatus.NotFound) {
      Logger.logInfo(LogCategory.Frontend, error.message, error);
      return [];
    } else
      throw error;
    }
  }
}
```

This pattern exists in frontend code (IModelConnection.elements.getProps), but errors originate in backend IModelDb.elements.getElementProps and are propagated through RPC. With previous behavior, all errors were `IModelStatus.NotFound`, so this helper effectively returned `[]` for any errors. Now, this would throw in many cases that previously returned `[]`.

core/backend/src/IModelDb.ts
```diff
@@ -2349,15 +2346,24 @@ export namespace IModelDb {
...
+    }
 
-      return model instanceof modelClass ? model : undefined;
+    private tryGetModelImpl<T extends Model>(
+      modelId: Id64String,
+      modelClass?: EntityClassType<Model>
+    ): Expected<T> {
+      return this
+        .tryGetModelPropsImpl<ModelProps>(modelId)
+        .map(modelProps => {
```
MichaelSwigerAtBentley: And here - if modelProps is an error, map will return it as an ExpectedError or do some work on it and then return it as ExpectedValue wrapped model.

MichaelSwigerAtBentley:
> Potentially problematic example I found?
> 
> ```ts
> async function tryGetElementProps(iModel: IModelConnection, arg: Id64Arg) {
>   try {
>     return await iModel.elements.getProps(arg);
>   } catch (error) {
>     if (error instanceof BentleyError && error.errorNumber === IModelStatus.NotFound) {
>       Logger.logInfo(LogCategory.Frontend, error.message, error);
>       return [];
>     } else
>       throw error;
>     }
>   }
> }
> ```
> 
> This pattern exists in frontend code (IModelConnection.elements.getProps), but errors originate in backend IModelDb.elements.getElementProps and are propagated through RPC. With previous behavior, all errors were `IModelStatus.NotFound`, so this helper effectively returned `[]` for any errors. Now, this would throw in many cases that previously returned `[]`.

but the intent to throw actual errors is there, its a bug that it currently returns [] on non NotFound errors - something we should not try to preserve

ben-polinsky:
> **a bug** that it currently returns [] on non NotFound errors - something we should not try to preserve

I'd love to solidify this into a policy. If we have a policy, we can let our users know - hey, we're allowed to break things in the name of fixing buggy, yet still workable behavior. The current implementation is buggy: but is it something that people rely on? We can sit here and debate all we want, but if we don't have a clear policy which drives user expectations, we're guessing. And our recent history says we err on the side of caution. If we want to change that, it can't be a one-off. Otherwise, we'll have this debate every time it happens.

Edit: I also don't think searching through our repos for examples is a satisfactory or good practice for this. Otherwise, our policy breaks down, and it's always a game of "well, it doesn't look too bad for these handful of users we know about."

MichaelSwigerAtBentley:
> I'd love to solidify this into a policy. If we have a policy, we can let our users know - hey, we're allowed to break things in the name of fixing buggy, yet still workable behavior. The current implementation is buggy: but is it something that people rely on? 

I also think its worth asking, in all of these cases where a user is checking for NotFound -> are they expecting us to not misreport the error - or are they _intentionally_ relying on our broken error reporting?

But yes, I agree, we should clarify policy around this.

ben-polinsky:
> I also think its worth asking, in all of these cases where a user is checking for NotFound -> are they expecting us to not misreport the error - or are they intentionally relying on our broken error reporting?

Yes, good point. 
My conscious still tells me to err on the side of caution. @johnnyd710?

johnnyd710:
> If we have a policy, we can let our users know - hey, we're allowed to break things in the name of fixing buggy, yet still workable behavior.

I agree with this ^

> > I also think its worth asking, in all of these cases where a user is checking for NotFound -> are they expecting us to not misreport the error - or are they intentionally relying on our broken error reporting?
> 
> Yes, good point. My conscious still tells me to err on the side of caution. @johnnyd710?

I don't know what the intention was - but the `else` never actually ran if all errors were NotFound errors. That probably wasn't intentional. Anyways we shouldn't use that one example as a reason to not do this. Correctness should be the goal, so preserving original error codes is the right direction.

core/backend/src/IModelDb.ts
kring: That's right. This is a super common pattern in functional programming - a restricted version of what Haskell calls the `Either` monad - but it's useful in more traditional languages, too. I think of it as kind of a middle ground between exceptions (which make error handling "easy" in that you don't need to write a lot of code to handle errors most of the time, but create non-linear control flow that is sometimes hard to follow and reason about) and explicit error returns (which require more manual handling everywhere, even in the common case when you just want to bubble the error up to the caller).

core/backend/src/IModelDb.ts
kring: Yes that's right.

I actually changed this spot just now to an `andThen` instead of a `map`. The difference is that `map` takes a function from `T` -> `U` and applies it to the value (if it's not an error). Just like `map` on Array applies a function to every element. While `andThen` instead takes a function from `T` -> `Expected<U>`, which means the function can "fail" without the need to throw an exception. Not that throwing an exception is terrible or anything, but it's a bit wasteful to throw an exception only to catch it and turn it into an Expected. `andThen` lets us go straight to the Expected.

You might notice that `andThen` works a lot like `Promise.then` when the continuation itself returns a Promise. That's not a coincidence!

kring:
I've just pushed a change to the api-support-policies document for everyone's consideration. It's admittedly self-serving in that this PR is ok by the rules I've written down. From my perspective, they're reasonable, but let me know what you think.

ben-polinsky:
> I've just pushed a change to the api-support-policies document for everyone's consideration. It's admittedly self-serving in that this PR is ok by the rules I've written down. From my perspective, they're reasonable, but let me know what you think.

Thanks. I agree these are mostly good additions. What I'm still uneasy about:

```md
The following are _not_ considered breaking changes...

- Thrown exceptions
  - Changing the message associated with an exception. _(Agree, no one should be writing logic off the message.)_
  - Changing the errorNumber, iTwinErrorId, or other code associated with an exception to make it more accurate or specific. _(This one still sticks out to me as problematic.)_
  - Producing a result in a scenario that previously threw an exception. _(Also seems problematic.)_
```

(Notes attempted inline...)
Summary: I'm in favor of keeping error message changes as non-breaking - but code changes and throw/return flips seem like potentially breaking - at a minimum needs to be clearly documented as bug-fix exception with migration notes? 

@karolis-zukauskas @johnnyd710 @MichaelSwigerAtBentley

karolis-zukauskas:
>Summary: I'm in favor of keeping error message changes as non-breaking - but code changes and throw/return flips seem like potentially breaking - at a minimum needs to be clearly documented as bug-fix exception with migration notes?

Agree.

All of these can be breaking changes - each case needs to be considered (impact). I'd suggest to not create hard rules and phrase it in a way where platform has the right to make this change if it deems necessary (value outweighs risk/damage). Surely such changes need to be documented properly.

For example I think the change to throw more accurate errors from getElement instead of always NotFound is fine. Changing error messages should be fine too in most cases. Though people tend to branch on error message because they have no other option .. I created multiple issues for platform in the past, where there was no way to properly catch and handle an error (should be way better now with `ITwinError` available)

ben-polinsky:
@kring and I spoke briefly on Teams, so I'll recap here. For now, we need to abide 100% by our breaking changes/api guidance. 

However, our current policy is a result of our failure to handle breaking changes gracefully in the past. It specifies a lengthy timeline to ensure our consumers re-gain confidence in API release stability.

We'd like to get to a place -  and whatever decisions we make must be codified - where we might be a bit more flexible with some of these policies, but work hand in hand with application teams ahead of the changes to ensure they hit the road running.

core/backend/src/IModelDb.ts
```diff
@@ -2274,21 +2275,23 @@ export namespace IModelDb {
...
-      return model;
+      return this.tryGetModelPropsImpl<T>(id).valueOrThrow();
     }
 
     /** Get the ModelProps with the specified identifier.
      * @param modelId The Model identifier.
-     * @returns The ModelProps or `undefined` if the model is not found.
-     * @throws [[IModelError]] if the model cannot be loaded.
+     * @returns The ModelProps or `undefined` if the model is not found or cannot be loaded.
      * @note Useful for cases when a model may or may not exist and throwing an `Error` would be overkill.
```
MichaelSwigerAtBentley, line 2284: @JulijaRamoskiene The issues around inconsistent Error returns from the get... functions should be corrected here. (along with the incorrect docs)

PR #8979, opened 2026-02-11
Presentation: Content fields selector

grigasp:
Some consumers use the `getElementProperties` API for loading properties of every iModel's element. In some iModels, elements have a huge number of properties (specific case that I looked at had 50k total number of properties), and consumers only need some of them - up to a hundred. Reducing the number of loaded properties can let us drop some unnecessary joins, speeding up the loading by a lot.

- Added `fieldsSelector` callback prop to `PresentationManager.getElementProperties` function to allow consumers specify which fields / properties they want.
- Added `fieldsSelector` attribute to content `Descriptor` and related types.
- Added `Descriptor.selectedFields` getter that filters its `fields` based on `fieldsSelector`. The default content parser used by `getElementProperties` now uses `selectedFields` instead of `fields`.
- Fixed a problem that caused the default content parser to drop some properties from the content.

### DIFF
--- a/common/changes/@itwin/core-common/bugfix-black-text-color-is-falsy_2026-02-12-14-39.json
+++ b/common/changes/@itwin/core-common/bugfix-black-text-color-is-falsy_2026-02-12-14-39.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
+      "comment": "Fix a bug in a 'if' statement with a color def TBGR numeric value.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-common"
+}
--- a/core/common/src/geometry/ElementGeometry.ts
+++ b/core/common/src/geometry/ElementGeometry.ts
@@ -435,7 +435,7 @@ export namespace ElementGeometry {
         let result: boolean;
         if (entry.text) {
           result = this.appendTextString(new TextString(entry.text));
-        } else if (entry.color) {
+        } else if (undefined !== entry.color) {
           const params = geomParams?.clone() ?? new GeometryParams(Id64.invalid);
           if (entry.color !== "subcategory") {
             params.lineColor = ColorDef.fromJSON(entry.color);


presentation/backend/src/presentation-backend/PresentationManager.ts
```diff
@@ -707,8 +707,18 @@ export class PresentationManager {
...
-      this.getContentDescriptor({ ...contentOptions, displayType: DefaultContentDisplayTypes.Grid, contentFlags: ContentFlags.ShowLabels, ...partialProps });
+    const descriptorGetter = async (partialProps: Pick<ContentDescriptorRequestOptions<IModelDb, KeySet, RulesetVariable>, "rulesetOrId" | "keys">) => {
+      const descr = await this.getContentDescriptor({
+        ...contentOptions,
+        displayType: DefaultContentDisplayTypes.Grid,
+        contentFlags: ContentFlags.ShowLabels,
+        ...partialProps,
+      });
+      if (descr) {
+        descr.fieldsSelector = fieldsSelectorOption?.(descr);
```
ben-polinsky, line 718: nit/question: Are the parens necessary: `?.(descr);`?

presentation/common/src/presentation-common/content/Descriptor.ts
```diff
@@ -453,6 +467,21 @@ export class Descriptor implements DescriptorSource {
...
   }
 
+  /** Get selected fields based on `fields` in this descriptor and `fieldsSelector`. */
+  public get selectedFields(): Field[] {
+    if (!this.fieldsSelector) {
+      return this.fields;
+    }
+
+    const { type, fields: selectedFields } = this.fieldsSelector;
+    switch (type) {
```
saskliutas: Maybe it's worth caching result of `selectedFields` until `fieldsSelector` changes?

ben-polinsky:
Also, maybe worth mentioning in presentation docs.

presentation/backend/src/presentation-backend/PresentationManager.ts
grigasp: this is a function call, so yes

presentation/common/src/presentation-common/content/Descriptor.ts
```diff
@@ -453,6 +467,21 @@ export class Descriptor implements DescriptorSource {
...
+      case "include":
+        return exclusivelyIncludeFields(this.fields, selectedFields);
+      case "exclude":
+        return excludeFields(this.fields, selectedFields);
```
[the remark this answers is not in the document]
grigasp: we use typescript to ensure we don't get unexpected values...

presentation/common/src/presentation-common/content/Descriptor.ts
```diff
@@ -474,6 +503,145 @@ export class Descriptor implements DescriptorSource {
...
+    ...field,
+    nestedFields: [...field.nestedFields],
+  });
+}
+
+function exclusivelyIncludeFields(descriptorFields: Field[], targetFieldDescriptors: FieldDescriptor[]): Field[] {
+  const rootFields: Field[] = [];
+  const clones = new Map<Field, NestedContentField>();
+
+  for (const targetFieldDescriptor of targetFieldDescriptors) {
```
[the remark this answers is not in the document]
grigasp: I don't fields order is relevant for our use cases. Even if it is for some consumer, field descriptors are build from the fields list, so consumers can control the order and keep it the same as in fields list, if necessary.

presentation/common/src/presentation-common/PresentationManagerOptions.ts
```diff
@@ -219,6 +219,13 @@ export interface MultiElementPropertiesBaseRequestOptions<TIModel, TParsedConten
    */
   contentParser?: (descriptor: Descriptor, item: Item) => TParsedContent;
 
+  /**
+   * A callback that allows to specify which fields should be included or excluded in the result based on the given content descriptor. This
```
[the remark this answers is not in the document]
grigasp: fixed

presentation/common/src/presentation-common/ElementProperties.ts
```diff
@@ -177,11 +177,17 @@ class ElementPropertiesAppender implements IPropertiesAppender {
...
-    }
+    // create an ordered list of categories/appenders starting with categories that have the most ancestors and finishing with categories
+    // that have no ancestors, so that when we call `finish` on appenders, child categories are finished before parent categories, otherwise
+    // we may skip parent categories, thinking they have no items
+    const categoriesNestedToRoot = new SortedArray<{ category: CategoryDescription; appender: IPropertiesAppender }>(
+      (lhs, rhs) => countAncestors(rhs.category) - countAncestors(lhs.category),
+      DuplicatePolicy.Allow,
+    );
+    Object.entries(this._categoryItemAppenders).forEach(([_, { category, appender }]) => {
+      categoriesNestedToRoot.insert({ category, appender });
```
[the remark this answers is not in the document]
grigasp: good point, fixed

presentation/common/src/presentation-common/content/Descriptor.ts
grigasp: good idea, made the change

presentation/backend/src/presentation-backend/PresentationManager.ts
ben-polinsky: right, thanks.

PR #8983, opened 2026-02-12
ElementGeometry.appendTextBlock. Fix incorrect 'if' comparison against color

a-gagnon:
### Fix:

This PR fixes an issue where the `entry.color` is black, meaning its value is `0`. I ended up in the `else` statement and couldn't understand why there was always a line segment in my text annotation.

It might be a good idea to turn on [@typescript-eslint/strict-boolean-expressions](https://typescript-eslint.io/rules/strict-boolean-expressions/) to avoid getting such error.

### DIFF
--- a/common/changes/@itwin/core-common/bugfix-black-text-color-is-falsy_2026-02-12-14-39.json
+++ b/common/changes/@itwin/core-common/bugfix-black-text-color-is-falsy_2026-02-12-14-39.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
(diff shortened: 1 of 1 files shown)


markschlosseratbentley (approved):
LGTM. Thanks for this catch.

> It might be a good idea to turn on [@typescript-eslint/strict-boolean-expressions](https://typescript-eslint.io/rules/strict-boolean-expressions/) to avoid getting such error.

@ben-polinsky @hl662 What do you think of the above suggestion?

a-gagnon:
@mergifyio backport release/5.6.x

ben-polinsky:
> @ben-polinsky @hl662 What do you think of the above suggestion?

sure

PR #8988, opened 2026-02-12
Make unit system optional in SchemaFormatsProvider

### LINKED ISSUE #8980: @itwin/presentation-common no longer supports implicit unit conversion since 5.1.0
**Describe the bug**

`KoqPropertyValueFormatter.format` no longer automatically convert values and units when presented with an override format specification string, e.g. `"Formats.DefaultRealU(6)[Units.SQ_YRD]"`, the persistence unit system is "metric" (SQ_M), but the presentation system is supposed to be "imperial" (SQ_YRD).

Specifically, prior to 5.1.0, format would return a value as "42.673 yd²". Since 5.1.0, format returns "35.68 m²" without any unit conversion.

**To Reproduce**

1. Checkout version 5.0.5 of itwinjs-core.
2. Modify [KoqPropertyValueFormatter.test.ts](https://github.com/iTwin/itwinjs-core/blob/release/5.0.5/presentation/common/src/test/KoqPropertyValueFormatter.test.ts)
3. Add a new test:
```ts
    it("FIXME formats value using an implicit unit system conversion override", async () => {
      const formatted = await formatter.format(1.5, {
        koqName: "TestSchema:TestKoqImplicitOverride"
      });
      expect(formatted).to.be.eq(`1,5 ${imperialUnit.label}`);
    });

    // etc...

const koqImplicitOverrideProps: KindOfQuantityProps = {
  schemaItemType: SchemaItemType.KindOfQuantity,
  name: "TestKoqImplicitOverride",
  schema: "TestSchema",
  relativeError: 6,
  presentationUnits: ["TestSchema.SiFormat[TestSchema.ImperialUnit]"],
  persistenceUnit: "TestSchema.MetricUnit"
}

// etc...

const schemaProps: SchemaProps = {
  $schema: "https://dev.bentley.com/json_schemas/ec/32/ecschema",
  name: "TestSchema",
  alias: "test",
  version: "1.0.0",
  items: {
    [koqImplicitOverrideProps.name!]: koqImplicitOverrideProps,
  },
};
```
4. Verify the test passes.
5. Check out the latest and repeat the test.

The test will fail:

```
  1) KoqPropertyValueFormatter
base.js:337
       with default props
base.js:337
         format
base.js:337
           FIXME formats value using an implicit unit system conversion override:
base.js:337

base.js:337
      AssertionError: expected '1.5 Metric' to equal '1,5 Imperial'
base.js:337
      + expected - actual
base.js:337

base.js:337
      -1.5 Metric
base.js:337
      +1,5 Imperial
```

**Expected behavior**
It is expected that when presented with a presentation override format specification, that the value and unit system be honored.

**Screenshots**
N/A

- OS: Windows
- Browser: N/A
- Version: N/A
- iTwin.js Version 5.6.0

**Additional context**

This issue was encountered upgrading Bentley's NextGen connectors to use the latest itwinjs. Feel free to reach out to me on Teams for more context. Our specific test case was "Update Unit to compatible unit" which would take an existing BIM file via a briefcase and then add several ECClass to it. One of them contained a new ECProperty with the unit YARD_SQUARED. The test then formatted an element and expected the value to be converted from meters to "yd²", but it did not. It formatted as "m²" without any conversion.

I debugged the issue and identified an unexpected change of behavior in 5.1.0 in itwin-core in `KoqPropertyValueFormatter` of **@itwin/presentation-common**. Specifically, [this](https://github.com/iTwin/itwinjs-core/blob/2cd6fe900cd4fd7df9d1aef56920f99de14f7873/presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts#L195) was removed in 5.1.0:

```ts
  if (koq.defaultPresentationFormat) {
    return getFormatProps(await koq.defaultPresentationFormat);
  }
```

I can see there was significant refactoring in this code, with the inclusion of a new [SchemaFormatsProvider.ts](https://github.com/iTwin/itwinjs-core/blob/release/5.0.5/core/ecschema-metadata/src/SchemaFormatsProvider.ts#L111-L147), which now handles the `defaultPresentationFormat`.

However, `SchemaFormatsProvider` seems to have an affinity for [picking formats based on the persistence unit system](https://github.com/iTwin/itwinjs-core/blob/2cd6fe900cd4fd7df9d1aef56920f99de14f7873/core/ecschema-metadata/src/SchemaFormatsProvider.ts#L94-L100). When defined, wouldn't the `persistenceUnit` _always_ match the `persistenceUnitSystem`?

```ts
    // If no matching presentation format was found, use persistence unit format if it matches unit system.
    const persistenceUnit = await kindOfQuantity.persistenceUnit;
    const persistenceUnitSystem = await persistenceUnit?.unitSystem;
    if (persistenceUnitSystem && unitSystemMatchers.some((matcher) => matcher(persistenceUnitSystem))) {
      this._formatsRetrieved.add(itemKey.fullName);
      return getPersistenceUnitFormatProps(persistenceUnit!);
    }
```

Nevertheless, disabling that chunk of code fixes that specific issue (but causes another).

hl662:
Closes #8980

This pull request updates the logic for handling unit systems in format selection, making the unit system optional in `SchemaFormatsProvider` and ensuring that default presentation formats are used when a unit system is not specified. It also improves how cross-system presentation format overrides are handled, and adds comprehensive test coverage for these scenarios.

**Format selection logic improvements:**

* Made the `unitSystem` parameter optional in the `SchemaFormatsProvider` constructor and property, allowing format retrieval without requiring a unit system. If not provided, the default presentation format is used directly.

**Presentation logic and formatter updates:**

* Modified `KoqPropertyValueFormatter` to avoid defaulting to the metric unit system when none is provided, and to support undefined unit systems throughout the formatting logic. 

**Testing and validation:**

* Added new tests to verify correct behavior when using cross-system format overrides, and when no unit system is provided, including new test assets and KindOfQuantity definitions.

### DIFF
--- a/presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
+++ b/presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
@@ -164,7 +164,7 @@ export class KoqPropertyValueFormatter {
     // `SchemaFormatsProvider` will fall back to default presentation format, but we want to fall back
     // to default formats' map first, and only then to the default presentation format. All of this can
     // be removed with the removal of default formats map.
-    if (this._defaultFormats && (!formatProps || (await getUnitSystemKey(this._unitsProvider, formatProps)) !== unitSystem)) {
+    if (unitSystem && this._defaultFormats && (!formatProps || (await getUnitSystemKey(this._unitsProvider, formatProps)) !== unitSystem)) {
       const defaultFormatProps = await getFormatPropsFromDefaultFormats({
         schemaContext: this._schemaContext,
         formatsMap: this._defaultFormats,
@@ -228,7 +228,7 @@ async function getFormatPropsFromDefaultFormats({
   schemaContext: SchemaContext;
   // eslint-disable-next-line @typescript-eslint/no-deprecated
   formatsMap: FormatsMap;
-  unitSystem: UnitSystemKey | undefined;
+  unitSystem: UnitSystemKey;
   koqName: string;
 }): Promise<FormatProps | undefined> {
   const koq = await getKoq(schemaContext, koqName);
@@ -249,7 +249,7 @@ async function getFormatPropsFromDefaultFormats({
     for (const defaultUnitSystemFormat of Array.isArray(defaultPhenomenonFormats)
       ? /* c8 ignore next */ defaultPhenomenonFormats
       : [defaultPhenomenonFormats]) {
-      if (unitSystem && defaultUnitSystemFormat.unitSystems.includes(unitSystem)) {
+      if (defaultUnitSystemFormat.unitSystems.includes(unitSystem)) {
         return defaultUnitSystemFormat.format;
       }
     }


rschili (approved):

presentation/common/src/presentation-common/KoqPropertyValueFormatter.ts
```diff
@@ -250,7 +249,7 @@ async function getFormatPropsFromDefaultFormats({
     for (const defaultUnitSystemFormat of Array.isArray(defaultPhenomenonFormats)
       ? /* c8 ignore next */ defaultPhenomenonFormats
       : [defaultPhenomenonFormats]) {
-      if (defaultUnitSystemFormat.unitSystems.includes(unitSystem)) {
+      if (unitSystem && defaultUnitSystemFormat.unitSystems.includes(unitSystem)) {
```
saskliutas: Looks like call to `getFormatPropsFromDefaultFormats` can be skipped if `unitSystem` is `undefined` as this condition will always fail.

hl662: makes sense - I reverted my changes to getFormatPropsFromDefaultFormats and added a check against `unitSystem` in getFormattingProps where this method is used

jamie-bentley:
@hl662, do you know when this will be released?

hl662:
Hi jamie thanks for following up! this will be available in version 5.7.0, that is slated to launch next week. We have released the first release candidate for 5.7.0: `5.7.0-dev.16` if you are able to test locally

jamie-bentley:
@hl662, I can confirm that `5.7.0-dev.16` fixes the issue locally.

PR #8989, opened 2026-02-13
Validate in ViewDefinition2d constructor that provided baseModelId is valid.

### LINKED ISSUE #8150: `getElement` call on `ViewDefinition` element right after inserting is failing in iTwin.JS 5.x
**Describe the bug**
`iModel.elements.getElement` operation on an `ViewDefinition` element immediately after insertion fails as `Not Found: Element=...` despite insertion operation not throwing. Here:
```ts
// Insert operation below does not throw, returns elementId
const drawingViewId = DrawingViewDefinition.insert(
      iModel,
      definitionModelId,
      "Drawing View",
      "0",
      drawingCategorySelectorId,
      displayStyle2dId,
      new Range2d(0, 0, 100, 100)
    );
// getElement operation below throws with `Not Found: Element=...`
iModel.elements.getElement(drawingViewId);
```

**To Reproduce**
I have setup a minimal repo with the reproduction of this issue:
1. Clone https://github.com/anmolshres98/core-backend-minimal-repro
2. Run `pnpm i` then `pnpm clean:build:start` - this will throw error in console at location where `getElement` is called
3. You can edit package.json to use `4.11.1` then redo step 2 - this will not throw any errors
4. You can also edit the `const cfg: IModelHostOptions = {};` in line 12 so that it is `disableThinnedNativeInstanceWorkflow: true`. This also does not throw the error
5. You can switch through other dev versions I have mentioned above to pinpoint the problem release
> Note: In my reproduction repo, I used the code that lead to this issue in my app repo but there are likely other workflows that trigger this issue

**Expected behavior**
Either, the `getElement` operation should return as normal without throwing or the `insert` operation should throw if the provided props are not correct

- OS: MacOS
- Node Version: v22
- iTwin.js Version: 5.0.0-dev.114

**Additional context**
Upon further debugging, I can see that the issue is coming from within the call to `classDef.deserialize` inside `IModelDb.js` [here](https://github.com/iTwin/itwinjs-core/blob/eecefb202cb43c5d65b9e363193a038443bcf355/core/backend/src/IModelDb.ts#L2051). Inside that call, an attempt to access `rowInstance.baseModel.id` is made [here](https://github.com/iTwin/itwinjs-core/blob/eecefb202cb43c5d65b9e363193a038443bcf355/core/backend/src/ViewDefinition.ts#L726). In this case `baseModel` is undefined so `baseModel.id` access attempt fails and throws leading to the above error to the final calling code.

The assumption in the code seems to be that `baseModel` is always defined. Is that assumption correct? If that is the case this should be enforced when the element is inserted as well. If not, the code should be modified so that `baseModel.id` is not attempted to access and instead should probably be replaced with `baseModel?.id`.

kring:
This makes it consistent with similar validation of `categorySelectorId` and `displayStyleId` in `ViewDefinition`.

Fixes #8150 

I haven't been able to convince myself one way or the other yet if there _also_ needs to be some validation of this property somewhere in imodel-native. But I think the change in this PR makes sense either way.

### DIFF
--- a/core/backend/src/test/standalone/ViewDefinition.test.ts
+++ b/core/backend/src/test/standalone/ViewDefinition.test.ts
@@ -317,7 +317,6 @@ describe("ViewDefinition", () => {
 
   describe("DrawingViewDefinition", () => {
     it("fails on insert without a valid baseModelId", () => {
-      const sourceFileName = "some.bim";
       const subjectId = Subject.insert(
         iModel,
         IModel.rootSubjectId,


ben-polinsky:
Need to lint, but otherwise looks good. Thanks.

Issue #8966, opened 2026-02-09
Presentation: Allow specifying which properties to load in `getElementProperties` backend API


grigasp:
Consumer is using the backend `PresentationManager.getElementProperties` API to load properties for all elements in the iModel. Some specific iModels have a huge amount of properties (close to 50k), making the API very slow. However, the consumer only needs a much smaller set of properties (up to a 100), and reducing the amount of properties would make the API substantially quicker.

We should provide a way for consumers to specify which fields / properties to load with `getElementProperties`.

grigasp:
Closing the issue as complete. The changes made don't cover the single element case, but that's not something that consumers requested, so probably not worth spending our time on.

PR #8990, opened 2026-02-16
Add fieldsSelector to single element properties API

AbhijeetFasate13:
- Added fieldsSelector property to SingleElementPropertiesRequestOptions
- Implemented filtering logic in getSingleElementProperties method
- Addresses issue #8966

### DIFF
--- a/common/api/core-bentley.api.md
+++ b/common/api/core-bentley.api.md
@@ -1385,9 +1385,13 @@ export type NonFunctionPropertyNamesOf<T> = {
 // @public
 export class ObservableSet<T> extends Set<T> {
     constructor(elements?: Iterable<T> | undefined);
+    addAll(items: Iterable<T>): number;
     clear(): void;
     delete(item: T): boolean;
+    deleteAll(items: Iterable<T>): number;
     readonly onAdded: BeEvent<(item: T) => void>;
+    readonly onBatchAdded: BeEvent<() => void>;
+    readonly onBatchDeleted: BeEvent<() => void>;
     readonly onCleared: BeEvent<() => void>;
     readonly onDeleted: BeEvent<(item: T) => void>;
 }
--- a/common/api/core-frontend.api.md
+++ b/common/api/core-frontend.api.md
@@ -1971,6 +1971,7 @@ export namespace CatalogConnection {
 export class CategorySelectorState extends ElementState {
     constructor(props: CategorySelectorProps, iModel: IModelConnection);
     addCategories(arg: Id64Arg): void;
+    addCategoriesBatched(arg: Id64Arg): void;
     // (undocumented)
     get categories(): Set<string>;
     set categories(categories: Set<string>);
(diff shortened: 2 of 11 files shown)


AbhijeetFasate13:
@grigasp Can you review this?

presentation/backend/src/presentation-backend/PresentationManager.ts
```diff
@@ -650,14 +650,36 @@ export class PresentationManager {
...
+      const emptyDescriptor = new Descriptor({
+        connectionId: "",
+        inputKeysHash: "",
+        displayType: DefaultContentDisplayTypes.PropertyPane,
+        selectClasses: [],
+        fields: [],
+        categories: [],
+        contentFlags: 0,
+      });
+      const selection = fieldsSelector(emptyDescriptor);
```
grigasp, line 674: Empty descriptor is not a valid argument for `fieldsSelector` callback - it needs to receive a real descriptor with fields so it has something to choose from. 

To pass it a valid argument, we'd first have to request a descriptor, then pass it to `fieldsSelector`, then call `getContent`. However, splitting the content request into 2 requests would actually make the performance worse for 1 element case.

presentation/common/src/presentation-common/PresentationManagerOptions.ts
```diff
@@ -201,6 +201,13 @@ export interface SingleElementPropertiesRequestOptions<TIModel, TParsedContent =
   /** ID of the element to get properties for. */
   elementId: Id64String;
 
+  /**
+   * A callback that allows specifying which fields should be included or excluded in the result based on the given content descriptor. This
+   * is useful when requesting properties of elements with a very large number of fields, where it may be desirable to only get a
+   * subset of all available fields for performance improvement.
+   */
+  fieldsSelector?: (descriptor: Descriptor) => DescriptorFieldsSelector | undefined;
```
grigasp, line 209: `SingleElementPropertiesRequestOptions` is also used by frontend's [PresentationManager.getElementProperties](https://github.com/iTwin/itwinjs-core/blob/3303839f731aa23f515e2e51901e89cbb212edb9/presentation/frontend/src/presentation-frontend/PresentationManager.ts#L727), so would have to implemented there, too.

grigasp (commented):
[The issue](https://github.com/iTwin/itwinjs-core/issues/8966) doesn't state it, but we have the performance problem, when loading properties for many elements, and that's been fixed with https://github.com/iTwin/itwinjs-core/pull/8979.

The single-element case would be a good improvement for consistency, however, as stated in one of the comments, it can't be implemented through descriptor overrides without making the performance worse. The approach I'd take is to:
1. Request content like we did before your changes.
2. If `fieldsSelector` is provided, clone the content's descriptor and replace `fields` with original descriptor's `selectedFields`.
3. Pass the cloned descriptor to `parser` instead of the original one.

Also, because `SingleElementPropertiesRequestOptions` is used by both - frontend and backend - `PresentationManager.getElementProperties` methods, this needs to be implemented on both sides.

PR #8998, opened 2026-02-18
Change CloudSqlite `validateDbVersion` to fallback to default when ve…

hl662:
This pull request updates the CloudSqlite version handling logic in the `@itwin/core-backend` package to ensure that any falsy database version value (now including an empty string) will now default to `"0.0.0"`. This improves robustness when dealing with missing or empty version strings. The update also includes new tests to verify this fallback behavior.

* Changed the `validateDbVersion` function in `CloudSqlite.ts` to use a fallback to `"0.0.0"` for any falsy version value, not just `undefined` or null.

Little tidbit on difference between `??` and `||`:
`a ?? b` uses `b` only when a is `null` or `undefined`
`a || b` uses `b` when `a` is any falsy value: `undefined`, `null`, `""`, 0, `false`, `NaN`.

### DIFF
--- a/common/changes/@itwin/core-backend/nam-fix-csqlite-versioning_2026-02-16-22-32.json
+++ b/common/changes/@itwin/core-backend/nam-fix-csqlite-versioning_2026-02-16-22-32.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Change CloudSqlite `validateDbVersion` to fallback to default when version is any falsy value",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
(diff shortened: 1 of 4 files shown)


anmolshres98:
> Little tidbit on difference between `??` and `||`:
> `a ?? b` uses `b` only when a is `null` or `undefined`
> `a || b` uses `b` when `a` is any falsy value: `undefined`, `null`, `""`, 0, `false`, `NaN`.

wow, good to know. thanks

ben-polinsky:
Would it be stronger to make "0.0.0" the default for the argument?

hl662:
> Would it be stronger to make "0.0.0" the default for the argument?

No, it wouldn't solve the side effect - a `""` being passed in will not lead to `0.0.0` be assigned if you changed the argument

PR #8999, opened 2026-02-18
Implemented CurveFactory.fromFilletedLineString

saeeedtorabi:


aruniverse:
@bbastings @FelixGirard mind taking a pass reviewing?

bbastings:
@FelixGirard @saeeedtorabi I've previously questioned whether this can actually be used since it fails if it encounters a single arc that is not a fillet. What does your current/custom implementation do Felix? To be generally useful, the method ought to return the non-fillet arcs, and non-linear segments as CurvePrimitives in the output along with the points/radius for the path sections that could be parsed.

FelixGirard:
From what I understood from the input for a valid fillet, we should be ok to use what it produce. We are more flexible as we support non tangent arc. We could still use this by having it run as relaxed and then remove zero-length segments.

Our current implementation  basically only parse the geometry and extract start and end points from the linestrings (and radius for an Arc3d). 
It does also add a point without radii when two curves are not parallel. I will share you the link to the full implementation.

saeeedtorabi:
> I've previously questioned whether this can actually be used since it fails if it encounters a single arc that is not a fillet. 

@bbastings Well, `CurveFactory.fromFilletedLineString` is reverse of `CurveFactory.createFilletsInLineString`.

`createFilletsInLineString` takes a set of points and radii and creates a filleted line string, it cannot have non-fillet arcs. Therefore, `fromFilletedLineString` does not allow non-fillet arcs.

I assume we can have a method that allows non-fillet arcs but it would be a different requirement and a different API.

FelixGirard:
After discussion with Brien, I realized what we actually need is probably not related to fillets.
We support any combination of arcs and linestrings. 

So basically a function to parse a geometry to an array of point and radii is what we need. 
They don't need to be filleted. We might still don't want cusp though.

I'm sorry if our requirements were not clear to begin with.

dassaf4:
I haven't looked at this PR yet, but as I understand the algorithm Saeed and I have been discussing:

It is still possible to use arcs that don't join with their neighbor with parallel tangents. You would have to insert a zero-length line segment at the end of the arc that isn't tangent to the neighbor. Arcs that are adjacent to each other would also need a zero-length line in between them. And arcs whose sweep is >= 180 degrees need to be split by a zero-length line.  These zero length lines are markers for the case where the arc (as a fillet) completely subsumes the adjacent segment(s). Without them, we cannot recreate that missing segment in the pre-filleted line string.

What is not handled by `CurveFactory.fromFilletedLineString`:
* elliptical arcs. These are not fillets where core-geometry is concerned.
* cusps (where the join tangents are anti-parallel).

saeeedtorabi:
> It is still possible to use arcs that don't join with their neighbor with parallel tangents. You would have to insert a zero-length line segment at the end of the arc that isn't tangent to the neighbor. Arcs that are adjacent to each other would also need a zero-length line in between them. And arcs whose sweep is >= 180 degrees need to be split by a zero-length line. These zero length lines are markers for the case where the arc (as a fillet) completely subsumes the adjacent segment(s). Without them, we cannot recreate that missing segment in the pre-filleted line string.

These are good test cases for me. I will add them to my tests to confirm this will work not only for filleted line strings but also more general cases. I want to know if `fromFilletedLineString` can handle paths like
image

Please note that if you set `relaxedValidation = true` then you dont have to insert a zero-length line segment at the end of the arc that isn't tangent to the neighbor (the API does it for you). You still need to break arcs with sweeps larger than 180.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -129,6 +129,28 @@ export interface CreateFilletsInLineStringOptions {
...
 }
 
+/**
+ * Options used for method [[CurveFactory.fromFilletedLineString]].
+ * @public
+ */
+export interface FilletedLineStringOptions {
+  /**
+   * Options for [[Vector3d.isParallelTo]].
+   *  Default: [[radianSquaredTol: Geometry.smallAngleRadiansSquared, distanceSquaredTol: Geometry.smallMetricDistanceSquared]].
```
dassaf4: Remove duplicate documentation of defaults. Reader can refer to the single source of truth: the documentation for `PerpParallelOptions`.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -129,6 +129,28 @@ export interface CreateFilletsInLineStringOptions {
...
+   */
+  parallelOptions?: PerpParallelOptions;
+  /**
+   * Distance tolerance for detecting equal points.
+   * Default: [[Geometry.smallMetricDistance]].
+   */
+  distanceTol?: number;
+  /**
+   * Whether to allow relaxed validation of the filleted linestring, which allows arcs to be adjacent to arcs.
```
dassaf4: This also allows arcs to meet linear elements at a corner. But I would not go into specifics here, as this can become stale text. Just say this option allows more chains to serve as valid input, leave it at that, and let the user see what this really means in the method doc.

saeeedtorabi:
@FelixGirard @bbastings 
I updated `CurveFactory.fromFilletedLineString`. Now you can set `relaxedValidation: true` and pass a path (with line segment, line string, or circular arc children) with any tangent between children and any arc sweep between 0 and 360. Now you will get a result rather than `undefined` and you dont need to insert zero-length line segments or break arcs. We do it for you under `relaxedValidation: true`.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+   * tangent. If `Arc3d` start/end tangent is non-parallel or is anti-parallel to the adjacent curve's end/start tangent,
+   * then caller should either set `options.relaxedValidation` to `true`, or add a zero-length `LineSegment3d` between
+   * the `Arc3d` and the adjacent curve to make it a valid filleted linestring.
+   * @param filletedLineString A filleted linestring usually created by [[CurveFactory.createFilletsInLineString]].
+   * @param options (optional) tolerances for distance and radian angle.
+   * @returns Array of [point, radius] pairs extracted from the filleted linestring, or `undefined` if the input is
+   * not a valid filleted linestring.
+   */
+  public static fromFilletedLineString(
+    filletedLineString: Path, options?: FilletedLineStringOptions,
```
dassaf4: Please change the input curve type to `CurveChain` for widest utility.

Internally all you have to do to a `Loop` is first thing create a `Path` with the same children.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+   * `Arc3d` then caller should either set `options.relaxedValidation` to `true` to allow it, or must add a zero-length
+   * `LineSegment3d` between the two `Arc3d` objects to make it a valid filleted linestring.
+   *   * Each `Arc3d` start/end tangent should be parallel (and not anti-parallel) to the adjacent curve's end/start
+   * tangent. If `Arc3d` start/end tangent is non-parallel or is anti-parallel to the adjacent curve's end/start tangent,
+   * then caller should either set `options.relaxedValidation` to `true`, or add a zero-length `LineSegment3d` between
+   * the `Arc3d` and the adjacent curve to make it a valid filleted linestring.
+   * @param filletedLineString A filleted linestring usually created by [[CurveFactory.createFilletsInLineString]].
+   * @param options (optional) tolerances for distance and radian angle.
+   * @returns Array of [point, radius] pairs extracted from the filleted linestring, or `undefined` if the input is
+   * not a valid filleted linestring.
```
dassaf4:
```suggestion
   * Extract points and radii from a valid filleted linestring.
   * * A valid filleted linestring is a `CurveChain` that satisfies the following conditions:
   *   * Its children have type `Arc3d`, `LineSegment3d`, or `LineString3d`.
   *   * Each `Arc3d` is circular.
   *   * Each `Arc3d` sweep is less than 180 degrees.
   *   * Each `Arc3d` cannot be adjacent to another `Arc3d`.
   *   * Each `Arc3d` is G1 continuous with each of its neighbors, i.e., at their common point, the curves have the same tangent direction.
   * * To treat more input chains as valid, pass `options.relaxedValidation = true`. Internally, this setting performs several transformations on the input to produce a valid filleted linestring:
   *   * Each `Arc3d` whose sweep is more than 180 degrees is split into two arcs separated by a zero-length `LineSegment3d`.
   *   * Adjacent `Arc3d`s are separated by a zero-length `LineSegment3d`.
   *   * An `Arc3d` that is not G1 continuous with its neighbor is separated from its neighbor by a zero-length `LineSegment3d`.
   * @param filletedLineString A linestring with corner fillets, e.g., as created by {@link createFilletsInLineString}.
   * @param options optional validation settings.
   * @returns Array of [point, radius] pairs extracted from input, or `undefined` if the input is not valid. A radius of zero means no fillet at the vertex.
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
     }
     return path;
   }
+  /** If an open filletedLineString starts/ends with an arc, add a zero-length line segment to its start/end. */
+  private static insertZeroLengthSegmentsAtArcBoundariesForOpenPath(filletedLineString: Path): Path {
```
dassaf4:
```suggestion
  private static validateOpenPathStartEnd(filletedLineString: Path): Path {
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+    if (firstChild instanceof Arc3d) {
+      const tempPath = Path.create();
+      tempPath.tryAddChild(LineSegment3d.create(firstChild.startPoint(), firstChild.startPoint()));
+      for (let i = 0; i < numOfChildren; i++) {
+        tempPath.tryAddChild(filletedLineString.children[i]);
+      }
+      filletedLineString = tempPath;
+    }
+    if (lastChild instanceof Arc3d) {
+      filletedLineString.tryAddChild(LineSegment3d.create(lastChild.endPoint(), lastChild.endPoint()));
```
dassaf4: 1. You've just changed the input, which is not good practice.
2. Write a test that fails with the current code and succeeds with your fix.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+    const numOfChildren = filletedLineString.children.length;
+    const firstChild = filletedLineString.children[0];
+    const lastChild = filletedLineString.children[numOfChildren - 1];
+    if (firstChild instanceof Arc3d) {
+      const tempPath = Path.create();
+      tempPath.tryAddChild(LineSegment3d.create(firstChild.startPoint(), firstChild.startPoint()));
+      for (let i = 0; i < numOfChildren; i++) {
+        tempPath.tryAddChild(filletedLineString.children[i]);
+      }
+      filletedLineString = tempPath;
```
dassaf4:
```suggestion
      const startPoint = firstChild.startPoint();
      filletedLineString = Path.create(LineSegment3d.create(startPoint, startPoint), ...filletedLineString.children);
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
     }
     return path;
   }
+  /** If an open filletedLineString starts/ends with an arc, add a zero-length line segment to its start/end. */
+  private static insertZeroLengthSegmentsAtArcBoundariesForOpenPath(filletedLineString: Path): Path {
+    const numOfChildren = filletedLineString.children.length;
```
dassaf4, line 259: You are implicitly assuming the caller has checked that the children array is nonempty, so you should add an `assert` here.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+   * Verify each neighboring curve to an arc is a line segment/string. Also verify arc tangents are parallel
+   * (but not anti-parallel) to neighboring line segment/string tangents.
+   */
+  private static validateArcNeighbors(
+    validatedFilletedLineString: Path, i: number, perpOptions?: PerpParallelOptions,
+  ): boolean {
+    const numOfChildren = validatedFilletedLineString.children.length;
+    const child = validatedFilletedLineString.children[i];
+    if (!(child instanceof Arc3d))
+      return false;
```
dassaf4: Safer:

```suggestion
    const child = validatedFilletedLineString.getChild(i);
    if (!child || !(child instanceof Arc3d))
      return false;
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+    // arc start tangent must be parallel (but not anti-parallel) to previous line segment
+    const arcStartTangent = child.fractionToPointAndDerivative(0).direction;
+    const prevChildEndTangent = prevChild.fractionToPointAndDerivative(1).direction;
+    if (!arcStartTangent.isParallelTo(prevChildEndTangent, false, true, perpOptions))
+      return false;
+    // (note: an open validatedFilletedLineString never ends with an arc, i.e, i=numOfChildren-1 never reaches here)
+    // next child after arc must be line segment/string
+    const nextChild = validatedFilletedLineString.children[(i + 1) % numOfChildren];
+    if (!(nextChild instanceof LineSegment3d) && !(nextChild instanceof LineString3d))
+      return false;
```
dassaf4: Faster:

```suggestion
    // next child after arc must be line segment/string
    const nextChild = validatedFilletedLineString.cyclicCurvePrimitive(i + 1); // ASSUME: i === numOfChildren - 1 only if the path is closed
    if (!nextChild || !(nextChild instanceof LineSegment3d || nextChild instanceof LineString3d))
      return false;
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+  ): boolean {
+    const numOfChildren = validatedFilletedLineString.children.length;
+    const child = validatedFilletedLineString.children[i];
+    if (!(child instanceof Arc3d))
+      return false;
+    // (note: an open validatedFilletedLineString never starts with an arc, i.e, i=0 never reaches here)
+    // previous child before arc must be line segment/string
+    const prevChild = validatedFilletedLineString.children[(i - 1 + numOfChildren) % numOfChildren];
+    if (!(prevChild instanceof LineSegment3d) && !(prevChild instanceof LineString3d))
+      return false;
```
dassaf4:
```suggestion
    // previous child before arc must be line segment/string
    const prevChild = validatedFilletedLineString.cyclicCurvePrimitive(i - 1); // ASSUME: i === 0 only if the path is closed
    if (!prevChild || !(prevChild instanceof LineSegment3d || prevChild instanceof LineString3d))
      return false;
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+   * between them.
+   * * If there is an arc with sweep >= 180 degrees, break the arc into 2 pieces and add a zero-length line segment
+   * between them.
+   */
+  private static updatePathForRelaxedValidation(
+    filletedLineString: Path, isClosed: boolean, parallelOptions?: PerpParallelOptions,
+  ): Path {
+    const newFilletedLineString = new Path();
+    const numOfChildren = filletedLineString.children.length;
+    for (let i = 0; i < numOfChildren; i++) {
```
dassaf4:
```suggestion
    for (let i = 0; i < numOfChildren; i++) { // examine each child and its predecessor
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+      const child = filletedLineString.children[i];
+      if (!isClosed && i === 0) {
+        newFilletedLineString.tryAddChild(child);
+        continue; // skip first child for open path since it won't have a previous child
+      }
+      const prevChild = filletedLineString.children[(i - 1 + numOfChildren) % numOfChildren];
```
dassaf4:
```suggestion
      const prevChild = filletedLineString.cyclicCurvePrimitive(i - 1);
      assert(prevChild !== undefined, "Cyclic neighbor is defined within loop extents");
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+      const twoConnectedArcs = child instanceof Arc3d && prevChild instanceof Arc3d;
+      // apply relaxed validation rules
+      if (twoConnectedArcs || twoNonParallelArcLsPair)
+        newFilletedLineString.tryAddChild(LineSegment3d.create(child.startPoint(), child.startPoint()));
+      if (child instanceof Arc3d && Math.abs(child.sweep.sweepDegrees) >= 180) {
+        const firstHalf = child.clonePartialCurve(0, 0.5);
+        const secondHalf = child.clonePartialCurve(0.5, 1);
+        newFilletedLineString.tryAddChild(firstHalf);
+        newFilletedLineString.tryAddChild(LineSegment3d.create(firstHalf.endPoint(), secondHalf.startPoint()));
+        newFilletedLineString.tryAddChild(secondHalf);
```
dassaf4:
```suggestion
      const arcLineStringCorner = !childrenAreParallel &&
        ((child instanceof Arc3d && (prevChild instanceof LineSegment3d || prevChild instanceof LineString3d)) ||
          ((prevChild instanceof Arc3d && (child instanceof LineSegment3d || child instanceof LineString3d))));
      const twoConnectedArcs = child instanceof Arc3d && prevChild instanceof Arc3d;
      // apply relaxed validation rules
      const linePoint = Point3d.createZero();
      if (twoConnectedArcs || arcLineStringCorner) {
        child.startPoint(linePoint);
        newFilletedLineString.tryAddChild(LineSegment3d.create(linePoint, linePoint));
      }
      if (child instanceof Arc3d && Math.abs(child.sweep.sweepDegrees) >= 180) {
        const firstHalf = child.clonePartialCurve(0, 0.5);
        const secondHalf = child.clonePartialCurve(0.5, 1);
        child.fractionToPoint(0.5, linePoint);
        newFilletedLineString.tryAddChild(firstHalf);
        newFilletedLineString.tryAddChild(LineSegment3d.create(linePoint, linePoint));
        newFilletedLineString.tryAddChild(secondHalf);
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+    // Algorithm:
+    // Each arc contributes a point with the arc's radius. If arc consumed the entire edge (2 points), we make sure to
+    // add both points (one with the arc's radius and one with zero radius).
+    // Each line segment contributes its start point with zero radius, except when it follows an arc, in which case
+    // it is ignored since the arc already contributes that point with the correct radius.
+    // For open validatedFilletedLineString (which is guaranteed to start and end with a line segment) we also add
+    // start and end points with zero radius.
+    const result: Array<[Point3d, number]> = [];
+    const numOfChildren = validatedFilletedLineString.children.length;
+    let ignoreLineSegment = false; // flag to ignore line segment that follows an arc
```
dassaf4:
```suggestion
    const lastChild = validatedFilletedLineString.cylicCurvePrimitive(-1);
    let ignoreLineSegment = isClosed && lastChild && lastChild instanceOf Arc3d;
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+      } else if (child instanceof LineSegment3d) {
+        // path is closed and starts with a line segment that follows an arc; ignore it
+        if ((isClosed && i === 0 && validatedFilletedLineString.children[numOfChildren - 1] instanceof Arc3d))
+          continue;
+        if (ignoreLineSegment) { // ignore line segment that follows an arc
+          ignoreLineSegment = false;
+          continue;
+        }
+        if (!ignoreLineSegment || (!isClosed && i === 0)) // add start point of open path
+          result.push([child.startPoint(), 0]);
```
dassaf4: Your last `if` statement in the LineSegment3d scope was always true. Try this simplification (together with the previous change to initialize `ignoreLineSegment`).

```suggestion
        if (ignoreLineSegment)
          ignoreLineSegment = false;
        else
          result.push([child.startPoint(), 0]);
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+      } else if (child instanceof LineString3d) {
+        for (let j = 0; j < child.numPoints() - 1; j++) {
+          if (ignoreLineSegment) // ignore line segment that follows an arc
+            ignoreLineSegment = false;
+          else
+            result.push([child.pointAtUnchecked(j), 0]);
+        }
```
dassaf4:
```suggestion
        const j0 = ignoreLineSegment ? 1 : 0;
        ignoreLineSegment = false;
        for (let j = j0; j < child.numPoints() - 1; j++)
          result.push([child.pointAtUnchecked(j), 0]);
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+      }
+    }
+    if (!isClosed) // add end point of open path
+      result.push([validatedFilletedLineString.children[numOfChildren - 1].endPoint(), 0]);
```
dassaf4:
```suggestion
    if (!isClosed) {
      const endPoint = validatedFilletedLineString.endPoint();
      if (endPoint)
        result.push([endPoint, 0]);
    }
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+    const child = validatedFilletedLineString.children[i];
+    if (!(child instanceof Arc3d))
+      return;
+    const prevChild = validatedFilletedLineString.children[(i - 1 + numOfChildren) % numOfChildren];
+    const prevPrevChild = validatedFilletedLineString.children[(i - 2 + numOfChildren) % numOfChildren];
+    const childStartTangent = child.fractionToPointAndDerivative(0).direction;
+    const prevPrevChildEndTangent = prevPrevChild.fractionToPointAndDerivative(1).direction;
+    const arcsAreParallel = childStartTangent.isParallelTo(prevPrevChildEndTangent, false, true, perpOptions);
+    if (prevChild instanceof LineSegment3d && prevChild.curveLength() === 0 && prevPrevChild instanceof Arc3d && !arcsAreParallel)
+      result.push([prevChild.startPoint(), 0]);
```
dassaf4: 1. What happens when an arc whose sweep is greater than 180 has been split by a ZLL into two "parallel" arcs? Your code will _not_ add the ZLL vertex between the arcs, which is incorrect, because the returned data when fed into `CurveFactory.createFilletsInLineString` will produce a chain with a different shape. I believe we want to add the ZLL vertex even if the arcs are parallel (see below).
2. Write a test that fails on the current code and succeeds on your fix.

```suggestion
    const child = validatedFilletedLineString.getChild(i);
    if (!child || !(child instanceof Arc3d))
      return;
    const prevChild = validatedFilletedLineString.cyclicCurvePrimitive(i - 1);
    if (!prevChild || !(prevChild instanceof LineSegment3d) || prevChild.curveLength() > 0)
      return;
    const prevPrevChild = validatedFilletedLineString.cyclicCurvePrimitive(i - 2);
    if (!prevPrevChild || !(prevPrevChild instanceof Arc3d))
      return;
    result.push([prevChild.startPoint(), 0]);
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,222 @@ export class CurveFactory {
...
+      const prevChildEndTangent = prevChild.fractionToPointAndDerivative(1).direction;
+      const childrenAreParallel = childStartTangent.isParallelTo(prevChildEndTangent, false, true, parallelOptions);
+      const twoNonParallelArcLsPair = !childrenAreParallel &&
+        ((child instanceof Arc3d && (prevChild instanceof LineSegment3d || prevChild instanceof LineString3d)) ||
+          ((prevChild instanceof Arc3d && (child instanceof LineSegment3d || child instanceof LineString3d))));
+      const twoConnectedArcs = child instanceof Arc3d && prevChild instanceof Arc3d;
+      // apply relaxed validation rules
+      if (twoConnectedArcs || twoNonParallelArcLsPair)
+        newFilletedLineString.tryAddChild(LineSegment3d.create(child.startPoint(), child.startPoint()));
+      if (child instanceof Arc3d && Math.abs(child.sweep.sweepDegrees) >= 180) {
```
dassaf4: 1. I would split the arc even if it's slightly less than 180, e.g., `Math.abs(child.sweep.sweepDegrees) > 180 - Geometry.smallAngleDegrees`.
2. Here's an interesting edge case that you don't currently handle. What if `arc.sweep.isFullCircle` returns `true`? Two ways to handle this case under `relaxedValidation` are to split the arc twice (three 120-degree arcs) or thrice (four 90-degree arcs).
3. Write another test on an input Path/Loop with a single full-sweep circular arc. Verify that the output data represent an equilateral triangle or a square (depending on if you split twice or thrice) with fillets at each corner.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,233 @@ export class CurveFactory {
...
+      filletedLineString = Path.create(LineSegment3d.create(startPoint, startPoint), ...filletedLineString.children);
+    }
+    if (lastChild instanceof Arc3d) {
+      const endPoint = lastChild.endPoint();
+      filletedLineString = Path.create(...filletedLineString.children, LineSegment3d.create(endPoint, endPoint));
+    }
+    return filletedLineString;
+  }
+  /** Split `arc` at each interior fraction and append the pieces with zero-length line segments at split points. */
+  private static splitArc(output: Path, arc: Arc3d, fractions: number[]): void {
```
dassaf4: Rename and clarify.

```suggestion
  /**
   * Split `arc` according to the partition given by `fractions` and append the pieces to `output` with zero-length line segments in between.
   * @param fractions a complete partition of the fractional parameter space, e.g., [0, 0.5, 1] splits the arc into two pieces. 
   */
  private static splitAndAppendArc(output: Path, arc: Arc3d, fractions: number[]): void {
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,233 @@ export class CurveFactory {
...
+    }
+  }
+  /**
+   * Update the path for relaxed validation:
+   * * If there are 2 connected arcs, add a zero-length line segment between them.
+   * * If there is a pair of arc and line segment/string with non-parallel tangents, add a zero-length line segment
+   * between them.
+   * * If there is an arc with sweep between 180 and 360 degrees, break the arc into 2 pieces and add a zero-length
+   * line segment between them. Break 360 degrees arc to 3 pieces. Return `undefined` if there is an arc with sweep
+   * greater than 360 degrees.
```
dassaf4:
```suggestion
   * * If there is an arc with sweep degrees in [180, 360), break the arc into 2 pieces separated by a zero-length
   * line segment. Similarly, break a 360-degree arc into 3 pieces separated by 2 zero-length line segments. Return
   * `undefined` if there is an arc with sweep greater than 360 degrees.
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,233 @@ export class CurveFactory {
...
+      const childStartTangent = child.fractionToPointAndDerivative(0).direction;
+      const prevChildEndTangent = prevChild.fractionToPointAndDerivative(1).direction;
+      const childrenAreParallel = childStartTangent.isParallelTo(prevChildEndTangent, false, true, parallelOptions);
+      const arcLineStringCorner = !childrenAreParallel &&
+        ((child instanceof Arc3d && (prevChild instanceof LineSegment3d || prevChild instanceof LineString3d)) ||
+          ((prevChild instanceof Arc3d && (child instanceof LineSegment3d || child instanceof LineString3d))));
+      const twoConnectedArcs = child instanceof Arc3d && prevChild instanceof Arc3d;
+      // apply relaxed validation rules
+      if (twoConnectedArcs || arcLineStringCorner) {
+        newFilletedLineString.tryAddChild(LineSegment3d.create(child.startPoint(), child.startPoint()));
```
dassaf4: Save an allocation...

```suggestion
        const linePoint = child.startPoint();
        newFilletedLineString.tryAddChild(LineSegment3d.create(linePoint, linePoint));
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,233 @@ export class CurveFactory {
...
+      }
+      if (child instanceof Arc3d && Math.abs(child.sweep.sweepDegrees) > 180 - Geometry.smallAngleDegrees
+        && Math.abs(child.sweep.sweepDegrees) < 360) {
```
dassaf4: Be precise with strict and non-strict inequalities...

* [0, 180 - e]  --- no split
* **(180 - e, 360 - e] --- 1 split**
* (360 - e, 360 + e] --- 2 splits
* (360 + e, infinity) --- `undefined`

```suggestion
      if (arcSweep !== undefined && arcSweep > 180 - sweepTol && arcSweep <= 360 - sweepTol) {
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,233 @@ export class CurveFactory {
...
+        CurveFactory.splitArc(newFilletedLineString, child, [0, 0.5, 1]); // 2 pieces
+      } else if (child instanceof Arc3d && Math.abs(child.sweep.sweepDegrees) === 360) {
```
dassaf4: Be precise with strict and non-strict inequalities...

* [0, 180 - e]  --- no split
* (180 - e, 360 - e] --- 1 split
* **(360 - e, 360 + e] --- 2 splits**
* (360 + e, infinity) --- `undefined`

```suggestion
      } else if (arcSweep !== undefined && arcSweep > 360 - sweepTol && arcSweep <= 360 + sweepTol) {
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,233 @@ export class CurveFactory {
...
+   * greater than 360 degrees.
+   */
+  private static updatePathForRelaxedValidation(
+    filletedLineString: Path, isClosed: boolean, parallelOptions?: PerpParallelOptions,
+  ): Path | undefined {
+    const newFilletedLineString = new Path();
+    const numOfChildren = filletedLineString.children.length;
+    for (let i = 0; i < numOfChildren; i++) { // examine each child and its predecessor
+      const child = filletedLineString.children[i];
+      if (child instanceof Arc3d && Math.abs(child.sweep.sweepDegrees) > 360)
```
dassaf4: Be precise with strict and non-strict inequalities... 
 
* [0, 180 - e]  --- no split
* (180 - e, 360 - e] --- 1 split
* (360 - e, 360 + e] --- 2 splits
* **(360 + e, infinity) --- `undefined`**

```suggestion
      const arcSweep = child instanceof Arc3d ? Math.abs(child.sweep.sweepDegrees) : undefined;
      const sweepTol = Geometry.smallAngleDegrees;
      if (arcSweep !== undefined && arcSweep > 360 + sweepTol)
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,233 @@ export class CurveFactory {
...
+    let validatedFilletedLineString = filletedLineString;
+    if (!isClosed)
+      validatedFilletedLineString = this.validateOpenPathStartEnd(validatedFilletedLineString);
+    if (relaxedValidation) {
+      const ret = this.updatePathForRelaxedValidation(validatedFilletedLineString, isClosed, options?.parallelOptions);
+      if (ret === undefined)
+        return undefined;
+      validatedFilletedLineString = ret;
+    }
+    const numOfChildren = validatedFilletedLineString.children.length;
```
dassaf4, line 381:
```suggestion
    const relaxedValidation = options?.relaxedValidation ?? false;
    let validatedFilletedLineString: Path | undefined = filletedLineString;
    if (!isClosed)
      validatedFilletedLineString = this.validateOpenPathStartEnd(validatedFilletedLineString);
    if (relaxedValidation)
      validatedFilletedLineString = this.updatePathForRelaxedValidation(validatedFilletedLineString, isClosed, options?.parallelOptions);
    if (!validatedFilletedLineString)
      return undefined;
    const numOfChildren = validatedFilletedLineString.children.length;
```

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,233 @@ export class CurveFactory {
...
+   *   * Each `Arc3d` is circular.
+   *   * Each `Arc3d` sweep is less than 180 degrees.
+   *   * Each `Arc3d` cannot be adjacent to another `Arc3d`.
+   *   * Each `Arc3d` is G1 continuous with each of its neighbors, i.e., at their common point, the curves have the same
+   * tangent direction.
+   * * To treat more input chains as valid, pass `options.relaxedValidation = true`. Internally, this setting performs
+   * several transformations on the input to produce a valid filleted linestring:
+   *   * Each `Arc3d` whose sweep is between 180 and 360 degrees is split into 2 arcs of equal sweep separated by a
+   * zero-length `LineSegment3d`. A 360 degrees arc is split into 3 arcs of equal sweep with 2 zero-length
+   * `LineSegment3d`s. Arcs with sweep greater than 360 degrees are not allowed.
```
dassaf4:
```suggestion
   * zero-length `LineSegment3d`. A 360-degree arc is split into 3 arcs of equal sweep separated by 2 zero-length
   * `LineSegment3d`s. Arcs with sweep greater than 360 degrees are not allowed.
```

core/geometry/src/test/curve/CurveFactory.test.ts
dassaf4, line 189: In general:
* Prefer shorter, precisely targeted tests. These are much easier to understand, debug, and extend. For example, I just wanted to run the "case 6: full circle" logic and see only its output.
* Split a long test into a `describe` block of multiple short `it` blocks.
* Break out common logic into helper functions to share across `it` blocks.

No action item for this PR, but please remember these tips for next time.

core/geometry/src/test/curve/CurveFactory.test.ts
```diff
@@ -327,6 +324,1138 @@ describe("CurveFactory", () => {
...
+    chain = Path.create(Arc3d.createXY(Point3d.create(0, 0), 3, AngleSweep.createStartEndDegrees(0, 360)));
+    GeometryCoreTestIO.captureCloneGeometry(allGeometry, chain, x0, y0);
+    pointsAndRadii = CurveFactory.fromFilletedLineString(chain);
+    ck.testUndefined(
+      pointsAndRadii,
+      "expect fromFilletedLineString to return undefined for 360 degree sweeps case 6"
+    );
+    // call with relaxed validation
+    childCountBefore = chain.children.length;
+    pointsAndRadii = CurveFactory.fromFilletedLineString(chain, { relaxedValidation: true });
```
dassaf4, line 1201: I expect the resulting linestring to be closed.  It is not.

For each Loop and physically closed Path test, please verify that `CurveFactory.fromFilletedLineString` returns an array whose first and last entries are equal.

core/geometry/src/test/curve/CurveFactory.test.ts
```diff
@@ -327,6 +324,1138 @@ describe("CurveFactory", () => {
...
+      Point3d.create(4, 0, 0),
+      Point3d.create(4, 4, 0),
+      Point3d.create(0, 4, 0),
+    );
+    const path = CurveFactory.createFilletsInLineString(lineString, radius, { allowCusp: false, filletClosure: true })!;
+    GeometryCoreTestIO.captureCloneGeometry(allGeometry, path, x0, y0);
+    const loop = Loop.create(...path.children);
+
+    pointsAndRadii = CurveFactory.fromFilletedLineString(loop);
+    ck.testDefined(pointsAndRadii, "expect fromFilletedLineString to work on Path");
```
dassaf4:
```suggestion
    ck.testDefined(pointsAndRadii, "expect fromFilletedLineString to work on Loop");
```

core/geometry/src/test/curve/CurveFactory.test.ts
saeeedtorabi: Yes, I personally prefer smaller unit tests. However, depending on the tests cases, there could be 2 problems with that. Problem 1 is probably ok but problem 2 is not.

1- Many times (not for this PR), I create multiple geoms (like curve primitives) and then test different combinations. If I want to use small unit tests I have to duplicate those geoms leading to larger test files.

2- I try to do non-visual tests but many times I have to also do visual tests (sometime only visual tests are possible or non-visuals dont test many things). When making src code changes, I prefer to see the visuals for all test cases at once. If I have many small unit tests, when I make a src code change, I have to open each visual separately to confirm I did not break anything. Having all visuals under one imjs file is much easier (unless there is a way to output visuals of all small unit tests into a shared imjs file).

core/geometry/src/test/curve/CurveFactory.test.ts
saeeedtorabi: That's not how I designed the algorithm. Currently I dont put duplicate entry to the end of array if path is closed. I assume I should. I will modify the src code and tests.

core/geometry/src/test/curve/CurveFactory.test.ts
dassaf4: 1. Put the common geometry creation code in a helper function. Or if it's widely useful, add it to GeometrySamples.ts or ImportedSamples.ts. 
2. Ask copilot to concatenate the output files into one file you can import. Tell it to organize them spatially so they don't overlap. Ask it to write a script/skill to do this.

core/geometry/src/curve/CurveFactory.ts
```diff
@@ -232,6 +254,240 @@ export class CurveFactory {
...
+          result.push([child.pointAtUnchecked(j), 0]);
+      }
+    }
+    if (!isClosed) {
+      const endPoint = validatedFilletedLineString.endPoint();
+      if (endPoint)
+        result.push([endPoint, 0]);
+    }
+    if (isClosed)
+      result.push(result[0]);
```
dassaf4: 1. For clarity, combine both if statements into an `if..else` statement.  Put the `isClosed` logic first, then the `else` clause has the `!isClosed` logic.
2. It costs little to be paranoid here: in the `isClosed` case, you should not push the first entry if the result array is empty, or if the last entry already equals the first entry.
3. Maybe you can come up with an input Path that results in output with duplicate last two entries because of the current `isClosed` logic. If you can, turn it into a test that verifies your fix.

saeeedtorabi: ok I modify it to 
```ts
    if (isClosed) {
      if (result.length > 0)
        result.push(result[0]);
    } else {
      const endPoint = validatedFilletedLineString.endPoint();
      if (endPoint)
        result.push([endPoint, 0]);
    }
```
Last 2 entries of results cannot be equal because:

1- For open we dont push `result[0]` as the last entry.
2- For closed 
- (a) if last child is line segment, it is ignored.
- (b) if last child is line string, its last line segment is ignored.
- (c) if last child is arc, there is no way that it can produce duplicate point/radius pairs.

dassaf4: 2. A future dev won't easily know this. I would still add a sanity check: for closed results with at least two entries, append the first entry if and only if the last entry is not exactly equal to the first. This is basically a free check, and a future-proof that your assumption won't be violated by future changes. An assert would also suffice, but we don't always remember to turn those on during testing (as we found out with the spiral close approach PR).

saeeedtorabi: ok I changed it to

```ts
    if (isClosed) {
      const [firstPoint] = result[0] ?? [];
      const [lastPoint] = result[result.length - 1] ?? [];
      if (firstPoint !== undefined && !firstPoint.isAlmostEqual(lastPoint, options?.distanceTol))
        result.push(result[0]);
    }
```

I only need to compare first and last points. No need to check first and last radii as they could be equal but related to 2 different points. Also I dont expect to have first and last points equal but with different radii. If that happens we wont push and the last radii will be ignored if data is fed to `createFilletsInLineString`

dassaf4: You lost the empty array guard and `firstPoint` is never undefined.

```
  if (isClosed) {
      if (result.length > 0 && !result[0][0].isAlmostEqual(result[result.length - 1][0], options?.distanceTol))
        result.push(result[0]);
      }
    }
```

saeeedtorabi: `firstPoint !== undefined` implies `result.length > 0` because if results is empty then `const [firstPoint] = [];` and therefore, `firstPoint === undefined` (notice the brackets around `firstPoint`) .

You can run below code and see it:
```ts
    const result: Array<[Point3d, number]> = [];
    const [firstPoint] = result[0] ?? [];
    if (firstPoint === undefined)
      console.log("firstPoint is undefined and result is empty");
```

However, I see that could be confusing. I thought using variable names for first and last points makes it less confusing.  I change it to
```ts
  if (isClosed) {
      if (result.length > 0 && !result[0][0].isAlmostEqual(result[result.length - 1][0], options?.distanceTol))
        result.push(result[0]);
      }
    }
```

PR #9002, opened 2026-02-19
Allow Passing Error Objects to Logger.logError()

MichaelSwigerAtBentley:
Adds an overload to Logger.logError() that takes an object as a parameter in place of a message. This allows us to pass in errors directly, and log them as LoggerMetadata. This matches the current functionality as logException(), but unifies it under logError.

This also flags logException as deprecated in favor of logError. This should simplify and alleviate confusion around how to log errors in iTwin. 

Fixes: https://github.com/iTwin/itwinjs-backlog/issues/1867

### DIFF
--- a/core/bentley/src/Logger.ts
+++ b/core/bentley/src/Logger.ts
@@ -315,8 +315,6 @@ export class Logger {
     const exceptionType = error?.constructor?.name ?? "<Unknown>";
     if (metaData === undefined) {
       return () => ({ exceptionType, ...error });
-    } else if (typeof metaData === "function") {
-      return () => ({ exceptionType, ...error, ...metaData() });
     }
     return () => ({ exceptionType, ...error, ...BentleyError.getMetaData(metaData) });
   }


core/bentley/src/test/Logger.test.ts
```diff
@@ -492,4 +493,43 @@ describe("Logger", () => {
...
+      (c, m, d) => outtrace = [c, m, BentleyError.getMetaData(d)]);
+    Logger.setLevel("testcat", LogLevel.Error);
+
+    clearOutlets();
+    try {
+      throw new Error("error message");
+    } catch (err: any) {
+      Logger.logError("testcat", err);
+    }
+    checkOutlets(["testcat", "Error: error message", { ExceptionType: "Error" }], [], [], []);
```
MichaelSwigerAtBentley: Upon closer inspection of these tests, I've realized that this checkOutlets function is not working correctly. It appears that even the old logException was not logging as expected. Seeing if I can work this stuff out in this PR now

MichaelSwigerAtBentley: I've fixed the tests and added better ways to handle metadata

core/bentley/src/test/Logger.test.ts
```diff
@@ -492,4 +493,43 @@ describe("Logger", () => {
...
+    expect(outerr[2]).deep.equal({ ...e2 }); // message and stack should be stripped off
+
+    clearOutlets();
+    expect(() => Logger.logError("testcat", undefined)).to.not.throw("undefined exception");
+    checkOutlets(["testcat", "Error: err is undefined.", { ExceptionType: "Error" }], [], [], []);
+
+    clearOutlets();
+    expect(() => Logger.logError("testcat", null)).to.not.throw("null exception");
+    checkOutlets(["testcat", "Error: err is null.", { ExceptionType: "Error" }], [], [], []);
+  });
```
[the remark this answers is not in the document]
MichaelSwigerAtBentley: Perhaps take a look at the very next test, titled: "should log errors with metadata when passed into logError"

core/backend/src/internal/IntegrityCheck.ts
```diff
@@ -356,6 +356,6 @@ export async function performSpecificIntegrityCheck(iModel: IModelDb, check: Int
       return results;
     }
     default:
-      throw new IModelError(IModelStatus.BadRequest, `Unknown integrity check type: ${check}`);
+      throw new IModelError(IModelStatus.BadRequest, `Unknown integrity check type`);
```
ben-polinsky, line 359: Did you mean to drop the `${check}` here?

core/bentley/src/Logger.ts
```diff
@@ -259,13 +259,35 @@ export class Logger {
...
-      Logger._logError(category, message, metaData);
+  public static logError(category: string, message: string, metaData?: LoggingMetaData): void;
+  /**
+   * Log the specified error to the **error** stream.
+   * @param category The category of the message
+   * @param error Error to log
+   * @param metaData Optional data for the message
+   * @note For legacy [[BentleyError]] exceptions, the special "exceptionType" property will be added as metadata. Otherwise, all enumerable members of the exception are logged as metadata.
+   */
+  public static logError(category: string, error: any, metaData?: LoggingMetaData): void;
```
ben-polinsky: Usually `unknown` is preferred over `any`. It's a bit safer, but not sure we get much here from it. Your choice.  

I do think the branch nesting could be cleaned up a bit.

core/bentley/src/Logger.ts
```diff
@@ -279,11 +301,28 @@ export class Logger {
     return BentleyError.getErrorMessage(err) + stack;
   }
 
+  /**
+   * Merged passed metaData with error properties into one LoggingMetaData, with the passed metaData taking precedence in case of conflict.
+   * @param error The error to be logged as metadata
+   * @param metaData Optional metadata to be merged with the error
+   * @returns A function returning the merged metadata
+   */
+  private static getExceptionMetaData(error: any, metaData?: LoggingMetaData): LoggingMetaData {
```
ben-polinsky, line 314: This error could also be `unknown` but you'd have to check before spreading below.

core/backend/src/internal/IntegrityCheck.ts
MichaelSwigerAtBentley: Yes, due to strong typing + switch, lint detects that this check's type is technically "never" and throws a lint warning.

core/bentley/src/Logger.ts
MichaelSwigerAtBentley: Yeah, I thought about using unknown, but I ended up sticking to any since logException used any. Swapping to unknown should be fine however

core/backend/src/internal/IntegrityCheck.ts
ben-polinsky: I'm assuming it's not that important then?

ben-polinsky:
A `NextVersion.md` note might be helpful as well.

core/backend/src/internal/IntegrityCheck.ts
MichaelSwigerAtBentley: Yeah, as lint is pointing out, its actually impossible for this default to ever be hit. But I opted to keep it around just for clarity. Removing the ${check} gets rid of the lint warning

core/bentley/src/Logger.ts
MichaelSwigerAtBentley: Yes, I'd end up just casting as an object at some point, and that doesn't feel right either

core/bentley/src/Logger.ts
MichaelSwigerAtBentley: Swapped to unknown + cleaned up the if statements a bit

MichaelSwigerAtBentley:
> A `NextVersion.md` note might be helpful as well.

Added

core/bentley/src/Logger.ts
```diff
@@ -259,31 +259,74 @@ export class Logger {
...
+   * @note For legacy [[BentleyError]] exceptions, the special "exceptionType" property will be added as metadata. Otherwise, all enumerable members of the exception are logged as metadata.
+   */
+  public static logError(category: string, error: unknown, metaData?: LoggingMetaData): void;
+  // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
+  public static logError(category: string, messageOrError: string | unknown, metaData?: LoggingMetaData) {
+    if (Logger._logError && Logger.isEnabled(category, LogLevel.Error)) {
+      if (typeof messageOrError === "string") {
+        Logger._logError(category, messageOrError, metaData);
+      } else if (BentleyError.isError(messageOrError)) {
+        // For backwards compatibility, log BentleyError old way
```
tcobbs-bentley, line 281: Did logError allow BentleyError objects before? It doesn't look like that is the case. Is this because you updated the `logException` code to instead use `logError`?

core/bentley/src/Logger.ts
```diff
@@ -259,31 +259,74 @@ export class Logger {
...
+   * @param error The error to be logged as metadata
+   * @param metaData Optional metadata to be merged with the error
+   * @returns A function returning the merged metadata
+   */
+  private static getExceptionMetaData(error: any, metaData?: LoggingMetaData): LoggingMetaData {
+    const exceptionType = error?.constructor?.name ?? "<Unknown>";
+    if (metaData === undefined) {
+      return () => ({ exceptionType, ...error });
+    } else if (typeof metaData === "function") {
+      return () => ({ exceptionType, ...error, ...metaData() });
```
tcobbs-bentley: Since `BentleyError.getMetaData` already does this, I don't think this `else if` clause is needed.

tcobbs-bentley: See [here](https://github.com/iTwin/itwinjs-core/blob/9b9d90ef1752d39cc49dc6923c4231ac90ec0c84/core/bentley/src/BentleyError.ts#L451-L453).

bbastings (approved):

core/bentley/src/Logger.ts
MichaelSwigerAtBentley: No, this change deprecates logException and adds a new overload to logError that covers logExceptions functionality. (previously logError only took a message and metadata and logException only logged an error) Now, logError can log either a message and metaData or an error and metaData.

core/bentley/src/Logger.ts
MichaelSwigerAtBentley: Good catch

MichaelSwigerAtBentley: fixed

PR #9009, opened 2026-02-20
Add dta-repro agent

ben-polinsky:
Add agent to assist in minimal reproductions of bugs in our display-test-app viewer. For those unfamiliar with the DTA viewer, it can be a bit frustrating to figure out integration. This agent handles writing the repro based on information, github issue, or code and registers a key-in to trigger.

Initial testing was promising. I am wondering if a skill is a better fit.


ben-polinsky:
dupe #9312

PR #9011, opened 2026-02-20
Add DataTestId support to PropertyEditorParams

arome:
When multiple enum list are used in one tool, they all share the same data-testid. I'm adding an option to provide data-testid as an additional params so it can be passed to the editor

### DIFF
--- a/ui/appui-abstract/src/appui-abstract/properties/EditorParams.ts
+++ b/ui/appui-abstract/src/appui-abstract/properties/EditorParams.ts
@@ -70,6 +70,7 @@ export enum PropertyEditorParamTypes {
   // SuppressUnitLabel = "UiAbstract-SuppressUnitLabel",
   SuppressEditorLabel = "UiAbstract-SuppressEditorLabel",
   CheckBoxImages = "UiAbstract-CheckBoxImages",
+  DataTestId = "UiAbstract-DataTestId",
 }
 
 /**
@@ -78,6 +79,7 @@ export enum PropertyEditorParamTypes {
  */
 export interface BasePropertyEditorParams {
   type: string;
+  "data-testid"?: string;
 }
 
 /**
@@ -295,7 +297,6 @@ export interface ImageCheckBoxParams extends BasePropertyEditorParams {
   type: PropertyEditorParamTypes.CheckBoxImages;
   imageOn: string;
   imageOff: string;
-
 }
 
 // /**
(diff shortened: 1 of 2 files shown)


ben-polinsky:
Hey - where's this coming from and what's the need?

arome:
> Hey - where's this coming from and what's the need?

hey, I updated the description of the PR. It's because we have tools that have multiple properties and they all share the same data-testids when its a similar component. I'm creating this PR here to be able to provide different test-ids to target different components and will create another one in appui to use the data-testid the user will be passing

ben-polinsky:
I don't think this is the place for the `data-testId` - we don't provide any currently.  I assume that the current data-testIds you're referring to are set in AppUI? Could you set this there? 

`data-testId`s are also a last-resort for locating elements in tests - we don't want to enforce that at the abstract UI level.

arome:
> I don't think this isn't the place for the data-testId - we don't provide any currently. I assume that the current data-testIds you're referring to are set in AppUI? Could you set this there?

yes its set here in appui, https://github.com/iTwin/appui/blob/17e28f44bae899f9277e82814975ec61e3be4eb3/ui/components-react/src/components-react/editors/EnumEditor.tsx#L183
problem is thats a generic component, and it can appear multiple times in a page, so we shouldnt ideally have a fixed data-testid for it. 

The only way I saw to pass a testid was through this additional params, do you have a better approach?

ben-polinsky:
> yes its set here in appui, https://github.com/iTwin/appui/blob/17e28f44bae899f9277e82814975ec61e3be4eb3/ui/components-react/src/components-react/editors/EnumEditor.tsx#L183 problem is thats a generic component, and it can appear multiple times in a page, so we shouldnt ideally have a fixed data-testid for it.
> 
> The only way I saw to pass a testid was through this additional params, do you have a better approach?

Why do you need a data-testId in the first place? Is there no other way to locate the element?

arome:
> > yes its set here in appui, https://github.com/iTwin/appui/blob/17e28f44bae899f9277e82814975ec61e3be4eb3/ui/components-react/src/components-react/editors/EnumEditor.tsx#L183 problem is thats a generic component, and it can appear multiple times in a page, so we shouldnt ideally have a fixed data-testid for it.
> > The only way I saw to pass a testid was through this additional params, do you have a better approach?
> 
> Why do you need a data-testId in the first place? Is there no other way to locate the element?

It's a bug filed from our QA team, they need to be able to target specific components in ToolSettings for their tests. As you can see there are no way to differentiate between the two. The only approach i could see would be to target based on the child div position, which isnt ideal when you're writing tests.
image

ben-polinsky:
> and will create another one in appui to use the data-testid the user will be passing

Can't you reuse one of the other properties here to set the data-testid in appui? I'm not very keen on adding data-testId support in the library. 

>  The only approach i could see would be to target based on the child div position, 

What about the text inside of the span (closed in your screenshot)?

arome:
> > and will create another one in appui to use the data-testid the user will be passing
> 
> Can't you reuse one of the other properties here to set the data-testid in appui? I'm not very keen on adding data-testId support in the library.
> 
> > The only approach i could see would be to target based on the child div position,
> 
> What about the text inside of the span (closed in your screenshot)?

adding @chriscroucherBentley, since he's the one mainly concerned by this issue

PR #9012, opened 2026-02-22
fix oneAtATimeAction so it doens't start a new request until the old one finishes

kabentley:
Fixes a race in OneAtATimeAction where a replaced pending request could trigger execution of the next pending request before the active request completed.

Added new test that shows this behavior and verified that old implementation failed the new test


kabentley:
@hl662 Can you please take this over for me? I can't see the build logs and I don't know what is failing.

aruniverse (approved):

PR #9016, opened 2026-02-24
Decreased coordinate compare tolerance to something more adequate

AlainRobertAtBentley:
Reduced tolerance for coordinate compare operation for Coordinate Reference System from default (0.000001) to a more valid value of 0.001 which is plenty in terrain coordinates

### DIFF
--- a/common/changes/@itwin/core-backend/alainr-fix-gcs-test-tol_2026-02-24-15-48.json
+++ b/common/changes/@itwin/core-backend/alainr-fix-gcs-test-tol_2026-02-24-15-48.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Changed the tolerance for comparing coordinate in CRS tests",
+      "type": "none"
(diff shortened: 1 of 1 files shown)


ben-polinsky:
FYI: this helps us get Vertical CRS support through in https://github.com/iTwin/imodel-native/pull/1308

aruniverse:
> FYI: this helps us get Vertical CRS support through in [iTwin/imodel-native#1308](https://github.com/iTwin/imodel-native/pull/1308)

should this be a linked pr then?
cc @DanRod1999

DanRod1999:
I personally only recommend linking if the PRs are dependent of each other. Otherwise its just more work and longer wait times for no difference. If they are dependent I would link, however since this PR is passing without any native changes I don't know if it's necessary.

ben-polinsky:
> I personally only recommend linking if the PRs are dependent of each other. Otherwise its just more work and longer wait times for no difference. If they are dependent I would link, however since this PR is passing without any native changes I don't know if it's necessary.

This was the thought process.

Issue #8873, opened 2025-12-19
QueryBinder should not fail on bing a empty array of ids


khanaffan:
`QueryBinder`should fail on empty array of ids

Following should not throw error

```ts
QueryBinder.from([elementIds]),
```

reported by @a-gagnon

PR #9036, opened 2026-02-27
QueryBinder should not fail on bind empty array

naveedkhan8067:
Fixes: https://github.com/iTwin/itwinjs-core/issues/8873

### DIFF
--- a/core/backend/src/test/ecdb/QueryReaders.test.ts
+++ b/core/backend/src/test/ecdb/QueryReaders.test.ts
@@ -240,6 +240,14 @@ describe("QueryReaders - createQueryReader() and withQueryReader() api tests", (
       const reader = iModel.createQueryReader("SELECT ECInstanceId FROM meta.ECSchemaDef ORDER BY ECInstanceId ASC", undefined, { limit: { count: 5 } });
       await readerCallback(reader);
     });
+
+    it("Should not fail on empty array", async () => {
+      const idSet: Id64String[] = [];
+      const binder = QueryBinder.from([idSet]);
+      const reader = iModel.createQueryReader("SELECT ECInstanceId, ECClassId, Name from ecdbf.ExternalFileInfo WHERE InVirtualSet(?, ECInstanceId)", binder);
+      assert.isFalse(await reader.step());
+    });
+
   });
 
   describe("Common usages", () => {


core/common/src/ConcurrentQuery.ts
```diff
@@ -609,7 +609,7 @@ export class QueryBinder {
       params.bindPoint2d(nameOrId, val);
     } else if (val instanceof Point3d) {
       params.bindPoint3d(nameOrId, val);
-    } else if (val instanceof Array && val.length > 0 && typeof val[0] === "string" && Id64.isValidId64(val[0])) {
+    } else if (val instanceof Array && ((val.length > 0 && typeof val[0] === "string" && Id64.isValidId64(val[0])) || val.length == 0)) {
```
RohitPtnkr1996: Should we check all the elements instead of just the first one in the array when binding?
Something like
else if (Array.isArray(val) && (val.length === 0 || val.every(v => typeof v === "string" && Id64.isValidId64(v))))

naveedkhan8067: Good point. I have updated it.

naveedkhan8067: @RohitPtnkr1996, @khanaffan  do you know why the older expression was set to `else if (val instanceof Array && val.length > 0 && typeof val[0] === "string" && Id64.isValidId64(val[0]))` (only checking the validity of the first item and allowing others) because I see [tests](https://github.com/iTwin/itwinjs-core/blob/master/core/backend/src/test/ecdb/ECSqlQuery.test.ts) verifying this as intended behaviour:

```
  it("concurrent query bind idset with invalid values in IdSet virtual table", async () => {
    const ids: string[] = ["0x1", "ABC", "YZ"];

    const reader = imodel1.createQueryReader("SELECT * FROM BisCore.element, IdSet(?) WHERE id = ECInstanceId ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES", QueryBinder.from([ids]));
    let props = await reader.getMetaData();
    assert.equal(props.length, 12); // 11 for BisCore.element and 1 for IdSet
    let rows = 0; // backend will bind successfully but some of the values are not valid for IdSet VT so those values will be ignored
    while (await reader.step()) {
      rows++;
    }
    assert.equal(rows, 1);
    props = await reader.getMetaData();
    assert.equal(props.length, 12); // 11 for BisCore.element and 1 for IdSet
    assert.equal(reader.stats.backendRowsReturned, 1);
    assert.isTrue(reader.stats.backendCpuTime > 0);
  });
  it("concurrent query bind idset with invalid values in IdSet virtual table", async () => {
    const ids: string[] = ["ABC", "0x1", "YZ"]; // as first value is not an Id so QueryBinder.from will throw error of "unsupported type"

    try {
      imodel1.createQueryReader("SELECT * FROM BisCore.element, IdSet(?) WHERE id = ECInstanceId ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES", QueryBinder.from([ids]));
    } catch (err: any) {
      assert.equal(err.message, "unsupported type");
    }
  });
```

Now with the updated check, the first test case will also fail with the `unsupported type` error message. What should be the expected behaviour?

RohitPtnkr1996: This seems like a weirdly specific behavior that does not make sense to me.
The binding fails when the first value is invalid Id64, but all others are valid.
On the other hand, binding succeeds when only the first is valid by ignoring the subsequent invalid ones.

I can see if we want to ignore the invalid values and bind only the valid Ids, but not sure why we have it this way.

@soham-bentley

soham-bentley: Yes I can't remember any specific reason to have the check only on the first element of the list, although as the logic was already just checking first element, I added a test case to document that behavior. But I agree a more robust approach is checking all the elements of the list instead of the first one. I can't think of any specific scenario where checking only the first element in the list makes sense.

naveedkhan8067: Then these tests should be fixed. I'll do that.

PR #9037, opened 2026-02-27
Fix GHSA-7r86-cg39-jmmj

hl662:
Fixes GHSA-7r86-cg39-jmmj

Rush update --full couln't bump it, have to add new pnpm override entry, just targeting `api-extractor` - `test-exclude` also uses minimatch, but an older major version that is not affected by the cve.


hl662:
@mergifyio backport release/5.7.x

Issue #8749, opened 2025-11-10
HubMock doesn't handle locks properly


rschili:
From our documentation:

> You may only obtain the Exclusive lock on an Element if your BriefcaseDb.changeset.index is equal or greater than the ChangesetIndex
specified the last time the lock was released. That is, you may only 
acquire the Exclusive lock on an Element if your briefcase holds its 
most recent state.

https://www.itwinjs.org/learning/backend/concurrencycontrol/

HubMock does not do that. Instead, you can just always obtain a lock unless it is currently held by somebody else.

This leads to data corruption in tests that cover this area.

Pasting a failing test from a different PR here which demonstrates the problem:

```ts
  it("Pending incoming element update with local data transform change", async () => {
    // Setup iModel
    await setupIModel(user1Token);
    const user1Briefcase = await openBriefcase(user1Token);
    briefcases.push(user1Briefcase);

    // Import base schema
    await user1Briefcase.importSchemaStrings([testSchemas.baseSchema]);
    user1Briefcase.saveChanges();
    assert.isFalse(user1Briefcase.holdsSchemaLock);
    assert.isTrue(user1Briefcase.holdsSchemaTableLock);
    await user1Briefcase.pushChanges({ description: "import initial schema", accessToken: user1Token });

    // Insert element using base schema
    await user1Briefcase.locks.acquireLocks({ shared: IModel.dictionaryId });
    const [modelId, categoryId] = insertModelAndCategory(user1Briefcase);
    const elementId = insertTestElement(user1Briefcase, modelId, categoryId, "INITIAL_A", "INITIAL_C");

    // Insert a second element which will not be modified later, so we can verify data transformation worked correctly
    const secondElementId = insertTestElement(user1Briefcase, modelId, categoryId, "SECOND_ELEM_A", "SECOND_ELEM_C");

    user1Briefcase.saveChanges();
    await user1Briefcase.pushChanges({ description: "insert element", accessToken: user1Token });
    assert.isFalse(user1Briefcase.holdsSchemaLock);
    assert.isFalse(user1Briefcase.holdsSchemaTableLock);
    const user2Briefcase = await openBriefcase(user2Token);
    briefcases.push(user2Briefcase);

    // User1 modifies element locally
    await user1Briefcase.locks.acquireLocks({ exclusive: elementId });
    updateElementProperty(user1Briefcase, elementId, "propA", "UPDATED_A");
[...]

PR #376, opened 2020-12-07
Rework and simplify BriefcaseManager and BriefcaseDb

kabentley:
Note this PR contains breaking changes to the @beta api BriefcaseManager.download and @internal api NativeApp

PR #9045, opened 2026-03-02
Add the ability to release the locks associated with reversed Txns

### LINKED ISSUE #8552: Automatically release locks when undoing Txns
Locks are held unnecessarily when reversing a Txn. Once you make an edit that requires a lock, the only way the framework will automatically release them is to:
1. Push changes
1. Undo ALL changes to a briefcase since last sync via `discardChanges`
1. Release the briefcase

I would like functions that undo changes to release locks that were acquired during those Txns. For example, `TxnManager.reverseTxns` and `TxnManager.cancelTo` should release the locks acquired from those Txns. (perhaps as an optional argument/setting)

Manually keeping track of and diffing which locks were created during a specific Txn so that they can manually be reversed on undo.

kring:
Fixes #8552 by making `ServerBasedLocks` Txn-aware, so locks from reversed or canceled Txns can be safely abandoned and later reacquired for redo, without weakening existing lock guarantees.

This PR:

* Tracks lock acquisition per Txn in the `ServerBasedLocks` lock cache, including upgrades of Shared -> Exclusive.
* Adds async versions of `TxnManager` functions for Txn reverse and cancel that automatically perform lock abandon, and functions for Txn reinstate that automatically perform lock reacquisition.
* Makes lock abandonment on reverse optional, but, if locks are abandoned, then they _must_ be reacquired in order to reinstate.
* Keeps existing synchronous `TxnManager` functions intact and does not change their semantics, to preserve backward compatibility.
* Ensures even the old synchronous `reinstateTxn` fails with LockNotHeld when required locks were previously abandoned, so lock semantics cannot be violated. This does not contradict the point above, and is not a breaking change, because clients using the synchronous reverse/cancel functions will not abandon the locks in the first place.
* Clears stale Txn-lock records when Txns become unreachable (canceled), preventing incorrect behavior when Txn IDs are reused.
* Switches discard-changes lock release semantics to abandon mode so lock release does not stamp a changeset when nothing was actually pushed, avoiding unnecessary pull requirements for other users.
* Introduces optional hub access abandon APIs with compatibility fallbacks to existing lock update calls, so older hub implementations continue to work.

Backward compatibility:

* Existing synchronous Txn APIs remain available and behavior-compatible unless apps opt into the new async lock-aware flows.
* New `BackendHubAccess` methods are optional and `ServerBasedLocks` knows how to use fallbacks when they're not available.
* `LockControl` is not available for external implementation, so new methods there do not represent breaking changes.

Lock-correctness guarantees:

* Locks are only abandoned after 1) unsaved changes have been abandoned, and 2) the Txn in which the locks were acquired has been reversed.
* Reinstating a reversed Txn requires the previously-acquired locks to be held. Users can't accidentally (or purposely) reinstate changes without first accquiring the necessary locks.
* In the (unlikely?) event lock abandonment fails while canceling a Txn, the transaction remains reinstateable rather than being irreversibly canceled. This allows the abandonment to be retried, or alternatively the client can proceed to cancel without abandoning the locks.

### DIFF
--- a/core/backend/src/IpcHost.ts
+++ b/core/backend/src/IpcHost.ts
@@ -343,57 +343,27 @@ class IpcAppHandler extends IpcHandler implements IpcAppFunctions {
   }
 
   public async reverseTxns(key: string, numOperations: number): Promise<IModelStatus> {
-    const db = IModelDb.findByKey(key);
-    if (db.isBriefcaseDb()) {
-      return db.txns.reverseTxns(numOperations);
-    } else {
-      return db[_nativeDb].reverseTxns(numOperations);
-    }
+    return BriefcaseDb.findByKey(key).txns.reverseTxns(numOperations);
   }
 
   public async reverseTxnsAsync(key: string, numOperations: number, args?: ReverseTxnArgs): Promise<void> {
-    const db = IModelDb.findByKey(key);
-    if (!db.isBriefcaseDb()) {
-      throw new IModelError(IModelStatus.BadRequest, "reverseTxnsAsync is only supported for briefcase iModels");
-    }
-
-    return db.txns.reverseTxnsAsync(numOperations, args);
+    return BriefcaseDb.findByKey(key).txns.reverseTxnsAsync(numOperations, args);
   }
 
   public async reverseAllTxn(key: string): Promise<IModelStatus> {
-    const db = IModelDb.findByKey(key);
-    if (db.isBriefcaseDb()) {
-      return db.txns.reverseAll();
-    } else {
-      return db[_nativeDb].reverseAll();
-    }
+    return BriefcaseDb.findByKey(key).txns.reverseAll();
   }
 
   public async reverseAllTxnsAsync(key: string, args?: ReverseTxnArgs): Promise<void> {
-    const db = IModelDb.findByKey(key);
-    if (!db.isBriefcaseDb()) {
-      throw new IModelError(IModelStatus.BadRequest, "reverseAllTxnsAsync is only supported for briefcase iModels");
-    }
-
-    return db.txns.reverseAllTxnsAsync(args);
+    return BriefcaseDb.findByKey(key).txns.reverseAllTxnsAsync(args);
   }
 
   public async reinstateTxn(key: string): Promise<IModelStatus> {
-    const db = IModelDb.findByKey(key);
-    if (db.isBriefcaseDb()) {
-      return db.txns.reinstateTxn();
-    } else {
-      return db[_nativeDb].reinstateTxn();
-    }
+    return BriefcaseDb.findByKey(key).txns.reinstateTxn();
   }
 
   public async reinstateTxnAsync(key: string, args?: ReinstateTxnArgs): Promise<void> {
-    const db = IModelDb.findByKey(key);
-    if (!db.isBriefcaseDb()) {
-      throw new IModelError(IModelStatus.BadRequest, "reinstateTxnAsync is only supported for briefcase iModels");
-    }
-
-    return db.txns.reinstateTxnAsync(args);
+    return BriefcaseDb.findByKey(key).txns.reinstateTxnAsync(args);
   }
 
   public async restartTxnSession(key: string): Promise<void> {


core/backend/src/internal/ServerBasedLocks.ts
```diff
@@ -200,12 +238,155 @@ export class ServerBasedLocks implements LockControl {
     return this.acquireAllLocks(locks);
   }
 
+  private async releaseLocksAfterAbandon(locks: LockMap): Promise<void> {
+    if (IModelHost[_hubAccess].releaseLocksAfterAbandon === undefined) {
+      // If the IModelHub doesn't support an explicit abandon, call release with a null changeset.
+      await IModelHost[_hubAccess].acquireLocks({
```
rschili, line 289: This looks sus. Calling acquireLocks with a null changeset? What is the expected outcome of this call?

core/backend/src/BackendHubAccess.ts
```diff
@@ -224,12 +224,42 @@ export interface BackendHubAccess {
...
   /** Release all currently held locks */
   releaseAllLocks: (arg: BriefcaseDbArg) => Promise<void>;
 
+  /**
+   * Releases all currently held locks when none of the associated elements have been or will be modified.
+   * It is only valid to call this method when none of the elements protected by the locks have been edited,
+   * or if all edits have been reversed or abandoned without pushing them.
+   *
+   * The locks are released on the IModelHub, but the changeset associated with the locks is not updated.
+   * This is equivalent to calling {@link BackendHubAccess.releaseAllLocks} with an invalid changeset index.
```
GytisCepk: Why are we introducing more methods which functionality is already covered by other methods ?

If goal is to improve clarity, I think, it would be better to fix types and better document other methods (e.g. making changeset optional in `releaseAllLocks` and documenting it better what happens if it's undefined). Adding additional methods like these just bloats API and makes functionality harder to discover, in my opinion.

rschili:
Just linking this here for reference. HubMock has some existing limitations that may limit how we can test this, moving back and forth between txns/changesets. https://github.com/iTwin/itwinjs-core/issues/8749

core/backend/src/test/standalone/ServerBasedLocks.test.ts
```diff
@@ -328,4 +328,301 @@ describe("Server-based locks", () => {
       expectUnlocked();
     });
   });
-});
+
+  describe("abandonLocksForReversedTxn", () => {
```
khanaffan, line 334: Can you provide an example or test showing how a downstream application is expected to use this API? It would help to see different scenarios where these APIs would realistically be called.

Right now, from just looking at the API surface, these appear to be endpoints a user *can* call, but it’s unclear *when* they are supposed to call them. These APIs also don’t appear to enforce the requirement that a txn must be successfully reversed or deleted before its lock is released. Otherwise, we risk pushing a txn whose locks have been released without proper cleanup, which could cause errors when another briefcase later pulls and applies that txn (because two people might be change same element).

It feels like these APIs should be internal and only invoked as part of specific workflows, such as:

1.  Deleting reversed txn during pull
2.  Undoing/reversing a txn
3.  Redoing/reinstating a txn

core/backend/src/internal/ServerBasedLocks.ts
kring: Oh yes this is fun!

Ok so `acquireLocks` is more or less a thin wrapper over Hub's [Update Locks](https://developer.bentley.com/apis/imodels-v2/operations/update-imodel-locks) service. So, despite the name, it can be used to _both_ acquire and release locks.

Here's an explanation of how the changeset parameter to that Update Locks service is meant to work, copied from https://github.com/iTwin/imodels-api/issues/224#issuecomment-3965955415:

> ### How changesetId Works
>
> The changesetId property is optional when acquiring or releasing locks.
>
> When changesetId is NOT set:
>
> * Indicates the lock (Shared or Exclusive) was acquired and released without modifying the element
ReleasedWithChangeSetIndex will not be set in the database
> * Other users won't need to pull to modify the element
>
> When changesetId IS set:
> 
> * ReleasedWithChangeSetIndex will be set in the database
> * Other users must pull up to that changeset before modifying the element
> * This happens even if the element wasn't actually modified

Except I don't think this was widely understood. iTwin.js was pretty much _always_ specifying a valid changeset when releasing locks, even when the associated changes were discarded. This meant that even discarded changes could end up forcing another user to pull unnecessarily.

core/backend/src/BackendHubAccess.ts
kring: It's true, we could solve this with an additional parameter and documentation, though I think that would less usable. 

Based on the discussion in iTwin/imodels-api#224, my understanding is that these are the _valid_ lock release operations:

1. _release all locks after push_ - The user locked some elements, edited them, and pushed the changes. They now want to release the locks. It's essential that the changeset associated with each element be updated within Hub's lock database so that we enforce the rules about, for example, only acquiring an Exclusive lock on an element (and its model/parent subtree) when your briefcase has the latest version of those elements.
2. _release all locks after abandon_ - The user locked some elements, maybe edited them, and then abandoned their changes. They now want to release the locks, but they haven't actually edited anything. This time, it's important that the changeset associated with the element is _not_ updated, because if we updated it, other users would be forced to pull to edit these elements even though nothing actually changed. Before this PR, we were getting this wrong (see my comment above).
3. _release a subset of currently-held locks after edits to those particular elements are abandoned_ - This is new in this PR, to support releasing locks for a single abandoned Txn (not _all_ the user's Txns).

Prior to the changes to `BackendHubAccess`, (1) was implemented via `releaseAllLocks`. (2) was also implemented by calling `releaseAllLocks`, which was incorrect. And (3) could be implemented by calling `acquireLocks` in just the right way, but this was hacky.

So in this PR, I've made those three operations explicit as three functions. Backward compatibility required the new ones to be optional.

I think there's a reasonable argument to be made that (1) and (2) can be satisfied with a single `releaseAllLocks` method with a parameter indicating which mode you're in. Technically, we don't even need a new parameter, because the changeset parameter already provided to the function achieves the purpose _if people know how to use it_. But it's subtle and confusing and easy to get wrong.

If we add a new explicit "abandon" parameter to `releaseAllLocks`, this creates backward compatibility headaches. An old implementation of the interface will ignore the parameter and do the wrong thing, and we have no way of seeing from the outside that this will be the case. With the separate, optional `releaseAllLocksAfterAbandon` method, we can tell whether the implementation supports the new mode by the presence of the method.

Also, with (3), the only allowed mode is the "abandon" version. So I didn't like inconsistency of having a single `releaseAllLocks` method with an "abandon" parameter that defaults to false (for backward compatibility), plus a `releaseLocks` method that can _only_ be used after abandon. The explicit methods make this all much clearer and harder to misuse accidentally, IMO.

core/backend/src/test/standalone/ServerBasedLocks.test.ts
kring: My expectation is that immediately after calling `TxnManager.cancelTo`, an application will loop over the range of canceled Txns, from newest to oldest, and call `abandonLocksForReversedTxn` on each.

Or, similarly, the "undo session changes" operation can call `abandonLocksForReversedTxn` for each Txn in the session (newest to oldest once again).

Finally, as mentioned in the top post, if an application that can tolerate "redo" throwing an exception when it can't acquire a lock, then it's possible to call `abandonLocksForReversedTxn` immediately after `TxnManager.reverseSingleTxn`, and, when redoing, call `acquireLocksForReinstatingTxn` immediately before `TxnManager.reinstateTxn`.

My view here is probably incomplete though. I would definitely appreciate some feedback on it from @simihartstein  and others.

kring: I wrote my comment above before your edit. Some new thoughts based on what you added:

> These APIs also don’t appear to enforce the requirement that a txn must be successfully reversed or deleted 
before its lock is released.

`releaseLocksForReversedTxn` does require that either:
1. The txn is the current one, and it has no unsaved changes, or
2. The txn has been reversed.

I added that relatively late, though, so maybe you reviewed it before I did.

One thing that is _not_ enforced (yet is probably important) is that `releaseLocksForReversedTxn` / `acquireLocksForReinstatingTxn` are called in a stack-like fashion. If you have txn A, B, and C, saved in that order, you must release the locks for C, then B, then A. And when re-acquiring, you must acquire for A before B and B before C. You get the idea. Enforcing this would add some complexity, but I'm happy to do that if it seems valuable.

> It feels like these APIs should be internal and only invoked as part of specific workflows

Yeah, that makes sense. You could kind of make that argument about most of `LockControl`, right? 

Anyway, to do that here, we'll have to make some decisions. For example, do we release locks on undo? 😬

core/backend/src/internal/ServerBasedLocks.ts
rschili: Oh wow, this is horrible but powerful at the same time. I feel like we should document and wrap this so it doesn't feel as weird. This actually allows us to do things we didn't think were possible.
Maybe it's enough if we just put paste this documentation on acquireLocks().

core/backend/src/BackendHubAccess.ts
```diff
@@ -224,12 +224,42 @@ export interface BackendHubAccess {
...
+   * it may be returned to the {@link LockState.Shared} state or released entirely. It is only
+   * valid to call this method when none of the elements protected by the locks have been edited, or if all edits
+   * have been reversed or abandoned without pushing them.
+   *
+   * The locks are released on the IModelHub, but the changeset associated with the locks is not updated.
+   * This is equivalent to calling {@link BackendHubAccess.acquireLocks} with an invalid changeset index.
+   *
+   * It is an error to specify {@link LockState.Exclusive}. It is also an error to specify
+   * {@link LockState.Shared} for a lock that is not currently held exclusively, but an
+   * implementation may choose whether to throw an exception in this scenario or simply ignore it.
```
GytisCepk: I would argue, that whether and what errors are thrown should be part of API and not implementation detail. There is no point in abstraction if user has to know implementation details to handle errors.

I would use `@throws` tag to document thrown error or just document that it fails silently.

core/backend/src/BackendHubAccess.ts
GytisCepk: Maybe I'm being too pedantic, but all this sounds more like responsibility of `LockControl` implementation and not `BackendHubAccess`.

As you pointed out elsewhere, these methods are thin wrappers around iModels API so ideally it would remain just that and would provide just a simple methods to execute requests to iModels API while `LockControl` would be the one to actually do the logic of deciding what and how needs to be acquired or released.

core/backend/src/BackendHubAccess.ts
kring: Well, I agree with you in general. But the way I see this particular case is that specifying `LockState.Shared` for a lock that is not already held exclusively is improper use of the API. A violation of the API's preconditions. A developer error. So not something that should ever need to be _handled_.

The reason to leave the behavior unspecified is to leave room for the implementation to do whatever it thinks best in this case that should never happen in normal use. Perhaps the implementation of the "release" is just a SQL statement, but if we need to handle it explicitly in order to throw an exception, then it would require an extra query. Or, conversely, perhaps there would be a risk of inadvertently acquiring a Shared lock in the course of a "release", and the best way to defend against that is to explicitly check for it, and as long as we're checking the natural thing is to bail early with an exception.

So that's why I left this unspecified, while still calling attention to it so that implementors will think about it. I don't feel too strongly about it though, so I don't mind pinning it down if people think that would be better. Probably go the exception route in that case, because I think that's the more idiomatic approach in iTwin.js.

core/backend/src/BackendHubAccess.ts
kring: That's definitely possible! Among other things, it puts us in the weird situation of using a method on `BackendHubAccess` called `acquireLocks` to release locks, though. 😁 I guess it would have been nice if the method were called `updateLocks` and more explicitly matched the backend service API.

kabentley:
I don't understand why we don't just always abandon locks on undo.

kabentley:
>  This is mostly ok, except that cancelTo can also effectively obliterate txns and then I believe those txn IDs can be reused. That probably won't go well.

I think we can probably change `cancelTo` to not reuse TxnIds. Did you look into that?

core/backend/src/internal/ServerBasedLocks.ts
```diff
@@ -105,7 +111,26 @@ export class ServerBasedLocks implements LockControl {
     return this[_releaseAllLocks]();
   }
 
-  private insertLock(id: Id64String, state: LockState, origin: LockOrigin): true {
+  public async releaseAllLocksAfterAbandon(): Promise<void> {
+    if (this.briefcase.txns.hasLocalChanges) {
+      throw new Error("Locks cannot be released while the briefcase contains local changes");
```
kabentley: I think you should create a new `ServerBasedLockErrors extends ITwinError` interface and use it for all lock-related exceptions.

core/backend/src/internal/ServerBasedLocks.ts
```diff
@@ -105,7 +111,26 @@ export class ServerBasedLocks implements LockControl {
     return this[_releaseAllLocks]();
   }
 
-  private insertLock(id: Id64String, state: LockState, origin: LockOrigin): true {
+  public async releaseAllLocksAfterAbandon(): Promise<void> {
```
kabentley: I still don't understand why this isn't `abandonAllLocks`?

core/backend/src/test/standalone/ServerBasedLocks.test.ts
simihartstein: I didn't know about those functions!

Having it as a separate call feels a little clunky. My first thought is that it would be an option on `cancelTo`.

Edit:
Upon reflection, the problem with that is cancelTo is applicable to standaloneDb (or generally speaking briefcases that wouldn't use locks). I suppose it could no-op in that case, but it feels clunky either way...

kring:
> I don't understand why we don't just always abandon locks on undo.

I believe the concern was that redo will then require re-acquiring the locks, and this can fail. Applications will have to deal with that new possibility, which is a breaking change. If a redo fails because someone else edited the iModel, we'll have to either rebase (!) or inform the user their changes are lost. If redo fails because someone else acquired the necessary locks, the redo will be blocked until they're released.

I agree with you that undo _should_ abandon the associated locks, but it sounds like the logistics of doing that could get complicated. That's why Simi suggested starting with the simpler (but still important) use-case of unlocking when discarding session changes.

kring:
> I think we can probably change cancelTo to not reuse TxnIds. Did you look into that?

I haven't. My entire basis for saying "I believe those txn IDs can be reused" is this line in imodel-native:
https://github.com/iTwin/imodel-native/blob/7f7aa3d949f44d4b9c427e4121b03e417895e571/iModelCore/iModelPlatform/DgnCore/TxnManager.cpp#L2604

core/backend/src/internal/ServerBasedLocks.ts
kring: I mentioned this in the top post:

> @kabentley suggested calling this new method something like abandonAllLocks. I thought the term "abandon" had a misleading meaning here, though, because in the context of a lock, "abandoned" usually means "still locked but forgotten about." It's not really the lock that is abandoned, but the changes associated by that lock. So I tried to capture that in the name as best I could. Suggestions welcome.

kabentley: i don't follow that at all. It seems this should be simple: you acquire a lock to reserve the right to perform some action. When you abandon the lock, you abandon that right, presumably with the understanding that you never used it. Releasing the lock indicates that you're done, but you don't say one way or the other whether you actually used it (and we really don't now). 

Why isn't that straightforward?

simihartstein:
> I believe the concern was that redo will then require re-acquiring the locks, and this can fail. Applications will have to deal 
with that new possibility, which is a breaking change. If a redo fails because someone else edited the iModel, we'll have to either rebase (!) or inform the user their changes are lost. If redo fails because someone else acquired the necessary locks, the redo will be blocked until they're released.
> 
> I agree with you that undo _should_ abandon the associated locks, but it sounds like the logistics of doing that could get complicated. That's why Simi suggested starting with the simpler (but still important) use-case of unlocking when discarding session changes.

@kring I was just about to post that, but you beat me to it!

You're right. The concern was that the user might expect redo to be a guaranteed operation. I'm not a big fan of that argument personally, but it's something we can validate in our upcoming user labs.

Having it as an option would be the most flexible and avoid breaking things if we're trying to be completely safe. We could make it the default in a new major if that's desired.

That being said, handling a failed redo seems trivial to me. It amounts to showing an error message and doing nothing else. The only hard part I can think of is disabling the redo button if attempting it would cause a failure. Not a huge deal if we didn't do that imo.

We might be able to handle this at an application framework level (Studio), so each application wouldn't even need to know about it.

kabentley:
> > I don't understand why we don't just always abandon locks on undo.
> 
> I believe the concern was that redo will then require re-acquiring the locks, and this can fail. Applications will have to deal with that new possibility, which is a breaking change. If a redo fails because someone else edited the iModel, we'll have to either rebase (!) or inform the user their changes are lost. If redo fails because someone else acquired the necessary locks, the redo will be blocked until they're released.
> 
> I agree with you that undo _should_ abandon the associated locks, but it sounds like the logistics of doing that could get complicated. That's why Simi suggested starting with the simpler (but still important) use-case of unlocking when discarding session changes.

No, redo failing isn't something applications need to deal with, or even know happens. If someone else reserved the lock while the Txn was reversed, they're in the same state they would have been if they attempted to make a change but were denied due to locking. They didn't lose any work, they have undone it. There's nothing to rebase, and nothing they can do other than wait for the other user to push their changes. THIS is why locks are a terrible concept as i've said repeatedly.

core/backend/src/internal/ServerBasedLocks.ts
kring: Nothing complicated about it, but the term "abandon a lock" means something completely different to me. Here's the AI overview in a Google search (the best source for all information, as we know!), which agrees with the definition in my head:

> An abandoned lock in computer science occurs when a thread or process acquires a lock (mutex, critical section) but terminates, crashes, or stalls before releasing it. This causes a permanent blocking state, where other threads waiting for that resource wait indefinitely, leading to system-wide deadlock or severe performance degradation.

If you feel strongly, I can rename it, of course.

simihartstein:
> redo failing isn't something applications need to deal with, or even know happens

This is true that the "end state" is the same if acquiring a lock fails. The most meaningful difference is the frontend. We prevent manipulators and disable tools that would create a lock conflict if used. The same should be true of the redo button.

>They didn't lose any work, they have undone it

Let me devil's advocate here with a real-world scenario that I do all the time.

1. I delete some recent code I wrote thinking I'm done with it
2. I realize, "_shoot_ I needed some of that code!"
3. I undo until the deleted code I needed comes back, copy the relevant section, redo my changes, and paste it back.

I would definitely be angry and feel like my changes were lost if the redo failed.

Clearly, this is where the "locks are a terrible concept" comes in...

core/backend/src/internal/ServerBasedLocks.ts
kabentley: Yes, I do feel we should make a clear distinction between “abandoning” a lock and releasing it. What your Google search doesn’t know is that the term “lock” in this context means reserving a right to perform an action, not the action itself. When we release a lock, there is no guarantee that it was ever used, because we simply don’t know. We really mean “presume it was used”. Further, the reason we store “change sets” is because a change to one element may depend on changes to other elements. So, there is no valid way to release  _any_ lock without releasing all of them.

Now, this issue arises solely for the (rare) case where we acquire a lock, and immediately change our mind and say “oops, I don’t need that lock and I pinky promise I never used it.”  In that rare case, it is ok to “abandon the right under the lock” for a single lock. 

So you can see why a lock function that says release-after-abandon can make one’s head explode.

kring: > What your Google search doesn’t know is that the term “lock” in this context means reserving a right to perform an action, not the action itself.

A "lock" in computer science terms always represents the right to perform an action or use a resource, never the action/resource itself. I don't think that's a difference.

> So you can see why a lock function that says release-after-abandon can make one’s head explode.

I think you have a strong association between "release" and "I actually modified the protected element" that our users won't necessarily share. After all, I can acquire a regular old mutex, maybe read the resource protected by that mutex (or not), and then decide not to modify the resource. Now I'm done with that mutex, so what do I do? I release it. It would be weird (misleading, even!) to say I'm going to now "abandon" the mutex.

So no, saying I want to release my lock after I abandoned my changes to the elements that the lock was meant to protect doesn't make my head explode. That sounds like normal terminology to me.

But terminology is always contextual and maybe we've trained our users to expect that "release" means "modified", and so my terminology would be confusing. That's not really apparent to me, but I certainly trust your judgement on it. I will rename back to `abandonAllLocks`.

kabentley:
> I would definitely be angry and feel like my changes were lost if the redo failed.
> 
> Clearly, this is where the "locks are a terrible concept" comes in...

Let’s be clear about what you’d be angry about. You acquire a lock and perform an action reserved by it. This blocks other users, who are now angry at you blocking their progress. You relent to their bitching and undo your action and abandon your lock. While you’re staring at your now clean screen, your teammate reserves the lock you abandoned and likely makes progress with it. You change your mind that you really did like your previous edit and attempt to redo it. You can’t and you’re mad at your teammate.  

Locks are indeed a work inhibitor. The much worse scenario is deadlock; you hold a lock they need and they hold a lock you need. In that case SOMEONE will have to lose work.

kring:
It sounds like everyone (that has chimed in so far, at least!) prefers for undo to abandon the associated locks, even if there are some worries. So I'll move this PR in that direction. Thanks for all the comments!

kabentley:
> > I think we can probably change cancelTo to not reuse TxnIds. Did you look into that?
> 
> I haven't. My entire basis for saying "I believe those txn IDs can be reused" is this line in imodel-native: https://github.com/iTwin/imodel-native/blob/7f7aa3d949f44d4b9c427e4121b03e417895e571/iModelCore/iModelPlatform/DgnCore/TxnManager.cpp#L2604

I undoubtedly made that comment many years ago anticipating this day that someone would need to know whether we reuse TxnIds. The alternative is to have “holes” in the TxnId sequence. As I sit here today I can’t think of what problem that would cause. Maybe @khanaffan can think of one.

Frankly, I’d ask ChatGPT or copilot to consider it.

core/backend/src/internal/ServerBasedLocks.ts
kring: I changed this to `throw new IModelError(ChangeSetStatus.HasLocalChanges, ...)`. I looked at introducing a new `ITwinError`, as you suggested, but I'm not totally clear on how that's supposed to work, since `ITwinError` is marked `@beta`. FWIW, this class is already throwing `IModelError` in a few places, so changing those to ITwinError would be a breaking change I believe.

kabentley: That's why I bring this up. We should stop throwing Error or IModelError and instead use `ITwinError`. You're right, we can't change existing error types for public apis, but we should certainly not create any new ones. ErrorNumber is useless since there is no way to enforce uniqueness across apis. ITwinError uses a 2 part string for that.

I suggest you add your new interface to `ITwinCoreErrors.ts`, Look there for the pattern. For example `ChannelControlError`.

kring:
I'm looking at making `reverseTxns` automatically abandon the locks associated with the reversed txn. I thought this would be straightforward, but I now believe it's impossible without a large breaking change. Abandoning locks is an async operation. It has to be, because the Hub API is involved. But `reverseTxns` is synchronous, so it can't include an async operation.

Thoughts?

My best idea is to introduce a new async version of `reverseTxns` and deprecate the old one. Naming it will be awkward.

kring:
I think what I'll do, assuming no one has a better suggestion, is introduce new async functions along the lines of `reverseTxnsAndAbandonLocks`. I'll leave the synchronous `reverseTxns` and similar in place and not deprecate them. I'm assuming that forcing an async reverseTxns on everyone is going to have some very painful ripple effects. But this approach does mean users/applications will need to opt into the new lock abandonment behavior.

rschili:
@kring 
> I think what I'll do, assuming no one has a better suggestion, is introduce new async functions along the lines of `reverseTxnsAndAbandonLocks`. I'll leave the synchronous `reverseTxns` and similar in place and not deprecate them. I'm assuming that forcing an async reverseTxns on everyone is going to have some very painful ripple effects. But this approach does mean users/applications will need to opt into the new lock abandonment behavior.

I believe this is the right choice.
(Leaving pure `txn` methods sync, not deprecate, introduce new async ones alongside which bundle the lock stuff)

khanaffan:
> > > I think we can probably change cancelTo to not reuse TxnIds. Did you look into that?
> > 
> > 
> > I haven't. My entire basis for saying "I believe those txn IDs can be reused" is this line in imodel-native: https://github.com/iTwin/imodel-native/blob/7f7aa3d949f44d4b9c427e4121b03e417895e571/iModelCore/iModelPlatform/DgnCore/TxnManager.cpp#L2604
> 
> I undoubtedly made that comment many years ago anticipating this day that someone would need to know whether we reuse TxnIds. The alternative is to have “holes” in the TxnId sequence. As I sit here today I can’t think of what problem that would cause. Maybe @khanaffan can think of one.
> 
> Frankly, I’d ask ChatGPT or copilot to consider it.

I do not think there is issue with having holes in txnId. In fact I consider edit as requirement for  [StashManager](https://github.com/iTwin/itwinjs-core/blob/master/core/backend/src/StashManager.ts). So we can have stash1 and stash2 and they should not have overlapping ids. Though in practice since we only use stash manager for recovery when rebase fails application like OS+ uses it to abort and restore briefcase exactly as it was before pulll/merge. The reason for failure is generally application code. 

Also not I found another bug where ElementId sequence  was reset on pull/merge to MAX(elementid) this also caused issue during rebase as we try to insert new element and it failed due duplicate Elementid. So we make sure when you do/undo or do pull-merge a given briefcase never reuses ElementId sequence.

khanaffan:
@kring @simihartstein If user really like shelf changes e.g. going on vacation then we can also improve our stashing api. It allow you to shelf current changes and optionally release locks. We currently use it as recovery mechanism when rebase fails. It implementing is different in that it copies txns and locks into a stash file (Sqlite db). Then it can move briefcase back in time and practically restore it to exactly point on which user started rebase include. undo history and session. So user can still go on and do undo or redo or create new txns continuing previous work.

core/backend/src/test/standalone/ServerBasedLocks.test.ts
khanaffan: `cancelTo()` is synchronous and we need like async api to release lock. We can have new function `abortTxns({start: Id64String, end? Id64String, releaseLocks?: false ): Promise<void>`. I just think application should not be responsible to release lock. Any mistake there can lead to bad data being pushed which have cascading issues when pull by other users.

kabentley:
> I believe this is the right choice. (Leaving pure `txn` methods sync, not deprecate, introduce new async ones alongside which bundle the lock stuff)

(sorry about that fat finger closing your PR).

I agree. The native layers are sync and should stay that way. Adding an async layer in TS is ok, but remember this whole locking process should be unnecessary one fine day.

core/backend/src/internal/ServerBasedLocks.ts
kring: This is done. In `ServerBasedLocks`:

1. Places that were previously throwing `Error` now throw `ServerBasedLocksError` instead, with the same message. This should not be a breaking change.
2. Places that were previously throwing `ModelError` continue to do so, because switching to `ITwinError` here would be breaking.
3. All _new_ exceptions are `ServerBasedLocksError`.

However, there are some new `IModelErrors` in `TxnManager` in this PR. This is because:

1. They are turning an `IModelStatus` returned by an existing synchronous API into a Promise rejection. That works well for an IModelError (`throw new IModelError(status, IModelError.getErrorKey(status));`) but I can't see an obvious way to turn it into an ITwinError.
2. In `reinstateTxnAndAcquireLocks`, there's an explicit throw of an `IModelError` that could easily be made into an `ITwinError`. But it seems wrong for that one case to be an `ITwinError`, and different from the synchronous version of the function, when everything else is `IModelError`.

core/backend/src/BackendHubAccess.ts
kring: I think I'm going to leave this as-is. The revised `BackendHubAccess` in this PR I believe reflects where we want to head with the iModels API itself. That is, there should be clearly-defined valid operations, rather than a free-for-all "update locks". The `ServerBasedLocks` class has all of the logic necessary to use this improved API, if it's available, or to provide backward compatibility if not.

core/backend/src/BackendHubAccess.ts
kring: I've changed the doc to state that an exception is expected.

kring:
This PR now adds the following new methods to `TxnManager`:

* `reverseTxnsAndAbandonLocks`
* `reverseSingleTxnAndAbandonLocks`
* `reverseAllTxnsAndAbandonLocks`
* `reverseToTxnAndAbandonLocks`
* `cancelToTxnAndAbandonLocks`
* `reinstateTxnAndAcquireLocks`

These are all async methods.

For the reverse and cancel methods, the lock abandonment happens after the reversal. So, in the (unlikely?) event the abandonment fails, the txn will remain reversed but the locks will still be held. You can "try again" by calling `locks.abandonLocksForReversedTxn`.

This is true also for `cancelToTxnAndAbandonLocks`. If the lock abandonment fails, the txn will only be reversed, not canceled. This means you can retry the lock abandonment. Or call `cancelTo` to cancel the txns without abandoning locks.

In the reinstate method, if any of the locks cannot be acquired, the promise will reject and the txn will not be reinstated.

simihartstein:
> This PR now adds the following new methods to `TxnManager`:
> 
> * `reverseTxnsAndAbandonLocks`
> * `reverseSingleTxnAndAbandonLocks`
> * `reverseAllTxnsAndAbandonLocks`
> * `reverseToTxnAndAbandonLocks`
> * `cancelToTxnAndAbandonLocks`
> * `reinstateTxnAndAcquireLocks`
> 
> These are all async methods.


@kring From an API perspective, instead of duplicating every function to handle locks specifically, it might be cleaner to have identically named async versions that take an options argument. This would provide more flexibility in case we have need for other options in the future.

* `reverseTxnsAsync({abandonLocks: boolean})`
* `reverseSingleTxnAsync({abandonLocks: boolean})`
* `reverseAllTxnsAsync({abandonLocks: boolean})`
* `reverseToTxnAsync({abandonLocks: boolean})`
* `cancelToTxnAsync({abandonLocks: boolean})`
* `reinstateTxnAsync({acquireLocks: boolean})`

kring:
Good idea, thanks @simihartstein. I'll default all those booleans to `true`.

kring:
Well, I keep making pretty significant changes to this. For good reasons, I'd say, but it's probably a sign I don't totally have a handle on it yet. But I think it's getting closer. Recent changes:

* Implemented Simi's suggestion to make generic "Async" versions of the reverse/cancel/reinstate functions, and pass options to control lock behavior.
* Rather than an option called `abandonLocks` in the reverse functions, I added an option called `retainLocks`. This makes it more consistent with `discardChanges` and also avoids the slightly awkward situation where an arg value defaults to true.
* I also did not add an `acquireLocks` option to `reinstateTxnAsync`, because allowing edits without holding the necessary locks would be a violation of the locking policy. If the locks were previously abandoned, they _must_ be reacquired prior to reinstating the txn.
* `reinstateTxn` now returns an error if the necessary locks have been abandoned and not yet re-acquired. There's no way for this synchronous method to re-acquire the locks, so an error is the only option (other than allowing violation of the locking policy). Is this a breaking change? I say definitely no, because it only comes up if you use the new lock abandonment capability in some places, but don't account for it in all places. Existing code will continue to work as it did before, where locks are not abandoned, and do not need to be re-acquired.
* To support the above, I added two more methods to `LockControl`: `abandonLocksForCurrentUnsavedTxn` and `holdsNecessaryLocksForReinstatingTxn`.
* The reinstate args do have a `retainLocks` option. It only affects the limited scenario where the current unsaved txn holds locks before reinstating a txn, and allows the user to request that these not be abandoned.

kring:
Probably a stupid question, but why did imodeljs-admin make a commit to this branch with a bunch of conflicts? 😕

core/backend/src/internal/ServerBasedLocks.ts
```diff
@@ -114,9 +162,26 @@ export class ServerBasedLocks implements LockControl {
...
+    // be cleared (no longer reinstateable) on commit.
+    this.lockDb.withPreparedSqliteStatement("INSERT INTO txn_locks(txnId,elementId,state,origin,abandoned) VALUES (?,?,?,?,FALSE)", (stmt) => {
+      stmt.bindId(1, this._unsavedChangesTxnId);
+      stmt.bindId(2, id);
+      stmt.bindInteger(3, state);
+      stmt.bindInteger(4, origin);
+      const rc = stmt.step();
+      if (DbResult.BE_SQLITE_DONE !== rc)
+        ServerBasedLocksError.throwError("lock-database-problem", `can't insert txn lock record into database (error code ${rc})`);
+    });
```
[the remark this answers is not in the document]
kring: Fixed by making (txnId,elementId) the primary key and updating on conflict.

core/backend/src/internal/ServerBasedLocks.ts
```diff
@@ -200,12 +267,249 @@ export class ServerBasedLocks implements LockControl {
...
+  }
+
+  public clearTxnLockRecords(txnId: Id64String) {
+    this.lockDb.withPreparedSqliteStatement("DELETE FROM txn_locks WHERE txnId!=? AND txnId>=?", (stmt) => {
+      stmt.bindId(1, this._unsavedChangesTxnId);
+      stmt.bindId(2, txnId);
+      const rc = stmt.step();
+      if (DbResult.BE_SQLITE_DONE !== rc)
+        ServerBasedLocksError.throwError("lock-database-problem", `can't delete txn lock records from database (error code ${rc})`);
+    });
```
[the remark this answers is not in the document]
kring: Added the saveChanges.

core/backend/src/internal/ServerBasedLocks.ts
```diff
@@ -39,20 +39,48 @@ export class ServerBasedLocks implements LockControl {
...
+      // so clear out the record of their locks. If the locks are still held, sorry, it's too late!
+      this.clearTxnLockRecords(committedTxnId);
+
+      // All of the "current" changes are now part of a real txn, so update the txn id accordingly.
+      this.lockDb.withPreparedSqliteStatement("UPDATE txn_locks SET txnId=? WHERE txnId=?", (stmt) => {
+        stmt.bindId(1, committedTxnId);
+        stmt.bindId(2, this._unsavedChangesTxnId);
+        const rc = stmt.step();
+        if (DbResult.BE_SQLITE_DONE !== rc)
+          ServerBasedLocksError.throwError("lock-database-problem", `can't mark txn locks as abandoned in database (error code ${rc})`);
```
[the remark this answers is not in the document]
kring: Fixed copy/paste error.

core/backend/src/LocalHub.ts
```diff
@@ -649,11 +649,23 @@ export class LocalHub {
     this.db.saveChanges();
   }
 
+  public abandonLocks(locks: LockMap, arg: BriefcaseIdArg) {
+    const releaseArg = { briefcaseId: arg.briefcaseId, changesetIndex: 0 };
+    for (const props of locks)
+      this.releaseLock({ id: props[0], state: props[1] }, releaseArg);
```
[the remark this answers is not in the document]
kring: True, but covered by #8749 and IMO outside the scope of this PR.

ben-polinsky:
> Probably a stupid question, but why did imodeljs-admin make a commit to this branch with a bunch of conflicts? 😕

It's because you've linked the native PR in your description. Once that PR merges, the native addon is merged back into this PR.

core/backend/src/LocalHub.ts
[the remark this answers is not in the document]
kring: This was addressed in #9102.

kring:
I've updated the top post to better summarize the state of this PR. I believe it is ready to merge, subject to further review of course.

Two things I'd like to see (but not strictly required) before merging:
1. I'd love to see Simi or someone from the OpenSite+ team try this out to make sure it addresses the original need expressed in #8552. I'm reasonably confident it will, but it would be great to have the confirmation. I can also have a go at this myself if that sounds better to people.
2. Confirmation that making the new TxnManager and LockControl methods "public" right from the start is the right approach. Or would it be better if they start out as beta?

DanRod1999:
It seems like this PR is fairly close to being wrapped up. We do plan on releasing the next minor version of itwinjs by end of next week. Do we think this PR will be ready by then and/or do we consider it high priority to get into this next minor

kring:
@DanRod1999 there's no reason - to my knowledge - to rush to get it into next week's release. From my perspective it is ready, but it will depend on whether the appropriate people find the time to review before then, and whether they have any concerns.

common/api/summary/core-backend.exports.csv
```diff
@@ -473,6 +473,7 @@ beta;type;QueryWorkspaceResourcesCallback
 alpha;interface;RebaseHandler
 alpha;class;RebaseManager
 beta;class;RecipeDefinitionElement
+public;interface;ReinstateTxnArgs
```
aruniverse: can we start these off at beta and then promote them down the road when we feel confident in them?
id prefer we dont start off the bat w/ public apis that we now cant break for another 2 major version

simihartstein:
@kring 
> 1. I'd love to see Simi or someone from the OpenSite+ team try this out to make sure it addresses the original need expressed in [Automatically release locks when undoing Txns #8552](https://github.com/iTwin/itwinjs-core/issues/8552). I'm reasonably confident it will, but it would be great to have the confirmation. I can also have a go at this myself if that sounds better to people.

@smmr-dn Will be testing this out in OpenSite+

core/backend/src/internal/ServerBasedLocks.ts
```diff
@@ -200,12 +281,251 @@ export class ServerBasedLocks implements LockControl {
...
+
+    // So we simply have to find all non-abandoned locks associated with the given txn or later, abandon them (or restore
+    // them to their previous state), and mark them as abandoned in txn_locks.
+
+    // Find all locks associated with the given txnId or later.
+    // For each elementId, find the previous state of the lock before this Txn (if any), or None otherwise.
+    // This is the state that we will restore the element's lock to. The reason we do this is to account for
+    // lock upgrades. If an earlier Txn acquired a Shared lock on this element, and this Txn acquired an
+    // Exclusive lock, we should restore the Shared lock.
+    // TODO: Keith says this isn't necessary, but I don't fully understand why not. So I'm including it for now.
```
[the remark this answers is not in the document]
kring: I'm going to remove this TODO. I'm increasingly convinced this logic is necessary.

I think Keith's point, when he previously told me it wasn't necessary, was that reverting an Exclusive lock back to Shared would be complicated to do, and the missing Shared lock would be detected later anyway, e.g. as a conflict during rebase. I don't doubt this is true, but I think it's much better to preserve the normal lock invariants rather than try to make a case that it's ok not to preserve them in this particular case. Especially given that I don't think the implementation here that achieves it is overly complicated.

kring:
All the new stuff is now marked `@beta`.

kring:
I built Studio against this branch and ran OS+ (unmodified) in it. And much to my surprise, OS+ undo/redo operations don't call into the code I modified at all! TIL about `IpcHost.ts`. The reverse and reinstate functions on that class don't call the methods on `IModelDb`. Instead, they directly pull the `_nativeDb` field out of the instance and call straight into the native code. See [here](https://github.com/iTwin/itwinjs-core/blob/3cebdeb1babefa4a6700bad0b03d9f0a286665dd/core/backend/src/IpcHost.ts#L345).

So I guess I will fix that. Looks like I will also need to add the new async functions here and to BriefcaseTxns so that they can be called from the frontend.

kring:
I've now tried this branch in OS+, and made some further changes based on that experience. A few notes:

1. I previously hadn't exposed the new async reverse/reinstate methods through the frontend's `BriefcaseTxns`. Now I have.
2. Previously, calls to the sync reverse/reinstate functions were going directly to the imodel-native layer, instead of calling into `TxnManager`. I've now changed this in this PR. However, it did require this slightly awkward structure because not all `IModelDb` instances have a `TxnManager`:

```ts
    const db = IModelDb.findByKey(key);
    if (db.isBriefcaseDb()) {
      return db.txns.reverseTxns(numOperations);
    } else {
      return db[_nativeDb].reverseTxns(numOperations);
    }
```

I don't _think_ it makes any sense to call `reverseTxns` on something that's not a briefcase DB, and so it would be fine to throw an exception instead, but I wasn't sure. For the new async methods, where there are fewer backward compatibility concerns, I've chosen to throw.

3. I was unwittingly passing element IDs as decimal strings (rather than hexadecimal) to `BackendHubAccess` methods. HubMock was ok with this, but the real Hub was not. Now fixed.
4. I had a lot of trouble running OS+ / local Studio / custom itwinjs-core in the VS Code debugger. The Extension Host process hosting the javascript debugger kept exceeding the heap space limit (4GB) and crashing. After trying many things to fix this, none of which worked, I eventually learned I could use the Chrome DevTools instead, which was much more reliable. I have instructions for this whole process written down, which I should share somewhere.
5. OS+'s undo/redo appears to be handled entirely within itwinjs-core, specifically `core/frontend/src/tools/ToolAdmin.ts`. So now this PR modifies that code to use the new async methods, which automatically abandon and reacquire locks. OS+ should get this new behavior without any further changes.
6. The lock abandonment seems to be working as expected. I tried opening the same iModel in OS+ on two different computers. When I make a change on one, the object turns red on the other. When I undo that change, the red highlight on the other disappears.
7. If I make a change on one computer, undo it, make a conflicting change on the other computer, and then attempt to Redo on the first, OS+ silently does nothing. I see the following in the log:

> 2026-04-10 11:29:06.724 [error] |iTwinStudio.Backend| ConflictWithAnotherUser: Lock(s) is owned by another Briefcase. Conflicting locks:
> 1. Object id: 0x200000004ef, lock level: exclusive, briefcase ids: 3
>
> Response from requested url: undefined, status code: undefined 

Could be some room for improvement here, but it doesn't seem too bad.

8. I noticed that OS+ acquires a bunch of shared locks at startup. This happens in `apps/booster/src/backend/Site.ts` `startEditScope` method. In this PR, those shared locks get associated with the initial Txn. Which means that if we edit something in that Txn, and then undo it, those locks are abandoned as well. If we then redo that operation, the locks come back, which is fine. But if instead of redoing we make some _other_ edit, those shared locks will never be reacquired. I don't know exactly why these shared locks are needed, but I'm guessing this is a problem. I'd appreciate any thoughts on this.

kring:
@kabentley you have a couple of unresolved comments above, which I think are now the only thing stopping this PR from being auto-merged. Mind taking a look and seeing if you have any concerns? Thanks!

pmconne:
> not all IModelDb instances have a TxnManager

Use `BriefcaseDb.findByKey` instead. EDIT: that function uses `as`. It should be changed to throw if an `instanceof BriefcaseDb` check fails.

kring:
Thanks @pmconne.

> Use BriefcaseDb.findByKey instead. EDIT: that function uses as. It should be changed to throw if an instanceof BriefcaseDb check fails.

Looks like `BriefcaseDb.tryFindByKey` should work well. But are you suggesting I do this in all methods, or only in the new async ones?

pmconne:
> Looks like `BriefcaseDb.tryFindByKey` should work well. But are you suggesting I do this in all methods, or only in the new async ones?

I spelunked into the rabbit hole and found that `BriefcaseDb.findByKey` calls `super.findByKey` which calls `this.tryFindByKey` in which case `this` is the class `BriefcaseDb` which will invoke `BriefcaseDb`'s override of `tryFindByKey` which returns `undefined` if the iModel is not a `BriefcaseDb`. That's not at all clear from even a moderately attentive reading of the code (see [here](https://github.com/iTwin/itwinjs-core/pull/376#discussion_r540380149)), but it means that wherever you require a `BriefcaseDb` (e.g., when reversing txns, which only makes sense for briefcases), you should use `BriefcaseDb.findByKey` and not have to worry about handling the "not a briefcase" path, nor the `undefined` path - it'll throw, as desired, if no briefcase is associated with the specified key.

kring:
Ok thanks @pmconne, I've changed IpcHost.ts to use `BriefcaseDb.findByKey` in all the reverse and reinstate methods, including both the existing sync and the new async ones.

PR #9052, opened 2026-03-03
Native Bulk Element Deletion API

RohitPtnkr1996:
imodel-native: https://github.com/iTwin/imodel-native/pull/1344

I have added performance tests comparing deleteElement (old) vs deleteElements (new) and deleteDefinitionElements (old) vs deleteElements (new).

The new APIs were expected to do additional validations like checking for intra and inter set FK violations, auto-handling parent child deletions within the same call, expanding sub-models if a modeling element is being deleted, making the deletions element-order-agnostic (all of which are not handled by the old API calls).

Performance stats from the new full-stack-tests perftest added:

1. deleteElement (loop) vs deleteElements (bulk)

Element Count | deleteElement | deleteElements | Improvement %
-- | -- | -- | --
5 | 7.39 ms | 3.95 ms | 46.5%
25 | 11.67 ms | 7.61 ms | 34.8%
50 | 18.12 ms | 12.45 ms | 31.3%
250 | 266.90 ms | 57.92 ms | 78.3%
1,000 | 1.00 secs | 245.92 ms | 75.6%
10,000 | 12.51 secs | 2.90 secs | 76.8%
100,000 | 3.56 mins | 34.84 secs | 83.7%
1 M | 19.04 mins | 6.06 mins | 68.2%
1.7 M | 32.77 mins | 10.73 mins | 67.3%

2. deleteDefinitionElements vs deleteElements 

Element Count | deleteDefinitionElements | deleteElements | Improvement %
-- | -- | -- | --
5 | 4.38 ms | 12.81 ms | -192.3%
10 | 5.90 ms | 15.18 ms | -157.2%
50 | 18.75 ms | 25.99 ms | -38.6%
100 | 35.83 ms | 41.46 ms | -15.7%
1,000 | 357.16 ms | 262.02 ms | 26.6%
10,000 | 6.16 secs | 3.24 secs | 47.4%
100,000 | 1.47 mins | 35.73 secs | 59.7%
1 M | 17.57 mins | 7.16 mins | 59.2%

When definition elements exist in the delete set, we run checks to ensure there is no element which is being used, along with validations for FK violations, parent-child and sub-models cascade delete support.
This gives us a better deletion confidence and a reduction in API call failures, but at the expense of a performance hit.
Adding to that are the domain handler callbacks that need to be called before and after deleting every element, which take up a chunk of the processing time.

Hence, the validations add significant overhead to the running time.
A more realistic performance check will be to run the new Api against the internal tool that the user has written to do the validations and compare the performances.

@khanaffan, will you please take a look at the tests and the [native API](https://github.com/iTwin/imodel-native/pull/1344) ?

### DIFF
--- a/common/api/core-backend.api.md
+++ b/common/api/core-backend.api.md
@@ -5716,6 +5716,7 @@ export type QueryWorkspaceResourcesCallback = (resources: Iterable<{
 
 // @alpha
 export interface RebaseHandler {
+    dispose?(): void;
     recompute(txn: TxnProps): Promise<void>;
     shouldReinstate(txn: TxnProps): boolean;
 }
@@ -5729,6 +5730,7 @@ export class RebaseManager {
         handler: (args: RebaseChangesetConflictArgs) => DbConflictResolution | undefined;
     }): void;
     canAbort(): boolean;
+    dispose(): void;
     inProgress(): boolean;
     get isAborting(): boolean;
     get isMerging(): boolean;
--- a/common/changes/@itwin/core-backend/copilot-add-dispose-method-on-rebaser_2026-04-13-13-59.json
+++ b/common/changes/@itwin/core-backend/copilot-add-dispose-method-on-rebaser_2026-04-13-13-59.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Add dispose method to RebaseManager",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/core/backend/src/TxnManager.ts
+++ b/core/backend/src/TxnManager.ts
@@ -324,6 +324,7 @@ export class RebaseManager {
   private _conflictHandlers?: IConflictHandler;
   private _customHandler?: RebaseHandler;
   private _aborting: boolean = false;
+  private _disposed: boolean = false;
 
   /** Event raised before pull merge process begins.
    * @alpha
@@ -446,6 +447,30 @@ export class RebaseManager {
   }
 
   public constructor(private _iModel: BriefcaseDb) { }
+
+  /** Disposes of this RebaseManager, clearing all event listeners.
+   * Also calls [[RebaseHandler.dispose]] on the registered custom handler, if any.
+   * Subsequent calls are ignored.
+   * @alpha
+   */
+  public dispose(): void {
+    if (this._disposed)
+      return;
+    this._disposed = true;
+    this._customHandler?.dispose?.();
+    this.onPullMergeBegin.clear();
+    this.onRebaseBegin.clear();
+    this.onRebaseTxnBegin.clear();
+    this.onRebaseTxnEnd.clear();
+    this.onRebaseEnd.clear();
+    this.onPullMergeEnd.clear();
+    this.onApplyIncomingChangesBegin.clear();
+    this.onApplyIncomingChangesEnd.clear();
+    this.onReverseLocalChangesBegin.clear();
+    this.onReverseLocalChangesEnd.clear();
+    this.onDownloadChangesetsBegin.clear();
+    this.onDownloadChangesetsEnd.clear();
+  }
 
   /**
    * Resumes the rebase process for the current iModel, applying any pending local changes
@@ -865,6 +890,7 @@ export class TxnManager {
     this.rebaser = new RebaseManager(_iModel);
     _iModel.onBeforeClose.addOnce(() => {
       this._isDisposed = true;
+      this.rebaser.dispose();
     });
   }
 
@@ -1595,4 +1621,11 @@ export interface RebaseHandler {
    * @alpha
    */
   recompute(txn: TxnProps): Promise<void>;
+  /**
+   * Called when the owning [[RebaseManager]] is disposed (e.g. when the iModel is closed).
+   * Override this method to unsubscribe from events or release resources held by this handler.
+   *
+   * @alpha
+   */
+  dispose?(): void;
 }
--- a/core/backend/src/test/standalone/TxnManager.test.ts
+++ b/core/backend/src/test/standalone/TxnManager.test.ts
@@ -13,7 +13,7 @@ import {
   _nativeDb,
   ChangeInstanceKey,
   ChannelControl,
-  EditTxn, IModelDb, IModelJsFs, PhysicalModel, PhysicalPartition, setMaxEntitiesPerEvent, SpatialCategory, StandaloneDb, SubCategory, SubjectOwnsPartitionElements, TxnChangedEntities, TxnManager, withEditTxn,
+  EditTxn, IModelDb, IModelJsFs, PhysicalModel, PhysicalPartition, RebaseHandler, setMaxEntitiesPerEvent, SpatialCategory, StandaloneDb, SubCategory, SubjectOwnsPartitionElements, TxnChangedEntities, TxnManager, withEditTxn,
 } from "../../core-backend";
 import { IModelTestUtils, TestElementDrivesElement, TestPhysicalObject, TestPhysicalObjectProps } from "../IModelTestUtils";
 import { IModelNative } from "../../internal/NativePlatform";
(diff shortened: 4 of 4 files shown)


core/backend/src/IModelDb.ts
```diff
@@ -2851,6 +2851,17 @@ export namespace IModelDb {
       });
     }
 
+    /**
+     * Delete multiple elements from the model.
+     * @param ids The identifiers of the elements to delete.
+     * @returns A set of identifiers for any elements that could not be deleted.
+     * @beta
+     */
+    public deleteElements(ids: Id64Array, skipHandlerCallbacks: boolean = false): Id64Set {
```
ben-polinsky: Please document the `skipHandlerCallbacks` argument.

Also, if you think there's a chance we'll add more arguments in the future, an options object is preferred.

RohitPtnkr1996: I was just experimenting with skipping domain handlers to measure the performance impact.
Removed the argument.

rschili (approved):
I approve, but those performance numbers make me wonder if this is really good. The gain is not big, and the slowdown with <50 deletes is significant. Maybe the typescript deleteElements method could check the size of the provided ID set and then call one or the other internally?

RohitPtnkr1996:
> I approve, but those performance numbers make me wonder if this is really good. The gain is not big, and the slowdown with <50 deletes is significant. Maybe the typescript deleteElements method could check the size of the provided ID set and then call one or the other internally?

Yeah, the performance numbers are a little unfair to be honest.
The old api simply deletes the elements one at a time without any validations on the elements being deleted.
The users have written their own tool in typescript that does all the validation before calling deleteElement.

My plan was to ask the users to try out this beta which does everything on the native side instead of their TS tool and share their findings on the performance.

core/backend/src/IModelDb.ts
```diff
@@ -2886,6 +2886,26 @@ export namespace IModelDb {
...
+     * @param ids The ids of the elements to delete. Any invalid ids will be ignored.
+     * @returns A set of valid ids for any elements that could not be deleted.
+     * @beta
+     */
+    public deleteElements(ids: Id64Array): Id64Set {
+      const failedToDelete = this._iModel[_nativeDb].deleteElements(ids);
+      const failedToDeleteSet = failedToDelete ? Id64.toIdSet(failedToDelete) : new Set<Id64String>();
+
+      for (const id of ids) {
+        if (Id64.isValidId64(id) && !failedToDeleteSet.has(id)) {
```
tcobbs-bentley: Maybe it would be better to check the input before calling `deleteElements` on `_nativeDb` and throw an exception for invalid input? By the time we get here, the id hopefully is either valid or in `failedToDelete`. If you do that, please add a test that tests invalid input.

docs/learning/backend/BulkElementDeletion.md
```diff
@@ -0,0 +1,97 @@
...
+iModel.elements.deleteElements([viewId, categorySelectorId, modelSelectorId, displayStyleId]);
+```
+
+## Constraint violations
+
+An element is a _constraint violator_ and will **not** be deleted (neither will its subtree) if, after removing intra-set references, any of the following are true for elements **outside** the delete set:
+
+| Constraint | Blocked element |
+|---|---|
+| Another element uses this element as its **CodeScope** | The code-scope element (and its' subtree as well as all its ancestors up to the highest element present in the delete set) |
```
tcobbs-bentley:
```suggestion
| Another element uses this element as its **CodeScope** | The code-scope element (and its subtree as well as all its ancestors up to the highest element present in the delete set) |
```

docs/learning/backend/BulkElementDeletion.md
```diff
@@ -0,0 +1,97 @@
...
+- **`ElementDrivesElement`** — rows where either the source or target element is deleted.
+- **`ModelSelectorRefersToModels`** — rows where the target model (modeled element) is deleted.
+
+No manual cleanup of these tables is required.
+
+## Lifecycle callbacks
+
+`deleteElements` fires the standard lifecycle callbacks for every element in the final delete set (after constraint violators have been pruned):
+
+- `_OnDelete` / `_OnDeleted` — fired for every element that was originally in the input array.
```
tcobbs-bentley: Just to confirm, the children that are automatically deleted don't get reported at all? Is this intentional? It seems like a bunch of stuff will get deleted with no callbacks.

```
A -> B -> C -> D
```

So deleting `A` would cause `_OnDelete` / `_OnDeleted` calls to happen for A and that's it (assuming the arrows point to children)?

core/backend/src/IModelDb.ts
RohitPtnkr1996: Done!

docs/learning/backend/BulkElementDeletion.md
RohitPtnkr1996: I made the bulk deletion API consistent with the single element deletion.

The _OnDelete is essentially a veto/pre-notification hook, so we only call it for the elements the user passed for deletion.
However, the _OnDeleted needs to react to the deletion and that gets called for every element that is being deleted.

I have updated the doc.
Thanks!

hl662:
Is there a new entry on NextVersion.md that perhaps mentions this is experimental and encourage users to try using it instead of our existing deleteElement API?

hl662 (commented):
Reviewed the `deleteElements` bulk deletion API across four phases (API surface, core logic, tests, documentation). Overall this is a solid addition — strong test coverage for cascading, constraint checking, and edge cases. A few items below, roughly ordered by importance.

**File-level notes** (no specific line):
- All tests go through `EditTxn.deleteElements` — the public `IModelDb.Elements.deleteElements` wrapper (with its own input validation, cache eviction, and return-set construction) isn't exercised directly. A small integration test calling `iModelDb.elements.deleteElements(...)` would cover that path.
- The `ModelSelectorRefersToModels` link-table cleanup is well covered — any appetite for a similar test with `ElementRefersToElements` or `ElementDrivesElement`? The learning doc mentions those are cleaned up too, but there's no test verifying it.

core/backend/src/EditTxn.ts
```diff
@@ -225,6 +225,17 @@ export class EditTxn {
...
   }
 
+  /**
+     * Delete multiple elements from the model.
+     * @param ids The ids of the elements to delete. All ids must be well-formed and valid [[Id64String]]s.
+     * @returns A set of ids for any elements that could not be deleted.
+     * @throws [[IModelError]] if any of the supplied ids are not well-formed/valid [[Id64String]]s.
+     * @beta
+     */
+  public deleteElements(ids: Id64Array): Id64Set {
```
hl662, line 235: Missing `verifyWriteable()` call — this is the only mutating method on `EditTxn` that doesn't have it. Looks like it got lost in the shuffle? Every sibling method (`deleteElement`, `deleteDefinitionElements`, `insertElement`, etc.) has it as the first line.

core/backend/src/IModelDb.ts
```diff
@@ -2874,6 +2874,37 @@ export namespace IModelDb {
...
     }
 
+    /**
+     * Delete multiple elements from the model.
+     * @param ids The ids of the elements to delete. All ids must be well-formed and valid [[Id64String]]s.
+     * @returns A set of ids for any elements that could not be deleted.
+     * @throws [[IModelError]] if any of the supplied ids are not well-formed/valid [[Id64String]]s.
+     * @beta
+     */
+    public deleteElements(ids: Id64Array): Id64Set {
```
hl662, line 2884: It looks like the implementation here predates the `EditTxn` introduction — the April 9 merge picked up `EditTxn` and a thin wrapper was added, but the logic stayed in `IModelDb.Elements`. To complete the migration, might be worth moving the validation + native call + cache cleanup into `EditTxn.deleteElements` (with `verifyWriteable()`), and having this method delegate through `_implicitTxn` with a `@deprecated` tag — matching `deleteElement` and `deleteDefinitionElements` right above.

core/backend/src/EditTxn.ts
```diff
@@ -225,6 +225,17 @@ export class EditTxn {
     });
   }
 
+  /**
+     * Delete multiple elements from the model.
```
hl662, line 229: Nit: "from the model" → "from the iModel" — matches every other delete method and avoids ambiguity with `Model`.

core/backend/src/EditTxn.ts
```diff
@@ -225,6 +225,17 @@ export class EditTxn {
     });
   }
 
+  /**
```
hl662, line 228: Nit: the JSDoc block here uses 5-space indent while the rest of the file uses the standard `/** ... \n * @param` format. Mind reformatting to match?

core/backend/src/EditTxn.ts
```diff
@@ -225,6 +225,17 @@ export class EditTxn {
...
+     * Delete multiple elements from the model.
+     * @param ids The ids of the elements to delete. All ids must be well-formed and valid [[Id64String]]s.
+     * @returns A set of ids for any elements that could not be deleted.
+     * @throws [[IModelError]] if any of the supplied ids are not well-formed/valid [[Id64String]]s.
```
hl662, line 232: Minor: `@throws [[IModelError]]` — the neighboring `deleteElement` documents `@throws [[ITwinError]]`. Since the codebase is migrating toward `ITwinError`, might be worth aligning.

common/changes/@itwin/core-backend/rohitptnkr-bulk-element-deletion_2026-03-03-14-06.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Added native apis to bulk delete definition and non-definition elements",
```
hl662, line 5: The rush change comment mentions "native apis" which is an implementation detail — consumers won't see that. Something like "Added `deleteElements` method for bulk element deletion with automatic cascade, constraint checking, and cleanup" would better describe the user-facing addition.

core/backend/src/test/element/DeleteDefinitionElements.test.ts
```diff
@@ -250,4 +299,1063 @@ describe("DeleteDefinitionElements", () => {
...
+
+      it("throws for invalid ids", () => {
+        assert.throws(() => txn.deleteElements([Id64.invalid, Id64.invalid]));
+      });
+
+      it("keeps an in-use definition element in the failed set and does not delete it", () => {
+        // geometryPartId is referenced by PhysicalObject1's geometry stream
+        const failed = txn.deleteElements([geometryPartId]);
+        assert.isTrue(failed.has(geometryPartId), "in-use definition element must be in failed set");
+        assert.isDefined(geometryPartId, "in-use definition element must not be deleted");
```
hl662, line 558: Six of the `assert.isDefined(...)` calls in the definition element tests (lines 558, 589, 600, 649, 657, 663) assert on the raw `Id64String` variable instead of doing an element lookup — so they always pass. Should be `assert.isDefined(iModelDb.elements.tryGetElement(id), ...)` to actually verify the element survived deletion.

core/backend/src/test/standalone/DeleteElements.test.ts
```diff
@@ -0,0 +1,820 @@
...
+    txn.saveChanges();
+    return id;
+  };
+
+  /** Assert that the element with the given id exists or has been deleted. */
+  const assertExists = (id: Id64String, msg: string) => assert.isDefined(iModelDb.elements.tryGetElement(id), msg);
+  const assertDeleted = (id: Id64String, msg: string) => assert.isUndefined(iModelDb.elements.tryGetElement(id), msg);
+
+  // Run deleteElements, then verify each id in `deleted` is gone and each id in `retained` is still present.
+  const executeTestCase = (label: string, idsToDelete: Id64Array, deleted: Id64Array, retained: Id64Array) => {
```
hl662, line 73: The `executeTestCase` helper here discards the `Id64Set` returned by `deleteElements` — unlike the equivalent helper in `DeleteDefinitionElements.test.ts` which asserts `sameMembers` on the failed set. Might be worth adding an `expectedFailed` parameter here too? The code-scope pruning tests especially would benefit, since the failed set is the primary signal distinguishing "not requested" from "constraint-blocked."

docs/learning/backend/BulkElementDeletion.md
```diff
@@ -0,0 +1,98 @@
...
+Pass an array of element IDs to delete. The method returns a set of IDs that could **not** be deleted because they would violate an integrity constraint (see [Constraint violations](#constraint-violations) below).
+
+```typescript
+const failed: Id64Set = iModel.elements.deleteElements([idA, idB, idC]);
+if (failed.size > 0) {
+  // Some elements could not be deleted; inspect `failed` to see which ones.
+}
+```
+
+You do **not** need to call [IModelDb.Elements.deleteDefinitionElements]($backend) for DefinitionElements. `deleteElements` handles both ordinary elements and DefinitionElements with indentical semantics.
```
hl662, line 16: A few typos in lines 16-17: "indentical" → "identical", "silenty" → "silently", "where as" → "whereas".

docs/learning/backend/BulkElementDeletion.md
```diff
@@ -0,0 +1,98 @@
...
+The only difference being `deleteDefinitionElements` will silenty ignore non-definition elements in the set, where as `deleteElements`, being able to handle both will delete everything that it can (definition/non-definition).
```
hl662, line 17: Minor wording nit — line 16 says "identical semantics" and then line 17 immediately describes a difference. Maybe something like: "…handles both ordinary elements and DefinitionElements — it performs the same usage checks for definitions. The only behavioral difference is that `deleteDefinitionElements` silently ignores non-definition elements in the input, whereas `deleteElements` will attempt to delete everything it can."

docs/learning/backend/BulkElementDeletion.md
```diff
@@ -0,0 +1,98 @@
...
+- **`ElementDrivesElement`** — rows where either the source or target element is deleted.
+- **`ModelSelectorRefersToModels`** — rows where the target model (modeled element) is deleted.
+
+No manual cleanup of these tables is required.
+
+## Lifecycle callbacks
+
+`deleteElements` fires the standard lifecycle callbacks for every element in the final delete set (after constraint violators have been pruned):
+
+- `_OnDelete` — fired for every element that was originally in the input array.
```
hl662, line 82: The callback names here (`_OnDelete`, `_OnDeleted`, etc.) are the native C++ names — TypeScript consumers would be looking for `Element.onDelete`, `Element.onDeleted`, `Element.onChildDelete`, etc. Might be worth switching to the TS names and linking to the API docs?

docs/learning/backend/BulkElementDeletion.md
```diff
@@ -0,0 +1,98 @@
...
+> **Note:** Callbacks cannot veto a deletion in `deleteElements`. Any veto attempt from a callback is ignored. If you need per-element veto semantics, use [IModelDb.Elements.deleteElement]($backend) on individual elements instead.
+
+Use `deleteElements` when:
+
+- Deleting a tree of elements (parent + all descendants).
+- Deleting partitions/modeled elements (so their sub-models are cleaned up too).
+- Deleting a mix of ordinary and definition elements.
+
+Use `deleteElement` when:
+- You are deleting less than 10 elements at a time.
```
hl662, line 96: The "less than 10 elements" guidance feels a bit arbitrary — the real distinction seems to be about semantics (hard failure vs. silent exclusion, veto support) rather than element count. Maybe reframe around when you'd want per-element error handling instead of a threshold?

core/backend/src/EditTxn.ts
RohitPtnkr1996: Done!

core/backend/src/IModelDb.ts
RohitPtnkr1996: Done!
deleteElements API now exists only in EditTxn.ts.

core/backend/src/EditTxn.ts
RohitPtnkr1996: Done!

core/backend/src/EditTxn.ts
RohitPtnkr1996: Done.

core/backend/src/EditTxn.ts
RohitPtnkr1996: Done!

common/changes/@itwin/core-backend/rohitptnkr-bulk-element-deletion_2026-03-03-14-06.json
RohitPtnkr1996: Done

core/backend/src/test/element/DeleteDefinitionElements.test.ts
RohitPtnkr1996: Done!

core/backend/src/test/standalone/DeleteElements.test.ts
RohitPtnkr1996: Done!
Tests updated to check for failed elements.

docs/learning/backend/BulkElementDeletion.md
RohitPtnkr1996: Done!

docs/learning/backend/BulkElementDeletion.md
RohitPtnkr1996: Done!

docs/learning/backend/BulkElementDeletion.md
RohitPtnkr1996: Done!

docs/learning/backend/BulkElementDeletion.md
RohitPtnkr1996: Done!

PR #9058, opened 2026-03-04
Enable example code extraction in core-backend and move workspace examples

hl662:
## Why

Backend workspace and settings example code was still owned by `example-code/snippets`, even though the APIs and extraction point now belong to `core/backend`. That split made the documentation examples easier to drift from the package they describe and kept the normal backend test flow from proving those examples still worked where the APIs actually live.

The cloud-backed examples also need Azurite. Pulling them into `core/backend` without extra scoping would either broaden the default backend harness to require Azurite everywhere or leak those tests into other surfaces like the iOS webpack bundle. This keeps the examples package-local without pushing their runtime requirements onto unrelated test paths.

## What

| Asset | Why it exists |
|---|---|
| `core/backend/src/test/example-code/WorkspaceExamples.test.ts` (test) | Backend workspace and settings snippets need to live beside the package that owns the APIs so extraction and example coverage stop depending on `example-code/snippets`. |
| `core/backend/src/test/example-code/AzuriteTest.ts` (test helper) | The cloud workspace examples need a local blob-backed test helper so they stay executable without real Azure resources. |
| `core/backend/scripts/runWorkspaceExamples.mjs` (test runner) | Azurite-backed examples need a narrowly scoped runner so they execute in backend CI without re-running the full backend suite or inheriting unrelated Mocha config. |
| `core/backend/scripts/runAllTests.mjs` (test orchestrator) | `core/backend` test runs need to cover both the main backend suite and the migrated workspace examples every time the package is validated. |
| `core/backend/package.json` (package wiring) | The backend package needs to own extraction, test orchestration, and iOS test scoping now that these examples live under `core/backend`. |
| `common/config/azure-pipelines/templates/publish-test-results.yaml` (CI config) | The dedicated workspace-example run emits its own JUnit file, so CI needs to publish both backend result sets. |
| `tools/internal/ios/ios.webpack.config.js` (test bundling config) | Azurite-only example tests should stay out of the iOS webpack bundle so that surface does not inherit backend-only runtime requirements. |
| `common/config/rush/browser-approved-packages.json` and `common/config/rush/pnpm-lock.yaml` (dependency metadata) | Adding Azurite as a backend dev dependency updates the Rush-managed dependency metadata that keeps installs and browser approvals consistent. |

## Validation

- `cd core/backend && npm run -s build:cjs`
- `cd core/backend && npm run -s extract`
- `cd core/backend && npm run -s test:workspace-examples`
- `cd core/backend && npm run -s test`
- `cd core/backend && npm run -s ios:webpack:tests`

---
*Nambot 🤖*


hl662:
closing as unplanned

PR #9059, opened 2026-03-04
Fix reality data not being reprojected correctly when its CRS is different than iModel

eringram:
There is a bug where when using the reproject parameter of `TileTreeReference.createGeometryTreeReference`, if the reality data is not in the same CRS than the iModel, the reprojection will put the geometry at the wrong place.

This was because in `RealityTileLoader.loadGeometryFromStream`, the geometry returned by `RealityTileLoader.readGltfAndCreateGeometry` is in iModel coordinates, but the reprojection transform is in the reality tree's coordinates.

The fix is to take this difference into account by doing the following:
1. multiply polyface by inverse of tree's `iModelTransform` to "convert" polyface from iModel CRS->tile tree CRS
2. apply the original reprojection transform `xForm`
3. multiply by original `iModelTransform` to convert back from tile tree CRS->iModel CRS

Tested with a couple different iModels with reality data in different CRS. I also added some unit tests for this.

### DIFF
--- a/common/changes/@itwin/core-backend/hl662-fix-getContainerAsync-token-precedence_2026-03-09-15-26.json
+++ b/common/changes/@itwin/core-backend/hl662-fix-getContainerAsync-token-precedence_2026-03-09-15-26.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Fix operator precedence bug in WorkspaceImpl.getContainerAsync that silently discarded explicit access tokens",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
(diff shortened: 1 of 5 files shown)


aruniverse:
@mergifyio backport release/5.6.x release/5.7.x

markschlosseratbentley:
> Nice work! Do we need to run image tests?

@eringram I think we should just to be safe.

eringram:
No image tests differences of note.

The V4 tests had some missing reality data, but I confirmed in the logs it was related to flaky requests for the data and not due to the rendering. The same reality data looked correct in the V1 tests, and since it's reality data and not iModel geometry, the tiles are the same in both versions. (correct me if that doesn't sound right, @markschlosseratbentley ..?)

eringram:
> Does a NextVersion.md entry make sense?

Added

markschlosseratbentley:
> > Does a NextVersion.md entry make sense?
> 
> Added

Thanks. LGTM.

eringram:
Doing some more testing of this before merging

anmolshres98:
@eringram are you planning on getting this merged in today? I am waiting on this to release the next 5.6 patch release. No worries if you are not planning on merging today, just wanted to check :)

eringram:
@anmolshres98 Hey sorry for the delay, yes planning to merge today. I'll fix the conflicts to do that now.

PR #9060, opened 2026-03-05
Introduce `LockService` interface to `BriefcaseConnection`

GytisCepk:
## Motivation

Tools and other UI elements sometimes need to know whether element is locked and who owns the lock. Tools may query this information frequently and expect a quick response (e.g. on element hover tools want to show that element is not available due to element being locked by other user).

Some generic core level API is needed to make sure that solution is reachable from any package.

## Solution

Have a locks cache available via `BriefcaseConnection`.

Cache would prefetch locks from iModelHub and regularly update them. Cache would be abstracted away via `LockService` interface and implementation would be provided by applications for now (later down the line, some generic default implementation could be available in core packages).

Benefits of having this available via `BriefcaseConnection`:
- Easily accessible to tools and UI components, regardless of the package they're in
- Easier to ensure that only one cache exists per iModel since multiple caches would create unnecessary network traffic.


core/frontend/src/BriefcaseConnection.ts
```diff
@@ -277,9 +310,19 @@ export class BriefcaseConnection extends IModelConnection {
...
+
+  /** Open a BriefcaseConnection to a [BriefcaseDb]($backend).
+   * @alpha
+   */
+  public static async openFile(briefcaseProps: OpenBriefcaseProps, lockServiceFactory?: LockServiceFactory): Promise<BriefcaseConnection>
+
+  public static async openFile(briefcaseProps: OpenBriefcaseProps, lockServiceFactory?: LockServiceFactory): Promise<BriefcaseConnection> {
     const iModelProps = await IpcApp.appFunctionIpc.openBriefcase(briefcaseProps);
     const connection = new this({ ...briefcaseProps, ...iModelProps }, briefcaseProps.readonly ? OpenMode.Readonly : OpenMode.ReadWrite);
+    if (lockServiceFactory)
```
ben-polinsky, line 323: Would be good to add a unit test for the optional locks assignment?

kring:
This looks reasonable to me. Do you have an implementation of `LockService` somewhere already? It might be interesting to see it, if so. Maybe it would even be useful for testing if we had an implementation for `LocalHub`? The foreign locks would be simulated, of course.

GytisCepk:
> Do you have an implementation of LockService somewhere already?

Partial one exists, but currently it's internal (you can find a link in linked issue).

> Maybe it would even be useful for testing if we had an implementation for LocalHub?

At some point, maybe. Currently, no one is using this so there is nothing to test. 

Also, interface is simple enough right now so writing a simple mock should not be difficult for anyone.

core/frontend/src/BriefcaseConnection.ts
```diff
@@ -191,6 +191,25 @@ class ModelChangeMonitor {
...
+ * @see [[BriefcaseConnection.locks]]
+ * @alpha
+ */
+export interface LockService {
+  /** Get all elements with exclusive locks owned by other briefcases. */
+  getExclusiveForeignLocks(): Promise<Id64Set>;
+  /** Get all elements with shared locks owned by other briefcases. */
+  getSharedForeignLocks(): Promise<Id64Set>;
+  /** Check whether it's possible to acquire a lock for the element. */
+  checkElementLockAvailability(elementId: Id64String, lock: LockState): Promise<boolean>;
```
rschili, line 205: Should we include information on WHO is holding those locks here?

core/frontend/src/BriefcaseConnection.ts
```diff
@@ -191,6 +191,25 @@ class ModelChangeMonitor {
   }
 }
 
+/**
+ * Provides access to lock information in the iModel.
+ * @see [[BriefcaseConnection.locks]]
+ * @alpha
+ */
+export interface LockService {
```
rschili, line 199: Who is going to implement LockService? Seems like something that should go into core as well, no?

GytisCepk: Currently, consumers. 

However, once we have reusable implementation internally, we'll make it public via one of the core packages.

core/frontend/src/BriefcaseConnection.ts
GytisCepk: You're right, knowing who blocks you by owning certain lock is important.

That being said, I want to first prove viability of `BriefcaseConnection.locks` before expanding it further, since I have a few technical concerns regarding this:
- Locks are not directly associated with user, locks are associated with briefcase and briefcase is associated with user. That means we have to cache more and send more request to the iModelHub.
- What if the reason for failing to acquire lock is not that someone has lock on that element, but rather that someone has an exclusive lock on parent or maybe even root model ? A more descriptive message might be needed instead of just lock owner. I'll admit, I haven't fully figured out how all this will work.

We will need to add this eventually, but currently, consumers have other UI components to show lock owners.

core/frontend/src/BriefcaseConnection.ts
GytisCepk: Added

PR #9063, opened 2026-03-05
pin KaTeX for docs builds until fix is released

ben-polinsky:
Unblock our docs builds until KaTeX releases a [fix for blunder](https://github.com/KaTeX/KaTeX/issues/4152).


ben-polinsky:
It's out! Nice work KaTeX devs.

PR #9065, opened 2026-03-05
docs: Rewrite Settings and Workspaces documentation

hl662:
## Rewrite and restructure Workspace & Settings learning docs

Rewrites the three linked learning docs (`WorkspacesAndSettings.md`, `Workspace.md`, `Settings.md`) and adds a one-page tutorial (`WorkspaceTutorial.md`) so the whole settings/workspace flow is simpler, beginner-friendly, and factually accurate.

### Goals
- Explain **why** workspaces and settings exist before diving into how
- Give newcomers a clear path from concept → code
- Fix factual errors and outdated API references
- Add an end-to-end, user-friendly walkthrough of the full flow

### Key changes
- **WorkspacesAndSettings.md** — trimmed verbose sections, replaced sequence diagram with concise prose, and clarified the read/discovery vs write/create split for iTwin settings
- **Workspace.md** — fixed incorrect claim that `iModel.workspace` auto-includes iTwin settings, added scope-bridging guidance
- **Settings.md** — replaced deprecated `IModelDb.saveSettingDictionary`/`deleteSettingDictionary` guidance with `EditTxn` equivalents, clarified session-only vs persisted dictionaries, added `close()` notes, and documented the default single-container iTwin settings workflow
- **WorkspaceTutorial.md** — added a one-page walkthrough that covers the end-to-end settings/workspace flow in a beginner-friendly format
- **Code examples** — fixed wrong comments, added missing `close()` calls inside extract boundaries, restored accidentally removed extract, and tightened the advanced settings-container guidance
- **TSDoc** — updated `Settings` class-level doc and `addDictionary` to reference current APIs

## Validation
Targeted verification:
- Reviewed the affected docs/snippets and API comments for consistency with the current `getITwinWorkspace` / `saveSettingDictionary` behavior
- Fixed the invalid `SettingsPriority.*` links in the tutorial and verified the remaining touched markdown passes `git diff --check`

Known baseline issues:
- None noted during this doc-only pass


docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,150 @@
...
+
+At runtime, settings and resources are accessed through one of three workspace scopes:
+
+| Workspace | Scope | Access |
+|---|---|---|
+| [IModelHost.appWorkspace]($backend) | Application-wide defaults and configuration | Available immediately after [IModelHost.startup]($backend) |
+| [IModelHost.getITwinWorkspace]($backend) | iTwin-scoped settings shared across all iModels in an iTwin | Requires an iTwinId; no iModel needed |
+| [IModelDb.workspace]($backend) | iModel-specific overrides, inherits from app and iTwin scopes | Available when an iModel is open |
+
+Each scope layers on top of the previous one through the [Settings priority stack](./Settings.md#settings-priorities). See [Choosing the right workspace](./Workspace.md#choosing-the-right-workspace) for guidance on when to use each scope.
```
johnnyd710: Not implicitly true. If the iModel has a pointer to an iTwin settings db, then it will use that, but it doesn't happen automatically.

`IModelDb.workspace` will fallback to `IModelHost.appWorkspace`
`IModelHost.getITwinWorkspace` will fallback to `IModelHost.getITwinWorkspace`

but `IModelDb.workspace` will not fallback to `IModelHost.getITwinWorkspace`. Arguably it should? But I didn't want to mess with fetching an async workspace when opening an iModel, that sounded wrong to me.

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,150 @@
...
+|---|---|---|
+| **Purpose** | Application configuration | Data resources |
+| **Content** | JSON key-value dictionaries | Strings, blobs, embedded files |
+| **Container type** | `"settings"` | `"workspace"` |
+| **Discovery** | Via `SettingsEditor.queryContainers()` — no iModel needed | Referenced from settings values |
+| **Resolution order** | Loaded first | Loaded second, via settings pointers |
+| **Write API** | [SettingsEditor]($backend) | [WorkspaceEditor]($backend) |
+| **Versioning** | Semver — immutable once published | Semver — immutable once published |
+
+## How settings and resources connect
```
johnnyd710, line 20: I would consider omitting this section. What benefit to the user in knowing about SettingsDbs?

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,405 @@
...
+
+What about "landscapePro/ui/availableTools"? In the LandscapePro schema, the corresponding `settingDef` has [SettingSchema.combineArray]($backend) set to `true`, meaning that when multiple dictionaries provide a value for the setting, they are merged into a single array, eliminating duplicates, and sorted in descending order by dictionary priority.
+
+### How settings are stored
+
+The sections below cover three progressively richer ways to work with settings:
+
+- **Session-only (in-memory):** [Settings.addDictionary]($backend) loads a dictionary for the current session. Good for testing, app defaults, and short-lived overrides.
+- **iTwin / iModel persisted:** [IModelHost.saveSettingDictionary]($backend) writes named dictionaries to cloud storage scoped to an iTwin or iModel. Changes persist across sessions and are shared with other users.
+- **SettingsDb (cloud-managed storage):** A [SettingsDb]($backend) stores individual settings as key-value pairs in a dedicated container with its own versioning and access control — see [SettingsDb](#settingsdb).
```
johnnyd710: I would hesitate to even mention SettingsDbs here. Every SettingsDb is linked to an iTwin anyways - how could you create a SettingsDb without implying its not iTwin persisted?

iModel-scoped SettingsDbs should be exposed in a similar manner - `IModelDb.saveSettingsDictionary` should be the only way to do that

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,405 @@
...
+
+## iTwin settings
+
+Each iTwin has its own workspace with its own [Settings]($backend) that can override and/or supplement the application workspace's settings. These settings are stored as named [SettingsDictionary]($backend)s in a [SettingsDb]($backend) (a settings-formatted [WorkspaceDb]($backend)) scoped to that iTwin. Whenever [IModelHost.getITwinWorkspace]($backend) is called, all named dictionaries in the container are loaded into the returned [Workspace.settings]($backend) at [SettingsPriority.iTwin]($backend). An application working in the context of a particular iTwin should resolve setting values by asking [IModelHost.getITwinWorkspace]($backend), which will fall back to [IModelHost.appWorkspace]($backend) if the iTwin's setting dictionaries don't provide a value for the requested setting.
```
johnnyd710: this is the correct way to talk about settingsDbs imo. In the context of iTwin Settings - almost an implementation detail

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,405 @@
...
+
+```ts
+[[include:WorkspaceExamples.ReadITwinSettings]]
+```
+
+iTwin settings are loaded with [SettingsPriority.iTwin]($backend) priority.
+
+## iModel settings
+
+> **⚠️ Caution — prefer SettingsDb for new work.** Storing settings directly in an iModel (via [IModelDb.saveSettingDictionary]($backend)) ties configuration to a single iModel file. Settings cannot be discovered without opening the iModel, cannot be versioned independently, and require an exclusive write lock on the entire iModel to modify. For new projects, store settings in a cloud-hosted [SettingsDb](#settingsdb) instead — it is discoverable by iTwinId, versioned independently, and does not require an iModel to be open.
```
johnnyd710: "require an exclusive write lock on the entire iModel to modify" - I keep seeing this but where do we actually enforce that? I don't think this is true.

johnnyd710: I wouldn't caution against iModel settings - there are legitimate use-cases for them and the only option is to store them in the iModel until we modify saveSettingsDictionary to *not* do that

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,405 @@
...
+
+Each [IModelDb]($backend) has its own [Workspace]($backend), accessible via [IModelDb.workspace]($backend). This workspace inherits all app-level and iTwin-level settings, and layers on settings stored inside the iModel itself. Because these iModel-level dictionaries are loaded at [SettingsPriority.iModel]($backend) — the highest built-in priority — they override any lower-priority setting with the same name.
```
johnnyd710: The workspace does not inherit iTwin-level settings by-default today

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,405 @@
...
+
+### Pinning iTwin settings versions
+
+If you need to pin the iModel to a specific version of the iTwin settings — so that its configuration does not change even when the iTwin settings are updated — save the version alongside the container props:
+
+```ts
+[[include:WorkspaceExamples.VersionAndPinITwinSettings]]
+```
+
+## SettingsDb
```
johnnyd710: I wouldn't add this to the docs. Imo, a SettingsDb is the underlying persistence layer (and the first layer) to *any* workspace (today is just for the iTwin workspace). But why does a user need to know that? I don't think we want users creating SettingDbs at will (remember, there can only be *one* per iTwin and so its important its created correctly).

core/backend/src/workspace/Settings.ts
```diff
@@ -256,6 +256,8 @@ export interface Settings {
 
   /** Add a new [[SettingsDictionary]] with the priority, name, and [[WorkspaceDb]] specified by `props` and setting values supplied by `settings`.
    * @note If a dictionary with the same name and [[WorkspaceDb]] already exists, it will be replaced.
+   * @note Values added via this method exist only for the current session. They are not persisted and will be lost when the iModel is closed.
+   * To persist settings across sessions, use [[IModelDb.saveSettingDictionary]].
```
johnnyd710: Settings.addDictionary is not just for iModel settings, its for all setting scopes, so its inaccurate to say "To persist settings across sessions, use [[IModelDb.saveSettingDictionary]]"

core/backend/src/IModelDb.ts
```diff
@@ -2259,6 +2259,11 @@ export abstract class IModelDb extends IModel {
    * @param name The name for the SettingDictionary. If a dictionary by that name already exists in the iModel, its value is replaced.
    * @param dict The SettingDictionary object to stringify and save.
    * @note All saved `SettingDictionary`s are loaded into [[workspace.settings]] every time an iModel is opened.
+   * @note **Prefer [[SettingsDb]] for new work.** Storing settings directly in an iModel couples configuration to a single iModel file:
```
johnnyd710: I don't think its fair to say prefer `SettingsDb`. There are valid use-cases for this and we don't have an alternative.

What I was eventually going to do, if we really need to, was change saveSettingsDictionary to also use a SettingsDb under the hood. But the user would still call this method.

core/backend/src/IModelDb.ts
```diff
@@ -2267,6 +2272,7 @@ export abstract class IModelDb extends IModel {
 
   /** Delete a SettingDictionary, previously added with [[saveSettingDictionary]], from this iModel.
    * @param name The name of the dictionary to delete.
+   * @note **Prefer [[SettingsDb]] for new work.** See the caution on [[saveSettingDictionary]] for details.
```
johnnyd710: Same here

docs/learning/backend/WorkspacesAndSettings.md
hl662: Fixed — the table now says `IModelDb.workspace` falls back to `appWorkspace` only. I also added a note in Workspace.md ("Accessing resources") explaining that `IModelDb.workspace` does **not** auto-inherit iTwin settings, and pointing to the "Referencing iTwin settings from an iModel" section in Settings.md for the bridging pattern.

docs/learning/backend/WorkspacesAndSettings.md
hl662: Agreed — trimmed the section down. The overview doc no longer goes into SettingsDb internals. It just describes the two container types (settings vs workspace) and how discovery works via `getITwinWorkspace`.

docs/learning/backend/Settings.md
hl662: Makes sense. The Settings.md doc now treats SettingsDb as an implementation detail — it's only mentioned in the "Settings containers" section as context for how `WorkspaceEditor` works, not as something users create directly. The primary path leads with `getITwinWorkspace`.

docs/learning/backend/Settings.md
hl662: 👍 That's the framing I've adopted throughout — SettingsDb is mentioned only as the underlying persistence layer, not as a user-facing concept.

docs/learning/backend/Settings.md
hl662: Good catch — I softened this. The current text says the lock is "obtained automatically by `withEditTxn`" rather than claiming it's enforced at the settings level. If that's still overstating it, happy to remove the lock mention entirely and just say ordinary users shouldn't do this. What do you think?

hl662: Fair point. I've reframed the note — it now describes the tradeoff (tied to iModel file, not discoverable without opening it, not independently versioned) without steering users away. The tone is informational rather than cautionary.

docs/learning/backend/Settings.md
hl662: Fixed in both Settings.md and Workspace.md. The docs now explicitly state that `IModelDb.workspace` does not auto-include iTwin settings, and the "Referencing iTwin settings from an iModel" section shows the manual bridging pattern.

docs/learning/backend/Settings.md
hl662: Agreed — removed the standalone SettingsDb creation guidance. The "Settings containers" section now only mentions `WorkspaceEditor` in the context of admin tooling, with a callout that `saveSettingDictionary` handles the single-container constraint automatically.

core/backend/src/workspace/Settings.ts
hl662: Removed that line from the TSDoc. The `addDictionary` doc no longer mentions `saveSettingDictionary` at all — it just describes what `addDictionary` does (session-scoped, all priority levels).

core/backend/src/IModelDb.ts
hl662: Removed the "Prefer SettingsDb" note from `saveSettingDictionary`'s TSDoc. The method stands on its own now without steering users elsewhere.

core/backend/src/IModelDb.ts
hl662: Same fix — removed the "Prefer SettingsDb" note from `deleteSettingDictionary` as well.

hl662: one sec the change disappeared from my staged commits @@
Edit: Pushed

docs/learning/frontend/Preferences.md
```diff
@@ -2,6 +2,6 @@
 
 User preferences allow applications to persist settings across a user's sessions, where the settings can be customized by the user of the application. The preferences may be applicable for the application itself or dependent on a specific iTwin or iModel.
 
-User preferences are separate from any iTwin or iModel configuration intended to be shared across multiple users and administered by an admin rather than each individual user. See [Workspace](../backend/Workspace.md) for more details on shared workspace and settings.
+User preferences are separate from any iTwin or iModel configuration intended to be shared across multiple users and administered by an admin rather than each individual user. See [Workspaces and Settings](../backend/WorkspacesAndSettings.md) for an overview, [Settings](../backend/Settings.md) for configuration details, and [Workspace resources](../backend/Workspace.md) for binary data assets.
```
hl662, line 5: binary data assets.... what would we rather say? data?  Non BIS-aligned data?

docs/changehistory/5.8.0.md
```diff
@@ -100,7 +100,7 @@ Previously, settings and binary resources (fonts, textures, templates) were stor
 
 ##### Creating a local SettingsDb
 
-See [SettingsDb]($docs/learning/backend/Workspace.md#settingsdb) for full documentation.
```
khanaffan, line 103: Isn't change history is immutable?

hl662: I believe it would be a bad link after the changes in this PR, so the docs build would fail...
@ben-polinsky is that right

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+
+At runtime, settings and resources are accessed through one of three workspace scopes:
+
+| Workspace | Scope | Access |
+|---|---|---|
+| [IModelHost.appWorkspace]($backend) | Application-wide defaults and configuration | Available immediately after [IModelHost.startup]($backend) |
+| [IModelHost.getITwinWorkspace]($backend) | iTwin-scoped settings shared across all iModels in an iTwin | Requires an iTwinId; no iModel needed |
+| [IModelDb.workspace]($backend) | iModel-specific overrides; falls back to `appWorkspace` for unresolved settings. Does **not** automatically include iTwin-scoped settings | Available when an iModel is open |
+
+Each scope layers on top of the previous one through the [Settings priority stack](./Settings.md#settings-priorities). See [Choosing the right workspace](./Workspace.md#choosing-the-right-workspace) for guidance on when to use each scope.
```
tcobbs-bentley: I feel that the wording here is now questionable at best, since it is now in conflict with the updated Scope entry in the table.

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+> For advanced scenarios (admin tooling, custom discovery logic), [WorkspaceEditor.queryContainers]($backend) provides lower-level access to enumerate containers by iTwinId and `containerType`.
+
+## Scope and priority
+
+Settings from multiple sources are merged using a priority stack. A higher-priority dictionary overrides a lower-priority one for any given setting name.
+
+```mermaid
+graph TD
+    subgraph AppWorkspace["IModelHost.appWorkspace"]
+        D["defaults (100)"]
```
tcobbs-bentley: The text here needs more contrast.

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+        A["application (200)"]
+        O["organization/iTwin (400)"]
+    end
+
+    subgraph IModelWorkspace["IModelDb.workspace"]
+        B["branch (500)"]
+        M["iModel (600)"]
```
tcobbs-bentley: The text here needs more contrast.

docs/learning/backend/WorkspacesAndSettings.md
hl662: I'm confused, is the text not showing up? It looks fine on my end: 
image

docs/learning/backend/WorkspacesAndSettings.md
hl662: same here: 
image

docs/learning/backend/WorkspacesAndSettings.md
hl662: Gotcha, updated this small paragraph to mention the difference and how iModel level workspaces would not automatically include iTwin level settings

docs/learning/backend/WorkspacesAndSettings.md
tcobbs-bentley: Turns out this is a dark mode problem. I switched to light mode before commenting, but it turns out that mermaid doesn't regenerate the image when that happens, so it was still showing me the (broken) dark mode version. Here's what it looks like in dark mode:
image

tcobbs-bentley: Note that if the official docs site doesn't allow dark mode, then this problem can probably be ignored.

docs/changehistory/5.8.0.md
ben-polinsky: We do not link check changehistory for this reason. Well, not for this exact reason - we do it because we don't have version-aware link checking. In this case, the change is for our current version, so it makes sense. But are you replacing it with something else?

hl662: The `#settingsdb` anchor no longer exists in Settings.md after this PR (we removed the standalone SettingsDb section since we're treating it as an implementation detail now). Leaving it as-is would result in a dead link. Updated it to point to `#settings-containers` which covers the same concept. Fixed in f895423.

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
+# Workspaces and Settings
+
+Every non-trivial iTwin.js application needs two things at run-time: **configuration** (which tools are available, what units to use, which data sources are active) and **resources** (binary assets like fonts, textures, and templates). iTwin.js provides two complementary systems to address these needs:
+
+- **[Settings]($backend)** — a priority-ordered stack of key-value configuration pairs. Configuration flows from cloud-hosted settings containers into the active [Settings]($backend) runtime, where values can be read by name.
+- **[Workspace resources](./Workspace.md)** — versioned binary assets stored in [WorkspaceDb]($backend) containers. Settings tell the application *which* `WorkspaceDb`s to load; the application then retrieves resources from them.
+
+These two systems are deliberately separate. Settings containers are discoverable without opening an iModel. `WorkspaceDb` containers are referenced *by* settings. This separation eliminates the circular dependency that would arise if settings had to be loaded from a `WorkspaceDb` just to discover which `WorkspaceDb` to open.
```
johnnyd710: > Settings containers are discoverable without opening an iModel.

Technically so are workspace containers.

> This separation eliminates the circular dependency that would arise if settings had to be loaded from a `WorkspaceDb` just to discover which `WorkspaceDb` to open.

Yeah, if we stored everything in workspace containers, we wouldn't know which workspace container to start with, that's the key point

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+
+At runtime, settings and resources are accessed through one of three workspace scopes:
+
+| Workspace | Scope | Access |
+|---|---|---|
+| [IModelHost.appWorkspace]($backend) | Application-wide defaults and configuration | Available immediately after [IModelHost.startup]($backend) |
+| [IModelHost.getITwinWorkspace]($backend) | iTwin-scoped settings shared across all iModels in an iTwin | Requires an iTwinId; no iModel needed |
```
johnnyd710: getITwinWorkspace falls back to `appWorkspace` too

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+| [IModelDb.workspace]($backend) | iModel-specific overrides; falls back to `appWorkspace` for unresolved settings. Does **not** automatically include iTwin-scoped settings | Available when an iModel is open |
+
+`appWorkspace` and `IModelDb.workspace` share the same priority stack, so iModel-level settings automatically override application defaults. `getITwinWorkspace` is independent — its settings are only available to an iModel if explicitly referenced (see [Referencing iTwin settings from an iModel](./Settings.md#referencing-itwin-settings-from-an-imodel)). See [Choosing the right workspace](./Workspace.md#choosing-the-right-workspace) for guidance on when to use each scope.
```
johnnyd710: I don't follow this. iTwin-level settings override application defaults too.

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+
+## The two container types
```
johnnyd710: I don't think this is worth going into detail.

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+        SDB["<b>Settings</b><br/>Key-value config<br/>JSON dictionaries<br/>Priority-based merge"]
+    end
+
+    subgraph WorkspaceContainer["Workspace container (containerType: &quot;workspace&quot;)"]
+        WDB["<b>Resources</b><br/>Named resources<br/>Strings, blobs, files<br/>On-demand lookup"]
+    end
+
+    CloudSqlite --- SettingsContainer
+    CloudSqlite --- WorkspaceContainer
+    SDB -->|"settings point to"| WDB
```
johnnyd710: I find this diagram hard to understand, specifically the relationship between CloudSqlite and the containers. CloudSqlite is the mechanism for which dbs are stored *inside* containers, not beside them?

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+| **Discovery** | Automatic via [IModelHost.getITwinWorkspace]($backend) — no iModel needed | Referenced from settings values |
+| **Resolution order** | Loaded first | Loaded second, via settings pointers |
+| **Write API** | [WorkspaceEditor]($backend) with `containerType: "settings"` | [WorkspaceEditor]($backend) |
+| **Versioning** | Semver — immutable once published | Semver — immutable once published |
+
+## How settings and resources connect
+
+At runtime the flow always starts from settings:
+
+1. **Discover and load** — call [IModelHost.getITwinWorkspace]($backend) with an iTwinId. This automatically discovers settings containers for the iTwin, loads their dictionaries into the [Settings priority stack](./Settings.md#settings-priorities), and returns a [Workspace]($backend) ready to use. No iModel is required.
```
johnnyd710: why mention `IModelHost.getITwinWorkspace` here, but not `IModelDb.workspace` or `IModelHost.appWorkspace`? just say *the workspace*.

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+2. **Resolve resources** — read settings values that point to [WorkspaceDb]($backend) containers, then open those containers to access resources.
+
+> For advanced scenarios (admin tooling, custom discovery logic), [WorkspaceEditor.queryContainers]($backend) provides lower-level access to enumerate containers by iTwinId and `containerType`.
+
+## Scope and priority
+
+Settings from multiple sources are merged using a priority stack. A higher-priority dictionary overrides a lower-priority one for any given setting name.
+
+```mermaid
```
johnnyd710: nit: I'm not sure this diagram provides any value

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,110 @@
...
+In practice:
+- **Organization-wide defaults** are stored in a settings container and loaded at [SettingsPriority.iTwin]($backend) (400).
+- **iModel-specific overrides** are loaded at [SettingsPriority.iModel]($backend) (600) — iModel wins over iTwin.
+- **Application defaults** are loaded at [SettingsPriority.application]($backend) (200) — overrideable by any cloud-backed settings.
+
+> **Note:** The diagram above simplifies organization and iTwin into one level. The full priority stack includes a separate [SettingsPriority.organization]($backend) (300) level — see [Settings priorities](./Settings.md#settings-priorities) for details.
+
+`IModelHost.appWorkspace` holds dictionaries at `application` priority or lower. `IModelDb.workspace` holds higher-priority dictionaries and falls back to `appWorkspace` when a setting is not found.
+
+## Container discovery
```
johnnyd710: I also don't think this is worth mentioning.

johnnyd710 (commented):
I found the documentation was easier to understand and flowed better when there was one document that went through a single user-story (LandscapePro). It only brought technical details when necessary. In general I find examples easier than concepts/diagrams but maybe that's just me

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,418 @@
+# Settings
+
+[Settings]($backend) are how administrators configure an iTwin.js application for end-users. A setting is a named value — a `string`, `number`, `boolean`, object, or array — that controls application behavior at run-time.
+
+> **New to this topic?** Start with the [Workspaces and Settings overview](./WorkspacesAndSettings.md) to understand how [Settings]($backend), settings containers, and [WorkspaceDb]($backend) relate before diving in here.
```
ben-polinsky, line 5: nice :)

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,418 @@
...
+
+```
+  "landscapePro/ui/defaultToolId"
+  "landscapePro/ui/availableTools"
+  "landscapePro/flora/preferredStyle"
+  "landscapePro/flora/treeDbs"
+  "landscapePro/hardinessRange"
+```
+
+Each name begins with the "landscapePro" schema prefix followed by a forward slash. Forward slashes create logical groupings, similar to file paths.
```
ben-polinsky:
```suggestion
Forward slashes create logical groupings, similar to file paths.
```

docs/learning/backend/WorkspacesAndSettings.md
hl662: i removed the mermaid diagrams so this isn't relevant to us anymore

docs/learning/backend/WorkspacesAndSettings.md
hl662: removed the mermaid diagrams

docs/learning/backend/WorkspacesAndSettings.md
hl662: Good points — removed the bootstrapping explanation and the discoverability framing entirely. The overview now just states what settings and workspace resources are and links to the deep-dive pages.

docs/learning/backend/WorkspacesAndSettings.md
hl662: Fixed — updated the wording so the fallback chain is accurate. The section no longer singles out specific API entry points.

docs/learning/backend/WorkspacesAndSettings.md
hl662: You're right — rewrote the scope/priority paragraph. It now says all three scopes share the same priority stack (iModel > iTwin > app defaults) without implying only iModel settings override.

docs/learning/backend/WorkspacesAndSettings.md
hl662: Agreed — removed the entire "Two container types" section (diagram, comparison table, ~30 lines). The overview no longer goes into container internals.

docs/learning/backend/WorkspacesAndSettings.md
hl662: Removed — both mermaid diagrams are gone from the page.

docs/learning/backend/WorkspacesAndSettings.md
hl662: Good call — replaced the specific API mention with generic language ("the workspace"). The deep-dive pages cover the actual entry points.

docs/learning/backend/WorkspacesAndSettings.md
hl662: Removed along with the other diagram.

docs/learning/backend/WorkspacesAndSettings.md
hl662: Agreed — removed the entire container discovery section.

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,42 @@
...
+
+| Workspace | Scope | Access |
+|---|---|---|
+| [IModelHost.appWorkspace]($backend) | Application-wide defaults and configuration | Available immediately after [IModelHost.startup]($backend) |
+| [IModelHost.getITwinWorkspace]($backend) | iTwin-scoped settings shared across all iModels in an iTwin; falls back to `appWorkspace` | Requires an iTwinId; no iModel needed |
+| [IModelDb.workspace]($backend) | iModel-specific overrides; falls back to `appWorkspace` for unresolved settings. Does **not** automatically include iTwin-scoped settings | Available when an iModel is open |
+
+All three scopes use the same [Settings priority stack](./Settings.md#settings-priorities) — iTwin-level and iModel-level settings both override application defaults. `getITwinWorkspace` is independent from `IModelDb.workspace` — its settings are only available to an iModel if explicitly referenced (see [Referencing iTwin settings from an iModel](./Settings.md#referencing-itwin-settings-from-an-imodel)). See [Choosing the right workspace](./Workspace.md#choosing-the-right-workspace) for guidance on when to use each scope.
+
+## How settings and resources connect
```
johnnyd710: I would omit this section because you kind of already said this here:
> These two systems are deliberately separate. If everything were stored in workspace containers, there would be no way to know which container to open first. Settings containers solve this bootstrapping problem — they are loaded first and tell the application which workspace containers to open.

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,418 @@
...
+    end
+
+    D -->|"overridden by"| A
+    A -->|"overridden by"| O
+    O -->|"overridden by"| B
+    B -->|"overridden by"| M
+
+    style M fill:#d4edda,stroke:#28a745
+    style D fill:#f8f9fa,stroke:#6c757d
+```
```
johnnyd710: I would omit this mermaid diagram. Its problematic because it says iTwin settings are in IModelHost.appWorkspace which isn't true, but also I just don't see the value beyond what you already explained

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,418 @@
...
+
+Use iModel settings when a particular iModel needs configuration that differs from the rest of its iTwin, or when you want to persist metadata (like an iTwin settings container reference) inside the iModel so it is available in future sessions.
+
+### Persisted vs session-only dictionaries
+
+There are two ways to supply settings dictionaries to an iModel's workspace scope:
+
+| Method | Scope | Persistence |
+|--------|-------|-------------|
+| [Settings.addDictionary]($backend) | Current session only | Values exist in memory only and are lost when the iModel is closed. |
```
johnnyd710: you already mentioned how addDictionary is session-only in `## Settings dictionaries`, this seems repetitive

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,418 @@
...
+| [EditTxn.saveSettingDictionary]($backend) (via `withEditTxn`) | All future sessions | Values are written to the iModel's `be_Props` table and automatically reloaded every time the iModel is opened. |
+
+Use `addDictionary` for transient overrides — for example, to inject ephemeral configuration while a particular tool is active. Use `saveSettingDictionary` only if you need settings to persist as part of an iModel's permanent record and cannot use a settings container instead.
```
johnnyd710: same here

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,418 @@
...
+The dictionary name (e.g. `"landscapePro/iModelSettings"`) identifies the dictionary within the iModel. If a dictionary with that name already exists, it is replaced; otherwise a new one is created. You can save multiple dictionaries under different names.
+
+Note that modifying iModel settings requires an exclusive write lock on the entire iModel (obtained automatically by `withEditTxn`) — ordinary users should never do this.
+
+### Deleting iModel settings
+
+To remove an entire settings dictionary from an iModel, use [EditTxn.deleteSettingDictionary]($backend) (via `withEditTxn`):
+
+```ts
+await withEditTxn(iModel, async (txn) => txn.deleteSettingDictionary("landscapePro/iModelSettings"));
```
johnnyd710: how come this isn't in WorkspaceExamples?

example-code/snippets/src/backend/WorkspaceExamples.test.ts
```diff
@@ -554,17 +576,105 @@ describe("Workspace Examples", () => {
...
+      // the iTwin's settings are updated later.
+      const iTwinWorkspaceToPin = await IModelHost.getITwinWorkspace(iTwinId);
+      const floatingRefs = iTwinWorkspaceToPin.settingsSources;
+      assert(undefined !== floatingRefs);
+
+      // Resolve each settings source to its current version and bake that into the reference.
+      const sources = Array.isArray(floatingRefs) ? floatingRefs : [floatingRefs];
+      const pinnedRefs = await Promise.all(
+        sources.map(async (source) => {
+          const db = await iTwinWorkspaceToPin.getWorkspaceDb(source);
```
johnnyd710: you shouldn't have to get the workspaceDb here, can't you just return `{ ...source, version: "1.0.0" }`?

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,418 @@
...
+
+### Pinning iTwin settings versions
+
+If you need to pin the iModel to a specific version of the iTwin settings — so that its configuration does not change even when the iTwin settings are updated — save the version alongside the container props:
+
+```ts
+[[include:WorkspaceExamples.VersionAndPinITwinSettings]]
+```
+
+## Settings containers
```
johnnyd710, line 275: This makes me think we should remove the "1 settings container per iTwin" restriction if we are encouraging people to make their own settings containers with non-standard attributes (metadata, manifest, etc.) since there can only be one of them.

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,418 @@
...
+
+> **When do you need this section?** The [IModelHost.saveSettingDictionary]($backend) and [IModelHost.deleteSettingDictionary]($backend) APIs shown in [iTwin settings](#itwin-settings) are sufficient for most application code — they manage the underlying settings container automatically. The `WorkspaceEditor`-based workflow described here is for **administrators** who need fine-grained control: creating containers explicitly, versioning, inspecting individual resources, or updating specific settings without replacing an entire dictionary.
```
johnnyd710: I don't follow - what use-case would someone actually need to create a container themselves? Remember, getITwinWorkspace will create a settings container for you, and you can use it to inspect resources, update specific settings, etc.

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,418 @@
...
+
+```ts
+const settingsRef = iModel.workspace.settings.getSetting<WorkspaceDbSettingsProps>("landscapePro/itwinSettingsRef");
+if (settingsRef !== undefined) {
+  const iTwinWorkspace = await IModelHost.getITwinWorkspace(settingsRef);
+  // iTwinWorkspace.settings now contains the iTwin-level settings.
+  // ... use the settings, then close when finished:
+  iTwinWorkspace.close();
+}
+```
```
johnnyd710, line 263: should this be in workspace examples?

docs/learning/backend/WorkspacesAndSettings.md
hl662: Done — removed this section.

docs/learning/backend/Settings.md
hl662: Agreed, removed the diagram. The bullet list above it covers the same ground.

docs/learning/backend/Settings.md
hl662: Removed — already covered in the Settings dictionaries section.

docs/learning/backend/Settings.md
hl662: Removed along with the table above.

docs/learning/backend/Settings.md
hl662: Moved to an extract snippet: `WorkspaceExamples.DeleteIModelSettingDictionary`.

example-code/snippets/src/backend/WorkspaceExamples.test.ts
hl662: Good call — simplified to `{ ...source, version: "1.0.0" }`.

docs/learning/backend/Settings.md
hl662: The nearby inline snippet for loading settings from an iModel has been moved to an extract: `WorkspaceExamples.LoadITwinSettingsFromIModel`.

docs/learning/backend/Settings.md
hl662: I've moved the admin workflows (discover, create, update, version) into a separate "Advanced: managing settings containers" section, framed generically rather than iTwin-specific. The conceptual intro stays under "Settings containers" with a note that getITwinWorkspace handles the common case automatically.
 
 I think we should document that these APIs are generic, and if needed, can be used to create multiple settings containers. They might not be iTwin level settings containers in the cloud, but something else. Domain rules and constraints of 1 settings container per iTwin is a rule we shouldn't enforce across the whole concept of settings....

docs/learning/backend/Settings.md
hl662: I added an advanced section now has a clear callout that most apps don't need this, and points to 
 getITwinWorkspace. I do want to keep the part on discovering, updating, creating containers available, docs extensibility sake (and openness)

docs/learning/backend/Settings.md
johnnyd710: > They might not be iTwin level settings containers in the cloud, but something else. Domain rules and constraints of 1 settings container per iTwin is a rule we shouldn't enforce across the whole concept of settings....

The thing is, you always create a container associated to an iTwin, its a required part of the API. And we throw an error when calling `getITwinContainerId` if there is more than 1 settings container associated to the iTwin. We did it like that for a reason - given an iTwinId, I need to find the settings container, if there's more than one because some admin somewhere created a bunch, what do we do? Its the same problem as before... I suppose you could just say load all settings containers associated to an iTwin and treat them collectively as the "settings"... but the code does not work like that today.

Today, if a user follows this tutorial and creates a new settings container in an iTwinId after someone has already created a settings container using `getITwinWorkspace` in the same iTwin, `getITwinWorkspace` will no longer work because of the constraint.

core/backend/src/workspace/Settings.ts
```diff
@@ -207,7 +207,7 @@ export interface SettingsDictionaryProps extends SettingsDictionarySource {
  * [[SettingGroupSchema]]s) at this time.
  *
  * iModel-specific settings are stored in the iModel's property table and loaded into [[IModelDb.workspace]] when the iModel is first opened.
- * You can add and remove a [[SettingsDictionary]] from the property table using [[IModelDb.saveSettingDictionary]] and [[IModelDb.deleteSettingDictionary]].
+ * You can add and remove a [[SettingsDictionary]] from the property table using [[EditTxn.saveSettingDictionary]] and [[EditTxn.deleteSettingDictionary]].
```
ben-polinsky, line 210: Do you think this will be jarring for anyone who is used to `IModelDb.` access and will now have to use `EditTxn`. I suppose I am getting at some sort of note here about the previous, API.

ben-polinsky: Similar to the "New to this topic?" callout you add below.

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,352 @@
...
+
+### Schema validation behavior
+
+Validation occurs lazily when you **retrieve** a setting value — not when the value is stored. When you call methods like [Settings.getString]($backend), [Settings.getObject]($backend), or [Settings.getBoolean]($backend), the value is validated against the registered schema for that setting name:
+
+- If **no schema** is registered for the setting name, the value passes through unchecked.
+- If a schema **is** registered, type mismatches throw an error — for example, if the schema declares a setting to be a `string` but the dictionary supplies a `number`.
+- For `object` types, `required` fields are enforced and `extends` references are resolved recursively.
+
+Because validation happens on retrieval, register your schemas early — ideally via [IModelHost.onAfterStartup]($backend) — so that type errors are caught as soon as the setting is first accessed.
```
ben-polinsky, line 79: Slightly unclear why you wouldn't register in `iModelHost.onAfterStartup`. Several lines above, you tell users explicitly to "Register schemas shortly after invoking [IModelHost.startup]($backend), using the [IModelHost.onAfterStartup]($backend) event"

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Settings dictionaries
+
+The values of [Setting]($backend)s are provided by [SettingsDictionary]($backend)s. The [Settings]($backend) for the current session can be accessed via the `settings` property of [IModelHost.appWorkspace]($backend).
```
ben-polinsky:
```suggestion
Values of [Setting]($backend) are provided by [SettingsDictionary]($backend). The [Settings]($backend) for the current session can be accessed via the `settings` property of [IModelHost.appWorkspace]($backend).
```

I think it reads alright without the 's'? A pet peeve of mine.

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,352 @@
...
+
+### Settings priorities
+
+Configurations are often layered. [SettingsPriority]($backend) defines which dictionaries take precedence — the highest-priority dictionary that provides a value for a given setting name wins.
+
+Specific values:
+
+- [SettingsPriority.defaults]($backend) (100) — settings loaded automatically at startup.
+- [SettingsPriority.application]($backend) (200) — settings supplied by the application to override or supplement defaults.
+- [SettingsPriority.organization]($backend) (300) — settings applying to all iTwins in an organization.
```
ben-polinsky: "settings applied" here and below to match the tense: "settings supplied" and "settings loaded" above?

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,352 @@
...
+- [SettingsPriority.iTwin]($backend) (400) — settings applying to a particular iTwin and all its iModels.
+- [SettingsPriority.branch]($backend) (500) — settings applying to all branches of a particular iModel.
+- [SettingsPriority.iModel]($backend) (600) — settings applying to one specific iModel.
+
+[SettingsDictionary]($backend)s of `application` priority or lower reside in [IModelHost.appWorkspace]($backend). iTwin-scoped settings are loaded into the workspace returned by [IModelHost.getITwinWorkspace]($backend) — see [iTwin settings](#itwin-settings). Settings of even higher priority (branch and iModel) are stored in an [IModelDb.workspace]($backend) — see [iModel settings](#imodel-settings).
```
ben-polinsky:
```suggestion
A [SettingsDictionary]($backend) of `application` priority or lower resides in [IModelHost.appWorkspace]($backend). iTwin-scoped settings are loaded into the workspace returned by [IModelHost.getITwinWorkspace]($backend) — see [iTwin settings](#itwin-settings). Settings of even higher priority (branch and iModel) are stored in an [IModelDb.workspace]($backend) — see [iModel settings](#imodel-settings).
```

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,352 @@
+# Settings
```
ben-polinsky, line 1: Really thorough document here, @hl662, nice work. I would suggest giving users a bit more up front in the introductory paragraphs. Otherwise, readers won't get an idea of things like priority or the fact that there are different levels and locations of settings until they come across way down the page.

I wouldn't go into detail, but a sentence about how they operate at different levels of persistence and have different priorities could be helpful to know up front.

docs/learning/backend/Settings.md
```diff
@@ -0,0 +1,352 @@
...
+
+[Settings]($backend) are how administrators configure an iTwin.js application for end-users. A setting is a named value — a `string`, `number`, `boolean`, object, or array — that controls application behavior at run-time.
+
+> **New to this topic?** Start with the [Workspaces and Settings overview](./WorkspacesAndSettings.md) to understand how [Settings]($backend), settings containers, and [WorkspaceDb]($backend) relate before diving in here.
+
+## What is a setting?
+
+Be careful to avoid confusing **settings** with **user preferences**, which are configured by individual users. For example, an application might provide a check box to toggle "dark mode". Each individual user can make their own choice — it is a user preference, not a setting. But an administrator may define a setting that controls whether users can see that check box in the first place.
```
ben-polinsky: A bit weird that the first sentences in the "What is a setting?" section are about what a setting is _not_. I might move this paragraph down below, after settings are introduced a bit more. Or rephrase it, so you lead with the example dark mode checkbox, then mention the choice/value is a user preference.

docs/learning/backend/Settings.md
hl662: > Today, if a user follows this tutorial and creates a new settings container in an iTwinId after someone has already created a settings container using `getITwinWorkspace` in the same iTwin, `getITwinWorkspace` will no longer work because of the constraint.

Small clarification on the creation path: the first default settings container comes from the write side via [`IModelHost.saveSettingDictionary`](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/core/backend/src/IModelHost.ts#L493-L500), which goes through [`SettingsEditor.constructForITwin`](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/core/backend/src/internal/workspace/SettingsEditorImpl.ts#L53-L69) and falls through to [`createNewCloudContainer(..., containerType: "settings")`](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/core/backend/src/internal/workspace/WorkspaceImpl.ts#L520-L525) when no existing container is found. [`IModelHost.getITwinWorkspace(iTwinId)`](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/core/backend/src/IModelHost.ts#L458-L465) is the read/discovery side via [`SettingsContainers.getITwinSettingsSources`](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/core/backend/src/workspace/SettingsEditor.ts#L100-L123). So I think the “created using `getITwinWorkspace`” part is a little off, but your broader point about multiple settings containers breaking automatic selection still stands; that failure mode is already covered in [ITwinWorkspace.test.ts](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/core/backend/src/test/standalone/ITwinWorkspace.test.ts#L114-L121).

I’ve updated the docs/snippets to align with that. The overview now calls out the read/write split in [WorkspacesAndSettings.md](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/docs/learning/backend/WorkspacesAndSettings.md#L12-L18), the tutorial now says first creation happens on first write in [WorkspaceTutorial.md](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/docs/learning/backend/WorkspaceTutorial.md#L84-L99), and the reference/docs/snippet now make the default single-container assumption explicit in [Settings.md](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/docs/learning/backend/Settings.md#L157-L186), [Settings.md](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/docs/learning/backend/Settings.md#L280-L330), and [WorkspaceExamples.test.ts](https://github.com/iTwin/itwinjs-core/blob/nam/settings-workspace-docs/example-code/snippets/src/backend/WorkspaceExamples.test.ts#L637-L643).

docs/learning/backend/WorkspacesAndSettings.md
```diff
@@ -0,0 +1,38 @@
+# Workspaces and Settings
+
+Every non-trivial iTwin.js application needs two things at run-time: **configuration** (which tools are available, what units to use, which data sources are active) and **resources** (binary assets like fonts, textures, and templates). iTwin.js provides two complementary systems to address these needs:
```
ben-polinsky: Maybe nit: "Every non-trivial iTwin.js application needs two things" reads like these are the only two things an app would need :) 

Hmm struggling with a phrase than encompasses both resources and configuration. "needs to persist resources"? "Needs some application-level persistence"? I don't really like either of those...

core/backend/src/workspace/Settings.ts
hl662: Good point — I added a short migration note here calling out the deprecated `IModelDb.saveSettingDictionary` / `deleteSettingDictionary` path and pointing readers to `EditTxn.*`. Addressed in d92fc82919.

hl662: Yep — followed that pattern and added a small migration callout in the learning docs iModel settings section too. Addressed in d92fc82919.

docs/learning/backend/Settings.md
hl662: Good catch — I kept the recommendation to register in `IModelHost.onAfterStartup`, and added a note that some later extract snippets call `addGroup(...)` directly only for brevity. Addressed in d92fc82919.

docs/learning/backend/Settings.md
hl662: Agreed — I rewrote this rather than taking the suggestion verbatim. It now says a `SettingsDictionary` supplies values for one or more `Setting`s. Addressed in d92fc82919.

docs/learning/backend/Settings.md
hl662: Makes sense — updated these bullets to use `settings applied ...` for consistency with the surrounding phrasing. Addressed in d92fc82919.

docs/learning/backend/Settings.md
hl662: Agreed — switched this sentence to singular generic wording. Addressed in d92fc82919.

docs/learning/backend/Settings.md
hl662: Good suggestion — I added a short sentence up front that settings can live at different persistence levels and are resolved by priority, so readers get that framing earlier. Addressed in d92fc82919.

docs/learning/backend/Settings.md
hl662: Agreed — I rewrote that section to lead with what a setting does, then contrasted it with user preferences using the dark mode example. Addressed in d92fc82919.

docs/learning/backend/WorkspacesAndSettings.md
hl662: Fair nit — I softened that opener to `Two recurring run-time needs...` so it doesnt read as exhaustive. Addressed in d92fc82919.

ben-polinsky (approved):

johnnyd710 (approved):

PR #9069, opened 2026-03-09
Fix operator precedence in getContainerAsync token resolution

### LINKED ISSUE #9068: Bug: Operator precedence error in getContainerAsync silently discards access tokens
## Summary

`getContainerAsync` in `WorkspaceImpl.ts` has a JS operator precedence bug where `??` binds tighter than `?:`, causing explicitly-provided access tokens to be silently discarded.

## Affected Code

`core/backend/src/internal/workspace/WorkspaceImpl.ts` — two instances:

### Line ~367 (read path)
```typescript
const accessToken = props.accessToken ?? ((props.baseUri === "") || props.isPublic) ? "" : await CloudSqlite.requestToken({ ...props, accessLevel: "read" });
```

### Line ~515 (write path)
```typescript
const accessToken = props.accessToken ?? (props.baseUri === "") ? "" : await CloudSqlite.requestToken({ ...props, accessLevel: "write" });
```

## Root Cause

JS evaluates `??` (precedence 5) before `?:` (precedence 4):

```typescript
// Intended:
accessToken = props.accessToken ?? ((baseUri === "") ? "" : requestToken(...))

// Actual:
accessToken = (props.accessToken ?? (baseUri === "")) ? "" : requestToken(...)
```

When `props.accessToken` is a non-empty string (truthy), `??` returns it as the left operand. Then the ternary sees a truthy value and picks the `""` branch — silently discarding the real token.

## Impact

Callers passing an explicit `accessToken` via `WorkspaceContainerProps` will have it silently replaced with `""`, potentially causing auth failures against cloud containers. The bug may be masked in code paths that never pass explicit tokens.

## Fix

Wrap the ternary fallback in parentheses:

```typescript
// Line ~367
const accessToken = props.accessToken ?? (((props.baseUri === "") || props.isPublic) ? "" : await CloudSqlite.requestToken({ ...props, accessLevel: "read" }));

// Line ~515
const accessToken = props.accessToken ?? ((props.baseUri === "") ? "" : await CloudSqlite.requestToken({ ...props, accessLevel: "write" }));
```

hl662:
The `??` operator binds tighter than `?:`, causing the ternary to evaluate against the resolved `??` result rather than serving as the fallback. This silently replaced explicitly-provided access tokens with "".

Fixes #9068

### DIFF
--- a/common/changes/@itwin/core-backend/2026-03-09-17-56.json
+++ b/common/changes/@itwin/core-backend/2026-03-09-17-56.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-backend"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -27,8 +27,8 @@ importers:
         specifier: ^12.28.0
         version: 12.28.0
       '@bentley/imodeljs-native':
-        specifier: 5.8.10
-        version: 5.8.10
+        specifier: 5.8.11
+        version: 5.8.11
       '@itwin/object-storage-azure':
         specifier: ^3.0.4
         version: 3.0.4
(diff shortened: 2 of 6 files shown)


core/backend/src/test/standalone/Workspace.test.ts
```diff
@@ -191,4 +191,58 @@ describe("WorkspaceFile", () => {
...
+
+      // The container constructor will attempt to connect to the cloud URI, which will fail
+      // in a test environment. The key assertion is that requestToken is NOT called — the
+      // explicit token should be passed through without re-requesting.
+      try {
+        await workspace.getContainerAsync(props);
+      } catch (_e) {
+        // expected — no real cloud endpoint
+      }
+      expect(requestTokenStub.called).to.be.false;
```
hl662, line 212: I don't think this assertion actually protects the precedence regression.

In the buggy expression, a truthy `props.accessToken` still leads to `""` and `CloudSqlite.requestToken()` is never called, so `requestTokenStub.called === false` passes both before and after the fix.

Could we spy on `getContainer()` (or otherwise inspect the forwarded `accessToken`) so the test proves the explicit token survives, and ideally add the same proof for `EditorImpl.getContainerAsync` since that path changed too?

core/backend/src/internal/workspace/WorkspaceImpl.ts
```diff
@@ -500,7 +500,7 @@ class EditorImpl implements WorkspaceEditor {
     return this.workspace.findContainer(props.containerId) as EditableWorkspaceContainer | undefined ?? new EditorContainerImpl(this.workspace, props);
   }
   public async getContainerAsync(props: WorkspaceContainerProps): Promise<EditableWorkspaceContainer> {
-    const accessToken = props.accessToken ?? (props.baseUri === "") ? "" : await CloudSqlite.requestToken({ ...props, accessLevel: "write" });
+    const accessToken = props.accessToken ?? ((props.baseUri === "") ? "" : await CloudSqlite.requestToken({ ...props, accessLevel: "write" }));
```
ben-polinsky: Fix looks good, but do you think it might be cleaner to read if we break it up? One way would be to return `props.accessToken` early if it exists. Then move on to the fallback logic.

johnnyd710: I agree, breaking it up would make this easier to read. Keep it to no more than one condition on a single line?

PR #9071, opened 2026-03-09
Bugfix/default civilunits length koq

### LINKED ISSUE #9070: Default KOQ for CivilUnits.Length always points to USSurvey
**Describe the bug**
The default koq for `CivilUnits.LENGTH` points to `QuantityType.LengthSurvey` regardless of the active unit system. It will always return a formatter that outputs survey feet which is wrong.
Refer to https://github.com/iTwin/itwinjs-core/blob/7abfd08e282fc3ea5d129336bc0e7e60d952ad6f/core/frontend/src/quantity-formatting/QuantityFormatter.ts#L345

I believe @Maxime-Brassard already made @hl662 aware of this, and I ended up uncovering the problem on my own while looking at some other teams tests.

a-gagnon:
### Fix:
This PR fixes the default mapping of the `CivilUnits.LENGTH` KOQ to `QuantityType.LengthSurvey` and instead points to `QuantityType.LengthEngineering`.

Fixes #9070

### DIFF
--- a/common/changes/@itwin/core-frontend/bugfix-default-civilunits-length-koq_2026-03-09-18-52.json
+++ b/common/changes/@itwin/core-frontend/bugfix-default-civilunits-length-koq_2026-03-09-18-52.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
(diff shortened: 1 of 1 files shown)


a-gagnon:
@Mergifyio backport release/5.7.x

PR #9077, opened 2026-03-10
Add dedicated SettingsDb and SettingsEditor APIs

hl662:
This pull request introduces a new dedicated API for settings storage, separate from the existing `WorkspaceDb` functionality. The main addition is the `SettingsDb` and `SettingsEditor` APIs, which provide a streamlined interface for managing settings containers and dictionaries. The changes also include new interfaces, implementation files, and exports to support these features.

New APIs for dedicated JSON settings storage:

* Added `SettingsDb`, `SettingsDbManifest`, `SettingsDbProps`, and related interfaces for managing settings databases, dictionaries, and manifests. 
* Introduced `SettingsEditor` namespace and interface for constructing and managing settings containers, including methods for creating and retrieving editable settings containers. 
* Added `EditableSettingsCloudContainer` (extends from new base interface `CloudSqliteContainer`) and `EditableSettingsDb` interfaces for editing settings containers and dictionaries. 
* Implemented `SettingsDbImpl` in a new file (`core/backend/src/internal/workspace/SettingsDbImpl.ts`) to provide the internal logic for settings-only database access and dictionary management.
* Exported new modules (`SettingsDb`, `SettingsEditor`) in `core-backend.ts` for public API access.

Enhancements to workspace and cloud container APIs:

* Extended `WorkspaceContainer` to inherit from new base interface `CloudSqliteContainer`, and updated `WorkspaceDbCloudProps` to be an interface including `DbCloudContainerInfo`.
* Added `DbCloudContainerInfo` interface for standardized cloud container metadata. 
* Introduced new argument interfaces for creating and retrieving settings databases and containers (`CreateNewSettingsContainerArgs`, `CreateNewSettingsDbVersionArgs`, `CreateSettingsDbArgs`, `GetSettingsDbArgs`).


core/backend/tsconfig.tsbuildinfo
johnnyd710, line 1: accidental commit?

common/api/summary/core-backend.exports.csv
```diff
@@ -69,6 +69,7 @@ internal;interface;BcvStats
 internal;interface;BcvStatsFilterOptions
 internal;interface;LockAndOpenArgs
 internal;function;transferDb
+beta;interface;CloudSqliteContainer
```
johnnyd710, line 72: why export CLoudSqliteContainer?

common/api/summary/core-backend.exports.csv
```diff
@@ -97,8 +98,11 @@ beta;interface;CreateChangeSummaryArgs
 alpha;interface;CreateFontFileFromRscBlobArgs
 beta;interface;CreateFontFileFromShxBlobArgs
 public;interface;CreateNewIModelProps
+beta;interface;CreateNewSettingsContainerArgs
```
johnnyd710, line 101: nit: any reason for beta over alpha? not sure what the typical policy is

core/backend/src/internal/workspace/SettingsDbImpl.ts
```diff
@@ -0,0 +1,122 @@
...
+  public readonly [_implementationProhibited] = undefined;
+  public readonly sqliteDb = new WorkspaceSqliteDb();
+  public readonly dbName: string;
+  public readonly dbFileName: string;
+  protected readonly _container: CloudSqliteContainer;
+  public readonly priority: SettingsPriority;
+  protected _manifest?: SettingsDbManifest;
+
+  public constructor(props: SettingsDbProps, container: CloudSqliteContainer, priority: SettingsPriority) {
+    this.dbName = props.dbName;
```
johnnyd710: nit: `?? "settings-db"` the dbName default should be here, not in workspace getContainer

core/backend/src/internal/workspace/SettingsImpl.ts
```diff
@@ -183,3 +194,255 @@ export class SettingsImpl implements Settings {
...
+        return props;
+      }
+    }
+    return CloudAccess.initializeSettings(args);
+  }
+
+  public async createNewCloudContainer(args: CreateNewSettingsContainerArgs): Promise<EditableSettingsCloudContainer> {
+    const cloudContainer = await this.initializeContainer(args);
+    if (!IModelHost.authorizationClient)
+      throw new Error("IModelHost.authorizationClient must be configured to create cloud settings containers");
```
johnnyd710: nit: you had to of already used `IModelHost.authorizationClient` to get to this point (in createBlobContainer)

core/backend/src/internal/workspace/SettingsImpl.ts
```diff
@@ -183,3 +194,255 @@ export class SettingsImpl implements Settings {
...
+    if (existing)
+      return existing;
+    const baseContainer = this._workspace.getContainer(args);
+    const editable = new EditableSettingsContainerImpl(baseContainer);
+    this._containers.set(args.containerId, editable);
+    return editable;
+  }
+
+  public async getContainerAsync(props: WorkspaceContainerProps): Promise<EditableSettingsCloudContainer> {
+    const accessToken = props.accessToken ?? ((props.baseUri === "") ? "" : await CloudSqlite.requestToken({ ...props, accessLevel: "write" }));
```
johnnyd710: nit: like your other PR, multi-conditionals on one line like this are hard to read

docs/learning/backend/Workspace.md
```diff
@@ -231,3 +231,55 @@ If we configure the setting to use version 1.1.0, then `allTrees` will not inclu
 We could also configure the version more precisely using [semantic versioning](https://semver.org) rules to specify a range of acceptable versions. When compatible new versions of a `WorkspaceDb` are published, the workspace would automatically consume them without requiring any explicit changes to its [Settings]($backend).
 
 It may be tempting to "optimize" by calling `getAvailableTrees` once when your application starts up and caching the result to reuse throughout the session, but remember that the list of trees is determined by a setting, and settings can change at any time during the session. If you must cache, make sure you listen for the [Settings.onSettingsChanged]($backend) event to be notified when your cache may have become stale.
+
+## SettingsDb
```
johnnyd710: It would be better to modify the original story - adding a SettingsDb where it makes sense - than adding this at the end.

For example, everything before WorkspaceResources should use a SettingsDb (instead of just runtime dictionaries?)

docs/learning/backend/Workspace.md
```diff
@@ -231,3 +231,55 @@ If we configure the setting to use version 1.1.0, then `allTrees` will not inclu
...
+
+A [WorkspaceDb]($backend) is a general-purpose database that can store settings dictionaries alongside binary resources like fonts, textures, and images. But sometimes all you need is a place to store pure configuration - JSON key-value dictionaries with no binary resources at all. For that, iTwin.js provides [SettingsDb]($backend), a dedicated database type focused exclusively on storing [SettingsDictionary]($backend) values.
+
+### When to use SettingsDb vs WorkspaceDb
+
+Use [SettingsDb]($backend) when you only need to store and retrieve JSON settings dictionaries. A `SettingsDb` provides a focused, dictionary-only API surface, making it a natural fit for pure configuration data. Its containers use `containerType: "settings"` in their metadata, enabling them to be discovered independently - without opening an iModel.
```
johnnyd710: technically workspaceDbs can be discovered independently too, the difference is, settingsDbs are meant for settings so you know where to start

You always start by finding and loading settingsDbs

docs/learning/backend/Workspace.md
```diff
@@ -231,3 +231,55 @@ If we configure the setting to use version 1.1.0, then `allTrees` will not inclu
...
+
+Use [WorkspaceDb]($backend) when you need to store binary resources (images, fonts, templates), embedded files, or a mixture of resources and settings. A `WorkspaceDb` exposes a broader API for strings, blobs, and files, and is configured via settings that reference [WorkspaceDbProps]($backend).
+
+In short: if your data is all JSON configuration, prefer `SettingsDb`. If it includes binary content or mixed resource types, use `WorkspaceDb`.
```
johnnyd710: this is true, but to me, its more about knowing where to start. settingsDbs are where you start - settings point to resources, start with settings, find resources, etc.

core/backend/src/workspace/Workspace.ts
```diff
@@ -361,6 +362,14 @@ export interface Workspace {
   /** Get a single [[WorkspaceDb]].  */
   getWorkspaceDb(props: WorkspaceDbCloudProps): Promise<WorkspaceDb>;
 
+  /** Get a [[SettingsDb]] from a previously-loaded settings container in this workspace.
+   * @param args The arguments identifying which SettingsDb to retrieve.
+   * @returns The SettingsDb from the matching container.
+   * @throws if no container matching [[GetSettingsDbArgs.containerId]] has been loaded into this workspace, or if the container does not hold a SettingsDb.
+   * @note The container must already be loaded via [[getContainer]] or [[getContainerAsync]] before calling this method.
+   */
+  getSettingsDb(args: GetSettingsDbArgs): SettingsDb;
```
johnnyd710, line 375: nice to have: a find settings db method that does the blob container service query for me using the "settings" type

core/backend/src/internal/workspace/SettingsImpl.ts
```diff
@@ -183,3 +194,255 @@ export class SettingsImpl implements Settings {
...
+  return new SettingsEditorImpl();
+}
+
+class SettingsEditorImpl implements SettingsEditor {
+  public readonly [_implementationProhibited] = undefined;
+  private readonly _workspace: OwnedWorkspace;
+  private _containers = new Map<string, EditableSettingsContainerImpl>();
+
+  public constructor() {
+    this._workspace = constructWorkspace(new SettingsImpl(), { containerDir: join(IModelHost.cacheDir, settingsEditorName) });
```
johnnyd710: I think this is causing a bug. I'm getting `container is not writeable` when trying to write to the settings db.

```
Error: container is not writeable
    at EditableSettingsContainerImpl.acquireWriteLock (/Users/john.dimatteo/Repos/studio/node_modules/.pnpm/@itwin+core-backend@file+..+..+..+..+tmp+itwin-core-packs+itwin-core-backend-5.8.0-dev._e5e9366f916c7b24cdcd7df232a7c8ab/node_modules/@itwin/core-backend/lib/cjs/internal/workspace/SettingsImpl.js:304:33)
    at saveSettings (/Users/john.dimatteo/Repos/studio/packages/apis/workspace-utils/lib/index.js:211:18)
    at async BackendIpcHandlers.saveSettings (/Users/john.dimatteo/Repos/studio/packages/framework/lib/backend/settings/WorkspaceSettings.js:178:5)
    at async /Users/john.dimatteo/Repos/studio/packages/framework/lib/common/ipc/StudioIpcEvents.js:40:34
    at async fn (/Users/john.dimatteo/Repos/studio/packages/framework/lib/backend/ipc/StudioIpcHost.js:59:48)
```

SettingsEditor uses constructWorkspace, which uses the default Workspace cloud cache. Unlike WorkspaceEditor (which uses an editor-specific `cacheName`), SettingsEditor uses the same cache as the workspace, which allows container reuse across read and write paths for the same baseUri/containerId. That can surface as acquireWriteLock failing with container is not writeable if the container was first opened in read mode.

Can we give SettingsEditor an isolated cache namespace (for example a SettingsEditor-specific WorkspaceImpl/getCloudCache override), similar to WorkspaceEditor (look at EditorWorkspaceImpl)?

core/backend/tsconfig.tsbuildinfo
hl662: yep... removed

common/api/summary/core-backend.exports.csv
hl662: It needs to be exported — without it, users can't type containers returned by the workspace APIs and the SettingsDb.container property would reference an unexported type, which would fail `rush extract-api`.
As for making it possibly internal, our new interface `EditableSettingsCloudContainer` is marked beta, this extends from CloudSqliteContainer. If a beta interface extends from an internal interface we would also fail rush extract-api

common/api/summary/core-backend.exports.csv
hl662: Since the APIs are documented with tsdocs, we'll go with beta, thats our usual policy. We can still change the APIs even while it's in beta

core/backend/src/internal/workspace/SettingsDbImpl.ts
hl662: got it, updated

core/backend/src/internal/workspace/SettingsImpl.ts
hl662: kind of wanted to avoid having to deal with `no-non-null-assertions` lint error by having this, but it's okay, i did some optional null chaining instead

core/backend/src/internal/workspace/SettingsImpl.ts
hl662: got it, changed to multiple lines

docs/learning/backend/Workspace.md
hl662: got it, I changed the learning doc a bit, moved around the sections so SettingsDb will come before Workspaces. This puts settingsDb section right after section on settingsDictionaries.

With our separation of workspaces and settings being much clearer as a result of this PR, I'll split this learning doc up accordingly in a follow up PR too

common/api/summary/core-backend.exports.csv
```diff
@@ -185,6 +189,8 @@ public;class;ECSqlValueIterator
 deprecated;class;ECSqlValueIterator
 public;class;ECSqlWriteStatement
 beta;interface;EditableCatalogDb
+beta;interface;EditableSettingsContainer
```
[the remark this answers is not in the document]
hl662: just had to regen extract-api

core/backend/src/workspace/SettingsEditor.ts
```diff
@@ -0,0 +1,234 @@
...
+ * @note Settings containers are separate from workspace containers, providing independent write locks.
+ * Editing settings does not block workspace resource editing, and vice versa.
+ * @beta
+ */
+export interface EditableSettingsCloudContainer extends CloudSqliteContainer {
+  /**
+   * Create a copy of an existing [[SettingsDb]] in this container with a new [[WorkspaceDbVersion]].
+   * The copy should be modified with new content before the write lock is released,
+   * and thereafter may never be modified again.
+   * @param args - The properties that determine the source SettingsDb and the version increment to apply.
```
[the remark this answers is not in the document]
hl662: just had to regen extract-api

core/backend/src/workspace/Workspace.ts
hl662: Not too sure what you mean... using the blob container service involves getting containers based on the settings type. do you mean a find settingsContainer?

core/backend/src/internal/workspace/SettingsImpl.ts
hl662: sure, i've created a SettingsEditorWorkspaceImpl that has a different cache name, and the cache size is set to 2 GB. hopefully that's good for now

docs/learning/backend/Workspace.md
hl662: I added a section at the very start of 'SettingsDb', at line 140, that hopefully explains this well... but it'll be better once we split workspaces and settings into their own respective markdown files, and we can guide users to read settings first before workspaces

docs/learning/backend/Workspace.md
hl662: see comment above

core/backend/src/workspace/Workspace.ts
johnnyd710: Yeah I meant container. I just don't see why a consumer should have to do 
```ts
    const containers = await service.queryContainersMetadata(userToken, {
      containerType: "settings",
      iTwinId,
    });
```
Imo "settings" is an internal detail. Something like `SettingsEditor.getContainer({ iTwinId })` would save me a few steps, and I wouldn't have to read the docs to know about `containerType: "settings"`

hl662: Okay that makes more sense.
I added a findContainers() method to SettingsEditor, figured 'find' semantically makes more sense than 'get' because we're returning the metadata of the container. Look at this [example](https://github.com/iTwin/itwinjs-core/pull/9077/changes#diff-93a71664783ee643fbe9c1860faca1ec4142f2ba5f7befb6e04dcd7bf3df6bf9R612-R640) from a user standpoint, because the blob container service can return an array of containers, I followed that similarly. It's just a technical implementation under the hood that our own service is designed to be 1 container per iTwin, but if we change it up in the future at least we are prepared ahead of time. For now you can return the first entry in the array (if there exists one)

Edit: Ignore the first sentence, findContainers does return the settings container instance itself. It's just we already have a getContainer method in SettingsEditor, but you would have to pass in the containerId, not the iTwin/iModel id. Maybe I should rename it to findSettingsContainers or getSettingsContainers

core/backend/src/internal/workspace/SettingsImpl.ts
johnnyd710: Working now, so far so good

core/backend/src/workspace/Workspace.ts
johnnyd710: Nice - `findContainers` is very helpful. Less code I have to worry about. I realized it would actually be nice to have the same method on WorkspaceEditor too.

```diff
- const [settingsMetadata] = await service.queryContainersMetadata(userToken, {
-   containerType: "settings",
-   iTwinId,
- });
-
- const settingsWriteToken = await service.requestToken({
-   containerId: settingsMetadata?.containerId,
-   userToken,
-   accessLevel: "write",
- });
-
- const settingsEditor = SettingsEditor.construct();
- let settingsContainer: ReturnType<typeof settingsEditor.getContainer> | undefined;
- settingsContainer = await settingsEditor.getContainerAsync({
-   containerId: settingsContainerId,
-   baseUri: settingsWriteToken.baseUri,
-   storageType: settingsWriteToken.provider,
-   writeable: true,
- });
+ const settingsEditor = SettingsEditor.construct();
+ const [settingsContainer] = await settingsEditor.findContainers({ iTwinId });
```

core/backend/src/workspace/SettingsDb.ts
```diff
@@ -0,0 +1,105 @@
...
+  /** The [semver](https://github.com/npm/node-semver) version string or range for the desired [[SettingsDb]].
+   * If not specified, the latest available version is used.
+   */
+  readonly version?: string;
+}
+
+/** Arguments for obtaining a [[SettingsDb]] from a previously-loaded container.
+ * @beta
+ */
+export interface GetSettingsDbArgs {
```
johnnyd710, line 47: Why does this have a dbName but not a version? How do I read from a specific settings db version?

hl662: that was a mistake, i added an optional version string to this interface - if not passed in, we'll resolve to the latest db version

docs/learning/backend/Workspace.md
```diff
@@ -133,6 +133,55 @@ The next time we open the iModel, the new settings dictionary will automatically
 
 The "hardinessRange" setting is obtained from the iModel's settings dictionary, while the "defaultTool" falls back to the value defined in `IModelHost.appWorkspace.settings`.
 
+## SettingsDb
```
johnnyd710, line 156: Documentation still needs work - this is where we could use more reviewers

docs/learning/backend/Workspace.md
johnnyd710, line 1: there's a workspace anatomy diagram at the top that needs to be updated

core/backend/src/workspace/SettingsDb.ts
johnnyd710: nice works now

core/backend/src/internal/workspace/SettingsImpl.ts
```diff
@@ -183,3 +196,268 @@ export class SettingsImpl implements Settings {
...
+      throw new Error(`EditableSettingsDb.close() failed`);
+    }
+  }
+
+  public updateManifest(manifest: SettingsDbManifest): void {
+    this.sqliteDb[_nativeDb].saveFileProperty(settingsManifestProperty, JSON.stringify(manifest));
+    this._manifest = undefined;
+  }
+
+  public updateSettingsDictionary(name: string, settings: SettingsContainer): void {
```
johnnyd710: nice to have: a method to update just one setting inside a settings dictionary. it'd avoid me having to call `getSettingsDb`, `getDictionary`, and the merging with my changes. it also avoids the possibility of me accidentally overriding settings not related to my changes

core/backend/src/internal/workspace/SettingsDbImpl.ts
```diff
@@ -0,0 +1,144 @@
...
+  public readonly [_implementationProhibited] = undefined;
+  public readonly props: SettingsDictionaryProps;
+  private readonly _settings: SettingsContainer;
+
+  public constructor(props: SettingsDictionaryProps, settings: SettingsContainer) {
+    this.props = { ...props };
+    this._settings = settings;
+  }
+
+  public getSetting<T extends Setting>(settingName: string): T | undefined {
```
johnnyd710: nice to have: getAllSettings/toJSON?

core/backend/src/internal/workspace/SettingsImpl.ts
johnnyd710: We might want to consider storing all settings inside a single dictionary to simplify this - its what I planned on doing from iTS side

hl662: nice yeah, I added a updateSetting (and updateSettings) method to the EditableSettingsDb interface.

And I'm dropping remove/updateSettingsDictionary from that same interface too, check the api changes in the md linked [here](https://github.com/iTwin/itwinjs-core/pull/9077/changes/e79e79259f98ce70781e1b883e10da9dbb272466#diff-296ede995d70f162e9ac5594f3a209a60cb11782c6e46f971121878a551d363fL2528)

core/backend/src/internal/workspace/SettingsDbImpl.ts
hl662: I added both... well, they're in different interfaces:

   - SettingsDb.getSettings() — on the SQLite-backed database. Reads the JSON blob from the strings table, parses it, returns the full SettingsContainer. This is the low-level storage API.
   - SettingsDictionary.toJSON() — on the in-memory dictionary object (which wraps a SettingsContainer and participates in the priority-based resolution chain via Settings). Returns a deep clone of its internal state so callers can inspect/mutate it without affecting the live dictionary.

  They're not duplicates — SettingsDb is the persistence layer (read/write SQLite), while SettingsDictionary is the runtime layer (one of potentially many dictionaries that Settings resolves across by priority). A SettingsDictionary doesn't necessarily come from a SettingsDb — it could be constructed in-memory.

core/backend/src/workspace/Workspace.ts
hl662: Added findContainers to WorkspaceEditor too now

core/backend/src/internal/workspace/SettingsImpl.ts
johnnyd710: awesome, simpler and less failure points for the consumer.

docs/learning/backend/Workspace.md
```diff
@@ -1,8 +1,28 @@
...
+      WC1["WorkspaceDb<br/>fonts, textures"]
+      WC2["WorkspaceDb<br/>templates"]
+    end
+  end
+
+  SC1 -->|"loads as<br/>SettingsDictionary"| S
+  SC2 -->|"loads as<br/>SettingsDictionary"| S
+  S -->|"settings point to"| WC1
+  S -->|"settings point to"| WC2
+```
```
johnnyd710, line 23: for some reason the docs site cuts off part of the last letter on all boxes

Image

docs/learning/backend/Workspace.md
```diff
@@ -133,13 +153,93 @@ The next time we open the iModel, the new settings dictionary will automatically
...
+
+When a `SettingsDb` is loaded into the runtime via [Workspace.getSettingsDb]($backend), its contents become **one** [SettingsDictionary]($backend) in the [Settings]($backend) priority stack. The data flow is:
+
+`SettingsDb` → JSON → `Settings.addJson()` → one `SettingsDictionary` in the priority stack
+
+Each `SettingsDb` occupies a single slot in the [priority system](#settings-priorities). Multiple `SettingsDb`s — for example, one scoped to an iTwin at [SettingsPriority.iTwin]($backend) and one scoped to an iModel at [SettingsPriority.iModel]($backend) — become separate dictionaries with separate priorities. The runtime resolves conflicts using the standard priority rules: the highest-priority dictionary that provides a value for a given setting name wins.
+
+### Discovering settings containers
+
+Because every settings container is tagged with `containerType: "settings"`, you can find all settings containers for a given iTwin without knowing their IDs in advance. Use [SettingsEditor.queryContainers]($backend) to query by iTwinId (and optionally iModelId):
```
johnnyd710:
```suggestion
You can find all settings containers for a given iTwin by using [SettingsEditor.queryContainers]($backend) to query by iTwinId (and optionally iModelId):
```

docs/learning/backend/Workspace.md
```diff
@@ -133,13 +153,93 @@ The next time we open the iModel, the new settings dictionary will automatically
...
 
+## SettingsDb
+
+So far, we have loaded [SettingsDictionary]($backend)s directly at run-time — building them in code and adding them to [Workspace.settings]($backend). This works for small-scale configuration and testing, but real-world deployments need a way to **store settings in the cloud**, version them over time, and discover them without opening an iModel. That is the role of [SettingsDb]($backend).
+
+A `SettingsDb` is a dedicated [CloudSqlite]($backend) database that stores settings as a flat JSON object — [SettingName]($backend)s mapped to [Setting]($backend) values, with no binary resources. Its containers are tagged with `containerType: "settings"` in their cloud metadata, making them discoverable independently of any iModel. This is what distinguishes a `SettingsDb` from a [WorkspaceDb]($backend): a `SettingsDb` is where you **start**. You load settings from a `SettingsDb`, and those settings tell your application which [WorkspaceDb]($backend)s hold the binary resources (fonts, textures, images) it needs.
```
ben-polinsky:
```suggestion
A `SettingsDb` is a dedicated [CloudSqlite]($backend) database that stores settings as a flat JSON object — [SettingName]($backend) keys mapped to [Setting]($backend) values, with no binary resources. Its containers are tagged with `containerType: "settings"` in their cloud metadata, making them discoverable independently of any iModel. This is what distinguishes a `SettingsDb` from a [WorkspaceDb]($backend): a `SettingsDb` is where you **start**. You load settings from a `SettingsDb`, and those settings tell your application which [WorkspaceDb]($backend)s hold the binary resources it needs.
```
No need to repeat the binary parenthetical. It's mentioned above.

docs/learning/backend/Workspace.md
```diff
@@ -133,13 +153,93 @@ The next time we open the iModel, the new settings dictionary will automatically
...
+3. **Acquire the write lock** — [EditableSettingsCloudContainer.acquireWriteLock]($backend). Only one user can hold the lock at a time.
+4. **Open an EditableSettingsDb** — [EditableSettingsCloudContainer.getEditableDb]($backend) returns an [EditableSettingsDb]($backend).
+5. **Write settings** — use [EditableSettingsDb.updateSettings]($backend) to replace all settings, or [EditableSettingsDb.updateSetting]($backend) to patch a single key.
+6. **Release the lock** — [EditableSettingsCloudContainer.releaseWriteLock]($backend) publishes your changes. Alternatively, [EditableSettingsCloudContainer.abandonChanges]($backend) discards them.
+
+> **Important**: Always release the write lock when you are done. Failing to release it will prevent other administrators from modifying the container until the lock expires.
+
+### Updating individual settings
+
+Often you need to change a single setting without touching the rest. [EditableSettingsDb.updateSetting]($backend) reads the existing settings, patches the specified key, and writes the result back — other settings are preserved:
```
ben-polinsky: Patches the key? Or patches the "entry"?

hl662: i changed it to 'entry', that makes more sense

docs/learning/backend/Workspace.md
hl662: Interesting, it's showing up fine on my end... 
image

johnnyd710: maybe just a browser issue, think we can ignore as long as yours look ok

johnnyd710 (approved):

PR #9080, opened 2026-03-11
Security: fix high-severity CVE-2026-30951 (GHSA-6457-6jrx-69cr) in sequelize

hl662:
## Summary

Resolves **GHSA-6457-6jrx-69cr** (CVE-2026-30951) — high-severity SQL injection in sequelize v6 via JSON column cast type.

### Before
- 2 high, 8 moderate, 4 low vulnerabilities

### After
- **0 high/critical**, 7 moderate, 4 low vulnerabilities

## Fix Details

| CVE | Severity | Package | Fix |
|-----|----------|---------|-----|
| CVE-2026-30951 | High | sequelize 6.37.7 → 6.37.8 | lockfile refresh via `rush update --full` |

**Dependency path (transitive):** azurite@3.35.0 → sequelize@^6.31.0
Patched version 6.37.8 falls within azurite's semver range — no `package.json` or `globalOverride` changes needed.

### Affected packages
- `example-code/snippets`
- `full-stack-tests/backend`
- `full-stack-tests/core`

## Validation

- ✅ `rush audit`: 0 high/critical vulnerabilities
- ✅ `rush build`: passed
- ✅ `rush test`: passed (1 pre-existing WebGL GPU test failure on master, unrelated)
- ⏭️ `rush extract-api`: skipped (lockfile-only change)
- ✅ `rush change --verify`: no change files required

## Advisory
- https://github.com/advisories/GHSA-6457-6jrx-69cr

PR #9086, opened 2026-03-12
Agent for adding support for new Electron major versions

GytisCepk:
Running locally this worked without issues, however trying to trigger this agent from other GitHub agent produced [mixed results](https://github.com/iTwin/itwinjs-core/pull/9085). Since we can't run these agents off branch in GitHub, I'm not sure how exactly we suppose to tests them in "real world" conditions.

Once this gets merged, I'll initialize a new GitHub agent run, this time starting agent directly.

### DIFF
--- a/.github/agents/electron-version-bump.agent.md
+++ b/.github/agents/electron-version-bump.agent.md
@@ -11,6 +11,7 @@ You are an agent that adds support for a new Electron major version in the iTwin
 - Review Electron's breaking changes for the new major version and apply any required code fixes.
 - Add a new Electron major version to the supported peer dependency ranges.
 - Update all dev dependency pinned Electron versions across the repo.
+- Check for newer versions of `@itwin/electron-authorization` that support the new Electron version and update if available.
 - Document the change in the changelog and create Rush change files.
 - Validate that build and tests pass after the change.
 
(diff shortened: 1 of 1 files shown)


ben-polinsky:
> trying to trigger this agent from other GitHub agent produced https://github.com/iTwin/itwinjs-core/pull/9085.

@hl662 has seen similar results

hl662:
Yeah, I ran into the same thing with the CVE remediation agent — cloud agent ([#9079](https://github.com/iTwin/itwinjs-core/pull/9079)) got stuck on step 1 of 6, never made a single file change. Local CLI ([#9080](https://github.com/iTwin/itwinjs-core/pull/9080)) did the whole thing in ~50 min and merged.

The main issue is **we don't have a `copilot-setup-steps.yml`** in this repo. Without it the cloud agent gets a vanilla Ubuntu Actions runner — no Rush, no pnpm, no correct Node.js version. It can understand *what* to do (the CVE agent produced a perfect checklist) but can't *execute* because `rush install` never succeeds. Even if we add the setup file, the bootstrapping cost for this monorepo is heavy — hundreds of packages, native deps — which eats into the agent's session budget before it starts real work.

> Since we can't run these agents off branch in GitHub, I'm not sure how exactly we suppose to tests them in "real world" conditions.

That's the other pain point. You basically have to merge first and hope it works, or test locally with Copilot CLI and treat the cloud run as a verification step after merge. Not great.

I think the concrete next steps are:
1. **Add `.github/workflows/copilot-setup-steps.yml`** — install Node.js, Rush, run `rush install`. This alone would unblock most cloud agent tasks.
2. **Use local CLI for anything that needs build/test validation** — the cloud agent is better suited for tasks that don't need a full monorepo build (research, simple file edits, etc.)
3. For testing agent instructions pre-merge, local CLI is really the only viable path right now.

Cloud agent sessions do leverage git hub actions, including any secrets we may set up for authenticating packages. As long as we make sure agent invocation is restricted to only people in the org, we can mitigate the risk of abuse of these agents by malicious third parties. 

That's one thing we have to think more about - This is a public repo, how do we avoid abuse? 

Anyways, before we fix cloud agent invocation, I think it's better for us to move our ADO CI pipelines to GH. That way, we'll have solved any issues with secrets ahead of time, and will have prepared our GH Actions runners for more automated usage

PR #9090, opened 2026-03-16
Refactor backend editing around explicit EditTxn and deprecate legacy mutators

kabentley:
`EditTxn` is now the primary backend edit surface. This change moves transaction lifecycle and mutating operations onto explicit transaction objects instead of relying on implicit mutation through IModelDb, elements, models, and relationships.

The old mutating APIs are now deprecated compatibility wrappers. They still work, but new code is expected to edit through `EditTxn` or subclasses like `EditCommand`. This makes transaction ownership explicit, keeps active-txn state in one place, and removes the need for separate admin/state helpers.

### DIFF
--- a/core/backend/src/EditTxn.ts
+++ b/core/backend/src/EditTxn.ts
@@ -22,6 +22,11 @@ export class EditTxn {
   /** The iModel this EditTxn may modify. */
   public readonly iModel: IModelDb;
 
+  /** Convenience accessor for the iModel this EditTxn may modify. */
+  public get db(): IModelDb {
+    return this.iModel;
+  }
+
   protected constructor(iModel: IModelDb) {
     this.iModel = iModel;
   }
@@ -69,11 +74,12 @@ export class EditTxn {
     this.end(true, args);
   }
 
-  /** Cancel the changes in this EditTxn.
+  /** Abandon database changes while keeping this EditTxn active.
    * @throws EditTxnError if this EditTxn is not active.
    */
-  public cancel(): void {
-    this.end(false);
+  protected abandonChanges(): void {
+    this.requireActive();
+    this.iModel.abandonChanges();
   }
 
   /** Save changes with additional arguments.
--- a/core/backend/src/TxnManager.ts
+++ b/core/backend/src/TxnManager.ts
@@ -1231,7 +1231,7 @@ export class TxnManager {
    */
   public reverseTxns(numOperations: number): IModelStatus {
     if (this._nativeDb.hasUnsavedChanges()) {
-      this._iModel.activeTxn.cancel();
+      this._iModel.activeTxn.end(false);
     }
     return this._nativeDb.reverseTxns(numOperations);
   }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -43,6 +43,8 @@
     "**/lib": true,
     "common/temp": true
   },
+  "files.autoSave": "afterDelay",
+  "files.autoSaveDelay": 1000,
   "files.trimTrailingWhitespace": true,
   "importSorter.generalConfiguration.sortImportsInDirectory": false,
   "importSorter.generalConfiguration.sortOnBeforeSave": false,
@@ -105,6 +107,30 @@
   "terminal.integrated.env.windows": {
     "NODE_ENV": "development"
   },
+  "terminal.integrated.defaultProfile.windows": "PowerShell (NoProfile)",
+  "terminal.integrated.profiles.windows": {
+    "PowerShell (NoProfile)": {
+      "path": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
+      "args": [
+        "-NoLogo",
+        "-NoProfile"
+      ]
+    },
+    "Command Prompt (No Autorun)": {
+      "path": "C:\\Windows\\System32\\cmd.exe",
+      "args": [
+        "/d"
+      ]
+    }
+  },
+  "terminal.integrated.automationProfile.windows": {
+    "path": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
+    "args": [
+      "-NoLogo",
+      "-NoProfile"
+    ]
+  },
+  "terminal.integrated.shellArgs.windows": [],
   "typescript.tsdk": "common/temp/node_modules/typescript/lib",
   "vitest.maximumConfigs": 20,
   "github.copilot.chat.confirmChanges": false,
--- a/core/backend/src/test/ElementDrivesElement.test.ts
+++ b/core/backend/src/test/ElementDrivesElement.test.ts
@@ -15,8 +15,22 @@ import { ElementDrivesElement, ElementDrivesElementProps } from "../Relationship
 import { Schema, Schemas } from "../Schema";
 import { HubWrappers } from "./IModelTestUtils";
 import { KnownTestLocations } from "./KnownTestLocations";
-import { editTxnOf } from "./TestEditTxn";
+import { TestEditTxn } from "./TestEditTxn";
 chai.use(chaiAsPromised);
+
+function getTestTxn(iModelDb: IModelDb): TestEditTxn {
+  const activeTxn = iModelDb.activeTxn;
+  if (activeTxn instanceof TestEditTxn)
+    return activeTxn;
+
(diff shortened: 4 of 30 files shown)


kring:
This makes sense to me, and represents much less of a breaking change than I expected when you described it to me before. Nice! Two high level observations / questions for you:

### Exceptions

I see you changed the doc for several methods to say they throw an `EditTxnError` instead of `IModelError`. It looks like this may not be entirely true yet, but that may just be because it's a work in progress. More importantly, isn't that a breaking change?

### activeTxn

I see the current `activeTxn` is readily accessible on the `IModelDb`. A major purpose of this PR (as I understand it) is to defend against code that thoughtlessly invokes `updateElement` (for example) without considering what else is happening. So I worry that such code will now be written (still thoughtlessly) as `db.activeTxn().updateElement(...)`, and thus have the exact same problem it did before.

We could avoid this by making the `activeTxn` inaccessible. That would require that anyone that wants to use the active txn to be given it explicitly. Better safety, but it puts some burden on users to pass that reference around. Do you think that could be a worthwhile tradeoff?

kabentley:
> ### Exceptions
> I see you changed the doc for several methods to say they throw an `EditTxnError` instead of `IModelError`. It looks like this may not be entirely true yet, but that may just be because it's a work in progress. More importantly, isn't that a breaking change?

This PR is still very much WIP, and to be clear all changes were authored by copilot. No changes anywhere were typed by me. But, you're right it did put comments about `@throws` that aren't right yet. It does now throw new exceptions, but all of the old exceptions are unchanged. The documentation is wrong as it stands.

> ### activeTxn
> I see the current `activeTxn` is readily accessible on the `IModelDb`. A major purpose of this PR (as I understand it) is to defend against code that thoughtlessly invokes `updateElement` (for example) without considering what else is happening. So I worry that such code will now be written (still thoughtlessly) as `db.activeTxn().updateElement(...)`, and thus have the exact same problem it did before.

That's why all of the methods on `EditTxn` are protected, so you *can't* call `updateElement` on the activeTxn.

Stay tuned. I'm working on converting the tests. That should make it clearer how new code should use the `EditTxn` apis.

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,417 @@
...
+    } else {
+      this.iModel.abandonChanges();
+    }
+    this.restoreLegacyTxn();
+  }
+
+  /** Commit the changes in this EditTxn.
+   * @throws EditTxnError if this EditTxn is not active.
+   */
+  protected commit(args?: string | SaveChangesArgs): void {
```
tcobbs-bentley, line 68: Is there a reason that this function is protected? Since all it does is call the public `end` function, having it be protected doesn't seem to accomplish anything, and `cancel` below is public. I see that `saveChanges` is also protected, along with the various edit functions. I assume that there is a reason for this, but I don't know what it is, so I think that documentation somewhere should indicate why. Is this class intended to always be subclassed?

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,417 @@
...
+   */
+  protected saveChanges(args?: string | SaveChangesArgs): void {
+    this.requireActive();
+    this.iModel.saveChangesImpl(args);
+  }
+
+  /** Insert a new element into the iModel.
+   * @param elProps The properties of the new element.
+   * @returns The newly inserted element's Id.
+   * @throws EditTxnError if this EditTxn is not active.
```
tcobbs-bentley, line 91: Should this also indicate other things that might get thrown. A bunch of functions in this file are documented to throw `EditTxnError` while also throwing other errors once they get past the `requireActive` check. It seems strange to me to document the `EditTxnError` throws while omitting the others, but maybe this was intentional.

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,417 @@
...
+      err.metadata = { elProps };
+      throw err;
+    }
+  }
+
+  /** Delete elements from the iModel.
+   * @param ids The Ids of the elements to delete.
+   * @throws EditTxnError if this EditTxn is not active.
+   */
+  protected deleteElement(ids: Id64Arg): void {
```
tcobbs-bentley, line 141: Is there a reason that this isn't named `deleteElements`?

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,423 @@
...
+    this.requireActive();
+    const iModel = this.iModel;
+    Id64.toIdSet(ids).forEach((id) => {
+      try {
+        this.iModel.elements[_cache].delete({ id });
+        this.iModel.elements[_instanceKeyCache].deleteById(id);
+        iModel[_nativeDb].deleteElement(id);
```
tcobbs-bentley, line 154: Why do the first two lines use `this.iModel` while the last uses `iModel`? Given the mostly single-threaded nature of JS, can the values ever differ?

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,423 @@
...
+    const error = this.iModel[_nativeDb].updateModelGeometryGuid(modelId);
+    if (error !== IModelStatus.Success)
+      throw new IModelError(error, `Error updating geometry guid for model ${modelId}`);
+  }
+
+  /** Delete models from the iModel.
+   * @param ids The Ids of the models to delete.
+   * @throws EditTxnError if this EditTxn is not active.
+   */
+  protected deleteModel(ids: Id64Arg): void {
```
tcobbs-bentley, line 213: Is there a reason this isn't named `deleteModels`?

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,423 @@
...
+    return props.id = this.iModel[_nativeDb].insertLinkTableRelationship(props);
+  }
+
+  /** Update an existing relationship in the iModel.
+   * @param props the properties of the relationship to update.
+   * @throws EditTxnError if this EditTxn is not active.
+   */
+  protected updateRelationship(props: RelationshipProps): void {
+    this.requireActive();
+    this.iModel[_nativeDb].updateLinkTableRelationship(props);
```
tcobbs-bentley, line 247: Should this throw if the return value of `updateLinkTableRelationship` is not BE_SQLITE_OK? The same question applies to `deleteRelationship` and `deleteRelationships`.

core/backend/src/IModelDb.ts
```diff
@@ -421,6 +422,15 @@ export abstract class IModelDb extends IModel {
   /** @internal */
   protected _codeService?: CodeService;
 
+  /** @internal */
+  public readonly [_legacyEditTxn]: LegacyEditTxn;
+
+  /** @internal */
+  public [_activeTxn]!: EditTxn;
```
tcobbs-bentley, line 429: I can't find where this is initialized. `EditTxn` sets it to itself in `start` and sets it to `[_legacyEditTxn]` in `restoreLegacyTxn`, but those are the only places that I can find that it is set, and yet it is being declared that it will always be defined before use.

core/backend/src/TxnManager.ts
```diff
@@ -147,7 +148,7 @@ class ChangedEntitiesProc {
 
   public static maxPerEvent = 1000;
 
-  public static process(iModel: BriefcaseDb | StandaloneDb, mgr: TxnManager): void {
+  public static process(iModel: BriefcaseDb, mgr: TxnManager): void {
```
tcobbs-bentley, line 151: Is this internal? I can't find documentation comments indicating that it is, but it's also undocumented (along with the `ChangedEntitiesProc` class). The same question applies to the other breaking changes here.

full-stack-tests/core/src/backend/backend.ts
```diff
@@ -47,6 +47,16 @@ function loadEnv(envFile: string) {
...
 let electronAuth: ElectronMainAuthorization;
 
+interface EditTxnAccess {
+  insertElement(props: ElementProps): Id64String;
+  updateElement(props: ElementProps): void;
+  insertModel(props: ModelProps): Id64String;
+}
+
+function editTxnOf(iModelDb: IModelDb): EditTxnAccess {
+  return iModelDb.activeTxn as unknown as EditTxnAccess;
```
tcobbs-bentley, line 57: This seems to answer my earlier question about protected members, but documenting this unusual usage pattern in the `EditTxn` code would probably be good.

PR #9093, opened 2026-03-17
Strict schema loading

rschili:
imodel-native: https://github.com/iTwin/imodel-native/pull/1356

Preparation work for EC 3.3.
See linked story for details


rschili:
Abandoning, we will revisit once we get to planning a new EC spec release, which has been deferred for the moment

PR #9094, opened 2026-03-17
Migrate full-stack-tests/core from Certa+Mocha+webpack to Vitest via vitest-certa-bridge

hl662:
## Motivation

Certa — our custom test harness for \`full-stack-tests/core\` — has been a long-standing source of friction:

- **No ESM support**: Electron doesn't support ESM natively, and Certa has no active development to work around it. Since our full-stack integration tests depend on Certa, this blocks our packages from dropping dual CJS/ESM shipping and going ESM-only
- **Slow**: Synchronous IPC-based console throttling makes test runs ~10× slower than they need to be
- **Webpack-locked**: Tied to webpack while the rest of the ecosystem has moved to Vite
- **Fragile CI**: GPU crashes on Windows (WARP/SwiftShader) required manual reruns with no automated retry

Vitest browser mode + Playwright is the modern replacement — but browser tests can't call Node.js APIs directly. This PR bridges that gap.

## What This PR Does

1. **Introduces \`@itwin/vitest-certa-bridge\`** — a Vite plugin + Electron test orchestrator that replaces Certa's backend↔frontend callback bridging
2. **Migrates \`full-stack-tests/core\`** from Certa + Mocha + webpack → Vitest, using the bridge plugin
3. **Removes Certa** from \`core/i18n\` and \`core/electron\` (migrated to Vitest browser or Node.js Vitest)

<details>
<summary><strong>Bridge plugin details</strong></summary>

- **Chrome mode**: Vite plugin with \`/__certa_bridge\` HTTP middleware for backend callbacks
- **Electron mode**: Parallel sharded test runner with crash retry, bisect, grep filtering, and configurable timeouts
- \`electronArgs\` option passes extra flags (e.g. \`--disable-gpu\` on CI), preserved across crash retries
- \`importRewritePatterns\` and \`rendererSetup\` hooks decouple iTwin-specific patches from the generic harness

</details>

<details>
<summary><strong>full-stack-tests/core migration details</strong></summary>

- Removed webpack config and Certa infrastructure
- Added \`vitest.config.mts\` (Chrome/Playwright) and \`vitest.electron.config.mts\`
- Chrome runs a smoke subset (non-GPU tests); Electron runs the full suite including rendering/tile tests
- Backend split: \`BackendServer.ts\` (standalone RPC server) + \`BridgeInit.ts\` (callback registration)
- Integration/performance tests gated behind \`#integration\`/\`#performance\` grep tags
- 14 suites skip in Electron (some validate the HTTP bridge path only present in Chrome; others skip for non-HTTP reasons such as DOM availability)

</details>

## CI Hardening

- **GPU stability**: \`--disable-gpu\` via \`electronArgs\` on Linux and Windows CI
- **Crash retry**: Electron shards that exit without writing \`test-results.json\` (covers native crashes, OOM kills, and unexpected clean exits) retry once with \`--disable-gpu\` and a 2s cooldown; if the retry also produces no results and the shard has >1 file, the shard is bisected into two parallel sub-shards to isolate the problematic file
- **Shard config**: Linux 2 shards, macOS 1 shard, Windows 2 shards (halves per-process memory pressure that triggers fast-fail on long sequential runs with the native addon)

## Test Results

| Platform | Chrome (browser) | Electron | Notes |
|----------|-----------------|----------|-------|
| Linux | 128 tests ✅ | 272 passed ✅ | 2 shards |
| macOS | 128 tests ✅ | 272 passed ✅ | 1 shard |
| Windows | 128 tests ✅ | 272 passed ✅ | 2 shards, \`--disable-gpu\` |

## Validation

- Targeted verification: All three CI platforms pass Chrome + Electron suites end-to-end
- Known baseline issues: None — all flakes addressed in this PR

<details>
<summary><strong>Sacrifices — things that may look suspect without context</strong></summary>

- **\`SheetViewState.test.ts\` — 6 transient \`areAllTilesLoaded === false\` assertions removed**
  Under real Electron IPC, the response resolves before the \`onElementsChanged\` push notification — the viewport hasn't been invalidated yet. The transient loading state was never an API-contract guarantee. Reload + render + data correctness assertions are all preserved.

- **\`BackgroundMap\`, \`PlanProjection\`, \`PlanarClipMask\` skipped in Electron**
  Each calls \`vp.waitForAllTilesToRender()\` with real Bing Maps fetches + main-thread IMDL decode + SwiftShader on CI → 150–200s per test. Chrome and Electron share the exact same rendering pipeline; Chrome still runs all of them.

- **8 transitive deps added to \`full-stack-tests/core\` devDependencies**
  Packages like \`js-base64\`, \`flatbuffers\`, \`fuse.js\`, etc. Vite's dep scanner only finds direct imports — when a transitive dep is discovered mid-test, Vite reloads the page and kills the test run. Listed in both \`devDependencies\` and \`optimizeDeps.include\` with a cross-linking \`@see\` comment.

</details>

---
*Nambot 🤖*


tools/vitest-certa-bridge/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@itwin/vitest-certa-bridge",
+  "version": "5.8.0-dev.6",
+  "description": "Vitest plugin that bridges browser test code to Node.js backend callbacks, replacing @itwin/certa",
```
hl662: once electron bridge is complete, update description

hl662:
todo: update CODEOWNERS for certa, and add codeowners for this vitest plugin

aruniverse (changes requested):
just holding pr from merging until i can review

core/electron/src/common/ElectronIpcTransport.ts
```diff
@@ -165,6 +165,10 @@ export class FrontendIpcTransport extends ElectronIpcTransport<RpcRequestFulfill
 
     const protocol = this._protocol;
     const request = protocol.requests.get(message.id) as ElectronRpcRequest;
```
GytisCepk:
```suggestion
    const request = protocol.requests.get(message.id) as ElectronRpcRequest | undefined;
```

core/electron/src/test/backend/RunElectronBackendTests.ts
```diff
@@ -13,7 +13,10 @@ async function spawnElectronMainProcess(suiteToRun: string, testToRun: string) {
   const command = require("electron/index.js"); // eslint-disable-line @typescript-eslint/no-require-imports
 
   const args = [
-    path.join(__dirname, "./RunSingleTest.js"),
+    // RunSingleTest runs inside a spawned Electron main process (needs real
+    // Electron APIs), so it must use compiled JS — tsx/esm loaders don't
```
GytisCepk: I would remove `(needs real Electron APIs)` part. What makes Electron APIs "real" ?

core/electron/package.json
```diff
@@ -14,8 +14,9 @@
     "lint-deprecation": "eslint --fix -f visualstudio --no-inline-config -c ../../common/config/eslint/eslint.config.deprecation-policy.js \"./src/**/*.ts\"",
     "test": "",
     "test:integration": "npm run test:integration:backend && npm run test:integration:frontend",
-    "test:integration:backend": "mocha --config src/test/backend/.mocharc.json",
-    "test:integration:frontend": "certa -r electron --config src/test/frontend/utils/certa.json",
+    "test:integration:backend": "vitest run --config vitest.config.mts",
+    "test:integration:frontend": "vitest run --config vitest.frontend.config.mts",
+    "test:integration:frontend:legacy": "certa -r electron --config src/test/frontend/utils/certa.json",
```
GytisCepk: What is the reason for keeping legacy test script ?

core/electron/src/test/frontend/utils/RunRendererTests.ts
```diff
@@ -0,0 +1,121 @@
...
+  // where no GPU is available. Must be set before app.whenReady().
+  app.commandLine.appendSwitch("enable-unsafe-swiftshader");
+
+  await app.whenReady();
+  await ElectronHost.startup();
+
+  const win = new BrowserWindow({
+    show: false,
+    webPreferences: {
+      nodeIntegration: true,
```
GytisCepk, line 28: Tests should not be run with `nodeIntegration`

core/electron/src/test/frontend/utils/RunRendererTests.ts
GytisCepk, line 1: If I understand correctly, each test package needs to define a file like this for electron tests ? Is there a reason why this couldn't be handled by the new package ?

GytisCepk (commented):
I think, core-electron tests migration should have been another PR on it's own, considering how hacky current tests can be there.

tools/vitest-certa-bridge/src/electron/runner.ts
```diff
@@ -0,0 +1,249 @@
...
+  const pkgRoot = path.resolve(__dirname, "../../..");
+  return path.join(pkgRoot, "lib", "cjs", "electron", "session.js");
+}
+
+/** Discover test files matching a glob-like pattern in testDir. */
+function findTestFiles(testDir: string, pattern: string): string[] {
+  // Support simple glob patterns: "**/*.test.js" → recursive search for .test.js files
+  const isRecursive = pattern.includes("**");
+  const ext = path.extname(pattern); // e.g. ".js"
+  const suffix = pattern.replace("**/", "").replace("*", ""); // e.g. ".test.js"
```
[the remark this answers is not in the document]
hl662: Outdated — this code was refactored and the flagged line no longer exists in the current diff.

core/i18n/vitest.config.mts
```diff
@@ -0,0 +1,60 @@
...
+      // This runs BEFORE Vite's internal middleware
+      server.middlewares.use((req, res, next) => {
+        if (req.url) {
+          const match = req.url.match(/\/locales\/([^?]+\.json)(?:\?.*)?$/);
+          if (match) {
+            const filePath = path.resolve(__dirname, "src/test/public/locales", match[1]);
+            if (fs.existsSync(filePath)) {
+              res.setHeader("Content-Type", "application/json");
+              res.setHeader("Access-Control-Allow-Origin", "*");
+              fs.createReadStream(filePath).pipe(res);
```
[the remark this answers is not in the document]
hl662: Good catch — added a path traversal guard that resolves the full path and verifies it stays within the `localesDir` before serving.

tools/vitest-certa-bridge/src/plugin.ts
```diff
@@ -0,0 +1,142 @@
...
+          const mod = await import(moduleUrl);
+          const result = mod.default ?? mod;
+          if (typeof result === "function") {
+            cleanupFn = result;
+          } else if (typeof result?.then === "function") {
+            const resolved = await result;
+            if (typeof resolved === "function") {
+              cleanupFn = resolved;
+            }
+          }
```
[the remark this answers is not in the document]
hl662: Good catch — updated the docstring to accurately describe the actual behavior: the module initializes via side effects when `import()`ed, and if the default export is a function, it's stored as a cleanup function called on server close (not as an init function).

full-stack-tests/core/package.json
```diff
@@ -74,18 +71,19 @@
     "@itwin/itwins-client": "^1.6.1",
     "@itwin/object-storage-core": "^3.0.4",
     "@itwin/oidc-signin-tool": "^5.0.2",
-    "@types/chai": "4.3.1",
-    "@types/chai-as-promised": "^7",
+    "@itwin/vitest-certa-bridge": "workspace:*",
     "@types/fs-extra": "^4.0.7",
```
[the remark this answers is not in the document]
hl662: Outdated — this was flagged against an earlier revision. The current diff no longer touches those dependency lines.

core/electron/package.json
hl662: Removed the legacy scripts (`test:integration:frontend:legacy` and `webpack:test`) along with the dead `certa.json` and `webpack.config.js` files they referenced.

core/electron/src/test/backend/RunElectronBackendTests.ts
hl662: Removed, good call.

core/electron/src/common/ElectronIpcTransport.ts
hl662: Removed the unnecessary type assertion entirely — `Map.get()` already returns `ElectronRpcRequest | undefined`, so the cast was redundant. Thanks.

core/electron/src/test/frontend/utils/RunRendererTests.ts
hl662: Each test package does need its own runner file, but it's intentionally lightweight — the bridge handles all the heavy lifting (test discovery, sharding, retry, IPC reporting).

The runner file only contains app-specific initialization that varies per package:
- Which `ElectronHost.startup()` options to use
- Which RPC interfaces to register
- Whether to call `IModelHost.startup()` vs `IModelApp.startup()`
- Package-specific test fixtures or environment setup

For example, `core/electron` registers `ElectronApp` + presentation RPC, while `full-stack-tests/core` sets up `IModelHost` with a test iModel. These can't be generalized without making the bridge aware of every consumer's domain — which would invert the dependency.

The bridge stays generic (Vitest ↔ Electron IPC plumbing), and each package owns its init. It's the same pattern as how each package has its own `vitest.config.mts` — the framework is shared, but config is per-consumer.

core/electron/src/test/frontend/utils/RunRendererTests.ts
hl662: Good point — `nodeIntegration: true` is intentionally used here because this is a test harness, not production app code.

The test bundle (`ElectronApp.test.ts`) calls `ElectronApp.startup()`, which internally resolves its IPC transport via:
```ts
// ElectronApp.ts:43-46
// use the methods on window.itwinjs exposed by ElectronPreload.ts,
// or ipcRenderer directly if running with nodeIntegration=true (**only** for tests).
this._api = window.itwinjs ?? require("electron").ipcRenderer;
```

The modern best practice would be to disable `nodeIntegration`, enable `contextIsolation`, and use a preload script with `contextBridge` instead. But to do that here, we'd need to:
1. Wire up `ElectronPreload.ts` as the preload script for the test `BrowserWindow`
2. Bundle the compiled test code (currently CJS with `require()` calls) so it can run without Node.js in the renderer — which means reintroducing webpack or another bundler just for this test

That defeats the purpose of this PR. The existing `nodeIntegration` fallback in `ElectronApp` was specifically designed for this test scenario (as the comment documents). Happy to revisit if we want to tackle it separately, but it felt out of scope here.

aruniverse (commented):
TEST - ignore

aruniverse:
## Architectural Review — Staff/Principal Level

> *Generated with AI assistance (GitHub Copilot), reviewed and approved before posting.*

This PR introduces `@itwin/vitest-certa-bridge` to migrate `full-stack-tests/core` from Certa+Mocha+webpack to Vitest. The migration goal is sound, but the implementation has several design issues ranging from correctness bugs to fragile abstractions that will make this package difficult to maintain and extend. Individual findings follow as separate comments.

**Summary by severity:**
- 🔴 Critical (2): correctness bugs that silently change test behavior or break security abstractions
- 🟠 Significant (3): architectural decisions that will hurt maintainability long-term  
- 🟡 Should-fix (6): smaller issues worth addressing before GA

> **Note:** GitHub's inline review API returned an internal error for all diff-anchored comments on this PR (likely due to its size — 134 commits, 155 files). Findings are posted as timeline comments with direct line links instead.

aruniverse:
### 🔴 `renderer-harness.ts` — This is a custom Mocha runner, not Vitest

📍 [`renderer-harness.ts:24`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/electron/renderer-harness.ts#L24) · `buildRendererHarness()`

`buildRendererHarness()` generates an inline JavaScript **string** that re-implements `describe`, `it`, `before`, `after`, grep filtering, and timeouts from scratch. This is a bespoke Mocha-compatible runner injected as a string literal into the Electron renderer process.

**Why this matters:**
- The PR is titled as a migration *to* Vitest, but in Electron mode the tests run against this hand-rolled harness. Vitest APIs (`vi.mock`, `vi.spyOn`, coverage, snapshots, `expect.soft`, `--reporter`) are **entirely unavailable** in the renderer process.
- The string-construction approach produces no source maps, so test failures show offsets into the injected bundle rather than source locations.
- Any future Vitest upgrade that changes the harness API requires updating the string-generation code in lock-step — this coupling is invisible at compile time.

**Suggestion:** Consider running Vitest itself in the renderer process rather than synthesizing a Mocha-compatible shim. If a full Vitest port is out of scope now, the PR description should clearly state which test environments get real Vitest and which get a compatibility shim, so consumers know what they're signing up for.

aruniverse:
### 🔴 `renderer-harness.ts:114` — `it.only` is silently swallowed in Electron mode

📍 [`renderer-harness.ts:114`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/electron/renderer-harness.ts#L114)

```ts
globalThis.it.only = globalThis.it;  // effectively a no-op
```

When a developer uses `it.only()` or `describe.only()` during local debugging, every test in the suite runs anyway. There is no warning, no error, and no indication that the focus filter was ignored. This is a silent correctness regression compared to the old Certa+Mocha setup.

**Why this matters:** A developer who `it.only`s a flaky test to debug it will see the full suite run and assume their change is good, when in fact they may have introduced a regression elsewhere that the narrowed run would have caught. Worse, they may accidentally commit the `.only` and not realize tests are still running fully.

**Suggestion:** Either implement proper `only` semantics in the harness, or throw at runtime: `globalThis.it.only = () => { throw new Error('it.only is not supported in Electron renderer mode') }`.

aruniverse:
### 🔴 `plugin.ts:125` — `dispatchCallback` token validation is a no-op

📍 [`plugin.ts:125`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/plugin.ts#L125)

```ts
await dispatchCallback(body.name, body.args, bridgeToken, bridgeToken);
//                                                         ^^^^^^^^^^^  ^^^^^^^^^^^
//                                                         token        expectedToken  <-- same value!
```

📍 [`callbackRegistry.ts:61`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/callbackRegistry.ts#L61)

```ts
export async function dispatchCallback(name, args, token, expectedToken) {
  if (token !== expectedToken) throw ...; // always false — never throws
}
```

The caller passes `bridgeToken` as **both** `token` and `expectedToken`, so the guard `token !== expectedToken` is permanently false and the check never fires. The real protection is the `x-bridge-token` HTTP header check upstream, but the abstraction is actively misleading: it implies a two-party token validation pattern while providing none.

**Suggestion:** Either (a) remove the `expectedToken` parameter from `dispatchCallback` and validate the token inside the registry against a stored value, or (b) have the call site pass the registry's stored expected token rather than the same bridgeToken it's already holding.

aruniverse:
### Remaining findings (🟠 Significant + 🟡 Should-fix)

---

#### 🟠 `renderer-harness.ts:316` — Global `Module._extensions` hook rewrites `import()` via regex

📍 [`renderer-harness.ts:316`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/electron/renderer-harness.ts#L316)

The bridge installs a `Module._extensions['.js']` hook that rewrites every `.js` file loaded in the renderer with a regex replacing `import()` → `require()`. This is global (affects all loaded files, including third-party), matches inside strings and comments, and will silently break if the TypeScript compiler's import transform changes. ESM/CJS incompatibility should be addressed at build time, not via a post-hoc regex rewrite over compiled output.

---

#### 🟠 `callbackRegistry.ts:61` — `dispatchCallback` signature makes correct usage accidental

📍 [`callbackRegistry.ts:61`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/callbackRegistry.ts#L61)

The signature `dispatchCallback(name, args, token, expectedToken)` requires the caller to supply both values. This is easy to get wrong (as evidenced by the bug in finding #3 above). Move the `expectedToken` inside the registry so the function validates against its own stored value and callers can't pass the same value twice by accident.

---

#### 🟠 `electron/types.ts:77` — `rendererSetup` documents `require.cache` mutation as `@beta` API

📍 [`electron/types.ts:77`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/electron/types.ts#L77)

The `rendererSetup` option is documented and typed as `@beta`, meaning it carries a stability promise. But it exposes patching `require.cache` (a Node.js internal) as the intended public API contract. This enshrines module-system internals manipulation as a supported pattern that consumers can depend on. If this must exist, it should be `@alpha` or marked clearly that it will be removed once the ESM transition is complete.

---

#### 🟠 `electron/session.ts:43` — 12+ environment variables as the orchestrator↔session interface

📍 [`electron/session.ts:43`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/electron/session.ts#L43)

The bridge passes configuration between the Vitest plugin and the Electron session via 12+ environment variables including `ELECTRON_RENDERER_SETUP` (base64-encoded JavaScript) and `ELECTRON_IMPORT_REWRITE_PATTERNS` (JSON-encoded array). This interface is invisible to the type system, unvalidatable at compile time, and undiscoverable. A typed options object serialized to `--bridge-config` (or similar) would be far more maintainable.

---

#### 🟡 `callbackRegistry.ts:22` — `getCallbacksRegisteredOnBackend()` returns a live mutable map

📍 [`callbackRegistry.ts:22`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/callbackRegistry.ts#L22)

The function returns a direct reference to the internal map. Any caller can add or remove entries without going through the registry's registration logic. Return `ReadonlyMap` or a shallow copy.

---

#### 🟡 `preferEsm.ts:51` — Synchronous `existsSync` in hot Vite `resolveId` path

📍 [`preferEsm.ts:51`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/src/preferEsm.ts#L51)

`existsSync` is called on every module resolution. Cache results after the first lookup per path to avoid repeated synchronous disk I/O during test runs with many imports.

---

#### 🟡 `.github/CODEOWNERS` — Single-person ownership for now-critical shared infrastructure

📍 [`.github/CODEOWNERS:35`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/.github/CODEOWNERS#L35) and [line 156](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/.github/CODEOWNERS#L156)

`@itwin/vitest-certa-bridge` and its consumer `full-stack-tests/core` are listed with a single owner (`@hl662`). This is now a shared CI infrastructure dependency for the monorepo. Add at least one more owner to avoid a bus-factor-1 situation.

---

#### 🟡 `vitest-certa-bridge/package.json:74` — `dotenv` listed as optional peer but used unconditionally

📍 [`vitest-certa-bridge/package.json:74`](https://github.com/iTwin/itwinjs-core/blob/f78b521a98cb1daecf5564e66b40a37827d43c47/tools/vitest-certa-bridge/package.json#L74)

`dotenv` is declared `optionalPeerDependencies` but consumer configs (`vitest.config.mts`) import it unconditionally at the top level. If it's truly optional, the import should be guarded; if it's required, move it to `peerDependencies`.

---

#### 🟡 `full-stack-tests/core/package.json` — Stale Certa remnants

`@itwin/certa` remains in `devDependencies` and `certa.json` appears to still be present. If the migration is complete, these should be removed to avoid dead weight and confusion about which test harness is canonical.

PR #9101, opened 2026-03-18
Several improvements to SettingsDb APIs

hl662:
Working on adding more tests than just happy path and skeleton assertions, discovered a few problems around SettingsDb/WorkspaceDb versioned lookups and editable cache behavior.

- Fix editable SettingsDb / WorkspaceDb caching so repeated requests reuse the same instance for the same resolved db, with regression coverage for local lookups, semver-equivalent selectors, and default-name/version cases.
- Honor requested versions when creating cloud SettingsDb / WorkspaceDb entries, and default omitted cloud versions to `0.0.0`.
- Add prototype guarding for `getSetting()` using `Object.hasOwn()`.
- Remove the circular dependency by extracting `SettingsEditorImpl.ts` from `SettingsImpl.ts`.
- Clarify editor docs around local `createEmptyDb(...)` path/overwrite behavior and resolved-db caching semantics.

### DIFF
--- a/common/changes/@itwin/core-backend/nam-settingsDb-improvements_2026-03-18-16-16.json
+++ b/common/changes/@itwin/core-backend/nam-settingsDb-improvements_2026-03-18-16-16.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/core-backend",
-      "comment": "Improve SettingsDb",
+      "comment": "Fix WorkspaceDb/SettingsDb instance caching, cloud version defaulting, and `getSetting` prototype-chain guard.",
       "type": "none"
     }
   ],


core/backend/src/workspace/WorkspaceEditor.ts
```diff
@@ -34,6 +34,8 @@ export namespace WorkspaceEditor {
 
   /**
    * Create a new, empty, [[EditableWorkspaceDb]] file on the local filesystem for importing [[Workspace]] resources.
+   * @note The caller is responsible for ensuring that `localFileName` refers to a safe, intended path.
```
ben-polinsky: This note is scary in its ambiguity. What are you getting at here?

core/backend/src/workspace/WorkspaceEditor.ts
```diff
@@ -34,6 +34,8 @@ export namespace WorkspaceEditor {
...
+   * This function will overwrite any existing file at that location.
```
ben-polinsky: Should we provide an option to not?

core/backend/src/internal/workspace/SettingsEditorImpl.ts
```diff
@@ -0,0 +1,309 @@
...
+} from "../../workspace/SettingsEditor";
+import { BlobContainer } from "../../BlobContainerService";
+import { settingsDbDefaultName, SettingsDbImpl, settingsManifestProperty } from "./SettingsDbImpl";
+import { SettingsSqliteDb } from "./SettingsSqliteDb";
+import { constructSettingsEditorWorkspace, OwnedWorkspace } from "./WorkspaceImpl";
+import { _implementationProhibited, _nativeDb } from "../Symbols";
+import { SettingsImpl } from "./SettingsImpl";
+import { IModelJsFs } from "../../IModelJsFs";
+
+// ==================== SettingsEditor implementation ====================
```
johnnyd710: nit: don't need this

core/backend/src/internal/workspace/SettingsEditorImpl.ts
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  public getEditableDb(props?: SettingsDbProps): EditableSettingsDb {
+    const key = this._cacheKey(props);
+    const dbName = props?.dbName ?? settingsDbDefaultName;
+    let db = this._settingsDbs.get(key);
+    if (undefined === db) {
+      db = new EditableSettingsDbImpl(props ?? { dbName }, this);
+      this._settingsDbs.set(key, db);
+    }
```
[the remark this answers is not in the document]
johnnyd710: I think copilot is right here ^, you are caching by *request* instead of by resolution

core/backend/src/workspace/WorkspaceEditor.ts
hl662: Yeah fair. I rewrote this to be more concrete: don’t pass an untrusted or unintended path in localFileName, and this helper creates or overwrites the file at that location.

core/backend/src/internal/workspace/SettingsEditorImpl.ts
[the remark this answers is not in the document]
hl662: fair point, this is fixed now to cache keys off the resolved db rather than the requested. Added tests. Also you brought up EditorContainerImpl doesn't use a version keyed cache and I added the same fix there too, so workspace dbs will get the same fix

core/backend/src/workspace/SettingsEditor.ts
```diff
@@ -166,6 +167,8 @@ export interface EditableSettingsCloudContainer extends CloudSqliteContainer {
 
   /**
    * Get an editable [[SettingsDb]] to add, delete, or update settings *within a newly created version* of a SettingsDb.
+   * Editable DBs are cached per `(dbName, version)` pair — requesting a different version of the same `dbName`
+   * returns a distinct instance.
```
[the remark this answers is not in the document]
hl662: Updated. The docs now describe caching in terms of repeated calls that resolve to the same SettingsDb returning the same cached instance, instead of saying it is cached per (dbName, version) pair.

core/backend/src/workspace/WorkspaceEditor.ts
hl662: I didn’t add an overwrite option in this PR since that would expand the public API. Instead I clarified the current behavior in the docs and noted that callers needing fail-if-exists behavior should check.

Of course, the API is @beta, so maybe?

core/backend/src/internal/workspace/SettingsEditorImpl.ts
hl662: thanks, removed

core/backend/src/internal/workspace/WorkspaceImpl.ts
```diff
@@ -142,28 +142,29 @@ class WorkspaceContainerImpl implements WorkspaceContainer {
...
-      return join(this.dirName, `${props.dbName}.${workspaceDbFileExt}`); // local file, versions not allowed
+      return join(this.dirName, `${dbName}.${workspaceDbFileExt}`); // local file, versions not allowed
 
-    return CloudSqlite.querySemverMatch({ ...props, container, dbName: workspaceDbNameWithDefault(props.dbName) });
+    return CloudSqlite.querySemverMatch({ ...props, container, dbName });
   }
 
-  public addWorkspaceDb(toAdd: WorkspaceDb) {
-    if (undefined !== this._wsDbs.get(toAdd.dbName))
+  public addWorkspaceDb(toAdd: WorkspaceDb, cacheKey: string) {
```
johnnyd710: I don't think cacheKey should be an arg here (or elsewhere). A cache key is something you calculate, an impl detail hidden to the end-user - can you just use toAdd.dbFilename?

hl662: when I made this change I thought we only had dbName, not dbFileName @@ 
Also thought it was okay since end user cannot access this method, its marked internal. But yes you right, I've dropped the arg.

common/changes/@itwin/core-backend/nam-settingsDb-improvements_2026-03-18-16-16.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Improve SettingsDb",
```
ben-polinsky: This comment should be more explicit about changes or else removed.

hl662: yep, got a concise brief comment on the biggest changes now as the comment.

PR #9102, opened 2026-03-19
Improve LocalHub's lock handling

### LINKED ISSUE #8749: HubMock doesn't handle locks properly

kring:
* Fixed #8749 
* Implemented @kabentley's improved algorithm for checking and updating changesets associated with particular element locks. This is described [here](https://github.com/iTwin/imodels-api/issues/224#issuecomment-3980847794). Keith believes (and I agree) the previous algorithm was incorrect. In real Hub, I believe this would result in locks being denied (pull required) more aggressively than necessary. This was less a problem in LocalHub due to other problems with the implementation, as described in #8749.
* Added tests to demonstrate how the new algorithm works under different scenarios.
* Made a clearer distinction between releasing and abandoning locks, though this isn't providing much value currently without #9045.

Just as a bit of history, I originally implemented this by building on top of #9045. Mostly because the new ServerBasedLocks tests in that PR are also a good way of testing these changes. Unfortunately, that will lead to conflicts in this PR when that PR is merged because of our "squash" merge strategy. I am strongly opposed to squash merging and will happily die on that hill if anyone wants to discuss it. 😁 But anyway, I rebased my changes on master for this PR.

### DIFF
--- a/common/config/rush/version-policies.json
+++ b/common/config/rush/version-policies.json
@@ -2,7 +2,7 @@
   {
     "policyName": "prerelease-monorepo-lockStep",
     "definitionName": "lockStepVersion",
-    "version": "5.8.0-dev.11",
+    "version": "5.8.0-dev.12",
     "nextBump": "prerelease"
   }
 ]
--- a/core/backend/package.json
+++ b/core/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-backend",
-  "version": "5.8.0-dev.11",
+  "version": "5.8.0-dev.12",
   "description": "iTwin.js backend components",
   "main": "lib/cjs/core-backend.js",
   "module": "lib/esm/core-backend.js",
--- a/core/bentley/package.json
+++ b/core/bentley/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-bentley",
-  "version": "5.8.0-dev.11",
+  "version": "5.8.0-dev.12",
   "description": "Bentley JavaScript core components",
   "main": "lib/cjs/core-bentley.js",
   "module": "lib/esm/core-bentley.js",
--- a/core/common/package.json
+++ b/core/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/core-common",
-  "version": "5.8.0-dev.11",
+  "version": "5.8.0-dev.12",
   "description": "iTwin.js components common to frontend and backend",
   "main": "lib/cjs/core-common.js",
   "module": "lib/esm/core-common.js",
--- a/core/ecschema-editing/package.json
+++ b/core/ecschema-editing/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-editing",
-  "version": "5.8.0-dev.11",
+  "version": "5.8.0-dev.12",
   "description": "ECSchema editing and validation API",
   "license": "MIT",
   "main": "lib/cjs/ecschema-editing.js",
--- a/core/ecschema-locaters/package.json
+++ b/core/ecschema-locaters/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-locaters",
-  "version": "5.8.0-dev.11",
+  "version": "5.8.0-dev.12",
   "description": "EC Schema file locaters",
   "license": "MIT",
   "main": "lib/cjs/ecschema-locaters.js",
--- a/core/ecschema-metadata/package.json
+++ b/core/ecschema-metadata/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/ecschema-metadata",
-  "version": "5.8.0-dev.11",
+  "version": "5.8.0-dev.12",
   "description": "ECObjects core concepts in typescript",
   "license": "MIT",
   "main": "lib/cjs/ecschema-metadata.js",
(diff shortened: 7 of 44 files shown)


core/backend/src/LocalHub.ts
```diff
@@ -649,11 +752,23 @@ export class LocalHub {
     this.db.saveChanges();
   }
 
+  public abandonLocks(locks: LockMap, arg: BriefcaseIdArg) {
```
MichaelSwigerAtBentley, line 755: Do we want to expose this in HubMock? It might be useful for some of the multi-user unit tests down the line.

kring: It is already in #9045. That PR also exposes a "safe" form of `abandonLocks` in the public API all the way up through `LockControl` and `TxnManager`.

Do you think it's worthwhile to bring the `HubMock` change alone over to this PR? I guess it depends how long that other PR is likely to be held up in review, and how useful this one is in isolation in the meantime.

MichaelSwigerAtBentley: Cool, just took at look. I don't have an immediate need for it, so it can stay in the other PR. 👍

Issue #8754, opened 2025-11-11
tool argument in IModelApp.toolAdmin.activeToolChanged sometimes undefined


adrianclinansmith:
https://github.com/iTwin/itwinjs-core/blob/d5cadb6d38838f74049f0b519d2519d5a027964e/core/frontend/src/tools/ToolAdmin.ts#L893

Code was throwing because tool is sometimes undefined. The signature doesn't show that, so I have to cast and check:

```ts
IModelApp.toolAdmin.activeToolChanged.addListener((tool) => {
    if (!(tool as any))
      return;
```

(We're not raising the event ourselves, just listening)

bbastings:
Should never be undefined. For example, in the case where there is no default tool, startPrimitiveTool calls the event listeners with the idleTool.

```
    this.onActiveToolChanged(undefined !== newTool ? newTool : this.idleTool, StartOrResume.Start);
```

There are no calls to onActiveToolChanged in core/frontend where tool should be undefined...need to see a callstack...

bbastings:
Seems like this might be what is happening...

App is causing startDefaultTool to be called from iModel onClose event (which documents that you should not call asynchronous methods).
The default tool fails to run because the viewports have been closed.
startPrimitiveTool is then called with undefined, which calls onActiveToolChanged with this.idleTool.
The only place that clears this._idleTool is ToolAdmin.onShutdown, the public get accessor this.idleTool just asserts it's defined.

Could do the following to protect against this scenario, but that post shutdown call should be addressed/removed.

```
-this.onActiveToolChanged(undefined !== newTool ? newTool : this.idleTool, StartOrResume.Start);
+const tool = newTool ?? this._idleTool;
+if (tool) // Protect against erroneous calls after onShutDown which sets this._idleTool to undefined...
+  this.onActiveToolChanged(tool, StartOrResume.Start);
```

hl662:
I've read the sentry logs, the stack traces point to quantityFormatter setUnitsProvider() raising an active tool changed event as well. Let me see how I can harden against this

hl662:
Oop, forgot to close this following #9115. Closing now

PR #9115, opened 2026-03-20
core-frontend: harden ToolAdmin and QuantityFormatter against post-shutdown races

hl662:
## Summary

Fixes a crash (#8754) where calling `QuantityFormatter.resetToUseInternalUnitsProvider()` as fire-and-forget during iModel close could race against `IModelApp.shutdown()` and cause an `activeToolChanged` event to be raised with an `undefined` tool, crashing any listener that accessed `tool.toolId`.

## Root cause

The shutdown sequence clears `ToolAdmin._idleTool` inside `onShutDown()`. If a detached async formatter reset completes after that point, `startPrimitiveTool(undefined)` accesses `this.idleTool` (an asserting getter) and then raises `activeToolChanged` with the result — which is `undefined` in production where assertions are disabled.

## Changes

**`ToolAdmin.startPrimitiveTool`** (`core/frontend/src/tools/ToolAdmin.ts`)
Skip the `activeToolChanged` emission if `_idleTool` has been cleared by `onShutDown()`. This prevents any listener from receiving an `undefined` tool.

**`QuantityFormatter.setUnitsProvider`** (`core/frontend/src/quantity-formatting/QuantityFormatter.ts`)
Remove the call to `startDefaultTool` when `IModelApp.initialized`. This cuts the race from happening entirely.

## Testing

Two regression tests added to `IModelApp.test.ts` (verified to fail before the fixes):
1. `startPrimitiveTool(undefined)` after `toolAdmin.onShutDown()` — must not emit `activeToolChanged`
2. `resetToUseInternalUnitsProvider()` after `IModelApp.shutdown()` — must not call `startDefaultTool`

## Notes

The consumer-side root cause (firing `void resetToUseInternalUnitsProvider()` from a sync `onClose` handler) remains a misuse pattern. This PR ensures `core-frontend` does not crash when that pattern occurs.

### DIFF
--- a/core/frontend/src/test/IModelApp.test.ts
+++ b/core/frontend/src/test/IModelApp.test.ts
@@ -200,16 +200,16 @@ describe("IModelApp startup tests", () => {
  */
 class NonBasicUnitsProvider {
   private readonly _delegate = new BasicUnitsProvider();
-  async findUnit(unitLabel: string, schemaName?: string, phenomenon?: string, unitSystem?: string): Promise<UnitProps> {
+  public async findUnit(unitLabel: string, schemaName?: string, phenomenon?: string, unitSystem?: string): Promise<UnitProps> {
     return this._delegate.findUnit(unitLabel, schemaName, phenomenon, unitSystem);
   }
-  async getUnitsByFamily(phenomenon: string): Promise<UnitProps[]> {
+  public async getUnitsByFamily(phenomenon: string): Promise<UnitProps[]> {
     return this._delegate.getUnitsByFamily(phenomenon);
   }
-  async findUnitByName(name: string): Promise<UnitProps> {
+  public async findUnitByName(name: string): Promise<UnitProps> {
     return this._delegate.findUnitByName(name);
   }
-  async getConversion(fromUnit: UnitProps, toUnit: UnitProps): Promise<UnitConversionProps> {
+  public async getConversion(fromUnit: UnitProps, toUnit: UnitProps): Promise<UnitConversionProps> {
     return this._delegate.getConversion(fromUnit, toUnit);
   }
 }


core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -700,8 +700,11 @@ export class QuantityFormatter implements UnitsProvider {
       return;
     }
 
-    // force default tool to start so any tool that may be using cached data will not be using bad data.
-    if (IModelApp.toolAdmin)
+    // Force default tool to start so any tool that may be using cached data will not be using bad data.
+    // Skip if IModelApp has already been shut down — startDefaultTool is not meaningful post-shutdown
+    // and the idle tool may have been cleared, which would cause startPrimitiveTool to emit an
+    // activeToolChanged event with an undefined tool.
+    if (IModelApp.toolAdmin && IModelApp.initialized)
```
bbastings: The IModelApp.toolAdmin check isn't doing anything...

  public static get toolAdmin(): ToolAdmin { return this._toolAdmin; }

core/frontend/src/tools/ToolAdmin.ts
```diff
@@ -1706,7 +1706,12 @@ export class ToolAdmin {
         await this.setPrimitiveTool(newTool);
       }
       // it is important to raise event after setPrimitiveTool is called
-      this.onActiveToolChanged(undefined !== newTool ? newTool : this.idleTool, StartOrResume.Start);
+      const activeTool = undefined !== newTool ? newTool : this._idleTool;
```
bbastings: Using the nullish coalescing operator is cleaner.

const activeTool = newTool ?? this._idleTool;

But yeah, adding this check as was suggested in the issue is a good idea.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
bsy-nicholasw, line 632: @hl662 Here is another issue. Setting units providers will cause `onInitialized` to get called again, but it doesn't cleanup the existing event listener.

core/frontend/src/tools/ToolAdmin.ts
hl662: Done, updated to `newTool ?? this._idleTool`.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: Thanks. Fixed by tearing down `_removeFormatsProviderListener` at the top of `onInitialized()` before re-registering. The cleanup already existed in `[Symbol.dispose]`, so I reused the same pattern there.

hl662:
> It's wrong that setUnitsProvider has the side effect of calling IModelApp.toolAdmin.startDefaultTool to begin with. I would expect this would be something apps normally call when an iModel is opened, when the active tool is a non-issue.

That's fair, I removed the `startDefaultTool` call entirely. It was a pre-existing design issue that this PR accidentally surfaced, and the guard I added was really just papering over it.
 
 The `onUnitsProviderChanged` event still fires so listeners can react, but restarting the active tool is now the caller's responsibility via `IModelApp.toolAdmin.restartPrimitiveTool()`. I added a `@note` to the JSDoc of both `setUnitsProvider` and `resetToUseInternalUnitsProvider` documenting this.
 
The `_idleTool` guard in `startPrimitiveTool` is still in place as a general safety net, but it's no longer the primary defense as we are solving the root cause which is the quantity formatter being able to start a default tool.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: Removed this if check entirely now that we don't call startDefaultTool when resetting to use an internal units provider

Issue #7316, opened 2024-11-04
Refine workspace APIs and settings persistence


pmconne:
# User problem

Administrators of infrastructure projects want to be able to configure how applications work, including the resources available for their use. We refer to this configuration of settings and resources as a "workspace". Those familiar with MicroStation complain of a lack of control over their workspace configuration. We want to offer them better, easier-to-understand tools when working with iTwins and iModels. For example, iTwin workspaces should meet the following requirements:
- Prohibit ordinary (non-admin) users from modifying the workspace.
- Permit reuse of common settings and resources across different contexts (e.g., all iModels within a specific iTwin; or all iTwins of a specific project type) without duplication.
- A single app in which administrators can configure workspaces across their organization with clear descriptions of available settings and possible values.
- Administrators can quickly push workspace configuration changes to their users without significantly disrupting their work.
- It must be possible to configure settings outside of the context of an iModel.

# General solution

We have implemented [workspace and settings](https://www.itwinjs.org/learning/backend/workspace/) APIs that store resources in versioned distributed cloud databases. Settings are configured at run-time to point to those resources. Settings can be defined at various levels, in order of ascending priority: defaults, application-supplied, organization, iTwin, branch, and iModel. Settings of levels iTwin and above can be stored in an iModel's property table and automatically loaded when the iModel is opened. Settings of organization level are not stored anywhere. Default and application-supplied settings are delivered with the application.

Really, there are two workspaces:
- `IModelHost.appWorkspace` comprising application defaults and organization-level settings; and
- `IModelDb.workspace` comprising the rest of the settings, in the context of a single iModel. It includes and in some cases overrides the settings defined in `IModelHost.appWorkspace`.

# Problems with current APIs and implementation

The current implementation contradicts some of the design goals.

- core-backend provides [Settings](https://www.itwinjs.org/reference/core-backend/workspace/settings/) APIs that any app or application can use to modify the in-memory contents of the workspace at run-time, violating the rule that only admins should have that ability.
- Any user (not just an admin) can modify the workspace by writing settings into an iModel's property table.
- iTwin- and branch-level settings must be written into the iModel's property table, because there is nowhere else to store them. This encourages duplication and risks them becoming out of sync. It would be very hard for an admin to update the iTwin-level settings across all iModels belonging to that iTwin - they would need to obtain the schema lock on each one, forcing all users to first push or abandon their local changes.
- There is no place to store organization-level settings.
[...]

PR #9116, opened 2026-03-23
iTwin workspace

johnnyd710:
Attempt to address some of the problems from https://github.com/iTwin/itwinjs-backlog/issues/1898 and https://github.com/iTwin/itwinjs-core/issues/7316

This PR adds a high-level API for persisting iTwin-scoped settings dictionaries through IModelHost, and aligns the iTwin workspace model with the existing iModel pattern.

```typescript
await IModelHost.getITwinWorkspace(iTwinId); // analogous to `IModelDb.workspace`
await IModelHost.saveSettingDictionary(iTwinId, settingName, value); // analogous to `IModelDb.saveSettingDictionary`
await IModelHost.deleteSettingDictionary(iTwinId, settingName); // analogous to `IModelDb.deleteSettingDictionary`
```

Instead of requiring callers to:
1. find *the* iTwin settings container (and what if it doesn't exist)
2. initialize it
3. acquire and release write locks
4. discover or create an editable prerelease SettingsDb
5. open and close the db
6. handle abandon/cleanup paths

> Why `IModelHost`? 
> I don't know if `IModelHost` is the right place for these APIs, but I ended up choosing it because (1) where else would we put it, there's no `ITwin` or `ITwinDb` equivalent to `IModelDb`? and (2) `appWorkspace` is already there. 
> Open to other suggestions

## Why

We already have a simple API for the iModel workspace, and two methods to persist settings scoped to an iModel:

```typescript
iModel.workspace;
iModel.saveSettingDictionary(name, dict);
iModel.deleteSettingDictionary(name, dict);
```

I'm introducing the "iTwin workspace" and attempting to match the iModel workspace's functionality. There are some key differences between the workspaces:

### `IModelHost.appWorkspace`
- Application-scoped.
- The "source-of-truth" is the JSON schema files registered at startup.
- Holds app/default settings and containers.

### `IModelDb.workspace`

- iModel-scoped.
- The "source-of-truth" is the dictionaries explicitly persisted into the iModel itself via `saveSettingDictionary(...)`.
- The settings themselves stored in the iModels are iModel-scoped, but you can store *pointers* to iTwin settings, which can be used to snapshot or persist iTwin settings into the iModel when that is desired (i.e., to support ISO 19650 workflows - for example the *latest* iTwin workspace might be on v2.1.7, but there may be an iModel on the project given to a contractor that you want to stay on v1.1.0, ~v1.9.0, etc.)

### `IModelHost.getITwinWorkspace(iTwinId)` (new)
- iTwin-scoped.
- The "source-of-truth" are the dictionaries persisted into the iTwin `SettingsDb` via `saveITwinSettingDictionary(...)`.
- Useful for admin tooling and iTwin settings management without requiring an iModel.

### DIFF
--- a/docs/changehistory/5.8.0.md
+++ b/docs/changehistory/5.8.0.md
@@ -100,8 +100,6 @@ Previously, settings and binary resources (fonts, textures, templates) were stor
 
 ##### Creating a local SettingsDb
 
-[[include:SettingsDb.createLocal]]
-
 See [SettingsDb]($docs/learning/backend/Workspace.md#settingsdb) for full documentation.
 
 #### Container type convention


core/backend/src/workspace/SettingsEditor.ts
```diff
@@ -61,6 +61,37 @@ export namespace SettingsEditor {
...
+
+  /**
+   * Query the [[BlobContainer]] service for the single settings container associated with a given iTwin.
+   * @returns The containerId of the single settings container, or `undefined` if no container exists.
+   * @throws if more than one settings container is found for the given iTwin.
+   */
+  export async function getITwinSingletonContainerId(iTwinId: GuidString): Promise<string | undefined> {
+    const containers = await queryContainers({ iTwinId });
+    if (containers.length > 1)
+      ITwinError.throwError<ITwinError>({
```
kabentley: I suggest you make a new interface/namespace called `ITwinSettingsErrors` (or something like that) so callers can know what the error keys are. Follow the pattern in ITwinCoreErrors.ts

core/backend/src/IModelHost.ts
```diff
@@ -309,6 +312,23 @@ class ApplicationSettings extends SettingsImpl {
   }
 }
 
+/**
+ * Settings for an iTwin. May only include settings priority for iTwin and organization.
+ */
+class ITwinWorkspaceSettings extends SettingsImpl {
+  protected override verifyPriority(priority: SettingsPriority) {
+    if (priority <= SettingsPriority.application)
+      throw new Error("Use IModelHost.appSettings");
```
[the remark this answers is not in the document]
kabentley: This should be an ITwinSettingError

core/backend/src/workspace/SettingsEditor.ts
johnnyd710: Done, added `ITwinSettingsError` with keys for some common failure points:
```ts
  export type Key =
    "failed-to-obtain-container-token" |
    "multiple-itwin-settings-containers" |
    "no-cloud-container" |
    "blob-service-unavailable" |
    "invalid-priority";
```

core/backend/src/workspace/Settings.ts
```diff
@@ -171,13 +172,13 @@ export interface SettingsDictionary {
  * @beta
  */
 export interface SettingsDictionarySource {
-  /** The name of the dictionary, which must be unique within its [[workspaceDb]], or - if [[workspaceDb]] is undefined - unique among all dictionaries not associated with any [[WorkspaceDb]]. */
+  /** The name of the dictionary, which must be unique within its [[workspaceDb]], or - if [[workspaceDb]] is undefined - unique among all dictionaries not associated with any source database. */
   readonly name: string;
-  /** The [[WorkspaceDb]] from which the dictionary originated. */
-  readonly workspaceDb?: WorkspaceDb;
+  /** The [[WorkspaceDb]] or [[SettingsDb]] from which the dictionary originated. */
+  readonly workspaceDb?: WorkspaceDb | SettingsDb;
```
hl662: this looks bad, if we know the API is in beta, and that we've had low (or minimal) adoption of these APIs I'm curious if we can reword this property to be better

core/backend/src/IModelHost.ts
```diff
@@ -309,6 +312,23 @@ class ApplicationSettings extends SettingsImpl {
...
+    if (priority <= SettingsPriority.application)
+      ITwinSettingsError.throwError("invalid-priority", { message: `Settings with priority ${priority} cannot be added to an iTwin workspace.` });
+    if (priority > SettingsPriority.iTwin)
+      ITwinSettingsError.throwError("invalid-priority", { message: `Settings with priority ${priority} cannot be added to an iTwin workspace.` });
+  }
+
+  public override * getSettingEntries<T extends Setting>(name: string): Iterable<{ value: T, dictionary: SettingsDictionary }> {
+    yield* super.getSettingEntries(name);
+    yield* IModelHost.appWorkspace.settings.getSettingEntries(name);
+  }
```
hl662, line 328: including application level settings entries as part of the returned values - implicitly - is risky if we don't disclose that to the user. I would be surprised if I found appWorkspace level settings included as part of getting iTwin settings entries, at least without a @note in the jsdoc for this method?

core/backend/src/IModelHost.ts
```diff
@@ -395,6 +415,142 @@ export class IModelHost {
...
+
+    return workspacePromise;
+  }
+
+  /** Save a [[SettingsDictionary]] for an iTwin workspace.
+   * If no iTwin settings container exists for `iTwinId`, one is created.
+   * The dictionary is stored directly in the root settings db under the supplied `name`, similar to [[IModelDb.saveSettingDictionary]].
+   * @beta
+   */
+  public static async saveITwinSettingDictionary(iTwinId: GuidString, name: string, dict: SettingsContainer): Promise<void> {
```
hl662: this and deleteITwinSettingDictionary seem to be near identical methods, except for the db.update/removeSetting method. Maybe extract into a shared private static method?

```ts
private static async withITwinSettingsDb(
   iTwinId: GuidString,
   action: (db: EditableSettingsDb) => void
 ): Promise<void> {
  ...
  settingsDb.open()
  action(settingsDb)
  settingsDb.close()
  ...
```

then save/deleteITwinSettingDictionary can become one liners:

```ts
 public static async saveITwinSettingDictionary(...) {
   return this.withITwinSettingsDb(iTwinId, (db) => db.updateSetting({ settingName: name, value: Setting.clone(dict) }));
 }
 public static async deleteITwinSettingDictionary(...) {
   if (await SettingsContainers.getITwinContainerId(iTwinId) === undefined) return;
   return this.withITwinSettingsDb(iTwinId, (db) => db.removeSetting(name));
 }
```

core/backend/src/IModelHost.ts
```diff
@@ -395,6 +415,142 @@ export class IModelHost {
...
+
+      const settingsDb = workspace.getSettingsDb({ containerId: container.fromProps.containerId, priority: SettingsPriority.iTwin, includePrerelease: true });
+
+      try {
+        for (const [dictName, dict] of Object.entries(settingsDb.getSettings())) {
+          if (typeof dict !== "object" || undefined === dict || null === dict || Array.isArray(dict))
+            continue;
+
+          const dictProps = { name: dictName, workspaceDb: settingsDb, priority: SettingsPriority.iTwin };
+          if (workspace.settings.dictionaries.some((dictionary) => dictionary.props.name === dictName))
```
hl662: what happens if there's another settings already loaded? Like at application level? Do we want to discard itwin level settings if the application already defines those settings?

core/backend/src/IModelHost.ts
```diff
@@ -395,6 +415,142 @@ export class IModelHost {
...
+
+      const problems: WorkspaceDbLoadError[] = [];
+      const container = workspace.getContainer(containerEntry);
+      const cloudContainer = container.cloudContainer;
+      if (undefined !== cloudContainer) {
+        try {
+          if ((!ProcessDetector.isMobileAppBackend && !ProcessDetector.isElectronAppBackend) || getOnlineStatus())
+            cloudContainer.checkForChanges();
+        } catch {
+          // must be offline or temporarily unreachable
```
hl662: I would at least use Logger.logWarning that we are offline?

core/backend/src/workspace/Settings.ts
johnnyd710: I'm not sure what you mean - you think we should hide `SettingsDb` until we get usage? `WorkspaceDb` is also in beta btw

core/backend/src/IModelHost.ts
johnnyd710: The iTwin workspace behaves [the same way as the iModel workspace](https://www.itwinjs.org/learning/backend/workspace/#imodel-settings), they both fallback to application-level defaults.

Its now [documented in the `Settings` interface](https://github.com/iTwin/itwinjs-core/pull/9116/commits/2bea298faade317382f5d509f71aa42a368733b8)

core/backend/src/IModelHost.ts
johnnyd710: `IModelHost.getITwinWorkspace("").settings.dictionaries` are separate from `IModelHost.appWorkspace.settings.dictionaries`.

These represent dictionaries explicitly added to the iTwin-workspace, the fall back you are thinking of only happens for reading a setting value

core/backend/src/IModelHost.ts
johnnyd710: added - since we don't know if it failed because we are online and `getOnlineStatus` lied to us or if there was some other error I added
```ts
Logger.logWarning(loggerCategory, `Failed to check for changes to iTwin settings container ${containerEntry.containerId}: ${String(error)}`);
```

core/backend/src/IModelHost.ts
johnnyd710: Good idea - [added a convenience method on `EditableSettingsCloudContainer` doing what you described](https://github.com/iTwin/itwinjs-core/pull/9116/commits/302d04e66cf51071d83a27795885f5656ddf14f4)

core/backend/src/workspace/Settings.ts
hl662: no, I mean, change the field name.... `cloudSqlLiteDb`?

johnnyd710: Ah I see what you mean. I could rename to something like `sourceDb` or I could introduce a second param called `settingsDb` to avoid any breaking changes. Considering both options atm

johnnyd710: https://github.com/iTwin/itwinjs-core/pull/9116/commits/d75e348ee4d20282719da15b15f358f2ad409d0e

core/backend/src/test/standalone/ITwinWorkspace.test.ts
```diff
@@ -0,0 +1,260 @@
...
+  });
+
+  afterEach(() => {
+    BlobContainer.service = savedBlobContainerService;
+    setOnlineStatus(true);
+    sinon.restore();
+  });
+
+  after(async () => {
+    await TestUtils.startBackend();
```
hl662: should this be in the after all hook? I imagine should have been shutdown, since we would be tearing down

core/backend/src/test/standalone/ITwinWorkspace.test.ts
```diff
@@ -0,0 +1,260 @@
...
+        provider: "azure" as const,
+        expiration: new Date(Date.now() + 3600000),
+        metadata: { containerType: "settings", label: containerId },
+        baseUri: "",
+      }),
+    };
+  }
+
+  beforeEach(async () => {
+    await TestUtils.shutdownBackend();
```
hl662, line 71: I assume you're using this to clear out any cached workspace db files from previous test suites?

johnnyd710: We shutdown and startup before each test so that we are working with clear empty state on IModelHost each time (app workspace, settings schemas, cloud cache, etc)

core/backend/src/test/standalone/ITwinWorkspace.test.ts
johnnyd710: other tests might assume IModelHost is already running - at least that's what I read in `SemanticRebase.test.ts`. is that not true?

hl662: got it, you're right.

docs/changehistory/NextVersion.md
```diff
@@ -103,6 +104,23 @@ Previously, settings and binary resources (fonts, textures, templates) were stor
 
 See [SettingsDb]($docs/learning/backend/Workspace.md#settingsdb) for full documentation.
 
+#### iTwin-scoped workspace
+
+This branch adds an end-to-end iTwin settings workflow in [@itwin/core-backend]($backend):
```
hl662:
```suggestion
An end-to-end iTwin settings workflow in [@itwin/core-backend]($backend) was added:
```

Inaccurate to use the phrase 'this branch' as part of a minor release note

docs/changehistory/NextVersion.md
```diff
@@ -103,6 +104,23 @@ Previously, settings and binary resources (fonts, textures, templates) were stor
 
 See [SettingsDb]($docs/learning/backend/Workspace.md#settingsdb) for full documentation.
 
+#### iTwin-scoped workspace
+
+Adds an end-to-end iTwin settings workflow in [@itwin/core-backend]($backend):
```
hl662:
```suggestion
Adds an end-to-end iTwin settings workflow in [@itwin/core-backend]($docs/reference/core-backend)
```
This should fix docs pipeline, if not, just using standard backticks will work. the $backend/<anyPkg> only works for API class/method names

full-stack-tests/backend/src/integration/CloudWorkspace.test.ts
```diff
@@ -379,5 +379,105 @@ describe("Cloud workspace containers", () => {
...
+      const resourceContainer = await editor.createNewCloudContainer({
+        metadata: { label: "itwin1-resource-container", description: "resource workspace for iTwin1" },
+        scope: { iTwinId: iTwin1Id },
+        manifest: { workspaceName: resourceWorkspaceName },
+      });
+
+      resourceContainer.acquireWriteLock("itwin-resource-admin");
+      const resourceVersion = (await resourceContainer.createNewWorkspaceDbVersion({ versionType: "major" })).newDb;
+      const resourceDb = resourceContainer.getEditableDb(resourceVersion);
+      resourceDb.open();
```
hl662, line 409: after seeing the wrapper you made for SettingsEditor, having a settingsContainer.withOpenDb, is there value in doing the same here on the EditableWorkspaceContainer, like resourceContainer.withEditableDb()?
```ts
 withEditableDb(args: {
   user: string;
   versionType: "major" | "minor" | "patch";
   operation: (db: EditableWorkspaceDb) => void;
 }): Promise<void>;
 ```

File: core/backend/src/internal/workspace/WorkspaceImpl.ts [hunk 13]
```diff
@@ -814,4 +833,6 @@ export function throwWorkspaceDbLoadErrors(message: string, wsLoadErrors: Worksp
 export interface OwnedWorkspace extends Workspace {
   /** Only the owner of a Workspace may close it. */
   close(): void;
```
ben-polinsky, line 841: Just to clarify - Is there any time when an owner must close the OwnedWorkspace on their own?

core/backend/src/workspace/SettingsDb.ts
```diff
@@ -6,117 +6,38 @@
...
-import { _implementationProhibited } from "../internal/Symbols";
+import type { WorkspaceDb, WorkspaceDbName } from "./Workspace";
 
-/** Metadata stored inside a [[SettingsDb]] describing the database's contents, to help users understand
- * the purpose of the [[SettingsDb]] and who to contact with questions about it.
- * @note Only the `settingsName` field is required, and users may add additional fields for their own purposes.
- * @note Since the information is stored inside the [[SettingsDb]], it is versioned along with the rest of the contents.
- * @beta
+/** The fixed db name used by the iTwin settings [[WorkspaceDb]].
+ * This remains a normal [[WorkspaceDb]] name. The load-all behavior is applied by
```
ben-polinsky: What does a "normal" WorkspaceDb name mean?

docs/changehistory/NextVersion.md
```diff
@@ -76,44 +73,41 @@ db.withQueryReader(query, (reader) => {
 });
 ```
 
-### Dedicated SettingsDb for workspace settings
+### iTwin settings workspace
 
-A new [SettingsDb]($backend) type has been added to the workspace system, providing a dedicated database for storing JSON settings as key-value pairs, separate from general-purpose [WorkspaceDb]($backend) resource storage.
+Applications can now store and load named settings dictionaries in an iTwin-scoped workspace.
```
ben-polinsky: I'd emphasize Applications and maybe give a brief nod to how it differs from iModel-level settings.

docs/learning/backend/Workspace.md
```diff
@@ -129,129 +129,128 @@ A [SettingsPriority]($backend) is just a number, but specific values carry seman
...
+```ts
+[[include:WorkspaceExamples.VersionAndPinITwinSettings]]
+```
 
-This two-layer design keeps settings and resources in separate containers with independent access control, versioning, and write locks. The [workspace resources](#workspace-resources) section below shows how to create and access `WorkspaceDb`s, and the [accessing workspace resources](#accessing-workspace-resources) section shows how settings point to them.
 
 ## Workspace resources
 
-"Resources" are bits of data that an application depends on at run-time to perform its functions. The kinds of resources can vary widely from one application to another, but some common examples include:
+So far we have covered settings — the name-value pairs that configure an application's behavior. But applications also depend on **resources**: binary data files like fonts, textures, images, and templates. This section explains how resources are stored and accessed.
```
ben-polinsky: Nit: second usage of "So far" to being a paragraph in this document.

ben-polinsky: I'd just go with "We've now covered..."

core/backend/src/workspace/SettingsDb.ts
johnnyd710: It meant conforming to WorkspaceDbName type but I shouldn't have mentioned it at all. Updated to: `The fixed db name used by [[WorkspaceDb]]'s meant to hold settings. [[Workspace.loadSettingsDictionary]] will treat rows in the 'strings' table as setting dictionary entries when [[WorkspaceDbSettingsProps.dbName]] matches this value.`

docs/learning/backend/Workspace.md
johnnyd710: done

docs/changehistory/NextVersion.md
johnnyd710: Changed to:
```md
### iTwin settings workspace

*Applications* can now store and load named settings dictionaries in an iTwin-scoped workspace, separate from iModel-level settings so the same values can be shared across iModels in that iTwin.

Under the hood, that workspace uses a [SettingsDb]($backend), which is a settings-formatted [WorkspaceDb]($backend) named `settings-db`. In that db, each string resource is one settings dictionary:

- Resource name: dictionary name
- Resource value: JSON dictionary content

Developers still read and write settings dictionaries by name, while container management, versioning, and cloud sync follow the standard workspace model.
```

core/backend/src/internal/workspace/WorkspaceImpl.ts
johnnyd710: Yes, whenever you are done with the itwin workspace you should close it, that will close all the containers the workspace opened (See `close` method on `WorkspaceImpl`)

I added a note to IModelHost.getITwinWorkspace jsDocs to make it more clear that you need to close it when you are done with it

hl662: what are the consequences of an owner not closing it? I.e a lock held on all the containers?

johnnyd710: Not closing a container keeps the SQLite connection open, increasing memory unnecessarily until the process exits when all connections exit. I think there's also a network socket that would remain open. So nothing critical but definitely not optimal.

core/backend/src/IModelHost.ts
```diff
@@ -395,6 +415,108 @@ export class IModelHost {
...
+
+  /** Delete a named [[SettingsDictionary]] from the iTwin's settings container.
+   * If no iTwin settings container exists, this method does nothing.
+   * @param iTwinId The iTwin whose settings container should be updated.
+   * @param name The name of the dictionary (resource name) to delete.
+   * @note uses [[IModelHost.userMoniker]] as the user name for acquiring the write lock on the settings container.
+   * @beta
+   */
+  public static async deleteSettingDictionary(iTwinId: GuidString, name: string): Promise<void> {
+    const containerId = await SettingsContainers.getITwinContainerId(iTwinId);
```
hl662: `getITwinContainerId` here queries `BlobContainer.service.queryContainersMetadata`, and then `constructForITwin` on line 510 calls `getITwinContainerId` again internally (SettingsEditorImpl.ts:27) — two identical network roundtrips per delete.

The guard on line 506 is necessary (without it, `constructForITwin` would *create* a new container just to delete from it). But the redundancy could be avoided by splitting `constructForITwin` into two variants:

- `constructForITwin(iTwinId)` — get-or-create (used by `saveSettingDictionary`)
- `getForITwin(iTwinId)` — get-only, returns `undefined` if no container exists (used by `deleteSettingDictionary`)

That way delete collapses to a single flow with no repeated queries, and the create-vs-get intent is explicit at the API boundary.

core/backend/src/IModelHost.ts
```diff
@@ -395,6 +415,108 @@ export class IModelHost {
...
+   * @note The returned [[OwnedWorkspace]] is caller-owned. Call [[OwnedWorkspace.close]] when finished.
+   * @beta
+   */
+  public static async getITwinWorkspace(containerProps: GetWorkspaceContainerArgs): Promise<OwnedWorkspace>;
+  /** @internal */
+  public static async getITwinWorkspace(args: GuidString | GetWorkspaceContainerArgs): Promise<OwnedWorkspace> {
+    const isContainerProps = typeof args !== "string";
+    const workspace = constructWorkspace(new ITwinWorkspaceSettings());
+
+    const workspacePromise = (async () => {
```
hl662: Nit: the IIFE + `.catch()` pattern here could be simplified to a plain `try/catch` since the method is already `async`:

```typescript
const workspace = constructWorkspace(new ITwinWorkspaceSettings());
try {
  const containerProps = isContainerProps ? args : await SettingsContainers.getITwinContainerProps(args);
  if (undefined === containerProps)
    return workspace;
  // ... rest of the logic ...
  return workspace;
} catch (error) {
  workspace.close();
  throw error;
}
```

(`catch` rather than `finally` is correct here — on success the caller takes ownership and is responsible for closing.)

core/backend/src/internal/workspace/WorkspaceImpl.ts
```diff
@@ -405,26 +393,37 @@ class WorkspaceImpl implements Workspace {
...
+        for (const resourceName of resourceNames) {
+          const manifest = db.manifest;
+          const dictProps: SettingsDictionaryProps = { name: resourceName, workspaceDb: db, priority: prop.priority };
+          // don't load if we already have this dictionary. Happens if the same WorkspaceDb is in more than one list
+          if (undefined === this.settings.getDictionary(dictProps)) {
+            const settingsJson = db.getString(resourceName);
+            if (undefined === settingsJson)
+              throwWorkspaceDbLoadError(`could not load setting dictionary resource '${resourceName}' from: '${manifest.workspaceName}'`, prop, db);
+
+            db.close(); // don't leave this db open in case we're going to find another dictionary in it recursively.
```
hl662: `db.close()` fires on every loop iteration now, but it was originally placed here to avoid holding the db open during the recursive `loadSettingsDictionary` call below. With the new multi-resource loop (for SettingsDbs), this causes N-1 unnecessary close/reopen cycles — `getString()` on the next iteration auto-reopens via `withOpenDb`.

Consider moving the close to just before the recursive call (line 425), or keeping the db open for the entire loop and closing once at the end:

```typescript
for (const resourceName of resourceNames) {
  const settingsJson = db.getString(resourceName);
  // ...
  if (nested !== undefined) {
    db.close(); // close only before recursing into other containers
    await this.loadSettingsDictionary(nested, problems);
  }
}
```

core/backend/src/test/standalone/ITwinWorkspace.test.ts
```diff
@@ -0,0 +1,249 @@
...
+    const iTwinId = Guid.createValue();
+    createLocalSettingsDb("itwin-settings-a", {
+      "app/testA": "value-a",
+      "app/testB": "value-b",
+    });
+    BlobContainer.service = createSettingsContainerService(iTwinId, ["itwin-settings-a"]);
+
+    await IModelHost.startup(opts);
+
+    const workspace = await IModelHost.getITwinWorkspace(iTwinId);
```
hl662, line 96: Minor: the `OwnedWorkspace` returned by `getITwinWorkspace` is never closed in these tests (lines 94, 105, 147, 186). Consider closing them in the assertion block or in `afterEach` to avoid leaking resources across test runs.

core/backend/src/test/standalone/ITwinWorkspace.test.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    BlobContainer.service = createSettingsContainerService(iTwinId, [containerId]);
+    await IModelHost.startup(opts);
+
+    const workspace = await IModelHost.getITwinWorkspace(iTwinId);
+    expect(workspace.settings.getString("app/testA")).to.equal("value-a");
+    expect(workspace.settings.getString("app/testB")).to.equal("value-b");
+  });
+
+  it("saveSettingDictionary saves a named dictionary and closes the editor", async () => {
```
hl662, line 199: The save/delete tests verify the happy-path orchestration nicely. One gap: there's no test for the error path — e.g., what happens when `withEditableDb` throws? The `finally { editor.close() }` in `saveSettingDictionary` / `deleteSettingDictionary` is critical for cleanup, but it's not exercised.

A quick addition like this would cover it:
```typescript
it("saveSettingDictionary closes editor even on error", async () => {
  const close = sinon.spy();
  sinon.stub(SettingsEditor, "constructForITwin").resolves({
    editor: { close } as any,
    container: { withEditableDb: sinon.stub().rejects(new Error("write failed")) } as any,
  });
  await expect(IModelHost.saveSettingDictionary(iTwinId, "x", {})).to.be.rejectedWith("write failed");
  expect(close.calledOnce).to.be.true;
});
```

johnnyd710: Thanks, added one for saving and one for deleting

core/backend/src/test/standalone/ITwinWorkspace.test.ts
johnnyd710: added cleanup

core/backend/src/internal/workspace/WorkspaceImpl.ts
johnnyd710: good catch, moved close() to the end of the func

core/backend/src/IModelHost.ts
johnnyd710: changed from IIFE to async try/catch

core/backend/src/IModelHost.ts
johnnyd710: Added getForITwin

core/backend/src/internal/workspace/WorkspaceImpl.ts
```diff
@@ -405,32 +393,42 @@ class WorkspaceImpl implements Workspace {
...
+          ? (() => {
+            const all = queryStringResourceNames(db);
+            // If resourceName is supplied, load it last so it has highest precedence at the same priority.
+            return prop.resourceName !== undefined
+              ? [...all.filter((name) => name !== prop.resourceName), prop.resourceName]
+              : all;
+          })()
+          : [prop.resourceName ?? settingsResourceName];
+
+        for (const resourceName of resourceNames) {
```
hl662, line 409: The `try/catch` at lines 395/428 wraps the entire inner `resourceNames` loop. For non-SettingsDbs this is fine (loop length is always 1), but for SettingsDbs with multiple dictionaries, one bad resource now prevents all subsequent siblings from loading — e.g., if `"appB/config"` is missing, `"appC/config"` never runs.

Consider moving the error handling inside the inner loop so each resource fails independently:

```typescript
for (const resourceName of resourceNames) {
  try {
    const manifest = db.manifest;
    const dictProps: SettingsDictionaryProps = { name: resourceName, workspaceDb: db, priority: prop.priority };
    if (undefined === this.settings.getDictionary(dictProps)) {
      const settingsJson = db.getString(resourceName);
      if (undefined === settingsJson)
        throwWorkspaceDbLoadError(...);
      this.settings.addJson(dictProps, settingsJson);
      // ... nested loading ...
    }
  } catch (e) {
    problems?.push(e as WorkspaceDbLoadError);
  }
}
```

The outer try/catch can remain for errors from `queryStringResourceNames` or the db open.

johnnyd710: Updated, thanks for the suggestion

johnnyd710: https://github.com/iTwin/itwinjs-core/pull/9116/commits/531fac7d1f18e1f2704c40256f73218367549148

core/backend/src/internal/workspace/WorkspaceImpl.ts
```diff
@@ -405,32 +393,46 @@ class WorkspaceImpl implements Workspace {
...
-          const dict = this.settings.getDictionary(dictProps);
-          if (dict) {
-            Workspace.onSettingsDictionaryLoadedFn({ dict, from: db });
-            // if the dictionary we just loaded has a "settingsWorkspaces" entry, load them too, recursively
-            const nested = dict.getSetting<WorkspaceDbSettingsProps[]>(WorkspaceSettingNames.settingsWorkspaces);
-            if (nested !== undefined) {
-              IModelHost.settingsSchemas.validateSetting<WorkspaceDbSettingsProps[]>(nested, WorkspaceSettingNames.settingsWorkspaces);
-              await this.loadSettingsDictionary(nested, problems);
+        const resourceNames = isSettingsDb(db)
+          ? (() => {
```
ben-polinsky: I'd prefer a named function - cleaner + much easier to debug.

core/backend/src/internal/workspace/WorkspaceImpl.ts
```diff
@@ -405,32 +393,46 @@ class WorkspaceImpl implements Workspace {
...
+            const all = queryStringResourceNames(db);
```
ben-polinsky: Part of me feels the condition below (name !== prop.resourceName) could go into `queryStringResourceNames`. Would that be useful other places? Otherwise, this is okay.

core/backend/src/internal/workspace/WorkspaceImpl.ts
```diff
@@ -629,6 +624,37 @@ class EditorContainerImpl extends WorkspaceContainerImpl implements EditableWork
...
+    this.acquireWriteLock(user);
+    try {
+      let db: EditableWorkspaceDb;
+      try {
+        db = this.getEditableDb({ ...props, includePrerelease: true });
+      } catch (error) {
+        if (!CloudSqliteError.isError(error, "already-published"))
+          throw error;
+
+        const newVersion = await this.createNewWorkspaceDbVersion({
```
ben-polinsky, line 640: Is it possible this.cloudContainer could still be undefined at this point?

johnnyd710: Since "already-published" error is only thrown if `this.cloudContainer` is defined, it'll never be undefined there.

core/backend/src/internal/workspace/WorkspaceImpl.ts
johnnyd710: Hmm I don't think it would be useful anywhere else other than loadSettingsDictionary tbh, its specific to loading a WorkspaceDbSettingsProps. Passing WorkspaceDbSettingsProps into queryStringResourceNames would feel wrong

core/backend/src/internal/workspace/WorkspaceImpl.ts
johnnyd710: I agree, updated

PR #9123, opened 2026-03-24
Fixed intersection of 2 tangent circles

saeeedtorabi:


core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
```diff
@@ -1917,6 +1917,12 @@ describe("CurveCurveIntersectXY", () => {
       AngleSweep.createStartEndDegrees(0, 360)
     );
     GeometryCoreTestIO.captureCloneGeometry(allGeometry, geometryB, x0);
+
+    // visualize the transformation of the problem to the intersection of unit circle and homogeneous ellipse
+    const c0 = Arc3d.createUnitCircle();
+    const c1 = Arc3d.create(Point3d.create(-0.4183722792273876, 1.2483930446909888), Vector3d.create(0.31624653205499503, 0.015624693816741206), Vector3d.create(-0.015624693816741225, 0.31624653205499503));
+    GeometryCoreTestIO.captureCloneGeometry(allGeometry, [c0, c1]);
```
saeeedtorabi: what's the reason for this visualization? I think it was for debug and you forgot to remove it? 

If you really want to keep it, please use x0. Currently it is visualized at (0,0) mixing with another test visual
Image

Also when you move it, note that the visualization for the original problem is far down right so I think transformation visual should be close to it

Image

core/geometry/src/numerics/Polynomials.ts
```diff
@@ -1224,27 +1224,38 @@ export class TrigPolynomial {
     PowerPolynomial.zero(coffs);
     let degree;
     // see core\geometry\internaldocs\unitCircleEllipseIntersection.md for derivation of these coefficients
-    if (Geometry.hypotenuseXYZ(axx, axy, ayy) > TrigPolynomial._coefficientRelTol * Geometry.hypotenuseXYZ(ax, ay, a)) {
```
saeeedtorabi, line 1227: so you reduced the original tol 1e-12 to 1-e24. why isnt it too small?

Also you said one could drive a truck through the Copilot tolerance 1e-10! Why is 1e-10 considered to be so large for this problem?

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
dassaf4: Please keep it.

* Definitely not using x0. The transformed intersection problem is centered at the origin, and it is highly useful to read off actual coordinates when debugging. It does not conflict with the other graphics there.
* Did not forget to remove. It is just like any other JSON we save.

core/geometry/src/numerics/Polynomials.ts
dassaf4: The original tol was not squared. Now it is. The inequality is the same, only everything is squared now (more efficient, sqrt is expensive).

The Copilot squared tol was 1.0e-10, which corresponds to 1.0e-5 distance, which is 10x bigger than the default metric tolerance!

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
saeeedtorabi: I still dont get the benefit of keeping it and think it makes visual confusing. We have 2 circles and when we transform the problem, we already knew that one circle is transformed to unit circle centered at origin and the other circle is still tangent to it.

Why is it highly useful to read off actual coordinates? It's not showing the values of axx, axy, and ayy which was important for the fix.

core/geometry/src/numerics/Polynomials.ts
saeeedtorabi: I see

core/geometry/src/test/curve/CurveCurveIntersectXY.test.ts
dassaf4: I was using it to understand the transformation. I found it very insightful. But I'm not dying on this hill.  You can remove it.

PR #9127, opened 2026-03-25
Introduce explicit `EditTxn` for iModel editing and deprecate legacy implicit-transaction mutators

kabentley:
`EditTxn` introduces an explicit transaction API for iModel write operations. It gives callers deliberate control over the start, scope, and end of a unit of work. This PR adds the new `EditTxn` API, the `withEditTxn(...)` helper, and txn-aware replacements for many existing mutating APIs across `IModelDb`, `Element`, `Model`, `ElementAspect`, `Relationship`, and domain-specific helpers.

Today, all existing write APIs work through an implicit transaction-unaware path. That model is convenient, but it also blurs transaction boundaries: unrelated edits can silently accumulate into the same unit of work and then be saved, abandoned, or reversed together. Explicit transactions make those boundaries deliberate so that save and abandon apply only to the intended scope of edits.

The migration from implicit writes to explicit transactions is intended to be incremental and backward-compatible by default.

Existing mutating APIs such as `iModel.elements.insertElement(...)`, `element.insert()`, and `iModel.saveChanges(...)` remain functional in this PR, but they are deprecated. Deprecated APIs continue to route through the legacy implicit write path so existing code does not need to migrate all at once.

This PR also adds `IModelHostOptions.implicitWriteEnforcement` to control how writes through the legacy implicit write path are handled:

- `"allow"` (default): implicit writes continue to work as before.
- `"log"`: implicit writes still succeed, but each use emits an error-level log with key `"implicit-txn-write-disallowed"`.
- `"throw"`: implicit writes fail with `EditTxnError` key `"implicit-txn-write-disallowed"`.

This allows applications to migrate in stages: preserve existing behavior with `"allow"`, inventory remaining implicit-write call sites with `"log"`, and switch to `"throw"` once write paths have been migrated.

To migrate a code path, wrap related writes in `withEditTxn(...)` or manage an `EditTxn` directly.

Before:

```ts
const id = iModel.elements.insertElement(props);
iModel.elements.insertAspect(aspectProps);
iModel.saveChanges("added element with aspect");
```

After:

```ts
const id = withEditTxn(iModel, "added element with aspect", (txn) => {
  const id = txn.insertElement(props);
  txn.insertAspect(aspectProps);
  return id;
});
```

Or, if the caller needs to control transaction lifetime directly:

```ts
const txn = new EditTxn(iModel, "added element with aspect");
txn.start();

try {
  const id = txn.insertElement(props);
  txn.insertAspect(aspectProps);
  txn.end();
  return id;
} catch (err) {
  if (txn.isActive)
    txn.end("abandon");

  throw err;
}
```

`withEditTxn(...)` creates an `EditTxn`, starts it, invokes the callback, saves on success, and abandons on error. For APIs that previously performed writes directly from an `IModelDb`, this PR adds txn-first-arg alternatives such as `insert(txn, ...)`, `update(txn, ...)`, and `delete(txn, ...)`. Some helper APIs also now accept an explicit transaction parameter when they may need to create or mutate persisted state.

Explicit transactions enforce their own rules independently of `implicitWriteEnforcement`. An explicit `EditTxn` must be active before it can perform writes. Explicit transactions cannot be nested: starting one while another explicit transaction is already active throws `EditTxnError("already-active")`. Starting one while unsaved changes (from a legacy implicit api) already exist throws `EditTxnError("unsaved-changes")`. These guards ensure clean transaction boundaries throughout the migration.

Beyond core element/model/relationship mutation, this PR also updates related write surfaces so they can participate in explicit transaction scopes, including schema import/drop flows, file-property writes, line-style helper creation paths, element-tree deletion helpers, annotation field-update flows, and backend/full-stack tests that exercise write behavior.


diegoalexdiaz (approved):

core/backend/test-output-schema-refactor.txt
```diff
@@ -0,0 +1,25 @@
...
++                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    + CategoryInfo          : NotSpecified: (:String) [], RemoteException
+    + FullyQualifiedErrorId : NativeCommandError
+ 
+[92m [0m[32m 0 passing[0m[90m (137ms)[0m
+[31m  1 failing[0m
+
+[0m  1) "before all" hook in "{root}":
+[0m[31m     Db is busy: Profile 
+[D:\dev\repos\itwinjs-core\core\backend\lib\cjs\test\.cache\profiles\default]
```
GytisCepk: Some of these new files look like something AI created for debugging and were committed by accident.

kabentley: yes, I see several copilot trash files - I will remove them. I also see several `tsconfig.tsbuildinfo` files here too. Shouldn't we gitignore them?

editor/backend/src/EditCommand.ts
```diff
@@ -67,9 +109,13 @@ export class EditCommand implements EditCommandIpc {
    * potentially showing the returned string to the user.
    */
   public async requestFinish(): Promise<"done" | string> {  // eslint-disable-line @typescript-eslint/no-redundant-type-constituents
-    this.onFinish(); // TODO: temporary, remove
+    await this.endEdits();
```
karolis-zukauskas: This will forcefully `saveChanges()` in the middle of work when someone from the Frontend attempts to start a new `EditCommand` via `EditTools.startCommand()`. Since the Frontend is chaotic, this will cause havoc.

Previously this worked by a "request loop" where if `command.requestFinish()` does not return "done", `EditCommandAdmin` throws `IModelStatus.ServerTimeout` which is caught by Frontends `EditTools.startCommand()` - this then wait a bit and tries again. Are we getting rid of this "random queue" mechanism?

editor/backend/tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
...216,402],[217,403],[199,404],[200,405],[215,406],[156,404],[160,407],[224,408],[225,409],[141,410],[226,411],[228,412],[229,413],[230,414],[231,415],[212,416],[232,417],[140,6],[165,418],[148,419],[233,420],[164,421],[146,414],[153,422],[145,423],[234,424],[144,425],[151,426],[221,427],[235,428],[143,429],[237,430],[223,431],[236,432],[239,433],[238,434],[149,435],[240,436],[220,437],[170,438],[179,6],[150,439],[241,440],[142,6],[242,441],[247,442],[147,443],[248,444],[246,445],[222,446],[186,447],[249,448],[250,422],[245,449],[251,6],[154,450],[252,6],[253,451],[254,452],[255,453],[271,454],[173,455],[264,453],[172,456],[265,457],[266,6],[176,458],[267,459],[269,460],[270,460],[268,461],[177,462],[207,463],[274,464],[175,465],[273,466],[272,467],[276,468],[275,469],[277,470],[174,6],[282,471],[280,6],[279,472],[281,473],[211,474],[201,475],[209,476],[202,477],[203,478],[210,479],[204,480],[208,481],[205,482],[206,483],[278,484],[481,6],[493,485],[482,221],[486,486],[484,6],[492,487],[491,488],[485,489],[483,6],[488,490],[487,491],[489,492],[490,492],[741,493],[740,494],[742,495],[739,496],[738,497],[737,6]],"affectedFilesPendingEmit":[[741,51],[740,51],[742,51]],"version":"5.6.2"}
```
karolis-zukauskas: dont want to commit this

core/backend/test-output-schema-refactor.txt
GytisCepk: In our case, `tsconfig.tsbuildinfo` shouldn't be in package root directory. They are placed in wherever `outDir` is specified for TS compiler (in our case it's `lib`, `lib/cjs` or `lib/esm`) and, as a result, are already covered by our `.gitignore`.

I assume, new `tsconfig.tsbuildinfo` were created by AI running `tsc` directly and not `pnpm build`. We can add additional gitignore line for this, but it may cause local repository to accumulate trash files as time goes on.

editor/common/src/EditorIpc.ts
```diff
@@ -44,4 +44,10 @@ export interface EditorIpc {
 export interface EditCommandIpc {
   /** Identify the current `EditCommand`'s name and version, optionally returning additional properties that describe its state. */
   ping: () => Promise<{ commandId: string, version: string, [propName: string]: any }>;
+  /** Save pending changes on this command's transaction.
+   * @param description Optional description saved with the changes.
+   */
+  saveChanges: (description?: string) => Promise<void>;
```
karolis-zukauskas, line 50: Perhaps this should be `SaveChangesArgs`?

core/backend/src/TxnManager.ts
```diff
@@ -1229,6 +1230,8 @@ export class TxnManager {
    * @note If numOperations is too large only the operations are reversible are reversed.
    */
   public reverseTxns(numOperations: number): IModelStatus {
+    if (this._nativeDb.hasUnsavedChanges())
```
karolis-zukauskas, line 1247: If an EditCommand is active (meaning it is actively computing and populating the current Txn), I would like to protect against other code that attempt to write, including undo/redo/save (perhaps even pull/push etc). Currently it seems like Frontend code could initiate an undo/redo while an EditCommand is working - this will lead to a partial Txn (or a crash because some changes got reverted that the EditCommand was aware of).

editor/backend/src/EditCommand.ts
kabentley: It is the "chaos" this PR attempts to resolve. EditCommands should be predictable when/how/if they save their changes. The default now is to save pending work when they end (at this line.) Any command that overrides `requestFinish` won't get that default and will have to determine whether to save pending changes. Every EditCommand has an EditTxn through which all changes should be made. That EditTxn cannot outlive it's EditCommand.

editor/backend/tsconfig.tsbuildinfo
kabentley: i deleted it and added tsconfig.tsbuildinfo to gitignore.

editor/common/src/EditorIpc.ts
kabentley: No, I don't think so. When EditCommands save their changes, they must store their commandId in the Txn record so rebasing will work. They can also save additional `appData` (stored on the EditCommand object itself.) There should not be any way to "lose" that. So, only the "undo message" can be supplied externally.

core/backend/src/TxnManager.ts
kabentley: Aren't EditCommands somehow notified when undo happens now (I don't know the answer)? This test is just the lower-enforcement that you can't reverse changes with pending (unsaved) changes.

bbastings: > Aren't EditCommands somehow notified when undo happens now (I don't know the answer)? This test is just the lower-enforcement that you can't reverse changes with pending (unsaved) changes.

@kabentley Not directly. The active primitive tool is called to restart (which calls finishCommand). From BriefcaseConnection:

      this.txns.onAfterUndoRedo.addListener(async () => { await IModelApp.toolAdmin.restartPrimitiveTool(); });

editor/backend/src/EditCommand.ts
karolis-zukauskas: > The default now is to save pending work when they end (at this line.)

Saving partial changes sounds like a bad idea. How should the software recover from or even identify such changes?

core/backend/src/TxnManager.ts
karolis-zukauskas: Even if an EditCommand is notified of an Undo - how is it supposed to react to that apart from abandoning everything it was working on?

A command will typically work on some Element Ids (e.g. update property of an Element) - I would like to be sure those elements aren't reverted in the middle of my command.

editor/backend/src/EditCommand.ts
karolis-zukauskas: For example - OpenSite+ has a command to create a "Parking Lot". That involves creating hundreds of elements - Authoring elements (edges, vertices), Physical elements (curbs, courses, surface markings), Plan elements (for drawings), a ton of Relationships, etc.

Current default allows to save a half created Parking Lot.

kabentley: i guess we could make abandon be the default? Do you think that will be more clear?

core/backend/src/TxnManager.ts
kabentley: It seems to me that if a user requests undo, the most sensible thing for the active command to do is abandon its pending changes. Otherwise they'll block undo. Do you think users will understand that?

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,681 @@
...
+ * - `log`: allow implicit writes but emit an error log for each usage.
+ * - `throw`: reject implicit writes.
+ *
+ * **During indirect changes (commit processing):** Use [[IModelDb.getIndirectChangesTxn]] to obtain the transaction
+ * in callbacks like [[Relationship.onRootChanged]] and [[Relationship.onDeletedDependency]] that fire during indirect transaction processing.
+ *
+ * Compatibility rule: implicit transaction writes must continue to work when enforcement is `allow`.
+ * @beta
+ */
+export class EditTxn {
```
rschili, line 36: "Edit" is the only kind of Txn there is, no?
In that case, we could call it just Txn or Transaction.
There is some inconsistency across the caller methods withEditTxn and withTxn.

rschili (approved):
Didn't do a deep code review, just general decisions + mechanism.

core/backend/src/EditTxn.ts
kabentley: Please read my comment at the top of EditTxn.ts. The problem is that there is *always* a SQLite transaction active (we call it the "default transaction"), for READ consistency. The idea is to introduce EditTxn to mean a transaction that can make changes.

There's also the further complication naming-wise that we already have the term `Txn` in our api to mean "the persistent record of the changes made from a call to `saveChanges`. Now, the appropriate explanation would be "a `Txn` is the result of an `EditTxn`".

You're right, I suppose the proper name for the functions that replace the nicely-named-but-now-deprecated functions would be xxxWithEditTxn.

core/backend/src/IModelDb.ts
```diff
@@ -2811,25 +2900,15 @@ export namespace IModelDb {
     /** Insert a new element into the iModel.
      * @param elProps The properties of the new element.
      * @returns The newly inserted element's Id.
-     * @throws [[ITwinError]] if unable to insert the element.
+     * @throws Error if insertion fails.
      * @note For convenience, the value of `elProps.id` is updated to reflect the resultant element's id.
      * However when `elProps.federationGuid` is not present or undefined, a new Guid will be generated and stored on the resultant element. But
      * the value of `elProps.federationGuid` is *not* updated. Generally, it is best to re-read the element after inserting (e.g. via [[getElementProps]])
      * if you intend to continue working with it. That will ensure its values reflect the persistent state.
+     * @deprecated Use EditTxn.insertElement instead, within an explicit EditTxn scope (or via withEditTxn). See EditTxn documentation for migration help.
```
kring, line 2845: I don't think we should mark this (and others) deprecated while the replacement is still marked beta. That seems to be setting users up for churn. I think it would be better to:
1. Just add a note to the documentation along the lines of "consider using EditTxn" rather than officially deprecating (yet), or
2. If we're confident in the new API, make it public from the start.

core/backend/src/IModelDb.ts
```diff
@@ -81,6 +82,48 @@ import { ECSqlSyncReader, SynchronousQueryOptions } from "./ECSqlSyncReader";
...
+      }
+
+      throw err;
+    }
+  }
+
+  public override abandonChanges(): void {
+    this.iModel.clearCaches({ instanceCachesOnly: true });
+    this.iModel[_nativeDb].abandonChanges();
+  }
```
kring: As far as I can tell, the only reason for this override is to avoid calling `verifyWriteable`, which in turn makes it ok to call `abandonChanges` even when implicit transaction write is disallowed. But I don't understand why this is desirable. For example, in `core/backend/src/test/standalone/DisplayStyle.test.ts`, the "preserves skybox test" has a call to the deprecated `abandonChanges`, even though the rest of the test uses an explicit EditTxn. That seems like a potential bug: some unruly code calls `abandonChanges` and ends up abandoning changes managed by some other EditTxn, even while the user thinks they are operating under the security of `implicitWriteEnforcement="throw"`.

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,681 @@
...
+ * An explicit EditTxn lets callers define a deliberate unit of work by choosing when editing
+ * starts (`start`) and how it ends (`end()` / `end("save")` or `end("abandon")`). This avoids mixing
+ * unrelated edits into one implicit unit of work and makes save/rollback boundaries explicit.
+ *
+ * Explicit EditTxn instances must be active before mutating operations are performed, regardless of enforcement level.
+ * In other words, explicit transaction behavior is independent of `implicitWriteEnforcement`.
+ *
+ *
+ * *During indirect changes (commit processing):* Use [[IModelDb.getIndirectChangesTxn]] to obtain the EditTxn
+ * in callbacks like [[Relationship.onRootChanged]] and [[Relationship.onDeletedDependency]] that fire during indirect processing.
```
kring: Why not pass the `EditTxn` to these callbacks? That seems cleaner than the back door `getIndirectChangesTxn` that is only usable in certain circumstances.

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,681 @@
...
+    this.verifyWriteable();
+    this.iModel.saveFilePropertyImpl(prop, strValue, blobVal);
+  }
+
+  /** Update the project extents of the iModel.
+   * @param newExtents The new project extents.
+    * @throws EditTxnError if this EditTxn is not active.
+    * @throws IModelError if extents are invalid.
+   */
+  public async updateProjectExtents(newExtents: Range3dProps): Promise<void> {
```
kring: There's a lot going on in this method (and in `updateEcefLocation` below) that wasn't in the the now-deprecated versions in IModelDb. Why? Maybe the reason is obvious to some, but it's not to me. 

Also, mostly unrelated to this PR, the original code looks dangerous:

```
    this.iModel.projectExtents = extents;
    this.iModel.updateIModelProps();
```

What if the next line is `this.iModel.abandonChanges`? The DB will have one project extent while the `IModelDb` instance will have another. With a brief look, I can't see any place the `projectExtents` property is updated or invalidated in that scenario. I don't think the new implementation in EditTxn changes this situation.

core/backend/src/IModelDb.ts
```diff
@@ -1179,7 +1248,7 @@ export abstract class IModelDb extends IModel {
    */
   public performCheckpoint() {
     if (!this.isReadonly) {
-      this.saveChanges();
+      this[_nativeDb].saveChanges();
```
kring, line 1249: You probably already confirmed this is fine, but mentioning it just in case: `this.saveChanges()` was previously doing some checks that `this[_nativeDb].saveChanges()` will not. For example, if the native saveChanges returns an error code, it will now be silently ignored.

core/backend/src/IModelDb.ts
```diff
@@ -1226,7 +1295,8 @@ export abstract class IModelDb extends IModel {
    * @throws [IModelError]($common) if the database if the operation failed.
    * @alpha
    */
-  public async dropSchemas(schemaNames: string[]): Promise<void> {
+  /** @internal */
+  public async dropSchemasImpl(schemaNames: string[]): Promise<void> {
```
kring: I can't see any reason for this new Impl method to exist. Used in a previous iteration perhaps?
If we do need to keep it, it should probably be named using a symbol, and the `@alpha` in the doc should be removed.

core/backend/src/IModelDb.ts
```diff
@@ -1433,14 +1513,15 @@ export abstract class IModelDb extends IModel {
...
+   * @throws [[EditTxnError]] if the schema lock cannot be obtained or there is a problem importing the schema.
    * @note Changes are saved if importSchemas is successful and abandoned if not successful.
    * - You can use NativeLoggerCategory to turn on the native logs. You can also control [what exactly is logged by the loggers](https://www.itwinjs.org/learning/common/logging/#controlling-what-is-logged).
    * - See [Schema Versioning]($docs/bis/guide/schema-evolution/schema-versioning-and-generations.md) for more information on acceptable changes to schemas.
    * @note This method should not be called from {TxnManager.withIndirectTxnModeAsync} or {RebaseHandler.recompute}.
    * @see querySchemaVersion
    */
-  public async importSchemas(schemaFileNames: LocalFileName[], options?: SchemaImportOptions): Promise<void> {
+  /** @internal */
+  public async importSchemasImpl(schemaFileNames: LocalFileName[], options?: SchemaImportOptions): Promise<void> {
```
kring: Same here, this `Impl` method doesn't appear to be necessary.

core/backend/src/IModelDb.ts
```diff
@@ -1433,14 +1513,15 @@ export abstract class IModelDb extends IModel {
    * You must import a schema into an iModel before you can insert instances of the classes in that schema. See [[Element]]
    * @param schemaFileName  array of Full paths to ECSchema.xml files to be imported.
    * @param {SchemaImportOptions} options - options during schema import.
-   * @throws [[IModelError]] if the schema lock cannot be obtained or there is a problem importing the schema.
+   * @throws [[EditTxnError]] if the schema lock cannot be obtained or there is a problem importing the schema.
```
kring: I believe `EditTxnError` here is wrong. Or at least it's different from `importSchemas` below, which still says it throws `IModelError`.

core/backend/src/IModelDb.ts
```diff
@@ -1451,18 +1532,38 @@ export abstract class IModelDb extends IModel {
...
    * @note Changes are saved if importSchemaStrings is successful and abandoned if not successful.
    * @note This method should not be called from {TxnManager.withIndirectTxnModeAsync} or {RebaseHandler.recompute}.
    * @see querySchemaVersion
    * @alpha
    */
-  public async importSchemaStrings(serializedXmlSchemas: string[], options?: SchemaImportOptions): Promise<void> {
+  /** @internal */
+  public async importSchemaStringsImpl(serializedXmlSchemas: string[], options?: SchemaImportOptions): Promise<void> {
```
kring: And another unnecessary Impl with incorrect exception doc.

core/backend/src/IModelDb.ts
```diff
@@ -1473,6 +1574,31 @@ export abstract class IModelDb extends IModel {
...
+   * @note This method should not be called from {TxnManager.withIndirectTxnModeAsync} or {RebaseHandler.recompute}.
+   * @see querySchemaVersion
+   * @alpha
+   */
+  public async importSchemaStrings(serializedXmlSchemas: string[], options?: SchemaImportOptions): Promise<void> {
+    await this.importSchemaStringsImpl(serializedXmlSchemas, options);
+  }
+
+  /** @internal */
+  public saveFilePropertyImpl(prop: FilePropertyProps, strValue: string | undefined, blobVal?: Uint8Array): void {
```
kring: This one _is_ used, from EditTxn. It should be hidden using a symbol.

core/backend/src/IModelDb.ts
```diff
@@ -2811,25 +2900,15 @@ export namespace IModelDb {
     /** Insert a new element into the iModel.
      * @param elProps The properties of the new element.
      * @returns The newly inserted element's Id.
-     * @throws [[ITwinError]] if unable to insert the element.
+     * @throws Error if insertion fails.
```
kring: Did you mean to replace `ITwinError` with `Error` here? Also in `updateElement` below.

core/backend/src/IModelDb.ts
```diff
@@ -4361,13 +4327,11 @@ export class SnapshotDb extends IModelDb {
...
-    if (this._createClassViewsOnClose) { // check for flag set during create
-      if (BentleyStatus.SUCCESS !== this[_nativeDb].createClassViewsInDb()) {
-        throw new IModelError(IModelStatus.SQLiteError, "Error creating class views");
-      } else {
-        this.saveChanges();
-      }
-    }
+    if (this._createClassViewsOnClose) {
+      withEditTxn(this, "create class views", () => {
+        this[_nativeDb].createClassViewsInDb();
```
kring: Previously, the return value of `createClassViewsInDb` was checked and an exception was thrown on error. Now an error code is silently ignored. Intentional?

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,681 @@
...
+    const iModel = this.iModel;
+    if (iModel.openMode === OpenMode.Readonly)
+      throw new IModelError(IModelStatus.ReadOnly, "IModelDb was opened read-only");
+
+    if (iModel.isBriefcaseDb() && iModel.txns.isIndirectChanges)
+      throw new IModelError(IModelStatus.BadRequest, "Cannot save changes while in an indirect change scope");
+
+    args ??= this.description;
+    const saveArgs = typeof args === "string" ? { description: args } : args;
+    const stat = iModel[_nativeDb].saveChanges(JSON.stringify(saveArgs));
```
kring, line 148: I asked Copilot to compare all of the methods that moved from IModelDb.ts to EditTxn.ts for any copy/paste errors or other differences. It noted that the behavior here changed slightly:

> The old code passed `undefined` to `nativeDb.saveChanges()` when no description was provided:
>
> The new code uses `args ??= this.description` (which for `ImplicitWriteTxn` defaults to "implicit"), so it always passes a JSON string:
>
> This means callers who previously passed no description will now get `{"description":"implicit"}` sent to native instead of `undefined`.

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,681 @@
...
+  private static readonly _settingPropNamespace = "settings";
+  private static readonly _viewStoreProperty: FilePropertyProps = { namespace: "itwinjs", name: "DefaultViewStore" };
+
+  /** Save a `SettingDictionary` in this iModel.
+   * @param name The name for the SettingDictionary. If a dictionary by that name already exists, its value is replaced.
+   * @param dict The SettingDictionary object to stringify and save.
+   * @throws EditTxnError if this EditTxn is not active.
+   * @beta
+   */
+  public saveSettingDictionary(name: string, dict: Record<string, unknown>): void {
```
kring: It also noted that `dict`'s type was `SettingsContainer` in `IModelDb`, instead of `Record<string, unknown>`. Intentional?

core/backend/src/EditTxn.ts
```diff
@@ -0,0 +1,681 @@
...
+    this.verifyWriteable();
+    const usageInfo = this.iModel[_nativeDb].queryDefinitionElementUsage(definitionElementIds);
+    if (!usageInfo)
+      throw new IModelError(IModelStatus.BadRequest, "Error querying for DefinitionElement usage");
+
+    const usedIdSet = usageInfo.usedIds ? Id64.toIdSet(usageInfo.usedIds) : new Set<Id64String>();
+    const deleteIfUnused = (ids: Id64Array | undefined, used: Id64Set): void => {
+      ids?.forEach((id) => {
+        if (!used.has(id))
+          this.deleteElement(id);
```
kring, line 295: Copilot pointed out that calling `this.deleteElement` here will cause a redundant call to `verifyWriteable` for every ID. Probably not a big problem, except that in `ImplicitWriteEnforcement` `log` mode there will potentially be many redundant warnings logged.

kring (commented):
I've mostly just looked at the changes to `IModelDb` and the new `EditTxn` class so far, and have a few observations and questions below. Overall it looks great! Compared to the original implementation that I looked at, I really appreciate that `EditTxn` is now meant to be a thing that is passed around, rather than a thing that is inherited from. That's a really big improvement in my mind.

core/backend/src/TxnManager.ts
karolis-zukauskas: Abandoning changes makes sense, but we also need to cancel "work". I mean the JavaScript will continue to execute code inside the EditCommand and pollute the now cleared Txn. Perhaps the `EditTxn` used by EditCommand could be marked as "dirty" on undo/redo and throw on attempt to write (or even read) - Frontend code could handle this error case - perhaps notify user that "operation was cancelled".

OS+ needs to be aware when this happens and be able to react since iModel is not the only state it modifies - it would need to also revert changes made on the Java (Siteops) process side.

editor/backend/src/EditCommand.ts
karolis-zukauskas: > Previously this worked by a "request loop" <...>

Morning clarity - I am so used to how we implemented EditCommands in OS+ that I forgot how the default implementation in core works (which returns "done" by default"). In OS+ we decided that 1 EditCommand == 1 Txn and another EditCommand cannot start if the active one did not do a commit/abandon, see [TxnEditCommandBase.requestFinish](https://github.com/iTwin/OpenSitePlus/blob/cdad38854a77407766a68c84f386d4f69fb092df/packages/authoring-support-backend/src/commands/TxnEditCommand.ts#L150) and [SiteEditCommand.isBusy](https://github.com/iTwin/OpenSitePlus/blob/cdad38854a77407766a68c84f386d4f69fb092df/apps/booster/src/backend/commands/SiteEditCommand.ts#L58). Perhaps it is fine to not have this behaviour in core and leave it for applications to implement if they need to.

> i guess we could make abandon be the default?

Yes - seems safer to me.

kabentley: abandonEdits is now the default.

core/backend/src/TxnManager.ts
kabentley: I think there should be another event that notifies the active command *before* undo, but I will leave that to another PR by somebody else.

core/backend/src/IModelDb.ts
kabentley: you're 100% correct. I don't know why that was added, but it is wrong. I removed it and fixed the test.

core/backend/src/IModelDb.ts
kabentley: Sorry, I think we *should* leave this as `@deprecated`/`@beta`. That's not at all uncommon. 

Not deprecating the implicit api will mean nobody will do anything to clean up their code because they won't be aware there's anything wrong, and that must happen soonest. Certainly nobody should be adding new code without considering the transaction implications.

I don't consider the new `EditTxn` api stable enough that it should become unchangeable. I hope and doubt it will need much change, but I'm certainly not ready to warrant that it will need none.

This is a fundamental and painful transition. It will only become more painful the longer we allow it to continue unchecked.

core/backend/src/EditTxn.ts
kabentley: I suppose that would be preferable, perhaps. I considered it, but decided to take this approach because adding arguments to callbacks requires native code changes and I didn't want to tie this PR to a native one. The "certain circumstances" where it works is the only circumstances where it's needed.

core/backend/src/IModelDb.ts
kabentley: Yes, the checkpoint operation referred to here is a SQLite operation, and it will not work if there are unsaved changes. This is fine.

core/backend/src/IModelDb.ts
kabentley: fixed

core/backend/src/EditTxn.ts
kabentley: The new behavior is preferable. We should have never allowed empty commit messages.

core/backend/src/EditTxn.ts
kabentley: fixed

core/backend/src/EditTxn.ts
kabentley: This is fine.

core/backend/src/IModelDb.ts
kabentley: I'm nearly certain this whole codepath is unused. The "views" referred to are SQLite views that we created at the very beginning of the project so you could look at your data with "pure" SQLite browsers. Throwing an exception on closing a Snapshot would have never helped anyway.

core/backend/src/IModelDb.ts
kabentley: removed

core/backend/src/IModelDb.ts
kabentley: removed

core/backend/src/IModelDb.ts
kabentley: fixed

core/backend/src/IModelDb.ts
kabentley: removed

core/backend/src/IModelDb.ts
kabentley: moved to EditTxn

core/backend/src/EditTxn.ts
kabentley: This method, and `updateProjectExtents` were moved from `EditBuiltinCommands.ts`. 

It didn't make sense to me to have two different implementations for the same concept, particularly when introducing a new api for it. I defer to @bbastings for their implementations.

bbastings: @kring @kabentley The "extra" code in updateProjectExtents was necessary when fixing the project extents for a connector so that an update/sync would know to not blow away the change. Similarly the stuff in updateEcefLocation was for a user/admin to remove an invalid GCS. We still need "clean" versions of these w/o the extra stuff.

> The DB will have one project extent while the IModelDb instance will have another.

The "project extents" tools that called the BasicManipulationCommand methods also called restartTxnSession right after saveChanges to avoid issues with undo. It also warns you beforehand if you have other changes you'd no longer be able to undo so you can cancel.

kabentley: @bbastings are you saying that the methods on EditTxn for updating project extents and ecefLocation should **not** have the "goop" i copied over from EditBuiltinCommand?

bbastings: @kabentley Correct, no "goop" on EditTxn, just want that stuff for the interactive tools.

kabentley: put them back to original implementations

core/backend/src/IModelDb.ts
kring: I see what you're saying, and I don't really even disagree with you. But I have trouble squaring that with the Studio update model. Number 1 publishing tip in the Studio documentation: [only use public APIs](https://qa-itwinstudio.bentley.com/getting-started/supporting-docs/publish-best-practices/#1-use-public-itwinjs-apis-only). Because if you use Beta APIs, Studio can auto-update and suddenly your app is running on a version of Studio that doesn't include that Beta API anymore (or it changed in a breaking way).

Making developers choose between deprecation warnings and the possibility that their application breaks on end-user systems later? That's not very nice.

This may be mostly a theoretical problem today, but as Studio becomes more widely used I think it's going to become a real problem, so we should think through it. My opinion: aggressive (but thoughtful!) API evolution is reasonable and necessary; the Studio update model is the problem here.

core/backend/src/EditTxn.ts
kring: If I'm not mistaken, the native code always invokes these callbacks via the protected `_onRootChanged` and `_onDeletedDependency` methods on `TxnManager`. So an extra property or parameter could be added by those methods without any need to change the native side.

But unfortunately the only parameter passed to the callbacks currently (other than the `IModelDb`) is `RelationshipProps` rather than a more general "args" instance. I don't think adding an `EditTxn` to `RelationshipProps` makes sense, so we'd be stuck introducing a new parameter. Or maybe changing the type of the first parameter to `RelationshipProps & { editTxn: EditTxn }`, but that seems worse.

I think I can make a case that it's worth it (keep IModelDb API surface smaller, provide users information they will need directly rather than make them go looking for it), but only just barely, so I won't argue (further! 😁) if you want to keep it as-is.

core/backend/src/EditTxn.ts
kring: > The "project extents" tools that called the BasicManipulationCommand methods also called restartTxnSession right after saveChanges to avoid issues with undo. It also warns you beforehand if you have other changes you'd no longer be able to undo so you can cancel.

Good to hear it's handled! Though I'll note that handling it at a higher level does not eliminate the risk posed by the way the lower-level API is structured. The doc doesn't mention this potential problem, and the example in the doc doesn't call `saveChanges`. `restartTxnSession` is marked `@internal`, so I guess it can't be called at all outside our own code. We have bigger fish to fry for sure. 😁

core/backend/src/IModelDb.ts
kring: IModelDb.insertElement and updateElement look good, but the same methods in EditTxn (plus deleteElement) still say `@throws Error if insertion fails` (or similar).

I'm happy to just push a fix for this nitpicky stuff rather than adding a comment. But I don't want to edit your PR without being sure you're ok with that, so let me know what you prefer.

core/backend/src/IModelDb.ts
kabentley: Studio should definitely *not* recommend against using @beta apis. Otherwise the whole purpose of @beta is lost - giving people a way to give feedback before something is finalized. How would we ever get anyone to switch from one api to another? They surely have no way of enforcing that anyway.

Marking something @beta is not a license to make breaking changes without careful consideration, and only as a last resort. In this case we want and need people to switch aggressively, but I'm certain we'll find issues that need to be addressed in the new apis.

I see the much bigger problem throughout the iTwin apis that we don't seem to promote from @beta to @public very aggressively or consistently.

core/backend/src/EditTxn.ts
kabentley: You're right! I forgot how I implemented relationship callbacks (in my defense, it was in 2018). Element callbacks are direct from native, but the relationship ones take a detour through TxnManager because the relationship class lookup isn't available in native code. Since I thought I'd have to change native code I didn't really consider the choice of changing the callback signatures.

It is possible, but I'm still not convinced it is right. For it to be useful, the value of IndirectTxn should not be optional. That would mean deprecating the current callback methods on Relationship (onRootChanged and onDeletedDependency) and adding new ones that would call the old ones by default. Subclasses would implement the new one. I'd like feedback on whether that feels better than requiring the new magic method on IModelDb. Does OS+ implement relationship callbacks @karolis-zukauskas? What's your vote?

common/api/core-backend.api.md
```diff
@@ -3908,11 +4026,13 @@ export abstract class IModelDb extends IModel {
...
+    // @deprecated
     saveChanges(description?: string): void;
-    // @alpha
+    // @alpha @deprecated
     saveChanges(args: SaveChangesArgs): void;
+    // @deprecated
     saveFileProperty(prop: FilePropertyProps, strValue: string | undefined, blobVal?: Uint8Array): void;
-    // @beta
+    // @beta @deprecated
     saveSettingDictionary(name: string, dict: SettingsContainer): void;
```
hl662, line 4024: @johnnyd710 Heads up — this PR deprecates IModelDb.saveSettingDictionary and moves it to 
 EditTxn.saveSettingDictionary. Your PR #9116 writes to a separate cloud SettingsDb so no
 functional conflict, but worth being aware of when merging.

hl662:
This PR needs a docs/changehistory/NextVersion.md entry for the EditTxn API. It's a major
 new pattern across 11 packages (54 new methods, 78 deprecations) — consumers need to
 understand what EditTxn is, the implicit→explicit migration path, the 
 implicitWriteEnforcement levels, and maybe one before/after code example.

common/api/core-backend.api.md
johnnyd710: Ah thanks for the heads up. my `IModelHost.saveSettingsDictionary` == `IModelDb.saveSettingsDictionary` loses some of the equivalence I wanted but I think its still fine

core/backend/src/IModelDb.ts
kckst8: Studio currently allows usage of beta apis but not without warning from the linter. I believe that app teams won't use them for fear of linter noise. Removing that noise would be risky and likely result in breaking changes that are not planned for properly, but perhaps we could highlight beta differently and add a policy that allows them to be changed in minors but not patches (Studio currently only upgrades to minor versions the first 2 weeks of the month, leaving 2 weeks for app testing pre-release and patches), or something along those lines?

kabentley: It's fine for there to be a lint warning about using @beta apis,, that makes sense - disallowing them wouldn't be right I don't think.

If I'm an app developer working on switching over from a now-deprecated to the now-beta replacement api, I'm going to be likely working piecemeal: silence the deprecation warnings with a TODO note about my new plan, then removing the deprecated calls with the new @beta ones (silencing their warnings with a TODO), finally after the new api goes @public removing the TODOs and any workaround code. That's the life of an app developer (or an agent working on their behalf.)

Api developers should be aware that making new code @beta means people will start using it (that is the goal!) and care should be taken to minimize churn unless it is unavoidable. That's the life of an api developer (or agents working on their behalf.)

Then, @beta api should be promoted to @public when they're deemed stable, they shouldn't stay @beta for extended cycles.

kabentley: One perhaps unexpected side effect of this approach is that it becomes possible for something to be both @beta and @deprecated. That actually happened here with the `settingsDictionary` api that @hl662 and @johnnyd710 observed below. Note that I purposely did not remove the beta-now-deprecated code for this reason.  

There's no excuse for that code to have existed for so long as @beta, but that's what has happened here.

core/backend/src/IModelDb.ts
kabentley: the ones in EditTxn.ts should now be fixed too.

kabentley:
> This PR needs a docs/changehistory/NextVersion.md entry for the EditTxn API. 

@hl662 @aruniverse I added a Nextversion.md and EditTxn.md learning articles. Let me know what you think

editor/frontend/src/ModifyElementTool.ts
```diff
@@ -104,7 +105,7 @@ export abstract class ModifyElementTool extends ElementSetTool {
 
   public override async processAgenda(ev: BeButtonEvent): Promise<void> {
     if (await this.applyAgendaOperation(ev))
-      return this.saveChanges();
```
bbastings, line 107: ModifyElementTool is an abstract base class that doesn't start an EditCommand, sub-classes choose the EditCommand. So we can't change this saveChanges call...

bbastings:
> Nextversion.md and EditTxn.md learning articles. Let me know what you think

These look good.

bbastings (approved):

core/backend/src/IModelDb.ts
kring: I'm not sure if I'm expressing my concern very well, so let me try again.

There are the facts as I see them:
1. Studio apps use `@beta` APIs.
2. `@beta` APIs can change and even be removed from one version to the next. For good reasons, but it can happen. Allowing for this possibility is the entire point of marking an API `@beta`.
3. Studio is updated automatically on end-user systems, and brings along a new version of iTwin.js. Applications that are already installed will immediately start running on top of the new version of Studio / iTwin.js.

Therefore:
A new version of iTwin.js that changes a `@beta` API risks causing a _runtime error on end-user systems_. The only mitigation I can think of for this is:
1. Don't change any APIs (`@beta` or otherwise) that are used by any Studio app, or
2. Synchronize the release schedule of _all_ Studio applications with the release schedule of Studio itself, so that breaking changes in iTwin.js can be accounted for in the application before Studio forces those breaking changes upon it.

Breaking a previously-working application on an end-user's system with an automatic update is about the worst problem I can imagine, so I'm pretty paranoid about this.

kring: I'm going to mark this conversation resolved since it's not directly related to this PR and doesn't need to be resolved before merging it.

core/backend/src/EditTxn.ts
kring: I don't think it should be necessary to add new methods because TS (unlike C++) allows an overriding method to have fewer parameters than the overridden one.

I'm going to leave this conversation unresolved until Karolis has a chance to chime in, but either of you should feel free to resolve it when you're ready, no need to wait for me to do it.

core/backend/src/annotations/TextAnnotationElement.ts
```diff
@@ -767,7 +775,13 @@ export class AnnotationTextStyle extends DefinitionElement {
 
     // Copy the style into the target iModel and remap its Id.
     const dstStyleProps = await context.cloneElement(srcStyle);
-    dstStyleId = context.targetDb.elements.insertElement(dstStyleProps);
+    const activeTxn = context.targetDb[_activeTxn];
+    if (undefined !== activeTxn) {
+      dstStyleId = activeTxn.insertElement(dstStyleProps);
+    } else {
+      dstStyleId = withEditTxn(context.targetDb, "Import AnnotationTextStyle", (txn) => txn.insertElement(dstStyleProps));
```
kring: The pattern I see throughout most of this PR is:
1. A new `doWhateverWithTxn` is added.
2. The old `doWhatever` is deprecated, and calls the above with the implicit EditTxn.

But here it's a bit different. The operation happens in whatever EditTxn happens to be active (this is a privilege user code does not have at all!), or, if there isn't one, the operation is done in a new EditTxn which is immediately committed.

Is this intentional? I don't have much context for how this method is used, but on the surface it seems like a fairly significant change in behavior.

core/backend/src/internal/ChannelAdmin.ts
```diff
@@ -114,9 +124,18 @@ class ChannelAdmin implements ChannelControl {
     if (this.queryChannelRoot(args.channelKey) !== undefined)
       ChannelControlError.throwError("root-exists", `Channel ${args.channelKey} root already exist`, args.channelKey);
 
-    const elementId = Subject.insert(this._iModel, args.parentSubjectId ?? IModel.rootSubjectId, args.subjectName, args.description);
-    this.makeChannelRoot({ elementId, channelKey: args.channelKey });
-    return elementId;
+    const activeTxn = this._iModel[_activeTxn];
```
kring: This is the other case I noticed that helps itself to the active EditTxn, or creates and immediately commits one if there is none active.

core/backend/src/EditTxn.ts
karolis-zukauskas: > Does OS+ implement relationship callbacks

No. I think @Rock2000 was looking at using them when implementing command rebase (?) - perhaps lack of async support was the reason to not use them. In general, we don't use and try to avoid other "domain handler" callbacks, like `onUpdate`, `onInserted`, etc.

> What's your vote?

I don't like static getters - it is unclear when to call them, what is available to use and they can be used in wrong places. If we are already deprecating a bunch of Element CRUD surface - lets improve Relationship part as well (i.e. pass EditTxn as parameter to those callbacks, if I got the discussion right).

core/backend/src/internal/ChannelAdmin.ts
kabentley: fixed

core/backend/src/annotations/TextAnnotationElement.ts
kabentley: fixed

core/backend/src/EditTxn.ts
kabentley: I added the editTxn to relationship callbacks. Element-based callbacks will still need to use the (internal) getIndirectChanges until a future PR adds the editTxn to the callbacks from native code.

docs/changehistory/NextVersion.md
```diff
@@ -3,3 +3,72 @@ publish: false
...
+[EditTxn.implicitWriteEnforcement]($backend) and [IModelHostOptions.implicitWriteEnforcement]($backend) control how legacy implicit writes behave while callers migrate:
+
+| Value | Behavior |
+| ----- | -------- |
+| `"allow"` | Preserve existing behavior and allow implicit writes. |
+| `"log"` | Allow implicit writes, but emit `implicit-txn-write-disallowed` errors to help inventory remaining migration work. |
+| `"throw"` | Reject implicit writes and require explicit [EditTxn]($backend) usage. |
+
+These levels are intended to support incremental adoption. Applications can start with `"allow"`, move to `"log"` to discover remaining legacy paths, and then switch to `"throw"` once those call sites have been migrated.
+
```
hl662, line 74: I'd love to see a little ending sentence that links this document to the EditTxn learning doc you created, like I did in here: https://github.com/iTwin/itwinjs-core/blob/master/docs/changehistory/5.6.0.md?plain=1#L38

hl662 (commented):
The documentation is really clear, thanks! I think we can improve the discoverability of these docs by adding a link to it, in the index page of docs/backend. My first thought was maybe somewhere under [iModelDb](https://github.com/iTwin/itwinjs-core/blob/ec8ae57425027f7c3c6897adb9382b066bdd0b0f/docs/learning/backend/index.md?plain=1#L24) heading?

I have a local MCP server that indexes the itwin docs site, my goal is to improve the discoverability of our docs so agentic AI, and devs using it, will always know what it can do with our APIs.

docs/changehistory/NextVersion.md
kabentley: done

hl662 (approved):

hl662:
Picked this up and pushed a few fixes to get CI green:

 - Docs build: removed duplicate @packageDocumentation tag in TxnProps.ts, and replaced 
broken doc links in importSchemas TSDoc (NativeLoggerCategory is @internal, console 
appender path was wrong) — both fatal under --strictLinkChecking.
 - Merged master: resolved conflicts with the iTwin settings workspace feature that landed
 in the meantime.
 - Lint: migrated three iModel.saveSettingDictionary() calls in doc examples to 
withEditTxn, since this branch deprecates the old pattern.

saskliutas (approved):
Presentation changes look good. Cleaned up tests a little bit.

ben-polinsky (approved):

core/backend/src/EditTxn.ts
hl662: @kring is this comment good to resolve?

core/backend/src/Element.ts
```diff
@@ -533,14 +574,43 @@ export class Element extends Entity {
...
+   */
+  public insertWithTxn(txn: EditTxn) {
+    return this.id = txn.insertElement(this.toJSON());
+  }
+
+  /**
+   * Update this Element in the iModel using the supplied transaction.
+   * @beta
+   */
+  public updateWithTxn(txn: EditTxn): void {
```
kring: At the risk of being annoying...

Previously, users would modify an `Element`, then call `update()` on it to actual modify the iModel. Simple enough.

Now, that's deprecated, and they have to call `updateWithTxn()`. In the context of this PR, that seems reasonable enough. But from an API design standpoint, that's a pretty strange method name. Am I updating my `Element` with a `Txn` somehow? What does that mean?

If we did this from the beginning, we obviously would have just called this method `update`, even though it takes an `EditTxn` as a parameter. In other languages, we could do an overload.

But wait (do I sound like Copilot yet?), we can do that here, too, right? TS supports overloading, it just has to be combined with a runtime check. Wouldn't that be better than a new method called `updateWithTxn`? Something like this (not tested):

```ts
/** Update this Element in the iModel.
 * @deprecated Use Element.updateWithTxn instead.
 */
public update(): void;

/**
 * Update this Element in the iModel using the supplied transaction.
 * @beta
 */
public update(txn: EditTxn): void;

public update(txn?: EditTxn): void {
    (txn ?? this.iModel[_implicitTxn]).updateElement(this.toJSON());
}
```

I think this would be a better pattern elsewhere, too. Or am I just being annoying?

core/backend/src/ElementTreeWalker.ts
```diff
@@ -174,7 +176,18 @@ function logModel(op: string, iModel: IModelDb, modelId: Id64String, scope?: Ele
...
  */
 export abstract class ElementTreeBottomUp {
-  constructor(protected _iModel: IModelDb) { }
+  private readonly _txn?: EditTxn;
+
+  protected get txn(): EditTxn {
+    if (undefined !== this._txn)
+      return this._txn;
+
+   return this._iModel[_activeTxn] ?? this._iModel[_implicitTxn];
```
kring: This is another case where something automatically inserts itself into the activeTxn, which seems wrong to me.

core/backend/src/ElementTreeWalker.ts
```diff
@@ -242,6 +255,15 @@ class SpecialElements {
   public definitions: Id64Array = [];
   public subjects: Id64Array = [];
 
+  public constructor(private readonly _txn?: EditTxn) { }
+
+  private getTxn(iModel: IModelDb): EditTxn {
+    if (undefined !== this._txn)
+      return this._txn;
+
+    return iModel[_activeTxn] ?? iModel[_implicitTxn];
```
kring: Here, too.

core/backend/src/ElementTreeWalker.ts
```diff
@@ -457,6 +485,8 @@ export class ElementSubTreeDeleter extends ElementTreeTopDown {
 export interface DeleteElementTreeArgs {
   /** The iModel containing the elements to delete. */
   iModel: IModelDb;
+  /** The EditTxn used to perform the deletes. Defaults to the current active transaction. */
```
kring: Assuming the use of the active txn isn't intentional, this comment needs to be updated, too.

kring (commented):
I've now (finally?) looked at all of the non-test code changes.

core/backend/src/Element.ts
karolis-zukauskas: Yes please

kabentley: excellent idea. I didn't realize you could have a deprecated overload of a method. I'll (tell copilot to) give it a try.

markschlosseratbentley:
Since this changes a significant number of files, I am kicking off an Image Tests run. Will post back when it finishes.
cc @eringram

markschlosseratbentley:
The Image Tests results look good. 👍

markschlosseratbentley (approved):

core/backend/src/Element.ts
kabentley: i made this change. Please review (it is big).

core/backend/src/ElementTreeWalker.ts
kabentley: fixed

core/backend/src/ElementTreeWalker.ts
kabentley: fixed

core/backend/src/ElementTreeWalker.ts
kabentley: fixed

core/backend/src/CodeSpecs.ts
```diff
@@ -177,17 +160,52 @@ export class CodeSpecs {
...
+    txnOrCodeSpec: EditTxn | CodeSpec | string,
+    codeSpecOrNameOrProps?: CodeSpec | string | CodeSpecProperties | CodeScopeSpec.Type,
+    props?: CodeSpecProperties | CodeScopeSpec.Type,
+  ): Id64String {
+    if (txnOrCodeSpec instanceof EditTxn) {
+      const txn = txnOrCodeSpec;
+      if (txn.iModel !== this._imodel)
+        EditTxnError.throwError("wrong-imodel", "EditTxn does not belong to this iModel", txn.iModel.key);
+      // CodeSpec insertion writes directly to SQLite, so enforce txn writability explicitly.
+      txn.verifyWriteable();
```
kring, line 176: In the same way that all the DB mutation methods in IModelDb moved to EditTxn, perhaps `insertCodeSpec` should move to EditTxn, too?

Related to that, I see there's also a `CodeSpec.updateProperties` below and it is not Txn aware. I think that's a bug.

core/backend/src/ElementTreeWalker.ts
```diff
@@ -177,16 +177,22 @@ function logModel(op: string, iModel: IModelDb, modelId: Id64String, scope?: Ele
...
 
-  constructor(protected _iModel: IModelDb, txn?: EditTxn) {
-    this._txn = txn;
+  constructor(iModelOrTxn: IModelDb | EditTxn) {
+    if (iModelOrTxn instanceof EditTxn) {
+      this._txn = iModelOrTxn;
+      this._iModel = iModelOrTxn.iModel;
+    } else {
+      this._iModel = iModelOrTxn;
+    }
```
kring: Assuming there's not any harm in grabbing the implicit txn at construction time (I can't think of any), this can be simplified to something like this:

```ts
constructor(iModelOrTxn: IModelDb | EditTxn) {
    this._txn = iModelOrTxn instanceof EditTxn ? iModelOrTxn : iModelOrTxn[_implicitTxn];
}
```

And we can get rid of the extra `_iModel` field.

kring (commented):
I looked at all the recent changes to non-test code. Some of those overloads got a lot more complicated than my simple example! 😅  But I still think it's a good change and it all looks correct to me.

I noticed two more small things that I noted below.

core/backend/src/CodeSpecs.ts
kabentley: fixed

core/backend/src/ElementTreeWalker.ts
kabentley: fixed

core/backend/src/IModelDb.ts
kring: And now I'm going to unresolve this comment. This is useful context, and github has a bug that prevents us from linking directly to resolved comments.

PR #9132, opened 2026-03-25
Clear non-null assertions in frontend packages

ben-polinsky:
Continuing with https://github.com/iTwin/itwinjs-core/issues/8376- cleaning up code caught by non-null assertions lint rule.

Frontend and somewhat associated packages.

### DIFF
--- a/common/changes/@itwin/core-electron/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
+++ b/common/changes/@itwin/core-electron/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-electron",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-electron"
+}
--- a/common/changes/@itwin/core-frontend/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
+++ b/common/changes/@itwin/core-frontend/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-frontend"
+}
--- a/common/changes/@itwin/core-mobile/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
+++ b/common/changes/@itwin/core-mobile/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-mobile",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-mobile"
+}
--- a/common/changes/@itwin/frontend-devtools/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
+++ b/common/changes/@itwin/frontend-devtools/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/frontend-devtools",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/frontend-devtools"
+}
--- a/common/changes/@itwin/webgl-compatibility/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
+++ b/common/changes/@itwin/webgl-compatibility/benpolinsky-non-null-frontend-and-friends_2026-03-27-13-48.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/webgl-compatibility",
+      "comment": "",
+      "type": "none"
+    }
+  ],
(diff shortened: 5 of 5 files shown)


core/electron/src/frontend/ElectronApp.ts
```diff
@@ -70,7 +70,9 @@ export class ElectronApp {
       this._ipc = new ElectronIpc();
       ElectronRpcManager.initializeFrontend(this._ipc, opts?.iModelApp?.rpcInterfaces); // eslint-disable-line @typescript-eslint/no-deprecated
     }
-    await NativeApp.startup(this._ipc!, opts);
+    const ipc = this._ipc;
+    assert(undefined !== ipc);
+    await NativeApp.startup(ipc, opts);
```
tcobbs-bentley: I feel that `expectDefined` and `expectNotNull` (from core-bentley) should be used instead of assert.

Note: the below isn't a suggestion since imports need to change.
```ts
    await NativeApp.startup(expectDefined(this._ipc), opts);
```

tcobbs-bentley: Note: I'm only tagging this one instance of `assert` in this PR, but the same probably applies to all of them.

core/frontend-devtools/src/widgets/GpuProfiler.ts
```diff
@@ -101,7 +101,11 @@ export class GpuProfiler {
   private _isRecording: boolean;
 
   public constructor(parent: HTMLElement) {
-    this._debugControl = IModelApp.renderSystem.debugControl!;
+    const debugControl = IModelApp.renderSystem.debugControl;
+    if (undefined === debugControl)
+      throw new Error("Render system debug control is not available.");
+
+    this._debugControl = debugControl;
```
tcobbs-bentley: You aren't using `assert`, but this is a good place for `expectDefined`, with a custom message.

core/frontend-devtools/src/widgets/GpuProfiler.ts
```diff
@@ -201,12 +205,12 @@ export class GpuProfiler {
...
-        const savedResults = this._results[index];
-        if (savedResults.values.length >= numSavedFrames) { // keep up to numSavedFrames values to average between
-          oldVal = savedResults.values.shift()!;
-        }
+        } else { // Edit old entry
+          let oldVal = 0.0;
+          const savedResults = this._results[index];
+          if (savedResults.values.length >= numSavedFrames) { // keep up to numSavedFrames values to average between
+            oldVal = savedResults.values.shift() ?? 0.0;
+          }
```
tcobbs-bentley: These new indents appear to be wrong.

core/frontend-devtools/src/widgets/GpuProfiler.ts
```diff
@@ -224,7 +228,7 @@ export class GpuProfiler {
 
     this._results.forEach((value, index) => {
       if (!changedResults[index]) { // if no data received on this item, add a value of 0.0 to the avg.
-        const oldVal = value.values.length >= numSavedFrames ? value.values.shift()! : 0.0;
+        const oldVal = value.values.length >= numSavedFrames ? (value.values.shift() ?? 0.0) : 0.0;
```
tcobbs-bentley: I'm not sure what this code is doing, but having `shift` fail due to an empty `values` array and then pushing 0.0 below seems like it is probably not correct behavior.

core/mobile/src/backend/MobileRpcServer.ts
```diff
@@ -91,7 +91,11 @@ export class MobileRpcServer {
         return;
       }
 
-      this._connection!.send(message, (err) => {
+      const connection = this._connection;
+      if (undefined === connection)
+        throw new IModelError(BentleyStatus.ERROR, "No connection.");
+
+      connection.send(message, (err) => {
```
tcobbs-bentley: Also a potential spot for expectDefined, although that throws Error and not IModelError, so maybe not.

core/mobile/src/backend/MobileRpcServer.ts
```diff
@@ -106,8 +110,12 @@ export class MobileRpcServer {
     if (this._pendingMessages === undefined)
       return;
 
+    const connection = this._connection;
+    if (undefined === connection)
+      throw new IModelError(BentleyStatus.ERROR, "No connection.");
```
tcobbs-bentley: And again.

core/frontend-devtools/src/tools/FrustumDecoration.ts
```diff
@@ -404,7 +404,11 @@ class ShadowFrustumDecoration {
   }
 
   public decorate(context: DecorateContext): void {
-    const frustum = this._targetVp.target.debugControl!.shadowFrustum;
+    const debugControl = this._targetVp.target.debugControl;
+    if (undefined === debugControl)
```
GytisCepk: `const frustum = this._targetVp.target.debugControl?.shadowFrustum;`

Would this prevent us from having additional undefined check ?

tcobbs-bentley: I agree; just changing the original `!` to `?` should fix the problem.

core/frontend-devtools/src/widgets/GpuProfiler.ts
ben-polinsky: It's the old indents which are wrong, but I'll align.

tcobbs-bentley: I don't see anything wrong with the indents on master.

ben-polinsky: Yeah, not sure what I was looking at... anyway fixed.

PR #9133, opened 2026-03-25
mitigate CVE-2026-33671 via rush update --full

ben-polinsky:
Mitigate https://github.com/advisories/GHSA-c2c7-rcm5-vvqj 

rush update --full


ben-polinsky:
@grigasp could you approve when you get in?

Issue #8376, opened 2025-07-30
Fix or disable all lint warnings


pmconne:
#8272 made `@typescript-eslint/no-non-null-assertion`, `@typescript-eslint/restrict-template-expressions`, etc into warnings. The output of `npm run lint` is now unusable. For example, running it in core-frontend produces 582 problems.

Non-null assertions should be treated as an error - overly-optimistic exclamation marks have been the cause of countless bugs/crashes in production. We should fix all the violations. If that will take more than, say, a week, we should disable the warnings until the violations are fixed. If we think other rules like restrict-template-expressions aren't really a big deal, we should not warn on them.

@augustasgri @aruniverse 

[lint.txt](https://github.com/user-attachments/files/21510649/lint.txt)

aruniverse:
Thanks for creating this, I agree re `no-non-null-assertions` and should look to incrementally get this down to 0.

+ @paulius-valiunas

tcobbs-bentley:
I'll start whittling things down.

tcobbs-bentley:
FYI, this PR gets things started: https://github.com/iTwin/itwinjs-core/pull/8391

ben-polinsky:
Good work has been done on this (see above PRs), but many assertions remain. 
Can pick up when bandwidth allows.

PR #9134, opened 2026-03-25
Clear non-null assertions in leaf packages

ben-polinsky:
Continuing with https://github.com/iTwin/itwinjs-core/issues/8376 - cleaning up code caught by non-null assertions lint rule.

- core-markup
- linear-referencing
- map-layers-formats
- appui-abstract

Fairly cautious here, using asserts where copilot wanted to return early.


core/markup/src/Markup.ts
```diff
@@ -170,7 +170,8 @@ export class MarkupApp {
   }
   /** @internal */
   public static getVpToScreenMtx(): Matrix {
-    const rect = this.markup!.markupDiv.getBoundingClientRect();
+    const markup = expectDefined(this.markup);
+    const rect = markup.markupDiv.getBoundingClientRect();
```
tcobbs-bentley: While it doesn't exactly hurt anything, I personally feel that using `expectDefined` inline makes it more clear in the git history that the only change was replacing `!` with `expectDefined`. It seems like you introduce new variables for this a lot (always?), and if you feel that doing so is preferable, just mark this comment as resolved without updating.

```suggestion
    const rect = expectDefined(this.markup).markupDiv.getBoundingClientRect();
```

tcobbs-bentley: Edit: looks like you switched to doing things inline later in the PR. I would recommend updating the places where you introduced new variables for `expectDefined` and `expectNotNull` to not do so.

core/markup/src/SelectTool.ts
```diff
@@ -667,12 +668,14 @@ export class SelectTool extends MarkupTool {
     const rightToLeft = (start.x > end.x);
     const overlapMode = (ev.isShiftKey ? !rightToLeft : rightToLeft); // Shift inverts inside/overlap selection...
     const offset = Point3d.create(vec.x < 0 ? end.x : start.x, vec.y < 0 ? end.y : start.y); // define location by corner points...
-    this.markup.svgDynamics!.clear();
-    this.markup.svgDynamics!.rect(width, height).move(offset.x, offset.y).css({ "stroke-width": 1, "stroke": "black", "stroke-opacity": 0.5, "fill": "lightBlue", "fill-opacity": 0.2 });
-    const selectBox = this.markup.svgDynamics!.rect(width, height).move(offset.x, offset.y).css({ "stroke-width": 1, "stroke": "white", "stroke-opacity": 1.0, "stroke-dasharray": overlapMode ? "5" : "2", "fill": "none" });
-    const outlinesG = isDynamics ? this.markup.svgDynamics!.group() : undefined;
+    const svgDynamics = expectDefined(this.markup.svgDynamics);
+    const svgMarkup = expectDefined(this.markup.svgMarkup);
```
tcobbs-bentley, line 672: Since `svgDynamics` is used multiple times, I do agree that using a variable for it is better than a bunch of inline `expectDefined` calls. And even though `svgMarkup` is only used once, using a variable for it also in this context does seem to make sense. (Just resolve this comment after reading it.)

extensions/map-layers-formats/src/GoogleMaps/GoogleMapsImageryProvider.ts
```diff
@@ -61,38 +62,7 @@ export class GoogleMapsImageryProvider extends MapLayerImageryProvider {
...
-
-    if (this._settings.properties?.overlay !== undefined) {
-      createSessionOptions.overlay = this._settings.properties.overlay as boolean;
-    }
-
-    if (this._settings.properties?.apiOptions !== undefined) {
-      createSessionOptions.apiOptions = this._settings.properties.apiOptions as string[];
-    }
-    return createSessionOptions;
+    return GoogleMapsUtils.getSessionOptionsFromMapLayer(settings);
```
tcobbs-bentley: It seems like this is changing the behavior. The old code used both `this._settings` and `settings`, while the new code just uses `settings`. Was the old code broken?

extensions/map-layers-formats/src/internal/GoogleMapsUtils.ts
```diff
@@ -89,27 +90,28 @@ export const GoogleMapsUtils = {
       Logger.logError(loggerCategory, msg);
       throw new BentleyError(BentleyStatus.ERROR, msg);
     }
+    const requiredProperties = expectDefined(properties);
```
tcobbs-bentley: Since it's impossible to get here with `properties` being undefined, I feel that this is an appropriate place to use `!` with an eslint disable. Note that in the future we ideally want to search our code base for expectDefined and get rid of most or all uses.

```suggestion
    // If properties is undefined, the above code will throw an exception, so ! is appropriate
    const requiredProperties = properties!;  // eslint-disable-line @typescript-eslint/no-non-null-assertion
```

extensions/map-layers-formats/src/OgcApiFeatures/OgcApiFeaturesProvider.ts
```diff
@@ -276,9 +276,10 @@ export class OgcApiFeaturesProvider extends MapLayerImageryProvider {
         tmpUrl = this.appendCustomParams(nextLink.href);
         response = await this.makeTileRequest(tmpUrl, this._staticModeFetchTimeout);
         json = await response.json();
-        if (json?.features)
-          data!.features = this._staticData?.features ? [...this._staticData.features, ...json.features] : json.features;
-        else
+        if (json?.features) {
+          const featureData = expectDefined(data);
+          featureData.features = featureData.features ? [...featureData.features, ...json.features] : json.features;
```
tcobbs-bentley: This doesn't seem to be equivalent to the old code. The old code used data that was in `this` (`this._staticData.features`), while the new code uses the features from the first call to `makeTileRequest`. Even if the new code is correct, I don't feel that it belongs in this PR.

extensions/map-layers-formats/src/Tools/MapFeatureInfoDecorator.ts
```diff
@@ -104,9 +104,8 @@ class DrapedPrimitives {
   }
 
   public addMeshes(drapedMeshes: Polyface[]) {
-    if (!this._meshes)
-      this._meshes = [];
-    drapedMeshes.forEach((ds) => this._meshes!.push(ds));
+    const meshes = this._meshes ?? (this._meshes = []);
+    drapedMeshes.forEach((ds) => meshes.push(ds));
```
tcobbs-bentley: The whole reason `!` was required was due to the use of `forEach` instead of `for of`, so I feel that switching to `for of` is a better (more easy to understand) fix.
```suggestion
    if (!this._meshes)
      this._meshes = [];
    for (const ds of drapedMeshes) {
      this._meshes.push(ds);
    }
```

core/markup/src/SelectTool.ts
ben-polinsky: ty

extensions/map-layers-formats/src/Tools/MapFeatureInfoTool.ts
```diff
@@ -60,8 +60,9 @@ class ActiveMapLayerState {
     const result = {exists: false, hidden: false};
     this.existsInDisplayStyle = false;
     this.isVisible = false;
-    if (this.hasMapLayers) {
-      const oldMls = this._activeMapLayers![0];    // consider only first layer for now
+    const activeMapLayers = this.activeMapLayers;
+    if (activeMapLayers && activeMapLayers.length > 0) {
+      const oldMls = activeMapLayers[0];    // consider only first layer for now
```
tcobbs-bentley: Based on changes that I saw in another one of these non-null-assertion PRs, I feel that the following more complicated fix is better:

First, define this type and function somewhere:
```ts
export type NonEmptyArray<T> = [T, ...T[]];

export function isNonEmptyArray<T>(arr?: T[]): arr is NonEmptyArray<T> {
  return !!arr && arr.length > 0;
}
```

Next, add this member function (cannot be `get`):
```ts
  public hasActiveMapLayers(): this is { activeMapLayers: NonEmptyArray<MapLayerInfoFromTileTree> } {
    return isNonEmptyArray(this.activeMapLayers);
  }
```

Finally, use the above function:
```ts
    if (this.hasActiveMapLayers()) {
      const oldMls = this.activeMapLayers[0];    // consider only first layer for now
```

By doing it this way, calling the new `hasActiveMapLayers` member function's type predicate asserts that `this.activeMapLayers` is a non-empty array, which lets the ts compiler know that `!` is not needed.

extensions/map-layers-formats/src/Tools/MapFeatureInfoTool.ts
```diff
@@ -87,16 +88,17 @@ class ActiveMapLayerState {
...
     }
     return result;
   }
 
   public updateWithScaleRangeVisibility(layerIndexes: MapLayerScaleRangeVisibility[]) {
-    if (this.hasMapLayers) {
-      const currentMls = this.activeMapLayers![0];    // consider only first layer for now
+    const activeMapLayers = this.activeMapLayers;
+    if (activeMapLayers && activeMapLayers.length > 0) {
+      const currentMls = activeMapLayers[0];    // consider only first layer for now
```
tcobbs-bentley: See above.

ui/appui-abstract/src/appui-abstract/quantity/BaseQuantityDescription.ts
```diff
@@ -22,20 +22,22 @@ export abstract class BaseQuantityDescription implements PropertyDescription {
...
     this.typename = StandardTypeNames.Number;
+
+    const editorParams: PropertyEditorParams[] = [];
+    editorParams.push({
+      type: PropertyEditorParamTypes.CustomFormattedNumber,
+      formatFunction: this.format,
+      parseFunction: this.parse,
+    } as CustomFormattedNumberParams);
+
     this.editor = {
```
tcobbs-bentley, line 37: I feel that this should be declared inline instead of using push.
```ts
const editorParams: PropertyEditorParams[] = [{
  ...
}];
```

tcobbs-bentley (approved):

extensions/map-layers-formats/src/GoogleMaps/GoogleMapsImageryProvider.ts
ben-polinsky: thats probably the safer move. I saw heavily duplicated code and tried to refactor, but yeah not in scope in this PR.

Reverted.

extensions/map-layers-formats/src/Tools/MapFeatureInfoTool.ts
ben-polinsky: Both seem fine. How much do you want this change?

tcobbs-bentley: It's certainly not necessary, but providing a useful fix for the semi-useless `hasMapLayers` seems like a win (that is related to non-null assertions). It's your call.

ui/appui-abstract/src/appui-abstract/quantity/BaseQuantityDescription.ts
ben-polinsky: Fixed, not sure why the "outdated" label is not shown.

extensions/map-layers-formats/src/Tools/MapFeatureInfoTool.ts
ben-polinsky: I changed to match your advice

extensions/map-layers-formats/src/OgcApiFeatures/OgcApiFeaturesProvider.ts
ben-polinsky: fixed

extensions/map-layers-formats/src/internal/GoogleMapsUtils.ts
ben-polinsky: Reworked without re-introducing `!`

ui/appui-abstract/src/appui-abstract/dialogs/UiLayoutDataProvider.ts
```diff
@@ -105,10 +105,13 @@ export abstract class UiLayoutDataProvider extends UiDataProvider {
...
-      this.loadItemsInternal(this.supplyDialogItems());
-    }
-    return this._items!;
+    const items = this._items;
+    if (undefined !== items)
+      return items;
+
+    const suppliedItems = this.supplyDialogItems() ?? [];
+    this.loadItemsInternal(suppliedItems);
+    return suppliedItems;
```
hl662: Since `loadItemsInternal` is a protected hook on a `@public` base class, might be worth returning `this._items` after initialization instead of `suppliedItems` here? That keeps the old behavior where subclasses can normalize/replace the array in `loadItemsInternal` and callers still observe the cached post-hook value.

extensions/map-layers-formats/src/GoogleMaps/GoogleMapsImageryProvider.ts
```diff
@@ -61,37 +62,51 @@ export class GoogleMapsImageryProvider extends MapLayerImageryProvider {
   }
 
   protected createCreateSessionOptions(settings: ImageMapLayerSettings): GoogleMapsCreateSessionOptions {
```
hl662, line 64: Since `initialize()` already goes through `GoogleMapsUtils.getSessionOptionsFromMapLayer`, might be worth either deleting `createCreateSessionOptions()` or making it delegate to the shared helper? Right now it validates `settings.properties` but reads most values from `this._settings.properties`, which feels a little easy to drift if this helper gets used again later.

ui/appui-abstract/src/appui-abstract/dialogs/UiLayoutDataProvider.ts
ben-polinsky: Yeah, fair point. But then we have to keep the null assertion or expectDefined, which changes behavior. The issue is of course the method signature.

Went with  explicit cast to satisfy the signature and preserve behavior.

extensions/map-layers-formats/src/GoogleMaps/GoogleMapsImageryProvider.ts
ben-polinsky: Fair point, but out of the scope of this PR. We'd be in these non-null prs forever if we tried to fix things like this everywhere.

ui/appui-abstract/src/appui-abstract/dialogs/UiLayoutDataProvider.ts
```diff
@@ -105,10 +105,13 @@ export abstract class UiLayoutDataProvider extends UiDataProvider {
...
-      this.loadItemsInternal(this.supplyDialogItems());
-    }
-    return this._items!;
+    const items = this._items;
+    if (undefined !== items)
+      return items;
+
+    const suppliedItems = this.supplyDialogItems() ?? [];
+    this.loadItemsInternal(suppliedItems);
+    return this._items as ReadonlyArray<DialogItem>;
```
tcobbs-bentley: In my opinion, this is just cheating to avoid using `!`, since it has the same end result. I think you should put it back to using `!` and add an eslint-disable comment and another comment explaining why `!` is ok in this case.

ui/appui-abstract/src/appui-abstract/dialogs/UiLayoutDataProvider.ts
```diff
@@ -105,10 +105,15 @@ export abstract class UiLayoutDataProvider extends UiDataProvider {
...
   public get items(): ReadonlyArray<DialogItem> {
-    if (undefined === this._items) {
-      this.loadItemsInternal(this.supplyDialogItems());
-    }
-    return this._items!;
+    const items = this._items;
+    if (undefined !== items)
+      return items;
+
+    const suppliedItems = this.supplyDialogItems() ?? [];
```
tcobbs-bentley: After the other changes, `?? []` is no longer needed here.

ui/appui-abstract/src/appui-abstract/dialogs/UiLayoutDataProvider.ts
ben-polinsky: Gotcha agree.

ui/appui-abstract/src/appui-abstract/dialogs/UiLayoutDataProvider.ts
ben-polinsky: done

hl662 (approved):
lgtm, quantity stuff in appui abstract looks just fine

extensions/map-layers-formats/src/Tools/MapFeatureInfoTool.ts
```diff
@@ -40,6 +40,12 @@ export interface MapFeatureInfoToolData {
   mapInfo?: MapFeatureInfo;
 }
 
+type NonEmptyArray<T> = [T, ...T[]];
```
eringram, line 43: Could these be worth adding to the core-bentley package like in UtilityTypes.ts, etc.?

eringram (approved):
Map-layers-formats changes look good

extensions/map-layers-formats/src/Tools/MapFeatureInfoTool.ts
tcobbs-bentley: I do think having `NonEmptyArray` and `isNonEmptyArray` in UtilityTypes.ts would be useful. It does mean new API, and we'd have to decide if they are beta or not. The current type and function are not exported, so they don't count as new API.

bbastings (approved):

aruniverse (approved):

PR #9135, opened 2026-03-25
Clear non-null assertions in examples and tools

ben-polinsky:
Part 4/4 in Continuing with https://github.com/iTwin/itwinjs-core/issues/8376 - cleaning up code caught by non-null assertions lint rule.

- example-code
- test-apps
- tools

After we merge all 4, we'll flip the rule in the next PR


test-apps/display-performance-test-app/src/backend/backend.ts
```diff
@@ -60,47 +60,58 @@ export async function initializeBackend() {
...
  */
-function checkEnvVars(...keys: Array<string>): boolean {
+function getEnvVars(...keys: Array<string>): string[] | undefined {
   const missing = keys.filter((name) => process.env[name] === undefined);
-  if (missing.length === 0)
-    return true;
+  if (missing.length === 0) {
+    return keys.map((name) => {
+      const value = process.env[name];
+      if (value === undefined)
```
hl662: isn't this redundant? since line 102 we are filtering out keys whose value in process.env is undefined. This would be dead code. Can you cast missing or cast const value at line 105 to always be defined? This is a good candidate for disabling non null assertion otherwise

ben-polinsky: TBH, I went quick with my review of the test and example code. My goal is to get rid of these warnings in a reasonable way while ensuring we don't short circuit in places where the null would not otherwise stop execution.

Perfection is not the goal here. We can make reasonable changes, though.

ben-polinsky: (I will take a look :))

ben-polinsky:
Addressed review feedback: I refactored getEnvVars in test-apps/display-performance-test-app/src/backend/backend.ts to remove the redundant undefined check in the all-present path. It now gathers missing and values in one pass and returns values directly when nothing is missing. Existing behavior for partially-missing vars (warning and skipping auth setup) is unchanged.

ben-polinsky:
> Have you tried running DTA after these changes?

Yes. Did not test every feature, but started fine and seemed to run fine.

Issue #8968, opened 2026-02-09
Using multiple `IdSet` in a query does not work


JonasDov:
**Describe the bug**
When multiple `IdSet` are used, query always returns: `Expression '<Alias>.id' in ECSQL statement is ambiguous.`.

**To Reproduce**

1. Open any iModel.
2. Run the following ecsql query:
```sql
SELECT
  Model.Id modelId,
  Category.Id categoryId
FROM
  BisCore.GeometricElement3d
LIMIT
  1
```
3. In the following query, replace `<modelId>` with `modelId` and `<categoryId>` with `categoryId` from the first query:
```sql
SELECT
  this.ECInstanceId AS ECInstanceId
FROM
  BisCore.GeometricElement3d this,
  IdSet ('["<categoryId>"]') AS categoryIdSetTable,
  IdSet ('["<modelId>"]') AS modelIdSetTable
WHERE
  this.Category.Id = categoryIdSetTable.id
  AND this.Model.Id = modelIdSetTable.id
  ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES
```

**Expected behavior**
Query should allow having multiple `IdSet` tables defined.

PR #9139, opened 2026-03-26
Minor bump on `IModelReadRpcInterface.interfaceVersion`

grigasp:
iTwin.js Core @ 5.7 fixed [a critical `IdSet` issue](https://github.com/iTwin/itwinjs-core/issues/8968), which allows us to use that ECSQL feature in our queries. However, it didn't bump RPC version, so there's no guarantee that RPC requests sent using `core-frontend @ 5.7+` will land in a `core-backend @ 5.7` - in theory it could be `5.5` or even `5.0`. To avoid that, we need to bump RPC minor version, which will prevent newer frontend sending requests to older backends, and will let us safely use the `IdSet` ECSQL language feature.

cc @aruniverse @khanaffan

### DIFF
--- a/common/changes/@itwin/core-common/master_2026-03-26-14-26.json
+++ b/common/changes/@itwin/core-common/master_2026-03-26-14-26.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-common",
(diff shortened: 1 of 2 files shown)


aruniverse:
image

We'll make sure to rollout the backend first before the official pkgs get published.

PR #9140, opened 2026-03-27
Add NativeChangesetReader API with filtering

soham-bentley:
imodel-native: https://github.com/iTwin/imodel-native/pull/1368

In Phase 1 (this PR):
- We give lazy property loading - using three property filters(InstanceKey, BisCoreElement, All)
- We can now open multiple changeset files in the form of a changeset group
- We can decode arrays properly.
- We use the standard EC-to-JSON adaptor
- We give data on a row by row basis using our JSON adaptor
- We support rich metadata like operation type and isIndirectChange flag
- We support filtering based on operation(Inserted or Updated or Deleted), table names and classNames.


soham-bentley:
@khanaffan @rschili I have added as much extensive testing as possible. But still if you fee something more can or should be tested
Please feel free to mention it in the PR

soham-bentley:
I have taken absolutely separate code path in my new `ChangesetReader` and `unifier` API
Because the existing `ChangesetECAdaptor` and unifier API will be `deprecated` and removed in some time, so forcefully unifying the logic didn't make sense. It would have made the code unnecessarily complex. Instead this new `unifier` and `ChangesetReader` API and its exported types can become the standard going forward (if everybody agrees).

core/backend/src/ECChangesetReader.ts
```diff
@@ -0,0 +1,675 @@
...
+
+/** Shape returned by the native getValue() call. Not exported — internal bridge only. */
+interface ECChangesetRowValue {
+  /** `false` for internal SQLite tables; `data` and `key` are `undefined` in that case. */
+  isECTable: boolean;
+  /** EC property bag. `undefined` when `isECTable` is `false`. */
+  data?: any;
+  /** Native instance key (encodes ECInstanceId + root class). `undefined` when `isECTable` is `false`. */
+  key?: string;
+}
```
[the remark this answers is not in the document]
soham-bentley: outdated

core/backend/src/ECChangesetReader.ts
```diff
@@ -0,0 +1,675 @@
...
+    return false;
+  }
+
+  /**
+   * SQLite opcode of the current change.
+   * Valid only after a successful call to {@link step}.
+   * @beta
+   */
+  public get op(): ECNativeChangeOp { return this._op!; }
+
```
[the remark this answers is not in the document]
soham-bentley: done

core/backend/src/ECChangesetReader.ts
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  /** Convert a IModelJsNative.ECChangesetReader.Mode enum value to its string name.
+   * @internal
+   */
+  private modeToString(mode: IModelJsNative.ECChangesetReader.Mode): string {
+    switch (mode) {
+      case IModelJsNative.ECChangesetReader.Mode.All_Properties: return "All_Properties";
+      case IModelJsNative.ECChangesetReader.Mode.Bis_Element_Properties: return "Bis_Element_Properties";
+      case IModelJsNative.ECChangesetReader.Mode.Instance_Key: return "Instance_Key";
```
hl662: very neat, is the native side extensible enough to add more modes easily in the future? We've been thinking about excluding geometry (blobs? still learning my way through our lower layer)
Or is that not a realistically valuable case

hl662 (commented):
Went through all 19 files. The main finding is a `SqliteStatement` leak in `NativeSqliteBackedInstanceCache.all()` — the generator doesn't use `try/finally`, so early `break` skips disposal. See inline comment for the fix.

The rest is mostly hardening — factory methods should clean up the native reader on throw, the test suite is missing error paths and `delete-full`, and the doc examples use `.has()` on a `string[]` (should be `.includes()`). Also flagged that the doc examples are inlined markdown rather than extracted snippets, which is how the `.has()` bug slipped through without CI catching it.

Nothing blocking — strong API design and genuinely useful documentation (especially the out-of-sync section).

core/backend/src/ECNativePartialChangeUnifier.ts
```diff
@@ -0,0 +1,290 @@
...
+    const sql = `
+      SELECT JSON_GROUP_ARRAY(JSON([value]))
+      FROM (
+        SELECT [value],
+               SUM(LENGTH([value])) OVER (ORDER BY [key] ROWS UNBOUNDED PRECEDING) / ${this.bufferedReadInstanceSizeInBytes} AS [bucket]
+        FROM ${this._cacheTable}
+      )
+      GROUP BY [bucket]`;
+
+    const stmt = this._db.prepareSqliteStatement(sql);
```
hl662, line 164: `SqliteStatement` leak when consumer breaks out of `all()` early. The `stmt[Symbol.dispose]()` on line 179 is only reached if the generator runs to completion.

If a consumer does:
```ts
for (const inst of cache.all()) {
  if (found) break;  // generator.return() called — stmt never disposed
}
```

Wrap in `try/finally` so `generator.return()` triggers cleanup:
```ts
public *all(): IterableIterator<ECNativeChangeInstance> {
    const stmt = this._db.prepareSqliteStatement(sql);
    try {
      while (stmt.step() === DbResult.BE_SQLITE_ROW) {
        const bucket = JSON.parse(stmt.getValueString(0), Base64EncodedString.reviver) as ECNativeChangeInstance[];
        for (const instance of bucket)
          yield instance;
      }
    } finally {
      stmt[Symbol.dispose]();
    }
}
```

The in-memory cache doesn't have this issue (no external resources).

core/backend/src/ChangesetECAdaptor.ts
```diff
@@ -352,6 +354,7 @@ class ECDbMap {
 /**
  * Record meta data for the change.
  * @beta
+ * @deprecated The ChangeMetaData interface is deprecated and will be removed in a future release.
```
hl662: The deprecation messages are tautological — "is deprecated and will be removed" doesn't add info beyond the `@deprecated` tag itself. Consider pointing consumers to the replacement:

```typescript
/** @deprecated Use ECNativeChangeMeta from ECChangesetReader instead. */
export interface ChangeMetaData { ... }

/** @deprecated Use ECNativeChangeInstance from ECChangesetReader instead. */
export interface ChangedECInstance { ... }

/** @deprecated Use ECNativeChangeUnifierCache.createInMemoryCache / createSqliteBackedCache instead. */
export interface ECChangeUnifierCache { ... }

/** @deprecated Use ECNativePartialChangeUnifier instead. */
export class PartialECChangeUnifier { ... }

/** @deprecated Use ECChangesetReader instead. */
export class ChangesetECAdaptor { ... }
```

Also: the member-level `@deprecated` tags on `appendFrom` and `instances` are redundant since the class itself is already deprecated.

docs/learning/backend/ECChangesetReader.md
```diff
@@ -0,0 +1,605 @@
...
+using pcu = new ECNativePartialChangeUnifier(ECNativeChangeUnifierCache.createInMemoryCache());
+while (reader.step()) pcu.appendFrom(reader);
+
+for (const instance of pcu.instances) {
+  // Instance property keys are camelCase due to useJsName:
+  console.log(instance.lastMod);        // JS name — correct
+  console.log(instance.category?.id);  // JS name — correct
+
+  // changeFetchedPropNames always uses original EC names, NOT JS names:
+  instance.$meta.changeFetchedPropNames.has("LastMod");       // ✅ correct
```
hl662: Bug: `.has()` is a `Set` method but `changeFetchedPropNames` is typed as `string[]` — these examples won't compile. Should be `.includes()` (as used correctly in the earlier example on line 62).

More broadly, the repo convention is to put doc examples in `example-code/snippets/src/backend/*.test.ts` with `__PUBLISH_EXTRACT_START__`/`END` markers and reference them in docs via `[[include:ExampleName]]`. This gives CI compilation so bugs like this are caught automatically.

I'd recommend moving the key examples (especially the worked example) into an extracted snippet file — the existing files in `example-code/snippets/src/backend/` are good references for the pattern.

core/backend/src/ECChangesetReader.ts
```diff
@@ -0,0 +1,386 @@
...
+   * @param args.invert When `true`, invert all operations (Insert↔Delete).
+   * @param args.valueOptions Row adaptor options controlling how EC property values are formatted.
+   * @param args.mode Controls which properties are included. Defaults to `All_Properties`.
+   * @beta
+   */
+  public static openFile(args: { readonly fileName: string } & ECChangesetReaderArgs): ECChangesetReader {
+    const reader = new ECChangesetReader(args.db);
+    reader._rowOptions = args.rowOptions;
+    reader._mode = args.mode ?? IModelJsNative.ECChangesetReader.Mode.All_Properties;
+    reader._nativeReader.openFile(args.db[_nativeDb], args.fileName, args.invert ?? false, reader._mode);
```
hl662: If the native `openFile()` call throws (bad path, corrupt file, schema mismatch), the `ECChangesetReader` constructed on line 129 is never returned to the caller — so they can't call `close()`. The native C++ `ECChangesetReader` allocated in the field initializer (line 38) leaks until GC collects.

Consider wrapping the native call in try/catch across all five factory methods:

```typescript
public static openFile(args: ...): ECChangesetReader {
    const reader = new ECChangesetReader(args.db);
    reader._rowOptions = args.rowOptions;
    reader._mode = args.mode ?? IModelJsNative.ECChangesetReader.Mode.All_Properties;
    try {
      reader._nativeReader.openFile(args.db[_nativeDb], args.fileName, args.invert ?? false, reader._mode);
    } catch (e) {
      reader.close();
      throw e;
    }
    return reader;
}
```

Same pattern applies to `openGroup`, `openLocalChanges`, `openInMemoryChanges`, and `openTxn`.

core/backend/src/test/standalone/ECChangesetReader.test.ts
```diff
@@ -0,0 +1,3929 @@
...
+    assert.deepEqual([...elemOld!.$meta.changeFetchedPropNames].sort(), ["BinProp", "BoolProp",
+      "DblProp", "DtProp", "ECInstanceId", "IntArrProp", "IntProp", "LastMod",
+      "LongProp", "Pt2dProp", "Pt3dProp.X", "Pt3dProp.Y", "RelatedElem.Id", "StrArrProp", "StrProp",
+      "StructArrProp", "StructProp.Label", "StructProp.Pt2d", "StructProp.Pt3d", "StructProp.X",
+      "StructProp.Y", "StructProp.Z"].sort());
+  });
+
+});
+
+describe("ECChangesetReader delete-partial", () => {
```
hl662, line 2341: The test matrix covers insert-full, insert-partial, update-full, delete-partial — but `delete-full` (deleting an element spanning multiple tables / overflow) is missing. This would exercise the unifier merge path for deletions across overflow tables.

core/backend/src/test/standalone/ECChangesetReader.test.ts
```diff
@@ -0,0 +1,3929 @@
...
+  return txn;
+}
+
+async function importSchemaStrings(txn: EditTxn, schemas: string[]): Promise<void> {
+  if (txn.isActive)
+    txn.saveChanges();
+  await txn.iModel.importSchemaStrings(schemas);
+}
+
+describe("ECChangesetReader insert-full", () => {
```
hl662, line 53: Great test breadth (60 cases!), but no tests for error/failure paths:

- `openFile` with invalid/nonexistent path
- `openGroup` with empty array (the guard on line 146 of ECChangesetReader.ts)
- Accessing `op`, `tableName`, `isECTable`, `isIndirectChange` before `step()` — these throw `IModelError` but the guards are untested
- Double `close()` idempotency

These are the first thing consumers of a `@beta` API will hit.

core/backend/src/test/standalone/ECChangesetReader.test.ts
```diff
@@ -0,0 +1,3929 @@
...
+  let rwIModel: BriefcaseDb;
```
hl662, line 54: The SQLite-backed cache is only exercised in `openLocalChanges` tests and perf tests. The perf tests assert instance *counts* match, but not instance *content*.

Consider adding a test that runs the same changeset through both InMemoryCache and SqliteBackedCache and asserts deep equality of the resulting instances. This would catch JSON round-trip issues (e.g., `undefined` vs missing key, numeric precision, Base64 encoding).

docs/learning/backend/ECChangesetReader.md
```diff
@@ -0,0 +1,605 @@
...
+
+// So only trust s.X for the before/after comparison:
+const oldX = elementOld.s.X; // 1.5 — correct
+const newX = elementNew.s.X; // 100 — correct
+// Do NOT compare s.Y from Old vs New — it was not part of this changeset.
+```
+
+### Summary
+
+This scenario is not tied to the type of change being opened by the `ECChangesetReader` i.e. it doesnot depend on whether a change group or a changeset or a transaction is opened. It might happen when the iModel's state is not in sync with the change being read. In other words it might happen when the iModel's state is not **at** the change being read.
```
hl662: Typo:
```suggestion
it does not depend on whether...
```

docs/learning/backend/ECChangesetReader.md
```diff
@@ -0,0 +1,605 @@
+# ECChangesetReader
+
+`ECChangesetReader` is a low-level API in `@itwin/core-backend` that reads EC-typed change data from a changeset file, a group of changeset files, an in-memory transaction, or local un-pushed changes. It is designed for use cases where you need to inspect what changed at the EC property level — for example, building audit trails, incremental projections, or custom synchronization logic.
+
+## Core concepts
+
+### How change data is stored
+
+iTwin.js stores changes at the SQLite table-row level. A single EC entity may typically map to multiple tables or single table. Each table row change is a separate entry in the changeset stream.
```
hl662: Minor grammar:
```suggestion
A single EC entity may typically map to multiple tables or a single table.
```

core/backend/src/ECChangesetReader.ts
```diff
@@ -0,0 +1,386 @@
...
+  /**
+   * Read pending (not yet pushed) local changes from an open IModelDb.
+   * @param args.db Must be an `IModelDb` (not `ECDb`).
+   * @param args.includeInMemoryChanges Also include in-memory (not yet saved to disk) changes.
+   * @param args.valueOptions Row adaptor options controlling how EC property values are formatted.
+   * @param args.mode Controls which properties are included. Defaults to `All_Properties`.
+   * @beta
+   */
+  public static openLocalChanges(
+    args: Omit<ECChangesetReaderArgs, "db"> & { db: IModelDb; includeInMemoryChanges?: true },
```
hl662: nit: `includeInMemoryChanges?: true` prevents consumers from passing a dynamic `boolean`:

```ts
const flag: boolean = getConfigFlag();
ECChangesetReader.openLocalChanges({ db, includeInMemoryChanges: flag });
// TS error: Type 'boolean' is not assignable to type 'true | undefined'
```

Consider `?: boolean` for consistency with `invert?: boolean` on the same interface. The `false` value is redundant (same as omitting) but harmless, and it makes the API composable with dynamic config values.

core/backend/src/ECNativePartialChangeUnifier.ts
```diff
@@ -0,0 +1,290 @@
...
+// ---------------------------------------------------------------------------
+// ECNativeChangeUnifierCache — interface + factory
+// ---------------------------------------------------------------------------
+
+/**
+ * Cache used by {@link ECNativePartialChangeUnifier} to accumulate and merge
+ * partial EC change instances.
+ * @beta
+ */
+export interface ECChangeCache extends Disposable {
```
hl662: nit: `ECChangeCache` breaks the `ECNative*` prefix convention used by its factory (`ECNativeChangeUnifierCache`) and consumer (`ECNativePartialChangeUnifier`). The old stack was consistent — `ECChangeUnifierCache` was both interface and namespace.

Suggest renaming to `ECNativeChangeCache` to align with the sibling symbols:

```typescript
export interface ECNativeChangeCache extends Disposable { ... }

export namespace ECNativeChangeUnifierCache {
  export function createInMemoryCache(): ECNativeChangeCache;
  export function createSqliteBackedCache(...): ECNativeChangeCache;
}
```

Everything is `@beta` so this is non-breaking.

full-stack-tests/backend/src/perftest/ECChangesetReader.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      watch.stop();
+
+      assert.equal(pcu.instanceCount, testCase.numElements, "Number of instances should match the number of inserted elements");
+      reporter.addEntry("ECChangesetReaderAPI", "ECNativeUnifier-InMemoryCache", "Execution time (seconds)", watch.elapsedSeconds, { iModelId: rwIModelId, changesetId: changesets[testCase.testCaseNum].id, changesetInserts: testCase.numElements });
+    }
+
+    rwIModel.close();
+  });
+
+  it("Large Changeset Performance - SqliteBackedCache", async () => {
```
hl662: nit: The two perf tests (InMemoryCache and SqliteBackedCache) are ~95% identical — the only difference is which cache factory is called on line 93 vs 169. Extract the shared setup + timing loop into a helper and parameterize the cache strategy to halve the code.

core/backend/src/ECNativePartialChangeUnifier.ts
```diff
@@ -0,0 +1,290 @@
...
+  private createTempTable(): void {
+    this._db.withSqliteStatement(`CREATE TABLE ${this._cacheTable} ([key] text primary key, [value] text)`, (stmt: SqliteStatement) => {
+      if (DbResult.BE_SQLITE_DONE !== stmt.step())
+        throw new Error("unable to create temp cache table");
+    });
+  }
+
+  private dropTempTable(): void {
+    if (this._db instanceof ECDb) {
+      this._db.saveChanges();
```
rschili: Do we need to save here? I think templ tables to not participate in main schema transactions, so this may not need a commit?
But this call to commit WILL commit any pending transactions that the user has on that connection.
Also, how does this overlap with Keith's EditTxn PR? It was merged recently, and I believe direct saveChanges calls are now discouraged.

core/backend/src/ECChangesetReader.ts
```diff
@@ -0,0 +1,386 @@
...
+
+  /**
+   * Concatenate multiple changeset files and read them as a single logical stream.
+   * @param args.changesetFiles Ordered list of changeset file paths.
+   * @param args.db Database with schema at or ahead of the last changeset.
+   * @param args.valueOptions Row adaptor options controlling how EC property values are formatted.
+   * @param args.mode Controls which properties are included. Defaults to `All_Properties`.
+   * @beta
+   */
+  public static openGroup(args: { readonly changesetFiles: string[] } & ECChangesetReaderArgs): ECChangesetReader {
```
rschili, line 161: Should this load everything into memory immediately? May be better to chunk.

core/backend/src/ECChangesetReader.ts
```diff
@@ -0,0 +1,386 @@
...
+
+  /**
+   * Restrict iteration to changes for the given EC class ids.
+   * That means both `inserted` and `deleted` fields will be undefined for changes related to
+   * classIds other than the ones passed in the filter.
+   * The rest of the metadata related data fields like `op`, `isIndirectChange`, `tableName`, etc. will still be populated for filtered-out rows.
+   * @param classIds ECClassId values (as hex Id64 strings) to include.
+   * @beta
+   */
+  public setClassIdFilters(classIds: Set<Id64String>): void {
```
rschili: The doc says "metadata fields like `op`, `isIndirectChange`, `tableName`, etc. will still be populated for filtered-out rows" but there's a hidden asymmetry: for table-name and opcode filters the caller can still inspect why a row is empty (`tableName`/`op` tell you what was blocked). For classId filters there is no equivalent — `getChangeMetadata()` does not expose the resolved classId, so the caller cannot distinguish "my classId filter blocked this row" from "this row genuinely had no EC data". In practice the only observable left is `!inserted && !deleted && isECTable`, which mixes both cases.

Two options worth considering before this ships:
1. Add `classId?: string` to the object returned by `getChangeMetadata()` (the C++ already resolves it before the filter check in `ReFetchValues`, so the cost is zero).
2. Or, skip filtered rows entirely inside `Step()` and remove this whole "hollow row" concept — filters would then behave like most users expect.

The current behaviour is fine for `@beta` if it stays documented explicitly, but the docblock should at minimum say "**note: the resolved classId is not exposed in metadata**, so callers cannot programmatically identify which classId caused a row to be suppressed."

soham-bentley: filtered rows entirely skipped

core/backend/src/ECChangesetReader.ts
soham-bentley: I guess we have done similar discussion in this comment https://github.com/iTwin/imodel-native/pull/1368#discussion_r3067784698

core/backend/src/ECNativePartialChangeUnifier.ts
soham-bentley: We now use temporary SQlite Db

core/backend/src/ECNativePartialChangeUnifier.ts
soham-bentley: Done

core/backend/src/ECChangesetReader.ts
soham-bentley: Done

docs/learning/backend/ECChangesetReader.md
soham-bentley: Done

docs/learning/backend/ECChangesetReader.md
soham-bentley: Done

docs/learning/backend/ECChangesetReader.md
soham-bentley: Done

core/backend/src/ECNativePartialChangeUnifier.ts
soham-bentley: Done

core/backend/src/ECChangesetReader.ts
soham-bentley: Good catch...Done

core/backend/src/ECChangesetReader.ts
soham-bentley: Yes we can very well do that but for 1st phase I think having these 3 options is enough as discussed with @khanaffan

core/backend/src/ECNativePartialChangeUnifier.ts
```diff
@@ -0,0 +1,277 @@
...
+
+class NativeSqliteBackedInstanceCache implements ECNativeChangeCache {
+  private readonly _cacheTable = `[${Guid.createValue()}]`;
+  public static readonly defaultBufferSize = 1024 * 1024 * 10; // 10 MB
+  private _db: SQLiteDb;
+  public constructor(
+    public readonly bufferedReadInstanceSizeInBytes: number = NativeSqliteBackedInstanceCache.defaultBufferSize,
+  ) {
+    this._db = new SQLiteDb();
+    this._db.createDb("", undefined, { skipFileCheck: true, rawSQLite: true });
```
soham-bentley: Putting this comment here to bring your attention @khanaffan . Only this new API uses the temporary SQLITE Db, kept the old deprecated ChangesetECAdaptor API unchanged.

core/backend/src/ChangesetECAdaptor.ts
soham-bentley: Done

full-stack-tests/backend/src/perftest/ECChangesetReader.test.ts
soham-bentley: Done

core/backend/src/test/standalone/ECChangesetReader.test.ts
soham-bentley: Added

core/backend/src/test/standalone/ECChangesetReader.test.ts
soham-bentley: Added

core/backend/src/test/standalone/ECChangesetReader.test.ts
soham-bentley: Added

hl662:
The TSDoc comments and learning doc should use `[Name]($pkg)` links so the docs site can generate clickable cross-references. Right now everything is plain backtick references.

**In `ECChangesetReader.ts`** — the `@param` docs for `openLocalChanges`, `openInMemoryChanges`, and `openTxn` reference `IModelDb` and `ECDb` as plain backticks. Can you switch these to `[IModelDb]($backend)` and `[ECDb]($backend)`? For example:

```diff
- * @param args.db Must be an `IModelDb` (not `ECDb`).
+ * @param args.db Must be an [IModelDb]($backend) (not [ECDb]($backend)).
```

**In `docs/learning/backend/ECChangesetReader.md`** — the doc currently has zero `[Type]($pkg)` links. Compare with `docs/learning/backend/IModelDb.md` which links types like `[BriefcaseDb]($backend)`, `[SnapshotDb.openFile]($backend)`, etc. Might be worth doing a pass to convert the main backtick references — at minimum `ECChangesetReader`, `ECNativePartialChangeUnifier`, `ECNativeChangeUnifierCache`, and `ECNativeChangeInstance` — to proper `[Name]($backend)` links?

Tip: Copilot can probably knock this whole thing out for you — just point it at the two files and ask it to convert backtick type references to `[Name]($backend)` doc links following the pattern in `IModelDb.ts` / `IModelDb.md`.

hl662:
Similarly, the deprecation comments in your doccomments can utilize the same so you should edit those too to be linked

rschili (approved):
The Two PR titles are both different tense from the story title. I suggest we make this a bit more verbose.

iTwin:
Add NativeECChangesetReader API with lazy loading and filtering
native:
Add native ECChangesetReader

khanaffan:
@copilot review this PR with intention to find new types being declare for which there is already types available and any potential duplicate code.

core/backend/src/ECChangesetReaderTypes.ts
```diff
@@ -0,0 +1,154 @@
...
+
+// ---------------------------------------------------------------------------
+// Type aliases
+// ---------------------------------------------------------------------------
+
+/**
+ * Operation that caused an EC change.
+ * @beta
+ */
+export type ECNativeChangeOp = "Inserted" | "Updated" | "Deleted";
```
khanaffan, line 18: Do not use EC or Native this will hit you to reuse existing type already declared. I ask copilot review this per again with duplicate type definitions.

core/backend/src/ECChangesetReaderTypes.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Which snapshot of the changed EC row.
+ * @beta
+ */
+export type ECNativeChangeStage = "Old" | "New";
```
khanaffan, line 24: I think we already had this but if not still do not use word EC or Native.

core/backend/src/ECChangesetReaderTypes.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Controls which properties are included in the output of [ECChangesetReader]($backend).
+ * @beta
+ */
+export enum ECChangesetMode {
+  /** All EC properties mapped to changed tables. */
+  All_Properties = 0,
```
khanaffan, line 32: We do not use _ in enum names. It interesting though linter did not complaint on it?.

core/backend/src/ECChangesetReaderTypes.ts
```diff
@@ -0,0 +1,154 @@
...
+ * Which snapshot of the changed EC row.
+ * @beta
+ */
+export type ECNativeChangeStage = "Old" | "New";
+
+/**
+ * Controls which properties are included in the output of [ECChangesetReader]($backend).
+ * @beta
+ */
+export enum ECChangesetMode {
```
khanaffan, line 30: Should be name something like PropertyFilter

core/backend/src/ECChangesetReaderTypes.ts
```diff
@@ -0,0 +1,154 @@
...
+  /** All EC properties mapped to changed tables. */
+  All_Properties = 0,
+  /** For classes whose base class is `BisCore:Element`, only `BisCore:Element` properties
+   * mapped to changed tables are returned. If no `BisCore:Element` class property changed,
+   * only `ECInstanceId` and `ECClassId` are returned. For other classes all mapped properties
+   * are returned. */
+  Bis_Element_Properties = 1,
```
khanaffan, line 37: May be 

```
{
All,
BisCore,
InstanceKey,
}
```

core/backend/src/ECChangesetReaderTypes.ts
```diff
@@ -0,0 +1,154 @@
...
+  /** Only `ECInstanceId` and `ECClassId`. */
+  Instance_Key = 2,
+}
+
+/**
+ * Row-formatting options for [ECChangesetReader]($backend) factory methods.
+ * Controls how EC property values are represented in the returned instances.
+ * @beta
+ */
+export interface ECChangesetRowAdapterOptions {
```
khanaffan, line 47: People using itwin.js do not really would underhand what is ChangesetRow adaptor is?

May be RowFormatOptions

core/backend/src/ECChangesetReaderTypes.ts
```diff
@@ -0,0 +1,154 @@
...
+  stage: ECNativeChangeStage;
+  /** Change-stream index positions. */
+  changeIndexes: number[];
+  /**
+   * Native instance key computed by the native layer.
+   * Encodes ECInstanceId and class Id.
+   */
+  nativeKey: string;
+  /** Reader mode that was active when this change row was captured. */
+  mode: string;
```
khanaffan, line 89: Mode seem to be enum or a type.

core/backend/src/ECChangesetReaderTypes.ts
```diff
@@ -0,0 +1,154 @@
...
+  op: ECNativeChangeOp;
+  /** Whether this is the pre-change (`"Old"`) or post-change (`"New"`) snapshot. */
+  stage: ECNativeChangeStage;
+  /** Change-stream index positions. */
+  changeIndexes: number[];
+  /**
+   * Native instance key computed by the native layer.
+   * Encodes ECInstanceId and class Id.
+   */
+  nativeKey: string;
```
khanaffan, line 87: Encode as what? and do not use native

core/backend/src/ECChangesetReaderTypes.ts
```diff
@@ -0,0 +1,154 @@
...
+   For compound data properties like point2d, point3d or navigation properties,
+  the full name of the property is returned in case all the components of the property are fetched from the change.
+  If all of the components are not fetched from the changes(meaning they did not change),
+  then the individual component names which changed are returned smartly by using `.` as a separator (e.g. "MyPoint.X", "MyPoint.Y" for a point3d property "MyPoint" if only X and Y changed).
+  For struct properties the property names are always returned in the "StructProp.MemberName" format.
+  So if only X changed for a point2d property named "Myp2d" inide a struct "CustomStruct", the returned property name will be "CustomStruct.Myp2d.X".
+  Similaly if both X and Y changed for the same point2d property, the returned property name will be "CustomStruct.Myp2d". */
+  changeFetchedPropNames: string[];
+  /** Row adaptor options that were active when this change row was captured. */
+  rowOptions?: ECChangesetRowAdapterOptions;
```
khanaffan, line 100: As you see you use `rowOptions` here but type is called `ECChangesetRowAdapterOptions;`. As I mention you can call it ChangeRowOption or formatOption etc.

core/backend/src/ECNativePartialChangeUnifier.ts
```diff
@@ -0,0 +1,284 @@
...
+  /** Insert or replace a cached instance. */
+  set(key: string, value: ECNativeChangeInstance): void;
+  /** Iterate over all cached instances. */
+  all(): IterableIterator<ECNativeChangeInstance>;
+  /** Number of instances currently in the cache. */
+  count(): number;
+}
+
+/** @beta */
+export namespace ECNativeChangeUnifierCache {
```
khanaffan, line 36: Again EC + Native

khanaffan (changes requested):

core/backend/src/ECChangesetReaderTypes.ts
soham-bentley: Yes changed to `RowFormatOptions`

core/backend/src/ECChangesetReaderTypes.ts
soham-bentley: Name and comment changed

core/backend/src/ECChangesetReaderTypes.ts
soham-bentley: Done

core/backend/src/ECChangesetReaderTypes.ts
soham-bentley: Done

core/backend/src/ECChangesetReaderTypes.ts
soham-bentley: Done

core/backend/src/ECChangesetReaderTypes.ts
soham-bentley: Done

core/backend/src/ECChangesetReaderTypes.ts
soham-bentley: Removed

core/backend/src/ECChangesetReaderTypes.ts
soham-bentley: Reused old type

core/backend/src/ECNativePartialChangeUnifier.ts
soham-bentley: Removed EC + Native

core/backend/src/ECChangesetReaderTypes.ts
soham-bentley: Reused old type

khanaffan (approved):

hl662 (approved):

aruniverse (approved):

PR #9143, opened 2026-03-27
QuantityFormatter: multi-system KoQ registry, FormatSpecHandle and add BeUnorderedEvent

hl662:
### Summary

Overhauls `QuantityFormatter`'s formatting spec management to support multi-unit-system caching and introduces new APIs for consuming format specs.

### Changes

**`@itwin/core-frontend` — `QuantityFormatter`**
- **Multi-system KoQ registry** — Now caches format specs for all 4 unit systems (metric, imperial, usCustomary, usSurvey) per KoQ, keyed by `name → persistenceUnit → system`. Enables `getSpecsByNameAndUnit({ name, persistenceUnitName, system? })` for direct access without triggering a reload.
- **`enqueueReload`** — Latest-wins serialized reload queue preventing thundering herd on rapid configuration changes.
- **`onBeforeFormattingReady`** — Collector-pattern event letting providers register async work that must complete before `onFormattingReady` fires.
- **`onFormattingReadyUnordered`** — Unordered variant of `onFormattingReady` for listeners that don't need guaranteed ordering.

**`@itwin/core-quantity`**
- **`FormatSpecHandle`** (`@beta`, `Disposable`) — Auto-refreshing handle that tracks a KoQ's formatter/parser specs and stays current via `onFormattingReadyUnordered`. Supports optional system pinning.
- **`FormattingReadyCollector`** — Awaits registered promises with configurable timeout.
- **`FormattingSpecProvider`** interface — Read-only projection of the KoQ registry for consuming format specs without direct `QuantityFormatter` coupling.
- **`FormattingSpecArgs`**, **`AddFormattingSpecArgs`**, **`FormatSpecHandleArgs`** — Structured args interfaces for registry and handle APIs.
- **`QuantityLoggerCategory`** — Structured logging categories (`Package`, `Formatting`).
- **`FormatsProvider.getFormat(name, system?)`** — Providers can now resolve formats for a specific unit system, enabling multi-system loading in a single pass.

**`@itwin/core-bentley`**
- **`BeUnorderedEvent` / `BeUnorderedUiEvent`** (`@beta`) — Set-backed event class with O(1) add, safe concurrent modification during emit via tombstone pattern.

**`@itwin/ecschema-metadata`**
- `SchemaFormatsProvider` and `FormatSetFormatsProvider` updated to pass `system?` through to `FormatsProvider.getFormat()`.

**Documentation**
- New advanced guide: `docs/quantity-formatting/usage/QuantityFormatterAdvanced.md`
- Changelog entry in `NextVersion.md`
- 10 compiled extract snippets covering all new APIs

**Tests**
- 14 new unit tests for `FormatSpecHandle` and `FormattingReadyCollector` in `@itwin/core-quantity` (caught a real constructor-order bug)
- 226 lines of `BeUnorderedEvent` tests in `@itwin/core-bentley` (re-entrant emit, concurrent modification, exception isolation)
- ~400 lines of new `QuantityFormatter` integration tests in `@itwin/core-frontend` (multi-system KoQ, collector pattern, FormatSpecHandle lifecycle)

### DIFF
--- a/example-code/snippets/src/frontend/Formatting.ts
+++ b/example-code/snippets/src/frontend/Formatting.ts
@@ -2,7 +2,7 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
-/* eslint-disable no-console */
+import { assert } from "@itwin/core-bentley";
 import { IModelApp, IModelConnection, QuantityType } from "@itwin/core-frontend";
 import { SchemaUnitProvider } from "@itwin/ecschema-metadata";
 
@@ -96,7 +96,7 @@ export async function waitForFormatterReady() {
   const formatterSpec = IModelApp.quantityFormatter.findFormatterSpecByQuantityType(QuantityType.Length);
   if (formatterSpec) {
     const formatted = IModelApp.quantityFormatter.formatQuantity(1.5, formatterSpec);
-    console.log(formatted);
+    assert(formatted.length > 0);
   }
 }
 // __PUBLISH_EXTRACT_END__
@@ -140,7 +140,7 @@ export function useFormatSpecHandle() {
 
   // format() returns a fallback string if specs aren't loaded yet
   const formatted = handle.format(1.5);
-  console.log(formatted);
+  assert(formatted.length > 0);
 
   // Explicitly dispose when done to unsubscribe from reload events
   handle[Symbol.dispose]();
@@ -157,7 +157,7 @@ export function useFormatSpecHandleWithUsing() {
 
   // The handle auto-refreshes when the formatter reloads
   const formatted = handle.format(2.75);
-  console.log(formatted);
+  assert(formatted.length > 0);
   // handle is automatically disposed at end of scope
 }
 // __PUBLISH_EXTRACT_END__
@@ -173,12 +173,11 @@ export function lookupSpecsByNameAndUnit() {
   if (entry) {
     // Format a value
     const formatted = IModelApp.quantityFormatter.formatQuantity(3.14, entry.formatterSpec);
-    console.log(formatted);
+    assert(formatted.length > 0);
 
     // Parse a user-entered string
     const result = entry.parserSpec.parseToQuantityValue("3.14 m");
-    if (result.ok)
-      console.log(`Parsed: ${result.value}`);
+    assert(result.ok && typeof result.value === "number");
   }
 }
 // __PUBLISH_EXTRACT_END__
@@ -192,7 +191,7 @@ export function formatKoqInMultipleSystems(meters: number) {
   if (metricSpec && imperialSpec) {
     const metricStr = IModelApp.quantityFormatter.formatQuantity(meters, metricSpec.formatterSpec);
     const imperialStr = IModelApp.quantityFormatter.formatQuantity(meters, imperialSpec.formatterSpec);
-    console.log(`Metric: ${metricStr}, Imperial: ${imperialStr}`);
+    assert(metricStr.length > 0 && imperialStr.length > 0);
   }
 }
 // __PUBLISH_EXTRACT_END__


core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -456,6 +574,65 @@ export class QuantityFormatter implements UnitsProvider {
...
+      await reloadFn();
+    } catch (err) {
+      Logger.logError(`${FrontendLoggerCategory.Package}.QuantityFormatter`, BentleyError.getErrorMessage(err));
+      this._reloadInFlight = false;
+      // If there's a pending reload, still try to run it
+      if (this._pendingReload) {
+        const next = this._pendingReload;
+        this._pendingReload = undefined;
+        return this.enqueueReload(next);
+      }
```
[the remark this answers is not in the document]
hl662: Good catch — replaced `wasReady` with a persistent `_hasEverBeenReady` flag set in `finalizeReload`. Now the catch block checks `_hasEverBeenReady` instead of a local that gets lost across chained `enqueueReload` calls. Fixed in 0c64670.

core/bentley/src/BeEvent.ts
```diff
@@ -141,6 +141,135 @@ export class BeUiEvent<TEventArgs> extends BeEvent<(args: TEventArgs) => void> {
...
+  listener: Listener | undefined;
+  scope: any;
+  once: boolean;
+}
+
+/**
+ * Manages a set of *listeners* for a particular event and notifies them when the event is raised.
+ * Unlike [[BeEvent]], this class uses a `Set` internally for O(1) add/remove and safe concurrent
+ * modification during emit. When a listener is removed during emit, it is marked for deferred
+ * removal instead of mutating the set immediately.
```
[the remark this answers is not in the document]
hl662: Fair point — `addListener` is O(1) but `removeListener` scans the Set, so the O(1) claim was inaccurate. Fixed the class doc to drop that claim, and added a note on `addListener`'s return value recommending the closure for O(1) removal. `removeListener` exists for API compatibility with `BeEvent`, not hot-path usage. Fixed in 0c64670.

core/frontend/src/test/QuantityFormatter.test.ts
```diff
@@ -736,3 +758,283 @@ describe("Test Formatted Quantities", async () => {
...
+  });
+
+  it("onFormattingReady fires AFTER onActiveFormattingUnitSystemChanged", async () => {
+    const qf = new QuantityFormatter();
+    await qf.onInitialized();
+    const callOrder: string[] = [];
+    qf.onActiveFormattingUnitSystemChanged.addListener(() => { callOrder.push("systemChanged"); });
+    qf.onFormattingReady.addListener(() => { callOrder.push("formattingReady"); });
+    await qf.setActiveUnitSystem("metric");
+    expect(callOrder).toEqual(["formattingReady", "systemChanged"]);
```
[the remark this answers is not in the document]
hl662: Title was wrong, not the assertion. `formattingReady` fires inside `finalizeReload`, then `systemChanged` fires in the `.then()` after `enqueueReload` resolves — so the array order is correct. Fixed the title to `"onActiveFormattingUnitSystemChanged fires AFTER onFormattingReady"`. Fixed in 0c64670.

core/quantity/src/FormatSpecHandle.ts
```diff
@@ -0,0 +1,112 @@
...
+  public format(value: number): string {
+    if (!this._formatterSpec)
+      return value.toString();
+    return this._provider.formatQuantity(value, this._formatterSpec);
+  }
+
+  /** Unsubscribe from reload events. Idempotent and safe to call during a pending reload.
+   * Also accessible via `[Symbol.dispose]` for use with `using` declarations.
+   */
+  public dispose(): void {
```
bsy-nicholasw: Why have a duplicate dispose method?

core/bentley/src/BeEvent.ts
```diff
@@ -141,6 +141,141 @@ export class BeUiEvent<TEventArgs> extends BeEvent<(args: TEventArgs) => void> {
...
+}
+
+/**
+ * Manages a set of *listeners* for a particular event and notifies them when the event is raised.
+ * Unlike [[BeEvent]], this class uses a `Set` internally to support safe concurrent modification
+ * during emit. When a listener is removed during emit, it is marked for deferred removal instead
+ * of mutating the set immediately.
+ * @beta
+ */
+export class BeUnorderedEvent<T extends Listener> {
```
bsy-nicholasw, line 165: I see a lot of mention about safe concurrent modification, here and also in the NextVersion.md. TBH the biggest motivation wasn't so much safety, but performance when there's thousands of listeners in play. The normal BeEvent should be safe, although perhaps not as much for nested removals after seeing some of your tests for the unordered one...

core/quantity/src/FormatSpecHandle.ts
hl662: Agreed — removed it. `[Symbol.dispose]()` is now the single implementation.

core/bentley/src/BeEvent.ts
hl662: Good point — the perf story is the primary motivation; I'll tone down the "safe concurrent modification" framing.

That said, looking into it I found `BeEvent` did have a real gap here: `_insideRaiseEvent` is a boolean, so nested `raiseEvent()` resets it and runs the array filter mid-outer-iteration. Also, a listener removing itself during emit never set the `dropped` flag — tombstone was never swept.

Ported the depth-counter fix back to `BeEvent` in this PR: `_emitDepth++/--`, sweep only at depth 0, post-callback tombstone check. Added tests for re-entrant emit, self-removal, and exception isolation.

core/bentley/src/BeEvent.ts
```diff
@@ -141,6 +145,141 @@ export class BeUiEvent<TEventArgs> extends BeEvent<(args: TEventArgs) => void> {
...
+          this._listeners.delete(ctx);
+      }
+    }
+  }
+
+  /** Determine whether this BeUnorderedEvent has a specified listener registered.
+   * @param listener The listener to check.
+   * @param scope Optional scope argument to match call to addListener.
+   */
+  public has(listener: T, scope?: any): boolean {
```
a-gagnon: I've got a few hot takes (🔥) about events.
Since this is new code that is not yet bound by API support policies, I thought I'd speak my mind.

- Order should never matter for events
The unordered event is definitely a step towards the right direction. If order **does matter**, we're doing it wrong and should use a different object/class/concept instead.

- I would refrain from exposing `has()` and `removeListener()`
Using `has()` always felt like an antipattern (at best). When registering an event, it is the caller's responsibility to capture the function returned by `addListener()` somewhere and call it at a later point. If that variable is not set, then it is not registered.

With that in mind, `removeListener()` becomes somewhat irrelevant. Adding to this, the current approach may end up looping over every entry of the map to remove a listener, while the function returned by `addListener()` could do it efficiently by leveraging the map, eg:

```ts
  // Set of listeners to drop. Used to collect entries when we're currently emitting.
  private _listenersPendingRemoval = new Set<UnorderedEventContext>();

  public addListener(listener: T, scope?: any): () => void {
    const ctx: UnorderedEventContext = { listener, scope, once: false };
    this._listeners.add(ctx);

    return () => {
      if (this._emitDepth > 0) {
        this._listenersPendingRemoval.add(ctx);
        return;
      }
      this._listeners.delete(ctx);
    };
  }
```

core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -456,6 +505,80 @@ export class QuantityFormatter implements UnitsProvider {
     }
   }
 
+  /** Enqueue an async reload. If no reload is in flight, runs immediately. If a reload is
+   * already in flight, stores `reloadFn` as pending (latest-wins: only the last queued reload
+   * is kept). `finalizeReload()` fires only when the queue is fully drained.
+   * @internal
+   */
+  protected async enqueueReload(reloadFn: () => Promise<void>): Promise<void> {
```
a-gagnon: If the last one wins, why do we need to wait until it's finished to kickoff `this._pendingReload` ?

Consider using [OneAtATimeAction](https://www.itwinjs.org/reference/core-bentley/utils/oneatatimeaction/). I believe it does exactly what you're looking for here.

.ecc-design/qty-improvements-pr-walkthrough.html
```diff
@@ -0,0 +1,769 @@
+<!DOCTYPE html>
```
rschili: Just as a reminder, we'll want to remove this before merge

core/frontend/tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
...68,1269,1270,1271,1272,1273,1274,1275,1277,1278,1279,1280,1281,1282,1283,1284,1285,1286,1287,1288,1289,1290,1291,1292,1293,1294,1295,1296,1297,1298,1299,1300,1301,1302,1303,1304,1305,1306,1307,1308,1309,1310,1311,1312,1313,1357,1358,1359,1360,1361,1362,1363,1364,1365,1366,1367,1368,1369,1370,1371,1372,1373,1374,1375,1376,1377,1378,1379,1380,1381,1382,1383,1384,1385,1386,1387,1388,1389,1390,1392,1393,1394,1395,1396,1397,1398,1399,1400,1401,1402,1403,1404,1405,1406,1407,1408,1409,1454,1464],[405,434,456,457,458,459,460,1208,1223,1262,1263,1264,1265,1266,1267,1268,1269,1270,1271,1272,1273,1274,1275,1277,1278,1279,1280,1281,1282,1283,1284,1285,1286,1287,1288,1289,1290,1291,1292,1293,1294,1295,1296,1297,1298,1299,1300,1301,1302,1303,1304,1305,1306,1307,1308,1309,1310,1311,1312,1313,1357,1358,1359,1360,1361,1362,1363,1364,1365,1366,1367,1368,1369,1370,1371,1372,1373,1374,1375,1376,1377,1378,1379,1380,1381,1382,1383,1384,1385,1386,1387,1388,1389,1390,1392,1393,1394,1395,1396,1397,1398,1399,1400,1401,1402,1403,1404,1405,1406,1407,1408,1409,1454,1464],[139,339,1208,1223,1262,1263,1264,1265,1266,1267,1268,1269,1270,1271,1272,1273,1274,1275,1277,1278,1279,1280,1281,1282,1283,1284,1285,1286,12
```
rschili: temp build artifact included in PR? There is another one in quantity

core/ecschema-metadata/src/Formatting/FormatSetFormatsProvider.ts
```diff
@@ -61,7 +61,7 @@ export class FormatSetFormatsProvider implements MutableFormatsProvider {
    * Retrieves a format definition by its name from the format set. If not found, it checks the fallback provider to find the format, else returns undefined.
    * If the format is a string reference to another format, it resolves the reference and returns the FormatDefinition.
    */
-  public async getFormat(input: string): Promise<FormatDefinition | undefined> {
+  public async getFormat(input: string, system?: UnitSystemKey): Promise<FormatDefinition | undefined> {
```
rschili, line 64: I just realized there is tension between UnitSystemKey and UnitSystem. UnitSystemKey is 4 static strings - compared to what our Units schema exposes in UnitSystem elements

```xml
    <UnitSystem typeName="SI" />
    <UnitSystem typeName="CGS" />
    <UnitSystem typeName="METRIC" />
    <UnitSystem typeName="IMPERIAL" />
    <UnitSystem typeName="MARITIME" />
    <UnitSystem typeName="USSURVEY" />
    <UnitSystem typeName="INDUSTRIAL" />
    <UnitSystem typeName="INTERNATIONAL" />
    <UnitSystem typeName="USCUSTOM" />
    <UnitSystem typeName="STATISTICS" />
    <UnitSystem typeName="FINANCE" />
    <UnitSystem typeName="CONSTANT" />
```

All of this is preexisting, so nothing you introduce in this PR. But for consumers this will be as confusing as it is for me. 5 EC systems (CGS, MARITIME, INDUSTRIAL, STATISTICS, CONSTANT) are not mapped to UnitSystemKey at all.

I wonder. We should at least add some documentation on UnitSystemKey which clarifies that this is a user preference that indicates which systems to prefer, and how they map to actual UnitSystems. If I had to design it from scratch I'd probably call it "UnitSystemGroup" or something like that, if I interpret its purpose right.

core/frontend/src/test/QuantityFormatter.test.ts
```diff
@@ -736,3 +758,455 @@ describe("Test Formatted Quantities", async () => {
     await testFormatting(QuantityType.Volume, 1000.0, "35314.4548 ft³");
   });
 });
+
+describe("Reload queue and onFormattingReady", () => {
```
rschili, line 762: I believe we need additional tests to cover:

1. **`_rebuildRegistryFromProvider` path** All current tests exercise `_reloadCore` (via `onInitialized`, `setActiveUnitSystem`, etc.), but the production path where `IModelApp.formatsProvider.onFormatsChanged` fires with specific KoQ names and triggers `_rebuildRegistryFromProvider` has no coverage. This includes the selective per-name rebuild, stale entry deletion, and the `anySystemHadFormat` cleanup. We should include a test that sets a custom `FormatsProvider`, fires `onFormatsChanged` with a specific `formatsChanged` list, and verifies the registry entries update correctly.

2. **`impliedUnitSystem` propagation looks like it has a race condition.** The `impliedUnitSystem` check and the `onActiveFormattingUnitSystemChanged` emit happen in a `.then()` after `enqueueReload` resolves - but since `enqueueReload` is latest-wins, two rapid `onFormatsChanged` events with different `impliedUnitSystem` values could cause the first `.then()` to fire after the second reload completes, emitting a stale system-changed event. A test that fires two `onFormatsChanged` with different `impliedUnitSystem` in quick succession would either confirm this is a bug or prove me wrong.

core/bentley/src/BeEvent.ts
hl662: This makes sense to me. Alright I made the changes:
- Removed has() and removeListener() entirely. The closure returned by addListener/addOnce is now the only way to unsubscribe. (As I type this out I realize maybe I should disclose the motivation in NextVer.md so others can avoid the antipattern moving forward)
- Closures capture the ctx object directly and call Set.delete(ctx) — O(1) removal instead of the O(n) scan removeListener was doing.
- Used a tombstone approach (set ctx.listener = undefined) during emit rather than a separate _listenersPendingRemoval set — avoids the extra allocation and the cleanup sweep can reuse the same iteration over _listeners

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: Looked into this — the queuing semantics are similar but I'm not sure that OneAtATimeAction is a fit here, enqueueReload handles a lot more cases, and OneAtATime has some differences:

1. **Error model mismatch.** `OneAtATimeAction` rejects superseded requests with `AbandonedError`, requiring every call site to catch it. `enqueueReload` silently drops superseded reloads — several call sites fire-and-void (`void this.enqueueReload(...)`) and don't await or catch. Wrapping every site in try/catch for an error we intentionally don't care about would be noise.... unless you want to await and catch?

2. **Ready-state lifecycle is interleaved.** `enqueueReload` isn't just a queue, it's a state machine that manages `_isReady`, `_hasEverBeenReady`, and `_reloadInFlight`. It sets `_isReady = false` at the start, calls `finalizeReload()` (which fires `onFormattingReady`, resolves `whenInitialized`, and runs the `onBeforeFormattingReady` collector) only when the queue fully drains, and restores stale ready state on error if the formatter was previously initialized. Extracting the queue into `OneAtATimeAction` would push all of that into wrapper logic around each call. Its the same complexity, just split across two places.

3. **"why do we need to wait until it's finished to kickoff this._pendingReload?"** because `finalizeReload` is async (it awaits provider work via `onBeforeFormattingReady`) and we need to know whether the queue is truly drained before firing ready events. If we kicked off the pending reload immediately, we'd either finalize prematurely or need a separate drain-detection mechanism.

hl662: Going deeper into the lifecycle:
 - _isReady — "Can consumers use format specs right now?" false during any reload, true after the queue fully drains and finalizeReload  completes.
 - _hasEverBeenReady — "Has the formatter ever successfully loaded?" Used for error recovery: if a reload fails but we were ready before, restore _isReady = true so consumers can still use stale-but-valid cached specs instead of being stuck in a permanently "not ready" state.
 - _reloadInFlight — "Is the queue currently processing?" Guards against re-entrance into enqueueReload.

.ecc-design/qty-improvements-pr-walkthrough.html
hl662: Thanks, removed

core/frontend/tsconfig.tsbuildinfo
hl662: thanks, removed

core/ecschema-metadata/src/Formatting/FormatSetFormatsProvider.ts
hl662: Yeah, as a stopgap I added documentation to close the gap you're describing:
 - Added JSDoc on UnitSystemKey itself explaining it's a user-facing preference key, not an EC schema identifier, and documenting how each key maps to EC systems (e.g. "metric" → SI, CGS, METRIC; "imperial" → IMPERIAL)
 - Updated docs/bis/ec/ec-unitsystem.md with a mapping table showing which EC UnitSystems are covered and which are unmapped (i.e the 5 you mentioned)
 - Updated docs/quantity-formatting/definitions/Units.md and Providers.md with cross-references

core/frontend/src/test/QuantityFormatter.test.ts
hl662: You were right on both counts.

1. **_rebuildRegistryFromProvider** - added two tests: one for the formatsChanged: "all" path (verifies custom entries get cleaned up when the provider doesn't return them), and one for the targeted string[] path (verifies only named formats are rebuilt while others are untouched).

2. **impliedUnitSystem race condition** - confirmed as a real bug. The .then() on enqueueReload fires immediately for superseded reloads (the promise resolves when the reload is queued, not when it runs), so rapid events would emit stale system-changed events.

- Fixed by moving the emission into a deferred pattern: the reloadFn sets _deferredSystemChangedEmit, and finalizeReload() emits it after isReady === true. Also clear the deferred field at reload cycle start so a failed reload can't leak a stale emit into the next successful finalization. I added 3 tests for the fix: deferred emit fires only after isReady, latest-wins supersession (two rapid events → only winning system emitted), and no spurious emit when impliedUnitSystem is undefined.

core/frontend/src/test/QuantityFormatter.test.ts
```diff
@@ -633,6 +635,26 @@ describe("Quantity formatter", async () => {
       expect(spy.mock.calls[1][0]).toEqual({ formatsChanged: "all" });
 
     });
+
+    it("should not leak listeners when formatsProvider is replaced multiple times", () => {
+      const provider1 = new QuantityTypeFormatsProvider();
+      const provider2 = new QuantityTypeFormatsProvider();
+
+      IModelApp.formatsProvider = provider1;
```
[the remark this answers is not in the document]
hl662: intentional. It registers an internal forwarding listener on provider1.onFormatsChanged. The whole point is proving that listener gets cleaned up when provider2 replaces it.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -456,6 +521,120 @@ export class QuantityFormatter implements UnitsProvider {
...
+   * already in flight, stores the intent as pending (latest-wins: only the last queued reload
+   * is kept). `finalizeReload()` fires only when the queue is fully drained.
+   *
+   * Reload intents:
+   * - `"full"` — rebuild entire registry + re-register provider listener (onInitialized, setUnitsProvider)
+   * - `"formatsChanged"` — patch registry from provider + load maps (formatsChanged listener)
+   * - `"activeSystem"` — reload format/parsing maps for current unit system (setActiveUnitSystem, reinitializeFormatAndParsingsMaps)
+   * @internal
+   */
+  protected async enqueueReload(intent: ReloadIntent): Promise<void> {
```
ben-polinsky: We've been discussing offline, but figured I would put my concerns down. Since this isn't really a queue, it could be misleading - only one intent can be stashed at a time. You do call this out "latest-wins" but maybe a note about how waiting for the emitted event is the only way to be sure that the change has been processed would be beneficial.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -773,7 +978,7 @@ export class QuantityFormatter implements UnitsProvider {
       return;
 
     this._activeUnitSystem = systemType;
-    await this.loadFormatAndParsingMapsForSystem(systemType);
+    await this.enqueueReload({ scope: "activeSystem" });
     // allow settings provider to store the change
     await this._unitFormattingSettingsProvider?.storeUnitSystemSetting({ system: systemType });
     // fire current event
```
ben-polinsky: If I am reading this correctly, you emit `onActiveFormattingUnitSystemChanged` without waiting for `this.finalizeReload()`. If no intent is yet queued, this will roll through and call `finalizeReload()` and the event is safe to emit after. However, if there is an intent queued, `enqueueReload()` will return early without `finalizeReload()` having been called and the event is thus not yet accurate?

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: Good call — renamed to scheduleReload and expanded the TSDoc to call out that the returned promise resolves when the method exits, not when the reload actually runs.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: Yeah you're right — that was a bug. When a reload is already in-flight, scheduleReload returns immediately and the emit fires before specs are rebuilt.
 
Fixed by threading emitSystemChanged through the ReloadIntent so the event gets deferred to finalizeReload(), after specs are ready

ben-polinsky: Thanks. Perhaps a test to solidify?

docs/changehistory/NextVersion.md
```diff
@@ -113,3 +250,41 @@ Delete it:
...
+
+##### `@itwin/core-quantity`
+
+- **`Units` namespace** — Typed constants for commonly used unit names (e.g., `Units.M`, `Units.FT`, `Units.RAD`). Eliminates magic strings when referencing units programmatically.
+- **`findPersistenceUnitForPhenomenon()`** — Maps phenomenon names to their canonical SI persistence unit.
+- **`FormatsChangedArgs.impliedUnitSystem`** — Allows a `FormatsProvider` to indicate which unit system its format set implies.
+
+##### `@itwin/core-frontend`
+
+- **[QuantityFormatter.onBeforeFormattingReady]($frontend)** — Pre-ready event that fires before [QuantityFormatter.onFormattingReady]($frontend). Providers use it to register async work (e.g., loading domain formats) via the [FormattingReadyCollector]($quantity) passed to listeners. The formatter awaits all pending work (with a 10-second timeout) before emitting `onFormattingReady`.
```
ben-polinsky: This is the timeout in `FormattingReadyCollector.awaitAll`? If so, that defaults to 20 seconds (says 10 here) and can be overridden, so I would revise this phrasing.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
```diff
@@ -434,13 +449,63 @@ export class QuantityFormatter implements UnitsProvider {
...
+  /** Whether the QuantityFormatter has completed at least one successful reload and is ready to format/parse.
+   * @beta
+   */
+  public get isReady(): boolean { return this._isReady; }
+
+  /** A promise that resolves after the first successful initialization. This is one-shot — it resolves once
+   * and stays resolved forever. For subsequent reloads, subscribe to [[onFormattingReady]].
+   * @beta
+   */
+  public get whenInitialized(): Promise<void> { return this._initializedPromise; }
```
ben-polinsky, line 486: Thinking through this, if one error occurs - this stays pending - that's okay as long as another attempt succeeds. But what if nothing is every re-attempted or the successive attempt also fails? 

Maybe that is alright, and your comment implies that behavior, but is there a time when you can definitively tell the initialization has failed (and will not recover) and you could reject?

pmconne:
> BeUnorderedEvent / BeUnorderedUiEvent (@beta) — Set-backed event class with O(1) add, safe concurrent modification during emit via tombstone pattern.

I am curious why an entirely new class was needed, vs modifying `BeEvent`.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: Good call — added a test: `"onActiveFormattingUnitSystemChanged deferred when reload already in-flight"`. It starts a full reload, calls `setActiveUnitSystem` while the reload is in-flight, and asserts the system-changed event fires *after* `onFormattingReady` (not before specs are ready). See 9ea476cabf.

docs/changehistory/NextVersion.md
hl662: You're right — the default is `20_000ms`. Fixed the docs and TSDoc to say "20-second default" in both `NextVersion.md` and the `QuantityFormatter.ts` comment.

core/frontend/src/quantity-formatting/QuantityFormatter.ts
hl662: By design — `whenInitialized` is a one-shot promise that resolves on first successful init but intentionally never rejects. There's no finite retry limit (any subsequent reload success will resolve it), so rejecting would close the door on recovery and force every caller to add `.catch()` for a scenario that may self-heal.

Added a clarifying TSDoc comment explaining this contract.

hl662:
> I am curious why an entirely new class was needed, vs modifying `BeEvent`.

Two things make `BeEvent` unsuitable for a Set backing:

1. **Add-during-emit semantics change.** `BeEvent` uses a snapshot-then-iterate array loop — listeners added mid-emit are never visited in that emission round. `Set.forEach` *does* visit entries added during iteration (per the ES spec). That's a silent behavior change across 103+ existing `@public` sites.

2. **`removeListener(fn, scope)` makes Set pointless.** It requires an O(n) linear scan regardless of backing store, since it matches by value equality on `(fn, scope)` pairs.

`BeUnorderedEvent` earns its existence by dropping `removeListener` entirely. The only way to unsubscribe is the disposal closure returned by `addListener()`, which does a single `Set.delete` — O(1). That's the pattern `FormatSpecHandle` needs: high-churn subscribe/dispose cycles with potentially thousands of handles.

That said, I collapsed the two events — `onFormattingReady` is now the single event, backed by `BeUnorderedUiEvent`. No need for a separate `onFormattingReadyUnordered` since both were new in this PR.

example-code/snippets/src/frontend/Formatting.ts
```diff
@@ -84,3 +85,175 @@ export async function enableMathematicalOperations() {
...
+/** Wait for the QuantityFormatter to complete its first initialization */
+export async function waitForFormatterReady() {
+  // Resolves after the first successful initialization; safe to call multiple times
+  await IModelApp.quantityFormatter.whenInitialized;
+
+  // Formatting APIs are now safe to use
+  const formatterSpec = IModelApp.quantityFormatter.findFormatterSpecByQuantityType(QuantityType.Length);
+  if (formatterSpec) {
+    const formatted = IModelApp.quantityFormatter.formatQuantity(1.5, formatterSpec);
+    console.log(formatted);
```
MichaelSwigerAtBentley: Any reason to console log in example code?

hl662: without the console.log we'll get no-unused-vars lint problems in the example code, which I guess I can disable line by line, but that still kind of sucks. so I left console.logs in here.

MichaelSwigerAtBentley: For unused-vars you can just put a `_` in front of the name right?

hl662: yeah.... I don't like the _, it could insinuate that this is not valuable and shouldn't be used (which is the opposite of what these extracted code docs should be communicating).

Instead, I removed the console.log() and replaced with assert() calls using assert() from core-bentley

PR #9151, opened 2026-03-29
SchemaView - light schema access layer

rschili:
imodel-native: https://github.com/iTwin/imodel-native/pull/1369

Some diagrams refer to this as "Runtime Schemas" which was my working title. I didn't update all the pictures, just be aware. The final name we're going with is SchemaView.

### Motivation

Yet another ECSchema layer for typescript, so let me explain:

- **ecschema-metadata** (`SchemaContext`) is the full-fidelity schema toolkit. It models every detail of the EC specification, but loading it is expensive: one synchronous RPC per schema, large memory footprint (~150-200 MB for a large iModel), and it **blocks the backend event loop during loading** which is the primary concern here.
- **Incremental schema loading** was implemented as a lighter alternative but remains incomplete and does not cover all use cases.
- **ECSQL against ECDbMeta** works for targeted queries but is awkward for traversal patterns (e.g. walking all inherited properties of a class) and does not cache on the TS side. For one-shot questions this is perfectly viable!
- Older backend metadata layer was discontinued and pulled single classes - this is somewhat similar to that again but loads one bigger chunk at once.

This PR introduces an optimized schema bundle which is transported via a single async call and parsed into an in-memory cache that provides fast synchronous access. It is lossy by design - it drops what runtime consumers do not need (units, formats, phenomena, full custom attribute instances) to keep the payload small and parsing fast.

Why one single bundle?
I did extensive testing. Of course it could be broken into smaller chunks, lazy loaded, but we have done all of that before - every approach has tradeoffs. This is keeping it simple and in testing performs rather well, even in scnearios in which it is weak by design.

The most extreme example is a single "check if class A IS of class B" on an iModel with huge schemas, magnitudes larger than an average iModel.
In this example, SchemaView takes about a .8 seconds spinup delay, then it responds instantaneously.
Asking the question directly via ECSql takes about 0.002s. With smaller iModels SchemaView catches up faster - ECSql is always available as an alternative option.

### Design

**Transport**: A new `PRAGMA schema_view(N)` in ECDb returns the binary blob via ConcurrentQuery. No new RPC methods - it flows through the existing `queryRows` path. The pragma accepts a format version parameter for forward compatibility.

**Binary format**: C++ writer reads `ec_` tables directly and produces a compact blob with:
- String lookup table (deduplicates repeated schema/class/property names)
- Property definition deduplication (many class-property pairs share identical definitions)
- Per-item `ecInstanceId` (row ID from `ec_` tables) for fallback to ECSQL when needed

**Cache invalidation**: Uses `PRAGMA checksum(ecdb_schema)` (SHA3-256) to detect schema changes. Schema imports or changeset pulls that modify schemas automatically invalidate the cache.

### Performance characteristics

Initial investigation into how much metadata we ingest, tested on an iModel with enormous metadata and poor performance:

grafik

Performance comparison walking all properties:

grafik

Performance comparison on simulated frontend with limited network bandwidth:

grafik

Memory footprint at runtime:

grafik


hl662:
When would users want to use this compared to the full fledged schema context?

rschili:
> When would users want to use this compared to the full fledged schema context?

I was hoping for the answer to be "always" unless they are trying to store/author schemas.

core/ecschema-metadata/src/RuntimeSchema.ts
```diff
@@ -0,0 +1,709 @@
...
+  /** Persistence unit as a full name string, e.g. "Units:M". */
+  public get persistenceUnit(): string {
+    const sid = this._data.persistenceUnitSid;
+    return sid !== 0 ? this._ctx.strings[sid] : "";
+  }
+  /** Presentation format strings, semicolon-separated. Empty string if none. */
+  public get presentationUnits(): string {
+    const sid = this._data.presentationUnitsSid;
+    return sid !== 0 ? this._ctx.strings[sid] : "";
+  }
```
grigasp: Shouldn't this be named `presentationFormats`? And maybe even have `Def`, `Spec` or similar suffix to highlight the fact that it's an unparsed string rather than an actual list of formats.

core/frontend/src/IModelConnection.ts
```diff
@@ -640,6 +642,52 @@ export abstract class IModelConnection extends IModel {
...
+   * Multiple concurrent callers share a single in-flight fetch.
+   *
+   * The returned `RuntimeSchemaContext` is a lightweight, read-only, synchronous API for
+   * navigating schema metadata - classes, properties, relationships, enumerations, etc.
+   * It is designed as a faster, lower-memory alternative to `schemaContext` (ecschema-metadata).
+   * @beta
+   */
+  public async getSchemas(): Promise<RuntimeSchemaContext> {
+    if (this._schemas !== undefined && !this._schemas.isOutdated && this._schemasPromise === undefined)
+      return this._schemas;
```
grigasp: Is there a point to have `this._schemas`? It seems having just the `this._schemasPromise` would suffice.

docs/learning/metadata/index.md
```diff
@@ -11,7 +14,7 @@ The API is split into these packages:
 
 ## Obtaining metadata from an imodel
 
-An [IModelDb]($backend) owns a [SchemaContext]($ecschema-metadata) which can be used to access all the meta information stored inside of the imodel.
+An [IModelDb]($backend) owns a [SchemaContext]($ecschema-metadata) which can be used to access all the meta information stored inside of the imodel. For a faster alternative optimized for runtime lookups, see [RuntimeSchemaContext](./RuntimeSchemaContext.md).
```
grigasp: To me personally, the added sentence doesn't explain when to use `RuntimeSchemaContext` vs `SchemaContext`.

docs/learning/metadata/RuntimeSchemaContext.md
```diff
@@ -0,0 +1,149 @@
+# RuntimeSchemaContext
+
+`RuntimeSchemaContext` is a high-performance, read-only schema metadata cache available in both backend and frontend. It loads a lossy optimized version of schemas from an iModel in a single call and provides synchronous access to schemas, classes, properties, enumerations, kinds of quantity, and relationship constraints.
+
+It lives in `@itwin/ecschema-metadata` and is the recommended way to access schema metadata at runtime when you need fast, repeated lookups - for example in presentation rules, property grids, or data-driven UI.
+
+For the binary transport format specification, see [RuntimeSchemaBinaryFormat.md](./RuntimeSchemaBinaryFormat.md).
+
+## Why not ecschema-metadata?
```
grigasp: Maybe use `SchemaContext` term instead of `ecschema-metadata`? Currently it's a bit confusing, since `RuntimeSchemaContext` also lives in `ecschema-metadata`.

core/ecschema-metadata/src/RuntimeSchema.ts
```diff
@@ -0,0 +1,709 @@
...
+    const labelSid = this._ref.labelSid;
+    return labelSid !== 0 ? this._ctx.strings[labelSid] : this.name;
+  }
+  public get description(): string {
+    const sid = this._def.descriptionSid;
+    return sid !== 0 ? this._ctx.strings[sid] : "";
+  }
+  public get kind(): PropertyKind { return this._def.kind; }
+  public get isReadOnly(): boolean { return this._def.isReadOnly; }
+  public get isHidden(): boolean { return this._def.isHidden; }
```
grigasp: What's the meaning of this property?

docs/learning/metadata/RuntimeSchemaContext.md
```diff
@@ -0,0 +1,149 @@
...
+
+## View objects and allocation
+
+`RuntimeSchema`, `RuntimeClass`, `RuntimeProperty`, and the other view types are lightweight wrappers holding only a context reference and an index. They do not cache data and are not identity-stable - calling `element.schema` twice returns two distinct objects that expose the same data. This means `===` comparison will fail; use `name` or `fullName` for equality checks.
+
+Calling `getProperties()` allocates a new `RuntimeProperty` wrapper for each property on every call. For hot loops, consider caching the result in a local variable. The underlying data is shared - only the thin wrapper objects are allocated.
+
+## Excluded schemas and data completeness
+
+`RuntimeSchemaContext` intentionally excludes infrastructure schemas that are not useful at runtime: Units, Formats, ECDb-internal schemas (ECDbSystem, ECDbMap, etc.), and pure custom-attribute schemas (CoreCustomAttributes, EditorCustomAttributes, etc.). The full list is defined in the C++ writer's `IsExcludedSchema()` function in `RuntimeSchemaWriter.cpp`.
```
grigasp: Not sure I agree that units and formats aren't useful at runtime. When formatting numeric property values, we:
1. look up their kind-of-quantity,
2. take its presentation formats (full spec, not just the name or definition),
3. each format is based on a unit, we take the unit and determine whether to use the format based on unit's unit system.

More details here: https://github.com/iTwin/presentation/blob/232e6e5dbbed67bb228a4bd51834b8e06beff2ab/packages/core-interop/src/core-interop/Formatting.ts#L109.

With the suggested API, we get a single presentation formats' string. Is there a path forward for us to use this string to achieve what we're currently doing with `SchemaContext` API?

grigasp (commented):
Overall looks good.

A few comments:

- Not sure the prefix `Runtime` fits the purpose of the API. Based on the docs, it sounds like `SchemaContext` is reserved for editing use cases, but that happens during runtime as well.

- I think a better distinction between `schemaContext` and `getSchemas()` on `IModelConnection` and `IModelDb` is needed. Why one is a getter and the other is a method?

  The `schemaContext` getter is still marked `@beta`, so I think we could at least rename it. However, considering it becomes a niche API used only for specific situations, I would consider removing it completely and instead exposing a global factory function to create a `SchemaContext` when it's needed.

- Since `RuntimeSchemaContext` is "lossy", I think the docs should be very clear with recommendations on what to do when that lost information is needed.

  For example, in my case I used `SchemaContext` API to traverse from `Property` -> `KindOfQuantity` -> `Format` -> `Unit`. With the new API, `KindOfQuantity` only has presentation formats string. What do I do from there?

- I think we'll want to bump `IModelReadRpcInterface.version` to ensure frontends can use the new pragma.

docs/learning/metadata/RuntimeSchemaContext.md
rschili: > I adjusted the text to say "The rationale for Units/Formats being: We are in the process of decoupling those from schemas. In a yet to be shipped API they will be loaded separately, so this API will already only expose identifiers which will be used to perform the lookup."

I also removed the "are not useful" bit, that was just misleading.

You are right, we will need some form of helper method to handle these strings.
@hl662 : When we decouple units/formats from schemas we should follow up on this.

core/ecschema-metadata/src/RuntimeSchema.ts
rschili: It's a populated version of the "HiddenProperty" CA which is in CoreCustomAttributes.
I was looking for CAs that consumers will frequently need and tried to incorporate them in the API so it's easier to use them.

Do you think it's more confusing than helpful? Maybe this is not necessary..

docs/learning/metadata/RuntimeSchemaContext.md
rschili: good suggestion, I changed it. Initially I had RuntimeSchemaContext outside of that package when I wrote this.

docs/learning/metadata/index.md
rschili: Changed it into:

> An [IModelDb]($backend) owns a [SchemaContext]($ecschema-metadata) which can be used to access all the meta information stored inside of the imodel. For a lighter read-only alternative, see [RuntimeSchemaContext](./RuntimeSchemaContext.md).

core/ecschema-metadata/src/RuntimeSchema.ts
rschili: Right, I renamed it. I didn't realize we had this mismatch between ECXml and ecschema-metadata already. It's called formats now.

rschili:
@grigasp 

>     * Not sure the prefix `Runtime` fits the purpose of the API. Based on the docs, it sounds like `SchemaContext` is reserved for editing use cases, but that happens during runtime as well.

Fair point - "Runtime" is ambiguous since `SchemaContext` is also used at runtime. I thought hard about this, considered alternatives like `CompactSchemaContext`, `LightSchemaContext`, `SchemaSnapshot`, etc., but none clearly communicated the intent better.

The distinction we're drawing is "optimized for runtime consumption" (pre-loaded, read-only, synchronous, compact) vs "optimized for schema introspection and editing" (lazy-loaded, mutable, full-fidelity). "Runtime" is imperfect, changing it now would be a substantial update.
The prefix does carry the connotation of "this is the thing you reach for in hot paths." Open to a better name if one comes to mind, but the alternatives seemed worse.
 
>     * I think a better distinction between `schemaContext` and `getSchemas()` on `IModelConnection` and `IModelDb` is needed. Why one is a getter and the other is a method?
>       The `schemaContext` getter is still marked `@beta`, so I think we could at least rename it. However, considering it becomes a niche API used only for specific situations, I would consider removing it completely and instead exposing a global factory function to create a `SchemaContext` when it's needed.

I think keeping it as a getter is the right call here. `schemaContext` and `getSchemas()` serve fundamentally different access patterns - `schemaContext` returns a thin wrapper object instantly (the real work happens lazily on nested calls like `getSchemaItem()`), while `getSchemas()` does real async work up front to fetch and deserialize the binary blob. Making one a getter and the other a method reflects that difference: synchronous, cheap construction vs. async, non-trivial I/O. A factory function would obscure that distinction and make the simple case harder to reach. I agree the naming asymmetry looks inconsistent at first glance, but I think it's actually communicating something useful about the cost model.

>       For example, in my case I used `SchemaContext` API to traverse from `Property` -> `KindOfQuantity` -> `Format` -> `Unit`. With the new API, `KindOfQuantity` only has presentation formats string. What do I do from there?

For now you would have to use ECSql to pull that remaining info on demand. Let's dive into that and better underatand what we need. Modelling the whole units object model into this seems overkill to me. But I agree, if you DO need it, the path should be simple. Maybe RuntimeProperty could expose a helper method that does the work for you?

>     * I think we'll want to bump `IModelReadRpcInterface.version` to ensure frontends can use the new pragma.

done

core/frontend/src/IModelConnection.ts
rschili: `_schemas` gives us a synchronous fast-path (no microtask yield), while `_schemasPromise` tracks in-flight fetches for request coalescing. If we merged them into a single promise and used `Promise.resolve()` for the cached case, `await` on a resolved promise still yields one microtask tick, which opens a window where concurrent callers can both detect a stale cache and start duplicate fetches. It's most relevant on the backend where `refreshSchemas()` can externally trigger staleness, but I kept the frontend consistent for the same reason.

core/ecschema-metadata/src/RuntimeSchema.ts
grigasp: We do need this custom attribute to know if property should be displayed or not, so I think it's useful. Could we also have it for `HiddenClass` and `HiddenSchema` custom attributes?

core/frontend/src/IModelConnection.ts
rschili: Actually, I reconsidered, I'm changing it to a single promise field. There may be a micro cost that comes with it, but it's so much cleaner. Thanks!

grigasp:
@rschili 

> Fair point - "Runtime" is ambiguous since `SchemaContext` is also used at runtime. I thought hard about this, considered alternatives like `CompactSchemaContext`, `LightSchemaContext`, `SchemaSnapshot`, etc., but none clearly communicated the intent better.
> 
> The distinction we're drawing is "optimized for runtime consumption" (pre-loaded, read-only, synchronous, compact) vs "optimized for schema introspection and editing" (lazy-loaded, mutable, full-fidelity). "Runtime" is imperfect, changing it now would be a substantial update. The prefix does carry the connotation of "this is the thing you reach for in hot paths." Open to a better name if one comes to mind, but the alternatives seemed worse.

I would consider `ReadonlySchemaContext`.

> I think keeping it as a getter is the right call here. `schemaContext` and `getSchemas()` serve fundamentally different access patterns - `schemaContext` returns a thin wrapper object instantly (the real work happens lazily on nested calls like `getSchemaItem()`), while `getSchemas()` does real async work up front to fetch and deserialize the binary blob. Making one a getter and the other a method reflects that difference: synchronous, cheap construction vs. async, non-trivial I/O. A factory function would obscure that distinction and make the simple case harder to reach. I agree the naming asymmetry looks inconsistent at first glance, but I think it's actually communicating something useful about the cost model.

Personally, a getter vs a method wouldn't make me realize that one is more expensive than the other. FWIW, I would probably think a getter is cheaper.

However, I'd like to reiterate on my suggestion to remove `schemaContext` getter completely. If we say that it's going to be needed in less than 10% of use cases, then why confuse all the consumers with two similar but different options? We could provide quick & simple schema access that suits 90% of consumers, and the rest could be forwarded to an API that lets them create `SchemaContext`. I think it's fair to say that more complex use cases require more complex setup.

> But I agree, if you DO need it, the path should be simple. Maybe RuntimeProperty could expose a helper method that does the work for you?

I don't think we'll be able to stop looking at schema-based formats/units anytime soon. Even when the full replacement is ready, we'll probably want to look at what's in schema as a fallback. It doesn't necessarily have to be simple, but it has to be clear.

My preference would be for `RuntimeKoQ.presentationFormats` to return something like `Array<{ formatName: string; precision: number; unitName: string }>`. Example:
```ts
// koq format spec:
//   f:DefaultRealU(4)[u:M_PER_SEC_SQ];f:DefaultRealU(4)[u:CM_PER_SEC_SQ];f:DefaultRealU(4)[u:FT_PER_SEC_SQ]
// formats array:
[
  { formatName: "Formats.DefaultRealU", precision: 4, unitName: "Units.M_PER_SEC_SQ" },
  { formatName: "Formats.DefaultRealU", precision: 4, unitName: "Units.CM_PER_SEC_SQ" },
  { formatName: "Formats.DefaultRealU", precision: 4, unitName: "Units.FT_PER_SEC_SQ" },
]
```

That way we could use `formatName` and `unitName` to pull the details we need.

hl662:
> I don't think we'll be able to stop looking at schema-based formats/units anytime soon. Even when the full replacement is ready, we'll probably want to look at what's in schema as a fallback. It doesn't necessarily have to be simple, but it has to be clear.

I agree, format set fotmatsProvider already has the concept of a fallback provider to lean into, a schema formats provider naturally fits that. If need be, we can extend the formatsProvider interface to allow an optional fallback formatsProvider within, bringing that functionality out of FSFormatsProvider. And do something similar with unitsProvider and a fallback of some sort.

core/backend/src/IModelDb.ts
```diff
@@ -1713,6 +1715,100 @@ export abstract class IModelDb extends IModel {
...
+
+  /** Check if the cached runtime schema context is stale and reload if necessary.
+   * Uses a lightweight `PRAGMA checksum(ecdb_schema)` to compare tokens without building the full
+   * binary blob. If the token matches, returns the existing cached context. If it changed, builds
+   * a new context and marks the old one as outdated.
+   *
+   * If no context is cached yet, this behaves identically to `getSchemas()`.
+   * @beta
+   */
+  public async refreshSchemas(): Promise<RuntimeSchemaContext> {
```
grigasp: I see this is a `@public` method - does that mean we expect consumers to call it? How do they know when to call it?

rschili: I am still struggling with the decision. The check via pragma checksum is cheap but not zero cost. I don't want to do it always when somebody accesses the context.
So I want to have an explicit call for this. I am calling it from importSchemas, applyChangeset etc automatically. On frontend, you'll probably have to call it more often manually.
I'll keep this comment open until I found a good solution.

grigasp: FWIW, I think schema context invalidation **must** be handled internally. I really hope that consumers don't need to check the `isOutdated` flag or call `refreshSchemas` themselves.

rschili: This was an unsolved problem with the old SchemaContext already. If you switch changesets or import schemas, the objects you hold in memory were outdated, but you could just continue using them.
In reality that's probably not often a real issue, but I want to create a mechanism that allows us to tackle this. The nuclear option would be to make all those objects throw errors once they are outdated, but I didn't want to do that. Seems way too disruptive. So the "isOutdated" flag is a less disruptive version. Consumers can just ignore it and keep working with the context that they have...

I can probably make it so the flag gets completely managed internally, but how do we make sure consumers notice it and reload their context?
I considered making the context a wrapper that can update its internal memory model, but if we do that, all the accessor methods would have to become async because they may have to reload stuff.

One more alternative could be that whenever we notice the context is outdated, we could just quietly go and load the new info in the background and replace the internal storage immediately. This is probably the most comfortable approach but it opens the door for race conditions.

What would you prefer?

grigasp: Here's how I'd handle this:
- we track internally when schemas **possibly** change:
  - do the checksum check, if they match - ignore,
  - if checksums show schema change:
    - set the promise to `undefined`,
    - raise some consumer-facing event, saying they should request a new schema context.

core/ecschema-metadata/src/SchemaView.ts
```diff
@@ -0,0 +1,1368 @@
...
+      return { name: formatString };
+
+    const name = nameMatch[1];
+    const precMatch = /\((\d+)\)/.exec(formatString);
+    const precision = precMatch ? parseInt(precMatch[1], 10) : undefined;
+
+    const bracketRgx = /\[([^\]]+)\]/g;
+    const units: Array<readonly [string, string | undefined]> = [];
+    let m;
+    while ((m = bracketRgx.exec(formatString)) !== null) {
```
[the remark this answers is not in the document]
rschili: not a ReDoS risk - single negated char class

core/backend/src/IModelDb.ts
rschili: So this is what I ended up with:
iModelDb internally automatically invalidates the cache.
iModelConnection has a protected invalidateSchemaViewIfChanged() method, it is called automatically called whenever schemas may have changed. The frontend does not have knowledge over what changed all the time, like, when pulling it does not know if a schema changeset was applied - the backend knows that. So for now, frontend will perform this additional checksum check. But it's all automatic now.

I did not add an event for now. We may add that later. The schemaContext has an "isOutdated" flag which we set to true when we know it changed. So if you cache it somewhere, you can check this flag sometimes. If you prefer an event, I guess we can do that, but it doesn't seem strictly necessary here.

docs/learning/metadata/RuntimeSchemaContext.md
grigasp: I think we'll be able to work around this missing piece of information with the recent change to `KindOfQuantity.presentationFormats`. Resolving the comment.

grigasp (approved):
I looked over the public API surface and how `SchemaView` is maintained by and exposed through `IModelDb` and `IModelConnection` - LGTM. The only thing I don't really like is that now we'll have `schemaContext` getter + `getSchemaView()` method, but it's just my opinion, I understand the unwillingness to break the API.

I didn't look at at the tests or internals of `SchemaView` implementation. Hopefully someone from the iModels / EC team can do that.

core/backend/src/IModelDb.ts
hl662: > The schemaContext has an "isOutdated" flag which we set to true when we know it changed. 
How do we flip it back 'off', if we do? Do we run the risk of needing to perform operations to make the schema context up to date, and if the processing to make it up to date fails a step somewhere, what do we do?

 I also noticed schema context has a isOutdated flag, and the resolved promise to the schemaView also has an isOutdated check as well... do we want that isOutdated check in schema view?

core/ecschema-metadata/src/SchemaView.ts
```diff
@@ -0,0 +1,1406 @@
...
+      return result;
+    }
+
+    /** IS-A check. Returns true if this class is, or derives from, `other` (transitively, including mixins).
+     * Accepts a `SchemaView.Class` or a qualified name string ("SchemaName:ClassName").
+     */
+    public is(classOrName: Class | string): boolean {
+      const targetIdx = typeof classOrName === "string"
+        ? this._ctx.resolveClassIdx(classOrName)
+        : classOrName.idx;
```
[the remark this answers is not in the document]
rschili: Technically correct but I'm going to say this seems like a paranoid check. It can do more harm than good. If somebody caches a Class object, and the context changes and he does mismatch them, having it always evaluate to false seems like a silent corruption problem. As an alternative we could throw, but that seems too harsh for this case. I think the current behavior is a fair tradeoff.

rschili:
Hey folks, this is my first mostly AI-coded story, I followed all the best practices, had a large plan, but every time I reviewed this myself I noticed major issues. By now I have probably caught most of them. But I do not trust this code much.

I realize that isn't going to change even if I review it 10 more times 
I would like to try and get this into 5.10. It's marked beta and rather isolated, so even if there are problems lurking, it should not be disruptive.

core/ecschema-metadata/src/SchemaViewBinaryReader.ts
```diff
@@ -0,0 +1,702 @@
...
+  public readU32(): number { const v = this._view.getUint32(this._pos, true); this._pos += 4; return v; }
+  public readI32(): number { const v = this._view.getInt32(this._pos, true); this._pos += 4; return v; }
+  public readF64(): number { const v = this._view.getFloat64(this._pos, true); this._pos += 8; return v; }
+  public get pos(): number { return this._pos; }
+  public set pos(v: number) { this._pos = v; }
+
+  /** Read a string-table reference (U32 index) and return the original string. */
+  public readSRef(): string {
+    const idx = this.readU32();
+    return idx < this._strings.length ? this._strings[idx] : "";
```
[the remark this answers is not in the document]
rschili: Fixed: readSRef now throws when the index is out of range so malformed/version-mismatched blobs fail fast instead of producing a silently incomplete view.

core/ecschema-metadata/src/SchemaViewBinaryReader.ts
```diff
@@ -0,0 +1,702 @@
...
+  const schemaCatCounts: number[] = [];
+  const schemaClassStarts: number[] = [];
+  const schemaClassCounts: number[] = [];
+
+  // Deferred class data for cross-reference resolution
+  const pendingClasses: PendingClass[] = [];
+
+  // ---- PropertyDefTable ----
+  expectTag(reader, Tag.PropertyDefTable);
+  const defCount = reader.readU32();
```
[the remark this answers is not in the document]
rschili: Fixed: readSRef now throws when the index is out of range so malformed/version-mismatched blobs fail fast instead of producing a silently incomplete view.

core/backend/src/IModelDb.ts
```diff
@@ -1656,15 +1662,16 @@ export abstract class IModelDb extends IModel {
...
    *
    * @example
    * ```typescript
    * // Current usage:
    * const classMetaData: EntityMetaData | undefined = iModel.classMetaDataRegistry.find("SchemaName:ClassName");
    *
    * // Replacement:
-   * const metaData: EntityClass | undefined = imodel.schemaContext.getSchemaItemSync("SchemaName.ClassName", EntityClass);
+   * const view = await imodel.getSchemaView();
+   * const cls = view.getClass("SchemaName:ClassName");
```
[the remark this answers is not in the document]
rschili: Fixed: added a validateCount(count, minBytesPerEntry, name) helper on BinaryReader and invoked it for PropertyDefTable, SchemaTable, EnumTable, KoQTable, PropCatTable, and ClassTable. Counts that exceed what the remaining buffer could plausibly hold now throw before any allocation/loop.

core/backend/src/ClassRegistry.ts
```diff
@@ -347,8 +347,8 @@ export class ClassRegistry {
  *
  * Replacement:
  * ```ts
- * const entityMetaData: EntityClass | undefined = iModel.schemaContext.getSchemaItemSync("SchemaName.ClassName", EntityClass);
- * const relationshipMetaData: RelationshipClass | undefined = iModel.schemaContext.getSchemaItemSync("SchemaName", "ClassName", RelationshipClass);
+ * const view = await iModel.getSchemaView();
+ * const cls = view.getClass("SchemaName:ClassName");
```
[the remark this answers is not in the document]
rschili: Fixed across all five sites in this file - the migration snippets now use view.findClass(...).

core/frontend/src/IModelConnection.ts
```diff
@@ -640,6 +645,96 @@ export abstract class IModelConnection extends IModel {
...
+        return;
+      const liveToken = result.value.sha3_256 as string;
+      if (liveToken !== existing.schemaToken) {
+        this._schemasPromise = undefined;
+        existing.markOutdated();
+      }
+    } catch {
+      // If the checksum check fails, leave the cache as-is. A stale view is preferable
+      // to propagating an error from what is essentially a background housekeeping step.
+    }
```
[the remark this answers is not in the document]
rschili: Fixed: invalidateSchemaViewIfChanged now drops the cached promise and calls markOutdated() on the existing view if either awaiting the cached promise or running the checksum query throws. The next getSchemaView() reloads. Trade-off (extra reload on transient query failure) is the right one here since this is called right after operations that may have changed schemas (e.g. pullChanges).

docs/learning/ECSqlReference/MetaQueries.md
```diff
@@ -2,6 +2,8 @@
 
 The `ECDbMeta` schema allows querying schema information from an ECDb.
 
+> For an object-model alternative that loads all schemas in one shot, see [SchemaView](../metadata/SchemaView.md).
```
hl662: I dont think this clearly states the tradeoffs of the SchemaView. what about
`For a runtime-optimized object-model alternative that loads a subset of schema metadata in one shot for traversal/lookups, see SchemaView.`
Its not all schemas is it? Since it would exclude some schemas and/or its data? Like CustomAttributes?

docs/learning/ECSQLTutorial/MetaQueries.md
```diff
@@ -2,6 +2,8 @@
 
 Every iModel includes the [ECDbMeta](../ECDbMeta.ecschema.md) ECSchema. It exposes the content of all schemas that the iModel contains. You can therefore use ECSQL against that schema to query for schemas, classes, properties etc.
 
+> For bulk traversal and repeated lookups rather than targeted SQL queries, consider [SchemaView](../metadata/SchemaView.md) which loads all schemas in a single call with synchronous access.
```
hl662: Same comment as my one other in ECSqlReference - Its not all schemas

docs/learning/metadata/SchemaView.md
```diff
@@ -0,0 +1,201 @@
...
+
+## View objects and allocation
+
+`SchemaView.Schema`, `SchemaView.Class`, `SchemaView.Property`, and the other view types are lightweight wrappers holding only a context reference and an index. They do not cache data and are not identity-stable - calling `element.schema` twice returns two distinct objects that expose the same data. This means `===` comparison will fail; use `name` or `fullName` for equality checks.
+
+Calling `getProperties()` allocates a new `SchemaView.Property` wrapper for each property on every call. For hot loops, consider caching the result in a local variable. The underlying data is shared - only the thin wrapper objects are allocated.
+
+## Excluded schemas and data completeness
+
+`SchemaView` intentionally excludes a select list of schemas: Units, Formats, ECDb-internal schemas (ECDbSystem, ECDbMap, etc.), and pure custom-attribute schemas (CoreCustomAttributes, EditorCustomAttributes, etc.). The full list is defined in the C++ writer's `IsExcludedSchema()` function in `SchemaViewWriter.cpp`.
```
hl662: Should we be more explicit about which schemas are excluded? (I.e remove all the etc).
It seems SchemaViewWriter.cpp in native checks which schemas are part of the exclusion.

docs/learning/metadata/SchemaView.md
```diff
@@ -0,0 +1,201 @@
...
+
+`SchemaView` is a high-performance, read-only schema metadata cache available in both backend and frontend. It loads a lossy optimized version of schemas from an iModel in a single call and provides synchronous access to schemas, classes, properties, enumerations, kinds of quantity, and relationship constraints.
+
+It lives in `@itwin/ecschema-metadata` and is the recommended way to access schema metadata at runtime when you need fast, repeated lookups - for example in presentation rules, property grids, or data-driven UI.
+
+For the binary transport format specification, see [SchemaViewBinaryFormat.md](./SchemaViewBinaryFormat.md).
+
+## Why not SchemaContext?
+
+SchemaContext is from the full-fidelity schema toolkit. It models every detail of the EC specification - units, formats, constants, phenomena, custom attribute instances, schema references, editing, and round-trip serialization to XML/JSON. It is indispensable for schema authoring, validation, and tooling that needs the complete EC object graph.
```
hl662: The first time SchemaContext is mentioned should be a doc link

docs/learning/metadata/SchemaView.md
```diff
@@ -0,0 +1,201 @@
...
+| **Memory**            | full object graph with cross-references                   | flat arrays, string dedup, property dedup, consumes  90-95% less memory |
+| **Parse time**        | Slow (JSON parse + object construction per schema)        | very fast (binary decode into typed arrays)                             |
+| **Loading mechanism** | Synchronously, can lock the backend during large schemas. | Async via QueryReader                                                   |
+| **Custom attributes** | Always available                                          | Excluded by default                                                     |
+| **Scope**             | Full EC spec                                              | Subset (based on what runtime consumers need)                           |
+
+**Use SchemaView when** you need fast, synchronous lookups at runtime - property grids, IS-A checks, class navigation, presentation logic.
+
+**Use SchemaContext when** you need to author schemas, validate against rules, serialize to XML, or access units/formats/phenomena.
```
hl662: I'm not sure about distinct UseX / UseY bullet points in this document, not when this document gravitates so much towards SchemaView. Thinking about it now, the header for this sub section shouldn't be framed as 'Why not SchemaContext?' Imagine a user just looking for a read projection of schema data, and they go to SchemaView. The first section they see mentions a completely different API. Why not frame the section more positively instead as 'When to use SchemaView'?

docs/learning/metadata/SchemaViewBinaryFormat.md
```diff
@@ -0,0 +1,312 @@
+# Schema View Binary Format
+
+This document specifies the binary blob format produced by `SchemaViewWriter` (C++, imodel-native) and consumed by `parseSchemaViewBlob()` (TypeScript, ecschema-metadata). It is the transport format for `SchemaView` - see [SchemaView](./SchemaView.md) for the consumer API.
```
hl662: should we mention parseSchemaViewBlob or SchemaView.fromBinary? Former is internal, latter is the public facing API call

core/ecschema-metadata/eslint.config.js
```diff
@@ -15,7 +15,8 @@ module.exports = [
         "error",
         {
           "ignored-barrel-modules": [
-            "./src/ECObjects.ts"
+            "./src/ECObjects.ts",
+            "./src/SchemaViewInterfaces.ts"
```
hl662: Why is this a barrel file? Why is ECObjects a barrel file? Why are they re exporting stuff that seemingly already is being exported?

Can we remove the `export { StrengthType, StrengthDirection } from "./ECObjects";` re-export from SchemaViewInterfaces.ts
Don't need to do the same for ECObjects reexport from property types yet, unless it blocks your ability to do the above

docs/learning/ECSqlReference/Pragmas.md
```diff
@@ -13,11 +13,12 @@ PRAGMA help
...
 | experimental_features_enabled | global | enable/disable experimental features                                            |
-| validate_ecsql_writes        | global | enable/disable validation for values in an ecsql statement                      |
+| validate_ecsql_writes         | global | enable/disable validation for values in an ecsql statement                      |
 | explain_query                 | global | explain query plan                                                              |
 | help                          | global | return list of pragma supported                                                 |
 | integrity_check               | global | performs integrity checks on ECDb                                               |
 | parse_tree                    | global | parse_tree(ecsql) return parse tree of ecsql.                                   |
+| schema_view                   | global | returns all schema metadata as a binary blob                                    |
 | disqualify_type_index         | class  | set/get disqualify_type_index flag for a given ECClass                          |
```
hl662: all or a subset 👀

example-code/snippets/src/backend/SchemaViewExamples.test.ts
```diff
@@ -0,0 +1,294 @@
...
+
+  after(() => {
+    if (iModel?.isOpen)
+      iModel.close();
+  });
+
+  it("obtaining the context", async () => {
+    // __PUBLISH_EXTRACT_START__ SchemaView.obtain
+    // Obtain the schema view (async, cached after first call)
+    const ctx = await iModel.getSchemaView();
```
hl662: this is getting a view, but the variable its assigned to is short for `context`. Why not align everything in these tests to all use smt like `schemaView` or `view`? (and update test desc in line 26 too)

common/changes/@itwin/core-backend/rschili-runtime-schema_2026-05-08-08-39.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "add schemaView to iModelDb",
```
hl662:
```suggestion
      "comment": "Add SchemaView to iModelDb",
```

common/changes/@itwin/core-frontend/rschili-runtime-schema_2026-05-08-08-39.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "add schemaView to iModelConnection",
```
hl662:
```suggestion
      "comment": "Add SchemaView to iModelConnection",
```

core/ecschema-metadata/src/SchemaViewInterfaces.ts
```diff
@@ -0,0 +1,240 @@
...
+import { StrengthDirection, StrengthType } from "./ECObjects";
+export { StrengthType, StrengthDirection } from "./ECObjects";
+
+/** Internal storage for a schema. Schemas own contiguous ranges of classes, enums, KoQs, and categories.
+ * @internal
+ */
+export interface SchemaData {
+  /** Row ID from ec_Schema - use with ECDbMeta queries, e.g. `SELECT * FROM meta.ECSchemaDef WHERE ECInstanceId = ?`. */
+  readonly ecInstanceId: number;
+  readonly nameSid: number;
```
hl662: Sid is too cryptic here, especially alongside ecInstanceId and *Idx. Its not immediately clear to me what these are. These are used for the deduped SchemaView.strings map? gotta be a better way to convey this

core/ecschema-metadata/src/SchemaView.ts
```diff
@@ -0,0 +1,1412 @@
...
+ * All data is stored in flat arrays. View objects (`SchemaView.Schema`, `SchemaView.Class`, etc.) are
+ * stateless wrappers that hold a reference to this view plus an index. They allocate nothing
+ * and cache nothing.
+ *
+ * The view is immutable after construction. Build it via `SchemaViewBuilder` or parse
+ * from a binary blob via `fromBinary`.
+ * @beta
+ */
+export class SchemaView {
+  /** @internal */ public readonly strings: readonly string[];
```
hl662: Have you considered using a module level private WeakMap variable, to remove the need for all this internal public readonly vars? 
```ts
// Not exported — module-private
const _d = new WeakMap<SchemaView, SchemaViewData>();

export class SchemaView {
  // NO public data fields
  constructor(data: SchemaViewData, schemaToken?: string) {
    _d.set(this, data);
    this._schemaToken = schemaToken ?? "";
  }
}

export namespace SchemaView {
  export class Schema {
    constructor(private readonly _ctx: SchemaView, public readonly idx: number) {}
    private get _view() { return _d.get(this._ctx)!; }   // same file = works
    get name() { return this._view.strings[this._view.schemas[this.idx].nameSid]; }
  }
}
```

transitiveBaseCache and derivedClassMap should not be part of the WeakMap though I don't think, since they are mutable

core/ecschema-metadata/src/SchemaViewBinaryReader.ts
```diff
@@ -0,0 +1,720 @@
...
+          builder.addEnumerator({
+            nameSid: builder.internString(name),
+            labelSid: builder.internString(v.DisplayLabel),
+            descriptionSid: builder.internString(v.Description),
+            value,
+          });
+          enumeratorCount++;
+        }
+      } catch (e) {
+        Logger.logWarning("ecschema-metadata.SchemaView", `Malformed EnumValues JSON for enumeration "${eName}": ${e}`);
```
hl662: Interesting, malformed EnumValues just get ignored? So an Enumeration could have no enums if all of the EnumValues are malformed?

core/frontend/src/IModelConnection.ts
```diff
@@ -640,6 +645,112 @@ export abstract class IModelConnection extends IModel {
...
+        this._schemasPromise = undefined;
+        existing.markOutdated();
+      }
+    } catch {
+      // The checksum check is called right after operations that may have changed schemas
+      // (e.g., pullChanges). If we cannot verify the cached view is still current, drop it
+      // rather than risk returning stale metadata indefinitely. The next getSchemaView() call
+      // will reload. We also mark the existing view outdated so any retained references can
+      // observe the invalidation.
+      if (this._schemasPromise === existingPromise) {
```
hl662, line 731: Can you keep the same identity guard on the success path too?

Right now `invalidateSchemaViewIfChanged()` captures `existingPromise` at lines 702-703, but when the schema token differs it unconditionally does `_schemasPromise = undefined` at lines 720-721. That means an interleaving like:

- this call starts with `existingPromise = P1`
- another path reloads and `getSchemaView()` installs a newer promise `P2`
- this call resumes, sees the token changed for `P1`, and clears `_schemasPromise` anyway

would clobber the newer cached promise. I don't think that returns stale/corrupt data directly, but it does lose the single-flight cache and can trigger duplicate `PRAGMA schema_view(...)` reloads.

Might be worth mirroring the guard you already have in the `catch` path / `getSchemaView()`?

```ts
if (liveToken !== existing.schemaToken) {
  if (this._schemasPromise === existingPromise)
    this._schemasPromise = undefined;
  existing.markOutdated();
}
```

(This one came from Nambot)

docs/learning/metadata/SchemaView.md
```diff
@@ -0,0 +1,201 @@
...
+
+## How does it relate to ECDbMeta ECSQL queries?
```
hl662: Instead of 'relate' maybe use 'differ'? I wonder if this either can't be moved further down the doc, or turn this from a ## section to a expandable block that acts as a note... The explanation on ECDbMeta ECSQL queries should be second to explanations of SchemaView

docs/learning/metadata/SchemaView.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Resolving format and unit names
+
+Presentation format names use schema aliases (e.g. `"f"` for `Formats`, `"u"` for `Units`). The Units and Formats schemas are excluded from `SchemaView` because they will be accessed through a separate dedicated API in the future.
+
+If you need the actual format definitions or unit details today, you can resolve the alias-qualified names via **ecschema-metadata** or **ECSQL**:
+
+#### Via ecschema-metadata (SchemaContext)
+
+Split the alias-qualified name at `:` - the left part is the schema alias (e.g. `"f"` -> `"Formats"`, `"u"` -> `"Units"`), the right part is the item name. Look up the schema item via `IModelDb.schemaContext`:
```
hl662, line 125: I would prefer all the code blocks in the docs to go through example-code-snippets. I think ecschema-metadata, inside tests/, has example-code folder? see if you can reuse the extracted snippets in there or make new ones?

hl662 (changes requested):
Documentation is split between places that says 'all schema metadata' is available in the SchemaView, when it seems like there are quite a few schemas, and/or their data, that is being excluded from this materialized view. That needs to all be made consistent

docs/learning/ECSqlReference/MetaQueries.md
rschili: Fixed - reworded to closely match your suggestion: "runtime-optimized object-model alternative that loads a curated subset of schema metadata in one shot for traversal and repeated lookups". The "all" framing is gone everywhere in the docs.

docs/learning/ECSQLTutorial/MetaQueries.md
rschili: Fixed - same wording adjustment here, no longer claims all schemas.

docs/learning/metadata/SchemaView.md
rschili: Done. The doc now has a dedicated "What is excluded" section that enumerates all four categories with concrete names instead of an "etc.":

- Custom attribute instances on schemas/classes/properties/relationship constraints (selected CAs - Views, Mixin, Hidden - are promoted to first-class concepts).
- Units, Formats, Phenomena, UnitSystems, Constants.
- ECDb-internal: ECDbSystem, ECDbMap, ECDbFileInfo, ECDbSchemaPolicies (note ECDbMeta is *not* excluded).
- Pure CA schemas: CoreCustomAttributes, BisCustomAttributes, EditorCustomAttributes, ECv3ConversionAttributes, SchemaLocalizationCustomAttributes, SchemaUpgradeCustomAttributes.

Points to `IsExcludedSchema()` in `SchemaViewWriter.cpp` as the authoritative list. Also added a short rationale: this is a deliberate size-vs-usefulness trade-off, the omitted data isn't "never needed" - it's needed rarely enough that pulling it on demand via SchemaContext or ECDbMeta beats carrying it in every runtime cache.

docs/learning/metadata/SchemaView.md
rschili: Done - first mention is now `[SchemaContext]($ecschema-metadata)`.

docs/learning/metadata/SchemaView.md
rschili: Reframed. The section is now `## When to use SchemaView` and leads with the runtime use cases (property grids, IS-A checks, presentation rules, property iteration). SchemaContext appears only as the secondary option for the cases SchemaView doesn't cover (authoring, editing, validating, serializing, or accessing excluded data). The comparison table follows with SchemaView in the left column. Visitors who land here looking for SchemaView see SchemaView first.

docs/learning/metadata/SchemaViewBinaryFormat.md
rschili: Switched to `SchemaView.fromBinary` (the public entry point). Good catch.

docs/learning/ECSqlReference/Pragmas.md
rschili: Changed to "a curated subset of schema metadata" both in the pragma table row and in the section description, with a link to the SchemaView "What is included" section so readers can see exactly what they get.

docs/learning/metadata/SchemaView.md
rschili: Did both: "relate" is now "differ", and the whole ECDbMeta comparison is now a collapsed `<details>` block at the end of the doc, after the SchemaView content (Dangling references section). So a reader landing on this page sees SchemaView material first and only opens the ECDbMeta comparison if they want it.

docs/learning/metadata/SchemaView.md
rschili: Most code blocks in this doc already use `[[include:SnippetName]]` and are extracted from `example-code/snippets/src/backend/SchemaViewExamples.test.ts`. The exceptions are the two inline ECSQL examples in "Resolving format and unit names" - I left those inline on purpose because they're illustrative ECSQL strings meant to be copied/adapted into queries, not runtime APIs whose compilation we'd want to enforce.

Happy to extract them for consistency if you feel strongly - the test harness would just hold them as constants.

core/ecschema-metadata/src/SchemaViewBinaryReader.ts
rschili: This is intentional. The blob is created by ourselves, we can assume that malformed entries are extremely rare. Skipped enum values are undesirable, however, having the whole parse fail would be catastrophc since nobody could access any schema information in that case. I think a log warning is the reasonable choice here.

common/changes/@itwin/core-backend/rschili-runtime-schema_2026-05-08-08-39.json
rschili: Applied your suggestion verbatim.

common/changes/@itwin/core-frontend/rschili-runtime-schema_2026-05-08-08-39.json
rschili: Applied your suggestion verbatim.

example-code/snippets/src/backend/SchemaViewExamples.test.ts
rschili: Renamed throughout the snippets file: `ctx` -> `view`, plus the test description on line 26 ("obtaining the context" -> "obtaining the schema view"). The one exception is the `views` test, where the inner `for (const view of schema.getViews())` would shadow the outer; I named the outer `schemaView` there to keep both readable. Updated the matching prose in [SchemaView.md](https://github.com/iTwin/itwinjs-core/blob/main/docs/learning/metadata/SchemaView.md) (the section is now "Obtaining the schema view") so the doc and the snippet agree.

core/frontend/src/IModelConnection.ts
rschili: Good catch (and please thank Nambot for me). Applied the same identity guard you proposed:

```ts
if (liveToken !== existing.schemaToken) {
  if (this._schemasPromise === existingPromise)
    this._schemasPromise = undefined;
  existing.markOutdated();
}
```

Now both the success-with-changed-token path and the catch path use `this._schemasPromise === existingPromise` before clearing, so a newer cached promise installed by an interleaving `getSchemaView()` call won't be clobbered.

core/ecschema-metadata/src/SchemaViewInterfaces.ts
rschili: Renamed across all three files (SchemaViewInterfaces.ts, SchemaView.ts, SchemaViewBinaryReader.ts): `*Sid` -> `*StringIdx`. So the SchemaData fields are now `nameStringIdx`, `aliasStringIdx`, `labelStringIdx`, `descriptionStringIdx`, etc.

Picked `*StringIdx` to be parallel with the existing `*Idx` suffix (`schemaIdx`, `classIdx`, `baseClassIdx`, etc.) used for typed-array indices, and clearly distinct from `ecInstanceId` which is a DB row ID. The string table is just another typed array, so `StringIdx` reads consistently with the rest. Open to `StringId` or `StrIdx` if you prefer those.

core/ecschema-metadata/eslint.config.js
rschili: Dropped the `export { StrengthType, StrengthDirection } from "./ECObjects"` re-export from SchemaViewInterfaces.ts. The plain `import` stays because those types are still used by the interfaces in the file (`SchemaData.strength`, `SchemaData.strengthDirection`, `navDirection`). Since SchemaViewInterfaces.ts no longer re-exports anything, I also removed it from the `ignored-barrel-modules` allowlist in eslint.config.js - the entry was only needed because of that re-export.

Left the ECObjects re-export from property types alone for now, per your suggestion.

core/ecschema-metadata/src/SchemaView.ts
rschili: Did the dig you nudged us toward, and you were right that the encapsulation could be tighter. Went with the Symbol-keyed-slot pattern rather than a WeakMap - same idea but no WeakMap.get cost on every accessor.

**Why not WeakMap:** every flyweight accessor would do 2x `_d.get(this._ctx)` calls. V8's WeakMap.get is materially slower than a hidden-class slot load (~5x in microbenchmarks). A property-grid render iterating 50 properties × ~6 field accesses each ends up doing 500-1000 WeakMap lookups where today (and now) it does direct slot reads. That cuts against the "optimized for hot paths" framing of the API.

**What I did instead:** a module-local `Symbol("SchemaView.storage")` (not exported, not registered) used as the key for a single `[_storage]: SchemaViewStorage` slot on `SchemaView`. This is the same pattern `IModelDb` uses for `_nativeDb` (see `core-backend/src/internal/Symbols.ts`), but a step stronger - because the symbol is module-local rather than `Symbol.for(...)`, no caller outside the file can reconstruct it. The 19 `@internal public readonly` data fields and 2 mutable cache fields collapse into a single slot:

```ts
const _storage = Symbol("SchemaView.storage");

interface SchemaViewStorage extends SchemaViewData {
  readonly transitiveBaseCache: Map<number, ReadonlySet<number>>;
  derivedClassMap: ReadonlyMap<number, readonly number[]> | undefined;
}

export class SchemaView {
  public readonly [_storage]: SchemaViewStorage;
  constructor(data: SchemaViewData, schemaToken?: string) {
    this[_storage] = { ...data, transitiveBaseCache: new Map(), derivedClassMap: undefined };
    ...
  }
}

export namespace SchemaView {
  export class Schema {
    constructor(private readonly _ctx: SchemaView, public readonly idx: number) {}
    public get name(): string { return this._ctx[_storage].strings[this._data.nameStringIdx]; }
  }
}
```

V8 stores Symbol-keyed properties in the same hidden-class slots as string-keyed ones, so monomorphic call sites are inline-cached as direct slot loads - same speed as the previous `this._ctx.strings`. The transitiveBaseCache and derivedClassMap live in the same storage bag (the slot itself is readonly; the cache fields inside it can mutate, same as your WeakMap suggestion would have allowed).

Result: the API surface for `SchemaView` shrinks from 21 `@internal public readonly` fields to a single `@internal` symbol-keyed slot. Consumer code can no longer reach into the internals without obtaining the module-local Symbol value, which isn't exported. The flyweight accessors in the namespace still get direct-slot perf.

Thanks for the push - this is a real improvement.

khanaffan (approved):

docs/learning/metadata/SchemaView.md
```diff
@@ -1,50 +1,66 @@
...
-| **Parse time**        | Slow (JSON parse + object construction per schema)        | very fast (binary decode into typed arrays)                             |
-| **Loading mechanism** | Synchronously, can lock the backend during large schemas. | Async via QueryReader                                                   |
-| **Custom attributes** | Always available                                          | Excluded by default                                                     |
-| **Scope**             | Full EC spec                                              | Subset (based on what runtime consumers need)                           |
+Reach for the full-fidelity [SchemaContext]($ecschema-metadata) instead when you are *working with schemas as schemas*: authoring, editing, validating, serializing to XML/JSON, or accessing data that `SchemaView` deliberately omits (see [What is included](#what-is-included)). `SchemaContext` is the more expensive option - one async RPC per schema, full object graph with cross-references, slow to load - so use it when its completeness is what you actually need.
```
hl662: > working with schemas as schemas

Eh... theres gotta be a better way to phrase this

docs/learning/metadata/SchemaView.md
```diff
@@ -1,50 +1,66 @@
...
+- **Units, Formats, Phenomena, UnitSystems, Constants** - the entire `Units` and `Formats` schemas are dropped. `KindOfQuantity` carries only the persistence-unit name and presentation-format strings; consumers resolve names against the dedicated units/formats APIs (today: `SchemaContext` or ECSQL - see [Resolving format and unit names](#resolving-format-and-unit-names)).
+- **ECDb-internal schemas** - `ECDbSystem`, `ECDbMap`, `ECDbFileInfo`, `ECDbSchemaPolicies`. These describe storage-layer mapping and are not relevant to runtime consumers. Note that `ECDbMeta` is *not* excluded - it remains queryable via ECSQL.
+- **Pure custom-attribute schemas** - `CoreCustomAttributes`, `BisCustomAttributes`, `EditorCustomAttributes`, `ECv3ConversionAttributes`, `SchemaLocalizationCustomAttributes`, `SchemaUpgradeCustomAttributes`. These contain only `CustomAttribute` and `Struct` definitions used for decoration; since CA instances are not transported, the definitions add little value.
 
-## Obtaining the context
+The authoritative list lives in `IsExcludedSchema()` in `SchemaViewWriter.cpp` (imodel-native).
```
hl662: I would prefer a hyperlink to the code block (once the PR in native is in)

docs/learning/metadata/SchemaView.md
rschili: good point, I rephrased it, please have a look

docs/learning/metadata/SchemaView.md
rschili: I added a hyperlink to the location where it will eventually be

core/ecschema-metadata/src/SchemaView.ts
```diff
@@ -0,0 +1,1380 @@
...
+     * //   { name: "f:DefaultRealU", precision: 4, unitAndLabels: [["u:M_PER_SEC_SQ", undefined]] },
+     * //   { name: "f:DefaultRealU", precision: 4, unitAndLabels: [["u:CM_PER_SEC_SQ", undefined]] },
+     * // ]
+     * ```
+     */
+    public get presentationFormats(): readonly PresentationFormat[] {
+      const raw = this.presentationFormatsRaw;
+      if (raw === "")
+        return [];
+      const formats: string[] = JSON.parse(raw);
```
RohitPtnkr1996, line 1088: With ref from Nam's comment https://github.com/iTwin/itwinjs-core/pull/9151#discussion_r3262003635

This presentationFormats getter also tries to parse a raw string which might be malformed and throw ?

rschili: I just checked, the concern is valid, but this scenario is effectively unreachable in normal flow.
The values in those tables are written by rapidJson and will always contain valid values unless there is a bug in the writer or external corruption, in both cases, this throwing seems like the correct behavior. For older FUS strings (pre 2018 iModels) this will not throw and instead just return nothing, which is documented on the .raw property.

RohitPtnkr1996 (approved):

hl662 (approved):

core/backend/src/Relationship.ts
```diff
@@ -50,7 +50,9 @@ export class Relationship extends Entity {
     return val;
   }
 
-  /** Query metadata for this relationship class from the iModel's schema. Returns cached metadata if available.*/
+  /** Query metadata for this relationship class from the iModel's schema. Returns cached metadata if available.
+   * @deprecated in 5.10.0. Will not be removed until after 2027-06-01. Obtain a `SchemaView` via `iModel.getSchemaView()` and call `view.findClass(relationship.classFullName)` to retrieve a `SchemaView.Class`. The returned type differs from `RelationshipClass` and offers a synchronous, snapshot-style API.
```
aruniverse: i believe you just need to mark it with `@deprecated in 5.X.0...` 
The date gets auto added by the linter

core/backend/src/IModelDb.ts
```diff
@@ -1656,15 +1662,16 @@ export abstract class IModelDb extends IModel {
 
   /** The registry of entity metadata for this iModel.
    * @internal
-   * @deprecated in 5.0 - will not be removed until after 2026-06-13. Please use `schemaContext` from the `iModel` instead.
+   * @deprecated in 5.0 - will not be removed until after 2026-06-13. Use `getSchemaView()` from the `iModel` instead.
```
grigasp, line 1665: What's the point to deprecate `@internal` api? Why not replace its internal usages with recommended path and just remove it straight away?

core/backend/src/Entity.ts
```diff
@@ -271,7 +277,9 @@ export class Entity {
     }
   }
 
-  /** @internal */
+  /** @internal
+   * @deprecated in 5.10.0 - will not be removed until after 2027-06-01. Internal callers should obtain a `SchemaView` via `iModel.getSchemaView()` (or retain a pre-loaded one) and call `view.findClass(...)` instead. Note that `getSchemaView()` is async.
```
grigasp: Same question - what's the point in deprecating `@internal` API?

common/changes/@itwin/core-frontend/rschili-runtime-schema_2026-05-08-08-39.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-frontend",
+      "comment": "Add SchemaView to iModelConnection",
```
grigasp:
```suggestion
      "comment": "Added `IModelConnection.getSchemaView()` function, which provides access to iModel's `SchemaView` - a lightweight, read-only, synchronous API for navigating schema metadata - classes, properties, relationships, enumerations, etc.",
```

common/changes/@itwin/core-backend/rschili-runtime-schema_2026-05-08-08-39.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-backend",
+      "comment": "Add SchemaView to iModelDb",
```
grigasp:
```suggestion
      "comment": "Added `IModelDb.getSchemaView()` function, which provides access to iModel's `SchemaView` - a lightweight, read-only, synchronous API for navigating schema metadata - classes, properties, relationships, enumerations, etc.",
```

core/backend/src/IModelDb.ts
rschili: even though it's internal, it is still being used by ClassRegistry. We are planning to eventually replace it, but until that happens, the deprecation is to prevent others from using the method.

core/backend/src/Entity.ts
rschili: There do not seem to be any existing callers for this. But the async counterpart of this has a lot of internal callers and it is public preview. I want to keep both around for the same timeframe, that's why this one is marked deprecated.

core/backend/src/IModelDb.ts
grigasp: > We are planning to eventually replace it

Do you have a work item for that? Linking it in the deprecation message would help with understanding why an internal API needs to be deprecated and tracking when it can be removed.

core/backend/src/Relationship.ts
rschili: ah, I wasn't aware. I changed all messages accordingly.

core/backend/src/Entity.ts
grigasp: It's internal, so I'd just remove it if there're no usages. I'm not an owner of this package, though, so your call. Feel free to resolve this comment.

rschili: Paul added this a few months ago. If it were by my group I'd drop it, but in this case I don't want to take any chances, I'll leave it until the timer runs out.

markschlosseratbentley (approved):

diegoalexdiaz (approved):

aruniverse (approved):
approving addon changes

hl662 (approved):
Re-approving with a confirmation that the deprecation comments will append deprecation dates at the end. Its a regex match, if theres no date mentioned in the comment, the script will run and based on the tag, it will decide the date.

The new deprecations in this PR, all but one will have a removal date of 3 months.
The one that has a one year deprecation date is Relationship.getMetadata.

A little refresher on what the date being added would be, depending on the tag:

Image

(Forgot myself that only preview tags would have a 3 month deprecation date, rest would have 1 year)

PR #9160, opened 2026-04-01
Rework api categories format

DanRod1999:
Trying to make it clearer that deprecated and preview tags are to be used on top of any of the other categories.

Also Preview does not link to anything and looks weird. Not sure if we should link to something or just remove the square brackets so it looks more normal

[Current docs](https://www.itwinjs.org/learning/api-support-policies/)


DanRod1999:
Formatted example:
## API categories

iTwin.js uses [API Extractor](https://api-extractor.com/) to help manage its APIs. Each API exported by a package is marked with one of four ["release tags"](https://api-extractor.com/pages/tsdoc/doc_comment_syntax/) categorizing its level of maturity.

- ["internal"](https://api-extractor.com/pages/tsdoc/tag_internal/) indicates an API that is intended strictly to be used inside the iTwin.js libraries - never by users.
- ["alpha"](https://api-extractor.com/pages/tsdoc/tag_alpha/) indicates a highly-experimental API likely to undergo significant, frequent change and/or abrupt removal. These are rare in iTwin.js as we are typically interested in early feedback.
- ["beta"](https://api-extractor.com/pages/tsdoc/tag_beta/) indicates an API currently under development. Users are encouraged to experiment with the API and provide feedback, with the understanding that breaking changes may occur at any time.
- ["public"](https://api-extractor.com/pages/tsdoc/tag_public/) indicates a stable API suitable for use in production, to which breaking changes will only be introduced under the strict policies described in this document.

In addition to any one of the above release tags, an API may also be tagged as:

- ["deprecated"](https://api-extractor.com/pages/tsdoc/tag_deprecated/), indicating that it should no longer be used and may be removed in a future release. A deprecated tag is typically accompanied by information about the package version in which it became deprecated and guidance for adjusting existing usage of the API, e.g., what API to use instead. A "public" deprecated API will only be removed according to the [deprecation policy](#api-deprecation-policy).
- ["preview"] indicates an API that is stable within the current major version, but may be changed or removed in the next major version. It is typically used in conjunction with the public tag to signal that while the API is production-ready for the time being, it is not guaranteed to remain stable across major version updates.

Only "public" and "beta" APIs are included in the [published documentation](https://www.itwinjs.org/reference/). An API typically starts out as "beta". It may evolve rapidly in response to feedback before stabilizing and being promoted to "public". On occasion, a "beta" API may be abandoned - not all experiments succeed.

The "public API" of a package comprises the set of all APIs it contains that are marked with the "public" release tag. The package's public API enjoys stability guarantees provided by the [package versioning policy](#package-versioning-policy).

Contributors to iTwin.js should follow the [provided guidelines](./guidelines/release-tags-guidelines.md) when applying release tags to their APIs.

docs/learning/api-support-policies.md
```diff
@@ -45,9 +45,10 @@ iTwin.js uses [API Extractor](https://api-extractor.com/) to help manage its API
...
 
-Additionally, ["preview"] indicates an API that is stable within the current major version, but may be changed or removed in the next major version. It is typically used in conjunction with the public tag to signal that while the API is production-ready for the time being, it is not guaranteed to remain stable across major version updates.
+- ["deprecated"](https://api-extractor.com/pages/tsdoc/tag_deprecated/), indicating that it should no longer be used and may be removed in a future release. A deprecated tag is typically accompanied by information about the package version in which it became deprecated and guidance for adjusting existing usage of the API, e.g., what API to use instead. A "public" deprecated API will only be removed according to the [deprecation policy](#api-deprecation-policy).
+- ["preview"] indicates an API that is stable within the current major version, but may be changed or removed in the next major version. It is typically used in conjunction with the public tag to signal that while the API is production-ready for the time being, it is not guaranteed to remain stable across major version updates.
```
DanRod1999: square brackets that lead to nothing that I was referring to ln51

MichaelSwigerAtBentley: Maybe link to an example, or link ot the api deprecation policy, and add a section to it that explains how preview tags might mean something is deprecated in the future?

docs/learning/api-support-policies.md
```diff
@@ -45,9 +45,10 @@ iTwin.js uses [API Extractor](https://api-extractor.com/) to help manage its API
...
 
 Only "public" and "beta" APIs are included in the [published documentation](https://www.itwinjs.org/reference/). An API typically starts out as "beta". It may evolve rapidly in response to feedback before stabilizing and being promoted to "public". On occasion, a "beta" API may be abandoned - not all experiments succeed.
```
MichaelSwigerAtBentley, line 53: For these notes at the bottom, maybe move them up near the section on public, beta, ect and use githubs
`> [!NOTE]` or `> [!TIP]` alerts?

DanRod1999: do note and tip alerts work in our docs pages?

MichaelSwigerAtBentley: [Yup](https://github.com/iTwin/itwinjs-backlog/blob/main/docs/itwinjs/iTwin.js%20Release%20Process.md)

You can see one at the top of that page

MichaelSwigerAtBentley: > [!NOTE]
> They even work here

PR #9165, opened 2026-04-02
Schema Localization Support

naveedkhan8067:
Requested in: https://github.com/iTwin/itwinjs-backlog/issues/1864

The localization json will be part of the respective schema package. After this PR, respective process will be validated and updated on that side.


core/ecschema-metadata/src/test/Localization/SchemaLocalization.test.ts
```diff
@@ -0,0 +1,800 @@
...
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import * as fs from "fs";
+import * as path from "path";
+import { SchemaContext } from "../../Context";
+import { Schema } from "../../Metadata/Schema";
+import { beforeAll, describe, expect, it } from "vitest";
+import { SchemaLocalization } from "../../Localization/SchemaLocalization";
+import { SchemaLocalizationJson } from "../../Localization/LocalizationTypes";
```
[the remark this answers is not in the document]
naveedkhan8067: Already fixed

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,321 @@
...
+    this._cache.set(cacheKey, localization || null);
+    return localization || null;
+  }
+
+  /**
+   * Get localized label for a schema.
+   * Fallback: localized label → original label → schema name
+   */
+  public async getSchemaLabel(schema: Schema): Promise<string> {
+    const localization = await this.getSchemaLocalizationJson(schema.name);
```
hl662, line 98: This is where getSchemaLabel does localization?.label → falls back to schema.label || schema.name, but never calls getBaseLocalizationJson().

Compare to getSchemaItemLabel at line 130 which does try getBaseLocalizationJson. The comment on line 97 even documents the shorter fallback chain: "localized label → original label → schema name" (missing base locale step).

Same issue on line 112 for getSchemaDescription, the fallback chain could be consistent?

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+  public async getLocalization(schemaName: string, locale: string): Promise<SchemaLocalizationJson | undefined> {
+    const cacheKey = `${schemaName}:${locale}`;
+
+    if (this._localizationCache.has(cacheKey)) {
+      return this._localizationCache.get(cacheKey);
+    }
+
+    let localizationData = await this._loader(schemaName, locale);
+
+    // Try fallback to language without region (e.g., "es-CO" → "es")
```
hl662, line 48: The locale fallback logic is split across two classes right now:
  - LocalizationProvider.getLocalization() does its own es-CO → es fallback when the specific file is missing
  - SchemaLocalization.getBaseLocalizationJson() independently calls provider.getLocalization(schema, "es") as a secondary lookup
 
If the es-CO file is missing entirely, the provider already falls back to es and caches the result under "schema:es-CO". Then SchemaLocalization fetches 'es' again separately — seems redundant . And when SchemaLocalization.locale is changed via the setter, only its own _cache is cleared, not  LocalizationProvider._localizationCache.
 
 I'd suggest picking one layer to own fallback — probably the provider, since it's closest to the data source?

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,321 @@
...
+   * @internal
+   */
+  private findItemLocalization(localization: SchemaLocalizationJson, item: SchemaItem): LocalizedText | undefined {
+    const itemName = item.name;
+
+    if (item.schemaItemType === "EntityClass" ||
+      item.schemaItemType === "StructClass" ||
+      item.schemaItemType === "CustomAttributeClass" ||
+      item.schemaItemType === "RelationshipClass" ||
+      item.schemaItemType === "Mixin") {
```
hl662, line 276: findItemLocalization() handles 13 of 14 SchemaItemType values but skips InvertedUnit. If someone localizes an inverted unit, it silently falls back to English with no warning. Should we fix that?
 
Also, the comparisons use raw string literals ("EntityClass", "StructClass", etc.) instead of the SchemaItemType enum values used everywhere else in the codebase. If enum string values ever change, this breaks silently.
 
 I'd recommend switching to an exhaustive switch on SchemaItemType with a default case that at least logs a warning — similar to how SchemaPartVisitorDelegate handles this. That way new item types can't slip through unnoticed

core/ecschema-metadata/src/Localization/LocalizationTypes.ts
```diff
@@ -0,0 +1,68 @@
...
+ * JSON structure for schema localization file.
+ * @beta
+ */
+export interface SchemaLocalizationJson {
+  /** Schema version identifier */
+  $schema?: string;
+  /** Name of the schema, this localization applies to */
+  name: string;
+  /** Version of the schema */
+  version?: string;
```
hl662, line 43: The JSON type includes version?: string, but the runtime never reads or validates it. This creates a misleading middle state — consumers might think versioned localization is supported, author files with version fields, and discover they're ignored.
 
The related GH issue agreed on "version-independent within major generation," but I don't think this is properly enforced?
 
 I think either: (a) remove version from the type until there's a real compatibility policy, or (b) validate it and at least warn when the major generation doesn't match

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!localizationData && locale.includes("-")) {
+      const baseLocale = locale.split("-")[0];
+      localizationData = await this._loader(schemaName, baseLocale);
+    }
+
+    if (localizationData) {
+      if (!localizationData.name || !localizationData.locale) {
```
hl662, line 55: LocalizationProvider checks that name and locale exist in the loaded JSON, but doesn't verify json.name === requestedSchemaName or json.locale === requestedLocale. A wrong file could be loaded and silently used. Localization files are typo-prone — misspelling a property key
 like "Hieght" just silently falls back to English. Some basic identity checks would help catch mistakes early...? 

Also, I went through copilot to check if there's any existing localization APIs and seems like presentation-common has LocalizationHelper? Should we check how that API can use this?

naveedkhan8067: I have fixed it. It should validate for requested schema and locale.

core/ecschema-metadata/src/Localization/LocalizationTypes.ts
naveedkhan8067: It now validates that the major schema version must be similar.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: The InvertedUnit is now covered and updated string literal to the SchemaItemType.

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
naveedkhan8067: The redundant behaviour should have fixed.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: It now covers the base localization.

rschili:
This deserves a thorough review. Sorry I haven't gotten around to it yet, but will try and take a look once my SchemaView work is done.

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+        throw new Error(`Localization JSON mismatch for ${schemaName}:${locale} - expected schema name "${schemaName}" but got "${localizationData.name}"`);
+      }
+
+      const expectedLocale = locale.includes("-") && !localizationData.locale.includes("-") ? locale.split("-")[0] : locale;
+      if (localizationData.locale !== expectedLocale) {
+        throw new Error(`Localization JSON mismatch for ${schemaName}:${locale} - expected locale "${expectedLocale}" but got "${localizationData.locale}"`);
+      }
+
+      const key = `${schemaName}:${localizationData.locale}`;
+      this._localizationCache.set(key, localizationData);
```
RohitPtnkr1996: We are caching the localization JSON under the fallback locale name, but the SchemaLocalization object still has the original locale set on it.
Won't this cause a cache miss every time a localized value is requested and force an unnecessary reload ?

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
RohitPtnkr1996: I think Nam makes a good point about the duplicate fallback and caching logic across the two classes.
Maybe you can combine them into a single private helper.

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
RohitPtnkr1996: Along with the redundant cache loading test, you can also add a test to ensure that setting a new locale cleares the cache and subsequent locale requests trigger a reload.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    return schema.description;
+  }
+
+  /**
+   * Get localized label for a schema item
+   * Fallback: localized label (specific locale) → localized label (base locale) → original label → item name
+   */
+  public async getSchemaItemLabel(item: SchemaItem): Promise<string> {
```
RohitPtnkr1996: I'm wondering if we should also have a combined call to get the label and description together, to avoid separate calls ?
This suggestion might be overkill, I admit I'm not aware of how the localization calls are made and if there is ever a need to get both at the same time.
But since you already have a LocalizedText defined, I thought why not ?

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,362 @@
...
+
+    if (this._cache.has(cacheKey)) {
+      return this._cache.get(cacheKey) || null;
+    }
+
+    const localization = await this._provider.getLocalization(schema.name, this._locale);
+    if (localization && localization.version) {
+      const localizationMajor = this.getMajorVersion(localization.version);
+      if (schema.schemaKey.readVersion !== localizationMajor) {
+        throw new Error(`Localization version mismatch for schema "${schema.name}". Schema major version is ${schema.schemaKey.readVersion.toString()}, but localization is for major version ${localizationMajor}.`);
```
RohitPtnkr1996: `getMajorVersion` returns `number | undefined`.
If it returns an undefined, it will make this error message a bit weird.
Maybe add a separate check for an undefined major version for better error reporting.

RohitPtnkr1996: We should add a cache entry with a null for this localization json before throwing to avoid possible future failures.

naveedkhan8067: Fixed

naveedkhan8067: Covered as part of another comment.

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+
+      return localizationData;
+    }
+
+    return undefined;
```
rschili: when neither the regional nor the base file exists, we return undefined without storing anything. Every subsequent call hits the loader twice (regional + base) for the same missing schema. For an iModel with dozens of schemas and no localization, that's hundreds of loader calls. SchemaLocalization's own _cache does cache the null, which partially mitigates - but Schema localization is per-instance, while the provider can be shared.

Suggest caching the negative result in the provider as well.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
rschili: I don't think we want to throw errors in these cases.
A major version mismatch is a real problem - I believe in that case our options are showing the default english labels or trying to apply the localization anyways.
Throwing an error could disrupt the caller so we don't show anything, that would be the worst possible outcome.
I believe logging a warning and returning null might be the right call here.

core/ecschema-metadata/src/Localization/LocalizationTypes.ts
rschili: Like I commented in the other file, I think throwing isn't a good use of this. The version is optional, specifying a version should probably raise a log warning but not lead to a hard throw.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,362 @@
...
+    return localization || null;
+  }
+
+  /**
+   * Load the base locale localization if the current locale has a region (e.g., "es" from "es-CO").
+   * @param schema The schema to load localization for
+   * @returns The base localization JSON, or null if not found or not applicable
+   * @internal
+   */
+  private async getBaseLocalizationJson(schema: Schema): Promise<SchemaLocalizationJson | null> {
```
karolis-zukauskas: Why are we using `null` instead of `undefined`?

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    if (this._localizationCache.has(cacheKey)) {
+      return this._localizationCache.get(cacheKey);
+    }
+
+    let localizationData = await this._loader(schemaName, locale);
+
+    // Try fallback to language without region (e.g., "es-CO" → "es")
+    if (!localizationData && locale.includes("-")) {
+      const baseLocale = locale.split("-")[0];
```
laurynas111: Would recommend documenting in some md file the format of JSON and its' name that the code expects

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    return item.description;
+  }
+
+  /**
+   * Get localized label for a property.
+   * Fallback: localized label (specific locale) → localized label (base locale) → original label → property name
+   */
+  public async getPropertyLabel(ecClass: ECClass, property: Property): Promise<string> {
```
karolis-zukauskas: I imagine this or similar function will be called a lot - having it be `async` has two issues:
1) Overhead (while in most cases it will hit caches and not perform any "real" async work)
2) Harder to consume from places that are not async friendly ..

Could we load all needed locale files in an async "constructor function" (have actual constructor private) and then have all of these methods sync? Or do you think this is a no issue?

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,362 @@
...
+ */
+export class SchemaLocalization {
+  private _provider: ILocalizationProvider;
+  private _locale: string;
+  private _cache: Map<string, SchemaLocalizationJson | null> = new Map();
+
+  /**
+   * Constructs a SchemaLocalization instance.
+   * @param provider The localization provider to use for loading localization files
+   * @param locale The target locale (e.g., "de", "fr", "es-CO")
```
karolis-zukauskas, line 46: Need better documentation of what is expected and allowed here .. what format are we using to specify the locale? Is it something like https://learn.microsoft.com/en-us/globalization/locale/standard-locale-names ?

Will we support locales like `de-DE-u-co-phonebk` ?

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,362 @@
...
+    return undefined;
+  }
+
+  /**
+   * Load and cache the localization JSON for a schema.
+   * @param schema The schema to load localization for
+   * @returns The localization JSON, or null if not found
+   * @internal
+   */
+  private async getSchemaLocalizationJson(schema: Schema): Promise<SchemaLocalizationJson | null> {
```
laurynas111: Use `undefined` instead of null

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
laurynas111: Bis allows changing both display labels and descriptions in minor schema upgrades.
If in a newer a description is changed, but a newer localization json is not released. The app would display a localized description that might be incorrect.

Do we want to handle this case? How can we handle it?
We could:

- Require the localization json to exact match the schema version number
- Duplicate English labels, descriptions in the json file. If schema contains a different English label than the one defined in the JSON - we could fall back to the schema one - same as it is done when no translation is found

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,368 @@
...
+      if (localizationMajor === undefined)
+        throw new Error(`Schema localization major version is undefined.`);
+
+      if (schema.schemaKey.readVersion !== localizationMajor)
+        throw new Error(`Localization version mismatch for schema "${schema.name}". Schema major version is ${schema.schemaKey.readVersion.toString()}, but localization is for major version ${localizationMajor}.`);
+    }
+
+    this._cache.set(cacheKey, localization || null);
+    return localization || null;
+  }
```
rschili, line 117: The public surface of `SchemaLocalization` currently takes ECObjects deserialized types - `Schema`, `SchemaItem`, `ECClass`, `Property`, `Enumeration`:

```ts
getSchemaLabel(schema: Schema): Promise<string>
getSchemaItemLabel(item: SchemaItem): Promise<string>
getPropertyLabel(ecClass: ECClass, property: Property): Promise<string>
getEnumeratorLabel(enumeration: Enumeration, enumeratorName: string): Promise<string>
```

That ties the feature to one specific way of accessing schema metadata.
I just published SchemaView, which is a separate API, localization this way is incompatible with it.
Also people querying through ECDbMeta also cannot use this API.
They would have to either spin up a full `SchemaContext` just to resolve labels

Reading through the bodies of the public methods, the only fields touched are:

- Schema: `name`, `schemaKey.readVersion`, `label`, `description`
- SchemaItem: `name`, `schemaItemType`, `schema.name`, `label`, `description`
- Property: `name`, `label`, `description` (plus owning class name + schema name)
- Enumerator: enum name, schema name, enumerator name, original label

Nothing here requires the heavy ECObjects types. And we don't even need an item-kind discriminator: schema items all live in one namespace within a schema (a class and an enumeration cannot share a name), and sub-items (properties on classes, enumerators on enumerations) are unique within their parent. A 3-segment path `schemaName.itemName.subItemName?` is enough to identify anything we localize.

I suggest restructuring around that, with thin convenience overloads for each metadata API on top:

```ts
// Core
getLabel(schemaName: string, itemName?: string, subItemName?: string): Promise<string | undefined>;
getDescription(schemaName: string, itemName?: string, subItemName?: string): Promise<string | undefined>;

// Convenience wrappers (kept in this class or alongside)
getSchemaItemLabel(item: SchemaItem): Promise<string>          // existing
getSchemaItemLabel(item: SchemaView.Class | SchemaView.Enumeration | ...): Promise<string>
```

The wrappers just shred the input down to the three strings. SchemaView and ECDbMeta consumers can either use those overloads or call the core directly with the row data they already have.

This also lets the JSON format simplify - the ten optional buckets in `LocalizationTypes.ts` (`classes`, `enumerations`, `units`, `invertedUnits`, `phenomena`, ...) collapse into a single `items` map keyed by schema-item name, with an optional `members` map for properties/enumerators:

```json
{
  "$schema": "ecschema-localization-v1",
  "name": "BuildingPhysical",
  "locale": "de",
  "label": "...",
  "description": "...",
  "items": {
    "Building":     { "label": "Gebäude",     "members": { "Height":      { "label": "Höhe" } } },
    "BuildingType": { "label": "Gebäudetyp",  "members": { "Residential": { "label": "Wohngebäude" } } }
  }
}
```

The `findItemLocalization` switch at lines 317-359 disappears entirely - the loader just does `items[name].members?.[subName]`.

One thing not covered today either way: relationship class role labels (`roleLabel` on a constraint). If we ever want to localize those, this shape extends naturally as `items[relName].members["source"|"target"]`.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: Fixed.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: Fixed.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: Now we show a warning instead of the error.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: I think its a good suggestion. I'll update it.

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
naveedkhan8067: Sure, we need to do that in bis-schemas repository because localization files will be packaged from there.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: I have restructured the code to support it.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: Definitely, we need documentation but it should be in bis-schemas repository where these localization files would be. As soon as this PR merge, I need to create a complete validation process there by using this api and create the complete documentation there as well. I have also added respective markdown file here.

At the moment, we only support a primary language tag and one optional subtag (es-CO) but it shouldn't be hard to support multiple optional subtags if we need them.

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
naveedkhan8067: Already moved.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: @diegoalexdiaz , what do you think?

diegoalexdiaz: My experience on the topic of schema-localization is based on what we did in OpenFlows. The scenario that @laurynas111 raises could also happen in that product line but was never considered a red-flag. That is, a dataset with newer schema (i.e. new classes/props or updated labels/descriptions) could be opened with a localized app that didn't have a perfect localization resource for it. 

While the localization of the app was catching up (months behind), the OpenFlows apps show localized strings that match the schema entries, and default/unlocalized strings for unmatched cases. That was considered acceptable.

Becoming very strict for everyone seems like an "all-or-nothing" approach that would cause localized apps to completely default to English while the localization resource is updated - even if its current resource covers 99% of the newer schema.

It seems to me that the approach in this PR supports controlling the applicability of a localization resource to a range of schema-versions on a per-schema basis (e.g. version: "01.XX.XX"), so if anyone needs it to be stricter for some particular use-case, it can still be done.

In summary, my vote is to leave it the way it is.

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
naveedkhan8067: I have also created a doc file for reference.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: I have fixed it.

core/ecschema-metadata/src/Localization/LocalizationProvider.ts
naveedkhan8067: I have added the test case.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,223 @@
...
+    const cacheKey = `${schemaName}:${locale}`;
+    if (this._cache.has(cacheKey))
+      return;
+
+    const localization = await this._provider.getLocalization(schemaName, locale);
+    if (localization?.version) {
+      const localizationMajor = this.getMajorVersion(localization.version);
+      if (localizationMajor === undefined || readVersion !== localizationMajor) {
+        // eslint-disable-next-line no-console
+        console.warn(`Localization version mismatch for schema "${schemaName}". Schema major version is ${readVersion.toString()}, but localization is for major version ${localizationMajor?.toString() ?? "undefined"}.`);
```
hl662: Maybe use `Logger` from `@itwin/core-bentley` instead of `console.warn` (+ eslint-disable)?`SchemaFormatsProvider.ts` in this same package does `Logger.logError(loggerCategory, ...)`

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
```diff
@@ -0,0 +1,223 @@
...
+  public get provider(): ILocalizationProvider {
+    return this._provider;
+  }
+
+  /**
+   * Load localizations for the given schema keys.
+   * @param schemaKeys The schema keys to load localization for
+   */
+  public async loadLocalizations(schemaKeys: Iterable<SchemaKey>): Promise<void> {
+    const baseLocale = this._locale.includes("-") ? this._locale.split("-")[0] : undefined;
```
hl662: Nit: this base-locale derivation is duplicated in `getCachedBaseLocalization`. Maybe a private getter method for baseLocale would be better? It would return string | undefined, and you can reuse it across the 2 places instead of code duplication.  Same for the `${schemaName}:${locale}` cache key, which is built in three places - a small private `cacheKey(name, locale)` helper could centralize and dedupe.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: I have added the logger now.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: Fixed.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
naveedkhan8067: I have combined them into one function.

laurynas111 (approved):

soham-bentley (approved):

docs/learning/metadata/SchemaLocalization.md
```diff
@@ -0,0 +1,134 @@
+# Schema Localization
```
rschili, line 1: I would like to see a small section in https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/metadata/SchemaView.md

And possibly in other places of the documentation as well, and this standalone page should probably go into the public documentation which gets published on our website. I guess like you mention the file format docs will be in the bis-schemas repo, but there needs to be some piece for this in our public documentation.

core/ecschema-metadata/src/Localization/SchemaLocalization.ts
rschili: Thanks, that has been resolved nicely, I like it!

rschili (approved):
All my comments have been resolved nicely, thanks. I left one more about documentation, please handle before merging. I'm still approving already.

docs/learning/metadata/SchemaLocalization.md
naveedkhan8067: I have added a section in SchemaView.md that point to this documentation.

The public documentation related to BIS schemas is generated from bis-schemas repository and then further consumed by itwinjs-core to publish. I think we should be ale to publish it from there.

hl662 (approved):

docs/learning/metadata/SchemaLocalization.md
naveedkhan8067: Thanks, it is now moved to `docs\learning\metadata\`

diegoalexdiaz (approved):

PR #9173, opened 2026-04-03
Add new ViewState.getRealityModelTreeRefs method

### LINKED ISSUE #9129: ScalableMeshModel and displayStyleState.contextRealityModels
For civil, we have a case where reality data has been added in different ways on two iModels, and we fail to pick up one of them for computing the cross section.

The first dataset has uses a persisted model (ie. `ScalableMesh.ScalableMeshModel`) with a valid `tilesetUrl`. As the model is turned on or off, the reality data is displayed properly, but it seems there is no way for us to obtain the underlying `TileTreeReference`.

The second dataset displays the reality data by attaching it to the display style using `displayStyle.realityModels`. The id of the model is a transientId since it is not backed by a persistent model. We are able to get a hold of the tile tree ref by iterating on `displayStyle.realityModels` and can compute the cross section if the tile tree outputs geometry.

**Expectations**
I would've expected that, since they're both reality models in the end, they'd both be listed under `displayStyle.realityModels`. It isn't the case.

I'd like a **uniform API** to be able to get to reality data. I don't get why both cases have to work differently, and no API user should care for those details either.

Please reach out if more info is needed.
Thanks,
Alex

eringram:
Fixes https://github.com/iTwin/itwinjs-core/issues/9129

This PR adds a new method `ViewState.getRealityModelTreeRefs` which yields both the context reality models attached to the `DisplayStyleState` as well as persistent reality models. It returns them in the form of a new `ViewRealityModel` interface which also includes their `name` and `description`. 

As explained in the issue, prior to this there was no unified public API to get both context and persistent reality models: you could either use `displayStyle.realityModels` to get only the context reality models, or `ViewState.getTileTreeRefs` to get ALL `TileTreeReference`s in the view (including non-reality models).

**Important note:** the new `ViewState.getRealityModelTreeRefs` returns all visible context reality models (excluding invisible ones), regardless of whether their tile trees have loaded. For persistent reality models, it can only return those whose tile trees have already loaded, because identifying a model as a reality model requires accessing the loaded tile tree to obtain the model ID.


core/frontend/src/ViewState.ts
```diff
@@ -44,6 +44,24 @@ import { ViewStatus } from "./ViewStatus";
 import { EnvironmentDecorations } from "./EnvironmentDecorations";
 import { _scheduleScriptReference } from "./common/internal/Symbols";
 
+/** Describes a reality model visible in a [[ViewState]], providing its [[TileTreeReference]] along with
+ * display metadata such as its name and description.
+ * @see [[ViewState.getRealityModelTreeRefs]]
+ * @beta
+ */
+export interface ViewRealityModel {
```
ben-polinsky, line 52: Is there any need for a field to determine the type of reality model?

core/frontend/src/ViewState.ts
```diff
@@ -44,6 +44,24 @@ import { ViewStatus } from "./ViewStatus";
...
+  /** The tile tree reference used for rendering, hit testing, and geometry collection. */
+  readonly treeRef: TileTreeReference;
+  /** Display name of the reality model, if available. For persisted reality models this comes from the
+   * [[ModelState]]; for context reality models, from the [[ContextRealityModelState]].
+   */
+  readonly name: string | undefined;
```
hl662: Quick question — `name` is typed `string | undefined` here, but both sources (`ContextRealityModel.name` and `ModelState.name`) always provide a `string` (defaulting to `""`). Was the `| undefined` intentional to reserve flexibility, or just symmetry with `description`? If it's always a string in practice, tightening to just `string` would save consumers a null check.

core/frontend/src/test/SpatialViewState.test.ts
```diff
@@ -111,4 +112,128 @@ describe("SpatialViewState", () => {
...
+          return { asSpatialModel: { isRealityModel: true }, name: "Persisted Reality" } as any;
+        if (id === nonRealityModelId)
+          return { asSpatialModel: { isRealityModel: false }, name: "Not Reality" } as any;
+        return undefined;
+      });
+
+      const contextState = view.displayStyle.attachRealityModel({ tilesetUrl: "https://fake.com/tileset.json", name: "Context Reality", description: "A context model" });
+
+      const refs = collectRefs(view);
+      expect(refs).to.have.length(2);
```
hl662: Nit: the new tests use Chai-chain style (`.to.equal()`, `.to.have.length()`) while the existing tests in this file use Vitest-native (`.toEqual()`, `.toHaveBeenCalled()`). Might be worth switching to `.toBe()` / `.toHaveLength()` / `.toBeUndefined()` for consistency — especially since `.to.equal()` is strict identity while `.toEqual()` is deep equality, and the intent isn't obvious when both appear in the same file.

core/frontend/src/test/SpatialViewState.test.ts
```diff
@@ -111,4 +112,128 @@ describe("SpatialViewState", () => {
...
+      const view = SpatialViewState.createBlank(iModel, { x: 0, y: 0, z: 0 }, { x: 1, y: 1, z: 1 });
+      view.getModelTreeRefs = () => [];
+
+      const visible = view.displayStyle.attachRealityModel({ tilesetUrl: "https://fake.com/a.json", name: "Visible" });
+      const invisible = view.displayStyle.attachRealityModel({ tilesetUrl: "https://fake.com/b.json", name: "Invisible" });
+      invisible.invisible = true;
+      const alsoVisible = view.displayStyle.attachRealityModel({ tilesetUrl: "https://fake.com/c.json", name: "Also Visible" });
+
+      const refs = collectRefs(view);
+      expect(refs).to.have.length(2);
```
hl662: Minor: the "yields only visible context models" test doesn't assert on `description` when no description is supplied. An `expect(refs[0].description).toBeUndefined()` (or `.toBe("")` depending on the contract) would make the expected behavior explicit.

hl662 (commented):
Reviewed the API surface, core logic, and tests. Clean, well-tested addition — just a few minor questions and suggestions.

ben-polinsky (approved):
LGTM. I noticed we don't really have a conceptual/learning doc for reality models. I don't  want to hold this PR up for that, but it might make sense to get one together.

common/changes/@itwin/core-frontend/eringram-scalable-mesh-reality-models-issue_2026-04-22-17-46.json
```diff
@@ -0,0 +1,10 @@
+{
```
markschlosseratbentley, line 1: NextVersion.md?

markschlosseratbentley (approved):

core/frontend/src/ViewState.ts
eringram: Agreed that it makes more sense as just `string`, fixed

core/frontend/src/test/SpatialViewState.test.ts
eringram: Fixed

core/frontend/src/test/SpatialViewState.test.ts
eringram: Fixed

hl662 (approved):
Noice

core/frontend/src/ViewState.ts
eringram: Good question. I could see it being useful if you want to distinguish them in a UI or something like that... Currently, I don't think there's a public way to tell just from obtaining the `TileTreeReference`. This new method treats them as identical.  
@a-gagnon Is this also necessary for your use case - a way to tell which of the returned `TileTreeReference`s are persisted or context models?

common/changes/@itwin/core-frontend/eringram-scalable-mesh-reality-models-issue_2026-04-22-17-46.json
markschlosseratbentley: @eringram Please feel free to self-resolve this comment so the PR is not held up, once you are satisfied.

core/frontend/src/ViewState.ts
a-gagnon: I don't really care for it right now, but it's certainly a nice to have for the future! Most likely it can be derived from the modelId somewhere in the Tile Tree props using `Id64.isTransient(modelId)`

eringram:
@ben-polinsky 
> LGTM. I noticed we don't really have a conceptual/learning doc for reality models. I don't want to hold this PR up for that, but it might make sense to get one together.

I agree, there isn't an obvious place where it's explained. The [2.17 changelog](https://www.itwinjs.org/changehistory/2.17.0/#reality-model-apis) explains the difference from when they were first introduced, and [this paragraph](https://www.itwinjs.org/bis/guide/physical-perspective/3d-guidance/#:~:text=There%20are%20some,the%20external%20repository.) I found in the BIS docs explains how persistent models are considered more legacy(?) at this point

core/frontend/src/ViewState.ts
eringram: Thanks. I won't add it for now to keep the new interface minimal and because `Id64.isTransient` likely works, but since this is beta we can revisit it if needed

ben-polinsky:
> @ben-polinsky
> 
> > LGTM. I noticed we don't really have a conceptual/learning doc for reality models. I don't want to hold this PR up for that, but it might make sense to get one together.
> 
> I agree, there isn't an obvious place where it's explained. The [2.17 changelog](https://www.itwinjs.org/changehistory/2.17.0/#reality-model-apis) explains the difference from when they were first introduced, and [this paragraph](https://www.itwinjs.org/bis/guide/physical-perspective/3d-guidance/#:~:text=There%20are%20some,the%20external%20repository.) I found in the BIS docs explains how persistent models are considered more legacy(?) at this point

@copilot please assemble a learning document pertaining to reality models. Consider the above comments.

ben-polinsky:
> @ben-polinsky
> 
> > LGTM. I noticed we don't really have a conceptual/learning doc for reality models. I don't want to hold this PR up for that, but it might make sense to get one together.
> 
> I agree, there isn't an obvious place where it's explained. The [2.17 changelog](https://www.itwinjs.org/changehistory/2.17.0/#reality-model-apis) explains the difference from when they were first introduced, and [this paragraph](https://www.itwinjs.org/bis/guide/physical-perspective/3d-guidance/#:~:text=There%20are%20some,the%20external%20repository.) I found in the BIS docs explains how persistent models are considered more legacy(?) at this point

@copilot+gpt-5.4 Could you compile this conceptual/learning documentation for reality models and add it in our docs?

ben-polinsky (approved):

eringram:
Moved the reality model docs here: https://github.com/iTwin/itwinjs-core/pull/9226

PR #9190, opened 2026-04-10
Fix CVEs issues in `release/4.11.x` branch

GytisCepk:
`rush update -full` + necessary bumps and overrides

### DIFF
--- a/common/config/rush/pnpm-config.json
+++ b/common/config/rush/pnpm-config.json
@@ -25,7 +25,7 @@
     "serialize-javascript": "^7.0.3", // https://github.com/advisories/GHSA-5c6j-r48x-rmvq mocha>serialize-javascript (related to CVE-2020-7660)
     "sequelize": ">=6.37.8", // https://github.com/advisories/GHSA-6457-6jrx-69cr azurite>sequelize
     "underscore": ">=1.13.8", // https://github.com/advisories/GHSA-qpx9-hpmf-5gmw json-schema-faker>jsonpath>underscore
-    "path-to-regexp@<0.1.13": ">=0.1.13", // https://github.com/advisories/GHSA-37ch-88jc-xwx2 express>path-to-regexp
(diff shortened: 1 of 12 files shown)


common/config/rush/pnpm-config.json
```diff
@@ -37,7 +43,12 @@
           "CVE-2025-58754", // https://github.com/advisories/GHSA-4hjh-wcwx-xvwj azurite>@azure/ms-rest-js>axios
           "CVE-2026-26996", // https://github.com/advisories/GHSA-3ppc-4f35-3m26 mocha>glob>minimatch
           "CVE-2025-59288", // https://github.com/advisories/GHSA-7mvr-c777-76hp @itwin/oidc-signin-tool>@playwright/test>playwright (playwright is dev only dep, requires active MitM to exploit)
-          "CVE-2026-1615" // https://github.com/advisories/GHSA-87r5-mp6g-5w5j presentation/common>json-schema-faker>jsonpath
+          "CVE-2026-1615", // https://github.com/advisories/GHSA-87r5-mp6g-5w5j presentation/common>json-schema-faker>jsonpath
+          // Fixes for following CVEs only available in newer Electron versions which are not supported by iTwin.js 4.x. Non of the CVEs should be exploitable in library itself.
```
GytisCepk, line 47: Oldest Electron version with all fixes available is 39. Highest supported by 4.x is 33.

I don't think, this is exploitable in library itself, so adding support for so many Electron versions might be overkill. If someones complains, we can look at revisiting this.

hl662 (approved):
i wonder if we should let copilot draw inspiration from this https://github.com/iTwin/itwinjs-core/pull/8861 and open a separate PR, so we can do rush update --full for 4.11.x branch

GytisCepk:
> i wonder if we should let copilot draw inspiration from this #8861 and open a separate PR, so we can do rush update --full for 4.11.x branch

I don't really want to start changing code in old branch. 

What if we (and this feels wrong typing) disabled those 2 new rules globally for this branch ? There isn't active development or backporting features to this branch, so benefit of those new rules might be non-existant

rschili (approved):

aruniverse (approved):

PR #9192, opened 2026-04-13
`iTwin Workspace` auto-resolves account iTwin workspace values

johnnyd710:
Now that blob container service supports [returning account iTwin containers](https://github.com/iTwin/blob-container-service/pull/25), we can automatically get settings containers for the account iTwin and treat those as a lower priority (smaller scope means higher priority, so iModel > iTwin > Organization > App) in the same workspace when getting the iTwin Workspace.

For example:
```typescript
AzuriteTest.userToken = AzuriteTest.service.userToken.admin;

await IModelHost.saveSettingDictionary(iTwin1Id, "app1/config", { "app1/max1": 17 });
await IModelHost.saveSettingDictionary(iTwin2Id, "org/defaults", { "app1/max1": 42, "app1/max2": 99 });
AzuriteTest.setParentITwin(iTwin1Id, iTwin2Id);

AzuriteTest.userToken = AzuriteTest.service.userToken.readWrite;

const workspace = await IModelHost.getITwinWorkspace(iTwin1Id);
try {
  expect(workspace.settings.getNumber("app1/max1")).equal(17); // iTwin-priority overrides org
  expect(workspace.settings.getNumber("app1/max2")).equal(99); // org-priority fallback
} finally {
  workspace.close();
}
```


core/backend/src/BlobContainerService.ts
```diff
@@ -73,6 +73,10 @@ export namespace BlobContainer {
    */
   export interface MetadataResponse extends Metadata {
     containerId: string;
```
hl662, line 76: nit: Can you help out with adding a simple doc comment for `containerId`?

core/backend/src/BlobContainerService.ts
```diff
@@ -158,6 +162,8 @@ export namespace BlobContainer {
     containerType?: GuidString;
     /** optional label of the containers to query */
     label?: GuidString;
+    /** whether to include parent iTwins in the query */
```
hl662: nit: The `@internal` `QueryArgs` version documents `Defaults to 'false'`. Could align here too:
```typescript
/** Whether to include parent iTwins in the query. Defaults to `false`. */
```

hl662 (commented):
Nice work on the parent iTwin workspace support! A few suggestions below.

**Additional notes (on lines outside the diff):**

- **`core/backend/src/workspace/SettingsEditor.ts` (~98):** The JSDoc here still says "the settings container" (singular) — now that this returns entries for parent iTwins too, might be worth updating to reflect the multi-entry return and the priority split?
- **`core/backend/src/IModelHost.ts` (~443):** Related: this line says settings load at `SettingsPriority.iTwin`, but parent iTwin containers now load at `SettingsPriority.organization`. Could this doc comment get a small update too?

core/backend/src/workspace/SettingsEditor.ts
```diff
@@ -102,24 +104,44 @@ export namespace SettingsContainers {
...
+          message: `Multiple iTwin settings containers were found for '${ownerITwinId}', so a container cannot be automatically selected.`,
+          iTwinId: ownerITwinId,
+        });
+      }
+    }
+
+    const results: WorkspaceDbSettingsProps[] = [];
+    for (const [ownerITwinId, group] of byITwin) {
+      const [container] = group;
+      const priority = ownerITwinId === iTwinId ? SettingsPriority.iTwin : SettingsPriority.organization;
```
hl662, line 131: I know we discussed our concern with multi-level ancestry earlier — if the service ever returns containers beyond the direct parent, they'd all share the same `organization` priority here. Might be worth a short comment noting the single-parent assumption?

core/backend/src/workspace/SettingsEditor.ts
```diff
@@ -102,24 +104,44 @@ export namespace SettingsContainers {
...
+      const tokenProps = await BlobContainer.service.requestToken({ containerId: container.containerId, accessLevel: "read", userToken });
```
hl662, line 132: Curious — these `requestToken` calls are independent, so they could run in parallel with `Promise.all`. In practice it's probably just two containers so the gain is marginal — totally fine to leave as-is if you'd rather keep it simple.

core/backend/src/workspace/SettingsEditor.ts
```diff
@@ -102,24 +104,44 @@ export namespace SettingsContainers {
       ITwinSettingsError.throwError("blob-service-unavailable", { message: "BlobContainer.service is not available." });
 
     const userToken = await IModelHost.getAccessToken();
-    const containers = await queryContainers({ iTwinId });
+    const containers = await queryContainers({ iTwinId, includeParentITwins: true });
     if (!containers || containers.length === 0) return undefined;
```
hl662: Boy scout help: `queryContainersMetadata` returns `Promise<MetadataResponse[]>`, so the `!containers` check here is unreachable — could simplify to just `if (containers.length === 0)`.

full-stack-tests/backend/src/integration/CloudWorkspace.test.ts
```diff
@@ -445,6 +445,19 @@ describe("Cloud workspace containers", () => {
...
       expect(deletedWorkspace.settings.getNumber("app1/max1")).equal(17);
       expect(deletedWorkspace.resolveWorkspaceDbSetting("app1/styles/textStyleDbs").length).equal(0);
+
+      // Verify organization-priority (parent iTwin) settings provide fallback for iTwin-priority settings.
+      await IModelHost.saveSettingDictionary(iTwin2Id, "org/defaults", { "app1/max1": 42, "app1/max2": 99 });
+      const orgWorkspace = await IModelHost.getITwinWorkspace(iTwin2Id);
+      assert(orgWorkspace.settingsSources !== undefined && deletedWorkspace.settingsSources !== undefined);
+      const orgSources = [orgWorkspace.settingsSources].flat().map((s) => ({ ...s, priority: SettingsPriority.organization }));
+      const childSources = [deletedWorkspace.settingsSources].flat();
+      const combinedWorkspace = await IModelHost.getITwinWorkspace([...orgSources, ...childSources]);
```
hl662: Could we extend the Azurite mock to support `includeParentITwins` so this can go through the actual `getITwinWorkspace(childITwinId)` codepath instead of manually combining sources? Would give us end-to-end coverage of the new settings discovery flow.

The filter to relax is in `full-stack-tests/backend/src/integration/AzuriteTest.ts` around line 196 — `queryContainersMetadata` currently matches only `m?.itwinid === iTwinId`.

core/backend/src/test/standalone/ITwinWorkspace.test.ts
```diff
@@ -276,4 +304,40 @@ describe("ITwin Workspace", () => {
     await expect(IModelHost.deleteSettingDictionary(iTwinId, "x")).to.be.rejectedWith("delete failed");
     expect(close.calledOnce).to.be.true;
   });
```
hl662, line 308: Would you consider adding a test for the case where only the parent iTwin has a settings container (no child container)? Seems like a natural scenario — org provides defaults but the child hasn't customized yet.

core/backend/src/test/standalone/ITwinWorkspace.test.ts
```diff
@@ -67,6 +67,34 @@ describe("ITwin Workspace", () => {
...
+    parentITwinId: string,
+    childContainerIds: string[],
+    parentContainerIds: string[],
+  ): BlobContainer.ContainerService {
+    return {
+      create: async () => ({ baseUri: "", containerId: childContainerIds[0], provider: "azure" as const }),
+      delete: async () => { },
+      queryScope: async () => ({ iTwinId: childITwinId }),
+      queryMetadata: async () => ({ containerType: "settings", label: "settings" }),
+      queryContainersMetadata: async () => [
```
hl662: Minor thought — the mock returns parent containers unconditionally, so if `includeParentITwins: true` got accidentally dropped, these tests would still pass. Probably fine if we get the Azurite integration coverage from the earlier comment, but a quick spy could also catch it cheaply.

johnnyd710: fixed, now it checks for parentITwinId arg

core/backend/src/test/standalone/ITwinWorkspace.test.ts
johnnyd710: added a test but to CloudWorkspace.test.ts

full-stack-tests/backend/src/integration/CloudWorkspace.test.ts
johnnyd710: extended azurite mocks

core/backend/src/workspace/SettingsEditor.ts
johnnyd710: thanks, fixed

core/backend/src/workspace/SettingsEditor.ts
johnnyd710: In this case, I think its better to keep it simple.

core/backend/src/workspace/SettingsEditor.ts
johnnyd710: added a comment about this

core/backend/src/BlobContainerService.ts
johnnyd710: added

core/backend/src/BlobContainerService.ts
johnnyd710: added

core/backend/src/BlobContainerService.ts
```diff
@@ -158,6 +163,8 @@ export namespace BlobContainer {
     containerType?: GuidString;
     /** optional label of the containers to query */
     label?: GuidString;
+    /** Whether to include parent iTwins in the query. Defaults to `false`. */
+    includeParentITwins?: boolean;
```
aruniverse, line 167: are there multiple parents?
or does this mean recursively go up and get the itwin hierarchy?

at a quick glance, im not sure about adding parentITwin to the lib here. i think its fine for the service, but wonder if we are bleeding our needs too much into the client lib

johnnyd710: Yeah unfortunately there can be multiple parents - im doing the simplest thing by just treating all parents as Org level priority.

I believe this is fine because you mostly only care about the current iTwin's settings, the rest are fallbacks, and if they happen to have the same settings dictionaries they will be overridden as you get higher in the stack

johnnyd710: not sure I follow your reservations about adding `includeParentITwins` to the Blob Container service type. Its already added to the client implementation, so i'm just matching the type to the implementation now

full-stack-tests/backend/src/integration/AzuriteTest.ts
```diff
@@ -137,6 +137,19 @@ export namespace AzuriteTest {
   }
 
   const fakeUser = () => `token ${Guid.createValue()}`;
+  /** Maps child iTwinId → parent iTwinId for simulating iTwin hierarchy in tests. */
+  const parentITwins = new Map<string, string>();
+  export const setParentITwin = (childITwinId: string, parentITwinId: string) => parentITwins.set(childITwinId, parentITwinId);
+  export const clearParentITwins = () => parentITwins.clear();
+
+  const isAncestorITwin = (ownerITwinId: string, requestedITwinId: string): boolean => {
+    for (let current = parentITwins.get(requestedITwinId); current !== undefined; current = parentITwins.get(current)) {
```
hl662, line 146: If `parentITwins` ever ends up with a cycle (A→B→A), this walks forever. Might be worth a small `visited` Set to break out?

Also curious whether the production `BlobContainerService` parent-walk has equivalent cycle protection — could mirror whatever it does here.

johnnyd710: After talking to @calebmshafer today, I'm going to change this a bit. Instead of recursively going up the chain and getting all parent iTwins, I'm just going to get [the *account* iTwin](https://developer.bentley.com/apis/itwins/operations/get-my-itwins/) and treat that as the org.

I'll need to update the blob container service again so I'll mark this as draft again

johnnyd710: https://github.com/iTwin/blob-container-service/pull/29 - waiting for a new release of blob container service

johnnyd710: done

johnnyd710:
Just trying to test this one more time in a cospace

Update: works!

markschlosseratbentley (commented):
## Review: `iTwin Workspace` auto-resolves account iTwin workspace values
 
 Nice change — the design is clean and the prior breaking-change concern was handled well by making the new `MetadataResponse` fields (`iTwinId`, `accountITwinId`) optional, so existing custom `BlobContainer.service` implementations keep compiling. Good test coverage too: the new "child overrides account", "child with no container inherits account", and "multiple containers for same account iTwin" cases cover the meaningful scenarios, and the Azurite end-to-end path is exercised. API tags look correct (optional additions to the `@beta` `BlobContainer` namespace are non-breaking; the removed `SettingsContainers.QueryArgs` is `@internal`). Changelog entries present.
 
 A couple of things worth a look:
 
 ### `includeParentITwins` boolean variant is unimplemented / ambiguous
 
 ```typescript
 includeParentITwins?: boolean | { filter: "accountOnly" };
 ```
 
 The `boolean` arm is never handled anywhere — both the production call in `SettingsEditor` and the Azurite mock only check the `{ filter: "accountOnly" }` object form:
 
 ```typescript
 const includeAccountITwin = typeof includeParentITwins === "object" && includeParentITwins.filter === "accountOnly";
 ```
 
 So passing `includeParentITwins: true` silently does nothing, yet the doc comment ("Whether to include the account iTwin in the query… Defaults to `false`") reads as if `true` is meaningful. Given the recursive parent-chain approach was dropped in favor of account-only, consider either narrowing the public `@beta` type to just `{ filter: "accountOnly" }`, or clearly documenting what `boolean` means and ensuring it's actually handled. Exposing an unhandled option on a public API tends to confuse consumers.
 
 ### Priority when the requested iTwin *is* its own account iTwin
 
 ```typescript
 const priority = container.accountITwinId === ownerITwinId ? SettingsPriority.organization : SettingsPriority.iTwin;
 ```
 
 If `getITwinWorkspace` is called on an iTwin that is itself the account iTwin (`ownerITwinId === accountITwinId`), its own settings load at `SettingsPriority.organization` rather than `SettingsPriority.iTwin`. That contradicts the updated `IModelHost.getITwinWorkspace` doc, which states the requested iTwin's resources load at `SettingsPriority.iTwin`. It's likely a benign edge case (account == organization), but a one-line doc clarification would avoid the mismatch.
 
 ### Minor
 
 - `AzuriteTest.ts`: double blank line after `getAccountITwinId` (test-only, take it or leave it).
 
 Image Tests aren't needed here — this is backend settings/container resolution with no rendering impact.
 
 > Generated by 🤖 markschlosserbot and manually reviewed by @markschlosseratbentley.

johnnyd710:
> A couple of things worth a look:
> 
> ### `includeParentITwins` boolean variant is unimplemented / ambiguous
> ```ts
> includeParentITwins?: boolean | { filter: "accountOnly" };
> ```
> 
> The `boolean` arm is never handled anywhere — both the production call in `SettingsEditor` and the Azurite mock only check the `{ filter: "accountOnly" }` object form:
> 
> ```ts
> const includeAccountITwin = typeof includeParentITwins === "object" && includeParentITwins.filter === "accountOnly";
> ```
> 
> So passing `includeParentITwins: true` silently does nothing, yet the doc comment ("Whether to include the account iTwin in the query… Defaults to `false`") reads as if `true` is meaningful. Given the recursive parent-chain approach was dropped in favor of account-only, consider either narrowing the public `@beta` type to just `{ filter: "accountOnly" }`, or clearly documenting what `boolean` means and ensuring it's actually handled. Exposing an unhandled option on a public API tends to confuse consumers.

Those types are just a re-export of the types from blob container service, I guess someone didn't want to introduce a dependency so they are re-created here. Its a good question though, should we try to be as close to the source-of-truth as possible, or just take what we need? I can remove the boolean option since we aren't using it, but [the boolean option is supported on `BlobContainer.service`](https://github.com/iTwin/blob-container-service/pull/25).

> ### Priority when the requested iTwin _is_ its own account iTwin
> ```ts
> const priority = container.accountITwinId === ownerITwinId ? SettingsPriority.organization : SettingsPriority.iTwin;
> ```
> 
> If `getITwinWorkspace` is called on an iTwin that is itself the account iTwin (`ownerITwinId === accountITwinId`), its own settings load at `SettingsPriority.organization` rather than `SettingsPriority.iTwin`. That contradicts the updated `IModelHost.getITwinWorkspace` doc, which states the requested iTwin's resources load at `SettingsPriority.iTwin`. It's likely a benign edge case (account == organization), but a one-line doc clarification would avoid the mismatch.

Yeah I would expect settings in `getITwinWorkspace(orgId)` to have priority `SettingsPriority.organization`. The impl is right but the doc is wrong. I will update the doc, thanks

> ### Minor
> * `AzuriteTest.ts`: double blank line after `getAccountITwinId` (test-only, take it or leave it).
> 
> Image Tests aren't needed here — this is backend settings/container resolution with no rendering impact.

Good points, will update

full-stack-tests/backend/src/integration/AzuriteTest.ts
```diff
@@ -142,12 +142,15 @@ export namespace AzuriteTest {
...
   export const clearParentITwins = () => parentITwins.clear();
 
-  const isAncestorITwin = (ownerITwinId: string, requestedITwinId: string): boolean => {
-    for (let current = parentITwins.get(requestedITwinId); current !== undefined; current = parentITwins.get(current)) {
-      if (ownerITwinId === current)
-        return true;
+  const getAccountITwinId = (requestedITwinId: string): string => {
+    const visited = new Set<string>();
+    let current = requestedITwinId;
+    for (let parent = parentITwins.get(current); parent !== undefined && !visited.has(parent); parent = parentITwins.get(current)) {
```
hl662: if the real implementation no longers 'walks' up iTwins to find the parent iTwin, maybe we don't need this cycle detection stuff, its inaccurate to the real behavior, and maybe just do a single parent lookup?

Like `parentITwins.get(requestedITwinId) ?? requestedITwinId`

hl662 (approved):

markschlosseratbentley (approved):
@johnnyd710 

 ## Re-review — follow-up
 
 Thanks for the quick turnaround. Confirmed resolved:
 
 - ✅ `IModelHost.getITwinWorkspace` doc now correctly states account-iTwin resources load at `SettingsPriority.organization` and the requested iTwin (when different) at `SettingsPriority.iTwin` — matches the impl. 👍
 - ✅ Double blank line in `AzuriteTest.ts` removed.
 
 ### Re: the `includeParentITwins: boolean | { filter: "accountOnly" }` type
 
 Thanks for the context — that it mirrors the `BlobContainer.service` source-of-truth type makes sense, and I agree keeping the client type close to the service contract is the right instinct. My only concern is the doc comment, which still reads as if `boolean` is meaningful here:
 
 ```typescript
 /** Whether to include the account iTwin in the query. Pass `{ filter: "accountOnly" }` to include only the account iTwin for the requested iTwin. Defaults to `false`. */
 includeParentITwins?: boolean | { filter: "accountOnly" };
 ```
 
 Since this repo's consumers only ever exercise the `{ filter: "accountOnly" }` path, I'd keep the `boolean` arm (to stay faithful to the service) but tighten the comment so the boolean semantics aren't ambiguous — e.g. note that `true` requests the full parent chain (as supported by the service) while `{ filter: "accountOnly" }` limits it to the account iTwin. That way no one has to read the service PR to understand what `true` does. Totally your call though — not a blocker.
 
 Everything else from the prior review is addressed. No rendering impact, so Image Tests aren't needed.
 
 > Generated by 🤖 markschlosserbot and manually reviewed/edited by @markschlosseratbentley

full-stack-tests/backend/src/integration/AzuriteTest.ts
johnnyd710: yup you're right. forgot about this test. simplified.

aruniverse (commented):
## Code Review — PR #9192
> Generated by 🤖 **arunalytics-bot**

The general approach is clean and the feature is well-motivated. Three issues landed across an adversarial multi-angle review (correctness, API design, and test quality). Two are blockers; the rest are optional.

---

## 🚨 Fixes Worth Doing Now

### Fix 1 — Remove `boolean` from `includeParentITwins` on `@beta` public API
**Severity: Blocker** · `core/backend/src/BlobContainerService.ts:167`

The type is `boolean | { filter: "accountOnly" }` with a doc comment saying `true` traverses the full parent chain. But **no implementation anywhere handles `boolean`** — the integration mock, unit mock, snippet mock, and only production call site all guard on `typeof x === "object"`. A caller passing `true` silently gets the same result as `false` (no parent traversal), contradicting the documented contract. This is dead code on a `@beta` API surface.

The `queryContainersMetadata` JSDoc on `ContainerService` (lines 188–191) also only mentions the object form, contradicting the field doc.

**Suggested fix — narrow the union:**
```typescript
// Before
includeParentITwins?: boolean | { filter: "accountOnly" };

// After
includeParentITwins?: { filter: "accountOnly" };
```
Update the doc comment and regenerate `common/api/core-backend.api.md`.

---

### Fix 2 — Dead `if` branch in `createAccountSettingsContainerService` mock
**Severity: Blocker** · `core/backend/src/test/standalone/ITwinWorkspace.test.ts:83,91`

The mock unconditionally `expect`s that `args.includeParentITwins` deep-equals `{ filter: "accountOnly" }`. If that `expect` passes, the subsequent `if (typeof args.includeParentITwins === "object" && ...)` is always `true` — dead code. Account containers are pushed regardless. This makes the mock fragile (reuse without the flag blows up with a confusing `AssertionError` instead of a descriptive test failure) and misleads reviewers.

**Suggested fix — drop the redundant `if` guard:**
```typescript
// Remove the if-check and push unconditionally,
// since the expect already enforces the contract:
results.push(...accountContainerIds.map((containerId) => ({
  containerId,
  containerType: "settings",
  label: containerId,
  iTwinId: accountITwinId,
  accountITwinId,
})));
```

---

## 🔧 Optional Improvements

### Opt-A — Add test for `missing-container-itwinid` error path
`SettingsEditor.ts:106–110` — This guard fires when `container.iTwinId === undefined`. There is zero test coverage: the `createSettingsContainerService` mock was updated in this PR to always set `iTwinId`, so the old tests don't hit it either. A single unit test with a mock that returns a container without `iTwinId` would close the gap.

### Opt-B — Add test for "child iTwin IS its own account iTwin" priority edge case
`SettingsEditor.ts:120`:
```typescript
const priority = container.accountITwinId === ownerITwinId ? SettingsPriority.organization : SettingsPriority.iTwin;
```
The unit mock omits `accountITwinId`, so `undefined !== ownerITwinId` → always `.iTwin`. But the AzuriteTest integration mock always sets `accountITwinId = iTwinId` when no parent is mapped, so existing integration tests (e.g., "can write and read settings to iTwin workspace") silently load at `.organization` instead of `.iTwin` after this PR. Unit and integration tests diverge in behavior for the same scenario. A targeted unit test where `accountITwinId === childITwinId` would document intent and close the gap.

### Opt-C — Add test for multiple *child* containers
`ITwinWorkspace.test.ts:349` — The "fails if multiple settings containers" test covers two **account** containers. The same `seenITwins` check also fires for two *child* containers, but there is no test for it.

### Opt-D — Add `@note` to `getITwinSettingsSources` re: service version requirement
The function throws a hard `"missing-container-itwinid"` error if `iTwinId` is absent on a `MetadataResponse`, but there is no doc-time warning. A single `@note` line would tell callers which service version is required to avoid this.

### Opt-E — Clean up settings containers in new integration tests
`CloudWorkspace.test.ts:458–459` — "child iTwin settings override account iTwin settings" calls `saveSettingDictionary(iTwin2Id, ...)`, creating a persistent container for the shared `iTwin2Id`. The `finally` only calls `clearParentITwins()` — the Azurite container is never deleted, and any subsequent test that uses `iTwin2Id` will find an unexpected container. A `deleteSettingDictionary` call in `finally` would close this.

---

## ✅ No Action Needed

| Finding | Reason |
|---|---|
| Double `getAccessToken()` call in `getITwinSettingsSources` | Functionally correct; caching is upstream. Minor nit, safe to defer. |
| `example-code/snippets` AzuriteTest ignores `includeParentITwins` / no `accountITwinId` | Example-only infrastructure; no test relies on account-iTwin traversal through this mock today. Safe follow-up. |
| Priority not explicitly asserted in integration tests | Functional behavior is verified; priority assertions would require internal API access and are more fragile than behavioral tests. |
| `ITwinSettingsError.Key` members undocumented | Pre-existing pattern; fixing all keys is wider than this PR scope. |

johnnyd710:
> ## 🚨 Fixes Worth Doing Now
> ### Fix 1 — Remove `boolean` from `includeParentITwins` on `@beta` public API
> **Severity: Blocker** · `core/backend/src/BlobContainerService.ts:167`
> 
> The type is `boolean | { filter: "accountOnly" }` with a doc comment saying `true` traverses the full parent chain. But **no implementation anywhere handles `boolean`** — the integration mock, unit mock, snippet mock, and only production call site all guard on `typeof x === "object"`. A caller passing `true` silently gets the same result as `false` (no parent traversal), contradicting the documented contract. This is dead code on a `@beta` API surface.

As discussed with @markschlosseratbentley above, [the implementation is in the blob container service](https://github.com/iTwin/blob-container-service/blob/main/core/src/BlobContainerService.ts). We don't use it, but we want to stay as close to the source-of-truth as possible.

> ### Fix 2 — Dead `if` branch in `createAccountSettingsContainerService` mock
> **Severity: Blocker** · `core/backend/src/test/standalone/ITwinWorkspace.test.ts:83,91`

Removed the if statement

> ## 🔧 Optional Improvements
> ### Opt-A — Add test for `missing-container-itwinid` error path
> `SettingsEditor.ts:106–110` — This guard fires when `container.iTwinId === undefined`. There is zero test coverage: the `createSettingsContainerService` mock was updated in this PR to always set `iTwinId`, so the old tests don't hit it either. A single unit test with a mock that returns a container without `iTwinId` would close the gap.

Added a test

> ### Opt-B — Add test for "child iTwin IS its own account iTwin" priority edge case
> `SettingsEditor.ts:120`:
> 
> ```ts
> const priority = container.accountITwinId === ownerITwinId ? SettingsPriority.organization : SettingsPriority.iTwin;
> ```
> 
> The unit mock omits `accountITwinId`, so `undefined !== ownerITwinId` → always `.iTwin`. But the AzuriteTest integration mock always sets `accountITwinId = iTwinId` when no parent is mapped, so existing integration tests (e.g., "can write and read settings to iTwin workspace") silently load at `.organization` instead of `.iTwin` after this PR. Unit and integration tests diverge in behavior for the same scenario. A targeted unit test where `accountITwinId === childITwinId` would document intent and close the gap.

Added a test

> ### Opt-E — Clean up settings containers in new integration tests
> `CloudWorkspace.test.ts:458–459` — "child iTwin settings override account iTwin settings" calls `saveSettingDictionary(iTwin2Id, ...)`, creating a persistent container for the shared `iTwin2Id`. The `finally` only calls `clearParentITwins()` — the Azurite container is never deleted, and any subsequent test that uses `iTwin2Id` will find an unexpected container. A `deleteSettingDictionary` call in `finally` would close this.

Added!

aruniverse (approved):

PR #9194, opened 2026-04-14
Changes to address issues with shortcuts being processed when focus is not on Home or AccuDraw.

bbastings:
Previously relied on checks in overrides of processShortcutKey or ui elements calling stopPropagation. For example, the keyin field used by display-test-app added stopPropagation checks to try to avoid this issue, every ui element should not be required to do this. The checks in processShortcutKey also don't stop ctrl or the active tool from handling keys when a ui element that doesn't stopPropagation has focus. (note: can see issue in studio test app where typing into keyin field runs AccuDraw shortcuts).

When focus is not Home or AccuDraw, pressing Escape now moves focus to Home so that shortcuts can be used. This does not prevent an application from using Escape to start the default tool in processShortcutKey while focus is Home. Escape is the most reliable key given it's default behavior to close popups, etc.

Fix onModifierKeyTransition for active tool. Previously, if focus was on an element that called stopPropagation the active tool would not be notified even though the user is interacting with the view. Moved this to the capture phase instead.

Example is using Shift modifier to invert the inside/overlap right->left/left->right direction for drag selection.
image


aruniverse (approved):

GerardasB (approved):
LGTM, as it improves the modifier key reliability AND allows customization of the default behavior via `processKeyboardEvent`.

PR #9200, opened 2026-04-15
LiteBriefcaseDb — a briefcase backed by a V2 checkpoint container

khanaffan:
imodel-native:  https://github.com/iTwin/imodel-native/pull/1385

### Lightweight briefcases for server and agent workflows (`LiteBriefcaseDb`)

The backend now supports opening a V2 checkpoint as a lightweight, writable briefcase via the new `LiteBriefcaseDb` class and `LiteBriefcaseOpenArgs` interface. Unlike a traditional `BriefcaseDb`, a lite briefcase **never downloads the full iModel** to disk — blocks are streamed on demand from the cloud container. This makes it ideal for short-lived edit sessions in server and agent workflows where only a small number of targeted changes are needed and full disk space is not available or practical.

#### What's new

- **`LiteBriefcaseDb`** (`@alpha`): a new `BriefcaseDb` subclass designed for lightweight, agent-friendly edit sessions. The full iModel is never downloaded; blocks stream from a V2 checkpoint container on demand. Local writes stay in a small local cache and are never uploaded back to the checkpoint container. Changesets can be pushed to iModelHub normally.
- **`LiteBriefcaseDb.openFromCheckpoint(args)`** (`@alpha`): static factory method that attaches the checkpoint container, acquires (or reuses) a briefcase ID, opens the database in read-write mode against the cloud container, and sets up SAS token refresh.
- **`LiteBriefcaseOpenArgs`** (`@alpha`): arguments interface for `openFromCheckpoint`, accepting a `CheckpointProps` and an optional pre-acquired `briefcaseId`.
- **`CloudSqlite.hasLocalChanges(container)`** (`@alpha`): helper to check whether a cloud container has uncommitted local changes that have not been uploaded.
- **`V2CheckpointManager.attachForBriefcase(checkpoint)`** (`@alpha`): attaches a V2 checkpoint and returns both the `dbName` and the live `CloudContainer`, throwing if a mock container is used (mocks do not support lite briefcase mode).

#### Key characteristics

- **No full download**: only the blocks actually read during your edits are fetched from the cloud.
- **Minimal disk footprint**: local storage is limited to changed blocks and the local cache — suitable for ephemeral server/agent environments.
- **Short-lived sessions**: designed for targeted edits (seconds to minutes), not long-running or full-file access workflows.
- **Agent/server friendly**: no BCV daemon required; blocks are fetched directly from Azure Blob Storage.

#### Constraints

- Local changes **must** be pushed to iModelHub before the briefcase is closed. Any unpushed changes are lost on close or crash — `LiteBriefcaseDb.close()` logs a warning if pending transactions exist.
- The checkpoint blob container is **never** modified; this is a read-from-cloud, write-local-only pattern.
- Pull/merge operations may trigger on-demand block fetching from the cloud container.
- Not suitable for workflows that read or modify large portions of the iModel — use a traditional `BriefcaseDb` in those cases.

#### Usage

```ts
import { LiteBriefcaseDb, withEditTxn } from "@itwin/core-backend";

const db = await LiteBriefcaseDb.openFromCheckpoint({
  accessToken,
  checkpoint: {
    accessToken,
    iTwinId,
    iModelId,
    changeset: { id: changesetId },
    expectV2: true,        // fail fast if no V2 checkpoint is available
    allowPreceding: true,  // accept the nearest preceding checkpoint
  },
});

// Make targeted edits — only needed blocks are fetched from the cloud
withEditTxn(db, "Add property", (txn) => {
  txn.saveFileProperty({ namespace: "MyApp", name: "key" }, "value");
});

// Push changesets to iModelHub when ready
await db.pushChanges({ accessToken, description: "My changes" });

db.close();
```


kabentley:
@khanaffan I don't understand this approach. Let's talk about it.

core/backend/src/IModelDb.ts
```diff
@@ -4089,8 +4089,172 @@ class RefreshV2CheckpointSas {
...
+    if (this._refreshSas) {
+      const token = accessToken ?? await IModelHost.getAccessToken();
+      await this._refreshSas.refreshSas(token, this);
+    }
+  }
+
+  /** Override close to warn if there are unpushed local changes. */
+  public override close(options?: CloseIModelArgs) {
+    if (this.isOpen && this[_nativeDb].hasPendingTxns()) {
+      Logger.logWarning(loggerCategory, "Closing CloudBriefcaseDb with unpushed local changes. These changes will be lost.");
```
tcobbs-bentley: Would it be better to throw an exception here telling the user to explicitly revert their changes?

khanaffan: So copilot did found this in review and suggest to log vs exception base on this been used in agent/server where there is not much you could do other then log.

core/backend/src/IModelDb.ts
```diff
@@ -4089,8 +4089,181 @@ class RefreshV2CheckpointSas {
...
+   *
+   * This is the preferred method for agent/server workflows where only targeted edits are needed
+   * and full disk space for a traditional briefcase is not practical.
+   *
+   * @param args parameters specifying the checkpoint and optional briefcaseId
+   * @returns a new LiteBriefcaseDb
+   * @throws IModelError if the checkpoint cannot be attached or if a briefcaseId cannot be acquired
+   * @alpha
+   */
+  public static async openCloud(args: OpenLiteBriefcaseArgs): Promise<LiteBriefcaseDb> {
```
tcobbs-bentley, line 4161: Should this perhaps just be `open`? The whole point of `LiteBriefcaseDb` is to make use of a V2 checkpoint container. Also, now that the class has been renamed, "cloud" feels more awkward here.

core/backend/src/IModelDb.ts
```diff
@@ -4089,8 +4089,181 @@ class RefreshV2CheckpointSas {
 
     return this._promise;
   }
+}
+
+/**
+ * Arguments to open a lite briefcase from a V2 checkpoint.
+ * @alpha
+ */
+export interface OpenLiteBriefcaseArgs extends TokenArg {
```
tcobbs-bentley, line 4098: I feel like this should be `LiteBriefcaseOpenArgs` (or `LiteBriefcaseOpenCloudArgs` if `openCloud` isn't renamed to `open`.

core/backend/src/CloudSqlite.ts
```diff
@@ -869,6 +869,15 @@ export namespace CloudSqlite {
     }
   }
 
+  /**
+   * Check whether a cloud container has local uncommitted changes that have not been uploaded.
```
tcobbs-bentley, line 873: Since the changes are never uploaded to the cloud container itself, I feel that the wording here is misleading. Perhaps replace "uploaded" with "published as a change set"?

docs/changehistory/NextVersion.md
```diff
@@ -5,6 +5,60 @@ publish: false
...
+
+#### What's new
+
+- **`LiteBriefcaseDb`** (`@alpha`): a new `BriefcaseDb` subclass designed for lightweight, agent-friendly edit sessions. The full iModel is never downloaded; blocks stream from a V2 checkpoint container on demand. Local writes stay in a small local cache and are never uploaded back to the checkpoint container. Changesets can be pushed to iModelHub normally.
+- **`LiteBriefcaseDb.openCloud(args)`** (`@alpha`): static factory method that attaches the checkpoint container, acquires (or reuses) a briefcase ID, opens the database in read-write mode against the cloud container, and sets up SAS token refresh.
+- **`OpenLiteBriefcaseArgs`** (`@alpha`): arguments interface for `openCloud`, accepting a `CheckpointProps` and an optional pre-acquired `briefcaseId`.
+- **`CloudSqlite.hasLocalChanges(container)`** (`@alpha`): helper to check whether a cloud container has uncommitted local changes that have not been uploaded.
```
tcobbs-bentley, line 17: See my comment about the `hasLocalChanges` inline documentation.

docs/changehistory/NextVersion.md
```diff
@@ -5,6 +5,60 @@ publish: false
...
+- **`V2CheckpointManager.attachForBriefcase(checkpoint)`** (`@alpha`): attaches a V2 checkpoint and returns both the `dbName` and the live `CloudContainer`, throwing if a mock container is used (mocks do not support lite briefcase mode).
+
+#### Key characteristics
+
+- **No full download**: only the blocks actually read during your edits are fetched from the cloud.
+- **Minimal disk footprint**: local storage is limited to changed blocks and the local cache — suitable for ephemeral server/agent environments.
+- **Short-lived sessions**: designed for targeted edits (seconds to minutes), not long-running or full-file access workflows.
```
tcobbs-bentley, line 24: This is awkward. All the other bold points qualify as key characteristics of the feature, while "Short-lived sessions" applies to what would be using the feature. Maybe "Intended for short-lived sessions"?

core/backend/src/IModelDb.ts
```diff
@@ -4089,8 +4089,181 @@ class RefreshV2CheckpointSas {
...
+    if (this._refreshSas) {
+      const token = accessToken ?? await IModelHost.getAccessToken();
+      await this._refreshSas.refreshSas(token, this);
+    }
+  }
+
+  /** Override close to warn if there are unpushed local changes. */
+  public override close(options?: CloseIModelArgs) {
+    if (this.isOpen && this[_nativeDb].hasPendingTxns()) {
+      Logger.logWarning(loggerCategory, "Closing LiteBriefcaseDb with unpushed local changes. These changes will be lost.");
```
tcobbs-bentley, line 4261: Maybe a comment explaining that since `LiteBriefcaseDb` is designed for use for backend server tasks, throwing an exception here would just cause problems.

core/backend/src/IModelDb.ts
khanaffan: I switch to `openFromCheckpoint` instead of `open` because if I use open I have to then also need to accept BreifcaseArgs and throw exception on if no checkpoint args is given. It look misleading so I just switch name of function.

core/backend/src/IModelDb.ts
khanaffan: fixed

khanaffan:
@copilot resolve the merge conflicts in this pull request

khanaffan:
@copilot resolve the merge conflicts in this pull request

khanaffan:
@copilot resolve the merge conflicts in this pull request

khanaffan:
@copilot resolve the merge conflicts in this pull request

PR #9205, opened 2026-04-16
Update integrityCheck to use primary connection

MichaelSwigerAtBentley:
Changes `iModelDB.integrityCheck()` function to use `usePrimaryConn: true` when querying. This prevents cases where integrityCheck would return 'no issues' if tables were corrupted but not saved.

Also fixes options keyword:
The parser does NOT treat options as the ECSQLOPTIONS keyword. So the experimental feature flag was not actually getting passed in properly. This caused some interesting false positives:

const queryWithOptions = "PRAGMA integrity_check options enable_experimental_features"; -> No corruption detected, All Tests Pass
const queryWithECSQLOPTIONS = "PRAGMA integrity_check ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES"; -> Detected Corruption, 1 Test fails

Adds a unit test that clears the cache and checks integrity

For: https://github.com/iTwin/itwinjs-backlog/issues/1811


ben-polinsky (approved):

khanaffan (approved):

MichaelSwigerAtBentley:
FYI @khanaffan I found another bug which I am fixing here while using this in OS+:

The parser does NOT treat options as the ECSQLOPTIONS keyword. So the experimental feature flag was not actually getting passed in properly. This caused some interesting false positives:

`const queryWithOptions = "PRAGMA integrity_check options enable_experimental_features"; `-> No corruption detected, All Tests Pass
`const queryWithECSQLOPTIONS = "PRAGMA integrity_check ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES";` -> Detected Corruption, 1 Test fails

aruniverse (approved):

